PR #719, opened 2014-12-23
Dossier workflow: Also allow Administrator role to activate deactivated dossiers.

lukasgraf:
Previously the transition `dossier-transition-activate` was only allowed for the `Publisher` role. This PR changes this to also allow users with the `Administrator` role to trigger this transition.

@phgross


deiferni:
:+1:

PR #725, opened 2015-01-14
Check action condition while getting tabs for tests comparisons.

deiferni:
Otherwise this leads to disabled tabs (e.g. proposals) being present. This also helps fixing tests for the `opengever.zug` package.

### DIFF
--- a/opengever/dossier/tests/test_dossier.py
+++ b/opengever/dossier/tests/test_dossier.py
@@ -5,6 +5,7 @@ from opengever.mail.behaviors import ISendableDocsContainer
 from opengever.testing import FunctionalTestCase
 from opengever.testing import index_data_for
 from Products.CMFCore.utils import getToolByName
+from Products.CMFCore.Expression import createExprContext
 
 
 class TestDossier(FunctionalTestCase):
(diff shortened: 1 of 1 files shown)


phgross:
Looks good for me .... 

Do we have other tests where the current tabs get checked? Then it would make sense do create a assertion method `assert_active_tabs` on the `FunctionalTestCase` otherwise :shipit:

deiferni:
good point - i did not find such tests with a quick search for `listActions` and `tabbedview-tabs`, so i guess not

opengever/dossier/tests/test_dossier.py
```diff
@@ -19,6 +20,16 @@ def setUp(self):
     def create_test_dossier(self):
         return create(Builder(self.builder_id))
 
+    def _get_active_tabbedview_tab_titles(self, obj):
+        types_tool = getToolByName(self.portal, 'portal_types')
+        ec = createExprContext(self.dossier, self.portal, self.dossier)
```
lukasgraf: Couldn't you use `listActionInfos()` instead (`Products.CMFCore.ActionProviderBase:89`)? This method already does the filtering for you, including checks for permissions, visibility and availability.

deiferni: awesome - did not see this method - but it does exactly what i want! updated.

PR #730, opened 2015-01-19
Make sure that an admin-unit is created before it is queried.

deiferni:
This fixes an issue while developing policies locally for customers that are still on the 4.0 release.

Also backport https://github.com/4teamwork/opengever.core/pull/722 to fix failing tests.


lukasgraf:
:+1:

PR #733, opened 2015-01-22
Move items: handle twice submitted forms or invalid selected paths.

### LINKED ISSUE #611: Move-Items: Raises key-error when submitting it twice.
When Submitting the move-items form twice or using the `backspace` key and moving a item again, the move items form raises an `KeyError`:
```
/home/zope/eggs/z3c.form-2.5.1-py2.7.egg/z3c/form/button.py:170→ __call__
/home/zope/eggs/opengever.core-3.4.1-py2.7.egg/opengever/dossier/move_items.py:80→ handle_submit
/home/zope/eggs/Zope2-2.13.18-py2.7.egg/OFS/Traversable.py:300→ unrestrictedTraverse
```
[Errbit-Error](https://errbit.4teamwork.ch/apps/524d6966383818361b00175b/problems/541c1cf9383818428100000d)

#### Proposal
The form should redirect to the current context and inform the user with a statusmessage:

phgross:
Redirects to current context and show a statusmessage, instead of raising a KeyError.
Fixes #611 

@deiferni or @lukasgraf please have a look.


opengever/dossier/move_items.py
```diff
@@ -77,30 +77,30 @@ def handle_submit(self, action):
...
-                src_name = src_object.title
-                src_id = src_object.id
+                src_name = obj.title
+                src_id = obj.id
 
                 # If parent isn't a dossier and obj is a document
                 # it's connected to a task and shouldn't be moved
-                if not IDossierMarker.providedBy(source_container) and \
-                    IDocumentSchema.providedBy(src_object):
+                if not IDossierMarker.providedBy(parent) and \
```
deiferni: *i know this was not introduced in this commit* - but: could we move this knowlegde closer to the model, i.e. the document?
maybe something like this:

```
if IDocumentSchema.providedBy(src_object) and not IDocumentSchema(src_object).is_movable():
    ...


class Document(Item):

    def is_movable(self):
        return not self.is_connected_to_task()

    def is_connected_to_task(self):
        ...
```

lukasgraf:
:+1: - but I really like the suggestion made by @deiferni, so if you still got time for it that would be great.

phgross:
@deiferni @lukasgraf I implemented dave's suggestions, please have another look ...

lukasgraf:
:+1:  :tada: :sparkles:

lukasgraf:
Unfortunately the [errbit-error](https://errbit.4teamwork.ch/apps/524d6966383818361b00175b/problems/541c1cf9383818428100000d) for this can't be ignored because the errbit ignores [only match against the exception message](https://github.com/4teamwork/errbit-python/blob/3322d5e9f3b08fd7e21a788433aa57e9ac3d8f07/errbit/client.py#L83), and `KeyError: 'dossier-42'` is much too generic to match against.

Searching for `opengever.dossier.businesscasedossier#move_items` works pretty well though, so I'll clean those messages from errbit manually.

Issue #612, opened 2014-10-09
Pasting a not allowed object_type raise ValueError.


phgross:
Example: When pasting an document to an repositoryfolder it raises the following `ValueError`
```
/home/zope/eggs/Products.PythonScripts-2.13.2-py2.7.egg/Products/PythonScripts/PythonScript.py:344→ _exec
/Script (Python):6→ object_paste
/home/zope/eggs/Zope2-2.13.18-py2.7.egg/OFS/CopySupport.py:206→ manage_pasteObjects
/home/zope/eggs/Products.CMFCore-2.2.7-py2.7.egg/Products/CMFCore/PortalFolder.py:421→ _verifyObjectPaste
Disallowed subobject type: opengever.document.document
```

[Errbit-Error](https://errbit.4teamwork.ch/apps/524d6966383818361b00175b/problems/53fdc7b5383818918600094f)

PR #734, opened 2015-01-22
Only show "Paste" action if objects in clipboard are in allowed addable types of target container.

### LINKED ISSUE #648: error when pasting object id disallowed
Instead of raising an error we should notify the user.

https://errbit.4teamwork.ch/apps/524d6966383818361b00175b/problems/544f93c6383818f6d1000098
https://errbit.4teamwork.ch/apps/534f83dc38381851ab005bcd/problems/544fc00f383818c7ee00003f

lukasgraf:
Only show the "Paste" action in the menu if all objects in clipboard are of a type that is allowed to be added to the target container. Fixes #648 and #612.

Since this check is being done pretty much on every single request, it's rather performance critical.
I therefore inlined some code from `OFS.CopySupport` because the previous `cb_dataValid()` does nothing else than trying to decode the clipboard. So if we were to call that, and then decode the clipboard again to get to the clipboard objects' types we would do it twice.

@deiferni @phgross

### DIFF
--- a/opengever/base/browser/pasting_allowed.py
+++ b/opengever/base/browser/pasting_allowed.py
@@ -1,5 +1,4 @@
 from five import grok
-from ZODB.POSException import ConflictError
 from zope.interface import Interface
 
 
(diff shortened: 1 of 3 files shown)


opengever/base/browser/pasting_allowed.py
```diff
@@ -17,16 +18,54 @@ class IsPastingAllowedView(grok.View):
...
+
+        # Try to decode the clipboard. As a side effect, this also verifies
+        # that the clipboard data is valid.
+        # For performance reasons we only decode the clipboard once, that's
+        # why there some code from CopySupport inlined here.
         try:
-            valid_cb_data = self.context.cb_dataValid()
-        except (ConflictError, KeyboardInterrupt, SystemExit):
-            raise
+            cp = _cb_decode(self.request['__cp'])
```
deiferni: I just noticed there is a method `cb_dataItems` that seems to do the same as the try/except and method inlined here. I think you could just do:

```
obj_list = self.context.cb_dataItems()
if not obj_list:
    return False
```

... and then drop ` cp = _cb_decode(self.request['__cp'])` and the inlined method. 
What do you think?

lukasgraf: As discussed: :+1:

lukasgraf:
@deiferni @phgross updated

deiferni:
:+1:

PR #740, opened 2015-01-29
Fix handling of mapped vocabulary terms in XlsSource

lukasgraf:
- Raise `KeyError` if a value can't be found in the translation mapping for fields with a vocabulary, instead of just setting the translated value anyway.
 - Add some more alternate spellings for the `archival_value` mapping.

Needs to be backported to `4.0-stable` and `4.1-stable`.


lukasgraf:
@deiferni

deiferni:
:+1:

PR #746, opened 2015-02-02
Make sure response descriptions are always unicode.

lukasgraf:
This PR makes sure that response descriptions are always returned as unicode, and all the **dynamic values in the mappings** of translated strings are unicode as well.

This fixes one bug in particular, where a document with non-ASCII characters in the title attached to a task caused the UTF-8 string from the [`Title()` accessor](https://github.com/4teamwork/opengever.core/blob/8ebcfafe23ce13a5efdcd493bc8af85770371183/opengever/task/response_description.py#L259) to be passed to the translation machinery as a dynamic value.

- Added tests for all the response descriptions except refusing forwardings and assigning forwardings to dossiers
- Modified those tests to use non-ASCII characters in as many places as possible
- Made sure response descriptions only use unicode, internally as well as for the return value
- Removed the old test at [`test_response.py`](https://github.com/4teamwork/opengever.core/blob/28387f082207687cd424a6d7ee6b1af3bc165bfb/opengever/task/tests/test_response.py) - one of them is skipped anyway, and the new tests pretty much cover what these tests were supposed to cover.

Currently I named the new tests `test_response_descriptions`, because they only do that for now: Do the absolute minimum to produce a task response and assert on it. This is because I needed some assurance that stuff still works after I modified task responses to use unicode.

But later these tests can be extended (an renamed) to actually test the task responses more in depth, and the descriptions would only be one of several assertions.

@deiferni @phgross - our Jenkins currently does not seem to test pull requests any more, but I ran the entire test suite locally -> green.


phgross:
:+1: nice!

opengever/task/tests/test_response_descriptions.py
```diff
@@ -0,0 +1,258 @@
+from datetime import datetime
+from ftw.builder import Builder
+from ftw.builder import create
+from ftw.testbrowser import browser as default_browser
```
deiferni: `ftw.testbrowser` hides the singleton browser instance by passing it to every decorated test method. Could we also pass the browser as paramter instead of using the singleton directly for our helper methods (`click_task_button`, `visit_overview` and `get_latest_answer`)?

deiferni:
:+1:

PR #747, opened 2015-02-05
Wrap checks in is_pasting_allowed in a try..except for safety.

lukasgraf:
This view is called on *every request*, and the checks may fail with a `KeyError` if an object is first copied, and then the same object is moved. Therefore we wrap all the checks in a `try..except` because this view must never fail.

### DIFF
--- a/opengever/base/browser/pasting_allowed.py
+++ b/opengever/base/browser/pasting_allowed.py
@@ -1,5 +1,6 @@
 from five import grok
 from zope.interface import Interface
+from ZODB.POSException import ConflictError
 
 
 class IsPastingAllowedView(grok.View):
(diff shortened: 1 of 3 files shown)


opengever/base/browser/pasting_allowed.py
```diff
@@ -49,3 +50,13 @@ def render(self):
...
         return True
+
+    def render(self):
+        allowed = False
+        try:
+            allowed = self.is_allowed()
+        except (ConflictError, KeyboardInterrupt, SystemExit):
+            raise
+        except Exception:
+            pass
```
deiferni: @lukasgraf this is quite unusual exception handling but - as discussed - necessary. Could you add an explanation why we do it like this?

lukasgraf:
@deiferni :+1: - updated.

deiferni:
:+1:

PR #755, opened 2015-02-15
Livesearch: Dropped not used icon <img> tag.

phgross:
Disable `plonetheme.teamraum` `livesearch.js` to avoid doubled javascript action. This fixes the bug #681.

Additionally I've dropped the not used icon `` tag. Because the livesearch use the sprite icon.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.2.0 (unreleased)
 ------------------
 
+- Livesearch: Dropped unused icon <img> tag.
+  [phgross]
+
 - Style meeting form.
   Add agendaitem sorting implementation.
   [deiferni, Kevin Bieri]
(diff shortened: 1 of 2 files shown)


phgross:
@lukasgraf, @deiferni: 

I've added the `jsregistry.xml` to the `mimetypes` profile, should I rename the `mimetypes` profile to `theme_overrides` (or did you have a better idea? :see_no_evil:)
For this reason I've don't added the upgradestep yet ...

deiferni:
@phgross looks :+1: so far.
I'm not sure if a rename is a good idea and whot to easily migrate already deployed installations. The mimetype profile is currently installed automatically https://github.com/4teamwork/opengever.core/blob/4d6e657e36ccde2c1f0018f77da419803c42e5ca/opengever/setup/deploy.py#L30. Maybe it would be best to delay renaming until we merge the other profiles  and then choose `overrides` as profile name since it can serve as a container for all necessary overrides?

phgross:
@deiferni 

I moved the disabling of the default `livesearch.js` to the `plonetheme.teamraum`package (see https://github.com/4teamwork/plonetheme.teamraum/commit/4934e5180e82c7c877dc387c1438f17a92a4c522).

Therefore this PR now only dropped the not necessary `` tag.

deiferni:
:+1:

PR #757, opened 2015-02-19
Allow submitting additional documents to an already made proposal.

deiferni:
This PR allows submitting additional documents to already made proposals. It introduces the following changes:
  - add actions/views to proposal and document/documents tabs
  - track submitted documents in a separate table
  - create entries in the tracking table on proposal submission for all attached documents
  - remove entries in the tracking table when a document is deleted

### DIFF
--- a/opengever/document/document.py
+++ b/opengever/document/document.py
@@ -7,7 +7,7 @@ from opengever.document import _
 from opengever.document.behaviors.related_docs import IRelatedDocuments
 from opengever.document.interfaces import ICheckinCheckoutManager
 from opengever.dossier.behaviors.dossier import IDossierMarker
-from opengever.meeting.model import SubmittedDocument
+from opengever.meeting.proposal import ISubmittedProposal
 from opengever.task.task import ITask
 from plone import api
 from plone.autoform import directives as form_directives
@@ -198,17 +198,15 @@ class Document(Item):
         return self.checked_out_by() is not None
 
     def is_movable(self):
-        return not self.is_inside_a_task()
+        return not self.is_inside_a_task() and not self.is_submitted_document()
 
     def is_inside_a_task(self):
         parent = aq_parent(aq_inner(self))
-        if ITask.providedBy(parent):
-            return True
-        else:
-            return False
+        return ITask.providedBy(parent)
 
     def is_submitted_document(self):
-        return SubmittedDocument.query.get_by_target(self) is not None
+        parent = aq_parent(aq_inner(self))
+        return ISubmittedProposal.providedBy(parent)
 
     def get_current_version(self):
         """Return the current document history version."""
--- a/opengever/document/document.py
+++ b/opengever/document/document.py
@@ -7,6 +7,7 @@ from opengever.document import _
 from opengever.document.behaviors.related_docs import IRelatedDocuments
 from opengever.document.interfaces import ICheckinCheckoutManager
 from opengever.dossier.behaviors.dossier import IDossierMarker
+from opengever.meeting.model import SubmittedDocument
 from opengever.task.task import ITask
 from plone import api
 from plone.autoform import directives as form_directives
@@ -205,6 +206,9 @@ class Document(Item):
             return True
         else:
             return False
+
+    def is_submitted_document(self):
+        return SubmittedDocument.query.get_by_target(self) is not None
 
     def get_current_version(self):
         """Return the current document history version."""
--- a/opengever/document/document.py
+++ b/opengever/document/document.py
@@ -205,3 +205,15 @@ class Document(Item):
             return True
         else:
             return False
+
+    def get_current_version(self):
+        """Return the current document history version."""
+
+        repository = api.portal.get_tool("portal_repository")
+        assert repository.isVersionable(self), \
+            'Document is not versionable'
+        history = repository.getHistoryMetadata(self)
+        current_version = history.getLength(countPurged=False) - 1
+        assert history.retrieve(current_version), \
+            'missing history entry for verion {}'.format(current_version)
+        return current_version
--- a/opengever/document/profiles/default/actions.xml
+++ b/opengever/document/profiles/default/actions.xml
@@ -76,6 +76,18 @@
       <property name="visible">True</property>
     </object>
 
+    <object name="submit_additional_documents" meta_type="CMF Action" i18n:domain="opengever.document">
+      <property name="title" i18n:translate="">Submit additional documents</property>
+      <property name="description" i18n:translate=""></property>
+      <property name="url_expr">string:submit_additional_documents:method</property>
+      <property name="icon_expr"></property>
+      <property name="available_expr">object/@@submit_additional_documents/available</property>
+      <property name="permissions">
+        <element value="opengever.meeting: Add Proposal"/>
+      </property>
+      <property name="visible">True</property>
+    </object>
+
 
   </object>
   <object name="object_checkin_menu" meta_type="CMF Action Category">
--- a/opengever/dossier/tests/test_templatedossier.py
+++ b/opengever/dossier/tests/test_templatedossier.py
@@ -310,7 +310,6 @@ class TestTemplateDossierListings(FunctionalTestCase):
         view = self.templatedossier.unrestrictedTraverse(DOCUMENT_TAB)
         self.assertEquals(['checkin_with_comment',
                            'checkin_without_comment',
-                           'submit_additional_documents',
                            'trashed',
                            'copy_items',
                            'zip_selected'],
--- a/opengever/dossier/tests/test_templatedossier.py
+++ b/opengever/dossier/tests/test_templatedossier.py
@@ -310,6 +310,7 @@ class TestTemplateDossierListings(FunctionalTestCase):
         view = self.templatedossier.unrestrictedTraverse(DOCUMENT_TAB)
         self.assertEquals(['checkin_with_comment',
                            'checkin_without_comment',
+                           'submit_additional_documents',
                            'trashed',
                            'copy_items',
                            'zip_selected'],
--- a/opengever/meeting/browser/documents/submit.py
+++ b/opengever/meeting/browser/documents/submit.py
@@ -1,4 +1,5 @@
 from opengever.base.source import DossierPathSourceBinder
+from opengever.base.utils import disable_edit_bar
 from opengever.document import _
 from opengever.meeting.proposal import IProposal
 from plone import api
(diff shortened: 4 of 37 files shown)


deiferni:
@lukasgraf plz have a look

opengever/meeting/command.py
```diff
@@ -21,16 +26,97 @@ def execute(self):
...
+        Transporter().transport_to(
+            self.document,
+            self.submitted_document.submitted_admin_unit_id,
+            self.submitted_document.submitted_physical_path,
+            view='update-submitted-document')
+
+        submitted_version = self.document.get_current_version()
+        submitted_document = SubmittedDocument.query.get_by_source(
+            self.proposal, self.document)
+        submitted_document.submitted_version = submitted_version
```
phgross: What happens if the `transport_to` fails ?
Is there any check in the `transport_to` method I haven't realized? If not I'd could lead to a invalid or not synched document ...

opengever/meeting/tests/test_proposal.py
```diff
@@ -181,3 +185,70 @@ def test_proposal_can_be_submitted(self, browser):
...
+        document = create(Builder('document')
+                          .within(self.dossier)
+                          .titled(u'A Document')
+                          .with_dummy_content())
+        proposal = create(Builder('proposal')
+                          .within(self.dossier)
+                          .titled(u'My Proposal')
+                          .having(committee=committee.load_model()))
+
+        self.assertFalse(proposal.is_submit_additional_documents_allowed())
```
phgross: I'd move this check in to a separate test

opengever/meeting/tests/test_proposal.py
```diff
@@ -181,3 +185,70 @@ def test_proposal_can_be_submitted(self, browser):
...
+                          .within(self.dossier)
+                          .titled(u'A Document')
+                          .with_dummy_content())
+        proposal = create(Builder('proposal')
+                          .within(self.dossier)
+                          .titled(u'My Proposal')
+                          .having(committee=committee.load_model())
+                          .relate_to(document))
+
+        proposal.execute_transition('pending-submitted')
```
phgross: Using `in_state` method when creating the proposal with the builder ...

opengever/meeting/tests/test_proposal.py
```diff
@@ -181,3 +185,70 @@ def test_proposal_can_be_submitted(self, browser):
...
+        self.assertTrue(proposal.is_submit_additional_documents_allowed())
```
phgross: If you move the allowed checks in to a separate test this check is no longer necessary

opengever/meeting/tests/test_proposal.py
```diff
@@ -181,3 +185,70 @@ def test_proposal_can_be_submitted(self, browser):
...
+        submitted_document = docs.pop()
+        self.assertEqual(0, submitted_document.get_current_version())
+
+        # create some new document versions
+        repository = api.portal.get_tool('portal_repository')
+        repository.save(document)
+        repository.save(document)
+        proposal.submit_additional_document(document)
+
+        self.assertEqual(1, submitted_document.get_current_version())
```
phgross: You create two additional versions above, but the current version is only `1` ??

opengever/dossier/tests/test_templatedossier.py
```diff
@@ -310,6 +310,7 @@ def test_enabled_actions_are_limited_in_document_tab(self):
         view = self.templatedossier.unrestrictedTraverse(DOCUMENT_TAB)
         self.assertEquals(['checkin_with_comment',
                            'checkin_without_comment',
+                           'submit_additional_documents',
```
phgross: I think the `submit_additional_documents` action makes no sense in a template dossier.

opengever/meeting/command.py
deiferni: Yes - currently we work under the assumption that only one deployment/plone-site is supported officially, but we try to prepare for multiple deployments whenever possible. So you're correct this code won't work for multiple plone-sites, but in case of one plone-site an exception is raised. (Same problem with the already merged `CopyProposalDocumentCommand`). I would suggest to  ignore this for now and implement as soon as we want to support multiple deployments. What do you think?

opengever/meeting/tests/test_proposal.py
deiferni: No this would not work, thats only for the plone `review_state` workflow. Note that this is our custom workflow implementation. Executing the transition calls the `Submit` transition implementation https://github.com/4teamwork/opengever.core/blob/master/opengever/meeting/proposal.py#L98 which copies the proposal's documents. I think this is also part of what this test validates.

Maybe we will implement `in_state` or `execute_transitions` for our custom builder at some point in the future but for now i'd suggest to call the transitions manually.

opengever/meeting/tests/test_proposal.py
deiferni: Yes thats exactly what i want to document here - the document versions run independently since these are two different document instances.

opengever/meeting/browser/documents/submit.py
```diff
@@ -0,0 +1,72 @@
...
+from z3c.relationfield.schema import RelationChoice
+
+
+class SubmittableProposalFilter(CustomFilter):
+    """Custom selectable filter to only allow selection of proposals which
+    allow submission of additional documents.
+
+    """
+    def __call__(self, brain, index_data):
+        obj = brain.getObject()
```
phgross: Is that unproblematic related to the performance ?

opengever/document/profiles/default/actions.xml
```diff
@@ -76,6 +76,18 @@
       <property name="visible">True</property>
     </object>
 
+    <object name="submit_additional_documents" meta_type="CMF Action" i18n:domain="opengever.document">
+      <property name="title" i18n:translate="">Submit additional documents</property>
+      <property name="description" i18n:translate=""></property>
+      <property name="url_expr">string:submit_additional_documents:method</property>
+      <property name="icon_expr"></property>
+      <property name="available_expr">object/@@submit_additional_documents/available</property>
```
phgross: Does the available expression also works for other objects than the expected (for example dossiers) ?

opengever/meeting/command.py
```diff
@@ -21,16 +26,97 @@ def execute(self):
...
+        self.document = document
+
+    def execute(self):
+        pass
+
+    def show_message(self):
+        portal = api.portal.get()
+        api.portal.show_message(
+            _(u'Document ${title} has already been submitted in that version',
+              mapping=dict(title=self.document.Title().decode('utf-8'))),
```
phgross: Using `self.document.title` instead of `self.document.Title()` would supersede the decoding ...

opengever/meeting/command.py
```diff
@@ -21,16 +26,97 @@ def execute(self):
...
+        submitted_version = self.document.get_current_version()
+        submitted_document = SubmittedDocument.query.get_by_source(
+            self.proposal, self.document)
+        submitted_document.submitted_version = submitted_version
+
+    def show_message(self):
+        portal = api.portal.get()
+        api.portal.show_message(
+            _(u'A new submitted version of document ${title} has been created',
+              mapping=dict(title=self.document.Title().decode('utf-8'))),
```
phgross: same as above

opengever/meeting/tests/test_submit_additional_documents.py
```diff
@@ -0,0 +1,163 @@
...
+        builder = (
+            Builder('proposal')
+            .within(self.dossier)
+            .titled(u'My Proposal')
+            .having(committee=self.committee.load_model()))
+        if attach_document:
+            builder = builder.relate_to(self.document)
+
+        proposal = create(builder)
+        proposal.execute_transition('pending-submitted')
```
phgross: Is it possible that the proposal builder supports the `in_state` method like all the other builders ?

opengever/meeting/browser/submitdocuments.py
```diff
@@ -0,0 +1,195 @@
...
+    within its context.
+
+    Returns JSON containing the created object's path and intid.
+
+    """
+    grok.name('update-submitted-document')
+    grok.require('cmf.AddPortalContent')
+    grok.context(IDocumentSchema)
+
+    def is_submitted_document(self):
```
phgross: not used the `is_submitted_document` method is only called on the document itself.

opengever/document/document.py
```diff
@@ -197,11 +198,24 @@ def is_checked_out(self):
...
-        return not self.is_inside_a_task()
+        return not self.is_inside_a_task() and not self.is_submitted_document()
 
     def is_inside_a_task(self):
         parent = aq_parent(aq_inner(self))
-        if ITask.providedBy(parent):
-            return True
-        else:
-            return False
+        return ITask.providedBy(parent)
```
phgross: Good catch :+1: :smile:

opengever/meeting/command.py
phgross: Ok, but could you add a small todo comment ...

opengever/meeting/tests/test_proposal.py
phgross: Ah now i got it. It's the copied document not the sql representation. Everything ok !

opengever/meeting/tests/test_proposal.py
deiferni: :+1:

opengever/meeting/tests/test_proposal.py
deiferni: :+1:

opengever/meeting/tests/test_submit_additional_documents.py
deiferni: yes but it would set the workflow state for the proxy object and would not call our own workflow implementation (sql-based, see above comment).

opengever/dossier/tests/test_templatedossier.py
deiferni: :+1:

opengever/meeting/browser/documents/submit.py
deiferni: Yes it is scoped by dossier and only gets objects only for all proposals in a dossier. This number should not exceed 100.

opengever/document/profiles/default/actions.xml
deiferni: yes, there are two views. one for `Interface` which handles submitting generically and one for `opengever.meeting.IProposal` which is type specific.

opengever/meeting/command.py
deiferni: :+1:

opengever/meeting/command.py
deiferni: :+1:

opengever/meeting/browser/submitdocuments.py
deiferni: :+1:

phgross:
:shipit:

opengever/meeting/command.py
deiferni: issue #728 has been updated with such a todo.

PR #758, opened 2015-02-20
Task workflow: Make workflow also manage add permissions for mails

lukasgraf:
Currently, the task workflow doesn't manage the two add permissions for mails, `ftw.mail: Add Mail` and `ftw.mail: Add Inbound Mail` at all.

This leads to a situation where mails could be added to a task in the state `task-state-tested-and-closed` by visiting `task/folder_factories`.

This PR changes the task workflow so that it manages those two add permissions for mails, and sets them the exact same way document add permissions are set for every workflow state.

The Upgrade-Step updates the corresponding workflow security.

@deiferni @phgross


opengever/task/profiles/default/workflows/opengever_task_workflow/definition.xml
```diff
@@ -53,6 +55,18 @@
    <permission-role>Owner</permission-role>
    <permission-role>Administrator</permission-role>
   </permission-map>
+  <permission-map name="ftw.mail: Add Mail" acquired="False">
+   <permission-role>Editor</permission-role>
+   <permission-role>Manager</permission-role>
+   <permission-role>Owner</permission-role>
```
phgross: I would drop `Owner` from the permission-map. 
Because we never change the owner, so it's always the creator of the task, but the `Creator`of the task is not relevant for the task permissions.

opengever/task/tests/test_workflow.py
```diff
@@ -1,11 +1,96 @@
...
+
+        browser.visit(self.task, view='++add++opengever.document.document')
+        form = browser.fill({'Title': 'foobar'})
+        with self.assertRaises(Unauthorized) as cm:
+            form.save()
+        self.assertEquals("Cannot create opengever.document.document",
+                          cm.exception.message)
+
+    @browsing
+    def test_adding_mail_in_state_tested_and_closed_not_allowed(self, browser):
```
phgross: This test is not really significant if the `test_adding_mail_in_state_open_allowed` also raise an unauthorized :see_no_evil: 

Maybe we can analyze the wrongly raised Exception next week together ...

phgross:
Aside from my comments PR looking :sparkles: :+1:

opengever/task/profiles/default/workflows/opengever_task_workflow/definition.xml
lukasgraf: I agree with the general idea, but I don't want to change existing functionality in a bugfix. And the fix here is just "Make sure mail permissions are treated exactly the same as documents" for now.

opengever/task/tests/test_workflow.py
lukasgraf: That's why I assert on the actual exception message in both those tests. It's not pretty, but it works for now as in the test failed before and passes after the fix.

phgross: ok ...

PR #759, opened 2015-02-24
Show proposal history on proposal overview.

deiferni:
Creates history entries for the following proposal-related actions:
 - creating a proposal
 - submitting a proposal
 - scheduling a proposal
 - removing a proposal from schedule
 - submit documents/additional documents
 - update submitted documents

It looks like this:

screen shot 2015-02-23 at 15 58 14

Note that the history should contain addtitional information for some actions and that icons are missing. These todos are tracked in #728.

### DIFF
--- a/opengever/meeting/model/proposalhistory.py
+++ b/opengever/meeting/model/proposalhistory.py
@@ -1,0 +1,103 @@
+from datetime import datetime
+from opengever.base.model import Base
+from opengever.meeting import _
+from opengever.ogds.base.actor import Actor
+from plone import api
+from sqlalchemy import Column
+from sqlalchemy import DateTime
+from sqlalchemy import ForeignKey
+from sqlalchemy import Integer
+from sqlalchemy import String
+from sqlalchemy.orm import relationship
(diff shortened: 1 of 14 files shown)


opengever/meeting/model/proposalhistory.py
```diff
@@ -0,0 +1,103 @@
...
+                 mapping={'user': Actor.lookup(self.userid).get_link()})
+
+
+class DocumentSubmitted(ProposalHistory):
+
+    __mapper_args__ = {'polymorphic_identity': 'documentsubmitted'}
+
+    css_class = 'addDocument'
+
+    def message(self):
```
phgross: It would be helpful to have the documents title or the link to the document included ... 

But its not urgent maybe we just :shipit: like that and adjust the messages in the future.

phgross:
Looks good for a first version

opengever/meeting/model/proposalhistory.py
deiferni: yeah as mentioned in the PR description this will be done soon, but in separate PR :wink: 
tracked in #728

PR #765, opened 2015-02-25
Use meeting icons from plonetheme.teamraum.

bierik:
Depends on 4teamwork/plonetheme.teamraum#277
Wrap submit input with span for displaying icon next to the button.
Change proposal history class for using new icon from plonetheme.teamraum.

### DIFF
--- a/opengever/meeting/model/proposalhistory.py
+++ b/opengever/meeting/model/proposalhistory.py
@@ -53,6 +53,8 @@ class Submitted(ProposalHistory):
 
     __mapper_args__ = {'polymorphic_identity': 'submitted'}
 
+    css_class = 'submitted'
+
     def message(self):
         return _(u'proposal_history_label_submitted',
                  u'Submitted by ${user}',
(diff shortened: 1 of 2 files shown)


opengever/meeting/model/proposalhistory.py
```diff
@@ -63,6 +65,8 @@ class Scheduled(ProposalHistory):
 
     __mapper_args__ = {'polymorphic_identity': 'scheduled'}
 
+    css_class = 'traktandiert'
```
deiferni: could you use an english class name? i.e. `scheduled`

opengever/meeting/model/proposalhistory.py
```diff
@@ -73,6 +77,8 @@ class RemoveScheduled(ProposalHistory):
 
     __mapper_args__ = {'polymorphic_identity': 'removescheduled'}
 
+    css_class = 'detraktandiert'
```
deiferni: same here i.e. `removeScheduled`

PR #767, opened 2015-02-25
Style protocol view.

bierik:
Made styling for protocol view.
Namespaced view with #opengever_meeting_protocol
Add two jquery plugins for autosizing textareas and sticky headings.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,10 @@ Changelog
 4.2.0 (unreleased)
 ------------------
 
+- Style protocol view in opengever.meeting.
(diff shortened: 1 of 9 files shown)


deiferni:
@bierik plz update changelog :wink:

opengever/meeting/browser/resources/flex_text.js
```diff
@@ -0,0 +1,273 @@
+/*!
```
lukasgraf: Is there a reason you're calling the file `flex_text.js` when the plugin is called `autosize`?

deiferni:
:+1: thanks!

Issue #754, opened 2015-02-13
Use plone.protect 3.x in order to get automatic CSRF protection


lukasgraf:
In order to get automatic CSRF protection for all our forms, we would need to use at least version 3.x of [`plone.protect`](https://github.com/plone/plone.protect).

`plone.protect` 2.x provides all the necessary primitives to implement CSRF protection yourself (by having a CSRF token generated, inserting it into the form and validating it in the view), but it does not yet patch any of the CRUD z3c forms.

`plone.protect` in Version 3.x however should do that, [according to the docs](https://github.com/plone/plone.protect/#automatic-csrf-protection). `plone.protect` 3.0 is what Plone 5 is supposed to be using, but at first glance it seems it should work just fine with Plone 4.x - pinning `plone.protect == 3.0.1` and `plone.keyring == 3.0.0` is all that's required to at least satisfy the version requirements:

```ini
[versions]
plone.protect = 3.0.1
plone.keyring = 3.0.0
```

However, in GEVER there is an issue where on some views, this confirmation dialog is shown:

![csrf_confirmation](https://cloud.githubusercontent.com/assets/405124/6185993/6a904e58-b374-11e4-9d13-0d910878fe7b.png)

This happens for example for the @@personal_overview, but also for most tabbed_views as far as I can tell. This dialog is caused by some code in [`plone.protect.auto.ProtectTransform._check()`](https://github.com/plone/plone.protect/blob/master/plone/protect/auto.py#L183-L189), but I haven't investigated the exact reason(s) in detail yet.

The things I'm currently suspecting to be causing, or at least be related to this problem, are:

- Redirects (301, 302)
- Writes on display
- Database modifications in GET requests

While we need to address these points in order for `plone.protect` to work properly, they aren't cool anyway and should be fixed. Redirects aren't necessary evil, but they should be used sparingly because they also affect performance (doubling round-trip time and making caching more difficult).

---

**TODO**: Provide actual examples where CSRF currently is possible in GEVER.

I don't have any concrete examples yet, but I'll construct one later in order to reproduce a case of CSRF and later verify that it's actually fixed. For now, this is how it would work on a conceptual level:

- Send an HTML email to a user you know to be currently logged into GEVER

#### `GET` requests

- Include an `<img />` element in the mail's HTML with a `src` of `http://gever/context/delete_this_thing`

#### `POST` requests
 
[...]

PR #771, opened 2015-03-03
Rewrite Redirector adapter to use cookies instead of sessions.

jone:
Accessing sessions causes the transaction to be a write, which is not very nice.
Required for #754

### DIFF
--- a/opengever/base/tests/test_redirector.py
+++ b/opengever/base/tests/test_redirector.py
@@ -1,6 +1,61 @@
 from opengever.base.interfaces import IRedirector
+from opengever.base.redirector import REDIRECTOR_COOKIE_NAME
+from opengever.base.redirector import RedirectorCookie
 from opengever.base.redirector import RedirectorViewlet
 from opengever.testing import FunctionalTestCase
+
+
+class TestRedirectorCookie(FunctionalTestCase):
+
+    def test_adding_string_to_cookie(self):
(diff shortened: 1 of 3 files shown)


jone:
@lukasgraf This one should work now :wink:

opengever/base/tests/test_redirector.py
```diff
@@ -1,7 +1,62 @@
...
+        RedirectorCookie(self.request).add({'foo': 'two'})
+        self.assertEquals([{'foo': 'one'}, {'foo': 'two'}],
+                          RedirectorCookie(self.request).read())
+        self.assertEquals([], RedirectorCookie(self.request).read())
+
+    def test_read_remove_flag(self):
+        RedirectorCookie(self.request).add({'foo': 'one'})
+        self.assertEquals([{'foo': 'one'}],
+                          RedirectorCookie(self.request).read(remove=False))
+        self.assertEquals([{'foo': 'one'}],
```
lukasgraf: Just a minor detail: This line works as an assertion and a rather implicit action to prepare for the following assertion at the same time (see [AAA](http://c2.com/cgi/wiki?ArrangeActAssert)). It might be better to split this up into two different tests.

lukasgraf:
Looks good, thank you! :+1: @phgross you also want to take a look at this?

phgross:
:+1: :sparkles:

PR #772, opened 2015-03-03
Replaced committeelisting tab with a committee overview listing

phgross:
:warning: The PR https://github.com/4teamwork/opengever.ogds.models/pull/26 has to be merged first.

Before:
bildschirmfoto 2015-03-03 um 13 16 58

After:
bildschirmfoto 2015-03-03 um 11 31 12

@deiferni, @lukasgraf  please have a look

### DIFF
--- a/opengever/meeting/browser/committeecontainertabs_templates/committee.pt
+++ b/opengever/meeting/browser/committeecontainertabs_templates/committee.pt
@@ -1,0 +1,46 @@
+<div i18n:domain="opengever.meeting" id="committees_view"
+     tal:define="committees view/committees">
+
+  <div tal:repeat="committee committees" class="committee_box">
+
+    <div class="header">
+      <a href="" tal:attributes="href committee/url">
+        <h2 i18n:translate="" tal:content="committee/title"></h2>
+      </a>
+    </div>
+
+    <div class="unscheduled_proposals"
+         tal:define="unscheduled_number committee/number_unscheduled_proposals">
+      <span i18n:translate="new_unscheduled_proposals">
+        New unschedulded proposals:
+      </span>
+      <a class="number"
+         tal:attributes="href string:${committee/url}#submittedproposals;
+                         class python: 'number unscheduled_number' if unscheduled_number else 'number'"
+         tal:content="unscheduled_number"></a>
+    </div>
+
+    <ul class="meetings" tal:define="next_meeting committee/next_meeting;
+                                     last_meeting committee/last_meeting">
+
+      <li tal:condition="last_meeting">
+        <span i18n:translate="label_last_meeting">Last Meeting:</span>
+        <a href="" tal:attributes="href last_meeting/get_url"
+           tal:content="last_meeting/get_date">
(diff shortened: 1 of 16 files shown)


opengever/meeting/service.py
```diff
@@ -16,8 +20,13 @@ def __init__(self, session):
     def _query_committee(self):
         return self.session.query(Committee)
 
-    def all_committees(self):
-        return self._query_committee().all()
+    def all_committees(self, text_filter=None):
```
deiferni: Please don't extend `meeting_service` - it will go away sooner or later. Better move these methods to the model.

opengever/meeting/service.py
```diff
@@ -34,3 +43,38 @@ def fetch_proposal(self, proposal_id):
 
     def fetch_agenda_item(self, agenda_item_id):
         return AgendaItem.query.get(agenda_item_id)
+
+    def get_next_meeting(self, committee):
```
deiferni: I'd suggest to move the query methods to https://github.com/4teamwork/opengever.core/blob/master/opengever/meeting/model/query.py in a new class `MeetingQuery`

opengever/meeting/browser/committeecontainertabs_templates/committee.pt
```diff
@@ -0,0 +1,46 @@
...
+    </div>
+
+    <div class="unscheduled_proposals"
+         tal:define="unscheduled_number committee/number_unscheduled_proposals">
+      <span i18n:translate="new_unscheduled_proposals">
+        New unschedulded proposals:
+      </span>
+      <a class="number"
+         tal:attributes="href string:${committee/url}#submittedproposals;
+                         class python: unscheduled_number and 'number unscheduled_number' or 'number'"
```
deiferni: python has a ternary expression since 2.5 so please use it: `python: 'number unscheduled_number' if unscheduled_number  else 'number'`

phgross:
@deiferni Updated and rebased ... please take another look ...

opengever/tabbedview/browser/base.py
```diff
@@ -141,6 +141,22 @@ def _public_trial_sorter(a, b):
...
+    def get_filter_text(self):
+        filter = self.request.form.get('searchable_text', None)
+
+        if not filter:
+            return []
+
+        if isinstance(filter, str):
+            filter = filter.decode('utf-8')
+
+        # remove trailing asterisk
```
deiferni: We should split first, then eliminate for every term `*`

opengever/meeting/model/query.py
```diff
@@ -75,3 +78,30 @@ def by_document(self, document):
             self._attribute('submitted_oguid') == oguid,
             self._attribute('oguid') == oguid
         ))
+
+
+class MeetingQuery(BaseQuery):
+
+    def _commite_meetings(self, committee):
```
deiferni: typo commite -> committee

deiferni:
thx! :+1:

PR #777, opened 2015-03-06
Disable content removal by default

lukasgraf:
This PR changes the rolemap so that by default no one except the `Manager` role has the `Remove GEVER content` permission.


phgross:
Looks good !

lukasgraf:
@phgross rebased

PR #779, opened 2015-03-06
Automatic CSRF protection with newest plone.protect

### LINKED ISSUE #754: Use plone.protect 3.x in order to get automatic CSRF protection

jone:
This PR updates to the newest plone.protect version, which has automatic CSRF protection.
Closes #754

How plone.protect works:
- All "write" requests are validated: a `_authenticator` request param is required.
- All `text/html` requests are parsed and for each `<form>` a hidden input field with the authenticator token is injected.
- The `X-Frame-Options` header is set, so that all contents from Opengever can no longer be included as frame in a page from another domain.
- ``request.SESSION`` should never be used when possible: it changes a lot of objects on each access, even when not changed!

**Opengever adjustements of plone.protect**
In `opengever.base.protect` the automatic CSRF protection is extended:
- A mechanism is added for whitelisting changed objects.
  - `portal_memberdata._members` is changed often and is therefore globally whitelisted for all requests.
- Patch: when there is no response content, don't try to parse it (saves us from a lot of errors in the log).
- Patch: when the response charset is set, the content is decoded (e.g. `ISO-8859-15` for some mails)

**Changes overview**

- The redirector is rewritten to use cookies instead of sessions (separate PR #771).

- Various counters and storages are whitelisted on access.
  Often all annotations are whitelisted so that we can add stuff.
  - reference number prefix adapter
  - sequence number adapter
  - OGDS sync stamp
  - tree portlet favorites
  - task response container

- Quickupload is extended / patched:
  - CSRF is disabled for JS initialization requests since the `SESSION` is accessed.
  - JS is injected for updating the quickuploader params with a token.
  - The file upload view is patched for moving the authenticator param to `request.form` (non-multipart file post).

- ``ftw.contentmenu`` is updated:
  - Transition links are required to have an `_authenticator` param since they are usually writes ( https://github.com/4teamwork/ftw.contentmenu/pull/10 )
  - The update have made the "advanced" link to appear (https://github.com/4teamwork/ftw.contentmenu/pull/5) which we remove in a customization now.

- `/edit` is always a write because of locking; CSRF is disabled for theese requests

- Write actions are extended with an authenticator:
  - Checkin / checkout actions on documents
  - Transition controller redirects
  - Journalized file downloads
  - Reference prefix manager

- dictstorage / tabbedview:
  - Enable sqlalchemy on tabebdview in order to make dictstorage not use annotations so that we don't have a change.

- Minor changes
  - testing: added logging stream handler so that it logs when the plone.protect confirmation page is rendered.
  - autocomplete widget test: CSRF disabled because the ftw.testbrowser testing layer is used.

**Developer infos**

- Always add a token to links which are supposed to make writes:
  ```python
  from plone.protect.utils import addTokenToUrl
  url = addTokenToUrl(url)
  ```
  ```xml
  <a tal:attributes="href string:${object/absolute_url}/@@view?_authenticator=${context/@@authenticator/token}" />
  ```

- Whitelist objects which may be written on GET requests but are not actual changes:
  ```python
  from opengever.base.protect import unprotected_write

  annotations = unprotected_write(IAnnotations(context))
  ```
  This example whitelists the annotations (actually the BTree of the annotations) but not its children.

- Do not use `SESSION`.

- Be aware that unexpected writes might trigger CSRF.

- When confused, open ``opengever.core.testing`` and switch the loggers to ``logging.DEBUG``.

- Do not disable CSRF whenever possible. When necessary, you can do it like this:
  ```python
  from plone.protect.interfaces import IDisableCSRFProtection
  from zope.interface import alsoProvides
  alsoProvides(self.request, IDisableCSRFProtection)
  ```

**Discussion**

- JavaScript is not tested and might have problems I didn't discover. Please help to test every AJAX request!

- Please test external tools carefully.

- *Placeholder for something important I wanted to write but forgot what it is*

**Todo**
- [x] Make Plone site creation work again.
- [x] Test external editor, probably not working.
- [x] Test mail in. *tested, works [lg]*
- [x] Test docugate (?) *the whole docugate story is on hold, will probably never make it into production [lg]*


sources.cfg
```diff
@@ -5,11 +5,13 @@ extensions = mr.developer
 development-packages =
   opengever.maintenance
   plonetheme.teamraum
+  ftw.contentmenu
 
 auto-checkout = ${buildout:development-packages}
 
 [branches]
 plone.formwidget.autocomplete = master
+ftw.contentmenu = plone-protect-support
```
jone: :construction: https://github.com/4teamwork/ftw.contentmenu/pull/10

jone:
I've fixed some other issues I encountered.
I've also tried external editor (mac version), which did work. I'm not sure though if I tested it good enough :wink:

opengever/base/subscribers.py
```diff
@@ -19,3 +23,14 @@ def create_initial_version(obj, event):
         # Create an initial version
         pr._recursiveSave(obj, {}, pr._prepareSysMetadata(comment),
             autoapply=pr.autoapply)
+
+
+@grok.subscribe(ILocking, IEditBegunEvent)
+def disable_plone_protect(obj, event):
+    """Disables plone.protect for requests beginning an edit.
+    Those requests cause the lockingbehavior to lock the content,
+    which causes a the transaction to be a write transaction.
```
deiferni: causes a the transaction -> causes the transaction

opengever/base/protect.py
```diff
@@ -0,0 +1,72 @@
...
+
+def unprotected_write(obj):
+    """Marks ``obj`` so that it does not trigger plone.protect's
+    write protection for GET request.
+    The flag is not applied recursively.
+    """
+    if obj is None:
+        return obj
+
+    # Unwrap unpersistent AttributeAnnotations
```
deiferni: unpersistent -> nonpersistent?

opengever/base/quickupload.py
```diff
@@ -19,6 +27,52 @@
...
+    We must disable plone.protect.
+    """
+
+    def upload_settings(self, *args, **kwargs):
+        alsoProvides(self.request, IDisableCSRFProtection)
+        return super(OGQuickUploadInit, self).upload_settings(*args, **kwargs)
+
+    def __call__(self, *args, **kwargs):
+        ori_js = super(OGQuickUploadInit, self).__call__(*args, **kwargs)
+        og_js = 'xhr_%(ul_id)s.setParams({"_authenticator": "%(token)s"});' % {
```
deiferni: please use `.format` here instead of the old `%` formatting

opengever/base/quickupload.py
```diff
@@ -19,6 +27,52 @@
...
+    For CSRF protection to work, we need to update the _authenticator in the form.
+    """
+
+    def quick_upload_file(self):
+        token = getDataFromAllRequests(self.request, '_authenticator')
+        if token:
+            self.request.form['_authenticator'] = token
+
+        result = super(OGQuickUploadFile, self).quick_upload_file()
+        self.request.response.setHeader('Content-Type', 'application/json; charset=utf-8')
```
deiferni: *afaik* the charset parameter is redundant since json default should always be utf-8.

*we don't do supply that param anywhere else. if it should be necessary for some reason we should also fix the rest of our code.*

opengever/base/quickupload.py
jone: This is tricky because I have JS objects (``{..}``) which conflict with `.format`.
Although it might be possible I think the old style formatting is more readable.

deiferni: Yeah but they can be escaped easily with double brackets :wink:. The new format statement would look like this, which i find pretty readable:
`'xhr_{ul_id}.setParams({{"_authenticator": "{token}"}});'.format(ul_id=self.uploader_id, token=createToken())`

deiferni:
apart from my comments looks :+1: :sparkles: 

i tested external editor on win7, this seems to work as well.

jone:
@deiferni I've eliminated the deficiencies :wink:

lukasgraf:
Great job, thanks! :+1:

lukasgraf:
One thing I noticed is that `/logout`still triggers a confirmation dialog.

jone:
I've fixed logout. I think it's ready :wink:

lukasgraf:
I customized the `confirm-action` dialog to translate it and reword it to be a bit less technical:

confirm-action

Still not very good, but it's the best I could come up with for now. Suggestions for better wording appreciated.
@phabegger @jone @deiferni @phgross

jone:
@lukasgraf I'm not sure if parameters are forwarded correctly when confirming the action. Can you confirm that it does this correctly?

lukasgraf:
@jone only query params in `GET` requests, right?

jone:
@lukasgraf yes, I think `GET` params should be passed on when confirming.
When the user has to confirm (and does) it is probably an error in the application (missing `_authenticator` token). If in this case the URL has params, the params should be passed on when confirming otherwise the URL might not work correctly.

lukasgraf:
@jone seems to work (tested with the `@@search` view which sets the `SearchableText` query param).

phgross:
@jone could you rebase and bump the next version to `4.3.0` please ...

jone:
@phgross done, untested, let jenkins finish :wink:

phgross:
@jone thx!!!

deiferni:
:tada: thank you.

PR #780, opened 2015-03-10
Automatically set predecessor's issuing_org_unit on successor.

phgross:
When accepting multiclient tasks, the `issuing_org_unit` of the successors is
wrongly be set to the current_org_unit instead of keeping the issuing_org_unit
from the predecessor. Therefore we reset the `issuing_org_unit` when setting the
predecessor with the SuccessorTaskController.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.2.0 (unreleased)
 ------------------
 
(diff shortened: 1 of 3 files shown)


phgross:
@deiferni rebased ...

deiferni:
:+1:

deiferni:
:+1:

PR #784, opened 2015-03-12
Fixed date_of_completion display in task overview.

phgross:
The `_format_date` did not format the given date, but format the tasks deadline.

@deiferni or @lukasgraf


deiferni:
ouch :astonished: - nice catch :+1:!

PR #785, opened 2015-03-13
SchemaMigration: Fix tracking table creation for DBs supporting transactional DDL

lukasgraf:
If the tracking table is created for the first time, it's not visible in `metadata` until the transaction has been committed.

Therefore we need to have a method that tries to fetch an existing tracking table and return it, or if it doesn't exist yet, creates it and then returns the created table.

Returning the table created by op.create_table requires `alembic >= 0.7.0`.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,19 @@ Changelog
 4.2.1 (unreleased)
 ------------------
 
+- SchemaMigration: Fix tracking table creation for DBs supporting transactional DDL:
+
+  If the tracking table is created for the first time, it's not visible in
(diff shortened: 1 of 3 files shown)


phgross:
Looks good :+1:

lukasgraf:
@phgross @deiferni updated PR to use a module global in order to keep a reference to the tracking table across several `SchemaMigration` upgrade steps

lukasgraf:
@phgross @deiferni updated once again: Made sure transaction is always marked as changed, even if it only contains DDL statements.

deiferni:
:+1: :sparkles:  nice, thanks!

PR #787, opened 2015-03-13
Replaced three date/time fields in favour of two datetime fields.

phgross:
Because Oracle has no Time data type, we change the Date column `date`  and the two time fields (`start_time` and `end_time`) into two DateTime columns `start` and `end`.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,7 +5,8 @@ Changelog
 4.2.1 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Drop three date/time fields in favour of two datetime fields.
+  [deiferni, phgross]
 
 
 4.2.0 (2015-03-12)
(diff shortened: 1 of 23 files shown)


opengever/meeting/model/query.py
```diff
@@ -87,14 +87,14 @@ def _committee_meetings(self, committee):
 
     def _upcoming_meetings(self, committee):
         query = self._committee_meetings(committee)
-        query = query.filter(self._attribute('date') >= date.today())
-        query = query.order_by(self._attribute('date'))
+        query = query.filter(self._attribute('start') >= date.today())
```
lukasgraf: Shouldn't this be `=> datetime.now()`? Because `start` is now a `datetime`, and `date.today()` is a `date`. And `TypeError: can't compare datetime.datetime to datetime.date`.

docs/HISTORY.txt
```diff
@@ -5,7 +5,8 @@ Changelog
 4.2.1 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Drop three date/time fields in favour of two dattime fields.
```
lukasgraf: `dattime` -> `datetime` (same in commit message)

lukasgraf:
Other than that: :+1:

PR #788, opened 2015-03-13
Remove lovely.buildouthttp from test-plone-4.2.x.cfg.

phgross:
Because the newest version of ftw.recipe.translations pins down `six` to <= `1.2.0`.
Because we don't need the `lovely.buildouthttp` for testing the package I've dropped the including.

In production buildouts we should use the `isotoma` plugin instead of `lovely.buildouthttp`.

@lukasgraf


lukasgraf:
:+1:

PR #792, opened 2015-03-17
Improve/fix schema migrations, support transactional and non-transactional DBMS.

deiferni:
Use `IdempotentOperations` only for MySQL, i.e. when no transactional DDL is available. Don't use `IdempotentOperations` any more when transactional DDL is available (e.g. for oracle and postgres). Improve `IdempotentOperations` - always update metadata when performing database changes.

*Note, currently not all `IdempotentOperations` methods are overwritten to be idempotent. This should be done on demand - we want to drop MySQL support sooner or later.*

Furthermore fix some existing migrations to work with MySQL and PostgreSQL (and Oracle). Don't use metadata tables for data migrations any more but create local table definitions using [sqlalchemy expressions](http://docs.sqlalchemy.org/en/latest/core/selectable.html).

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,25 @@ Changelog
 4.2.1 (unreleased)
 ------------------
 
+- SchemaMigration: Use IdempotentOperations only when executing migrations for MySQL.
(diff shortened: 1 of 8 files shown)


phgross:
:+1:

PR #812, opened 2015-03-23
Do not show resolved tasks as overdue.

phgross:
Added `task-state-resolved` to the overdue independent states, because already resolved tasks should not be flaged as overdue.

### DIFF
--- a/opengever/globalindex/tests/test_sql_task.py
+++ b/opengever/globalindex/tests/test_sql_task.py
@@ -101,11 +101,30 @@ class TestGlobalindexTask(TestCase):
         self.assertTrue(overdue.is_overdue)
 
     def test_is_overdue_respect_overdue_independent_states(self):
-        overdue = create(Builder('globalindex_task').having(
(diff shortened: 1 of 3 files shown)


opengever/globalindex/tests/test_sql_task.py
```diff
@@ -101,11 +101,30 @@ def test_is_overdue_compare_deadline_with_today(self):
...
+        closed_forwarding = create(Builder('globalindex_task')
+                                   .having(int_id=5, deadline=overdue_date,
+                                           task_type='forwarding',
+                                           review_state='forwarding-state-closed'))
+
+        self.assertFalse(closed.is_overdue)
+        self.assertFalse(cancelled.is_overdue)
+        self.assertFalse(resolved.is_overdue)
+        self.assertFalse(rejected.is_overdue)
+        self.assertFalse(closed_forwarding.is_overdue)
```
deiferni: could you add a test where `is_overdue` is expected to be `True`?

deiferni:
:+1:

Issue #731, opened 2015-01-21
Consider dependency order of tables during setup.


deiferni:
Issue from https://github.com/4teamwork/opengever.core/pull/724 by @lukasgraf:

I just noticed a slight annoyance (not really caused by your code, it just happens to surface here):
In [`GeverDeployment.drop_sql_tables()`](https://github.com/4teamwork/opengever.core/blob/0b6c672e53e79658898d50b1ae68a85fdfe25c4e/opengever/setup/deploy.py#L190-197) we don't consider the dependency order of the tables at all.

This can lead to foreign key constrains being violated when attempting to drop the tables:

``` pytb
2015-01-21 15:15:56 ERROR Zope.SiteErrorLog 1421849756.150.748551890844 http://localhost:8080/@@gever-create-deployment
Traceback (innermost last):
  ...
  Module opengever.setup.deploy, line 195, in drop_sql_tables
  ...
IntegrityError: (IntegrityError) (1217, 'Cannot delete or update a parent row: a foreign key constraint fails') '\nDROP TABLE proposals' ()
```

We should probably use something like [this recipe](http://docs.sqlalchemy.org/en/rel_0_7/core/schema.html?highlight=sorted_tables#reflecting-all-tables-at-once) to iterate over all **reflected** tables in **reverse dependency order**.

phgross:
Seems to be fixed with the current implementation. No problems known in recent times.

PR #813, opened 2015-03-24
Drop all SQL-tables during dev-setup.

deiferni:
This avoids problems when returning from a branch ahead of master which defines more tables. It also drops the upgrade-version tracking table.


lukasgraf:
:+1: - though #731 still needs to be addressed.

Issue #804, opened 2015-03-23
Backport all usermigration related changes to all 4.x branches


lukasgraf:
The following changes have only been backported to `4.0-stable` so far, and need to be backported to the other `4.x` branches as well:

- [x] #800 (Backports: #801 | #806 | #807 | #808)
- [x] #814 (Backports: #815 | #816 | #817 | #818)
- [x] #819 (Backports: #820 | #821 | #822 | #823)
- [x] #828 (Backports: #829 | #830 | #831 | #832)

PR #814, opened 2015-03-24
Add OGDSMigrator that migrates users and inbox groups in OGDS SQL tables

lukasgraf:
Add `OGDSMigrator` that migrates `users_group` and `inbox group` for `OrgUnits` in OGDS SQL tables (to be used with `ftw.usermigration`).

@phgross @deiferni


deiferni:
:+1:

PR #819, opened 2015-03-24
Add DictstorageMigrator that migrates user IDs in dictstorage keys (in SQL)

lukasgraf:
Add `DictstorageMigrator` that migrates user IDs in dictstorage keys in SQL (to be used with `ftw.usermigration`).

The approach is the following:

The migrator will replace the first occurence of the old user ID *from the right* if the key ends with 
`-olduserid`.

This approach will fail in the unlikely case that there exists a user ID that is a substring of another user ID with a dash (and that second user also has dictstorage entries).

For example: We have two users `user42` and `other-user42`, and `user42` gets mapped to `newuser42`. The keys of the second user will then be replaced with the possibly invalid user ID `other-newuser42`.

@deiferni @phgross

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,10 @@ Changelog
 4.3.1 (unreleased)
 ------------------
 
+- Add DictstorageMigrator that migrates user IDs in dictstorage keys (in SQL).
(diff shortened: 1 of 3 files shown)


deiferni:
:+1:

PR #828, opened 2015-03-24
Implement migrating participants in DossierMigrator.

lukasgraf:
Implement migrating participants in `DossierMigrator`.

Participations are stored as persistent `Participation` objects in the dossier's annotations. The `Participation` objects have a `contact` attribute that contains a string with a user ID of an OGDS user (or a contact, not relevant for this case).

I also sneaked in a commit that adds the `strict` keyword argument previously missing from the `OGDSMigrator`.

### DIFF
--- a/opengever/usermigration/tests/test_dossier_migrator.py
+++ b/opengever/usermigration/tests/test_dossier_migrator.py
@@ -1,6 +1,7 @@
 from ftw.builder import Builder
 from ftw.builder import create
 from opengever.dossier.behaviors.dossier import IDossier
+from opengever.dossier.behaviors.participation import IParticipationAware
 from opengever.testing import FunctionalTestCase
 from opengever.testing.helpers import obj2brain
 from opengever.usermigration.dossier import DossierMigrator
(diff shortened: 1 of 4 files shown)


opengever/usermigration/tests/test_dossier_migrator.py
```diff
@@ -54,12 +55,65 @@ def test_doesnt_raise_if_not_strict_and_user_doesnt_exist(self):
...
         results = migrator.migrate()
 
         self.assertEquals(
             [('/plone/dossier-1', 'old.user', 'new.user')],
-            results['responsible']['moved']
+            results['responsibles']['moved']
+        )
+
+
+class TestDossierMigratorForParticipants(FunctionalTestCase):
```
phgross: Adding a short tests with participations to an `opengever.contact.contact` would be good.

phgross:
Just my wish for the tests, otherwise looks :+1:

lukasgraf:
@phgross added a tests with a contact and rebased.

phgross:
:+1:

PR #833, opened 2015-03-25
Extend from 3.0.x KGS for 4.0-stable tests.

lukasgraf:


deiferni:
tests green but not reported correctly

deiferni:
:+1:

PR #835, opened 2015-03-25
Extend from 3.1.x KGS for 4.2-stable tests.

lukasgraf:


lukasgraf:
@phgross @deiferni test is green but reporting failed.

deiferni:
:+1:

PR #845, opened 2015-03-30
Customize Language Selector viewlet to display it as a menu.

phgross:
The styling is implemented in the https://github.com/4teamwork/plonetheme.teamraum/pull/286 PR.

Additionally I've added french to the supported languages in the examplecontent profile  to make sure that the language selector is activated during development.

### DIFF
--- a/opengever/base/viewlets/language_selector.pt
+++ b/opengever/base/viewlets/language_selector.pt
@@ -1,0 +1,32 @@
+<div id="portal-languageselector-wrapper"
+     tal:define="languages view/languages;
+                 here_url context/absolute_url;
+                 current view/get_current_language"
+     tal:condition="view/available">
+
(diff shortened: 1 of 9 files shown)


opengever/base/viewlets/language_selector.pt
```diff
@@ -0,0 +1,32 @@
+<div id="portal-languageselector-wrapper"
+     tal:define="languages view/languages;
+                 here_url context/absolute_url;
+                 current python:[lang for lang in languages if lang.get('selected')][0]"
```
deiferni: This could be a viewlet method

phgross: :+1:

phgross:
@deiferni updated ...

deiferni:
:+1:

PR #846, opened 2015-03-31
Fixed actor for all task activities.

phgross:
The TaskActivity class had wrongly returned objects creator as activity actor,
instead of the current user.


deiferni:
:+1:

PR #848, opened 2015-03-31
Notification mail fixes (translations and html structure)

phgross:
- Make sure that links in the activity summary are displayed correctly.
- Make sure that the activity kind is translated also in the mails. (The final implementation should configure a default language for each watcher. But for the moment (prototype) it's ok like that).


deiferni:
:+1:

PR #849, opened 2015-03-31
Add editing protocols, creating/downloading/updating protocol documents.

deiferni:
Also no longer display links to edit protocols/preprotocols in all states.


opengever/meeting/browser/meetings/meeting.py
```diff
@@ -143,6 +155,18 @@ def url_generate_pre_protocol(self):
     def url_download_pre_protocol(self):
         return DownloadGeneratedPreProtocol.url_for(self.context, self.model)
 
+    def url_protocol(self):
+        return EditPreProtocol.url_for(self.context, self.model)
```
phgross: should it not be 
```python
return EditProtocol.url_for(self.context, self.model)
```

opengever/meeting/model/meeting.py
```diff
@@ -107,6 +116,17 @@ def get_pre_protocol_filename(self):
             translate(_("Pre-Protocol"), context=getRequest()),
             normalizer.normalize(self.get_title()))
 
+    def get_protocol_title(self):
+        return u"{}-{}.docx".format(
```
phgross: The `title` should not return `docx` isn't it?

opengever/meeting/tests/test_protocol.py
```diff
@@ -0,0 +1,192 @@
...
+    def test_protocol_participants_can_be_edited(self, browser):
+        yesterday = date.today() - timedelta(days=1)
+        tomorrow = date.today() + timedelta(days=1)
+        peter = create(Builder('member'))
+        hans = create(Builder('member').having(
+            firstname=u'Hans', lastname=u'M\xfcller'))
+        create(Builder('membership').having(
+            member=peter,
+            committee=self.committee_model,
+            date_from=yesterday, date_to=tomorrow))
```
phgross: could we add a helper (`as_active`) or default values to the membership builder, which set `date_from` to yesterday and `date_to` to tomorrow?

deiferni: i like it! done.

opengever/meeting/browser/meetings/meeting.py
deiferni: nice catch :astonished:

opengever/meeting/model/meeting.py
deiferni: changed

deiferni:
@phgross updated

PR #852, opened 2015-04-07
Made DeactivatedFKConstraint compatible with newer alembic versions.

phgross:
The type parameter is now named type_ (see http://alembic.readthedocs.org/en/rel_0_6/ops.html#alembic.operations.Operations.drop_constraint.params.type_).


lukasgraf:
:+1:

PR #855, opened 2015-04-09
Document `bundler` installation in README.

lukasgraf:
Document installation of `bundler`, which is necessary for local development, in our README.

@deiferni @phgross


deiferni:
:+1: thanks!

PR #860, opened 2015-04-10
Made different SchemaMigration Upgradesteps more robust.

phgross:
I've adjusted some of our existing SchemaMigration because of problems I've run into while updating a GEVER installation from release `2.7.x` to `3.3`.

@lukasgraf please have a look.


lukasgraf:
:+1:

PR #865, opened 2015-04-13
Allow configuring sablon templates per committee.

deiferni:
Make sablon templates configurable per `Committee`. This configuration is optional. The templates defined on a `CommitteeContainer` will still be mandatory and act as fallback.

### DIFF
--- a/opengever/meeting/tests/test_excerpt.py
+++ b/opengever/meeting/tests/test_excerpt.py
@@ -1,0 +1,49 @@
+from ftw.builder import Builder
+from ftw.builder import create
+from ftw.testbrowser import browsing
+from opengever.core.testing import OPENGEVER_FUNCTIONAL_MEETING_LAYER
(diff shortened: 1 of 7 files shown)


deiferni:
@phgross plz have a look

opengever/meeting/tests/test_excerpt.py
```diff
@@ -0,0 +1,68 @@
...
+            Builder('sablontemplate')
+            .within(self.templates)
+            .with_asset_file('sablon_template.docx'))
+        container = create(Builder('committee_container').having(
+            pre_protocol_template=self.sablon_template,
+            protocol_template=self.sablon_template,
+            excerpt_template=self.sablon_template))
+
+        self.committee = create(Builder('committee').within(container))
+        self.proposal = create(Builder('proposal')
```
phgross: the following objects aren't used in any tests ...

deiferni:
@phgross updated

PR #871, opened 2015-04-16
Activate statefilter for subdossier listings.

### LINKED ISSUE #720: Activate state-filter for subdossier listings
The subdossier-listing should provide the same statefilter than the regular dossier listings  (`Dossiers` tab on a repositoryfolder) does.

phgross:
Inheriting from the Dossier tab instead of inherit from the BaseDossiersTab
class, includes the default dossier statefilter.

Fixes #720.

Feature flag is not necessary in my opinion: The idea was given the thumbs up (https://feedback.onegovgever.ch/t/abgeschlossene-subdossier-ausblenden-filterbar-machen/75)

bildschirmfoto 2015-04-16 um 07 28 01

### DIFF
--- a/opengever/tabbedview/browser/tabs.py
+++ b/opengever/tabbedview/browser/tabs.py
@@ -125,7 +125,17 @@ class Documents(OpengeverCatalogListingTab):
         ]
 
 
-class BaseDossiersTab(OpengeverCatalogListingTab):
+class Dossiers(OpengeverCatalogListingTab):
+    grok.name('tabbedview_view-dossiers')
+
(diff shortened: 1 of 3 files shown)


opengever/tabbedview/browser/tabs.py
```diff
@@ -193,7 +193,7 @@ class Dossiers(BaseDossiersTab):
     state_filter_available = True
 
 
-class SubDossiers(BaseDossiersTab):
```
deiferni: I think you could now drop the `BaseDossiersTab` tab.

deiferni:
:+1:

PR #876, opened 2015-04-17
Treeportlet: extend caching parameters with current language code.

phgross:
This allows us to display language dependent content in the Treeportlet. As we do for some customers.

@lukasgraf please have a look

### DIFF
--- a/opengever/base/utils.py
+++ b/opengever/base/utils.py
@@ -64,3 +64,11 @@ def ok_response(request=None):
 
 def disable_edit_bar():
     api.portal.get().REQUEST.set('disable_border', True)
+
+
+def get_preferred_language_code():
+    ltool = api.portal.get_tool('portal_languages')
+    language_code = ltool.getPreferredLanguage()
+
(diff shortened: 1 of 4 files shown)


opengever/base/utils.py
```diff
@@ -64,3 +64,12 @@ def ok_response(request=None):
...
 def disable_edit_bar():
     api.portal.get().REQUEST.set('disable_border', True)
+
+
+def get_preferred_language_code():
+    ltool = api.portal.get_tool('portal_languages')
+    language_code = ltool.getPreferredLanguage()
+    # special handling for combined languages
+    if len(language_code.split('-')) > 1:
+        return language_code[0]
```
lukasgraf: This should be `language_code.split('-')[0]` (or equivalent), no?

lukasgraf:
Other than that: :+1:

phgross:
@lukasgraf updated as we discussed ... 

`return language_code.split('-')[0]` works for both

lukasgraf:
:+1:

PR #883, opened 2015-04-22
Include ftw.footer and add content for the examplecontent profile.

phgross:
The ftw.footer is now included in every customer installation, it makes sense to incldue it to the core and add default content in the examplecontent profile.

This is the only means to ensure that the integration of ftw.footer works well.

Example footer (copied from dev.onegovgever.ch):
bildschirmfoto 2015-04-22 um 12 28 36


phgross:

deiferni:
:+1:

PR #885, opened 2015-04-23
Initialize annotations during object creation, to avoid CSRF messages on first object access.

phgross:
We have to initialize the annotations on every object. To avoid CSRF protection messages on first access of newly created objects, which haven't accessed the annotations during the creation process.

Because the PortletAssignments, for example `ftw.footer` portlets, are written to the objects annotations.

So if the annotations haven't been already accessed, the `AttributeAnnotations` mixin will create the `__annotations__` attribute on the object. This means that not the annotations are marked as changed but rather the object itself. And the we can't whitelist the object itself.

@deiferni @jone @lukasgraf

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,10 @@ Changelog
 4.3.1 (unreleased)
 ------------------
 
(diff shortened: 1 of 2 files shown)


lukasgraf:
:+1:

PR #889, opened 2015-04-27
Fix examplecontent repositories, make sure initial content can be created.

deiferni:
This PR changes two binary files (excel):

- Add missing repository folders to the repository in the `repository_minimal` profile
- Drop two unnecessary repository folders in the `repository_municipality` profile


deiferni:
tested and confirmed working with: 
- examplecontent opengever.core setup
- opengever.demo/gemeinde

lukasgraf:
:+1:

PR #895, opened 2015-04-28
Fix overwriting ploneform-macros

deiferni:
Currently the `ploneform-macros` view is overwritten to fix a bug with displaying status messages. Unfortunately the overwrite does not work reliably. This PR moves the overwrite to `overrides.zcml` to make sure it works correctly.

### DIFF
--- a/opengever/base/browser/configure.zcml
+++ b/opengever/base/browser/configure.zcml
@@ -9,20 +9,6 @@
           permission="zope2.View"
           />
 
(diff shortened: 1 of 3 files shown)


phgross:
Looks good for me any, @lukasgraf  any suggestions?

lukasgraf:
:+1: 

:yellow_heart: :green_heart: :purple_heart: for the test :grinning:

PR #897, opened 2015-04-29
Fix upgrades for PostgreSQL/MySQL

deiferni:
This PR addresses the following migration issues:
  - Fix opengever.policy.base upgrades, limit created tables to installed package.
  - Add idempotent MySQL implementations for creating/dropping indices and dropping tables.
  - Drop unnecessary/broken statements from migrations.
  - Drop a temporary table after the migration is finished.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,15 @@ Changelog
 4.3.1 (unreleased)
 ------------------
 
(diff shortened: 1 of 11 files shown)


phgross:
Looks good and works well.
I've tested the migration from `initial-meeting-implementation` to `deif-fix-migrations-yet-again` with MySQL and PostgreSQL.

PR #904, opened 2015-04-30
Fix missing agendaitem title in sablon templates.

deiferni:
This PR fixes missing agendaitem titles in sablon templates for agendaitems that are attached to a protocol.

It also fixes an issue with initial meeting content where an invalid template was created from non-unicode text.

*I think a changelog entry is not not necessary since recently introduced additions are fixed*


phgross:
:+1:

PR #908, opened 2015-05-04
Use clones for headings.

bierik:
The offset of the sticky headings are calculated once at startup based
on the DOM elements. When the content is getting higher, the position is
no longer valid because the offset is still referring to the
relative positioned elements. So a clone of each heading is being generated. When scrolling,
just the clone is being positioned relative. The offset is calculated
based on the static positioned original heading which has the correct
offset in the DOM.

### DIFF
--- a/opengever/meeting/browser/resources/sticky_heading.js
+++ b/opengever/meeting/browser/resources/sticky_heading.js
@@ -4,9 +4,13 @@ stickyHeading = function($){
   var onNoStickyCallback = function() {};
   var onCollisionCallback = function() {};
 
+  var clones = [];
+  var headings;
+  var doUpdate = false;
+
   var layouter = {
     reset: function(headings){
-      $.each(headings, function(i, heading) { heading.node.removeClass('sticky').css('top', 0); });
+      $.each(headings, function(i, heading) { heading.clone.removeClass('sticky').css('top', 0); });
     },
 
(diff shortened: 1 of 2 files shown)


opengever/meeting/browser/resources/sticky_heading.js
```diff
@@ -33,16 +37,27 @@ stickyHeading = function($){
...
+        el.addClass("original")
+        clone.insertAfter(el);
+        return { node: el, offset: offset, height: height, clone: clone};
+      });
+    }
+
+    function update() {
+      headings.each(function(i, e){
+        e.offset = e.node.offset().top;
+        //e.height = e.node.outerHeight();
```
phgross: :scissors:

opengever/meeting/browser/resources/sticky_heading.js
```diff
@@ -33,16 +37,26 @@ stickyHeading = function($){
...
-    function mapHeadings() {
+    function init() {
       headings = $(selector).map(function(i, e){
         var el = $(e);
-        return { node: el, offset: el.offset().top, height: el.outerHeight()};
+        var clone = el.clone();
+        var height = el.outerHeight();
+        var offset = el.offset().top;
+        clone.attr("id", el.attr("id") + "_clone");
+        el.addClass("original")
```
phgross: missing `;`

PR #911, opened 2015-05-05
Made featureflags (activity, meeting) more robust.

### LINKED ISSUE #909: Make feature flags checks more robust

phgross:
Calling the feature_flag methods (`is_activity_feature_enabled` etc.) should not fail if the corresponding registry entry is missing and only return False instead.

Fixes #909 


deiferni:
:+1:

PR #913, opened 2015-05-05
Fixed sql task synchronisation when task gets moved.

### LINKED ISSUE #802: When tasks get moved globalindex entry is not updated.
If a task gets moved, regardless of which only the task itself or the containing dossier, the task should be synched to globalindex. Because `reference_number` and `dossier_sequence_numer` should be updated.

Watched in `opengever.core` release `3.4.2`

phgross:
- Make sure that task's `reference_number` is updated in sql after moving the container: Because we can't control the order of event handlers, we have to sync all containing tasks manually, if the reference number of the dossier or subdossier has changed.
 - Reworked task's `get_dossier_sequence_number` method: Drop magic implementation using the catalog and replaced it with a simple implementation which uses acquisition to get the containing dossier and the `ISequenceNumber` adapter to get the number.

Fixes #802 

@lukasgraf please have a look ..

### DIFF
--- a/opengever/dossier/utils.py
+++ b/opengever/dossier/utils.py
@@ -1,0 +1,14 @@
+from Acquisition import aq_inner
+from Acquisition import aq_parent
+from opengever.dossier.behaviors.dossier import IDossierMarker
+from opengever.inbox.inbox import IInbox
+from Products.CMFPlone.interfaces.siteroot import IPloneSiteRoot
+
+
+def get_containing_dossier(obj):
+    while not IPloneSiteRoot.providedBy(obj):
+        if IDossierMarker.providedBy(obj) or IInbox.providedBy(obj):
+            return obj
+        obj = aq_parent(aq_inner(obj))
+
+    return None
--- a/opengever/inbox/tests/test_inbox.py
+++ b/opengever/inbox/tests/test_inbox.py
@@ -21,3 +21,8 @@ class TestInbox(FunctionalTestCase):
         inbox = create(Builder('inbox'))
 
         self.assertEqual(None, inbox.get_responsible_org_unit())
+
+    def test_get_sequence_number_returns_none(self):
+        inbox = create(Builder('inbox'))
+
+        self.assertEqual(None, inbox.get_sequence_number())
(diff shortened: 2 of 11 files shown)


lukasgraf:
@phgross hmm, tests are failing with `Unauthorized: Not authorized to access binding: context`.

At first glance it looks like there's a problem with the new meeting permissions and the events fired. Probably need to follow
https://plone.org/documentation/error/unauthorized-not-authorized-to-access-binding-context to debug this further :-/

opengever/inbox/tests/test_inbox.py
```diff
@@ -21,3 +21,8 @@ def test_get_responsible_org_unit_returns_none_when_no_org_unit_is_configured(se
         inbox = create(Builder('inbox'))
 
         self.assertEqual(None, inbox.get_responsible_org_unit())
+
+    def test_get_sequence_number_returns_none(self):
+        inbox = create(Builder('inbox'))
+
+        self.assertEqual(None, inbox.get_responsible_org_unit())
```
deiferni: this should call `inbox.get_sequence_number()`

opengever/task/task.py
```diff
@@ -318,30 +319,9 @@ def get_containing_subdossier(self):
...
-            if len(brains):
-                if brains[0].sequence_number:
-                    return brains[0].sequence_number
-                else:
-                    return None
-            else:
-                path = path[:-1]
-
-        return None
+        dossier = get_containing_dossier(self)
```
deiferni: i would also add a method `get_containing_dossier ` on `Task` and implement this in an object-oriented way.

opengever/dossier/utils.py
```diff
@@ -0,0 +1,12 @@
...
+from opengever.dossier.behaviors.dossier import IDossierMarker
+from opengever.inbox.inbox import IInbox
+from Products.CMFPlone.interfaces.siteroot import IPloneSiteRoot
+
+
+def get_containing_dossier(obj):
+    while not IPloneSiteRoot.providedBy(obj):
+        if IDossierMarker.providedBy(obj) or IInbox.providedBy(obj):
+            return obj
+        obj = aq_parent(aq_inner(obj))
```
deiferni: could you add an explicit `return None` to document what happens if we traverse up to the plone site

deiferni:
some comments, but otherwise nice work :+1:

PR #914, opened 2015-05-06
Redirect back to documents tab instead of trash tab, after trashing documents.

### LINKED ISSUE #511: Redirect after trashing a document should be to document tab.
Right now when trashing a document it redirects to the trash tab, this makes not much sense.

It should stay on the current tab, the document tab.

phgross:
Fixes #511.

Additionally I've completely reworked the trash tests, using `ftw.builder` and `ftw.testbrowser`.

@deiferni @lukasgraf

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.3.1 (unreleased)
 ------------------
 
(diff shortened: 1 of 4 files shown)


deiferni:
:+1:

PR #918, opened 2015-05-06
Set correct reference number starting point for example profiles.

### LINKED ISSUE #905: reference_prefix_starting_point is not consistent with new example repository
The new example repository introduced in #884 now is zero-based, while the previous one was one-based. But the value of `opengever.base.interfaces.IReferenceNumberSettings.reference_prefix_starting_point` still is set to `1`. This should be changed accordingly for the example content profile(s).

(The reference number formatter may have to be checked as well)

deiferni:
Fixes #905.


phgross:
:+1:

PR #919, opened 2015-05-07
Restrict allowed dossier meeting content in ConstrainTypesDecider.

### LINKED ISSUE #907: Adding meeting types should not be possible when feature is disabled
Currently, if the meeting feature flag is disabled, the meeting types only get [filtered in the menu](https://github.com/4teamwork/opengever.core/blob/bf07ab826bacd99eb9ca6cf1d8fa6f7b8b360ee8/opengever/dossier/menu.py#L16-L18), but they are still in the `allowedContentTypes`.

This fact also lead to a [change being necessary in the `opengever.ai`](https://github.com/4teamwork/opengever.ai/commit/042bc99db111c717c0f591546a7ce66375406d4c) tests which now expect proposals in the addable types, even though the feature is not yet enabled in that policy.

Instead the respective meeting types should actually not be addable if the feature is disabled. We already have a mechanism in place to check dynamically created `allowedContentTypes` constraints - the [`ConstrainTypesDecider`]( https://github.com/4teamwork/opengever.core/blob/master/opengever/dossier/base.py#L234-L269). It shouldn't be too hard to also use this for the meeting types in **Dossiers** and on the **Plone Site**.

deiferni:
Currently meeting types only get filtered in the add menu, but they are still in the allowedContentTypes and can be added through url-guessing. This PR changes this behaviour and allows or restricts meeting content in dossiers based on whether the meeting feature is enabled.

Closes #907.

### DIFF
--- a/opengever/dossier/templatedossier/__init__.py
+++ b/opengever/dossier/templatedossier/__init__.py
@@ -1,4 +1,5 @@
-from opengever.dossier.templatedossier.interfaces import ITemplateDossier
+from opengever.dossier.templatedossier.interfaces import ITemplateDossier  # keep!
+from opengever.dossier.templatedossier.templatedossier import TemplateDossier  # keep!
 from plone import api
 
 
--- a/opengever/dossier/templatedossier/templatedossier.py
+++ b/opengever/dossier/templatedossier/templatedossier.py
@@ -1,0 +1,5 @@
+from opengever.dossier.base import DossierContainer
+
+
+class TemplateDossier(DossierContainer):
+    """Base class for template dossiers."""
--- a/opengever/dossier/tests/test_dossier.py
+++ b/opengever/dossier/tests/test_dossier.py
@@ -1,11 +1,11 @@
 from ftw.builder import Builder
 from ftw.builder import create
 from ftw.contentmenu.menu import FactoriesMenu
+from opengever.core.testing import OPENGEVER_FUNCTIONAL_MEETING_LAYER
 from opengever.mail.behaviors import ISendableDocsContainer
 from opengever.testing import FunctionalTestCase
 from opengever.testing import index_data_for
 from Products.CMFCore.utils import getToolByName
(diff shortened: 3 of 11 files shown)


opengever/dossier/tests/test_dossier.py
```diff
@@ -94,3 +94,24 @@ def test_nested_subdossiers_is_not_possible_by_default(self):
...
+    layer = OPENGEVER_FUNCTIONAL_MEETING_LAYER
+
+    def setUp(self):
+        super(TestMeetingFeatureTypes, self).setUp()
+        self.dossier = create(Builder('dossier'))
+
+    def test_meeting_feature_enabled_addable_types(self):
+        self.assertEqual(
+            ['opengever.document.document', 'ftw.mail.mail',
+             'opengever.dossier.businesscasedossier', 'opengever.task.task'],
```
phgross: Hier müsste doch nun das `Proposal` addable sein, nicht?

deiferni:
@phgross nice catch, aktualisiert!

deiferni:
@phgross updated, class migration added. please double-check.

opengever/dossier/templatedossier/templatedossier.py
```diff
@@ -0,0 +1,7 @@
+from opengever.dossier.base import DossierContainer
+
+
+class TemplateDossier(DossierContainer):
+    """Base class for template dossiers."""
+
+    pass
```
lukasgraf: Not needed, docstring is enough.

opengever/dossier/templatedossier/__init__.py
```diff
@@ -1,4 +1,5 @@
-from opengever.dossier.templatedossier.interfaces import ITemplateDossier
+from opengever.dossier.templatedossier.interfaces import ITemplateDossier  # keep!
+from opengever.dossier.templatedossier.templatedossier import TemplateDossier  # keep!
```
lukasgraf: What about setting `__all__`?

PR #923, opened 2015-05-12
Drop Owner from opengever workflows.

### LINKED ISSUE #863: Owner should not be managed by any workflow
For example: 
https://github.com/4teamwork/opengever.core/blob/master/opengever/dossier/profiles/default/workflows/opengever_dossier_workflow/definition.xml#L36

Fix the following workflows:
- [x] document
- [x] contact
- [x] dossier
- [x] templatedossier
- [x] forwarding
- [x] inbox
- [x] mail
- [x] repository
- [x] repositoryroot
- [x] task

deiferni:
This PR removes the `Owner` role from all custom opengever workflows (With one exception: `Owner` still needs to be able to edit tasks as long as they are open). It also makes sure that `Manager` is granted all permissions that are granted to other roles in each workflow.

It also redefines the default roles for our functional tests to be `Contributor`, `Editor`, and `Reader`. This allows us to drop fiddling with permissions in several test-methods.

Fixes #863.


opengever/document/tests/test_forms.py
```diff
@@ -13,53 +14,55 @@ class TestDocumentIntegration(FunctionalTestCase):
...
-        self.browser.open('%s/edit' % self.document.absolute_url())
-
-        filename_display = '''<span class="named-file-widget namedblobfile-field" id="form-widgets-file">
-    <span>
-        <a class="link-overlay" href="http://nohost/plone/document-1/file_download_confirmation">test.txt</a>'''
+        browser.login().open(self.document, view='edit')
 
-        self.assertPageContains(filename_display)
+        self.assertEqual(
+            self.document.absolute_url() + '/file_download_confirmation',
```
phgross: use `'{}/file_download_confirmation'.format(url)`

opengever/document/tests/test_logout_overlay.py
```diff
@@ -1,30 +1,46 @@
...
+
+        self.dossier = create(Builder('dossier').within(self.repo_folder))
+
+        self.checkout1 = create(
+            Builder("document").titled("About Plone").within(self.dossier))
+        self.document = create(
+            Builder("document").titled("NOT checkedout").within(self.dossier))
+        self.checkout2 = create(
+            Builder("document").titled("About Python").within(self.dossier))
         getMultiAdapter(
```
phgross: The document builder support an easy way to create an already checked out documents, (https://github.com/4teamwork/opengever.core/blob/master/opengever/testing/builders/dx.py#L52)

opengever/document/tests/test_overview.py
```diff
@@ -223,11 +220,11 @@ class TestOverviewMeetingFeatures(FunctionalTestCase):
     def setUp(self):
         super(TestOverviewMeetingFeatures, self).setUp()
 
-        self.repository_root = create(Builder('repository_root'))
-        self.repository_folder = create(
-            Builder('repository').within(self.repository_root))
+        self.repo_folder_root = create(Builder('repository_root'))
```
phgross: `self.repo_folder_root` --> `self.repo_root `

opengever/document/tests/test_overview.py
```diff
@@ -223,11 +220,11 @@ class TestOverviewMeetingFeatures(FunctionalTestCase):
...
+        self.repo_folder_folder = create(
```
phgross: `self.repo_folder_folder --> self.repo_folder`

opengever/document/tests/test_overview.py
```diff
@@ -22,8 +22,13 @@ class TestDocumentOverview(FunctionalTestCase):
 
     def setUp(self):
         super(TestDocumentOverview, self).setUp()
-        self.grant('Manager')
-        self.document = create(Builder('document').with_dummy_content())
+        self.repo = create(Builder('repository_root'))
```
phgross: A lot of tests use the same initial content:
``` python
self.repo = create(Builder('repository_root'))
self.repo_folder = create(Builder('repository')
                          .within(self.repo))
self.dossier = create(Builder('dossier').within(self.repo_folder))
```
It would be nice if we could use a single builder, to create a dossier inside a repository tree. For example
``` python
dossier = create(Builder('dossier').within_repository_tree())
```

opengever/dossier/tests/test_templatedossier.py
```diff
@@ -418,6 +408,7 @@ def test_template_journal_tab(self, browser):
 
     @browsing
     def test_template_info_tab(self, browser):
+        self.grant('Manager')
```
phgross: Why is that necessary ?

opengever/document/tests/test_forms.py
deiferni: :+1:

opengever/document/tests/test_logout_overlay.py
deiferni: :+1:

opengever/document/tests/test_overview.py
deiferni: :+1:

opengever/document/tests/test_overview.py
deiferni: :smile:

opengever/document/tests/test_overview.py
deiferni: i agree, but i think we should separate dossier and repository creation to support creating multiple dossiers in the same setup. maybe something like:

```
self.repo, self.repo_folder = create(Builder('repo_tree'))
self.dossier = create(Builder('dossier').within(self.repo_folder))
```

opengever/dossier/tests/test_templatedossier.py
deiferni: We test stuff, that only `Manager` can see, https://github.com/4teamwork/opengever.core/blob/master/opengever/sharing/browser/sharing.py#L96

I've added a comment.

deiferni:
@phgross updated
@lukasgraf could you have a look as well?

lukasgraf:
:+1: - nice work! :cake:

Can you please rebase onto `master`?

PR #931, opened 2015-05-21
Ignore IObjectRemovedEvents in saveReferenceNumberPrefix handler.

phgross:
With the change of #913 we broke up the Plonesite removing, because the ObjectRemovedEvent inherit from the ObjectMovedEvent. This PR fix this problem.

@lukasgraf


lukasgraf:
:+1:

PR #933, opened 2015-05-21
Enable doc-properties for examplecontent and add example file.

deiferni:
This PR enables doc-property generation for our examplecontent profile and also adds a template file with pre-defined doc-properties.


lukasgraf:
Changelog :wink:

deiferni:
@lukasgraf updated

lukasgraf:
:+1:

Issue #957, opened 2015-05-29
Deal with reserved keywords in Oracle


lukasgraf:
I recently noticed that the `start` column of the `meeting` model is created in lowercase, while all other columns are usually `UPPERCASE` in Oracle:

![oracle_meeting](https://cloud.githubusercontent.com/assets/405124/7882529/b0085d6c-060e-11e5-9fd0-5d86d482dabc.png)

This is because of the combination of the following facts:
- Oracle distinguishes between case sensitive and case insensitive identifiers - `foo` is CI, while quoted identifiers like `"foo"` are considered CS
- CS identifiers are represented in Oracle's data dictionary in `UPPERCASE`
- SQLAlchemy however [considers all lower case identifiers](http://docs.sqlalchemy.org/en/latest/dialects/oracle.html#identifier-casing) to be CI - therefore we usually write `lowercase` in SQLAlchemy, get case insensitive behavior, and Oracle displays the names in `UPPERCASE`.
- So far so good. However, uppercase [`START` is a reserved keyword in Oracle].
- Therefore SQLAlchemy has to quote it (`"start"`) to make it a valid identifier for a column. This in turn causes Oracle to treat it as case sensitive, and store it in a different way than all our other column identifiers.

This now leads to the following problem:

A query for meetings, ordered by `start`, will cause `ORA-00936: missing expression`:

```python
>>> Meeting.query.order_by('start').all()
*** DatabaseError: (DatabaseError) ORA-00936: missing expression
 'SELECT meetings.id AS meetings_id, meetings.committee_id AS meetings_committee_id, meetings.location AS meetings_location, meetings."start" AS meetings_start, meetings.end AS meetings_end, meetings.workflow_state AS meetings_workflow_state, meetings.presidency_id AS meetings_presidency_id, meetings.secretary_id AS meetings_secretary_id, meetings.other_participants AS meetings_other_participants, meetings.pre_protocol_document_id AS meetings_pre_protocol_do_1, meetings.protocol_document_id AS meetings_protocol_document_id \nFROM meetings ORDER BY start' {}
```

The following query will however succeed (note the quoted `"start"`):
```python
>>> Meeting.query.order_by('"start"').all()
[<Meeting at "2015-05-30 00:00:00">]
```

[...]

PR #958, opened 2015-05-29
SqlTableSource: Wrap `sort_on` in a ColumnClause when extending query with ordering

lukasgraf:
This allows SQLAlchemy to properly quote column identifiers depending on the dialect (see #957).

While this fixes the immediate problem regarding ordering in table listings, we still should avoid using reserved keywords as identifiers, therefore I'm still leaving #957 open.

@phgross @deiferni


deiferni:
:+1:

PR #963, opened 2015-06-02
Configure visible language codes in language selector viewlet.

deiferni:
Plone does not display translated languages for combine language codes. This PR configures them globally on module level.

screen shot 2015-06-02 at 17 13 50

We want to move the language selector to another place anyway, so this should be only a temporary configuration.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.4.1 (unreleased)
 ------------------
 
+- Use short language codes "DE" and "FR" in language selector.
+  [deiferni]
+
 - Display a nested dossier-navigation on the dossier overview page. Display the navigation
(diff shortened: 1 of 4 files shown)


phgross:
Es gibt entsprechende Tests für den Language Selector, kannst du die anpassen so das diese language codes verwendet werden (`de-ch`, `fr-ch`)

phgross:
Und bitte auch noch die Default Spracheinstellungen anpassen (https://github.com/4teamwork/opengever.core/blob/master/opengever/examplecontent/profiles/default/portal_languages.xml#L4).

deiferni:
@phgross updated

phgross:
Changelog ?

deiferni:
@phgross changelog hinzugefügt

PR #967, opened 2015-06-03
Don't configure current_unit_id in policy templates.

deiferni:
This is done automatically by deploy.py and thus an unnecessary duplication.


phgross:
Klingt sinnvoll :+1:

PR #1016, opened 2015-06-22
Remove p.formwidget.namedfile monkeypatch

lukasgraf:
Remove `plone.formwidget.namedfile monkeypatch` and perform necessary changes for using `1.0.11` (bumped in https://github.com/4teamwork/kgs/commit/55e74b878b7a0e40dd7ce357390b28d12c021005)  which includes the fix from plone/plone.formwidget.namedfile#9

(`widget/file/file_size` now includes the `KB` suffix, which requires a few changes to our overridden templates. Also, an empty file is now reported as having a size of `1 KB` as opposed to `0 KB` before).

@phgross @deiferni

### DIFF
--- a/opengever/document/widgets/no_download_input.pt
+++ b/opengever/document/widgets/no_download_input.pt
@@ -28,7 +28,7 @@
     <span tal:condition="python: exists and action=='nochange'">
         <a class="link-overlay" tal:content="view/filename" tal:attributes="href download_url">Filename</a>
         <span class="discreet"> &mdash;
-            <span tal:define="sizekb view/file_size" tal:replace="sizekb">100</span> KB
+            <span tal:define="sizekb view/file_size" tal:replace="sizekb">100 KB</span>
         </span>
     </span>
     <div tal:condition="allow_nochange" style="padding-top: 1em;">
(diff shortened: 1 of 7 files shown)


opengever/document/widgets/no_download_input.pt
```diff
@@ -28,7 +28,7 @@
     <span tal:condition="python: exists and action=='nochange'">
         <a class="link-overlay" tal:content="view/filename" tal:attributes="href download_url">Filename</a>
         <span class="discreet"> &mdash;
-            <span tal:define="sizekb view/file_size" tal:replace="sizekb">100</span> KB
```
phgross: Ist das copy paste vom Standard-Widget? ansonsten könnte man sonst auch einfach:

``` html
 <span tal:replace="view/file_size">100</span> KB
```

phgross:
:+1:

opengever/document/widgets/no_download_input.pt
lukasgraf: @phgross yep, das ist copy & paste vom Original-Widget in `plone.formwidget.namedfile`.

deiferni:
:+1:

PR #1047, opened 2015-07-13
Fixed path limitation of the catalog query in save_reference_number_prefix subscriber.

phgross:
Additionally I've changed the method names.

@lukasgraf


lukasgraf:
:+1:

PR #1054, opened 2015-07-16
OGDSUpdater: Enforce uniqueness of user and group IDs on application side.

lukasgraf:
With this PR we enforce **uniqueness of user and group IDs** on application side.

If for example a user exists in LDAP twice, with two slightly different spellings of the supposedly unique user ID (casing, trailing whitespace), then a `UNIQUE` constraint in `SQL` may or may not consider those spellings equal, depending on the backend.

If it *doesn't* consider them equal, and allows the `INSERT` of the second user, then this leads to problems when using a case insensitive collation on the session: A `WHERE userid='john'` clause in an `UPDATE` statement will both match `'john'` and `'JOHN'`, and the update would therefore change both users even though it should only affect the one identified by the primary key value. SQLAlchemy detects this and raises

```
StaleDataError: UPDATE statement on table 'users' expected to update 1 row(s); 2 were matched.
```

Creating case-insensitive `UNIQUE` constraints in a cross-database way is non-trivial, we therefore enforce uniqueness on the application side.

Fixes 4teamwork/opengever.zug#112

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.5.0 (unreleased)
 ------------------
 
+- OGDSUpdater: Enforce uniqueness of user and group IDs on application side.
+  [lgraf]
+
 - Add internationalization support to the activity model.
 
   - Implement translatable activity columns with the help of SQLAlchemy i18n.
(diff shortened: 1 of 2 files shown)


phgross:
:+1:

PR #1058, opened 2015-07-20
OGDS Sync: Also track status of groups and group memberships

### LINKED ISSUE #713: OGDS Updater does not remove/inactivate no longer existing groups.
The group membership is synced correctly by the OGDSUpdater, but if a group is completely dropped from the LDAP, the group and also the group memberships still exists.The OGDS Updater should handle this, by removing or inactivating removed LDAP groups. 

:warning: When removing groups out of the OGDS, it should check for relations to `OrgUnits` (`users_group` or `inbox_group`)

lukasgraf:
This PR changes the OGDS import to also track the status of groups and group memberships.

Groups now have an `active` column that behaves the same way as for Users: Once a group that's present in the OGDS doesn't appear in LDAP any more, it gets set to inactive.

Also group memberships are kept up to date by clearing `Group.users` before the import of group memberships: This means that once a group disappears from LDAP, its group memberships will also be removed from the `groups_users` association table.

Additionally I dropped the `ogds_log_file` product config option, and instead use a log file with a fixed path `var/log/ogds-update.log` to log OGDS synchronization. This file will be rotated automatically (daily, 7 rotations are kept).

This PR depends on (and contains the upgrade-step for) 4teamwork/opengever.ogds.models#32

Fixes #713


docs/HISTORY.txt
```diff
@@ -5,6 +5,36 @@ Changelog
 4.5.0 (unreleased)
 ------------------
 
+- OGDS Sync: Clear memberships from groups_users association table before
```
phgross: Ich würde die ersten drei Einträge hier zusammenfassen, da die Änderungen meiner Meinung nach zusammengehören. Dies würde das Changelog übersichtlicher machen.

phgross:
Sieht gut aus :+1: , nur mein kleiner Wunsch bezüglich Changelog ...

lukasgraf:
@phgross updated

PR #1080, opened 2015-07-21
Make `author_cache_key` also discriminate on hostname

### LINKED ISSUE #1068: author_cache_key should include hostname
The `author_cache_key` in `opengever/tabbedview/helper.py` should include the hostname as a discriminator.

The current cache key leads to URLs being cached without respecting the hostname.

Steps to reproduce:

- Boot up a fresh `instance1`
- Create an ssh tunnel to the port of `instance1` and visit a Dossier's journal tab via `localhost:{port}`
- Visit the same `instance1` via its public URL.
- The `localhost:{port}` URL is now cached for the `@@user-details` link for the actors in the journal

lukasgraf:
This is required because this cache key is used to cache generated URLs, which are dependent on the hostname that is used to access the system (might be `localhost` + SSH tunnel).

Fixes #1068

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,12 @@ Changelog
 4.5.0 (unreleased)
 ------------------
 
+- Make `author_cache_key` also discriminate on hostname.
+  This is required because this cache key is used to cache generated
+  URLs, which are dependent on the hostname that is used to access the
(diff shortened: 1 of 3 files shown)


lukasgraf:
@deiferni @phgross rebased

deiferni:
:+1:

PR #1090, opened 2015-07-28
OGDS Updater: Fix some unicode errors while logging.

lukasgraf:
`userid`, `user_dn`, `groupid` and `group_dn` are converted to unicode (as soon as possible).

Therefore all message strings used in logging also need to be unicode.

@deiferni @phgross


deiferni:
:+1:

PR #1092, opened 2015-07-28
Set `coerce_to_unicode` for the CX Oracle dialect to always get unicode from SA

lukasgraf:
Make sure we always get unicode from SQLAlchemy for columns values by setting the `coerce_to_unicode` parameter for the cx_Oracle dialect. Necessary for `SQLAlchemy >= 0.9.2`

See http://docs.sqlalchemy.org/en/latest/dialects/oracle.html#unicode)


deiferni:
:+1:

PR #1113, opened 2015-07-30
Add upgrade-step to fix 'document_author' and 'title' for mails with incorrectly decoded header values.

lukasgraf:
Fixes the `document_author` and  `title` fields on the mail objects.

The upgrade-step tries to fix a possibly incorrectly decoded value in a field by

- Checking whether it looks like a raw, undecoded header
- Checking whether it is exactly the same as the corresponding raw header
  value (which means is hasn't been edited by a user)
- If both of the above are true, re-decode it by fetching it from the
  original message again using `get_header`
- If the new value is different, update it on the object, reindex the
  necessary indexes, and sync the filename to the title again if needed.

### DIFF
--- a/opengever/mail/upgrades/to4501.py
+++ b/opengever/mail/upgrades/to4501.py
@@ -1,0 +1,83 @@
+from ftw.mail.utils import ENCODED_WORD_WITHOUT_LWSP
+from ftw.mail.utils import get_header
+from ftw.upgrade import ProgressLogger
+from ftw.upgrade import UpgradeStep
+from plone import api
+import logging
+
+
+log = logging.getLogger('ftw.upgrade')
+
+
+def looks_incorrectly_decoded(text):
+    """Determine whether a field value looks like a raw header value that
+    `decode_header` failed to decode.
+    """
+    return bool(ENCODED_WORD_WITHOUT_LWSP.search(text))
+
+
+def fix_field_value(mail, field_name, header_name):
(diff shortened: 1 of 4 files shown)


opengever/mail/upgrades/to4501.py
```diff
@@ -0,0 +1,81 @@
...
+        new_value = get_header(mail.msg, header_name)
+
+        # safe_decode_headers returns UTF8, Dexterity wants unicode
+        new_value = new_value.decode('utf-8')
+
+        if value != new_value:
+            log.info("Fixing '{}' for mail {} ({} -> {})".format(
+                field_name, mail.absolute_url(), repr(value), repr(new_value)))
+            setattr(mail, field_name, new_value)
+            mail.update_filename()
```
phgross: Sollte nur gemacht werden wenn sich der Titel verändert oder ? wird aktuell auch verändert wenn sich nur der `document_author` verändert ...

lukasgraf: @phgross :+1: :+1: :+1: wollte ich eigentlich so machen, habs irgendwie übersehen

lukasgraf:
@phgross updated

deiferni:
:+1:

lukasgraf:
Affected Mails:

```
06-plone-sd-sds; 2
09-plone-gd-gds; 1
17-plone-di-dis; 5
```

PR #1135, opened 2015-08-04
Remove legacy folder and files.

deiferni:
The legacy folder contained old stuff from the time when the opengever ecosystem consistem of diverse separate packages. That's over now.

*imo no changelog entry is required*


lukasgraf:
:+1: 

snobqqe

PR #1156, opened 2015-08-10
Fix ftw.mail.inbound.createMailInContainer monkey patch

lukasgraf:
so it sets `message.contentType` to `unicode`, not `str` (see 4teamwork/ftw.mail#33).

Needs backport to `4.5-stable`


deiferni:
:+1:

PR #1166, opened 2015-08-13
Display notification tab also for regular users.

phgross:
Wrongly the `My notification` tab only gets added for users wich are allowed to watch additional tabs (inbox_group users).

@deiferni @lukasgraf 

Backport needed: `4.5-stable`

### DIFF
--- a/opengever/tabbedview/browser/personal_overview.py
+++ b/opengever/tabbedview/browser/personal_overview.py
@@ -110,12 +110,9 @@ class PersonalOverview(TabbedView):
         return tabs
 
     def get_tabs(self):
-        tabs = []
-
+        tabs = self.default_tabs + self.notification_tabs
         if self.is_user_allowed_to_view_additional_tabs():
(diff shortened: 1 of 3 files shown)


opengever/tabbedview/browser/personal_overview.py
```diff
@@ -115,7 +115,7 @@ def get_tabs(self):
         if self.is_user_allowed_to_view_additional_tabs():
             tabs = self.default_tabs + self.notification_tabs + self.admin_tabs
         else:
-            tabs = self.default_tabs
+            tabs = self.default_tabs + self.notification_tabs
```
lukasgraf: What about

```python
tabs = self.default_tabs + self.notification_tabs
if self.is_user_allowed_to_view_additional_tabs():
    tabs.extend(self.admin_tabs)
```

lukasgraf:
Other than that: :+1:

phgross:
@lukasgraf updated

PR #1168, opened 2015-08-13
Improve performance for task overviews

deiferni:
This PR improves performance for task overviews by:
- No longer inheriting from `DisplayForm`, this skips unnecessary widget preparations. Our data is from an SQL-source.
- No longer use plone transforms to `text/x-html-safe`, they are slow and not really required here since we build short markup fragments. Instead just use `escape_html`.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,11 @@ Changelog
 4.6.0 (unreleased)
 ------------------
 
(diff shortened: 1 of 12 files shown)


lukasgraf:
:+1: - but can you squash commits 1 and 3 together? Also, I'd say `no longer use text/x-html-safe transform` instead of just "plone transforms / portal transforms".

deiferni:
:+1: squashed, tests fixed

PR #1185, opened 2015-08-21
Fix autocomplete widget tests by using our own layer

lukasgraf:
Fix autocomplete widget tests by using our own layer instead of importing from `ftw.testbrowser`.

`ftw.testbrowser.testing` now imports from `plone.app.contenttypes` which we neither have or want). So we just define a form and view in `opengever.core.tests` to the our autocomplete widget against.

### DIFF
--- a/opengever/core/tests/views/z3cform.py
+++ b/opengever/core/tests/views/z3cform.py
@@ -1,0 +1,80 @@
+from plone.formwidget.autocomplete.widget import AutocompleteMultiFieldWidget
+from plone.z3cform.layout import FormWrapper
+from z3c.form.button import buttonAndHandler
+from z3c.form.field import Fields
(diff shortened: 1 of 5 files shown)


opengever/core/tests/views/z3cform.py
```diff
@@ -0,0 +1,86 @@
...
+        query_string = query_string.lower()
+        for term in self():
+            if query_string in term.title.lower():
+                yield term
+
+    def __call__(self, context=None):
+        return self
+
+
+def make_term_from_title(title):
```
phgross: Wird das wirklich noch gebraucht?

lukasgraf: Nein. Fixed :scissors:

PR #1188, opened 2015-08-25
ReferenceNumberPrefixAdpater: Fix conditions for lazily initialising PersistentMappings in annotations.

lukasgraf:
Any empty `PersistentMapping` is also falsy, therefore the previous check re-initialized an empty mapping again and again.

@phgross @deiferni


deiferni:
:+1:

PR #1194, opened 2015-08-31
Add debug view to trigger a plone.protect confirmation dialog.

lukasgraf:
Adds a debug view `@@trigger-csrf` (on Plone Site, with `ManagePortal` permission) that triggers a `plone.protect` confirmation dialog.

### DIFF
--- a/opengever/debug/browser/trigger_csrf.py
+++ b/opengever/debug/browser/trigger_csrf.py
@@ -1,0 +1,19 @@
+from five import grok
+from Products.CMFPlone.interfaces import IPloneSiteRoot
+
+
+class TriggerCSRF(grok.View):
+    """A view to trigger a plone.protect confirmation dialog.
+    """
+
+    grok.name('trigger-csrf')
+    grok.context(IPloneSiteRoot)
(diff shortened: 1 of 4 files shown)


opengever/debug/browser/trigger_csrf.py
```diff
@@ -0,0 +1,19 @@
...
+
+    grok.name('trigger-csrf')
+    grok.context(IPloneSiteRoot)
+    grok.require('cmf.ManagePortal')
+
+    def __call__(self):
+        plone = self.context
+
+        # Cause a write in a GET request
+        plone.some_fancy_attribute = True
```
deiferni: hm this will actually set `plone.some_fancy_attribute` if somebody confirms the view, right?

deiferni:
:+1:

opengever/debug/browser/trigger_csrf.py
lukasgraf: Indeed. I don't see that as a problem though, it's a debug view, and we're all consenting adults.

PR #1196, opened 2015-08-31
Fix missing journal entry due to CSRF.

### LINKED ISSUE #1195: Missing jounral entry when downloading files.
When downloading documents with the views `download_pdfpreview` or `download` no journal entry is written due to a CSRF protection issue.

deiferni:
This PR fixes an issue with missing journal entries when downloading a documents copy or its pdf-preview (if enabled).

The bug occurred due to CSRF aborting the transaction because a GET-request was writing the journal. Since the response did not contain `text/html` the content (the file) would be returned normally by `plone.protect`, the transaction however is aborted which drops the writes to the journal.

For details see [plone.protect sourcecode]
(https://github.com/plone/plone.protect/blob/3.0.1/plone/protect/auto.py#L183).

Needs backport to `3.4.3`.

Closes #1195.


lukasgraf:
:+1:

PR #1202, opened 2015-09-02
Validate task title length.

deiferni:
This prevents an ugly error being displayed.

Closes https://github.com/4teamwork/opengever.zug/issues/206


phgross:
:+1:

PR #1205, opened 2015-09-03
Fix duplicate activity recordings when a forwarding gets reassigned.

phgross:
Fixes https://github.com/4teamwork/opengever.zug/issues/200. 

With the PR #1011 the problem of doubled recorded activities gets fixed for tasks, this PR fix the Problem also for forwarding activities wich gets recorded twice.

@lukasgraf @deiferni 

Backport needed: `4.5-stable`

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.6.0 (unreleased)
 ------------------
 
+- Fix duplicate activity recordings when a forwarding gets reassigned.
+  [phgross]
+
 - Fix Upgrade-Step that updates lexicon: The affected indexes need to be
   cleared first before rebuilding them, otherwise the lexicon might not get
   updated properly.
--- a/opengever/inbox/tests/test_activities.py
+++ b/opengever/inbox/tests/test_activities.py
@@ -111,24 +111,78 @@ class TestForwardingActivites(FunctionalTestCase):
             [TEST_USER_ID],
             [watcher.user_id for watcher in self.center.get_watchers(task)])
 
-    @browsing
-    def test_reassing_forwarding_create_notifications_for_all_participants(self, browser):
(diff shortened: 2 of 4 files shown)


docs/HISTORY.txt
```diff
@@ -5,6 +5,9 @@ Changelog
 4.6.0 (unreleased)
 ------------------
 
+- Fix double activity recordings when a forwarding gets reassigned.
```
deiferni: double -> duplicate

opengever/inbox/tests/test_activities.py
```diff
@@ -110,3 +110,25 @@ def test_accepting_and_assign_forwarding_with_successor_and__updated_responsible
...
+
+        browser.login().open(forwarding)
+        browser.find('forwarding-transition-reassign').click()
+        browser.fill({'Responsible': 'peter.mueller',
+                      'Response': u'Peter k\xf6nntest du das \xfcbernehmen.'})
+        browser.find('Assign').click()
+
+        self.assertEquals(1, len(Activity.query.all()))
+        self.assertEquals(u'forwarding-transition-reassign',
+                          Activity.query.all()[0].kind)
```
deiferni: Could you add some assertions that include the watchers and explain what happens to them? Otherwise it is not really clear why the get set up.

phgross: Die  Funktionsweise der `TaskReassignActivity` wird in einem separaten Tests ausführlich getestet und dokumentiert (siehe  https://github.com/4teamwork/opengever.core/blob/master/opengever/task/tests/test_activities.py#L207).

Würdest du die Test duplizieren ?

phgross:
@deiferni @lukasgraf aktualisiert und rebased ...

deiferni:
:shipit: :+1:

PR #1210, opened 2015-09-03
Let's see if testrunner 4.4.9 still has problems.

deiferni:
:construction: We may also get a performance boost.


deiferni:
:bomb: :boom: :fire: :rotating_light: 

... uhm, yes.

lukasgraf:
:grin:

PR #1211, opened 2015-09-07
CSRF: Add view to enable and disable CSRF tracing monkey patches.

lukasgraf:
The `@@manage-csrf-tracing` view allows to enable **CSRF tracing**, which will monkey patch the method for object registration on the ZODB connection in order to **save a stack trace** at the point a persistent object has been registered (i.e. changed for the first time in a transaction).

Activating CSRF tracing will patch two methods: [`register()` on the ZODB connection](https://github.com/zopefoundation/ZODB/blob/e76c5b0234fff2ce773b875e024a32ba1e07800b/src/ZODB/Connection.py#L973-L991) and `_build_csrf_report()` on our own `OGProtectTransform`. The patch to `register()` will save a formatted stack trace to a module global every time a persistent object is modified for the first time in a transaction. This will then allow the patched `_build_csrf_report()` to include that stored stack trace in the CSRF incident log if a CSRF confirmation dialog is triggered.

csrf_tracing_inactive

Those patches can also be reverted from this view: The respective methods will then be restored to their original state.

Additionally, there is a timeout after which the patches will expire, and remove themselves the next time they're called.

csrf_tracing_active

Example stack trace that ends up in the `csrf-*.log`:

```pytb
================================================================================================
DB write to obj with OID 0x6050 from code in "opengever/debug/browser/trigger_csrf.py", line 17!
================================================================================================
  File "eggs/Zope2-2.13.21-py2.7.egg/ZServer/PubCore/ZServerPublisher.py", line 31, in __init__
    response=b)
  File "eggs/Zope2-2.13.21-py2.7.egg/ZPublisher/Publish.py", line 455, in publish_module
    environ, debug, request, response)
  File "eggs/Zope2-2.13.21-py2.7.egg/ZPublisher/Publish.py", line 249, in publish_module_standard
    response = publish(request, module_name, after_list, debug=debug)
  File "eggs/Zope2-2.13.21-py2.7.egg/ZPublisher/Publish.py", line 138, in publish
    request, bind=1)
  File "eggs/Zope2-2.13.21-py2.7.egg/ZPublisher/mapply.py", line 77, in mapply
    if debug is not None: return debug(object,args,context)
  File "eggs/Zope2-2.13.21-py2.7.egg/ZPublisher/Publish.py", line 48, in call_object
  File "opengever/debug/browser/trigger_csrf.py", line 17, in __call__
    plone.some_fancy_attribute = True
```

Needs backport to `4.5-stable`

@deiferni

### DIFF
--- a/opengever/debug/browser/manage_csrf_tracing.py
+++ b/opengever/debug/browser/manage_csrf_tracing.py
@@ -1,0 +1,244 @@
+from collections import namedtuple
+from datetime import datetime
+from datetime import timedelta
+from five import grok
+from opengever.base.protect import OGProtectTransform
+from Products.CMFPlone.interfaces import IPloneSiteRoot
+from Products.statusmessages.interfaces import IStatusMessage
+from ZODB.Connection import Connection
+from ZODB.POSException import ConflictError
+from ZODB.utils import u64
+import inspect
+import logging
+import threading
+import traceback
+
+
+log = logging.getLogger('opengever.debug')
+
+
+# References to original methods - used to restore them when "unpatching"
+_orig_register_func = Connection.register
+_orig_build_csrf_report_func = OGProtectTransform._build_csrf_report
+
+# Module global to store traceback (and OID) on DB write. This will be
+# used when building the CSRF report
+_tb_for_last_db_write = None
+
+# Monkey-patches expire after a timeout (in minutes) and remove themselves
+DEFAULT_PATCH_TIMEOUT = 60
+_patches_expire_at = None
+
+# Locks to make writing to module globals thread-safe
+expires_lock = threading.RLock()
+tb_lock = threading.RLock()
+
+
+# Lightweight object to keep a formatted traceback and associated info around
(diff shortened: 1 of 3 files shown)


opengever/debug/browser/manage_csrf_tracing.py
```diff
@@ -0,0 +1,240 @@
...
+    frame = current_frame.f_back.f_back
+
+    filename = frame.f_code.co_filename
+    line_no = frame.f_lineno
+    extracted_tb = traceback.extract_stack(frame, limit=tb_limit)
+    oid = hex(u64(obj._p_oid))
+    instruction = AnnotatedTraceback(oid, filename, line_no, extracted_tb)
+
+    # Write the traceback to the module global (in a thread-safe way)
+    with threading.RLock():
```
deiferni: As discussed, this should be instantiated globally

deiferni:
One comment, apart from that :+1: :sparkles:

lukasgraf:
@deiferni updated

PR #1220, opened 2015-09-10
Also log object oids in csrf logs.

deiferni:
This PR improves csrf logs by also logging registered object oids.


lukasgraf:
:+1:

PR #1221, opened 2015-09-15
Add logging to the task SQL synchronisation.

phgross:
The logging should help to detect problems with incorrectly synched or not even registered tasks.

@lukasgraf 

Needs-Backport: `4.5-stable`

### DIFF
--- a/opengever/globalindex/handlers/task.py
+++ b/opengever/globalindex/handlers/task.py
@@ -7,6 +7,10 @@ from Products.CMFPlone.interfaces import IPloneSiteRoot
 from zope.component import getUtility
 from zope.intid.interfaces import IIntIds
 from zope.lifecycleevent.interfaces import IObjectRemovedEvent
+import logging
+
+
+logger = logging.getLogger('opengever.globalindex')
 
 
 class TaskSqlSyncer(object):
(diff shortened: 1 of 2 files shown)


lukasgraf:
Hmm. Maybe some more detailed logging in [`get_object_id()`]( https://github.com/4teamwork/opengever.core/blob/pg_add_logging_to_task_synchronisation/opengever/globalindex/handlers/task.py#L38-L48) would be a good idea? There's at least two code paths that lead to an intid, and it would be interesting which one was taken.

opengever/globalindex/handlers/task.py
```diff
@@ -55,13 +60,20 @@ def get_sql_task(self):
 
     def sync(self):
         if self.is_uninstalling_plone():
+            logger.warn(
+                'Globalindex synchronisation of task {} skipped, '
+                'because plone site gets uninstalled.'.format(repr(self.obj)))
```
lukasgraf: You can use `{!r}.format(obj)` instead of `{}`.format(repr(obj))` to get the repr-formatting.

phgross:
@lukasgraf updated ...

opengever/globalindex/handlers/task.py
```diff
@@ -55,13 +60,20 @@ def get_sql_task(self):
...
 
         if self.obj_id is None:
+            logger.warn('Globalindex synchronisation of task {!r} skipped, '
+                        'because no intid exists.'.format(self.obj))
             return
 
         task = self.get_sql_task()
         task.sync_with(self.obj)
+        logger.info('Task {} has been successfully synchronized '
+                    'to globalindex ({!r}).'.format(self.obj, task))
```
lukasgraf: I think `self.obj` should also be repr'd, no? At least that's what you're doing above.

phgross:
@lukasgraf stimmt, hab's aktualisiert.

lukasgraf:
:+1:

PR #1224, opened 2015-09-15
Fix Upgrade-Step that updates lexicon: Affected indexes need to be cleared first

lukasgraf:
Our deferred version of the upgrade step removed the `catalog.manage_clearIndex([index_id])` [that the `plone.app.upgrade` upgrade step does](https://github.com/plone/plone.app.upgrade/blob/741f9e5996f12dff31189a0abed917b22769056a/plone/app/upgrade/v43/alphas.py#L65).

This leads to cases where the Plone Lexicon used by the `ZCTextIndexes` isn't rebuilt correctly.

Needs-Backport: `4.5-stable`

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,11 @@ Changelog
 4.6.0 (unreleased)
 ------------------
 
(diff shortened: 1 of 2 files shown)


phgross:
:+1:

PR #1227, opened 2015-09-15
CSRF tracing: Fix parameter name collision between actual and suggested timeout

lukasgraf:
I mistakenly used the same name `timeout` for both the **actual** timeout to be set (when submitting the form via `POST`) and the **suggested** timeout (query string parameter when first calling the view, just influencing the form default).

Because the query string parameter stays in the URL, they both get submitted when the form is `POST`ed, leading to Zope parsing them as a list and

```pytb
Traceback (innermost last):
  Module opengever.debug.browser.manage_csrf_tracing, line 183, in __call__
TypeError: int() argument must be a string or a number, not 'list'
```

With this change, the view should then obviously be invoked as `@@manage-csrf-tracing?suggested_timeout=700`

*(No changelog entry because the feature currently doesn't work as advertised in the existing changelog)*

Should probably be backported to `4.5-stable`.

### DIFF
--- a/opengever/debug/browser/manage_csrf_tracing.py
+++ b/opengever/debug/browser/manage_csrf_tracing.py
@@ -190,7 +190,7 @@ class ManageCSRFTracing(grok.View):
 
         # Allow for suggesting a timeout value via query string
         self._suggested_timeout = None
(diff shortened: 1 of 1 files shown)


deiferni:
:+1:

PR #1229, opened 2015-09-16
Fix creating policies from template

deiferni:
There was an error with cyclic imports during policy creation. This PR fixes that issue by moving a class to its own module from which it can be imported separately.

It also changes the `create_client` script to always pass remaining arguments to the template engine (mr.bob).

### DIFF
--- a/opengever/base/pathfinder.py
+++ b/opengever/base/pathfinder.py
@@ -1,0 +1,39 @@
+from App.config import DefaultConfiguration
+import App.config
+import os
+
+
+class PathFinder(object):
+    """Helper class to provide various Zope2 Instance related paths that are
+    otherwise cumbersome to access.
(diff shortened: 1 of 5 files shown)


deiferni:
@phgross plz have a look

opengever/base/pathfinder.py
```diff
@@ -0,0 +1,39 @@
+from App.config import DefaultConfiguration
+import App.config
+import os
+
+
+class PathFinder(object):
+    """Helper class to provide various Zope2 Instance related paths that are
+    otherwise cumbersone to access.
```
phgross: typo ...

phgross:
:+1:

deiferni:
@phgross updated

PR #1231, opened 2015-09-23
Extend logging of task SQL synchronisation, with tasks modification date.

phgross:
So it's possible to ensure, that the plone object which was used for the synchronisation (see https://github.com/4teamwork/opengever.core/blob/master/opengever/globalindex/handlers/task.py#L82) was up-to-date.

@deiferni


deiferni:
:+1:

PR #1239, opened 2015-10-05
VersionsTab: Use getHistoryMetadata() instead of getHistory() for better performance

lukasgraf:
VersionsTab: Use `getHistoryMetadata()` instead of `getHistory()` to avoid fetching the actual versioned objects and hence improve performance.

Iterating over the `LazyHistory` returned by `getHistory()` will cause a `retrieve()` for every version iterated over, causing the entire versioned object to be constructed, and therefore results in poor performance.

This change instead uses `getHistoryMetadata()`, which returns a [`ShadowHistory`](https://github.com/plone/Products.CMFEditions/blob/16391ecc2b4905cd286e448143b259f985e451f4/Products/CMFEditions/ZVCStorageTool.py#L784-L947) instance. The APIs for using `LazyHistory` and `ShadowHistory` are entirely different (`ShadowHistory` doesn't support iteration), that's why it's not just a method call that needed to be changed.

IMHO needs a backport to `4.5-stable`

@phgross @deiferni

### DIFF
--- a/opengever/document/browser/versions_tab.py
+++ b/opengever/document/browser/versions_tab.py
@@ -49,8 +49,10 @@ def translate_text(text):
 
 
 class VersionDataProxy(object):
-    """A proxy for CMFEditions `VersionData` objects (representing a single
-    version of a particular object).
+    """A proxy for CMFEditions `VersionData` dicts as returned by
+    ShadowHistory.retrieve().
+
+    This represents the metadata for a single version of a particular object.
 
     This proxy object provides easier attribute access to some of the
     VersionData's metadata, so we can use decent column names in the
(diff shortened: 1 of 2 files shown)


phgross:
:+1:

opengever/document/browser/versions_tab.py
```diff
@@ -152,25 +167,59 @@ def revert_link(self):
...
+
+        See Products.CMFEditions.browser.diff.DiffView.__call__()
+        for a more detailed example on how to use the ShadowHistory API.
+        """
+        # Order versions descending (most recent first)
+        # CMFEditions' storages call this the "selector" - it's basically the
+        # internal version number, with the low values being old versions
+        selector = self._length - index - 1
+
+        version_id = self._history.getVersionId(selector, countPurged=False)
```
deiferni: we include purged items to get the length `history.getLength(countPurged=True)` but here purged items are no longer included. Is this correct?

lukasgraf: Great catch! :catch: You're right, this is incorrect as it stands. Whether you include purged versions or not, the important thing is to use the same approach everywhere so the indexing / counts match up properly. My intention was to *not* count purged versions anywhere because that's what the [Plone `ContentHistoryViewlet` does](https://github.com/plone/plone.app.layout/blob/711b6833d2e3e43c13467025a614efde37fcb406/plone/app/layout/viewlets/content.py#L489-L492). Will fix this accordingly.

lukasgraf:
@deiferni updated

deiferni:
:ok_hand: :sparkles:

PR #1249, opened 2015-10-19
Register actor as watcher instead of every single user

phgross:
With the current implementation, the notification center has registered every single user of a inbox_group, when a task was assigned or issued by a inbox. This lead to following problems:
 - A user has been notified twice in some cases.
 - A user, which is no longer be part of the inbox_group, is still notified about this task, even if the user has no permissions.
 ...

Therefore we replaced the direct  relation between `Notification` and `Watcher` and create a Watcher for each actor instead. This means that now Watchers with the actor_id `inbox:fd` exist. 

With the described changes, the notification_center must have the ability to fetch all representatives for an actor. 
Right now the watcher model, has therefore a separate get_representatives method, which differs between users and inbox actors. This is only a temporarily solution, because the notification improvements should be available in the 4.4 release (backport). This method should use the Actor functionality, but right now it's part of the opengever.core instead of opengever.ogds.models. I'll improve that with a separate PR, which has not be backported to the 4.4 release.

@lukasgraf or @deiferni please review

### DIFF
--- a/opengever/activity/browser/resolve.py
+++ b/opengever/activity/browser/resolve.py
@@ -32,7 +32,7 @@ class ResolveNotificationView(ResolveOGUIDView):
         """Check if the current user is allowed to view the notification."""
 
         current_user = api.user.get_current()
-        return self.notification.watcher.user_id == current_user.getId()
+        return self.notification.userid == current_user.getId()
 
     def mark_as_read(self):
         self.notification.is_read = True
--- a/opengever/activity/model/watcher.py
+++ b/opengever/activity/model/watcher.py
@@ -11,8 +11,8 @@ from sqlalchemy.schema import Sequence
 
 class WatcherQuery(BaseQuery):
 
-    def get_by_userid(self, user_id):
-        return self.filter_by(user_id=user_id).first()
+    def get_by_actorid(self, actorid):
+        return self.filter_by(actorid=actorid).first()
 
 
 class Watcher(Base):
@@ -24,12 +24,12 @@ class Watcher(Base):
 
     watcher_id = Column('id', Integer, Sequence('watchers_id_seq'),
                         primary_key=True)
-    user_id = Column(String(USER_ID_LENGTH), nullable=False, unique=True)
+    actorid = Column(String(USER_ID_LENGTH), nullable=False, unique=True)
 
     resources = association_proxy('subscriptions', 'resource')
 
     def __repr__(self):
-        return '<Watcher {}>'.format(repr(self.user_id))
+        return '<Watcher {}>'.format(repr(self.actorid))
 
     def get_user_ids(self):
         """Returns a list of userids which represents the given watcher:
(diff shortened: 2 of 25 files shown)


opengever/activity/browser/resolve.py
```diff
@@ -32,7 +32,7 @@ def check_permission(self):
         """Check if the current user is allowed to view the notification."""
 
         current_user = api.user.get_current()
-        return self.notification.watcher.user_id == current_user.getId()
+        return self.notification.userid == current_user.getId()
```
lukasgraf: Shouldn't this be `actor_id`?

lukasgraf: Sorry, I actually meant `Watcher.user_id`. On notifications it's always a user, but on `Watcher` it's an actor, correct?

phgross: Ja, die Spalte sollte noch unbenannt werden. Korrigiere das noch .

lukasgraf:
Otherwise :+1:

opengever/activity/browser/resolve.py
phgross: Done

phgross:
@lukasgraf hab den PR noch um den Commit `Rename Watcher's user_id column to actorid.` ergänzt.

opengever/activity/upgrades/to4505.py
```diff
@@ -0,0 +1,73 @@
...
+from sqlalchemy import String
+from sqlalchemy.sql.expression import column
+from sqlalchemy.sql.expression import table
+
+
+USER_ID_LENGTH = 255
+
+
+class ReplaceNotificationsWatcherRelation(SchemaMigration):
+    """
```
deiferni: empty docstring, either write a docstring :wink:  or remove

opengever/activity/model/watcher.py
```diff
@@ -23,9 +24,25 @@ class Watcher(Base):
...
-    user_id = Column(String(USER_ID_LENGTH), nullable=False, unique=True)
+    actorid = Column(String(USER_ID_LENGTH), nullable=False, unique=True)
 
     resources = association_proxy('subscriptions', 'resource')
 
     def __repr__(self):
-        return '<Watcher {}>'.format(repr(self.user_id))
+        return '<Watcher {}>'.format(repr(self.actorid))
+
+    def get_representatives(self):
```
deiferni: `get_user_ids`?

opengever/activity/upgrades/to4506.py
```diff
@@ -0,0 +1,46 @@
...
+from sqlalchemy import Column
+from sqlalchemy import String
+from sqlalchemy.sql.expression import column
+from sqlalchemy.sql.expression import table
+
+
+USER_ID_LENGTH = 255
+
+
+class RenameUserIdColumn(SchemaMigration):
```
deiferni: what about `op.alter_column(..., new_column_name='actorid', ...)`?

Are there problems with such a rename on a specific backend?

phgross: Hmm doch, verwenden wir bereits in anderen upgradesteps, ich habe den Step entsprechend angepasst.

phgross:
@deiferni updated

deiferni:
:+1:

PR #1252, opened 2015-10-19
Merge changes made on minimal repository configuration to the municipality one.

phgross:
Some of the changed positions, are used by the examplecontent (see https://github.com/4teamwork/opengever.core/blob/af6b17f0cdedaa7c0115d70a3aee921056f7ac30/opengever/examplecontent/profiles/municipality_content/opengever_content/04_committees.json).

This change should fix the currently failing nightly build.

@deiferni


deiferni:
:+1:

PR #1261, opened 2015-10-21
Remove leftover related_items catalog metadata.

deiferni:
Index and column in catalog.xml has already been removed in an upgrade to
2602 - but not the actual metadata column in the catalog.

*For most customers this update is not critical and can be done eventually, in critical cases an update will be executed manually.*


deiferni:
@phgross plz have a look

PR #1268, opened 2015-10-22
Moved Upgradestep which ReRegister the Watchers after all the schemamigrations

phgross:
To avoid problems with difference between SQLAlchemy model definition and the DB, we moved the upgrade step which reregisters the watchers for each existing task to the bottom.

It doesn't make sense to copy the SQLAlchemy Model logic into the upgradestep, since it depends on the plone content type (task, forwarding), which cannot be versioned in such a way.

The Release 4.5.7 (which will contain all the activity changes) will be deployed soon.

@deiferni


deiferni:
:+1:

PR #1269, opened 2015-10-23
Notification dropdown fixes

phgross:
When deploying on the DEV system I found some issues on the reworked notification dropdown.
 - Fix javascript for sites with disabled activity feature.
 - Added missing javascript registration in the default profile of `opengever.base`
   The PR #1268 only adds the handlebar.js to the upgrade profile. So newly created GEVER sites will fail with javascript errors.
 - Fixed compression type for the handlebars.js, we already use the minified javascript, so no compression is needed anymore.

@lukasgraf 

Needs-backport: `4.5-stable`

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.6.0 (unreleased)
 ------------------
 
+- Notification viewlet: Fix javascript for sites with disabled activity feature.
+  [phgross]
+
 - Rework notification viewlet:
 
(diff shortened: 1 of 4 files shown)


lukasgraf:
:+1:

lukasgraf:
@phgross `handlebars.js` is still being registered in https://github.com/4teamwork/opengever.core/blob/master/opengever/activity/profiles/default/jsregistry.xml :frowning:

PR #1270, opened 2015-10-23
Fixed issue in Tasks `by_container` query, with similar dossier-ids.

phgross:
If a two dossier exist inside the same repository folder, on with the id `dossier-22` and the other with the id `dossier-222`. The task from the Dossier-222 are wrongly listed in the task listing of Dossier-22.

This change will fix that issue.

@lukasgraf 

I would say it makes sense to backport this fix to the `4.5-stable` ...


lukasgraf:
:+1:

PR #1271, opened 2015-10-23
Removed wrong and outdated handlebars.js registration from the activity jsregistry.

phgross:
The handlebars was moved to opengever.base, but I've not cleanely removed the javascript registration entry from the activity jsregistry.xml.

Backport: `4.5-stable`

@lukasgraf


lukasgraf:
:+1: - please rebase

PR #1279, opened 2015-10-30
Reassign: Make sure only the new responsible gets notified by mail.

phgross:
Make sure only the new responsible gets notified by mail, when a task gets reassigned.
To make sure that the old responsible is notified via batch icon, but not via mail, we change the role to `old_responsible` before record the activity, after notification creation, we remove the watcher definitely.

Theoretical it would also be possible to change the configuration, in a way that the old responsible gets also be notified by mail.

@lukasgraf please have a look.

Needs-Backport: `4.5-stable`

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,10 @@ Changelog
 4.6.0 (unreleased)
 ------------------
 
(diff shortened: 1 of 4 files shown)


lukasgraf:
Looks good to me! :+1:

PR #1284, opened 2015-11-03
Fixed NotificationDefault roles initialization in 4502 upgradestep.

phgross:
The current upgradestep ignored the settings without any configured roles, therefore this settings were initialized with Null instead of an empty json list `'[]'`. This change fix the ugpradestep retrospective, because the upgradestep is not released or deployed yet.

Additionally I've added the missing setting for the `forwarding-added` transition. An upgradestep is not needed because the setting was added manually in all Deployments.

@lukasgraf @deiferni 

Backport: `4.5-stable`


deiferni:
looks :+1:

lukasgraf:
:+1: :wink:

PR #1285, opened 2015-11-03
Include plone4.csrffixes and drop some of our custom write-on-read handling

lukasgraf:
This PR includes `plone4.csrffixes` in `opengever.core`.

This allows us to drop some of our custom handling of writes-on-read:

- Transforming responses with uncommon charsets
- Our own way of marking objects for safe writes
- Unprotecting writes to _dav_writelocks
- Handling context portlet assignments

It also should lead to *much* fewer false positive confirmation dialogs because of the referrer check included in `plone4.csrffixes`.

@phgross @deiferni


sources.cfg
```diff
@@ -18,3 +19,4 @@ auto-checkout = ${buildout:development-packages}
 
 [sources]
 plone.formwidget.autocomplete = git git@git.4teamwork.ch:opengever/plone.formwidget.autocomplete.git  branch=${branches:plone.formwidget.autocomplete}
+plone.protect = git git@github.com:plone/plone.protect.git
```
deiferni: This should not be necessary, plone.protect source information is available from [plonesource.org](http://plonesource.org/sources.cfg).

versions.cfg
```diff
@@ -18,3 +18,9 @@ zptlint = 0.2.4
 # which causes test isolation problems with PloneTestCase layers
 # which are not isolating properly.
 zope.testrunner = 4.4.4
+
+# Version bumps needed for plone4.csrffixes
+# Move to develop KGS once merged
+plone.keyring = 3.0.1
```
deiferni: could you move this to our dev kgs before the merge, that would be a bit easier ... or does it break current master?

deiferni:
looks :+1: !

sources.cfg
lukasgraf: Good point, updated. :+1:

versions.cfg
lukasgraf: Chances are it will break `master` - can't say for sure.

deiferni:
:ship:

PR #1288, opened 2015-11-03
Fixed destination version of the upgradestep which adds the MessageFactory JS

phgross:
The PR #1245 adds a new javascript `MessageFactory.js` including a upgradestep which register the new js in the registry. But the destination version for the upgradestep was wrong `4503` instead of `46xx` (see [configure.zcml](https://github.com/4teamwork/opengever.core/blob/21d31031c8cca117809f08c1561d73aa06f873be/opengever/base/upgrades/configure.zcml#L315-L322)). 

Because of backporting some Upgradesteps to 4.5.x releases, it's very important that Upgradestep is replaced with a correct destination version 4600, because the changes from #1245 are made for the 4.6 release.

@lukasgraf @deiferni


opengever/base/upgrades/configure.zcml
```diff
@@ -312,15 +312,6 @@
...
-    <!-- 4502 -> 4503 -->
-    <upgrade-step:importProfile
-        title="Regsiter MessageFactory javascript"
-        profile="opengever.base:default"
-        source="4502"
-        destination="4503"
-        directory="profiles/4503"
-        />
-
     <!-- 4503 -> 4504 -->
```
deiferni: Now there is a gap between 4502 and 4504, you should also adapt the source version of the following upgrade-step (4504) to 4502.

phgross: You're right, I've changed the source version.

deiferni:
:+1:

PR #1294, opened 2015-11-04
Lock protocol documents while a meeting is not closed.

### LINKED ISSUE #1061: Protokoll Dokument sperren solange Sitzung nicht abgeschlossen.
GEVER Dokument zur Bearbeitung sperren, bis die Sitzung abgeschlossen ist. 

Beim Dokument sollte ein entsprechender Hinweis dargestellt werden.

deiferni:
This PR lock protocol documents while a meeting is not closed to avoid that users get confused when editing the protocol document and the protocol through the web in parallel.

It takes a rather minimalistic approach with `plone.locking`. This has the advantage that we don't have to take further action on the document side (info message, disable actions like external editor, ...).
*The info message might need some customizing in the future though, i.e. show why the document is locked.*

Closes #1061.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.6.0 (unreleased)
 ------------------
 
+- Lock protocol documents while a meeting is not closed.
+  [deiferni]
+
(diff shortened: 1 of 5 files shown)


phgross:
:+1:

PR #1307, opened 2015-11-05
Eliminate duplicate notification when creating a successor after a task has been accepted.

phgross:
The problem is only obvious, when a task is assigned to a inbox. Then all inbox members, except the one who accept the task and creates the successor, are notified twice. 

@lukasgraf @deiferni 

Backport: `4.5-stable`

Note: unfortunately writing a test for this two AdminUnit Request problem, is not simple possible.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.6.0 (unreleased)
 ------------------
 
+- Eliminate duplicate notification, about a task acceptance, when creating a successor.
+  [phgross]
+
 - Lock protocol documents while a meeting is not closed.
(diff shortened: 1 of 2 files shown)


docs/HISTORY.txt
```diff
@@ -5,6 +5,9 @@ Changelog
 4.6.0 (unreleased)
 ------------------
 
+- Eliminate doubled notificaiton, about a task acceptance, when creating a successor.
```
deiferni: doubled notificaiton -> duplicate notification

deiferni:
looks :+1:

phgross:
@deiferni updated

PR #1322, opened 2015-11-09
Agendaitem table row width is not equal width original while dragging.

bierik:
Fixes https://github.com/4teamwork/opengever.core/issues/1299

Depends on https://github.com/4teamwork/plonetheme.teamraum/pull/371

bildschirmfoto 2015-11-09 um 17 12 32


phgross:
:+1:

PR #1324, opened 2015-11-10
Implement default error message for javascript generated messages.

bierik:


opengever/base/browser/templates/gever-macros.pt
```diff
@@ -104,5 +104,12 @@
 
     </metal:define>
 
+    <metal:define define-macro="js_default_error">
```
deiferni: js_default_error -> js_default_error_messages

bierik:
@deiferni thx... updated

phgross:
OK.

PR #1334, opened 2015-11-11
Mark destructive buttons.

bierik:
Depends on https://github.com/4teamwork/plonetheme.teamraum/pull/375


bierik:
@phgross updated

PR #1341, opened 2015-11-12
Javascript throws error when selecting a paragraph in the scrollspy.

bierik:
Scroll to the paragraph when selecting them in the navigation.


deiferni:
:+1:

PR #1348, opened 2015-11-12
Refactor blocking portlet inheritance, for easier inclusion in policies.

deiferni:
I'd like to include this in policies, similar to `assign_default_navigation_portlet`.


phgross:
:+1:

PR #1349, opened 2015-11-13
Safari does not set the left offset automatically on fixed elements.

bierik:
Fixes https://github.com/4teamwork/plonetheme.teamraum/issues/381

Depends on https://github.com/4teamwork/plonetheme.teamraum/pull/386


deiferni:
:+1:

PR #1369, opened 2015-11-18
Controller action callback does not create the required target.

bierik:
```event.target``` always refers to the dispatched object. So use ```event.currentTarget``` which refers to the event handler.


deiferni:
:+1:

PR #1376, opened 2015-11-20
Add last modified column to meeting.

deiferni:
Required by #1359.

*The column cannot be updated with onupdate since changes usually occur elsewehere, i.e. in an agenda item.*


deiferni:
@phgross plz have a look

phgross:
:+1:

PR #1386, opened 2015-11-23
Javascript is broken when development mode is activated.

bierik:
Disable compression for notify.js.
Include self compressed version from notify.js because the pre-compressed version doesn't work with plone (http://jscompress.com/).


phgross:
@bierik Änderungen sind :+1: aber bitte noch eine kurz Beschreibung bezüglich compression.

PR #1398, opened 2015-12-02
Save protocol only when clicking on save button.

bierik:
- Protocol gets saved when clicking cancel.
- Set event handler to save button instead of form.
- Discard localstorage saved protocol when hitting cancel.


deiferni:
:+1:

PR #1402, opened 2015-12-02
Filter proposals is not working.

bierik:
Fixes https://github.com/4teamwork/opengever.core/issues/1399

According to 84056e6 the filter functionality has been removed
accidentally.


deiferni:
:+1:

PR #1413, opened 2015-12-07
Avoid duplicated slashes when generating the CAS portal URL.

phgross:
Strip trailing slashes in admin_unit's public_url, when putting together the CAS portal URL.

@deiferni

### DIFF
--- a/opengever/base/casauth.py
+++ b/opengever/base/casauth.py
@@ -15,4 +15,4 @@ def is_cas_auth_enabled():
 
 def get_cas_portal_url():
     return CAS_SERVER_URL_FORMAT.format(
-        admin_unit_public_url=get_current_admin_unit().public_url)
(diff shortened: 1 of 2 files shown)


opengever/base/casauth.py
```diff
@@ -15,4 +15,4 @@ def is_cas_auth_enabled():
 
 def get_cas_portal_url():
     return CAS_SERVER_URL_FORMAT.format(
-        admin_unit_public_url=get_current_admin_unit().public_url)
+        admin_unit_public_url=get_current_admin_unit().public_url.strip('/'))
```
deiferni: you should use `rstrip` since we only strip on the right side. *Not that a public_url can start with a slash, but still ...*

deiferni:
:+1:

PR #1436, opened 2015-12-09
Fix an issue where proposals could not be rejected.

### LINKED ISSUE #1433: Antrag zurückweisen schlägt fehl
https://lab.onegovgever.ch/sitzungen/committee-17/submitted-proposal-71

deiferni:
The problem was caused by not removing all references to submitted_document from the proposal history entries. The submitted documents are deleted when rejecting the proposal because the `SubmittedProposal` proxy object is removed upon reject.

Fixes #1433.


deiferni:
@phgross plz have a look

PR #1450, opened 2015-12-09
Fix save link in meetings protocol view.

### LINKED ISSUE #1420: Protokollansicht: Speichern returned JSON Response
Beispiel auf Lab: https://lab.onegovgever.ch/sitzungen/committee-4/meeting-6/protocol

phgross:
Prevent the default event, when clicking on the save link in the protocol view.
Because the save method is redirecting manually back to the meeting view, after saving the changes via AJAX requests, we prevent the default link event. Fixes #1420.

@deiferni


deiferni:
:+1:

PR #1460, opened 2015-12-10
Disable unload protection for protocol.

bierik:
Fixes https://github.com/4teamwork/opengever.core/issues/1454

Implement enable_unload_protection configurable for each form. Will be fixed in plone.app.z3cform@0.7.6 -> https://pypi.python.org/pypi/plone.app.z3cform#id12
Disable unload_protection for protocol.
Disable unload protection for advanced search instead of overriding render method.


deiferni:
:+1:

PR #1464, opened 2015-12-11
Addded js workaroud to fix notification dropdown menu.

phgross:
With the PR #1378 the order in the jsregistry gets modified, so that plone's dropdown.js is now called earlier than our notification.js. This means that the notification dropdown menu does not work (don't drop down when clicking on the bell image).

The real problem is that the `toggleMenuHandler` in plone's `dropdown.js` return false instead of just prevent the default link event.

So to resolve this issue, we removed the actionMenuHeader class from the notification viewlet template.

@lukasgraf @deiferni

### DIFF
--- a/opengever/activity/browser/resources/notifications.js
+++ b/opengever/activity/browser/resources/notifications.js
@@ -77,7 +77,9 @@
       outlet: outlet,
       update: endpoints.readUrl
     });
-    $(".notificationsMenuHeader > a").on("click", function() {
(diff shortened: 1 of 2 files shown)


phgross:
Zur Info das `dropdown.js` wurde für Plone5 völlig überarbeitet (siehe https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/static/components/bootstrap/js/dropdown.js)

deiferni:
looks :+1: to me

PR #1467, opened 2015-12-14
Include ftw.profilehook, when required.

deiferni:
As documented in https://github.com/4teamwork/ftw.profilehook#usage we should always include ftw.profilehook in our zcml files.

This helps fixing an issue with failing opengever.maintenance tests.


lukasgraf:
:+1:

PR #1478, opened 2015-12-15
Fix CAS server and GEVER portal URLs

lukasgraf:
This PR introduces separate concepts for

- the base URL for the **GEVER cluster**
- the **CAS server** URL
- the **GEVER portal** URL

---

The **CAS server URL** is only constructed once (during setup) and used to configure the `ftw.casauth` PAS plugin. After that, the CAS server URL will always be retrieved from that plugin's property (for example to build the link for single-logout).

The **GEVER portal URL** however is constructed dynamically.

Both those URLs happen to be the same in our current setups, but they're always built using the **base URL of the GEVER cluster** as a base.

The GEVER cluster's base URL is determined through a heuristic: If the admin unit's `public_url` ends with the admin unit's ID, we're probably dealing with a multi-admin-unit setup, so we strip the admin unit ID from the end in order to get the correct base URL for the cluster.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,10 @@ Changelog
 4.6.2 (unreleased)
 ------------------
 
+- Cluster URL: Add heuristic to determine the correct cluster base URL
+  for both single- and multi-admin-unit setups.
+  [lgraf]
+
 - Make `my-portal` action use @@gever_state/gever_portal_url
   [lgraf]
 
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,7 +5,9 @@ Changelog
 4.6.2 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Only build URL to CAS server once during setup, after that fetch it
+  from the plugin when used as a base to build other URLs.
+  [lgraf]
 
 
 4.6.1 (2015-12-16)
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.6.2 (unreleased)
 ------------------
 
+- Introduce the concept of a base URL for the GEVER cluster.
+  [lgraf]
+
 - Only build URL to CAS server once during setup, after that fetch it
   from the plugin when used as a base to build other URLs.
   [lgraf]
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.6.2 (unreleased)
 ------------------
 
+- Make `my-portal` action use @@gever_state/gever_portal_url
+  [lgraf]
+
 - Add a `gever_state` view with traversable helper methods for use in TAL
   expressions and templates.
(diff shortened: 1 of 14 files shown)


deiferni:
Looks :+1: so far

opengever/setup/upgrades/casauth_profiles/4600/actions.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
+<object name="portal_actions" meta_type="Plone Actions Tool"
+        xmlns:i18n="http://xml.zope.org/namespaces/i18n">
+
+  <object name="user" meta_type="CMF Action Category">
+    <object name="my-portal" meta_type="CMF Action" i18n:domain="opengever.setup" insert-before="logout">
+      <property name="title" i18n:translate="">My portal</property>
```
deiferni: do you have to repeat all properties, even though they are not changed?

lukasgraf: Good point. We don't, updated accordingly.

lukasgraf:
@deif this one is ready for merging as far as I'm concerned. Changes since your last look at it:
- Dropped redundant properties from upgrade step
- Added heuristic to determine correct cluster URL
- Added some more tests

docs/HISTORY.txt
```diff
@@ -5,7 +5,25 @@ Changelog
 4.6.2 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Cluster URL: Add heuristic to determine the correct cluster base URL
+  for both single- and multi-admin-unit setups.
```
deiferni: please add an [lgraf] here, too :wink:

opengever/base/casauth.py
```diff
@@ -13,6 +12,50 @@ def is_cas_auth_enabled():
...
+    base_url = admin_unit.public_url
+    if not base_url.endswith('/'):
+        base_url = base_url + '/'
+
+    # XXX: Heuristic in order to determine the correct cluster base URL:
+    # If URL ends with the ID of the current admin unit, we're probably
+    # dealing with a deployment with multiple admin units. So we strip the
+    # admin unit ID from the end of the URL to get the correct cluster base.
+    #
+    # This heuristic will later be replaced with a configuration of the$
```
deiferni: the~~$~~

opengever/base/tests/test_gever_state.py
```diff
@@ -0,0 +1,29 @@
+from opengever.ogds.base.utils import get_current_admin_unit
+from opengever.testing import FunctionalTestCase
+from plone.app.testing import applyProfile
+
+
+class TestGeverStateView(FunctionalTestCase):
+
+    def setUp(self):
+        super(TestGeverStateView, self).setUp()
+        get_current_admin_unit().public_url = 'http://foo.org/cluster/fd'
```
deiferni: as discussed, `fd` is confusing here, i'd suggest to remove it

lukasgraf:
@deiferni updated.

PR #1481, opened 2015-12-16
Fixed missing `responsible` parameters in synced responses.

phgross:
When reassigning a multi AdminUnit task, the response on the foreign side, was not completly synchronised. The responsible attribute in the response changes was wrong. This PR fixes this problem.

Before:
kopierte_aufgabe_1024

After:
bildschirmfoto 2015-12-16 um 16 08 14

@lukasgraf or @deiferni

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.6.2 (unreleased)
 ------------------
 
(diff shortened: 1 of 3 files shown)


lukasgraf:
Looks good - will rebase onto master and fix a couple typos.

PR #1493, opened 2015-12-21
Update french translations (meeting, activity, task):

lukasgraf:
These should be all the newly translated french strings (and one fix) based on 8f0eb7e2.

See https://basecamp.com/2768704/projects/10521444/todos/205406450

@deiferni
/cc @phgross


lukasgraf:
@deiferni updated

deiferni:
:+1:

PR #1499, opened 2016-01-05
Reduce conflict errors by using uuid4 in CMFEditions.

jone:
Patch for Products.CMFEditions.historyidhandlertool.HistoryIdHandlerTool.register

The default "register" method uses the Products.CMFUid IUniqueIdGenerator utility for generating the history ID. This utility uses the auto-increment strategy, which generates a lot of conflicts.

In order to reduce the conflicts when generating the history id, we switch to the uuid4 implementation, generating a random number instead and thus not writing to the same place

--

The problem is when multiple users upload documents at the same time. The format of the used ID is not really relevant. Therefore no migration of existing objects is required; the old objects just use a smaller integer as ID.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,7 +5,9 @@ Changelog
 4.6.3 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Change CMFEdition IDs to uuid4 instead of auto-increment.
+  This reduces conflict errors.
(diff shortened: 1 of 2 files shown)


lukasgraf:
As discussed: Since we both tested this reasonably well, and the `opengever.core` ad `Products.CMFEditions` test suites are passing, lets :ship: this.

PR #1504, opened 2016-01-06
Eliminate modified_seconds index

jone:
The ``modified_seconds`` was used for generating the cache key for the navigation portlet (tree portlet).

The custom ``modified_seconds`` index caused a lot of database conflict (probably because it is not a ``DateIndex`` but a ``FieldIndex``) and we had trouble creating the index in the update because it took to much time.

**Detailed index problem**
For performance optimization, the ``modified`` index is not exactly precise, only down to the minute. If only the first item would have been used for the cache key, as it is done with the ``modified_seconds``, the cache may not have been invalidated correctly when multiple repository folders were changed in the same minute.

**New solution**
This PR replaces the cache key implementation to no longer use the ``modified_seconds`` index but the default ``modified`` index but considering all newest changed objects which were changed in the same minute, considering up to 100 objects.
This approach works quite well and would only not work correctly when 100 objects were changed in the same minute. 
(The number 100 was chosen without deeper thought, it just speeds up the catalog sorting.)

**Performance**
I've measured the method ``_get_newest_modification_timestamp`` for getting the timestamp with the old implementation and the new implementation, each running 100 times and calculating the mean:

- Old implementation: ~0.638ms
- New implementation: ~1.089ms

This calculation is done for every request rendering the main template. The performance drop is acceptable for me.


opengever/base/upgrades/configure.zcml
```diff
@@ -205,19 +205,11 @@
         />
 
     <!-- 4000 -> 4001 -->
-    <genericsetup:upgradeStep
-        title="Add 'modified_seconds' index for tree portlet."
-        source="4000"
-        destination="4001"
-        handler="opengever.base.upgrades.to4001.AddModifiedSecondsIndex"
-        profile="opengever.base:default"
-        />
```
lukasgraf: Have you considered leaving the actual upgrade-step in, and just replacing the code in the handler with some code that does nothing, and a comment describing that this index, and therefore the upgrade step, is not needed any more?

I don't feel to good about entirely removing an upgrade step that once existed and was executed on some systems. Not necessarily for technical reasons, but more to avoid human confusion. :wink:

opengever/base/browser/navigation.py
```diff
@@ -91,20 +91,54 @@ def _brain_to_node(self, brain):
...
+        if last_modified is not None:
+            version = get_distribution('opengever.core').version
+            username = getSecurityManager().getUser().getId()
+            return '-'.join((version, last_modified, username))
+        else:
+            return None
+
+    def _get_newest_modification_timestamp(self):
+        """Returns the timestamp in milliseconds of the latest modification
+        happend to any object in the navigation.
```
lukasgraf: -> `Returns the timestamp (in milliseconds) of the latest modification that happened to any object in the navigation.`

lukasgraf:
Other than that: Great, thank you! :+1:

opengever/base/upgrades/configure.zcml
jone: I agree, doing that.

opengever/base/browser/navigation.py
jone: :+1:

jone:
@lukasgraf updated & rebased.

jone:
@lukasgraf actually did the promised changes and gained a rebase skill level-up :grin:

lukasgraf:
I *thought* that diff of that first rebased version looked weird :stuck_out_tongue_winking_eye: Looks good now. :+1: 

Don't know whether I should wait on Jenkins failing because of low disk space or just merge right away...

jone:
Jenkins finished.

lukasgraf:
:+1:

PR #1517, opened 2016-01-11
Don't create savepoints in 46xx upgrade steps

lukasgraf:
MySQL implicitly commits savepoints as soon as a DDL statement is emitted (similar to Oracle).

That means we can't reliably use savepoints in upgrade paths that contain upgrades involving DDL.

@deiferni


deiferni:
:+1: thx!

PR #1520, opened 2016-01-12
Display state column in meeting overview tab of committee.

bierik:
Fixes https://github.com/4teamwork/opengever.core/issues/1456

As requested in https://github.com/4teamwork/opengever.core/issues/1456
display an additional column for the meeting state similar to the
proposals overview.

bildschirmfoto 2016-01-12 um 10 58 25

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.6.3 (unreleased)
 ------------------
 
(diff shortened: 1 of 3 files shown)


deiferni:
Looks :+1: thank you. Could you add a changelog entry?

bierik:
@deiferni updated

deiferni:
:+1:

PR #1524, opened 2016-01-13
Use trix widget for proposals form.

bierik:
Fixes https://github.com/4teamwork/opengever.core/issues/1522

Depends on https://github.com/4teamwork/plonetheme.teamraum/pull/409

bildschirmfoto 2016-01-13 um 16 09 39
bildschirmfoto 2016-01-13 um 16 09 51

### DIFF
--- a/opengever/base/widgets.py
+++ b/opengever/base/widgets.py
@@ -1,11 +1,16 @@
-""" Commen widget stuff
-"""
-
 from five import grok
 from plone.z3cform.textlines.textlines import TextLinesFieldWidget
 from z3c.form.interfaces import IFieldWidget
-from zope.schema.interfaces import ISequence, ITextLine
+from z3c.form.interfaces import IFormLayer
+from z3c.form.interfaces import ITextAreaWidget
+from z3c.form.widget import FieldWidget
+from z3c.form.widget import Widget
+from zope.component import adapter
+from zope.interface import implementer
+from zope.interface import implementsOnly
 from zope.publisher.interfaces.browser import IDefaultBrowserLayer
+from zope.schema.interfaces import ISequence
+from zope.schema.interfaces import ITextLine
 
 
 @grok.implementer(IFieldWidget)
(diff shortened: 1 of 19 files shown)


deiferni:
@lukasgraf plz have a look

opengever/base/widgets.py
```diff
@@ -14,3 +20,23 @@ def SequenceTextLinesFieldWidget(field, value_type, request):
     """ There is no default sequence widget for TextLine types
     """
     return TextLinesFieldWidget(field, request)
+
+
+class ITrixWidget(ITextAreaWidget):
+    """Text area widget with trix."""
+
+
+class TrixWidget(HTMLTextInputWidget, Widget):
```
lukasgraf: Is it really necessary to inherit from `Widget` here? My expectation would have been that `HTMLTextInputWidget` already inherits from `Widget`.

opengever/base/widgets.zcml
```diff
@@ -0,0 +1,22 @@
+<configure
+    xmlns="http://namespaces.zope.org/zope"
+    xmlns:z3c="http://namespaces.zope.org/z3c"
+    i18n_domain="z3c.form">
```
lukasgraf: `i18n_domain` isn't used anywhere

opengever/meeting/browser/proposalforms.py
```diff
@@ -16,6 +17,13 @@ class ProposalEditForm(ModelProxyEditForm, dexterity.EditForm):
     fields = field.Fields(Proposal.model_schema, ignoreContext=True)
     content_type = Proposal
 
+    fields['legal_basis'].widgetFactory = TrixFieldWidget
```
lukasgraf: Doesn't the `form.widget()` directive on the schema work? ([example](https://github.com/4teamwork/opengever.core/blob/995a95af00ec9d127c9343d2d34e6ee03bf2b471/opengever/tasktemplates/content/tasktemplate.py#L85))

lukasgraf:
Other than those points: I don't see anything obviously wrong with this :wink:

opengever/base/widgets.py
deiferni: Hmm, turns out `HTMLTextInputWidget` does not inherit from `Widget` but also is not necessary . It adds widget-configuration fields which we don't really use for our widget. I omitte it.

opengever/base/widgets.zcml
deiferni: thx, dropped it.

opengever/meeting/browser/proposalforms.py
deiferni: already tried that (and letting the interface inherit from `form.Schema`, too), unfortunately it does not work. Can't say why since we correctly inherit from `AutoExtensibleForm` via `dexterity.AddForm`. :disappointed:

deiferni:
@lukasgraf updated. streamlined widget inheritance a bit ... apart from that this seems to be the way to go

lukasgraf:
:+1:

PR #1531, opened 2016-01-18
Use new SAAS customer path in ldap template.

deiferni:
This PR catches up with recent changes in our LDAP structure and corrects the path used in LDAP templates.

It also fixes an error in a template where a variable was not substituted correctly.


deiferni:
@lukasgraf plz have a look.

lukasgraf:
:+1:

PR #1532, opened 2016-01-18
Add more meeting content to policy templates

deiferni:
This PR updates policy templates to:

- Include sample protocol/excerpt templates when meeting is enabled
- Include committee container when meeting is enabled
- Use german titles by default


deiferni:
@lukasgraf plz have a look

lukasgraf:
:+1:

PR #1534, opened 2016-01-19
Correctly use html content when filling templates.

deiferni:
The HTML structure is required for the template to be rendered correctly.


lukasgraf:
:+1:

PR #1487, opened 2015-12-21
Use Trix-Editor for editing protocol.

bierik:
According to the customers the current solution to edit the protocol with markdown was not satisfactory.
We use the latets [Sablon](https://github.com/senny/sablon) version which supports HTML. As clientside editor we use [Trix](https://github.com/basecamp/trix).

Depends on https://github.com/4teamwork/plonetheme.teamraum/pull/405

bildschirmfoto 2016-01-06 um 17 25 13

Closes #1518

PR #1543, opened 2016-01-20
Add migration for trix, convert fields from markdown to html.

### LINKED ISSUE #1527: Migration bestehende Markdown-Protokolle
Die bestehenden Protokolle im Markdown-Format müssen teilweise automatisch nach HTML migriert werden.

Mindestens Formatierungen und Listenaufzählungen sollten übernommen werden.

deiferni:
This migration converts fields from markdown to html for each `AgendaItem` and `Proposal`. This is required since we switched the editor to [trix](https://github.com/basecamp/trix), see #1487 and #1524. The migration is done independently for each table since the relation between agenda_items and proposals is weak and optional (there may be agenda_items without a proposal and vice versa).

The conversion is done with [Markdown](https://pypi.python.org/pypi/Markdown/2.0.3), the package seems to leave html alone when re-run with html input, so no additional checks are required.

I have tested the migration with data copied from a meeting on a productive system.

Closes #1527.


deiferni:
@lukasgraf plz have a look

opengever/meeting/upgrades/to4630.py
```diff
@@ -0,0 +1,121 @@
+from markdown import Markdown
```
lukasgraf: Doesn't this require a dependency in `setup.py`?

lukasgraf:
Other than the missing dependency: Looks good to me! :+1:

lukasgraf:
Rebased on behalf of @deiferni

deiferni:
@lukasgraf updated, `setup.py` dependency added

lukasgraf:
:+1:

PR #1551, opened 2016-01-22
Exclude non visible fields from the scrollspy navigation.

bierik:
Fixes https://github.com/4teamwork/opengever.core/issues/1548


deiferni:
:+1:

PR #1559, opened 2016-01-25
Fix encoding issue when displaying meeting state.

deiferni:
@lukasgraf plz have a look

*CL entry not necessary, the bug was added since the last release*


lukasgraf:
:+1:

PR #1571, opened 2016-02-02
Prevent protocol overwrites.

### LINKED ISSUE #1563: Hybrid Protokoll Pessimistisches/Optimistisches Locking
@senny:
> Den Pessimistic Locking Ansatz halte ich für nicht ausreichend. Es ist relativ schwer den für Webanwendungen sauber zu implementieren. Ich würde empfehlen Optimistic Locking für die Datensicherheit zu verwenden. Das funktioniert in etwa wie folgt:
> 
> Jeder Protokoll-Datensatz hat einen aufwärtszählenden Counter (Autoincrement oder Timestamp). Beim Laden des Formulars wird der Counter festgehalten. Beim Speichern kann man dann prüfen ob der Counter seit dem Laden nicht verändert wurde.
> 
> Das würde soweit einmal die Datenproblematik lösen. Wie genau mit fehlender Session oder Internetverbindung umgegangen werden soll, muss man separat anschauen.

deiferni:
Prevent that the protocol form can overwrite a newer version of the protocol.

If concurrent modifications occur we don't offer conflict resolution as of yet but leave the submitted form open to allow the user to take manual action.

Closes #1563.

*This problem should only occur in rare cases when users leave tabs open that contain unsaved changes or modify the same content in several tabs at the same time.*

screen shot 2016-02-02 at 11 47 30

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,11 @@ Changelog
 4.6.3 (unreleased)
 ------------------
 
+- Prevent that the protocol form can overwrite a newer version of the protocol.
+  If concurrent modifications occur we don't offer conflict resolution as of yet but
+  leave the submitted form open to allow the user to take manual action.
+  [deiferni]
+
 - Add migration for trix, convert fields from markdown to html.
   [deiferni]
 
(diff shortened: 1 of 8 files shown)


lukasgraf:
When testing locally (from 0e289fb), an attempt to overwrite was successfully prevented, but clicking the save button resulted in the error message quickly flashing, and then being redirected away from the form to `http://localhost:8080/fd/sitzungen/committee-2/undefined`, which resulted in a `404`.

lukasgraf:
Tested locally, works :tada: :wink: :+1:

PR #1573, opened 2016-02-02
Recycle reference numbers.

### LINKED ISSUE #1521: move to same folder/previously used folder does not preserve reference numbers
Wird ein Dossier von einer Ordnungsposition in dieselbe Ordnungsposition verschoben (Kann beim Formular so ausgewählt werden) wird für das entsprechende Dossier ein neues Aktenzeichen vergeben. Das Aktenzeichen sollte aber erhalten bleiben.

Analog: wenn man ein Dossier in eine Ordnungsposition verschiebt, in der es früher schon einmal war wird ein neues Aktenzeichen vergeben. Gewünscht wäre auch hier das Aktenzeichen beizubehalten.

Beim verschieben eines Dossier sollte man darauf achten, dass schon bestehende Einträge/Nummern im `prefix_mapping` wiederverwendet werden.

jone:
When moving an object back to a previous  parent, the previous reference number is recycled.

Solves these problems:

- When moving an object and selecting the current parent, which means that the object is not moved in fact, we now no longer change the reference number.

- When moving an object back to a container where it was before and already had a reference number, the reference number is now automatically recycled (unless explicitly set through the API).

Fixes #1521

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.6.4 (unreleased)
 ------------------
 
+- Recycle reference numbers. When moving an object back to a previous
+  parent, the previous reference number is recycled. [jone]
+
 - Remove "reference_number" from catalog metadata. [jone]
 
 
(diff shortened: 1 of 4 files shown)


deiferni:
:+1:

PR #1574, opened 2016-02-02
Prevent saving the protocol when it is locked by another user.

### LINKED ISSUE #1564: Protokoll kann trotz fremdem Lock gespeichert werden
Das Protokoll kann gespeichert werden, auch wenn dieses von einem anderen Benutzer gelockt ist.

deiferni:
Fixes #1564.

Remain on the form after a the error occurred, when the other user unlocks the protocol without modifications, it can be saved again by the current user.

screen shot 2016-02-02 at 16 58 53

Also, cleanup changelog.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.6.4 (unreleased)
 ------------------
 
+- Prevent saving the protocol when it is locked by another user.
+  [deiferni]
+
 - Rename "Kommission" to more generic "Gremium"
   [elioschmutz]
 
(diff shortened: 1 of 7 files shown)


deiferni:
@lukasgraf plz have a look

deiferni:
@lukasgraf rebased & updated

deiferni:
@lukasgraf rebased & updated

docs/HISTORY.txt
```diff
@@ -22,19 +25,21 @@ Changelog
...
-  parent, the previous reference number is recycled. [jone]
-
-- Remove "reference_number" from catalog metadata. [jone]
-
+  parent, the previous reference number is recycled.
+  [jone]
 
-4.6.3 (2016-01-25)
-------------------
+- Remove "reference_number" from catalog metadata.
```
lukasgraf: Rebase fail?

opengever/meeting/browser/meetings/protocol.py
```diff
@@ -201,11 +203,27 @@ def handleApply(self, action):
         if self.has_write_conflict():
             self._has_write_conflict = True
             return
+        if self.is_locked_by_another_user():
+            self._is_locked_by_another_user = True
+            return
```
lukasgraf: There's still a TOCTTOU race condition here - but I don't think we can really get rid of that without digging deep into Plone's internals.

lukasgraf:
Looks good to me :+1:

PR #1586, opened 2016-02-04
Add upgrade-step to update feedback link in footer (if necessary).

lukasgraf:
This is an upgrade step that basically applies the changes from #1554 to all deployments (if needed).

While the original change affects `opengever.examplecontent` and the policy template (in order not to set up new sites containing the wrong link), this upgrade step is in `opengever.base` and will be run for **all deployments**.

It's written in a way that should specifically fix the feedback forum URL if it occurs in an `ftw.footer` portlet, regardless of how that portlet was added.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.6.4 (unreleased)
 ------------------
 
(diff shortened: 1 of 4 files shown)


deiferni:
Looks :+1:

Issue #1598, opened 2016-02-11
Improve Controller.js documentation, rewrite event registration


deiferni:
[Event registation](https://github.com/4teamwork/opengever.core/blob/master/opengever/meeting/browser/resources/meeting.js#L173-L182) has become too cryptic to read, i'd suggest to replace the string with an object:
```javascript
    this.events = [{
            event: "click",
            target: ".delete-agenda-item",
            update: false,
            prevent: true,
            callback: somefunc
        }, {
            ....
        }];

```

[Controller.js](https://github.com/4teamwork/opengever.core/blob/master/opengever/base/browser/resources/Controller.js) has become an abstract base class for many different controllers. Unfortunately there is almost no documentation (today i spent quite some time figuring out how event registration works :wink:). Please add some documentation on how it works and how it should be extended.

- [ ] Refactor event registration
- [ ] Improve/Add Controller.js documentation

bierik:
@deiferni should I place this documentation in the README?

deiferni:
@bierik i'd suggest to place the docs in the js file with the code. that's were i'd look for it. If there is too much text we can always add minification :wink:

deiferni:
(documentation is not added yet)

PR #1603, opened 2016-02-12
Registering action on controllers has grown in complexity.

bierik:
Fixes https://github.com/4teamwork/opengever.core/issues/1598

Use explicit object configuration of actions instead of string notation.


deiferni:
Looks :+1:

++readability :sparkles:

PR #1604, opened 2016-02-15
Include lists in sample-template debug views.

deiferni:
This PR includes lists (ordered an unordered) in a view available to `Manager` that allows debugging sablon templates. This allows to detect problems in templates faster.


deiferni:
@lukasgraf plz have a look

lukasgraf:
:+1:

PR #1606, opened 2016-02-16
Fixes an issue with broken db-migrations.

### LINKED ISSUE #1605: Migrations fail
https://github.com/4teamwork/opengever.core/blob/master/opengever/meeting/upgrades/to4603.py#L14 fails when the model expects more columns (that have to be added in an upgrade-step not yet executed).

same here: https://github.com/4teamwork/opengever.core/blob/master/opengever/meeting/upgrades/to4633.py#L16

deiferni:
Don't load proposal model in upgrade steps but use sqlalchemy expression api for database access. This ensures that migrations also run in the future, when the model changes.

Also add a separate base-class for sql-based migrations that need to run once per plone site/admin-unit and not only once per cluster.

Fixes #1605

### DIFF
--- a/opengever/core/upgrade.py
+++ b/opengever/core/upgrade.py
@@ -240,16 +240,38 @@ class DeactivatedFKConstraint(object):
         return False
 
 
-class SchemaMigration(UpgradeStep):
-    """Baseclass for database-(schema) upgrade steps.
+class SQLUpgradeStep(UpgradeStep):
+    """Baseclass for upgrade steps that modify database content.
 
-    Maintains a tracking table to make sure that a database migration is only
+    This migration is run per plone site, if you want a migration that runs
+    only once have a look at `SchemaMigration`.
+
+    """
+    def __call__(self):
+        self._setup_db_connection()
+        self.migrate()
+
+    def migrate(self):
+        raise NotImplementedError()
+
+    def execute(self, statement):
+        return self.connection.execute(statement)
+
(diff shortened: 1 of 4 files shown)


deiferni:
@lukasgraf plz have a look

opengever/core/upgrade.py
```diff
@@ -240,16 +240,38 @@ def __exit__(self, exc_type, exc_value, exc_tb):
...
-    Maintains a tracking table to make sure that a database migration is only
+    This migration is run per plone site, if you want a migration that runs
+    only once have a look at `SchemaMigration`.
+
+    """
+    def __call__(self):
+        self._setup_db_connection()
+        self.migrate()
+
+    def migrate(self):
```
lukasgraf: I would tend to just call this `run()` or something similar - may not be an actual migration in every case

lukasgraf:
Looks good to me otherwise! :+1: :sparkles: :cake:

opengever/core/upgrade.py
deiferni: Hmm i'd like to have the same name here as a `SchemaMigration` where `migrate` makes more sense and is already in use. I could rename both methods to `run` but that would mean many retroactive changes for the old upgrade-steps. If this is ok for you i'd suggest to keep `migrate` here.

lukasgraf: :+1:

PR #1607, opened 2016-02-16
Introduce [api] extra to pull in plone.restapi and use it for tests and development

lukasgraf:
This is the first step for integrating [`plone.restapi`](https://github.com/plone/plone.restapi) into `opengever.core`.

This introduces an `[api]` extra for `opengever.core`, and uses it for tests and development.

For now, both `plone.rest` and `plone.restapi` will be used from source:
- plone.restapi doesn't have a release yet
- plone.rest is still changing heavily, we want to stay current for now

### DIFF
--- a/development.cfg
+++ b/development.cfg
@@ -25,6 +25,7 @@ ogds-db-driver = psycopg2
 zserver-threads = 4
 user = zopemaster:admin
 eggs +=
+    opengever.core[api]
     plonetheme.teamraum
     opengever.maintenance
     ${buildout:ogds-db-driver}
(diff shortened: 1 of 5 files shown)


development.cfg
```diff
@@ -25,6 +25,7 @@ ogds-db-driver = psycopg2
 zserver-threads = 4
 user = zopemaster:admin
 eggs +=
+    opengever.core[api, tests]
```
deiferni: why do you include the tests extra here?

lukasgraf: Good point, shouldn't be needed.

lukasgraf:
@deiferni updated, tests ran again and passed

deiferni:
:+1:

PR #1618, opened 2016-02-22
Avoid infinite loop in BasicReferenceNumber.get_parent_numbers()

lukasgraf:
This avoids an infinite loop in `BasicReferenceNumber.get_parent_numbers()` if the context is not acquisition wrapped: The loop also needs to be terminated when `aq_parent(obj)` returns `None`, which is the case for unwrapped objects.

@deiferni


deiferni:
:+1: 

*i guess we need to find another way to heat the server room, then.*

PR #1621, opened 2016-02-24
Prevent anonymous access to browser views on plone site.

### LINKED ISSUE #1608: Tabbedview-Views visible as Anonymous
Under certain circumstances some tabbedview-views may be visible as anonymous. For more details see: https://4teamwork.slack.com/archives/gever/p1455701691000002

deiferni:
This is a follow-up for #1278 to also prevent access when context does not provide `IPloneSiteRoot` but `IBrowserView` as is the case for e.g. a `TabbedView`. Fixes #1608.

It also contains a custom event handler for an event that is added in https://github.com/4teamwork/ftw.tabbedview/pull/51. It issues a page reload which then causes a redirect to the login form when we receive requests from the gever portal.

See also https://github.com/4teamwork/ftw.tabbedview/pull/51

### DIFF
--- a/opengever/base/subscribers.py
+++ b/opengever/base/subscribers.py
@@ -1,6 +1,8 @@
 from AccessControl import Unauthorized
 from AccessControl.SecurityManagement import getSecurityManager
+from Acquisition import aq_parent
 from five import grok
+from ftw.tabbedview.browser.tabbed import TabbedView
 from OFS.interfaces import IObjectClonedEvent
 from opengever.base import _
 from plone.app.lockingbehavior.behaviors import ILocking
@@ -22,7 +24,7 @@ from zope.lifecycleevent.interfaces import IObjectCreatedEvent
 from ZPublisher.interfaces import IPubAfterTraversal
 
 
-ALLOWED_VIEWS = set([
+ALLOWED_ENDPOINTS = set([
     'customlogo',
     'favicon',
     'list-open-dossiers-json',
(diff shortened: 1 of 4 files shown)


opengever/base/subscribers.py
```diff
@@ -123,6 +125,8 @@ def disallow_anonymous_views_on_site_root(event):
     user = getSecurityManager().getUser()
     if user is None or user.getUserName() == 'Anonymous User':
         context = event.request['PARENTS'][0]
+        if IBrowserView.providedBy(context) and not IResource.providedBy(context):
```
lukasgraf: Hmm - I feel that this might be too aggressive. This isn't restricted to just the site root, this would affect any browser view anywhere, wouldn't it? And it doesn't consider the `ALLOWED_VIEWS` whitelist.

deiferni: Hmm yeah it may be a bit extreme ...

For me an alternative would be to only address the know problem only, i.e. check if we are on a tabbed_view and only then raise `Unauthorized` and ignore all other cases of anonymous access to a browser view. What do you think?

lukasgraf: For all other cases than the site root, we shouldn't have to do anything for browser views anyway. BrowserViews do have a required permission associated with them that is automatically checked on the context the view is invoked on - the issue here is that everyone and their grandmother has `View` on the `SiteRoot` :wink: So that's the case we need to handle.

So what I don't quite understand yet is why this check in the existing code doesn't catch anonymous calls to tabbed views:
```python
             view_name = event.request['PUBLISHED'].__name__
              if view_name not in ALLOWED_VIEWS:
                 raise Unauthorized
```

deiferni: Ah yes, that's confusing. The Problem ist that we don't access a view on the `SiteRoot` but attributes of a `tabbed_view` (https://github.com/4teamwork/ftw.tabbedview/blob/master/ftw/tabbedview/browser/configure.zcml#L24). So our `context` in this case is a `TabbedView` instance. (And the `TabbedView` instances context is the `SiteRoot`.) 

The longer i think about this the more it seems that we should not be too intellligent here and maybe just better handle the known issue, e.g. let `TabbedView` provide a distinct interface e.g. `ITabbedView` and do
```python
if ITabbedView.providedBy(context):
  raise Unauthorized
```
What do you think?

lukasgraf: I would suggest something like this, conceptually:
```python
        is_site_root = ISiteRoot.providedBy(context)
        is_tabbed_view_on_site_root = isinstance(context, TabbedView) and \
            ISiteRoot.providedBy(aq_parent(context))

        if is_site_root or is_tabbed_view_on_site_root:
            endpoint_name = event.request['PUBLISHED'].__name__
            if endpoint_name not in ALLOWED_ENDPOINTS:
                raise Unauthorized
```

deiferni: :yellow_heart:

deiferni:
@lukasgraf updated as suggested.

lukasgraf:
:+1:

PR #1626, opened 2016-02-29
Avoid some more infinite loops for non-acquisition wrapped contexts

lukasgraf:
Avoid infinite loops in `find_parent_dossier()` and `set_default_with_acquisition()` if context is not acquisition wrapped.

@deiferni


deiferni:
:+1:

PR #1630, opened 2016-03-02
Email-Adressen von Teilnehmern werden in Meetings angezeigt.

elioschmutz:
Emailadressen werden nun zusätzlich zum Namen an folgenden Orten angezeigt:

Sitzung
----------

bildschirmfoto 2016-03-02 um 15 21 35

Protokoll bearbeiten
---------------------------
bildschirmfoto 2016-03-02 um 15 53 47

Protokoll
------------
bildschirmfoto 2016-03-02 um 15 22 09

### DIFF
--- a/opengever/meeting/protocol.py
+++ b/opengever/meeting/protocol.py
@@ -1,3 +1,4 @@
+from opengever.base.utils import escape_html
 from opengever.meeting import _
 from opengever.meeting.model.membership import Membership
 from opengever.ogds.base.utils import get_current_admin_unit
@@ -65,13 +66,14 @@ class ProtocolData(object):
             membership = Membership.query.fetch_for_meeting(
                 self.meeting, participant)
             members.append({
-                "fullname": participant.fullname,
+                "fullname": participant.get_title(show_email_as_link=False),
                 "role": membership.role if membership else None
             })
 
         participants = {
             'members': members
         }
+
         return participants
 
     def add_participants(self):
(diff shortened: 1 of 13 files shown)


opengever/meeting/protocol.py
```diff
@@ -1,11 +1,25 @@
 from opengever.meeting import _
 from opengever.meeting.model.membership import Membership
 from opengever.ogds.base.utils import get_current_admin_unit
+from plone import api
 from zope.globalrequest import getRequest
 from zope.i18n import translate
 import json
 
 
+def format_participant(member, show_email_as_link=False):
```
deiferni: Please move this method to `Member` so we can use it in an object oriented manner, e.g:

`member.get_title(show_email_as_link=True)`

opengever/meeting/protocol.py
```diff
@@ -1,11 +1,25 @@
...
+    transformer = api.portal.get_tool('portal_transforms')
+    participant = member.fullname
+    email = member.email
+
+    if email:
+        if show_email_as_link:
+            email = '<a href="mailto:{email}">{email}</a>'.format(email=email)
+        participant = u'{} ({})'.format(member.fullname, email)
+
+    return transformer.convert('safe_html', participant).getData()
```
deiferni: You don't need to use transforms here, they are relatively resource intensive, plus we don't want to transform and remove stuff, but display exactly what the user entered.

Use `escape_html` (`opengever.base.utils`) instead, it escapes  &, <, >, " and ' which is completely sufficient here.

opengever/meeting/tests/test_meeting.py
```diff
@@ -126,3 +126,76 @@ def test_close_meeting(self, browser):
...
+
+        self.assertEqual(
+            u'Hans M\xfcller',
+            vocabulary._terms[0].title)
+
+    def test_return_fullname_with_email_as_value(self):
+        member = create(Builder('member').having(
+            firstname=u'Hans',
+            lastname=u'M\xfcller',
+            email=u'mueller@foo.ch'))
```
deiferni: Please use the [example.com](www.example.com)/[example.org](www.example.org) domains for example/test data, thats what they are here for :wink:

elioschmutz:
@deiferni :+1: alle Anpassungen umgesetzt.

deiferni:
:+1:

PR #1635, opened 2016-03-08
Activate "remove GEVER content" by default for new policies.

deiferni:
SaaS customers should always be able to do so.


deiferni:
@lukasgraf plz have a look

lukasgraf:
:+1:

PR #1640, opened 2016-03-10
Implement sorter for 'no_client_id_grouped_by_three' reference number formatter

lukasgraf:
Because it has no leading `client_id`, the `no_client_id_grouped_by_three` formatter also needs its own sorting logic.

The `sorter()` method is an exact copy of the one from its superclass, minus the [line that tries to split by client ID](https://github.com/4teamwork/opengever.core/blob/f0d1e0ef7f5667fe70c47d7b8cf08c857063c272/opengever/base/reference_formatter.py#L126).

Only verified via tests, haven't actually tested this on a real site yet.

Also see: https://github.com/4teamwork/opengever.core/issues/1561 (We should probably use a `sortable_reference` index instead of sorting on demand).

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.7.1 (unreleased)
 ------------------
 
+- Implement sorter for 'no_client_id_grouped_by_three' reference number formatter.
+  [lgraf]
+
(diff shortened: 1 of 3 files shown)


deiferni:
:+1:

PR #1642, opened 2016-03-10
Fix protocol saving on IE.

bierik:
Since the update to 2.3.8 of bootstrap-datetimepicker the bug here
smalot/bootstrap-datetimepicker#122 is fixed.

So we dont need the workarround applying the timezone manually.

Making a UTC date extending a datestring with `UTC` does not work on IE.
So the backend received invalid date and the protocol could not be saved.


deiferni:
:+1:

PR #1643, opened 2016-03-10
Take existing date from server to feed the datetimepicker.

bierik:
Depends on https://github.com/4teamwork/opengever.core/pull/1642

When the user wants to edit a protocol the datetimepicker always
sets a new date instead of taking the existing date from the server.

So check if the datefileds are prefilled before inserting a new date.
A new date should only get inserted when creating a new meeting.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.7.2 (unreleased)
 ------------------
 
+- Keep existing date on a protocol instead of creating a new one.
+  [Kevin Bieri]
+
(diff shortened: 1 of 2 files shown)


docs/HISTORY.txt
```diff
@@ -5,6 +5,9 @@ Changelog
 4.7.2 (unreleased)
 ------------------
 
+- Keep exising date on a protocol instead of creating a new one.
```
deiferni: typo existing

deiferni:
:+1:

PR #1646, opened 2016-03-14
Update policy templates

deiferni:
This PR introduces two small changes to policy templates.
- Add deployment base config file (at the moment it must be edited manually)
- Let new policies inherit from Plone 4.3 testing config.


opengever/policytemplates/policy_template/opengever.+package.name+/deployment-theta-XX-+base.domain+.cfg.bob
```diff
@@ -0,0 +1,15 @@
+[buildout]
+extends =
+    https://raw.githubusercontent.com/4teamwork/gever-buildouts/master/standard-deployment.cfg
+    https://raw.githubusercontent.com/4teamwork/gever-buildouts/master/ogds-postgres.cfg
+    https://raw.githubusercontent.com/4teamwork/gever-buildouts/master/tika-standalone.cfg
+    https://raw.githubusercontent.com/4teamwork/ftw-buildouts/master/zeoclients/2.cfg
+    http://kgs.4teamwork.ch/release/opengever/XX
```
lukasgraf: I would use separate placeholders here - one for the KGS version, and one for the deployment number. ~~Like KKK for KGS...~~ maybe `XXX` and `YYY`? :stuck_out_tongue_winking_eye:

opengever/policytemplates/policy_template/opengever.+package.name+/deployment-theta-XX-+base.domain+.cfg.bob
```diff
@@ -0,0 +1,15 @@
...
+
+deployment-number = XX
+ogds-db-name = gever-{{{package.name}}}
+client-policy = opengever.{{{package.name}}}
+usernamelogger_ac_cookie_name = __ac_{{{adminunit.id}}}
```
lukasgraf: This won't be correct, unfortunately - the notation used in the cookie name is `underscore_lower`, it doesn't (necessarily) match the admin unit ID. I would probably replace it with something like `${REPLACE_ME}` so that buildout fails if someone forgets to correctly set it.

deiferni: Note this is the database identifier, not the abbreviation, which should always be chosen `underscore_lower` and is usually prefilled from the package name or the abbreviation and converted to lower case: https://github.com/4teamwork/opengever.core/blob/master/opengever/policytemplates/hooks.py#L32 or https://github.com/4teamwork/opengever.core/blob/master/opengever/policytemplates/hooks.py#L55

i'd suggest to leave the default as is, if somebody wants so shoot himself in the foot with upper-case ids its his problem :stuck_out_tongue_winking_eye:

opengever/policytemplates/policy_template/opengever.+package.name+/deployment-theta-XX-+base.domain+.cfg.bob
deiferni: :+1:

opengever/policytemplates/policy_template/opengever.+package.name+/deployment-theta-XX-+base.domain+.cfg.bob
lukasgraf: Cookie name is generated [here](https://github.com/4teamwork/opengever.core/blob/ca8af4e605d92773f517ea4bf8bf99ee5330f77d/opengever/setup/casauth.py#L45)

deiferni:
@lukasgraf updated

lukasgraf:
:+1:

PR #1652, opened 2016-03-17
Optimize `user_is_allowed_to_view()`.

buchi:
- use optimized method in `AdminUnit` to determine if a user is assigned to it
- memoize `user_is_allowed_to_view` per request as it's called multiple times in some views.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.7.3 (unreleased)
 ------------------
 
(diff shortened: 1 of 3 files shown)


lukasgraf:
@buchi looks good to me :+1: 

But I believe you'll have to add `opengever.ogds.models` to `development-packages` in `sources.cfg` for tests to pass (and also so we know we'll have to release and bump it when releasing `og.core`).

lukasgraf:
:tada:

PR #1662, opened 2016-03-31
Fix excel repositories.

deiferni:
- Fix retention period header in repository excel.
- Lock sheets/cells that must not be changed.
- Drop unsupported columns.
- Correctly setup data validation for columns.


deiferni:
cc @phabegger

PR #1702, opened 2016-04-11
No longer attempt to parse temporary excel files while creating repos.

deiferni:
For each open file MS office creates a temp file in the same directory. The temp file's filename starts with a tilde, but ends with the same file extension as the original file (`.xlsx`). Currently our import section attempts to parse these files which leads to an error.

This PR changes the behavior of the excel import section to also skip MS office temp files.

Tested and confirmed running with: 

```bash
foo.bar               ordnungssystem.xlsx   ~$ordnungssystem.xlsx ~blub.xlsx
```

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.7.5 (unreleased)
 ------------------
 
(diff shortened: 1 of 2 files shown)


lukasgraf:
:+1: :purple_heart:

PR #1731, opened 2016-04-19
API Docs: Add some code examples (Python)

lukasgraf:
Add some code examples in Python for the basic operations. As currently published on https://docs.onegovgever.ch/operations/

*(The CSS and JS for collapsibles will eventually go into a separate package `opengever.sphinx` as part of the theme, but I haven't gotten around to creating that package/repo just yet.)*

@deiferni


deiferni:
👍

PR #1732, opened 2016-04-19
Make default value for IDossier.start a defaultFactory instead of a form level default

lukasgraf:
This ensures this default is also respected when programmatically creating content (REST API, `plone.api`, `ftw.builder`, `transmogrifier`, ...).

See: http://docs.plone.org/4/en/external/plone.app.dexterity/docs/advanced/defaults.html#defaults

@phgross


deiferni:
👍

PR #1745, opened 2016-04-21
TranslatedTitleBrain: Also use Missing.Value as indicator to use regular title

lukasgraf:
TranslatedTitleBrain: Also use `Missing.Value` as indicator that a brain has no translated title, and should therefore return the regular title.

Objects that existed before the TranslatedTitle behavior was activated will have `Missing.Value` indexed for `title_de` and `title_fr` metadata. The existing check for `None` will therefore fail for those objects.


lukasgraf:
@phgross

PR #1769, opened 2016-04-29
bin/mtest: Remove bytecode before spawning parallel workers.

lukasgraf:
Remove `*.py[co]` files to avoid race conditions with parallel workers stepping on each other's toes when trying to clean up stale bytecode.

Setting `PYTHONDONTWRITEBYTECODE` is not enough, because running buildout also already precompiles bytecode for some eggs.

@phgross


phgross:
👍  thx!

PR #1801, opened 2016-05-10
Catch up with minor refactoring in ftw.showroom

deiferni:
Package organization was improved and files were
renamed, see
https://github.com/4teamwork/ftw.showroom/pull/6.


deiferni:
cc @phgross

PR #1818, opened 2016-05-20
Update GeverBumblebeeService for the new ftw.bumblebee version.

phgross:
With https://github.com/4teamwork/ftw.bumblebee/pull/75 ftw.bumblebee switched to the Bumblebee API v3, therefore we need to updated the `GeverBumblebeeService` as well.

@elioschmutz @deiferni

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.9.0 (unreleased)
 ------------------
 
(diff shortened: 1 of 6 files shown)


phgross:
@deiferni @elioschmutz https://github.com/4teamwork/ftw.bumblebee/pull/75 ist gemerged, demnach müsste man den hier auch 🚢

deiferni:
👍

PR #1826, opened 2016-05-23
 Add some basic tests for content creation via API

lukasgraf:
@deiferni

PR #1827, opened 2016-05-24
Have bin/mtest allocate a unique ZSERVER_PORT for each worker

lukasgraf:
Have `bin/mtest` allocate a unique `ZSERVER_PORT` for each worker based on `PORT[1-8]` (see 4teamwork/ci_governor#71) provided by Jenkins Port Allocator plugin.

This is necessary so that we don't have workers trying to start ZServers on conflicting ports once #1826 ist merged to `master`.

@deiferni

### DIFF
--- a/bin/mtest
+++ b/bin/mtest
@@ -3,6 +3,7 @@
 # Runs the tests in multiple threads.
 
 from collections import defaultdict
+from copy import copy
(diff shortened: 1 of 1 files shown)


deiferni:
👍

PR #1828, opened 2016-05-24
Implement broken image fallback.

bierik:
Replace source of preview image with a fallback image if loading failed.

bildschirmfoto 2016-05-24 um 10 54 49

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.9.0 (unreleased)
 ------------------
 
+- Bumblebee: Implement gallery preview image fallback.
+  [Kevin Bieri]
+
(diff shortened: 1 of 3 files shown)


deiferni:
how do you detect if an image is "broken"?

I'm not sure if we want to have a big red x symbol in evert case. Not having an image may be a valid use-case (i.e. when the document is not digitally available). The current symbol looks too much like an error for that case, so we should make sure it does only appear when an error ocurred.

bierik:
@deiferni The broken image is just for connection issues with the bumblebee server. This fallback does not handle the conversion errors that may occur in bumblebee.
I replaced the red `X` Symbol with a gray one.

deiferni:
i like it 👍

PR #1868, opened 2016-06-01
Dont quote advanced search parameters twice.

### LINKED ISSUE #1867: Dont quote and urlencode params
Currently advanced-search both `quotes` and `urlencodes` parameters passed to `@@search`. This is once too much and leads to errors since urlencode already applys `quote_plus` to its params, see https://docs.python.org/2/library/urllib.html#urllib.urlencode.

deiferni:
Fixes #1867.


deiferni:
@phgross plz have a look

PR #1877, opened 2016-06-02
Reindex documents after adding ftw.bumblebee.interfaces.IBumblebeeable.

### LINKED ISSUE #1870: Reindex documents after adding 'ftw.bumblebee.interfaces.IBumblebeeable'
We should reindex the object_provides index to update this correctly for all documents.

deiferni:
This makes sure that object_provides is up to date.

Closes #1870.

### DIFF
--- a/opengever/document/upgrades/20160601180404_reindex_object_provides_after_adding_i_bumblebeeable_interface/upgrade.py
+++ b/opengever/document/upgrades/20160601180404_reindex_object_provides_after_adding_i_bumblebeeable_interface/upgrade.py
@@ -1,0 +1,15 @@
+from ftw.upgrade import UpgradeStep
+from plone import api
+
+
+class ReindexObjectProvidesAfterAddingIBumblebeeableInterface(UpgradeStep):
+    """Reindex object_provides after adding IBumblebeeable interface.
+    """
+
+    def __call__(self):
+        catalog = api.portal.get_tool('portal_catalog')
+        query = {'portal_type': 'opengever.document.document'}
+        msg = 'Reindex object_provides for documents.'
+
+        for obj in self.objects(query, msg):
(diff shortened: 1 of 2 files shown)


deiferni:
@lukasgraf plz have a look

opengever/document/upgrades/20160601180404_reindex_object_provides_after_adding_i_bumblebeeable_interface/upgrade.py
```diff
@@ -0,0 +1,17 @@
+from ftw.upgrade import UpgradeStep
+from plone import api
+
+
+class ReindexObjectProvidesAfterAddingIBumblebeeableInterface(UpgradeStep):
+    """Reindex object_provides after adding IBumblebeeable interface.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
```
lukasgraf: Not needed - doesn't this [raise a `NoAssociatedProfileError`](https://github.com/4teamwork/ftw.upgrade/blob/ba2856585abbd32902da5a4738dbfdb015c002e2/ftw/upgrade/step.py#L275-L279) anyway?

lukasgraf:
Otherwise looks good 👍

opengever/document/upgrades/20160601180404_reindex_object_provides_after_adding_i_bumblebeeable_interface/upgrade.py
deiferni: 👍  there is no xml so i'll remove this line

(for new-style upgrade-steps id does not raise, the profile is generated and registered automagically. it is the same directory that contains `upgrade.py`, `20160601180404_reindex_object_provides_after_adding_i_bumblebeeable_interface` in our case. )

lukasgraf:
👍

PR #1878, opened 2016-06-03
Sortierreihenfolge wird im Dokumenttab wieder persistiert.

### LINKED ISSUE #1850: Sortierreihenfolge in Dokument-Reiter wird nicht mehr persistiert
Die Dokumentreihenfolge wird nicht mehr persistiert (auf test.opengever.ch funktionert es noch).

Wenn auf einem Dokumentreiter z.B. nach Laufnummer sortiert wird, sollte diese Sortierung gespeichert werden und bei einem späteren Zugriff auf diesen Reiter wieder verwendet werden.

record

Dieses funktionier leider nicht mehr.

- [x] Soriterreihenfolge in Dokuement-Reiter muss persitiert werden (siehe test.opengever.ch)

elioschmutz:
Dieser PR fixt das problem, dass Tablestates für das Dokumenttab nicht mehr persistiert wurden.

closes #1850

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,12 @@ Changelog
 4.9.0 (unreleased)
 ------------------
 
+- Fix: Add GeverTabbedviewDictStorage adapter to handle bumblebee proxy views.
+  Normally, the tabbedview is storing the last state (i.e. sorting) per user per tab.
+  Since bumblebee requires a proxy view to determine if the list or the gallery should be
+  rendered, the mechanism to store and get the state was broken.
+  [elioschmutz]
+
 - Initialize the showroom correctly for dynamically loaded items in the search-view.
   [elioschmutz]
 
--- a/opengever/dossier/tabbedviewstatestorage.py
+++ b/opengever/dossier/tabbedviewstatestorage.py
@@ -1,9 +1,13 @@
-from Products.CMFCore.utils import getToolByName
 from five import grok
+from ftw.dictstorage.base import DictStorage
 from ftw.tabbedview.interfaces import IGridStateStorageKeyGenerator
 from opengever.dossier.behaviors.dossier import IDossierMarker
-from zope.publisher.interfaces.browser import IBrowserView
+from opengever.tabbedview.interfaces import ITabbedViewProxy
+from Products.CMFCore.utils import getToolByName
 from zope.publisher.interfaces.browser import IBrowserRequest
+from zope.publisher.interfaces.browser import IBrowserView
+
+PROXY_VIEW_POSTFIX = "-proxy"
 
 
 class DossierGridStateStorageKeyGenerator(grok.MultiAdapter):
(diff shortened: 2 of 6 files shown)


docs/HISTORY.txt
```diff
@@ -5,6 +5,13 @@ Changelog
 4.9.0 (unreleased)
 ------------------
 
+- Fix: Add GeverTabbedviewDictStorage adapter to handle bumblebee proxy views.
+  Normally, the tabbedview is storing the last state (i.e. sorting) per user per tab.
+  Since bumblebee requires a proxy view to determine if the list or the gallery should be
+  rendered, the mechanism to storing and getting the state was broken. So it was no longer
```
deiferni: ... to store and get ...

then i would drop the last sentence(So it was no longer...), it does not really contain new information 😉

opengever/dossier/tabbedviewstatestorage.py
```diff
@@ -36,3 +38,57 @@ def get_key(self):
...
+    def __setitem__(self, key, value):
+        key = self.change_proxy_name(key)
+        return super(GeverTabbedviewDictStorage, self).__setitem__(key, value)
+
+    def __delitem__(self, key):
+        key = self.change_proxy_name(key)
+        return super(GeverTabbedviewDictStorage, self).__setitem__(key)
+
+    def change_proxy_name(self, value):
+        return value.replace('-proxy', '')
```
deiferni: - i'd suggest to move this to `ITabbedViewProxy.PROXY_VIEW_POSTFIX` to indicate that this is the naming convention that is used
- also we only want to do this if `if ITabbedViewProxy.providedBy(context)`, right?

opengever/tabbedview/interfaces.py
```diff
@@ -13,3 +13,9 @@ class IGeverTableSourceConfig(ITableSourceConfig):
 
 class IGeverCatalogTableSourceConfig(IGeverTableSourceConfig, ICatalogTableSourceConfig):
     pass
+
+
+class ITabbedViewProxy(Interface):
```
deiferni: when you create the constant `ITabbedViewProxy.PROXY_VIEW_POSTFIX` also write a short docu about it (i.e. view names are required to end with `-proxy`)

opengever/dossier/tabbedviewstatestorage.py
```diff
@@ -36,3 +38,57 @@ def get_key(self):
...
+
+    def change_proxy_context(self, context):
+        if ITabbedViewProxy.providedBy(context):
+            context = context.context.restrictedTraverse(
```
deiferni: i'm confused here, shouldn't context be the view? so you should be able to `context.restrictedTraverse`?

opengever/dossier/tabbedviewstatestorage.py
```diff
@@ -36,3 +38,57 @@ def get_key(self):
...
+
+    def __setitem__(self, key, value):
+        key = self.change_proxy_name(key)
+        return super(GeverTabbedviewDictStorage, self).__setitem__(key, value)
+
+    def __delitem__(self, key):
+        key = self.change_proxy_name(key)
+        return super(GeverTabbedviewDictStorage, self).__setitem__(key)
+
+    def change_proxy_name(self, value):
```
deiferni: I'd suggest to rename this to `strip_proxy_postfix(self, name)`

opengever/dossier/tabbedviewstatestorage.py
deiferni: and also use `value.rstrip(...)` since we always and only want to strip on the right side

deiferni:
Looks good in general, some comments though, especially to make the code a bit more defensive

opengever/dossier/tabbedviewstatestorage.py
elioschmutz: `context` is the view. But `restrictedTraverse` is only available for the object itself which is under  `context.context`

opengever/dossier/tabbedviewstatestorage.py
elioschmutz: As discussed => It's not possible to add a constant to the Interface

Excerpt from https://docs.zope.org/zope.interface/README.html:

Remember that interfaces aren’t classes. You can’t access attribute definitions as attributes of interfaces:

```
>>> IFoo.x
  File "<stdin>", line 1, in ?
AttributeError: 'InterfaceClass' object has no attribute 'x'
```

elioschmutz:
@deiferni could you review again?

opengever/dossier/tabbedviewstatestorage.py
```diff
@@ -36,3 +40,58 @@ def get_key(self):
...
+
+    def __delitem__(self, key):
+        key = self.strip_proxy_postfix(key)
+        return super(GeverTabbedviewDictStorage, self).__setitem__(key)
+
+    def strip_proxy_postfix(self, value):
+        if ITabbedViewProxy.providedBy(self.context):
+            return value.rstrip(PROXY_VIEW_POSTFIX, '')
+
+    def change_proxy_context(self, context):
```
deiferni: ah, now i see 😉 . I would change the naming slightly and add a variable to improve readability, something like this:

```python
def change_proxy_context(self, view):
    if ITabbedViewProxy.providedBy(context):
       non_proxy_view_name = self.strip_proxy_postfix(view.__name__)
       view = view.context.restrictedTraverse(non_proxy_view_name)

    return view
```

opengever/dossier/tests/test_tabbedviewstatestorage.py
```diff
@@ -0,0 +1,35 @@
...
+        self.document = create(Builder("document").within(self.dossier))
+
+    def test_call_storage_with_proxy_will_change_context_to_subview(self):
+        proxy_view = self.document.restrictedTraverse(
+            'tabbedview_view-documents-proxy')
+
+        sub_view = self.document.restrictedTraverse(
+            'tabbedview_view-documents')
+
+        self.assertIsNot(type(proxy_view), type(IDictStorage(proxy_view).context))
```
deiferni: i would rewrite these and use `assertIs[Not]Instance`

opengever/tabbedview/interfaces.py
deiferni: As discussed this cannot be done here since `zope.interface` prevents concrete attributes on interfaces, which is correct, of course. very correct.

deiferni:
two more comments, almost ready 👍

elioschmutz:
👍  all done!

PR #1890, opened 2016-06-06
Fixed ReindexObjectProvidesAfterAddingIBumblebeeableInterface upgradestep

phgross:
The existing upgradestep, which was introduced with #1877, also updates the metadata. This is unnecessary because there is only a index `object_provides` but not a metadata.

@deiferni


deiferni:
👍 😓

PR #1893, opened 2016-06-07
Strip whitespace when saving proposals or a protocol.

### LINKED ISSUE #1881: Trim Protocol/Proposal Text-Fields
Whitespace is not removed when saving a Text-Field in the Protocol or while editing a Proposal.

deiferni:
Since we process input from trix this is a bit more complicated than it looks at first glance. 

- Trix-Markup is always wrapped in a `div`, i had to adapt existing tests that were incorrect so far
- Trix converts spaces to `&nbsp;` and newlines to `br`
- the `safe_html` transform then also applies some conversions, that's why break is included multiple times

Closes #1881.


deiferni:
@phgross please have a look

opengever/base/widgets.py
```diff
@@ -53,4 +75,8 @@ def __init__(self, field, widget):
 
     def toFieldValue(self, value):
         safe_html = self.transformer.convert('safe_html', value).getData()
-        return super(TrixDataConverter, self).toFieldValue(safe_html)
+        # transform may return non-unicode empty string which raises validation
+        # errors on the field
+        safe_html = safe_html or u''
+        field_value = self.field.fromUnicode(safe_html or u'')
```
phgross: Fallback für safe_html ist doppelt ...

phgross:
Sonst 👍

deiferni:
@phgross updated

Issue #1906, opened 2016-06-09
Document-Overview broken


deiferni:
When i open the document overview nothing is rendered but there is a js backtrace:

![screen shot 2016-06-09 at 15 48 18](https://cloud.githubusercontent.com/assets/736583/15931845/a286510c-2e59-11e6-9110-2ad51c107c49.png)

PR #1907, opened 2016-06-09
Fix document overview.

bierik:
Fixes https://github.com/4teamwork/opengever.core/issues/1906

`NodeList` is not iterable so the `forEach` does not work here.
Convert the `NodeList` to an iterable Array to have the `forEach`.


deiferni:
👍

PR #1915, opened 2016-06-13
Update bumblebee documents after checkin and revert.

### LINKED ISSUE #1914: Storing not queued after document checkin/revert to version
There seems to be no `IObjectModifiedEvent` is fired. We should also register event handler for custom checking/revert events.

deiferni:
Register bumblebee event handler to also listen to checkin and revert events which don't fire an object-modified event.

Fixes #1914.


deiferni:
@phgross plz have a look

opengever/document/tests/test_bumblebee.py
```diff
@@ -43,8 +48,15 @@ def test_document_preview_is_hidden(self, browser):
 
 class TestBumblebeeIntegrationWithEnabledFeature(FunctionalTestCase):
 
+    maxDiff = None
```
lukasgraf: ✂️

PR #1923, opened 2016-06-15
Add lock-info and checkout-info viewlets to bumblebee-overlay.

elioschmutz:
Dieser PR fügt das Lock-Info Viewlet sowie das Checkout-Info-Viewlet dem Bumblebee-Overlay hinzu.

bildschirmfoto 2016-06-15 um 17 12 37

bildschirmfoto 2016-06-15 um 17 12 56

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.9.0 (unreleased)
 ------------------
 
+- Add lock-info and checkout-info viewlets to bumblebee-overlay.
+  [elioschmutz]
+
 - Lock excerpts in the dossiers and sync updates to the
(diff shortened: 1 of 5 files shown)


deiferni:
Id doesn't look pretty but it gets the job done 😉 . I'd say @bierik can improve the looks of this in a separate PR.

As discussed instantiating the viewlets directly is ok for me (and an easier solution than iterating over viewlet-managers and digging them out there, especially since these are our viewlets anyway), so unless @lukasgraf or @phgross disagrees i'll merge this one.

phgross:
Ist für mich i.0., bitte rebasen ...

deiferni:
👍

PR #1925, opened 2016-06-16
Rename build / publish docs script to avoid conflicts with the bin/instance script.

phgross:
@deiferni or @lukasgraf


deiferni:
thanks 👍

PR #1949, opened 2016-06-20
Getting default with AQ: Change lookup order so adapting context to schema interface is tried first

lukasgraf:
Getting default with AQ: Change lookup order so that we first try to properly adapt the context to the schema interface with `field.get(ISchema(obj))`, and only if that fails, try a simple `field.get(obj)`.

Previously this was reversed, and lead to `field.get(obj)` returning bogus values with the `DexterityContent.__getattr__` from `plone.dexterity == 2.3.0`.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,10 @@ Changelog
 4.9.0 (unreleased)
 ------------------
 
+- Getting default with AQ: Change lookup order so that we first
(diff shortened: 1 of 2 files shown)


lukasgraf:
Diff is 😵 💩  - basically it's just that the order of the `field.get(obj)` vs `field.get(adpt)` lookups got switched 😉

PR #1959, opened 2016-06-22
Fixed icon tooltip in document listings.

phgross:
With the bumblebee integration all document listings, gets wrapped in to a proxy view. Therefore the tooltip condition has to be changed to the proxy identifier as well.

@deiferni


deiferni:
👍

PR #1963, opened 2016-06-23
Don't truncate strings in sentry extra data.

deiferni:
We currently need full string length for errors reported by the
trix2sablon conversion.

### DIFF
--- a/opengever/base/transforms/trix2sablon.py
+++ b/opengever/base/transforms/trix2sablon.py
@@ -22,6 +22,9 @@ VALID_TAGS = {
 }
 
 
+SERVER_SIDE_STRING_MAX_LENGTH = 2**14
+
+
 _transform = SafeHTML(name='trix_to_sablon', valid_tags=VALID_TAGS)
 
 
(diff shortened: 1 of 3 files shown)


opengever/base/transforms/trix2sablon.py
```diff
@@ -55,4 +55,5 @@ def _log_unexpected_conversion_to_sentry(converted, markup):
         'Unexpected html during trix/sablon conversion',
         request=request,
         url=request.get('ACTUAL_URL', ''),
-        extra=extra)
+        extra=extra,
+        string_max_length=None)
```
lukasgraf: Can you set this to a very high limit instead of `None` please? If a user decides to paste a 20MB doc from Word, we simply don't want to send that to Sentry. So the limit should be reasonably high, but not unbounded.

lukasgraf:
+ changelog. Otherwise looks good to me 👍

deiferni:
@lukasgraf updated

lukasgraf:
👍

PR #1965, opened 2016-06-23
Use non self-closing br tags in trix transform.

deiferni:
Trix always renders breaks as simple `<br>`, thus we change the server-side transform to render the same markup.


deiferni:
@lukasgraf plz have a look

lukasgraf:
👍

PR #1970, opened 2016-06-27
Add helper script to generate JSON files for component import into Weblate.

lukasgraf:
This script is used to automate the component import into Weblate, where each `opengever.*` package (with translations) corresponds to one Weblate component.

*(Tests skipped with `[ci skip]`, this script isn't covered by tests)*

**Example usage:**

```bash
python create-component-import-json.py https://4teamworkTranslator:[access-token]@github.com/4teamwork/opengever.core.git
```

**Output:**

Two files, `main_component.json` and `components.json`.

**These can then be imported into Weblate like so:**

```bash
PROJ="<your_project_slug>" && python manage.py import_json --project ${PROJ} main_component.json && python manage.py import_json --project ${PROJ} --main-component opengever-activity components.json
```
(Where `<your_project_slug>` is something like `onegov-gever`)

@phgross @deiferni

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,10 @@ Changelog
 4.9.0 (unreleased)
 ------------------
 
+- Add helper script to generate JSON files for component import
+  into Weblate.
+  [lgraf]
+
 - Remove i18n markup for strings in opengever.setup:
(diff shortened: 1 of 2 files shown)


deiferni:
👍

PR #1972, opened 2016-06-27
Adjust decision in TranslatedTitleBrain when the the language specific title should be used.

phgross:
Decide with the help of the brains portal_type if the brain should use the language specific title or not.

Because `object_provides` is only a index but not a metadata, we have to check the portal_types FTI if the object supports TranslatedTitle or not. 

To avoid performance Issues the `support_translatedtitle` check is cached.

@lukasgraf as discussed.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,11 @@ Changelog
 4.9.0 (unreleased)
 ------------------
 
+- Adjust decision in TranslatedTitleBrain if the the language specific title
+  should be used or not. It checks now if the portal_type supports TranslatedTitle,
+  by making a lookup in the fti with the help of the brain's portal_type.
+  [phgross]
+
 - Add helper script to generate JSON files for component import
   into Weblate.
   [lgraf]
--- a/opengever/base/brain.py
+++ b/opengever/base/brain.py
@@ -1,5 +1,17 @@
+from opengever.base.behaviors.translated_title import ITranslatedTitle
 from opengever.base.utils import get_preferred_language_code
-import Missing
+from plone import api
+from plone.memoize import ram
+
+
+def cache_key(m, portal_type):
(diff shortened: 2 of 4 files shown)


docs/HISTORY.txt
```diff
@@ -5,6 +5,11 @@ Changelog
 4.9.0 (unreleased)
 ------------------
 
+- Adjust decision in TranslatedTitleBrain if the the language specific title
+  should be used or not. It checks now if the portal_type supports TranslatedTitle.
+  By making a lookup in the fti with the help of brains portal_type.
```
deiferni: . By -> by
of brains -> of the brain's

opengever/base/brain.py
```diff
@@ -1,5 +1,17 @@
...
+from plone import api
+from plone.memoize import ram
+
+
+def cache_key(m, portal_type):
+    return portal_type
+
+
+@ram.cache(cache_key)
+def support_translatedtitle(portal_type):
```
deiferni: supports_translated_title

opengever/base/tests/test_translated_title.py
```diff
@@ -126,7 +154,7 @@ def test_Title_on_brains_returns_title_in_preferred_language(self, browser):
                           obj2brain(repository_root).Title)
 
     @browsing
-    def test_Title_on_brains_use_Title_as_fallback_when_no_language_title_exists(self, browser):
+    def test_Title_on_brains_use_Title_when_type_does_not_support_translatedtitle(self, browser):
```
deiferni: test_Title_on_brains_uses_Title_when_type_does_not_support_translated_title

deiferni:
👍

phgross:
@deiferni updated ...

lukasgraf:
Looks good! 👍

PR #1987, opened 2016-07-01
Create SQL sample contacts in the opengever.examplecontent profile

phgross:
- Creates 100 individual person and 30 organisations when adding a new GEVER with the `examplecontent` profile.

@deiferni


opengever/contact/models/address.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Address(Base):
+    __tablename__ = 'addresses'
+
+    address_id = Column('id', Integer, Sequence('adresses_id_seq'),
+                        primary_key=True)
+    contact_id = Column('contact_id', Integer, ForeignKey('contacts.id'),
+                        primary_key=True)
+    contact = relationship("Contact", back_populates="addresses")
+    label = Column(String(256))
```
lukasgraf: `255` ❗❗❗

opengever/contact/models/url.py
```diff
@@ -0,0 +1,19 @@
+from opengever.base.model import Base
+from sqlalchemy import Column
+from sqlalchemy import ForeignKey
+from sqlalchemy import Integer
+from sqlalchemy import String
+from sqlalchemy.orm import relationship
+from sqlalchemy.schema import Sequence
+
+
+class Url(Base):
```
lukasgraf: `Url` -> `URL` (PEP8)

opengever/examplecontent/data/contacts.json
```diff
@@ -0,0 +1,1002 @@
+[
+  {
+    "title": "Ms.",
+    "street": "Binzmühlestrasse 135",
+    "city": "Sapin Haut",
+    "zip_code": 1907,
+    "mail": "SlainieCoupart@teleworm.us",
```
lukasgraf: Don't use real domains or phone numbers, that's a big no-no and a sure way to piss people off.

Use RFC 2606 sample domains (`example.org`, `example.com`, ... Subdomains of those if needed). For swiss phone numbers, I would use the `099` prefix.

opengever/examplecontent/contacts.py
```diff
@@ -0,0 +1,87 @@
...
+            self.add_mail(item, person, ADDRESS_LABELS)
+
+            org_role = OrgRole(person=person,
+                               organisation=random.choice(organisations))
+            self.db_session.add(org_role)
+
+    def add_address(self, item, contact, labels):
+        address = Address(label=random.choice(labels),
+                          street=item['street'],
+                          zip_code=str(item['zip_code']),
```
lukasgraf: Why this indirection? Why not just have the ZIP code as a string in the example data JSON?

opengever/examplecontent/data/contacts.json
phgross: oh yep didn't realized that they're real domains 😏 , I'll change that.

opengever/examplecontent/hooks.py
```diff
@@ -203,4 +204,7 @@ def municipality_content_profile_installed(site):
     creator = MeetingExampleContentCreator(site)
     creator.create_content()
 
+    creator = ExampleContactCreator(object)
```
deiferni: object -> site

deiferni: uhm you don't seem to use the site 😉  so maybe just omit it completely

opengever/examplecontent/contacts.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def __init__(self, site):
+        self.site = site
+        self.db_session = create_session()
+
+    def load_contacts(self):
+        path = resource_filename(
+            'opengever.examplecontent', 'data/contacts.json')
+        _file = open(path, 'r')
+        return json.loads(_file.read())
```
deiferni: i think you could use `json.load(_file)`

opengever/examplecontent/contacts.py
```diff
@@ -24,27 +26,62 @@ def load_contacts(self):
         _file = open(path, 'r')
         return json.loads(_file.read())
 
+    def load_organisations(self):
+        path = resource_filename(
+            'opengever.examplecontent', 'data/organisations.json')
+        _file = open(path, 'r')
+        return json.loads(_file.read())
```
deiferni: Same here, use `json.load(_file)`

opengever/examplecontent/contacts.py
```diff
@@ -24,27 +26,62 @@ def load_contacts(self):
         _file = open(path, 'r')
         return json.loads(_file.read())
 
+    def load_organisations(self):
```
deiferni: as suggested by @lukasgraf: for the sake of consistency `load_organizations`

deiferni:
👍

PR #1988, opened 2016-07-01
CSRF: Whitelist context portlet assignment annotations.

lukasgraf:
Fixes CSRF false positives on Portlet contexts like `vorlagen` or `ìnbox` when visiting them for the first time after site setup.

Fixes https://sentry.4teamwork.ch/sentry/onegov-gever/issues/806/

@phgross


phgross:
👍

PR #1993, opened 2016-07-05
Don't allow protocols for a meeting to be generated twice.

### LINKED ISSUE #1992: Duplicate protocol generation possible
Currently protocol can be generated multiple times by quickly opening the GET-view. This leads to one protocol being connected to the meeting, the other ones are disconnected. This leads to the locking viewlet being broken for all but the latest protocols.

We should lock the meeting for update when generating protocols and then prevent adding more than one protocol for a meeting.

See https://sentry.4teamwork.ch/sentry/onegov-gever/issues/757/

deiferni:
Lock the meeting row with `SELECT FOR UPDATE` while the protocol is being generated and check if the protocol is already present before generating the document from its template. This prevents multiple threads from generating a protocol twice for the same meeting.

Fixes #1992.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.10.0 (unreleased)
 -------------------
 
+- Don't allow protocols for a meeting to be generated twice.
+  [deiferni]
+
 - Configure offset for showroom overlay.
   [deiferni]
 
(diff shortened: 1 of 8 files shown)


deiferni:
@lukasgraf or @phgross plz have a look

lukasgraf:
Looks good to me! 👍

PR #1994, opened 2016-07-05
Don't store documents that are not digitally available in bumblebee.

### LINKED ISSUE #1976: Don't register non digitally available documents with bumblebee
Those will never ever have a preview, so we should not register them with bumblebee in the first place.

deiferni:
This has not happened anyway so far due to a side-effect of another statement. This change makes it explicit and adds tests, though.

Closes #1976.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.10.0 (unreleased)
 -------------------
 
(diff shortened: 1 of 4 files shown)


deiferni:
@phgross or @lukasgraf plz have a look

phgross:
👍  please rebase

lukasgraf:
👍

Issue #1778, opened 2016-05-02
Persist dexterity default values


lukasgraf:
The way Dexterity works, default values are sometimes not persisted in the DB, but instead dynamically generated on display.

This sometimes happens when creating content TTW, and nearly always when creating content programmatically (via `createContentInContainer()` for example). Default values not being persisted leads to values for an object suddenly changing retroactively once a default value (factory) is changed, which is unacceptable for a records management system.

How exactly default values are generated and possibly persisted is a result of complex interactions between (at least) these components:

- `zope.schema`
- `z3c.form`
- `plone.dexterity`
- `plone.behavior`

lukasgraf:
#### Pathological cases:

**1)**
Given:
- Field in base schema or a behavior using **AttributeStorage**
- Field has a `default` or a `defaultFactory`

Steps to reproduce:
- Add a new object using the `z3c.form` add form, leaving the field at its prefilled default value (Add & immediately save)
- Change the field's `default` or `defaultFactory` in the code

=> When displaying the field or reading it programmatically, the new default will be returned, indicating that the old default has never been persisted

---

This happens because `z3c.form` checks whether a field has been changed or not when applying changes, and only saves values that are different (see [`z3c.form.form.applyChanges()`](https://github.com/zopefoundation/z3c.form/blob/3.0.2/src/z3c/form/form.py#L45-L46)).

Because `plone.dexterity`'s [`DexterityContent` defines a `__getattr__` that falls back to a fields default instead of raising an `AttributeError`](https://github.com/plone/plone.dexterity/blob/2.1.2/plone/dexterity/content.py#L193-L205), for a newly created object, accessing an attribute that doesn't exist yet will result in the fields default value.

This will lead `z3c.form` to comparing the field's default value (prefilled in the form's widget) with the field's default value (returned by Dexterity's fallback when accessing a non-existent attribute). Because of that the field's value is never considered "changed", and therefore won't be written to the object.

On the surface things appear to have worked though, because Dexterity's fallback also will dynamically return the default on display / programmatic read access.


**2)**
Given:

- Field in a behavior using **AnnotationStorage**
- Field `default` that is equal to field's `missing_value`

Steps to reproduce:
- Add a new object using the `z3c.form` add form, leaving the field at its prefilled default value (Add & immediately save)
- Change the field's `default` or `defaultFactory` in the code

=> When displaying the field or reading it programmatically, the new default will be returned, indicating that the old default has never been persisted

---

[...]

PR #1996, opened 2016-07-05
Make sure Dexterity default values are persisted

lukasgraf:
This addresses several issues described in #1778:

These changes make sure that
- default values are actually persisted when adding content **TTW with `z3c.form`**
- default values are set when using **`createContentInContainer`**
- default values are set when using **`invokeFactory`**

(Will add / clean up changelog once merged into `master`

@deiferni @phgross

### DIFF
--- a/opengever/base/default_values.py
+++ b/opengever/base/default_values.py
@@ -1,0 +1,234 @@
+"""
+This module contains functions to help fix issues around Dexterity's
+behavior regarding default values.
+
+Default values not being persisted
+----------------------------------
+
+In some cases, default values that have been prefilled in a z3c.form form won't
+get persisted in the DB, causing an object's fields to "retroactively change"
+once the default[Factory] for a field is changed.
+
+Both these cases share a similar pattern: z3c.form checks whether a field's
+value has changed or not, and only writes fields whose values it believes
+have changed. However, that check gets fooled by __getattr__ fallbacks in
+DexterityContent and AnnotationsFactoryImpl that don't raise AttributeError,
+but return the field's default or missing_value respectively.
+
+1) - AttributeStorage
+
+Given:
+
+- Field in base schema or a behavior using AttributeStorage
+- Field has a default or a defaultFactory
+
+Steps to reproduce:
+
+- Add a new object using the z3c.form add form, leaving the field at its
+  prefilled default value (Add & immediately save)
+- Change the field's default or defaultFactory in the code
+
+=> The field's value will not have been persisted.
+
+(When displaying the field or reading it programmatically, the new default
+will be returned, indicating that the old default has never been persisted.)
+
+This happens because z3c.form checks whether a field has been changed or not
+when applying changes, and only saves values that are different
+(see z3c.form.form.applyChanges()).
+
+Because plone.dexterity's DexterityContent defines a __getattr__ that falls
+back to a field's default instead of raising an AttributeError, for a newly
+created object, accessing an attribute that doesn't exist yet will result in
+the fields default value.
+
+This will lead z3c.form to comparing the field's default value (prefilled in
+the form's widget) with the field's default value (returned by Dexterity's
+fallback when accessing a non-existent attribute). Because of that the field's
+value is never considered "changed", and therefore won't be written to the
+object.
+
+On the surface things appear to have worked though, because Dexterity's
+fallback also will dynamically return the default on display / programmatic
+read access.
+
+2) - AnnotationStorage
+
+Given:
+
+- Field in a behavior using AnnotationStorage
+- Field default that is equal to field's missing_value
+
+Steps to reproduce:
+
+- Add a new object using the z3c.form add form, leaving the field at its
+  prefilled default value (Add & immediately save)
+- Change the field's default or defaultFactory in the code
+
+=> The field's value will not have been persisted.
+
+In this case, the value isn't being persisted because there is a similar
+fallback in AnnotationsFactoryImpl: Instead of raising an AttributeError, it
+will fall back to returning the field's missing_value
+(see plone.behavior.annotation.AnnotationsFactoryImpl.__getattr__).
+
+This will result in z3c.form.form.applyChanges() comparing the default value
+for a field against the field's missing_value - if they happen to be the same,
+the field's value will never be persisted.
+
+Approach to fixing the issues
+-----------------------------
+
+We attempt to fix the z3c.form related issues where default values won't be
+persisted by patching z3c.form.util.changedField(). The idea is to patch it in
+a way so that it doesn't just rely on the DataManager to return a field's
+value (which would trigger the mentioned fallbacks), but instead uses the
+function below to really get to the persisted value for a field, taking the
+underlying storage into account. That way we can hopefully avoid any fallbacks
+that would fool the check.
+"""
+
+from Acquisition import aq_base
+from persistent.interfaces import IPersistent
+from plone.behavior.annotation import AnnotationsFactoryImpl
+from plone.dexterity.utils import iterSchemata
+from zope.schema import getFieldsInOrder
+from zope.schema._bootstrapinterfaces import IContextAwareDefaultFactory
+
+
+def get_persisted_value_for_field(context, field):
+    """Return the *real* stored value for a field, avoiding any fallbacks.
+
+    In particular, this circumvents the __getattr__ fallbacks in
+    DexterityContent and AnnotationsFactoryImpl that return the field's
+    default / missing_value.
+
+    Raises an AttributeError if there is no stored value for the field.
+    """
+    if not IPersistent.providedBy(context):
+        raise Exception(
+            "Attempt to get persisted field value for a non-persistent object")
+
+    # AQ unwrap object to avoid finding an attribute via acquisition
+    context = aq_base(context)
+    storage_impl = field.interface(context)
+
+    if isinstance(storage_impl, AnnotationsFactoryImpl):
+        # AnnotationStorage
+        name = field.__name__
+        if name not in storage_impl.__dict__['schema']:
+            raise AttributeError(name)
+
+        annotations = storage_impl.__dict__['annotations']
+        key_name = storage_impl.__dict__['prefix'] + name
+        try:
+            value = annotations[key_name]
+        except KeyError:
+            # Don't do the fallback to field.missing_value that
+            # AnnotationsFactoryImpl.__getattr__ does
+            raise AttributeError(name)
+        return value
+    else:
+        # Assume attribute storage
+        name = field.getName()
+        try:
+            # Two possible cases here: Field in base schema, or field in
+            # behavior with attribute storage. Either way, we look up the
+            # attribute in the objec's __dict__ in order to circumvent the
+            # fallback in DexterityContent.__getattr__
+            value = context.__dict__[name]
+        except KeyError:
+            raise AttributeError(name)
+        return value
+
+
+def get_persisted_values_for_obj(context):
+    values = {}
+    schemas = list(iterSchemata(context))
+    for schema in schemas:
+        fields = getFieldsInOrder(schema)
+        for name, field in fields:
+            try:
+                value = get_persisted_value_for_field(context, field)
+                values[name] = value
+            except AttributeError:
+                continue
+    return values
+
+
+def set_default_values(content, container, values):
+    """Set default values for all fields.
+
+    This is necessary for content created programmatically since dexterity
+    only sets default values in a view.
+
+    Parameters:
+    - content:   The object in creation. Might not be AQ wrapped yet
+    - container: The parent container the object will be added to
+    - values:    Mapping of *actual* values (not defaults) that will be or
+                 have been set on the object (not by us). I.e. kwargs to
+                 invokeFactory or createContentInContainer. Will be taken
+                 into consideration when determining whether defaults should
+                 apply or not.
+
+    """
+    marker = object()
+
+    def object_has_value_for_field(obj, field):
+        """Determine whether a value is persisted on `obj` for `field`.
+        """
+        try:
+            get_persisted_value_for_field(content, field)
+            return True
+        except AttributeError:
+            return False
+
+    def determine_default_value(field, container):
+        """Determine a field's default value during object creation.
+        """
+        # We deliberately ignore form level defaults here.
+
+        # zope.schema defaultFactory
+        #
+        # Based on zope.schema._boostrapfields.DefaultProperty. We don't use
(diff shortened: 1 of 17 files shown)


opengever/base/default_values.py
```diff
@@ -0,0 +1,234 @@
+"""
+This module contains functions to help fix issues around Dexterity's
+behavior regarding default values.
+
+Default values not being persisted
+----------------------------------
+
+In some cases, default values that have been prefilled in a z3c.form form wont
```
deiferni: won't

opengever/base/default_values.py
```diff
@@ -0,0 +1,234 @@
...
+- Add a new object using the z3c.form add form, leaving the field at its
+  prefilled default value (Add & immediately save)
+- Change the field's default or defaultFactory in the code
+
+=> The field's value will not have been persisted.
+
+In this case, the value isn't being persisted because there is a similar
+fallback in AnnotationsFactoryImpl: Instead of raising an AttributeError, it
+will fall back to returning the field's missing_value
+(see plone.behavior.annotation.AnnotationsFactoryImpl. __getattr__).
```
deiferni: whitespace: AnnotationsFactoryImpl. ✂️  \__getattr__

opengever/base/tests/test_default_values_for_types.py
```diff
@@ -0,0 +1,795 @@
...
+
+REPOFOLDER_DEFAULTS = {
+    'creators': (),
+    'description': u'',
+    'public_trial': u'unchecked',
+    'public_trial_statement': '',
+    'reference_number_prefix': u'1',
+    'title_de': DEFAULT_TITLE,
+}
+
```
deiferni: ✂️

opengever/base/tests/test_default_values_for_types.py
```diff
@@ -0,0 +1,795 @@
...
+TASK_DEFAULTS = {
+    'deadline': FROZEN_TODAY + timedelta(days=5),
+    'issuer': u'test_user_1_',
+    'relatedItems': [],
+    'responsible': u'test_user_1_',
+    'responsible_client': u'client1',
+    'title': u'My title',
+}
+
```
deiferni: ✂️

opengever/base/tests/test_default_values_for_types.py
```diff
@@ -0,0 +1,795 @@
...
+            browser.login().open()
+            factoriesmenu.add(u'RepositoryFolder')
+            browser.fill({u'Title': DEFAULT_TITLE}).save()
+            repofolder = browser.context
+
+        persisted_values = get_persisted_values_for_obj(repofolder)
+        defaults = self.get_z3c_form_defaults()
+
+        # XXX: Don't know why this happens
+        defaults['public_trial_statement'] = None
```
deiferni: i just got a bit confused here when reading, i would consider to call this variable `expected`, then set`expected['public_trial_statement'] = None`?

opengever/document/tests/test_document.py
```diff
@@ -252,7 +252,7 @@ def test_default_document_date_is_today(self, browser):
         self.assertEqual(today, document.document_date)
 
     @browsing
-    def test_preserverd_as_paper_default(self, browser):
+    def test_preserverd_as_paper_default_false(self, browser):
```
deiferni: preserverd -> preserved

opengever/document/tests/test_document.py
```diff
@@ -269,6 +269,10 @@ def test_preserverd_as_paper_default(self, browser):
         document = self.dossier['document-1']
         self.assertFalse(document.preserved_as_paper)
 
+    @browsing
+    def test_preserverd_as_paper_default_true(self, browser):
```
deiferni: preserverd -> preserved

opengever/base/tests/test_default_values_for_types.py
```diff
@@ -0,0 +1,795 @@
...
+            browser.find('Save').click()
+            committee = browser.context
+
+        persisted_values = get_persisted_values_for_obj(committee)
+        defaults = self.get_z3c_form_defaults()
+
+        linked_repo = persisted_values.pop('repository_folder')
+        self.assertEqual(linked_repo.to_object, repofolder)
+
+        self.assertDictEqual(defaults, persisted_values)
```
deiferni: Just a thought: `ftw.builder` currently implements its own default-value magic (https://github.com/4teamwork/ftw.builder/blob/master/ftw/builder/dexterity.py#L104) maybe it would be interesting to see that it creates the same default values as the three methods tested here? this would tell us that we use the same and thus correct default values in our tests. What do you think?

opengever/base/tests/test_default_values.py
```diff
@@ -143,3 +144,49 @@ def test_ann_behavior_with_default_equal_missing_value(self, browser):
 
         self.assert_value_persisted(
             field, initial_default=None, new_default=42, browser=browser)
+
+
+class TestDefaultValuePersistenceCreateContent(TestDVPersistenceBase):
+
+    def add_object(self, browser):
+        obj = createContentInContainer(self.portal, 'Dummy')
```
deiferni: `return createContentInContainer(self.portal, 'Dummy')`

opengever/base/tests/test_default_values.py
```diff
@@ -190,3 +190,104 @@ def test_ann_behavior_with_default_equal_missing_value(self):
...
+    PortalFolderBase.
+
+    Just to be sure our monkey patch affects both, we test both cases:
+    Calling invokeFactory on the Portal (this testcase), and calling it on
+    a DX Container (testcase below).
+    """
+
+    def add_object(self, browser):
+        portal = self.layer['portal']
+        obj = portal[portal.invokeFactory('Dummy', 'dummy')]
```
deiferni: `return portal[portal.invokeFactory('Dummy', 'dummy')]`

opengever/base/monkey/patches/default_values.py
```diff
@@ -87,3 +87,136 @@ def __getattr__(self, name):
...
+        from opengever.base.default_values import get_persisted_value_for_field
+        from persistent.interfaces import IPersistent
+        from z3c.form import interfaces
+        from z3c.formwidget.query.widget import QueryContext
+        import zope.schema
+
+        def changedField(field, value, context=None):
+            """Figure if a field's value changed
+
+            Comparing the value of the context attribute and the given value"""
```
deiferni: this docstring looks trapped, i'd add some newline(s) 😉

deiferni:
something i noticed while poking around in a local checkout: `CreateDocumentCommand` https://github.com/4teamwork/opengever.core/blob/master/opengever/base/command.py#L80 also sets default values, maybe it could be simplified or use `set_default_values` from `opengever.base.default_values` now?

opengever/base/tests/test_default_values.py
```diff
@@ -0,0 +1,145 @@
...
+        self.assertEqual(persisted_value, current_value)
+
+
+class TestDefaultValuePersistenceZ3CForm(TestDVPersistenceBase):
+
+    def add_object(self, browser):
+        browser.login().open()
+        factoriesmenu.add('Dummy')
+        browser.click_on('Save')
+        obj = browser.context
```
deiferni: `return browser.context`

hmm i'm starting to think that this is on purpose to improve readability, maybe just ignore my comments 😉

deiferni:
Looks 👍 ✨ 💛  apart from my nagging comments 😉

opengever/base/default_values.py
lukasgraf: Fixed

opengever/base/default_values.py
lukasgraf: Fixed

opengever/base/tests/test_default_values_for_types.py
lukasgraf: Fixed

opengever/base/tests/test_default_values_for_types.py
lukasgraf: Fixed

opengever/base/tests/test_default_values_for_types.py
lukasgraf: Great point - at this point, the values on the object aren't *defaults* any more anyway, they're just (expected) *values*, originating from defaults. Fixed (for all those tests).

opengever/document/tests/test_document.py
lukasgraf: Fixed

opengever/document/tests/test_document.py
lukasgraf: Fixed

opengever/base/tests/test_default_values_for_types.py
lukasgraf: Very good point, I'll definitely look into that. The challenge is, we can't really use the functions from opengever (`determine_default_values` etc..), that are supposed to do the right thing in a variety of circumstances, in `ftw.builder`. So we might end up having to patch builder from opengever 😢 

But regardless, having tests to assert a well-defined behavior regarding default values for builders is a must (even if they just point out and codify any inconsistencies that we haven't been able to fix yet).

opengever/base/monkey/patches/default_values.py
lukasgraf: I agree, but that's the way the docstring was formatted in the original function 😉  And I stuck with the approach of copying the original code as verbatim as possible for monkey patches, in order to make it easier if one eventually want's to diff it (no PEP8 cleanup, no line wrapping, ...).

opengever/base/tests/test_default_values.py
lukasgraf: Indeed 😉  - a temporary variable introduces a tiny bit more meaning (naming things), which is why I often use them even when things would easily fit on one line. Not sure whether it's warranted here, but I usually tend to lean towards the more verbose side.

lukasgraf:
> `CreateDocumentCommand` also sets default values [...]

Excellent point. That should be rewritten using the primitives from `opengever.base.default_values` (and probably tested together with the other methods of creation) - I'll look into that.

lukasgraf:
@deiferni I looked at the `CreateDocumentCommand`, and the way it's implemented, it currently not only sets defaults, but also [sets missing values](https://github.com/4teamwork/opengever.core/blob/master/opengever/base/command.py#L103) if no default can be found. So most objects that get created with that command end up having a value for every field that existed at creation time. If I now change the command to use `determine_default_values()`, that behavior will be altered and *only* default values will be set.

I'd propose to address this in a follup up PR, and also re-evaluate whether we don't want the behavior of setting values for *all* fields for all creation methods.

deiferni:
@lukasgraf regarding `CreateDocumentCommand`: sounds 👍  could you create an issue for this so we don't forget it?

opengever/base/tests/test_default_values.py
```diff
@@ -0,0 +1,315 @@
...
+        return obj
+
+    @unittest.skip("Not fixed yet")
+    @browsing
+    def test_ann_behavior_with_default_equal_missing_value(self, browser):
+        # This test fails because the persisted value is equal to missing
```
phgross: I would move this comment in to the `@unittest.skip(`

opengever/base/tests/test_default_values_for_types.py
```diff
@@ -0,0 +1,793 @@
...
+MAIL_FORM_INITVALUES = {
+    'delivery_date': None,
+    'document_type': None,
+    'foreign_reference': None,
+}
+
+
+TASK_DEFAULTS = {
+    'deadline': FROZEN_TODAY + timedelta(days=5),
+    'issuer': u'test_user_1_',
```
phgross: `TEST_USER_ID`

`from plone.app.testing import TEST_USER_ID`

opengever/base/tests/test_default_values_for_types.py
```diff
@@ -0,0 +1,793 @@
...
+
+    def test_create_content_in_container(self):
+        with freeze(FROZEN_NOW):
+            reporoot = createContentInContainer(
+                self.portal,
+                'opengever.repository.repositoryroot',
+                title_de=DEFAULT_TITLE)
+
+        persisted_values = get_persisted_values_for_obj(reporoot)
+        expected = self.get_type_defaults()
```
phgross: I think the indirection over the `type_defaults` getter is not necessary. I would say the following example would be shorter and easier to understand.

```
self.assertDictEqual(self.type_defaults, get_persisted_values_for_obj(reporoot))
```

opengever/contact/contact.py
```diff
@@ -187,10 +187,14 @@ def title(self):
             title = _(u'Contact')
         return title
 
+    @title.setter
```
phgross: are both of the setters necessary ?

opengever/testing/profiles/testing/types/Dummy.xml
```diff
@@ -0,0 +1,51 @@
...
+  <property name="schema">opengever.testing.types.IDummySchema</property>
+  <property name="klass">opengever.testing.types.Dummy</property>
+
+  <property name="behaviors">
+    <element value="opengever.testing.types.IDummyAnnotationStorageBehavior" />
+    <element value="opengever.testing.types.IDummyAttributeStorageBehavior" />
+  </property>
+
+  <!-- View information -->
+  <property name="link_target"></property>
```
phgross: Are all the view informations, method aliases and actions necessary for the tests? otherwise I would recommend to remove them.

phgross:
Just some minor things, otherwise excellent work ✨, thanks a lot!

opengever/base/tests/test_default_values_for_types.py
lukasgraf: It's maybe not needed in this case, but in several other places the contents of `expected` need to be modified slightly. Given that, I think it's less confusing to have the same style across all those tests.

opengever/contact/contact.py
lukasgraf: No, actually not attempting to set the `title` in the test works 👍  Fixed.

opengever/testing/profiles/testing/types/Dummy.xml
lukasgraf: Nope - removed.

opengever/base/tests/test_default_values_for_types.py
lukasgraf: Great catch! Fixed.

opengever/base/tests/test_default_values.py
lukasgraf: Fixed.

lukasgraf:
@phgross updated

PR #2015, opened 2016-07-11
Fixed contact SQLModel creation upgradestep.

### LINKED ISSUE #2008: Upgrade step 20160630155228_add_sql_model_for_new_contact_implementation broken
The upgrade won't work in combination with another upgrade step that adds more tables. It must be rewritten and define all its tables explicitly.

phgross:
Because the profile `opengever.contact` is already installed, we had to hardcode the current model in to the upgradestep, otherwise the upgradestep doesn't work with feature changes on the model.

Fixes #2008 

@deiferni or @lukasgraf


deiferni:
👍

PR #2017, opened 2016-07-11
Consequently use builders to create contact sql objects in tests.

### LINKED ISSUE #2013: Person unit-tests miss purpose
sry i should have noticed during a review 😞 , but the way these tests are implemented they wont detect errors since the objects are never added to a session/flushed, see for example: https://github.com/4teamwork/opengever.core/commit/27e6410c6395cd0f9f754dec843a0bca427bdd71#diff-985919f5fb6854b97e7611e942c669b4R15

we should at least add the created objects to the session and flush it, 
or even better use a builder

phgross:
Fixes #2013.

@deiferni


deiferni:
👍

PR #2018, opened 2016-07-11
Use qtip2 for document preview tooltip

### LINKED ISSUE #1997: Tooltip position improvable
The tooltip is positioned relatively to document link at the moment. this does not work very well vor long  document titles since they can be partially hidden by another table column: 

Slightly too long title, can result in an "unreachable" tooltip. The tooltip disappears when the mouse leaves the the title column, when users move the mouse a bit too slow the tooltip won't be accessible any more. This could be very annoying:

hahahaha

A very, very long title positions the overlay on the left side:

screen shot 2016-07-06 at 11 53 59

phgross:
- Remove no longer used `ftw.tooltip` dependency and customizations.
- Extract document tooltip from document link widget to a separate view.
- Load tooltip data asynchronous
- Wrap document link in to tooltip trigger div, to fixes #1997 

Styling in separate Issue in plonetheme.teamraum

tooltip

@deiferni or @lukasgraf


phgross:
@bierik would be nice if you could review the JS part (`tooltip.js`)

docs/HISTORY.txt
```diff
@@ -5,6 +5,14 @@ Changelog
 4.10.0 (unreleased)
 -------------------
 
+- Improve document tooltip:
+
+  - Replace ftw.tooltip implementation with our own javascript.
+  - Load tooltip data asynchronous.
```
deiferni: ..ly.

opengever/base/upgrades/20160708150759_add_tooltip_js/jsregistry.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<object name="portal_javascripts" meta_type="JavaScripts Registry">
+
+  <javascript cacheable="True" compression="safe" cookable="True" expression=""
+              enabled="True" id="++resource++opengever.base/tooltip.js" inline="False"
```
deiferni: please add `insert-after="..." ` it seems to have a dependency on jquery

opengever/base/profiles/default/jsregistry.xml
```diff
@@ -60,4 +60,8 @@
                 expression="" id="++resource++opengever.base/datetimepicker.js" inline="False"
                 insert-after="++resource++plone.app.jquery.js" />
 
+    <javascript cacheable="True" compression="safe" cookable="True" expression=""
+                enabled="True" id="++resource++opengever.base/tooltip.js" inline="False"
```
bierik: The `tooltip.js` depends on the base Controller. So use `insert-after` to make the tooltiip explicitly depending on the base Controller.

opengever/base/browser/resources/tooltip.js
```diff
@@ -0,0 +1,111 @@
+(function(global, $) {
+
+  global.Controller = Controller;
+
+  function timestamp() { return new Date().getTime(); }
```
deiferni: imho i don't like these one-line methods, for me they do not improve readability but instead are hard to "parse" at first glance

opengever/base/browser/resources/tooltip.js
```diff
@@ -0,0 +1,111 @@
...
+
+  function throttle(func, wait, options) {
```
bierik: Place a comment here that the function has been copied from `https://github.com/jashkenas/underscore`

opengever/base/browser/resources/tooltip.js
```diff
@@ -0,0 +1,111 @@
...
+
+    return throttled;
+  };
+
+  function TooltipController(options) {
+
+    options = $.extend({predelay: 500}, options);
+
+    this.settings = {
+      delay: 100,
```
bierik: How about making the delay configurable through the options from the tooltip controller.

opengever/base/browser/resources/tooltip.js
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    Controller.call(this);
+
+    this.init = function() {
+      this.registerActions();
+    };
+
+    this.fetch = function() {
+      return $.get(this.target.data('tooltip-url')) };
```
bierik:
``` javascript
this.fetch = function() {
  return $.get(this.target.data('tooltip-url'))
};
```

opengever/base/browser/resources/tooltip.js
```diff
@@ -0,0 +1,111 @@
...
+    ];
+
+    this.init();
+  }
+
+  global.TooltipController = TooltipController;
+
+
+  $(function() {
+    var tooltipcontroller = new global.TooltipController();
```
bierik:
```javascript
var tooltipcontroller = new TooltipController();
```

opengever/base/browser/resources/tooltip.js
deiferni: could you add a docstring that describes the behaviour of this method and its options?

opengever/base/browser/resources/tooltip.js
```diff
@@ -0,0 +1,111 @@
...
+      clearTimeout(timeout);
+      previous = 0;
+      timeout = context = args = null;
+    };
+
+    return throttled;
+  };
+
+  function TooltipController(options) {
+
```
deiferni: could you add a docstring here, too

bierik:
@phgross Check the semicolons in your JS code. I didn't mark every missing semicolon.

opengever/base/browser/resources/tooltip.js
```diff
@@ -0,0 +1,111 @@
+(function(global, $) {
+
+  global.Controller = Controller;
```
deiferni: ?

deiferni:
we should also drop `ftw.tooltip` from kgs

deiferni:
As discussed i'm still unhappy about the tooltip position: 

- it is positioned away from the mouse, long way
- it is hard to reach, when i leave the column it will disappear
- it will also appear when i hover the column white space

tooltip

bierik:
⚠️  Depends on https://github.com/4teamwork/plonetheme.teamraum/pull/443

I hijacked the branch and we now use the qtip library for generating the tooltip

See http://qtip2.com/

The library is able to load the preview asynchronously.
The tooltip will be placed next to the cursor as long as it fits the screen.
A delay of 300ms is set for both opening and closing the tooltip.

Perfect Fit:
bildschirmfoto 2016-07-18 um 14 57 41

Bottom touched:
bildschirmfoto 2016-07-18 um 14 58 00

Right side touched:
bildschirmfoto 2016-07-18 um 14 58 15

Bottom-right corner touched:
bildschirmfoto 2016-07-18 um 14 58 25

@phgross pls have a look

bierik:
@phgross updated

lukasgraf:
Positioning behavior looks good to me overall 👍 

Delay of 300ms is spot on, I think that's ideal for getting to the tooltip quickly, but also getting rid of it quickly enough. 👍 

If you're on the document overview, and trigger the tooltip on one of the recent documents (rightmost box), there's still a chance for the issue with overlapping the links below, but I don't see an easy way to fix this, unless we emulate context-menu behavior where the menu pops out inwards if it's too close to the screen edges.

opengever/document/widgets/tooltip.py
```diff
@@ -0,0 +1,42 @@
+from opengever.document.browser.download import DownloadConfirmationHelper
+from plone.app.contentlisting.interfaces import IContentListingObject
+from plone.protect.utils import addTokenToUrl
+from zope.publisher.browser import BrowserView
+import pkg_resources
+
+
+try:
```
deiferni: please use `opengever.base.pdfconverter.is_pdfconverter_enabled`

deiferni:
👍

phgross:
@deiferni updated 
@lukasgraf the issue with the overlapped links is known, but we don't found a solution in a reasonable time. In consideration of the fact, that we try to replace the tooltip functionality some how, it makes no sense to invest a lot of time to search a solution for this problem, ok?

PR #2023, opened 2016-07-12
Add comprehensive tests for fields using restricted vocabularies

lukasgraf:
This adds very detailed tests for the fields from `IClassification` and `ILifeCycle` behaviors that use **restricted vocabularies**.

This is in preparation for [untangling](http://tclhost.com/c7F6jTL.gif) and refactoring the [restricted vocabulary situation](https://github.com/4teamwork/opengever.core/blob/master/opengever/base/behaviors/utils.py#L37-L236).

@deiferni @phgross


opengever/base/tests/test_classification.py
```diff
@@ -0,0 +1,638 @@
...
+        # Start with a loose classification
+        self.set_classification(self.repofolder, u'unprotected')
+
+        browser.login().open(self.repofolder)
+        factoriesmenu.add(u'Business Case Dossier')
+        browser.fill({'Title': 'My Dossier'}).save()
+        dossier = browser.context
+
+        value = self.get_classification(dossier)
+        # Dossier should have inherited classification from repofolder
```
deiferni: you could use these comments as assertion msgs like so:
```python
self.assertEqual(u'unprotected', value, msg="Dossier should have inherited classification from repofolder")
```

opengever/base/tests/test_classification.py
```diff
@@ -0,0 +1,638 @@
...
+        browser.open(self.repofolder, view='edit')
+        browser.fill({'Classification': 'unprotected'}).save()
+        transaction.commit()
+
+        value = self.get_classification(dossier)
+        self.assertEqual(u'confidential', value)
+
+    @browsing
+    def test_propagation_is_depth_limited(self, browser):
+        """Propagation of classification is depth limited to 2 levels.
```
deiferni: maybe we should file an issue for this?
as discussed earlier this week, it most certainly is wrong and only implementet that way due to unclear documentation

opengever/base/tests/test_lifecycle.py
```diff
@@ -1,242 +1,866 @@
...
+            set(form_field.options_values))
+
+
+class TestCustodyPeriodPropagation(FunctionalTestCase):
+
+    def setUp(self):
+        super(TestCustodyPeriodPropagation, self).setUp()
+        self.repofolder = create(Builder('repository'))
+        self.field = ILifeCycle['custody_period']
+        self.grant('Manager')
```
deiferni: why is this necessary?

opengever/base/tests/test_classification.py
```diff
@@ -0,0 +1,638 @@
...
+            set(form_field.options_values))
+
+
+class TestClassificationPropagation(FunctionalTestCase):
+
+    def setUp(self):
+        super(TestClassificationPropagation, self).setUp()
+        self.repofolder = create(Builder('repository'))
+        self.field = IClassification['classification']
+        self.grant('Manager')
```
deiferni: Why is this necessary?

opengever/base/tests/test_lifecycle.py
```diff
@@ -1,242 +1,866 @@
...
-        super(TestRetentionPeriod, self).setUp()
-        add_languages(['de-ch'])
+        super(TestRetentionPeriodDefault, self).setUp()
+        self.repofolder = create(Builder('repository'))
+        self.field = ILifeCycle['retention_period']
 
-        self.grant('Administrator')
-        self.repo = create(Builder('repository')
-                           .having(retention_period=15))
+    def set_is_restricted(self, value):
```
deiferni: you could extract this as a testing helper function, i think it does not need to be a method

deiferni:
some comments, otherwise 👍 ✨

opengever/base/tests/test_classification.py
lukasgraf: @deiferni I personally don't like those at all, because they effectively suppress the **actual** value from the test failure output, unless one makes the effort to include `actual` and `expected` values in the assertion message by hand (string formatting), at which point those assertions are going to span multiple lines, and add a lot of noise.

Given that this is never being done, and the test runner has the assertion message *replace* the default assertion failure message, the actual value is effectively stripped from the output, and can only be inspected by setting a pdb breakpoint, printing it, etc.  Particularly for CI tests on Jenkins that isn't an option.

Long story short: If you've got a test failure, you'll need to look at the actual code for that test anyway, and if you got comments in the code describing more details about the intention, *and* the details mismatch of actual vs. expected in the test failure output, you've got both pieces of information, whereas with custom assertion messages, you're missing one of them.

opengever/base/tests/test_classification.py
lukasgraf: Issue #2027

opengever/base/tests/test_lifecycle.py
lukasgraf: Because otherwise I can't get editing of repofolders with `ftw.testbrowser` to work. In theory `self.grant('Administrator')` should be enough, but for some reason it isn't (still getting `Unauthorized`).

opengever/base/tests/test_classification.py
lukasgraf: Same as above, getting `Unauthorized` otherwise.

opengever/base/tests/test_lifecycle.py
lukasgraf: Done.

lukasgraf:
@phgross updated with changes according to Deif's comments.

opengever/base/tests/test_classification.py
```diff
@@ -0,0 +1,638 @@
...
+                           filter_content_types=False)
+        fti.behaviors = (
+            'opengever.base.behaviors.classification.IClassification',)
+        self.portal.portal_types._setObject('ClassificationFTI', fti)
+        fti.lookupSchema()
+        transaction.commit()
+
+    @browsing
+    def test_classification_behavior(self, browser):
+        # Defaul view doesnt work for system users
```
phgross: typo ...

lukasgraf: @phgross that's the test that already existed, I didn't touch that at all (just renamed the file for constistency). This test can eventually be removed, I'm just leaving it in while refactoring the restricted vocabularies.

opengever/base/tests/test_classification.py
```diff
@@ -0,0 +1,638 @@
...
+        self.assertDictContainsSubset({'type': 'Public trial modified',
+                                       'visible': True},
+                                      entry['action'])
+
+    @browsing
+    def test_public_trial_edit_form_journalizes_changes(self, browser):
+        browser.login().open(self.document, view='edit_public_trial')
+        browser.fill({'Public Trial': 'public'}).save()
+
+        journal = JournalHistory(self.document, self.document.REQUEST)
```
phgross: There is a assertion_method on the FunctionalTestCase (`assert_journal_entry`), please use this ...

phgross:
Ok, but keep in mind to fix the typo and use the assertion method, if the tests aren't removed during the refactoring ...

PR #2043, opened 2016-07-21
Add history model for the new sql contacts.

### LINKED ISSUE #2039: Modell Versionierung
Die Kontakte sollen versioniert werden. Dies wird mit einer separaten `history` Tabelle pro versioniertem Inhaltstyp implementiert.

contacts erd

deiferni:
Depends on https://github.com/4teamwork/plonetheme.teamraum/pull/455

Add history model and tables for contacts (persons and organizations), addresses, phonenumbers, mailaddresses and urls. 

Also add simple views for a person and organization history (Sample data is generated randomly when importing the examplecontent profile):

screen shot 2016-08-16 at 15 17 36
screen shot 2016-08-16 at 15 14 48

Closes #2039.


deiferni:
@phgross or @lukasgraf plz have a look

opengever/contact/tests/test_organizationhistory_unit.py
```diff
@@ -0,0 +1,33 @@
+from ftw.builder import Builder
+from ftw.builder import create
+from opengever.contact.models import OrganizationHistory
+from opengever.testing import MEMORY_DB_LAYER
+import unittest2
+
+
+class TestPerson(unittest2.TestCase):
```
lukasgraf: This should probably be `TestOrganizationHistory`? (Same for some other TestCase names below)

opengever/contact/tests/test_organizationhistory_unit.py
```diff
@@ -0,0 +1,33 @@
...
+        history1 = create(Builder('organizationhistory')
+                          .having(name=u'ACMOT corporation',
+                                  contact=organization))
+        history2 = create(Builder('organizationhistory')
+                          .having(name=u'ACMTT corporation',
+                                  contact=organization))
+
+        self.assertEqual([history1, history2], organization.history)
+
+    def test_is_contacthistory(self):
```
lukasgraf: -> `test_is_ organizationhistory`

opengever/contact/tests/test_organizationhistory_unit.py
```diff
@@ -0,0 +1,33 @@
...
+from opengever.contact.models import OrganizationHistory
+from opengever.testing import MEMORY_DB_LAYER
+import unittest2
+
+
+class TestPerson(unittest2.TestCase):
+
+    layer = MEMORY_DB_LAYER
+
+    def test_organization_can_have_multiple_histories(self):
```
lukasgraf: Hmm. Reading the name of this test first made me think "why should an org have multiple histories"? When in fact an `OrganizationHistory` entity refers to a **historical record** for an organization. Can we maybe come up with better naming? Maybe it could help to take advantage of the fact that SQLAlchemy allows us to control the table name independently of the entity name.

opengever/contact/tests/test_personhistory_unit.py
```diff
@@ -0,0 +1,33 @@
...
+        history1 = create(Builder('personhistory')
+                          .having(firstname=u'Hans', lastname=u'M\xfcller',
+                                  contact=person))
+        history2 = create(Builder('personhistory')
+                          .having(firstname=u'Fritz', lastname=u'M\xfcller',
+                                  contact=person))
+
+        self.assertEqual([history1, history2], person.history)
+
+    def test_is_contacthistory(self):
```
lukasgraf: => `test_is_personhistory`

opengever/contact/tests/test_phonenumberhistory_unit.py
```diff
@@ -0,0 +1,41 @@
+from ftw.builder import Builder
+from ftw.builder import create
+from opengever.testing import MEMORY_DB_LAYER
+import unittest2
+
+
+class TestAddressHistory(unittest2.TestCase):
```
lukasgraf: -> `TestPhoneNumberHistory`

opengever/contact/tests/test_history_tables.py
```diff
@@ -0,0 +1,61 @@
...
+from opengever.contact.models import PhoneNumberHistory
+from sqlalchemy import inspect
+from unittest2 import TestCase
+
+
+class TestHistoryTables(TestCase):
+    """Test that history tables contain all required columns.
+
+    When this test fails it most likely means that you have added columns to
+    a table but forgot to update the corresponding history table.
```
lukasgraf: 💚

lukasgraf:
Other than my comments, looks good to me! 👍

opengever/contact/tests/test_organizationhistory_unit.py
deiferni: as discussed, `Archived[TheEntityName]` should be best

deiferni:
regarding bae3abe, nice catch! someone else also renamed the module and now the wrong one got deleted by my commit. thus i dropped the commit.

deiferni:
@lukasgraf updated, the last commits need another careful review, the rename might have introduced new mistakes

lukasgraf:
Looks good to me from what I can tell. 👍 

But I wanted to give this a test locally, and realized that there's no views in place yet that actually use these models, or can be used to exercise them.

Therefore it's kind of hard for me to say whether this approach for versioning will actually work. The ideas seem sound, but the only way we'll be able to tell is when we use it with views that do the respective changes / queries.

I'd like to avoid merging model changes that need to be modified again, and require schema migrations. So if that's ok, we'll leave this open for now.

deiferni:
@lukasgraf updated with your suggestions, we now create random example data and also display the history for a contact.

opengever/contact/tests/test_person_view.py
```diff
@@ -106,6 +146,23 @@ def test_shows_phonenumbers_prefixed_with_label(self, browser):
...
+
+        create(Builder('archived_phonenumber')
+               .for_contact(peter)
+               .labeled('Rotary Club')
+               .having(phone_number=u'011 111 11 00'))
+
+        browser.login().open(self.contactfolder, view=peter.wrapper_id)
+
+        self.assertEquals([u'Rotary Club'],
+                          browser.css('.phone_number dt').text)
```
phgross: This tests does not really check if the archived phonenumber ist displayed as an archived one, i would add a currently active phone number and/or change the css selector.

I think its the same for all the other tests (Urls, mails ...)

phgross:
@deiferni  only a small thing in the test then ready to 🚢

opengever/contact/tests/test_person_view.py
deiferni: nice catch, this test should also use the `#contactHistory` selector as do the others!

deiferni:
@phgross updated

PR #2047, opened 2016-07-21
Refactor restricted vocabularies / acquired field values

lukasgraf:
This is a set of refactorings that are required in order to eventually rewrite the `ILifeCycle` and `IClassification` defaults as `defaultFactories` instead of form level defaults.

This is not exactly a *pure* refactoring (some subtle behavior changes were unavoidable), but overall the functional behavior on an application level shouldn't have changed. Instead I created [issues with the `default-values` label](https://github.com/4teamwork/opengever.core/issues?q=is%3Aissue+is%3Aopen+label%3Adefault-values) and referenced them where appropriate, in order to fix them at a later point.

@deiferni @phgross


opengever/base/behaviors/classification.py
```diff
@@ -128,13 +130,12 @@ def validate_children(folder, event):
...
 
-grok.global_utility(
-    utils.create_restricted_vocabulary(
-        IClassification['classification'],
-        CLASSIFICATION_OPTIONS,
-        message_factory=_),
-    provides=schema.interfaces.IVocabularyFactory,
-    name=u'classification_classification_vocabulary')
+# TODO: This will be rewritten to eliminate one level of factories
+classification_vf_factory = create_restricted_vocabulary(
```
deiferni: what does vf stand for? vocabulary factory? i'd recomment to not abbreviate here and use `classification_vocabulary_factory` for readability.

opengever/base/behaviors/classification.py
```diff
@@ -163,13 +161,12 @@ def validate_children(folder, event):
...
 
-grok.global_utility(
-    utils.create_restricted_vocabulary(
-        IClassification['privacy_layer'],
-        PRIVACY_LAYER_OPTIONS,
-        message_factory=_),
-    provides=schema.interfaces.IVocabularyFactory,
-    name=u'classification_privacy_layer_vocabulary')
+# TODO: This will be rewritten to eliminate one level of factories
+privacy_layer_vf_factory = create_restricted_vocabulary(
```
deiferni: same here

opengever/base/behaviors/classification.py
```diff
@@ -145,13 +146,10 @@ def validate_children(folder, event):
...
 
-
-grok.global_utility(
-    utils.create_simple_vocabulary(
-        options=PUBLIC_TRIAL_OPTIONS,
-        message_factory=_),
-    provides=schema.interfaces.IVocabularyFactory,
-    name=u'classification_public_trial_vocabulary')
+# TODO: This will be rewritten to eliminate one level of factories
+public_trial_vf_factory = create_simple_vocabulary(
```
deiferni: same here

opengever/base/behaviors/classification.py
lukasgraf: @deiferni Yes it does. Yes I did (earlier). It will be eliminated in a later commit in the same PR later, hence the comment 😉

opengever/base/upgrades/20160715122912_update_vocab_used_for_public_trial_default_value_registry_record/registry.xml
```diff
@@ -0,0 +1,3 @@
+<registry>
+    <records interface="opengever.base.behaviors.classification.IClassificationSettings" />
```
deiferni: If i understand correctly this just reloads the currently valid `IClassificationSettings` form the interface?
If so, please change the upgrade step and add the changed delta in an explicitly.

opengever/base/monkey/patches/default_values.py
```diff
@@ -228,3 +242,20 @@ def changedField(field, value, context=None):
...
+
+class PatchDexterityDefaultAddForm(MonkeyPatch):
+    """Patch DefaultAddForm.update() to have the request provide
+    IDuringContentCreation while content creation is in progress.
+    """
+
+    def __call__(self):
+
+        def update(self):
+            alsoProvides(self.request, IDuringContentCreation)
```
phgross: The interface stays on the request (`noLongerProvides` is missing), hard to tell for me if thats a problem or not ...

opengever/base/acquisition.py
```diff
@@ -38,9 +52,14 @@ def acquire_field_value(field, container):
...
 
 def set_default_with_acquisition(field, default=None):
-    """Sets a default value generator which uses the value
-    from the parent object, if existing, otherwise it uses
-    the given default value.
+    """Factory to provide a default value by acquiring it from parents.
+
+    First this function will attempt to acquire a value for the given field
+    by walking up the chain of parents. If no value can be found via
+    acuisition, we fall back on the provided static `default`, if given.
```
phgross: acquisition --> acquisitoin

phgross:
Otherwise 👍 🎉

opengever/base/upgrades/20160715122912_update_vocab_used_for_public_trial_default_value_registry_record/registry.xml
lukasgraf: The reason this upgrade step is that those schema interfaces only serve as a template when creating `plone.app.registry` records. The fact that the vocabulary for the `public_trial_default_value` changed means that without the upgrade step, `plone.app.registry` is still looking for a named global utility (vocabulary factory), even though the field on the schema changed. Re-importing the `IClassificationSettings` registry record fixes this.

I'll see if I can change the upgrade step to only include a delta, but it might not be possible without resorting to pure Python.

opengever/base/monkey/patches/default_values.py
lukasgraf: Shouldn't be a problem. Besides it being quite tricky to pin-point a function or method that gets called when adding content via `z3c.form` add form is over, it also shouldn't be necessary. Since adding content through `z3c.form` add forms is inherently browser-based, there's only ever going to be one content creation per request, and that request should finish and expire after that.

lukasgraf:
@phgross updated

PR #2049, opened 2016-07-21
Always enable bumblebee for development by default.

deiferni:
Most of our customers will enable bumblebee, so we should, too while developing.


deiferni:
@lukasgraf or @phgross

lukasgraf:
👍 🎉

PR #2050, opened 2016-07-21
Translations cleanup

phgross:
- Removed empty field descriptions
 - Rename plonemessage factory. So that the plone translations are ignored by the i18ndude.
 - Removed translations of intern labels as registry entries. 
 - updated all translation files except the ones from opengever.contact.

@lukasgraf as discussed


opengever/dossier/move_items.py
```diff
@@ -158,8 +158,8 @@ def create_statusmessages(
                 msg, type='error')
 
         if failed_resource_locked_objects:
-            msg = _(u'Failed to copy following objects: ${failed_objects}\
-                    . Locked via WebDAV',
+            msg = _(u'Failed to copy following objects: ${failed_objects}. '
+                    'Locked via WebDAV',
```
lukasgraf: `'Locked via WebDAV'` should also be an unicode literal: `u'Locked via WebDAV'`

phgross:
@lukasgraf updated ...

lukasgraf:
👍

PR #2055, opened 2016-07-22
Make tabbedview filter on sqltables listings case insensitive.

phgross:
The filterbox should be an easy and fast possibility to limit the entries of a listing, therefore it should work case insensitive.

Note: the filterbox on catalog listings already works case insensitive.

 --- 
Additionally in enable case_sensitive_like on sql lite connections.
This makes sure that all like queries in tests are case sensitive. So our test DB (SQLLite) works in the same manner than our production DBs (Oracle/PostgreSQL).

@lukasgraf

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.10.0 (unreleased)
 -------------------
 
+- Make tabbedview filter on sqltables listings case insensitive.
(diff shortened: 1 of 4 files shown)


lukasgraf:
@phgross looks good to me 👍 , please rebase

phgross:
@lukasgraf rebased

PR #2038, opened 2016-07-19
Emailformulare für Personen

elioschmutz:
🚧   MERGE https://github.com/4teamwork/plonetheme.teamraum/pull/448 FIRST 🚧 

Dieser PR fügt das E-Mailformular im Edit-Form einer Person hinzu.

emailform

TODO:

- [x] Tests
- [x] Übersetzungen
- [x] PR für plonetheme.teamraum

PR #2057, opened 2016-07-22
Reapply fix sticky heading in order to a refactoring of the baseclass.

elioschmutz:
Unfortunately the fix #2052 have been reverted from a earlier pullrequest #2038.

So we have to reapply it.

@lukasgraf


lukasgraf:
@elioschmutz the changelog entry for "fix sticky heading" already exists (hasn't been reverted). So there's no need to list this change again in the changelog.

elioschmutz:
👍

PR #2066, opened 2016-07-26
Rewrite form defaults for IClassification and ILifeCycle fields to defaultFactories

lukasgraf:
This rewrites the last remaining field defaults (that *shouldn't* be form defaults) as `defaultFactories`. Those are `IClassification` and `LifeCycle` fields that use fancy restricted vocabularies that had to undergo a major refactoring first (#2047).

@phgross


phgross:
👌

PR #2070, opened 2016-07-28
Deactivate 'Authentication' capability for LDAP plugin during setup for production deployments

lukasgraf:
In production, authentication will always be performed by CAS portal, and it should not be possible to authenticate against LDAP with Plone directly (e.g. via `/login_form`).

This is necessary so we can enforce the requirement for the second factor in 2FA via the CAS Portal.

⚠️  **Note** ⚠️ 
This means that it will be impossible to authenticate via LDAP *without the CAS Portal* for any deployment that isn't a local development setup (extending from `development.cfg`). *Every* GEVER deployment needs a CAS Portal from now on.

@phgross

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,11 @@ Changelog
 4.10.0 (unreleased)
 -------------------
 
+- Deactivate 'Authentication' capability for LDAP plugin during setup for
+  production deployments.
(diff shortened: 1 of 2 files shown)


phgross:
👍  please rebase

lukasgraf:
@phgross rebased

PR #2076, opened 2016-07-29
Restructure default value tests

lukasgraf:
This is a series of restructurings ad cleanups for the default value tests. These will eventually be needed in order to implement setting missing values.

@phgross

### DIFF
--- a/opengever/base/tests/test_default_values_for_types.py
+++ b/opengever/base/tests/test_default_values_for_types.py
@@ -20,12 +20,14 @@ FROZEN_NOW = datetime.now()
 FROZEN_TODAY = FROZEN_NOW.date()
 
 DEFAULT_TITLE = u'My title'
+DEFAULT_CLIENT = u'client1'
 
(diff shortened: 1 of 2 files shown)


opengever/base/tests/test_default_values_for_types.py
```diff
@@ -519,10 +550,12 @@ def test_create_content_in_container(self):
             task = createContentInContainer(
                 self.portal,
                 'opengever.task.task',
-                title=DEFAULT_TITLE,
-                issuer=TEST_USER_ID,
-                responsible=TEST_USER_ID,
-                responsible_client='client1')
+                title=TASK_REQUIREDS['title'],
```
phgross: Maybe just passing in *_REQUIREDS 

```python
 task = createContentInContainer(
                  self.portal, 'opengever.task.task', **TASK_REQUIREDS)
```

phgross:
👍

PR #2096, opened 2016-08-18
Fix showroom CSS loading order

bierik:
In some cases the ftw.showroom styling gets loaded after the GEVER styles. This causes the showroom styles defined in GEVER not to apply.
The upgrade step moves the showroom styles before the GEVER styles.

### DIFF
--- a/opengever/bumblebee/upgrades/20160818143144_fix_showroom_css_order/cssregistry.xml
+++ b/opengever/bumblebee/upgrades/20160818143144_fix_showroom_css_order/cssregistry.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0"?>
+<object name="portal_css">
+
+    <stylesheet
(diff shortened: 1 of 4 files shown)


opengever/bumblebee/upgrades/20160818143144_fix_showroom_css_order/cssregistry.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<object name="portal_css">
+
+    <stylesheet title=""
```
deiferni: You must also add this to `bumblebee/profiles/default` otherwise it will only be applied when upgrading, but not for a fresh installation

bierik:
@deiferni updated

deiferni:
👍  tested when upgrading and on a fresh installation, seems to work correctly now

PR #2098, opened 2016-08-18
Add views and forms for dossier participations

phgross:
Depend on https://github.com/4teamwork/opengever.ogds.models/pull/41

This PR adds views and forms for dossier participations using the new SQL contacts objects.

To make sure we can support both contacts implementation at once, the tab and the `add participation` action has to depend on the contact feature flag, which controls if the old Plone object implementation is active or the new SQL object implementation. Therefore I've reworked all dossier tabbedviews, to easily return the correct participation tab depending on the feature flag and register a redirect view which redirects to the correct Participation add form.

Participation Tab:
bildschirmfoto 2016-08-18 um 14 47 16
Add form
bildschirmfoto 2016-08-18 um 14 47 32
Edit form
bildschirmfoto 2016-08-18 um 14 49 15
Remove form
bildschirmfoto 2016-08-18 um 14 49 37

@deiferni 

Styling in a separate PR

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,10 @@ Changelog
 4.10.1 (unreleased)
 -------------------
 
+- Added participation tab and forms (add, edit, remove) using the new
+  contact implementation.
+  [phgross]
+
 - Disable local tab on contactfolder when contact feature is enabled.
   [phgross]
 
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.10.1 (unreleased)
 -------------------
 
+- Disable local tab on contactfolder when contact feature is enabled.
+  [phgross]
+
 - Move bumblebee installation script to opengever.maintenance.
   [phgross]
 
--- a/opengever/contact/browser/participation_add_view.py
+++ b/opengever/contact/browser/participation_add_view.py
@@ -1,0 +1,16 @@
+from opengever.contact import is_contact_feature_enabled
+from Products.Five.browser import BrowserView
+
+
+class ParticipationAddView(BrowserView):
+    """A simple redirector view which redirects to the corresponding add form,
+    depending on the is_contact_feature_enabled flag.
+
+    The view is called by dossiers `Add participation` factoriesmenu action.
+    """
+
+    def __call__(self):
+        if is_contact_feature_enabled():
+            return self.request.RESPONSE.redirect('add-sql-participation')
+
+        return self.request.response.redirect('add-plone-participation')
--- a/opengever/contact/browser/participation_forms.py
+++ b/opengever/contact/browser/participation_forms.py
@@ -11,7 +11,6 @@ from plone.formwidget.autocomplete import AutocompleteFieldWidget
 from plone.z3cform import layout
 from z3c.form.browser.checkbox import CheckBoxFieldWidget
 from z3c.form.interfaces import ActionExecutionError
-from z3c.form.interfaces import DISPLAY_MODE
 from z3c.form.interfaces import HIDDEN_MODE
 from z3c.form.interfaces import IDataConverter
 from zExceptions import Unauthorized
@@ -57,7 +56,6 @@ class ParticipationAddForm(z3c.form.form.Form):
 
     fields['contact'].widgetFactory = AutocompleteFieldWidget
     fields['roles'].widgetFactory = CheckBoxFieldWidget
-    fields['participation_id'].mode = HIDDEN_MODE
 
     @z3c.form.button.buttonAndHandler(_(u'button_add', default=u'Add'))
     def handle_add(self, action):
@@ -103,28 +101,25 @@ class ParticipationAddFormView(layout.FormWrapper, grok.View):
 
 class ParticipationEditForm(z3c.form.form.EditForm):
     ignoreContext = True
-    label = _(u'label_edit_participation', default=u'Edit Participation')
-    fields = z3c.form.field.Fields(IParticipation)
+    fields = z3c.form.field.Fields(IParticipation).omit('contact')
     participation = None
 
-    fields['contact'].mode = DISPLAY_MODE
     fields['participation_id'].mode = HIDDEN_MODE
     fields['roles'].widgetFactory = CheckBoxFieldWidget
+
+    @property
+    def label(self):
+        return _(u'label_edit_participation',
+                 default=u'Edit Participation of ${title}',
+                 mapping={'title': self.get_participation().contact.get_title()})
 
     def get_participation(self):
         participation_id = self.request.get('participation_id')
         if not participation_id:
-            return None
+            data, errors = self.widgets.extract()
+            participation_id = data.get('participation_id')
+
         return Participation.query.get(participation_id)
-
-    def update(self):
-        participation = self.get_participation()
-        if participation:
-            if not self.request.get('form.widgets.contact', None):
-                self.request.set('form.widgets.contact',
-                                 [participation.contact.contact_id])
-
-        super(ParticipationEditForm, self).update()
 
     def updateWidgets(self):
         super(ParticipationEditForm, self).updateWidgets()
--- a/opengever/contact/browser/participation_forms.py
+++ b/opengever/contact/browser/participation_forms.py
@@ -1,0 +1,92 @@
+from collective.elephantvocabulary import wrap_vocabulary
+from five import grok
+from opengever.base.model import create_session
+from opengever.base.oguid import Oguid
+from opengever.contact.models import Contact
+from opengever.contact.models import Participation
+from opengever.dossier import _
+from plone import api
+from plone.directives import form
+from plone.formwidget.autocomplete import AutocompleteFieldWidget
+from plone.z3cform import layout
+from z3c.form.browser.checkbox import CheckBoxFieldWidget
+from z3c.form.interfaces import ActionExecutionError
+from zope import schema
+from zope.interface import Interface
+from zope.interface import Invalid
+import z3c.form
+
+
+class IParticipation(form.Schema):
+    """ Participation Form schema
+    """
+
+    contact = schema.Choice(
+        title=_(u'label_contact', default=u'Contact'),
+        description=_(u'help_contact', default=u''),
+        vocabulary=u'opengever.contact.ContactsVocabulary',
+        required=True,
+    )
+
+    roles = schema.List(
+        title=_(u'label_roles', default=u'Roles'),
+        value_type=schema.Choice(
+            source=wrap_vocabulary(
+                'opengever.dossier.participation_roles',
+                visible_terms_from_registry='opengever.dossier'
+                '.interfaces.IDossierParticipants.roles'),
+        ),
+        required=True,
+        missing_value=[],
+        default=[],
+    )
+
+
+class ParticipationAddForm(z3c.form.form.Form):
+    ignoreContext = True
+    label = _(u'label_add_participation', default=u'Add Participation')
+    fields = z3c.form.field.Fields(IParticipation).omit('participation_id')
+
+    fields['contact'].widgetFactory = AutocompleteFieldWidget
+    fields['roles'].widgetFactory = CheckBoxFieldWidget
+
+    @z3c.form.button.buttonAndHandler(_(u'button_add', default=u'Add'))
+    def handle_add(self, action):
+        data, errors = self.extractData()
+
+        contact = Contact.query.get(data.get('contact'))
+        oguid = Oguid.for_object(self.context)
+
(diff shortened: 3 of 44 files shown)


phgross:
@lukasgraf @deiferni the PR is now ready for a review ...

docs/HISTORY.txt
```diff
@@ -5,6 +5,13 @@ Changelog
 4.10.0 (unreleased)
 -------------------
 
+- Added participation tab and forms(add, edit, remove) using the new
```
deiferni: missing whitespace 😉

opengever/contact/browser/participation_add_view.py
```diff
@@ -0,0 +1,17 @@
+from opengever.contact import is_contact_feature_enabled
+from Products.Five.browser import BrowserView
+
+
+class ParticipationAddView(BrowserView):
+    """A simple redirector view which redirects, depending on the
```
deiferni: A simple redirector view which redirects to the corresponding add form, depending on the `is_contact_feature_enabled` flag

docs/HISTORY.txt
```diff
@@ -5,6 +5,13 @@ Changelog
...
+  contact implementation.
+  [phgross]
+
+- Disable local tab on contactfolder when contact feature is disabled.
```
deiferni: this is the wrong war round, isn't it? the local tab is disabled with an active feature?

opengever/contact/models/participation.py
```diff
@@ -10,6 +13,16 @@
 from sqlalchemy.schema import Sequence
 
 
+class ParticipationQuery(BaseQuery):
+
+    def by_oguid(self, oguid):
+        return self.filter(Participation.dossier_oguid == oguid)
+
+    def by_oguid_and_contact(self, oguid, contact):
```
deiferni: i would avoid writing these methods, they are almost the same as Parcitipation.query.filter_by(dossier_oguid=some_oguid, contact=contact), so they don't add any value. 
I would only write new query methods when some work is required for the arguments, i.e. a conversion or something like that.

In this case i would add a method `by_dossier` and then create the dossier oguid within that method.

opengever/contact/tests/test_participation.py
```diff
@@ -44,3 +47,64 @@ def test_relation_to_dossier(self):
...
+        create(Builder('participation')
+               .having(contact=self.peter,
+                       dossier_oguid=Oguid.for_object(self.dossier)))
+
+        browser.login().open(self.dossier, view='add-contact-participation')
+        browser.fill({'Contact': str(self.peter.contact_id),
+                      'Roles': ['Final drawing', 'Regard']})
+        browser.click_on('Add')
+
+        self.assertEquals(['There were some errors.'], error_messages())
```
deiferni: a (nicer) alternative would be to not display already selected entities

opengever/contact/vocabularies.py
```diff
@@ -0,0 +1,46 @@
...
+class SQLSimpleVocabulary(SimpleVocabulary):
+
+    def search(self, query_string):
+        """search method for `z3c.formwidget.autocomplete` support. Returns
+        all matching contacts.
+        """
+
+        if isinstance(query_string, str):
+            query_string = query_string.decode('utf8')
+        query = query_string.lower().split(' ')
```
deiferni: i would use `split()` without parameter as it splits arbitrary whitespace and behaves slightly different (imho better for this use case), see https://docs.python.org/2/library/stdtypes.html#str.split

opengever/contact/vocabularies.py
```diff
@@ -0,0 +1,46 @@
...
+        if not value:
+            return False
+
+        for word in query:
+            if len(word) > 0 and word not in value.lower():
+                return False
+        return True
+
+
+class ContactsAndUsersVocabularyFactory(grok.GlobalUtility):
```
deiferni: i think we wanted to not add any more new grok stuff since it has to be removed in the distant (or near?) future

opengever/contact/browser/participation_forms.py
```diff
@@ -90,3 +99,82 @@ class ParticipationAddFormView(layout.FormWrapper, grok.View):
     def __init__(self, *args, **kwargs):
         layout.FormWrapper.__init__(self, *args, **kwargs)
         grok.View.__init__(self, *args, **kwargs)
+
+
+class ParticipationEditForm(z3c.form.form.EditForm):
```
deiferni: you should subclass `ModelEditForm` it already handles initial values for you

opengever/contact/models/participation.py
```diff
@@ -41,10 +41,33 @@ class Participation(Base):
...
-    def add_roles(self, roles):
-        for role_title in roles:
-            role = ParticipationRole(participation=self, role=role_title)
+    def add_roles(self, role_names):
+        for name in role_names:
+            role = ParticipationRole(participation=self, role=name)
             create_session().add(role)
 
+    def update_roles(self, role_names):
+        """Update the ParticipationRoles of the Participation, means:
```
deiferni: ~~, means~~

opengever/contact/browser/participations.py
```diff
@@ -0,0 +1,21 @@
...
+    template = ViewPageTemplateFile('templates/participations.pt')
+
+    def __call__(self):
+        return self.template()
+
+    def get_participations(self):
+        """Returns all participations for the current context.
+        """
+        oguid = Oguid.for_object(self.context)
+        return Participation.query.by_oguid(oguid).all()
```
deiferni: `query.by_dossier(self.context)`, see previous comment

opengever/contact/tests/test_participation_tab.py
```diff
@@ -0,0 +1,84 @@
...
+            firstname=u'Hans', lastname=u'M\xfcller'))
+        self.sandra = create(Builder('person').having(
+            firstname=u'sandra', lastname=u'Meier'))
+        self.peter_ag = create(Builder('organization')
+                               .having(name=u'Peter AG'))
+
+        create(Builder('participation')
+               .for_contact(self.hans)
+               .for_dossier(self.dossier)
+               .with_roles(['regard, final-drawing']))
```
deiferni: i guess this should be `['regard', 'final-drawing']` ?

opengever/contact/tests/test_participation_tab.py
```diff
@@ -0,0 +1,84 @@
...
+
+    @browsing
+    def test_roles_of_each_participation_is_visible(self, browser):
+        browser.login().open(self.dossier,
+                             view=u'tabbedview_view-participations')
+
+        row1, row2 = browser.css('#participation_listing > li')
+
+        self.assertEquals([u'Hans M\xfcller'], row1.css('.contact').text)
+        self.assertEquals(['regard, final-drawing'], row1.css('.roles').text)
```
deiferni: 😕

opengever/contact/tests/test_participation_tab.py
```diff
@@ -20,7 +20,7 @@ def setUp(self):
         create(Builder('participation')
                .for_contact(self.hans)
                .for_dossier(self.dossier)
-               .with_roles(['regard, final-drawing']))
+               .with_roles(['regard', 'final-drawing']))
```
deiferni: ah 👍  you already fixed it 😉

opengever/contact/models/participation_role.py
```diff
@@ -24,3 +26,16 @@ class ParticipationRole(Base):
 
     def __repr__(self):
         return u'<ParticipationRole {} >'.format(self.role)
+
+    def get_label(self):
+        vocabulary_registry = getVocabularyRegistry()
+        vocab = vocabulary_registry.get(
+            api.portal.get(),
+            'opengever.dossier.participation_roles')
```
deiferni: hmm not sure if we should keep the vdex vocab here. as far as i could see these terms are not customized by any customer. maybe a static list with valid terms would do ...

opengever/dossier/browser/tabbed.py
```diff
@@ -0,0 +1,125 @@
...
+        'class': None}
+
+    @property
+    def subdossiers_tab(self):
+        if self.context.show_subdossier:
+            return {'id': 'subdossiers',
+                    'title': _(u'label_subdossiers', default=u'Subdossiers'),
+                    'icon': None,
+                    'url': '#',
+                    'class': None}
```
deiferni: i'd add an explicit `return None` for readability and since the return value is important

opengever/dossier/browser/tabbed.py
```diff
@@ -0,0 +1,125 @@
...
+        'class': None}
+
+    journal_tab = {
+        'id': 'journal',
+        'title': _(u'label_journal', default=u'Journal'),
+        'icon': None,
+        'url': '#',
+        'class': None}
+
+    sharing_tab = {
```
deiferni: why not rename to `info_tab`?

opengever/contact/browser/participation_forms.py
```diff
@@ -173,3 +173,71 @@ class ParticipationEditFormView(layout.FormWrapper, grok.View):
     def __init__(self, *args, **kwargs):
         layout.FormWrapper.__init__(self, *args, **kwargs)
         grok.View.__init__(self, *args, **kwargs)
+
+
+class ParticipationRemoveForm(z3c.form.form.Form):
```
deiferni: i think this should also inherit form `ModelEditForm`

opengever/contact/vocabularies.py
```diff
@@ -1,46 +1,57 @@
...
 
-        for word in query:
-            if len(word) > 0 and word not in value.lower():
-                return False
-        return True
+    def getTermByToken(self, token):
+        return self.getTerm(token)
+
+    def search(self, query_string):
+        text_filters = query_string.split(' ')
```
deiferni: `split()`

opengever/contact/vocabularies.py
```diff
@@ -1,46 +1,57 @@
...
+from zope.schema.interfaces import IVocabularyTokenized
 from zope.schema.interfaces import IVocabularyFactory
 from zope.schema.vocabulary import SimpleTerm
-from zope.schema.vocabulary import SimpleVocabulary
 
 
-class SQLSimpleVocabulary(SimpleVocabulary):
+@implementer(IVocabularyTokenized)
+class ContactsVocabulary(object):
+    """A vocabulary of all contacts, means all persons and organizations.
```
deiferni: ~~means~~

opengever/contact/vocabularies.py
```diff
@@ -1,46 +1,57 @@
...
-        return True
+    def getTermByToken(self, token):
+        return self.getTerm(token)
+
+    def search(self, query_string):
+        text_filters = query_string.split(' ')
+        for query in self.queries:
+            query = query.by_searchable_text(text_filters=text_filters)
+            for obj in query.all():
+                yield self.createTerm(obj)
```
deiferni: how will this sort your terms? desired would be to sort alphabetically independent of the content-type. i fear this will show all matching persons first, then all matching contacts. i'd also add a test for this.

deiferni:
👍  apart from questions/comments

opengever/contact/tests/test_participation.py
phgross: I think the additional expense is not worth it, furthermore it could be confusing for a user who overlooked the existing participation and search for a contact and can't find him. 

Altogether I think a simple validation msg is the better way ...

opengever/contact/models/participation_role.py
phgross: For the bs installation we'll customize this vocabulary ... so I think it makes sense to still use a vdex vocab here...

opengever/contact/tests/test_participation.py
deiferni: hmm i have to disagree here, i think it leads to a bad user experience when the user can select and attempt to save an entity of which we already know that i won't be accepted. I see your point, omitting from the list can also be confusing, what about displaying an unselectable entity in the list?

opengever/contact/models/participation_role.py
deiferni: well in that case 😉 👍

opengever/contact/tests/test_participation.py
phgross: As discussed together, the best way would be to add an unselectable entry to the autocomplete widget. With the current autocomplete widget, this is not possible. Therefore we use the implementation with simple validator.

opengever/contact/vocabularies.py
phgross: The sorting is handled by the autocomplete, so its irrelevant in which sorting we return the terms.

phgross:
@deiferni updated, please have another look to 482ac6e

deiferni:
👍 ✨

PR #2112, opened 2016-08-26
Handle non-standard storages in get_persisted_value_for_field()

lukasgraf:
Before the `get_persisted_value_for_field()` function assumed that there would only ever be `AttributeStorage` or `AnnotationStorage` in play.

However, since we're also using other custom storages, we need to deal with those as well.

The new function docstring basically describes how we need to handle those:

> What we need to do here is basically to decide whether the storage implementation used for the given field on this context has any **fallbacks** (to default or missing value) in place. If it does, we need to **bypass** them and only return the actually persisted value, or raise an AttributeError.

@phgross

### DIFF
--- a/opengever/base/default_values.py
+++ b/opengever/base/default_values.py
@@ -90,6 +90,7 @@ that would fool the check.
 
 from Acquisition import aq_base
 from persistent.interfaces import IPersistent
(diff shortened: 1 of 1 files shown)


phgross:
👌

PR #2113, opened 2016-08-26
Add organizations column to the person listing.

phgross:
bildschirmfoto 2016-08-26 um 11 09 20

Besides I extract and improve the linked method in contact listings to a separate tabbeview helper, which now also prevent XSS.

@lukasgraf


lukasgraf:
Looks good 👍  But tests are still running 😉

PR #2118, opened 2016-08-26
Set and persist Dexterity default values

lukasgraf:
This feature branch contains changes that achieve the following:

- If values are set on a DX object that are equal to the default, they now actually get persisted
- When creating objects (e.g. using createContentInContainer() or invokeFactory()), **defaults** (for fields that have them) or **missing values** will be set for fields that aren't provided

Contains the following pull requests:

#1952 - Patch `DexterityContent.__getattr__` to correctly fall back to defaults from behavior schemas
#1956 - Use defaultFactory for `IClassification.public_trial` instead of form level defaults.
#1996 - Make sure Dexterity default values are persisted
#2023 - Add comprehensive tests for fields using restricted vocabularies
#2047 - Refactor restricted vocabularies / acquired field values
#2066 - Rewrite form defaults for `IClassification` and `ILifeCycle` fields to defaultFactories
#2076 - Restructure default value tests
#2077 - Also set missing values if no default can be found
#2112 - Handle non-standard storages in `get_persisted_value_for_field()`


thet:
I'm having the same problem ...

why not bringing these patches upstream?

lukasgraf:
@thet @agitator I would love to! That would make everything much easier for us instead of having to monkey patch everything. The problem is, it's anything else than trivial to fix it upstream.

Because it's not a single issue in one place, but a complex interaction of (well-intended) behavior all across the Plone ecosystem - `plone.dexterity`, `z3c.form`, `plone.behavior`, `zope.schema`, just to name a few.

And while in our own product I can freely make assumptions about what dependencies will always be there, and impart knowledge about how other dependencies behave into them via monkey patch, I can't simply do that in some of these packages that are supposed to be somewhat agnostic about Plone, like `z3c.form` and `zope.schema`.

But I intend to create an issue (probably in `plone.dexterity`, or maybe just `Products.CMFPlone`) where I outline all the little details of this mess, report our experiences with this set of changes, and start a discussion about how we can fix this issue at the root. I'll `@` mention you both once I get around to opening that issue.

PR #2122, opened 2016-08-29
[showroom] Fix document overview on searchpage

bierik:
Multiple showroom items get initialized for the same document.
So use a showroom reference for the preview image.


deiferni:
👍

PR #2124, opened 2016-08-30
Fix wrong pin position when trix content has been changed

bierik:
When the trix content grows, the pin item position is not
going to be updated. So trigger a refresh on `trix-change`


deiferni:
thx 👍 
also reported in https://feedback.onegovgever.ch/t/formatierungsbalken-laeuft-nicht-mit/262

PR #2132, opened 2016-08-31
Fix hint for local changes

bierik:
The synchronizer is not able to handle the same event twice.
So refresh the pin items in the trix synchronizer callback instead of creating a new synchronizer.


bierik:
@deiferni works on IE11

PR #2136, opened 2016-09-02
Contact detail views improvements

phgross:
- Link mail addresses and URLs 
- Only display archived-info collapsible header if archived information exists.

@deiferni

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,8 +5,12 @@ Changelog
 4.11.0 (unreleased)
 -------------------
 
+- Link mail addreses in contact detail views.
+  [phgross]
+
 - Whitelist @login, @logout and @login-renew REST API endpoints for anonymous
   access on Plone site root.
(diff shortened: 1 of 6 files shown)


opengever/contact/models/url.py
```diff
@@ -17,3 +18,8 @@ class URL(Base):
     contact = relationship("Contact", back_populates="urls")
     label = Column(String(CONTENT_TITLE_LENGTH))
     url = Column(String(CONTENT_TITLE_LENGTH))
+
+    def s_url(self):
```
deiferni: hmm we should find a better name here.

deiferni:
Looks 👍 , as discussed there are changes pending, though

phgross:
@deiferni @lukasgraf updated ...

phgross:
@lukasgraf rebased

lukasgraf:
:+1:

PR #2147, opened 2016-09-09
Fix whitelisting of anonymous views on site root

lukasgraf:
Fixes a mistake introduced in #2146:

`req.steps[-1]` doesn't normalize `@@view` to `view`, so depending on how a view was called in the URL, the whitelisted entries don't match any more.

So we revert back to using `req['PUBLISHED'].__name__` and instead prefix the `plone.rest` endpoints using `POST_application_json_@...`

@phgross


phgross:
👍

PR #2153, opened 2016-09-13
Prevent creating participations for inactive contacts.

### LINKED ISSUE #2152: Disallow creating participations for inactive contacts.
It should only be possible to select active contacts in the participation add form.

phgross:
Closes #2152


deiferni:
👍

PR #2159, opened 2016-09-16
Initialize trix editor on demand

### LINKED ISSUE #1958: Performance issues with many of trix-editors
Trix editors are initialized on load. This leads to performance problems on older internet explorers when a lot of editor instances are created (several seconds loading time, where the page freezes and cannot be used).

A possible solution could be to initialize the editors lazily, i.e. when the text-field is first accessed.

bierik:
⚠️  Depends on https://github.com/4teamwork/plonetheme.teamraum/pull/461

Closes #1958

/cc @phgross


opengever/base/browser/resources/MeetingStorage.js
```diff
@@ -0,0 +1,49 @@
+(function(global, $, GEVERStorage) {
+
+  var MeetingStorage = (function MeetingStorage() {
+    var storage;
```
deiferni: Please add a docstring to describe the purpose of this object.

docs/HISTORY.txt
```diff
@@ -5,6 +5,9 @@ Changelog
 4.11.0 (unreleased)
 -------------------
 
+- Load trix lazy
```
deiferni: Lazy load trix.

_Entries in the CL are sentences, and should always be terminated with a period 😉_

deiferni: Also handlebars gets updated, please mention that here

opengever/base/profiles/default/jsregistry.xml
```diff
@@ -1,73 +1,84 @@
 <?xml version="1.0"?>
 <object name="portal_javascripts" meta_type="JavaScripts Registry">
+    <!-- !!! WARNING: DO NOT DEFINE DEPENDENCIES MANUALLY USE THE TARJAN SCRIPT INSTEAD: https://gist.github.com/bierik/ece24ad316424d6d4909efd272940fb4 !!! -->
```
deiferni: could we put this script into the `og.core` repo instead? 

i'd expect a short readme entry and the script. if we don't have the `tarjan` package yet and if we don't want to ship it by default we could add a `development` extra.

You can do this in a separate PR if that's easier.

opengever/base/profiles/default/jsregistry.xml
```diff
@@ -1,73 +1,84 @@
...
-    <javascript cacheable="True" compression="none" cookable="True" enabled="on"
-                expression="" id="++resource++opengever.base/Storage.js" inline="False"
-                insert-after="++resource++opengever.base/Controller.js" />
+                enabled="True" id="++resource++opengever.base/prepoverlay.js" inline="False"
+                insert-after="++resource++plone.app.jquerytools.overlayhelpers.js" />
 
     <javascript cacheable="True" compression="none" cookable="True" enabled="on"
                 expression="" id="++resource++opengever.base/trix.js" inline="False"
-                insert-after="++resource++plone.app.jquery.js" />
+                insert-after="++resource++opengever.base/quickupload.js" />
```
deiferni: IMHO this introduces too many and "wrong" dependencies. for example there is no reason for trix to have a dependency on quickupload. trix is a stand-alone library.

I think this could be very confusing, i'd expect an `insert-after` in here to be a package dependency (direct/indirect). So in theory trix should not have any dependencies at all. I can see that you did this to always ensure the same package order, but maybe there is a less confusing way?
If we want to enforce a certain order for some reason we should do this in an upgrade step or for a fresh site maybe re-ordering the javascript definitions would be enough?

Maybe best discuss this in person 😉 .

opengever/meeting/browser/resources/Editor.js
```diff
@@ -0,0 +1,161 @@
...
+    Object.defineProperty(reveal, "toolbar", { get: function() { return toolbar; }});
+    Object.defineProperty(reveal, "enabled", { get: function() { return enabled; }});
+    Object.defineProperty(reveal, "content", { get: function() { return element.val(); }});
+    Object.defineProperty(reveal, "proposal", { get: function() { return id.split("-")[1]; }});
+    Object.defineProperty(reveal, "section", { get: function() { return id.split("-")[2]; }});
+    return Object.freeze(reveal);
+
+  }
+
+  function EditorProxy(proxyElement) {
```
deiferni: docstring

opengever/meeting/browser/resources/Editor.js
```diff
@@ -0,0 +1,161 @@
...
+      editor.element.detach();
+    }
+
+    reveal.revive = revive;
+    reveal.destroy = destroy;
+    Object.defineProperty(reveal, "editor", { get: function() { return editor; }});
+    return Object.freeze(reveal);
+  }
+
+  function EditorController(options) {
```
deiferni: docstring

opengever/meeting/browser/resources/Editor.js
```diff
@@ -0,0 +1,161 @@
...
+
+    Object.defineProperty(reveal, "id", { get: function() { return id; }});
+    Object.defineProperty(reveal, "element", { get: function() { return element; }});
+    Object.defineProperty(reveal, "link", { get: function() { return id + "-toolbar"; }});
+
+    return Object.freeze(reveal);
+
+  }
+
+  function Editor(id, options) {
```
deiferni: docstring

opengever/meeting/profiles/default/jsregistry.xml
```diff
@@ -1,37 +1,33 @@
 <?xml version="1.0"?>
 <object name="portal_javascripts" meta_type="JavaScripts Registry">
+    <!-- !!! WARNING: DO NOT DEFINE DEPENDENCIES MANUALLY USE THE TARJAN SCRIPT INSTEAD: https://gist.github.com/bierik/ece24ad316424d6d4909efd272940fb4 !!! -->
 
     <javascript cacheable="True" compression="safe" cookable="True" expression=""
-                enabled="True" id="++resource++opengever.meeting/jquery.autosize.js" inline="False"
-                insert-after="++resource++opengever.base/Storage.js"
-                />
+                enabled="True" id="++resource++opengever.meeting/Pin.js" inline="False"
```
deiferni: see comment in `opengever.base`, i think we should discuss this approach.

opengever/base/upgrades/20160915171602_fix_javascript_dependencies/upgrade.py
```diff
@@ -0,0 +1,9 @@
+from ftw.upgrade import UpgradeStep
+
+
+class FixJavascriptDependencies(UpgradeStep):
+    """Fix javascript dependencies.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
```
deiferni: after installing this upgrade i get multiple `Uncaught TypeError: Cannot read property 'compile' of undefined` errors, there seems to be an issue with `handblebars`.

screen shot 2016-09-19 at 11 42 13

deiferni: Also happens when i install a new plone site.

opengever/base/profiles/default/jsregistry.xml
bierik: The link in the commentsection above (https://gist.github.com/bierik/ece24ad316424d6d4909efd272940fb4) provides a script including the dependency graph.
I've already discussed this with @lukasgraf. He recommend no to store the script in the opengever.core package because this solution using tarjan and a dependency graph is not ready for production and should be an assistance for me. The dependency workflow for javascript is a mess anyway we should focus on a more reliable solution.

bierik:
As discussed with @deiferni we do not adjust the `insert-after` statement in the `jsregistry.xml` to produce a logical and human readable order. The linearized dependency graph produces by `tarjan` counts. Because some files have multiple dependencies we are not able to make a logical `jsregistry.xml` file with just using the `insert-after` statement. That's why we have to trust the `tarjan` produced linear dependency graph.

bierik:
@deiferni updated

opengever/base/profiles/default/jsregistry.xml
deiferni: if we don't want to introduce a tarjan dependency for production we could also add a `[dev]` extra to opengever.core. but if it is unclear whether this is a temporary solution or not the gist will do, too 👍 .

docs/HISTORY.txt
```diff
@@ -5,6 +5,9 @@ Changelog
 4.11.0 (unreleased)
 -------------------
 
+- Load trix lazily.
```
deiferni: plz also mention the handlebars update

opengever/meeting/browser/resources/Editor.js
```diff
@@ -0,0 +1,206 @@
...
+
+    var element = $(trixEditorTemplate({
+      toolbar: toolbar.link,
+      autofocus: options.autofocus,
+      id: id
+    }));
+
+    element.data("editor", reveal);
+
+    // function load(data) { element.editor.loadJSON(JSON.parse(data)); }
```
deiferni: please never ever commit commented-out code 😡 . it if is not needed, remove it.

opengever/meeting/browser/resources/Editor.js
```diff
@@ -0,0 +1,206 @@
...
+    /*
+      Disable the editor and its toolbar
+     */
+    function disable() {
+      enabled = false;
+      toolbar.disable();
+    }
+
+    reveal.save = save;
+    // reveal.load = load;
```
deiferni: 😡

opengever/meeting/browser/resources/Editor.js
```diff
@@ -0,0 +1,206 @@
...
+    the events for creating and deleting Editors on demand.
+   */
+  function EditorController(options) {
+
+    Controller.call(this);
+
+    this.activeProxy;
+
+    /*
+      Create a proxy for each proposal
```
deiferni: plz add colon, docstrings are sentences.

opengever/meeting/browser/resources/Editor.js
```diff
@@ -0,0 +1,206 @@
...
+     */
+    this.enableEditor = function(target) {
+      var editor = target.data("editor");
+      editor.enable();
+      var length = editor.editor.getDocument().toString().length;
+      editor.editor.setSelectedRange(length - 1);
+    };
+
+    /*
+      Replace the current proxy with an editor
```
deiferni: colon

bierik:
@deiferni updated

Issue #1990, opened 2016-07-04
Prevent multiple clicks on schedule button


deiferni:
Currently the schedule button can be clicked multiple times if the users is fast enough (or if the response is too slow) which leads to an error.

sentry: https://sentry.4teamwork.ch/sentry/onegov-gever/issues/493/

We should prevent that by e.g. displaying a spinner while the request is processed.

Adding some `sleep` to the schedule functions makes this easier to reproduce:
```diff
diff --git a/opengever/meeting/browser/resources/meeting.js b/opengever/meeting/browser/resources/meeting.js
index 13924c1..e06932c 100644
--- a/opengever/meeting/browser/resources/meeting.js
+++ b/opengever/meeting/browser/resources/meeting.js
@@ -306,7 +306,17 @@

     this.render = function(data) { return self.template({ proposals: data.items }); };

+    function sleep(milliseconds) {
+      var start = new Date().getTime();
+      for (var i = 0; i < 1e7; i++) {
+        if ((new Date().getTime() - start) > milliseconds){
+          break;
+        }
+      }
+    }
+
     this.schedule = function(target) {
+      sleep(2000);
       return $.post(target.attr("href")).done(function() { $("#filter-proposals").val(""); });
     };
```

bierik:
@deiferni a tip:
I would use the `network throttling` for testing this kind of issues in the chrome developer tools:
![screen shot 2016-09-16 at 10 45 22](https://cloud.githubusercontent.com/assets/1637820/18580297/03b108ca-7bfb-11e6-9632-a53ad5efd954.png)

deiferni:
https://sentry.4teamwork.ch/sentry/onegov-gever/issues/493/

PR #2160, opened 2016-09-16
Add loading option to controller

bierik:
⚠️  Depends on https://github.com/4teamwork/plonetheme.teamraum/pull/462

Closes https://github.com/4teamwork/opengever.core/issues/1990

This new behaviour of the controller makes it possible to turn a button in a loading button.

### DIFF
--- a/opengever/meeting/browser/resources/meeting.js
+++ b/opengever/meeting/browser/resources/meeting.js
@@ -19,10 +19,8 @@
       dialog.load();
     };
 
-    this.closeMeeting = function(target) {
-      target.addClass("loading");
+    this.closeMeeting = function() {
       return $.post(self.currentItem.attr("href"))
-              .always(function() { target.removeClass("loading"); })
               .done(function() { location.reload(); });
     };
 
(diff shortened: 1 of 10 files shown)


opengever/meeting/browser/resources/meeting.js
```diff
@@ -352,7 +348,8 @@
         target: ".schedule-proposal",
         callback: this.schedule,
         options: {
-          update: true
+          update: true,
+          loading: true
         }
       },
```
deiferni: what about `schedule-paragraph`?

opengever/meeting/browser/resources/meeting.js
```diff
@@ -352,7 +348,8 @@
         target: ".schedule-proposal",
```
deiferni: i just noticed that when scheduling a proposal/section with an empy text-field a request is sent which results in a server-side error. did the behavior change in this PR or is this an unrelated issue?

opengever/meeting/browser/resources/meeting.js
bierik: When the user adds a paragraph the input field is cleaned. So the user is not able to add the same paragraph again anyway.

bierik:
@deiferni updated

PR #2161, opened 2016-09-19
Add manual journal entry form

phgross:
bildschirmfoto 2016-09-19 um 12 30 05

bildschirmfoto 2016-09-19 um 12 28 56

See [Todo](https://basecamp.com/2768704/projects/12330648/todos/271609550) for more details

Besides the new functionality to add a journal entry manually, this PR fixes #2144 and changes the wrapper id for contact objects, it uses now `contact-{id}` instead of `person-{id}` or `organization-{id}`.

The styling adjustments are in separate [PR in plonetheme.teamraum](https://github.com/4teamwork/plonetheme.teamraum/pull/464).

@deiferni @lukasgraf

PR #2170, opened 2016-09-20
Make sure title of a manual journal entry is stored language independent

phgross:
It stores now the messagefactory object instead of the translated string as title of a manual journal entry.

Changelog not necessary just a small fix for #2161.


lukasgraf:
Pretty sure [those tests](https://jenkins.4teamwork.ch/job/opengever.core/1703/console) failed because they were running around midnight, and our test suite takes forever to complete 😢  - restarted the build.

PR #2207, opened 2016-10-04
Implement (basic) SQLSchemaDumper and also dump schemas for common SQL types.

lukasgraf:
This adds a minimal `SQLSchemaDumper` in order to also dump schemas for our SQL types.

The implementation currently is as minimal as possible, and well need to be built on later, but it already covers **listing available fields** (columns), and whether they're **required** (not nullable) or not.

The primary need for this is to validate SQL query results used for data migration against the structure we expect them to have.

@phgross @deiferni

### DIFF
--- a/docs/schema-dumps/_opengever.contact.models.OrgRole.json
+++ b/docs/schema-dumps/_opengever.contact.models.OrgRole.json
@@ -1,0 +1,40 @@
+{
+    "portal_type": "_opengever.contact.models.OrgRole",
+    "title": "OrgRole",
+    "schemas": [
+        {
+            "name": "OrgRole",
+            "fields": [
+                {
+                    "name": "id",
+                    "type": "Integer",
+                    "title": null,
+                    "desc": null,
+                    "required": true,
+                    "default": "Sequence(org_roles_id_seq)"
+                },
+                {
+                    "name": "function",
+                    "type": "String",
+                    "title": null,
+                    "desc": null,
+                    "required": false
+                },
+                {
+                    "name": "department",
+                    "type": "String",
+                    "title": null,
+                    "desc": null,
+                    "required": false
+                },
+                {
+                    "name": "description",
+                    "type": "UnicodeCoercingText",
+                    "title": null,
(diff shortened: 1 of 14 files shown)


docs/schema-dumps/_opengever.contact.models.OrgRole.json
```diff
@@ -0,0 +1,40 @@
...
+                {
+                    "name": "department",
+                    "type": "String",
+                    "title": null,
+                    "desc": null,
+                    "required": false
+                },
+                {
+                    "name": "description",
+                    "type": "UnicodeCoercingText",
```
deiferni: maybe you could map this to `Text`, afaik the coerce to unicode should be an internal thing only and not relevant for checks against the schema

opengever/base/schemadump/schema.py
```diff
@@ -87,6 +118,44 @@ def _get_fti(self, portal_type):
...
+            dottedname = sql_type.lstrip('_')
+            main_klass = resolve_dotted(dottedname)
+
+            schemas = []
+            schema_dumper = SQLSchemaDumper()
+
+            # Also consider columns from base classes (inheritance).
+            # We only consider one level of inheritance though, we don't look
+            # up base classes recursively.
+            bases = list(filter(_is_mapped_class, main_klass.__bases__))
```
deiferni: Hmm maybe also add an assertion for this? Otherwise we might get strange results when Inheritance hierarchy changes in the future?

docs/schema-dumps/_opengever.contact.models.OrgRole.json
lukasgraf: Yes, I intend to. 👍  I'm currently only doing something with `required`, using the types for validation will be the next step. I'll map this in an upcoming PR.

opengever/base/schemadump/schema.py
lukasgraf: 👍

lukasgraf:
@deiferni updated

PR #2224, opened 2016-10-10
Fix sorting in the journal listing.

### LINKED ISSUE #2221: Sorting of journal entries doesn't work
Clicking on a column in the journal tab just reverses the order of the journal entries.
However the entries are not sorted on the selected column.

The problem lies in the custom compare function in
https://github.com/4teamwork/opengever.core/blob/master/opengever/journal/tab.py#L123

The items in the results list are dicts, thus the sort key should be accessed by key.
Right now it's comparing empty string with empty string.

phgross:
Closes #2221 

@deiferni or @lukasgraf

### DIFF
--- a/opengever/journal/tab.py
+++ b/opengever/journal/tab.py
@@ -12,6 +12,7 @@ from opengever.journal import _
 from opengever.tabbedview import BaseListingTab
 from opengever.tabbedview import GeverTableSource
 from opengever.tabbedview.helper import linked_ogds_author
+from opengever.tabbedview.helper import readable_ogds_user
 from plone import api
 from zope.annotation.interfaces import IAnnotations
 from zope.browserpage.viewpagetemplatefile import ViewPageTemplateFile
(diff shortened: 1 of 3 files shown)


opengever/journal/tab.py
```diff
@@ -107,6 +109,13 @@ class JournalTableSource(GeverTableSource):
     grok.implements(ITableSource)
     grok.adapts(IJournalSourceConfig, Interface)
 
+    sorter = {
+        'time': lambda a, b: cmp(a.get('time', ''), b.get('time', '')),
```
deiferni: It may improve readability to extract the lambdas and create a function for each of them instead.

opengever/journal/tab.py
```diff
@@ -107,6 +109,13 @@ class JournalTableSource(GeverTableSource):
...
+        'title': lambda a, b: cmp(a.get('action', {}).get('title'),
+                                  b.get('action', {}).get('title')),
+        'actor': lambda a, b: cmp(linked_ogds_author(a, a.get('actor')),
```
phgross: @lukasgraf your right. I've updated the code, the actor sorting method uses now the cached helper method `readable_ogds_user`, which only returns the label of an actor.

phgross:
@deiferni @lukasgraf updated ...

PR #2226, opened 2016-10-10
SchemaDumper: Include max_length for stringish field types.

lukasgraf:
Includes `max_length` in the field dumps for string-like fields.

Also maps `SQLAlchemy` column types to corresponding `zope.schema` field types.

@phgross @deiferni

### DIFF
--- a/opengever/base/schemadump/field.py
+++ b/opengever/base/schemadump/field.py
@@ -183,6 +183,14 @@ class SQLFieldDumper(object):
     """Dumps a simple Python representation of a SQLAlchemy Column.
     """
 
+    # Mapping from SQLAlchemy column types to zope.schema field types
+    SQL_FIELD_TYPES = {
+        'Integer': 'Int',
+        'String': 'Text',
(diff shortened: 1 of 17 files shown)


opengever/base/schemadump/field.py
```diff
@@ -183,6 +188,14 @@ class SQLFieldDumper(object):
     """Dumps a simple Python representation of a SQLAlchemy Column.
     """
 
+    # Mapping from SQLAlchemy column types to zope.schema field types
+    SQL_FIELD_TYPES = {
+        'Integer': 'Int',
+        'String': 'Text',
```
phgross: Maybe a mapping to `TextLine` would correlate better?

lukasgraf: I briefly considered that, but no, I don't think so - `TextLine` is defined as "one line of text at most" (no newlines). An SQL `String` on the other hand can contain text passages of arbitrary length, including newlines and everything. So while for some of our models it might be intended as a line of text, we can't assume this for the general case, and should choose the more generic type.

PR #2216, opened 2016-10-05
Add address selection for recipients when creating documents from a template.

deiferni:
This PR adds address selection for recipients when creating documents from a template.
Contacts, Org-Roles and ogds-users are supported as recipients.

It also contains minor refactoring of the `create_document_from_template` form and introduces a new widget that render a radio-button selection list (see screenshots).

The form is now a two-step wizard when the contact-feature is enabled:

screen shot 2016-10-05 at 15 38 27
screen shot 2016-10-05 at 15 41 47

With a disabled contact-feature the form does not change and does not display the recipient field.
This is the second part of #2200 and closes #2186.

PR #2229, opened 2016-10-11
Fix messagefactory import.

### LINKED ISSUE #2228: Missing translation for empty table-selection widget
screen shot 2016-10-11 at 11 50 21

deiferni:
Fixes missing translation for message that is displayed when the widget is empty:

screen shot 2016-10-11 at 12 01 34

Closes #2228.

*CL should not be necessary, it is only a small fix for a recently merged feature (#2216).*


deiferni:
@phgross plz have a look

PR #2233, opened 2016-10-12
Use ooxml_docprops force mode to overwrite properties of a wrong type.

deiferni:
Depends on https://github.com/4teamwork/ooxml_docprops/pull/5

When creating templates users already add some doc-properties to fill template-slots. The values of these properties do never matter since they will always be automatically filled by the system at a later point with a different value. Thus we should be able to safely convert the properties to their proper type when writing.


deiferni:
@phgross plz have a look

PR #2241, opened 2016-10-18
Readd prepoverlay.js

phgross:
The prepoverlay js was removed by mistake in the javsascript dependencies cleanup PR (see #2159).

@deiferni 

cc @bierik


deiferni:
👍

PR #2247, opened 2016-10-20
Update showroom

bierik:
⚠️  Depends on https://github.com/4teamwork/ftw.showroom/pull/41


bierik:
@phgross Tested within GEVER

PR #2248, opened 2016-10-21
Mark workflow state with css classes

bierik:
Prepare the workflow state cells for styling

### DIFF
--- a/opengever/meeting/tabs/proposallisting.py
+++ b/opengever/meeting/tabs/proposallisting.py
@@ -16,7 +16,7 @@ def proposal_link(item, value):
 
 
 def translated_state(item, value):
-    wrapper = '<span class="{0}">{1}</span>'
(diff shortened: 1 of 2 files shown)


opengever/globalindex/model/task.py
```diff
@@ -212,7 +212,7 @@ def get_deadline_label(self, format="medium"):
             return u''
 
         if self.is_overdue:
-            label = u'<span class="overdue">{}</span>'
+            label = u'<span class="wf-task-state-overdue">{}</span>'
```
phgross: `overdue` is not a workflow state, its also displayed in a separate row in the task listings.

phgross: `deadline-state-overdue`

bierik:
@phgross updated

PR #2257, opened 2016-11-01
Add a tab to list all periods of a committee.

deiferni:
This PR adds a tab that lists all periods that have been created for a committee. It converts the committee tabs to a non fti-action based tabbedview. It also adds a base class that allows convenient definitions of non fti-action based tabbedviews.

screen shot 2016-07-01 at 21 27 51

Part of #1373.


opengever/tabbedview/tabbed.py
```diff
@@ -1,7 +1,50 @@
 from ftw.tabbedview.browser.tabbed import TabbedView
 
 
-class ModelProxyTabbedView(TabbedView):
+class GeverTabbedView(TabbedView):
```
phgross: Move both views `GeverTabbedView` and `ModelProxyTabbedView` to opengever.tabbedview.browser

deiferni: @phgross: short answer: i'd like to keep it there if you don't mind.

Currently the stuff in `opengever.tabbeview.browser` is all a concrete implementation of a tabbedview, e.g. the tabbedviews for tasks. These implementations should go away though and be moved to the package they're for. The package `opengever.tabbedview` should then become a package for tabbedview base-classes and global customizations.

For that reason i'd like to keep the base-classes apart from the concrete implementations, and i'd like to keep them in the base package for now. Also once the implementations are moved out the `opengever.tabbedview` package will become quite small, so the browser package might go away completely. 

Btw: the two classes are also re-imported in `__init__.py` so they might be imported conveniently as follows: `from opengever.tabbedview import GeverTabbedView `

deiferni: ok - as discussed we'll follow the convention as long as the browser package exists 👍

deiferni:
@phgross updated

PR #2266, opened 2016-11-04
Remove OGBrowser and use ftw.testbrowser instead

phgross:
This PR reworks all tests wich currently used the OGBrowser to consequently use the `ftw.testbrowser` instead.

Beside that I found two views (`OpenDossiersAsJSONView` and `DocumentsOfDossierAsJSONView`) which aren't used any more. I removed them.

@deiferni or @lukasgraf


opengever/task/tests/test_workflow.py
```diff
@@ -109,13 +109,14 @@ def test_document_in_a_closed_tasks_are_still_editable(self):
                           .titled(u'Letter for Peter')
                           .within(task))
 
-        self.browser.open('%s/edit' % (document.absolute_url()))
-        self.browser.fill({'Title': 'New Title'})
-        self.browser.click('Save')
+        browser.login().open(document, view='edit')
+        browser.fill({'Title': 'New Title'})
```
deiferni: what about `browser.fill({'Title': 'New Title'}).submit()`?

phgross: Unfortunately not it submits the "autocomplete widget form" 😏

PR #2328, opened 2016-11-24
Ungrok disposition subpackage

phgross:
This PR ungroks the disposition subpackage, where it's possible. 

Currently the `BaseCatalogListingTab` is a grok base class, so I can't ungrok the disposition listing for now.

Closes #2328 

@deiferni @lukasgraf


opengever/disposition/appraisal.py
```diff
@@ -1,27 +1,27 @@
...
 from opengever.disposition.interfaces import IDisposition
 from persistent.dict import PersistentDict
 from zope.annotation.interfaces import IAnnotations
+from zope.component import adapter
 from zope.component import getUtility
+from zope.interface import implementer
 from zope.intid.interfaces import IIntIds
 
 
-class Appraisal(grok.Adapter):
```
deiferni: There would also be the `zcml` adapter registration. might be something we need to discuss first, and agree upon a convention?

```
    <adapter factory=".." provides=".." for=".." />
```

deiferni: ok, as discussed we want to use the decorators, so 👍

PR #2329, opened 2016-11-24
Make disposition title editable, but prefill it with a default value.

phgross:
Default value format: `Disposition {admin unit abbreviation} {localized today's date}`

See https://basecamp.com/2768704/projects/12554551/todos/274416169

@deiferni or @lukasgraf

### DIFF
--- a/opengever/testing/builders/dx.py
+++ b/opengever/testing/builders/dx.py
@@ -440,5 +440,12 @@ builder_registry.register('private_dossier', PrivateDossierBuilder)
 class DispositionBuilder(DexterityBuilder):
     portal_type = 'opengever.disposition.disposition'
 
+    def __init__(self, session):
+        super(DispositionBuilder, self).__init__(session)
+        # OfferedDossiersValidator is broken when this is import is in
+        # module scope.
+        from opengever.disposition.disposition import title_default
+        self.arguments['title'] = title_default()
+
 
 builder_registry.register('disposition', DispositionBuilder)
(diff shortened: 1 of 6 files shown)


opengever/testing/builders/dx.py
```diff
@@ -440,5 +440,9 @@ class PrivateDossierBuilder(DossierBuilder):
 class DispositionBuilder(DexterityBuilder):
     portal_type = 'opengever.disposition.disposition'
 
+    def __init__(self, session):
+        super(DispositionBuilder, self).__init__(session)
+        self.arguments['title'] = 'Angebot FD 23.11.2010'
```
lukasgraf: This will be incorrect once the title default suggestion format changes. Couldn't you just call `title_default()` here?

opengever/disposition/defaults.py
```diff
@@ -0,0 +1,19 @@
+from datetime import date
+from opengever.disposition import _
+from opengever.ogds.base.utils import get_current_admin_unit
+from plone import api
+from zope.globalrequest import getRequest
+from zope.i18n import translate
+
+
+def title_default():
```
deiferni: I'd consider to move this back to `Disposition` as a `@staticmethod` to have the method close to its related class and to make their relationship clear to future coders. You could then write `Disposition.default_title()` which imo improves readability as well 😉 .

phgross:
@deiferni updated as discussed

Issue #2306, opened 2016-11-22
Create OGGBundle Transmogrifier Source


lukasgraf:
Create a Transmogrifier Source that is capable of reading an OGGBundle and yielding appropriate pipeline items.

lukasgraf:
Closed via #2337 and #2339

PR #2339, opened 2016-11-29
Add guid resolver section for oggbundles.

deiferni:
See https://github.com/4teamwork/opengever.core/issues/2306.

This PR adds a transmogrifier section that resolves and validate GUIDs.

It also yields items in an order that guarantees that parents are always positioned before their children. This is achieved by building a temporary tree, then re-yielding the nodes in pre-order.

This section also validates that:
- each item has a guid
- each guid is unique
- parent pointers are valid, should they exist

*CL will follow in a later PR*

### DIFF
--- a/opengever/setup/sections/resolveguid.py
+++ b/opengever/setup/sections/resolveguid.py
@@ -13,12 +13,17 @@ class DuplicateGuid(Exception):
     pass
 
 
+class MissingParent(Exception):
+    pass
+
+
 class ResolveGUIDSection(object):
     """Resolve and validate GUIDs.
 
     This section also validates that:
         - each item has a guid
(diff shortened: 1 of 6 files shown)


opengever/setup/sections/resolveguid.py
```diff
@@ -0,0 +1,50 @@
...
+class MissingGuid(Exception):
+    pass
+
+
+class DuplicateGuid(Exception):
+    pass
+
+
+class ResolveGUIDSection(object):
+    """Resolve and validate GUIDs.
```
phgross: what is the GUID? I would add a short description to the docstring.

deiferni:
@phgross updated

PR #2344, opened 2016-11-30
Clean up test output

### LINKED ISSUE #2288: Define the i18n translation domain for all subproducts
Currently running tests is unnecessarily verbose as zope.configuration is complaining about configure.zcml missing a translation domain for most of the fields per subproduct.

This also has a minor contribution to test fixture generation time.

Rotonen:
* Fixed missing i18n_domain related UserWarnings
* Squelched log levels DEBUG and INFO from bleeding onto the test output
* Added changelog entries

Fixes #2288

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,13 @@ Changelog
 4.13.1 (unreleased)
 -------------------
 
+- Add missing i18n_domain entries into subproduct configure.zcml files in
+  order to squelch UserWarning messages bleeding into test output.
+  [Rotonen]
+
+- Block log levels DEBUG and INFO from bleeding into test output.
+  [Rotonen]
+
 - Add new role Archivist and let them see dispositions and offered and
   archived dossiers.
   [phgross]
--- a/opengever/core/testing.py
+++ b/opengever/core/testing.py
@@ -48,7 +48,7 @@ import transaction
 
 
 loghandler = logging.StreamHandler(stream=sys.stdout)
-loghandler.setLevel(logging.DEBUG)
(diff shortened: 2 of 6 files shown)


docs/HISTORY.txt
```diff
@@ -5,6 +5,13 @@ Changelog
 4.13.1 (unreleased)
 -------------------
 
+- Add missing i18n_domain entries into subproduct configure.zcml files in
+  order to squelch UserWarning messages bleeding into test output.
+  [Rotonen]
+
+  - Block log levels DEBUG and INFO from bleeding into test output.
```
deiferni: too much whitespace before `-`

opengever/testing/test_case.py
```diff
@@ -34,6 +35,10 @@ class FunctionalTestCase(TestCase):
     layer = OPENGEVER_FUNCTIONAL_TESTING
     use_default_fixture = True
 
+    # Squelch INFO and DEBUG from bleeding onto stdout
+    # We previously had plone.protect garble test output with its INFO messages
+    logging.disable(logging.WARNING)
```
deiferni: i'd have a look at `opengever.core.testing.py` it already seems to fiddle with logging and installs some loghandlers. might be that you can remove some stuff there instead of adding this statement 😉 .

if this statement should still be necessary some way or the other i'd also move it to `opengever.core.testing.py` (module scope).

Rotonen: Why do we set the loglevel to DEBUG there? During testing DEBUG and INFO should not be interesting.

I will just set that to WARNING there instead, then.

Rotonen:
@deiferni ready for rereview.

PR #2353, opened 2016-11-30
Add separte permission "Download SIP Package"

### LINKED ISSUE #2326: Add a separate "SIP download" permission
The SIP download should be protected by a separate permission.

phgross:
The download is only available in the disposed state of a disposition for the Archivist and Record Manager Roles. For Managers its available in every state, because it could be useful for debugging purpose.

Closes #2326 

@deiferni


docs/HISTORY.txt
```diff
@@ -5,6 +5,9 @@ Changelog
 4.13.1 (unreleased)
 -------------------
 
+- Add separte permission "Download SIP Package".
```
deiferni: separate

opengever/disposition/profiles/default/workflows/opengever_disposition_workflow/definition.xml
```diff
@@ -1,79 +1,84 @@
 <?xml version="1.0"?>
 <dc-workflow workflow_id="opengever_disposition_workflow" title="DC Workflow Definition" description="" state_variable="review_state" initial_state="disposition-state-in-progress" manager_bypass="False">
-  <permission>View</permission>
```
deiferni: did you re-export and replace the file? please don't do that 😞  but modify the existing file in such a way that unchanged content is not re-ordered. this makes reviewing easier.

opengever/disposition/tests/test_overview.py
```diff
@@ -127,17 +127,22 @@ def test_lists_possible_transitions_in_actionmenu_as_buttons(self, browser):
                           browser.css('.transitions li').text)
 
     @browsing
-    def test_sip_download_is_active_in_disposed_state(self, browser):
+    def test_sip_download_is_only_available_in_disposed_state(self, browser):
         self.grant('Records Manager')
         browser.login().open(self.disposition, view='tabbedview_view-overview')
+        self.assertNotIn('Download disposition package',
```
deiferni: This is very fragile and won't fail if the action name is changed slighlty but starts appearing due to a bug. Instead i'd test against an explicit list of actions, something like `assertEquals(['action1', 'action 2'], browser.css('ul.ALL_ACTION_SELECTOR'))`

phgross: No its not, this test contains also a check where it checks the existence of the action. So the test would fail if the action gets renamed ...

opengever/disposition/profiles/default/workflows/opengever_disposition_workflow/definition.xml
phgross: Everything else is very fault-prone ... but can modify the xml if you like

opengever/disposition/tests/test_overview.py
deiferni: hm ok, but still you could forget to adapt the test that check for absence. i'd extract the action name into constant that is used by both tests.

opengever/disposition/profiles/default/workflows/opengever_disposition_workflow/definition.xml
deiferni: if manageable yes, please do. otherwise next time 😉

phgross:
@deiferni updated, the diff of the workflow xml should now be easier to read.

PR #2354, opened 2016-11-30
Replace INameFromTitle behavior with adapter for repository roots.

### LINKED ISSUE #2350: repositoryroot translated title behavior conflicts with namefromtitle
When both `INameFromTitle` and `ITranslatedTitle` are active for a content-type, choosing correct names based on the translated title fails.

`INameChooser` will always lookup the object itself with `INameFromTitle(obj)` instead (behavior) of the now necessary adapter which should return a translated title (just as for `IRepositoryFolder`s)

Proposed solution:
- drop `plone.app.content.interfaces.INameFromTitle` from `RepositoryRoot`
- add `RepositoryRootNameFromTitle` adapter see `RepositoryFolder`

deiferni:
This fixes choosing names from the title in combination with ITranslatedTitle.

Fixes #2350.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.13.1 (unreleased)
 -------------------
 
+- Replace INameFromTitle behavior with adapter for repository roots.
+  [deiferni]
+
 - Add separate permission "Download SIP Package".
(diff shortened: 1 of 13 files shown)


deiferni:
@phgross updated

Issue #2308, opened 2016-11-22
Build transmogrifier pipeline for OGGBundle import


lukasgraf:
Build the transmogrifier pipeline for OGGBundle import.

This will compose 
- OGGBundle Source
- GUIDResolverSection
- Local role setter section
- File loader section
- Review State setter
- ...

into a complete pipeline to create GEVER content from an OGGBundle.

lukasgraf:
Base pipeline has been implemented.

PR #2357, opened 2016-12-02
Add custom constructor section for oggbundle.

deiferni:
See https://github.com/4teamwork/opengever.core/issues/2308.

This PR adds a transmogrifier section that constructs instances.

*CL will follow in a later PR*

### DIFF
--- a/opengever/setup/sections/constructor.py
+++ b/opengever/setup/sections/constructor.py
@@ -1,0 +1,62 @@
+from collective.transmogrifier.interfaces import ISection
+from collective.transmogrifier.interfaces import ISectionBlueprint
+from plone import api
+from plone.dexterity.utils import createContentInContainer
+from zope.interface import classProvides
+from zope.interface import implements
+import logging
+
+
+logger = logging.getLogger('opengever.setup.constructor')
+
+
+class InvalidType(Exception):
+    pass
+
+
+class ConstructorSection(object):
+    classProvides(ISectionBlueprint)
+    implements(ISection)
+
+    def __init__(self, transmogrifier, name, options, previous):
+        self.previous = previous
+        self.transmogrifier = transmogrifier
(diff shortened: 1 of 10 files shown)


opengever/setup/sections/constructor.py
```diff
@@ -0,0 +1,62 @@
...
+            if fti is None:
+                raise InvalidType(portal_type)
+
+            try:
+                # we need the title sometimes to auto-generate ids
+                # XXX maybe be a bit more intelligent here and set all required
+                # fields while constructing?
+                obj = createContentInContainer(
+                    context, portal_type,
+                    title_de=item.get(u'title_de'),
```
phgross: Does this attribute has impact for object without the TranslatedTitle behavior? Maybe extending the tests would be a good thing ...

deiferni: hmm right, should be extra careful here, update implementation to check for `ITranslatedTitle` behaviour and only then user the language dependent fields.

deiferni:
@phgross updated

opengever/setup/sections/constructor.py
```diff
@@ -0,0 +1,71 @@
...
+            fti = self.ttool.getTypeInfo(portal_type)
+            if fti is None:
+                raise InvalidType(portal_type)
+
+            try:
+                # we need the title sometimes to auto-generate ids
+                # XXX maybe be a bit more intelligent here and set all required
+                # fields while constructing?
+                if self.has_translated_title(fti):
+                    title_keys = (u'title_de', u'title_fr')
```
phgross: Why not using the `TRANSLATED_TITLE_NAMES`?

deiferni: 🙈  that's why i've refactored it in the first place, thanks ...

deiferni:
@phgross let's try again 😉  - updated

PR #2372, opened 2016-12-05
Disposition history and removal protocol fixes

phgross:
- Make sure that destroying dossiers doesn't get logged to disposition history as an edited event.
Currently the modified event handler creates a history entry for each dossiers which gets removed, this is unnecessary and wrong. Therefore we mark the request with a marker interface, to ignore ModifiedEvents while destroying dossiers.
- Fixed transition label for the added and edited action.
- Fixed typo in german translations.
- Don't display `None` as transfer number in the removal protocol, if there is no transfer number.

@deiferni  or @lukasgraf

### DIFF
--- a/opengever/disposition/browser/removal_protocol.py
+++ b/opengever/disposition/browser/removal_protocol.py
@@ -123,7 +123,8 @@ class RemovalProtocolLaTeXView(grok.MultiAdapter, MakoLaTeXView):
              'value': self.context.title},
             {'label': _('label_transfer_number',
                         default=u'Transfer number'),
-             'value': self.context.transfer_number}
+             'value': self.context.transfer_number if
+                 self.context.transfer_number else u''}
         ]
 
         for row in config:
--- a/opengever/disposition/history.py
+++ b/opengever/disposition/history.py
@@ -80,6 +80,13 @@ class Added(DispositionHistory):
         return _('msg_disposition_added', default=u'Added by ${user}',
                  mapping=self._msg_mapping)
 
+    @property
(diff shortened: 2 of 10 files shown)


opengever/disposition/interfaces.py
```diff
@@ -11,3 +11,8 @@ class IHistoryStorage(Interface):
 
 class IDisposition(Interface):
     """The disposition Interface."""
+
+
+class IDuringDossierDestroyment(Interface):
```
deiferni: ...Destruction


_destroyment isn't even a word 😉_

opengever/disposition/browser/removal_protocol.py
```diff
@@ -123,7 +123,8 @@ def get_disposition_metadata(self):
              'value': self.context.title},
             {'label': _('label_transfer_number',
                         default=u'Transfer number'),
-             'value': self.context.transfer_number}
+             'value': self.context.transfer_number if
+             self.context.transfer_number else u''}
```
deiferni: indent more by one?

opengever/disposition/history.py
```diff
@@ -80,6 +80,13 @@ def msg(self):
         return _('msg_disposition_added', default=u'Added by ${user}',
                  mapping=self._msg_mapping)
 
+    @property
+    def transition_label(self):
```
deiferni: Could this be tested? Is it visible in the UI?

phgross:
@deiferni and @lukasgraf updated ...

Issue #2290, opened 2016-11-16
Scheduled maintenance tasks


phgross:
**Usecases**
GEVER:
 - daily/weekly removal of trashed documents in resolved GEVER Dossiers, when `Karrenzfrist` is expired
 - daily/weekly removal of the original file in resolved GEVER Dossiers, when `Karrenzfrist` is expired
 - Consistency checks, such as searching for dossiers in non-leaf-repository-folders, and notifying when found (email?)

myteamraum:
 - Scheduled sending of billing e-mails (once per month)
 
plone maintenance in general:
 - Catalog: remove brains of missing objects.
 - Catalog: scheduled/batched reindex of objects.
 - Add-On cleanups, such as [`ftw.copymovepatches`' fixes](https://github.com/4teamwork/ftw.copymovepatches#backport-of-dexterity-patch-set-copy-flags)

**Ideas, questions and thoughts**
- Add-ons should provide their maintenance tasks.
- Maintenance tasks can be checks for errors and / or fixes for those errors.
- There should be a standardized way for notifying on (unresolvable) problems (email? slack?).
- For very long running tasks, e.g. reindex `SearchableText` for all objects, the system should be able to split that up into batches and run them over a period of multiple nights.
- Maybe we need to be able to schedule tasks on a specific date / time (business use cases), but we also need to have "general maintenance" tasks which should be run "sometime in a night, maybe once a week".
- It should be possible to schedule a job once, e.g. for the next night, such as complete catalog reindexes. This is invoked when we suspect something to be wrong and have a fix, such as reindexing, but it takes too long (conflicts) and should be done in the night.
- There should be a nice control panel, where we can see what did happen (executed jobs) and which jobs are scheduled.
- One-time-jobs could be triggered through the control panel.
- It would be awesome if periodically jobs could be registered in buildout, so that we can schedule it for a different time per deployment on the same server.

Rotonen:
Initial thoughts:

* Zope's clock-server can be used to process maintenance tasks in guaranteedly small chunks at a high frequency
* I prefer cronjobs + `bin/instance run` for stuff which requires guarantees of when the jobs run
  * Cronjob management should be done via deployment configuration management tools
    * Ansible can be married into buildout - or the other way around
* There is no real need for an actual message queue based runner
  * This would complicate getting guaranteed feedback out of the tasks

There seems to be exist decent scaffolding to build upon:
https://github.com/plone/ansible.plone_server

Rotonen:
[...]

PR #2377, opened 2016-12-06
 Add maintenance job to erase source files when waiting period expires

phgross:
This PR adds the functionality to remove the source files, as soon the waiting period of the dossier has been expired.
This functionality should be triggered by a periodic maintenance job. But some general work and decisions has to be done (see #2290 for more information). For now there is just a zopectl_handler `bin/instance erase_source_files` to test the functionality.

The source file will only be removed, when an archival file exists.

The waiting period (range between the end date of the dossier and the source file removal) is configurable via registry entry. 2 years is the default. 
The erasement can be enabled  or disabled with a separate registry entry, by default the erasement is disabled.

@deiferni @lukasgraf

### DIFF
--- a/opengever/dossier/events.py
+++ b/opengever/dossier/events.py
@@ -24,3 +24,11 @@ class ParticipationRemoved(ObjectEvent):
     def __init__(self, obj, participant):
         self.object = obj
         self.participant = participant
+
+
+class SourceFileErased(ObjectEvent):
+    """The `SourceFileErased` is fired after the source file of a document
+    is erased during a maintenance job.
+    """
+
+    implements(interfaces.ISourceFileErased)
--- a/opengever/dossier/maintenance.py
+++ b/opengever/dossier/maintenance.py
@@ -1,6 +1,8 @@
 from datetime import date
 from dateutil.relativedelta import relativedelta
 from ftw.mail.mail import IMail
+from opengever.core.debughelpers import get_first_plone_site
+from opengever.core.debughelpers import setup_plone
 from opengever.document.behaviors import IBaseDocument
 from opengever.document.behaviors.metadata import IDocumentMetadata
 from opengever.document.document import IDocumentSchema
@@ -9,6 +11,12 @@ from opengever.dossier.events import SourceFileErased
 from opengever.dossier.interfaces import IDossierResolveProperties
 from plone import api
 from zope.event import notify
+import logging
+import transaction
+
+
+logger = logging.getLogger('opengever.dossier')
+logger.setLevel(logging.INFO)
 
 
 class SourceFileEraser(object):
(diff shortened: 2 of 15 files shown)


lukasgraf:
@phgross could you please add a bullet list to the PR body of the exact conditions that need to be fulfilled in order for a file to be deleted? E.g.

- Containing dossier is in state resolved (?)
- Waiting period has passed. What does that exactly mean though? `IDossier.end` + `waiting_period_source_file_removal` * 365 > today? Or same date as `waiting_period_source_file_removal`  years before?)
- `source_file_removal_enabled` feature flag is true
 - anything else?

opengever/dossier/maintenance.py
```diff
@@ -0,0 +1,46 @@
...
+        for dossier in self.get_dossiers_to_erase():
+            for document in api.content.find(dossier, object_provides=IBaseDocument):
+                self.erase_source_file(document.getObject())
+
+    def erase_source_file(self, document):
+        if not IDocumentMetadata(document).archival_file:
+            return
+
+        if IDocumentSchema.providedBy(document):
+            IDocumentSchema(document).file = None
```
lukasgraf: Does just setting `file` to `None` actually result in the blob being removed from blob storage? _(It doesn't for me in a quick test)_

opengever/dossier/events.py
```diff
@@ -24,3 +24,11 @@ class ParticipationRemoved(ObjectEvent):
     def __init__(self, obj, participant):
         self.object = obj
         self.participant = participant
+
+
+class SourceFileErased(ObjectEvent):
+    """The `SourceFileErased` is fired after the source file of a document
```
lukasgraf: The `SourceFileErased` -> The `SourceFileErased` event

opengever/dossier/events.py
```diff
@@ -24,3 +24,11 @@ class ParticipationRemoved(ObjectEvent):
...
+    is erased during a maintenance job.
```
lukasgraf: ` is erased` -> `has been erased`

opengever/dossier/maintenance.py
```diff
@@ -35,6 +37,8 @@ def erase(self):
             for document in api.content.find(dossier, object_provides=IBaseDocument):
                 self.erase_source_file(document.getObject())
 
+            notify(SourceFileErased(dossier))
```
lukasgraf: Same event for dossier and document? The meaning is slightly different in the two cases (files vs. file)

opengever/dossier/maintenance.py
```diff
@@ -45,6 +47,11 @@ def erase(self):
         dossiers = self.get_dossiers_to_erase()
 
         for dossier in dossiers:
+            if ISourceFileHasBeenErased.providedBy(dossier):
```
lukasgraf: Shouldn't this marker be on the document instead of dossier? What if documents get moved to a different dossier?

opengever/dossier/maintenance.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def erase(self):
+        if not self.is_enabled():
+            logger.info('Source file erasement aborted, feature is disabled.')
+            return
+
+        logger.info('Source file erasement started.')
+        dossiers = self.get_dossiers_to_erase()
+
+        for dossier in dossiers:
```
lukasgraf: This will walk over an increasingly large list of dossiers over time, most of which will already have had their documents' source files removed. That means performance will degrade substantially over time, particularly if 90% of dossier objects are being woken up all the time.

phgross:
With the change in the expiration calculation, we'll change that functionality to a manually triggered removal.

PR #2384, opened 2016-12-06
Remove unused variable in the ObjectCreatorCommand

elioschmutz:
Remove the `skip_defaults_fields` variable from the `ObjectCreatorCommand` and its subobjects because it's never used.


elioschmutz:
Wird in PR #2386 weiterbearbeitet.

PR #2385, opened 2016-12-06
Enable zip exports for tasks.

Rotonen:
* Create an upgrade step for the zipexport interface setting
* Add the interface
* Add tests for the export actions being there

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.13.1 (unreleased)
 -------------------
 
+- Enable zip exports for tasks
+  [Rotonen]
+
 - Hide private root from breadcrumbs.
   [deiferni]
 
--- a/opengever/task/tests/test_overview.py
+++ b/opengever/task/tests/test_overview.py
@@ -43,7 +43,7 @@ class TestTaskOverview(FunctionalTestCase):
             self.user.label(), browser.css('.issuer a').first.text)
 
     @browsing
-    def test_issuer_is_prefixed_by_current_org_unit_on_a_multiclient_setup(self, browser):
+    def test_issuer_is_prefixed_by_current_org_unit_on_a_multiclient_setup(self, browser): # noqa
         create(Builder('org_unit').id('client2')
                .having(admin_unit=self.admin_unit))
         task = create(Builder("task").having(task_type='comment',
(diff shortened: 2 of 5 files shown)


docs/HISTORY.txt
```diff
@@ -4,6 +4,8 @@ Changelog
 
 4.13.1 (unreleased)
 -------------------
+- Enable zip exports for tasks
```
phgross: Newline above is missing.

opengever/task/tests/test_overview.py
```diff
@@ -57,13 +57,19 @@ def test_issuer_is_prefixed_by_current_org_unit_on_a_multiclient_setup(self, bro
 
     @browsing
     def test_date_of_completion_is_displayed_correclty(self, browser):
-        task = create(Builder("task").having(task_type='comment',
-                                             date_of_completion=date(2015, 02, 02),
-                                             issuer=TEST_USER_ID))
+        task = create(
```
phgross: We ignore the pyflakes length limit in tests. It just makes everything more unreadable in most of the cases.

opengever/task/tests/test_overview.py
```diff
@@ -89,6 +95,41 @@ def test_documents_are_listed(self, browser):
             browser.css('#documentsBox a.document_link').text)
 
     @browsing
+    def test_zip_export_actions_are_available(self, browser):
+        dossier = create(
```
phgross: please use the following formatting when using the builder, as we use everywhere else:

```python
        task = create(Builder("task")
                      .within(dossier)
                      .titled(u'Aufgabe')
                      .having(text='Text blabla',
                              task_type='comment',
                              deadline=datetime(2010, 1, 1),
                              issuer=TEST_USER_ID,
                              responsible=TEST_USER_ID))
```

opengever/task/tests/test_overview.py
```diff
@@ -89,6 +95,41 @@ def test_documents_are_listed(self, browser):
...
+            .within(task)
+            )
+
+        browser.login()
+        browser.open(task)
+        self.assertIsNotNone(
+            browser.css('.actionicon-object_buttons-zipexport'))
+
+        browser.open(task, view='tabbedview_view-documents')
+        self.assertIn('Export as Zip',
```
phgross: The action was already available before the change, but hasn't worked, so a test wich really made an zip-export of documents on a task, would be good.

opengever/task/tests/test_overview.py
```diff
@@ -57,13 +57,19 @@ def test_issuer_is_prefixed_by_current_org_unit_on_a_multiclient_setup(self, bro
...
     @browsing
     def test_date_of_completion_is_displayed_correclty(self, browser):
-        task = create(Builder("task").having(task_type='comment',
-                                             date_of_completion=date(2015, 02, 02),
-                                             issuer=TEST_USER_ID))
+        task = create(
+            Builder("task")
+            .having(
+                task_type='comment',
+                date_of_completion=date(2015, 02, 02),
```
deiferni: i wouldn't use octal numbers here, people might start to wonder why `08` doesn't work 😉

Rotonen:
@deiferni @phgross ready for rereview.

PR #2404, opened 2016-12-08
OGGBundle pipeline: Add basic FileLoader section.

lukasgraf:
This adds a basic `FileLoader` section that loads files into documents based on paths.

The distinction between `opengever.document.document` and `ftw.mail.mail` (for `.eml`) is still to be done, otherwise this should be pretty complete.

@deiferni


opengever/setup/sections/fileloader.py
```diff
@@ -0,0 +1,109 @@
...
+from opengever.setup.sections.jsonsource import BUNDLE_PATH_KEY
+from opengever.setup.sections.jsonsource import JSON_STATS_KEY
+from zope.annotation.interfaces import IAnnotations
+from zope.interface import classProvides
+from zope.interface import implements
+import logging
+import os.path
+
+
+logger = logging.getLogger('opengever.bsmigration.fileloader')
```
deiferni: bsmigration 😉

opengever/setup/sections/fileloader.py
```diff
@@ -0,0 +1,109 @@
...
+            pathkey = self.pathkey(*keys)[0]
+
+            if self.key in item:
+                filepath = item[self.key]
+                if filepath is None:
+                    yield item
+                    continue
+
+                if pathkey not in item:
+                    logger.warning("Missing path key for file %s" % filepath)
```
deiferni: we might want to let a validator section check if the file exists and fail hard here instead

opengever/setup/sections/fileloader.py
```diff
@@ -0,0 +1,109 @@
...
+                    self.stats['errors']['files_not_found'][filepath] = path
+                    yield item
+                    continue
+
+                mimetype, _encoding = guess_type(filepath, strict=False)
+                if mimetype is None:
+                    logger.warning("Unknown mimetype for file %s" % filepath)
+                    mimetype = 'application/octet-stream'
+
+                obj = traverse(self.context, path, None)
```
deiferni: currently the constructed instance is also available as `item['_object']`

opengever/setup/sections/fileloader.py
```diff
@@ -0,0 +1,109 @@
...
+                if obj is None:
+                    logger.warning(
```
deiferni: this should not happen, ever. maybe also fail hard?

opengever/setup/sections/fileloader.py
```diff
@@ -0,0 +1,109 @@
...
+from zope.interface import classProvides
+from zope.interface import implements
+import logging
+import os.path
+
+
+logger = logging.getLogger('opengever.bsmigration.fileloader')
+
+
+class FileLoaderSection(object):
```
deiferni: I'd also test the section separately, not only in our pipeline blackbox tests.

opengever/setup/sections/fileloader.py
```diff
@@ -0,0 +1,109 @@
...
+        # configurable instead of hard coding it here.
+        self.key = 'filepath'
+        self.pathkey = defaultMatcher(options, 'path-key', name, 'path')
+
+        self.bundle_path = IAnnotations(transmogrifier)[BUNDLE_PATH_KEY]
+
+        # TODO: Handle .eml / ftw.mail.mail
+        self.file_field_class = IDocumentSchema['file']._type
+
+        self.stats = IAnnotations(transmogrifier)[JSON_STATS_KEY]
```
deiferni: i'd also add a short test-case to the sections tests for correct stats-generation

opengever/setup/sections/fileloader.py
lukasgraf: D'oh 🙈

opengever/setup/sections/fileloader.py
lukasgraf: 👍 Done.

opengever/setup/sections/fileloader.py
lukasgraf: 👍 Done

opengever/setup/sections/fileloader.py
lukasgraf: I agree. Until we actually have implemented this as a stand-alone validator, I'd prefer to leave it here, just so we can keep track of edge cases and validations we still need to do.

opengever/setup/sections/fileloader.py
lukasgraf: Yeah. `traverse()` is a transmogrifier utility function though that traverses to objects without being fooled by acquisition or views.

opengever/setup/sections/fileloader.py
lukasgraf: I agree. In the beginning to I think it's useful to keep track of the things that could happen, and Should Not Ever Happen [tm], and then later ramp up the strictness once we have a pre-validation stage in place, and maybe also introduce a strict mode.

lukasgraf:
@deiferni / @phgross updated

lukasgraf:
@deiferni replaced `traverse()` with `item['_object']` as discussed

PR #2406, opened 2016-12-09
Fix flaky test depending on journal entry sort order

lukasgraf:
Fix flaky test: Journal entry sort order was fixed in fb18056f, so that journal entries are now sorted in chronological order. In combination with the timestamp used in this test (`09.12.2016 09:40`) and another journal entry (we don't care about) using the non-frozen, dynamic `datetime.today()`, this resulted in the sort order changing today (2016-12-09), and the test started to fail.

@phgross


phgross:
thx!

PR #2411, opened 2016-12-13
Completely remove digitally_available, add sequence_number to jsonschema.

### LINKED ISSUE #2365: Attribute ergänzen/entfernen
JSON-Spez aktualisieren

- digitally_available für Dokumente entfernen
- Laufnummern (sequence numbers) hinzufügen

deiferni:
Completely remove `digitally_available` from all schema dumps. That field is de facto a property and is always set automatically depending on file availability. I think it is reasonable to also drop it from the schema dumps for the documentation.

Also add `sequence_number` to JSON-Schema for dossiers and documents.

Closes #2365

### DIFF
--- a/opengever/base/schemadump/schema.py
+++ b/opengever/base/schemadump/schema.py
@@ -327,8 +327,22 @@ class OGGBundleJSONSchemaBuilder(object):
 
         if portal_type == 'opengever.document.document':
             core_schema['properties']['filepath'] = {'type': 'string'}
+            core_schema['properties']['sequence_number'] = {
+                'type': 'integer',
+                'title': 'Laufnummer',
+                'description': u'Fortlaufend gez\xe4hlte Nummer eines '
+                               'Dokumentes.'
+            }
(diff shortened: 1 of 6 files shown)


opengever/base/schemadump/schema.py
```diff
@@ -327,9 +327,23 @@ def build_schema(self, portal_type):
 
         if portal_type == 'opengever.document.document':
             core_schema['properties']['filepath'] = {'type': 'string'}
+            core_schema['properties']['sequence_number'] = {
```
lukasgraf: I'm wondering whether `sequence_number` shouldn't be injected at the `JSONSchemaDumper` level instead, so we also have it in our API docs (and `field_order`)

deiferni: I'm not sure if we want to expose it there yet, afaik the API could not handle setting a custom `sequence_number`, for example. If you agree i'd suggest to leave it only in the schema for now.

PR #2418, opened 2016-12-14
Inlcude string length contraints in schema dumps

### LINKED ISSUE #2364: Include String-Length constraints in schema dumps
Certain String fields have length-constraints. These need to be included in Schema dumps.

deiferni:
This PR includes the string length constraints of all fields in schema dumps.

Closes #2364.

### DIFF
--- a/docs/public/api/schemas/ftw.mail.mail.inc
+++ b/docs/public/api/schemas/ftw.mail.mail.inc
@@ -134,16 +134,6 @@
        
 
 
-   .. py:attribute:: digitally_available
-
-       :Feldname: :field-title:`Digital verfügbar`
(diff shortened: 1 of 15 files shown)


deiferni:
Updated
- i dropped the task from the exported types again but left the overrides etc. in-place in case we want to re-enable the update
- also the public documentation will be updated with this `PR` since we now omit the field `digitally_available`

lukasgraf:
@deiferni `opengever.task.task.schema.json` is somehow still part of the diff

deiferni:
@lukasgraf oh i see, they are not deleted automatically - lets try again 🙈

PR #2427, opened 2016-12-14
Patch CMFCatalogAware object to improve performance.

elioschmutz:
Patch the Products.CMFCore.CMFCatalogAware indexObject, reindexObject
and unindexObject methods.

This patch is deactivated by default and can be activated through
the `DeactivateCatalogIndexing` context manager:

```
>>> with DeactivateCatalogIndexing():
...     object.reindexObject  # Does nothing
...     object.unindexObject  # Does nothing
...     object.indexObject  # Does nothing
```
If the patch is activated, it skips the catalog index-methods.

What's the motivation behind this patch?

While creating an object, the object will be reindexed up to 4 times.
This behavior takes a lot of time and is not performance-friendly. Creating
one object through the web might be not a performance issue. But in several
parts of the GEVER-system we're creating a lot of content programatically
(i.e. dossiertemplates).

To improve the performence in such case, you can decativate the index-methods

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.14.1 (unreleased)
 -------------------
 
+- Creates successor-tasks of forwardings with a default task type.
+  [phgross]
+
 - Display former contact id in the byline of contacts.
   [phgross]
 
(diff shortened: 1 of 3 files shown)


elioschmutz:
This PR was automatically closed after rebasing the branch...

see new branch at #2428

PR #2428, opened 2016-12-15
Patch CMFCatalogAware object to improve performance.

elioschmutz:
Patch the Products.CMFCore.CMFCatalogAware indexObject, reindexObject
and unindexObject methods.

This patch is deactivated by default and can be activated through
the `DeactivateCatalogIndexing` context manager:

```
>>> with DeactivateCatalogIndexing():
...     object.reindexObject  # Does nothing
...     object.unindexObject  # Does nothing
...     object.indexObject  # Does nothing
```
If the patch is activated, it skips the catalog index-methods.

What's the motivation behind this patch?

While creating an object, the object will be reindexed up to 4 times.
This behavior takes a lot of time and is not performance-friendly. Creating
one object through the web might be not a performance issue. But in several
parts of the GEVER-system we're creating a lot of content programatically
(i.e. dossiertemplates).

To improve the performence in such case, you can decativate the index-methods

see #2143

### DIFF
--- a/opengever/base/monkey/patches/cmf_catalog_aware.py
+++ b/opengever/base/monkey/patches/cmf_catalog_aware.py
@@ -1,0 +1,86 @@
+from opengever.base.monkey.patching import MonkeyPatch
+from zope.globalrequest import getRequest
+from zope.interface import alsoProvides
+from zope.interface import Interface
+from zope.interface import noLongerProvides
+
+
+class IDisableCatalogIndexing(Interface):
+    """Marker-interface to disable the catalog
+    indexing functions.
+
+    If this interface is provided by the request, all the
+    catalog-index-methods will be disabled.
+
+    Use the DeactivatedCatalogIndexing contextmanager to
+    get in use of this functionality.
+    """
+
+
+class DeactivatedCatalogIndexing(object):
+    """Contextmanager: Deactivates catalog-indexing
+    """
+    def __enter__(self):
+        alsoProvides(getRequest(), IDisableCatalogIndexing)
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        noLongerProvides(getRequest(), IDisableCatalogIndexing)
+
+
+class PatchCMFCatalogAware(MonkeyPatch):
+    """Patch the Products.CMFCore.CMFCatalogAware indexObject, reindexObject
+    and unindexObject methods.
+
+    This patch is deactivated by default and can be activated through
+    the DeactivatedCatalogIndexing context manager:
+
+    >>> with DeactivatedCatalogIndexing():
+    ...     object.reindexObject  # Does nothing
+    ...     object.unindexObject  # Does nothing
+    ...     object.indexObject  # Does nothing
+
+    If the patch is activated, it skips the catalog index-methods.
+
+    What's the motivation behind this patch?
+
+    While creating an object, the object will be reindexed up to 4 times.
+    This behavior takes a lot of time and is not performance-friendly. Creating
+    one object through the web might be not a performance issue. But in several
+    parts of the GEVER-system we're creating a lot of content programatically
+    (i.e. dossiertemplates).
+
+    To improve the performance in that case, you can decativate the index-methods
+    and do it manually at the end of your tasks.
+    """
+
+    def __call__(self):
+
+        def _is_indexing_disabled():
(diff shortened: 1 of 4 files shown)


opengever/base/monkey/patches/cmf_catalog_aware.py
```diff
@@ -0,0 +1,88 @@
...
+
+class IDisableCatalogIndexing(Interface):
+    """Marker-interface to disable the catalog
+    indexing functions.
+
+    If this interface is provided by the request, all the
+    catalog-index-methods will be disabled.
+
+    Use the DeactivateCatalogIndexing contextmanager to
+    get in use of this functionality.
```
deiferni: ~~get in~~ use ~~of~~ this functionality.

opengever/base/monkey/patches/cmf_catalog_aware.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+
+
+class DeactivateCatalogIndexing(object):
```
deiferni: i'd move this out of monkey patches, it can be reused

deiferni: also i'd rename to Deactivate**d**CatalogIndexing, this reads better: `with DeactivateCatalogIndexing()`:

opengever/base/monkey/patches/cmf_catalog_aware.py
```diff
@@ -0,0 +1,88 @@
...
+
+    What's the motivation behind this patch?
+
+    While creating an object, the object will be reindexed up to 4 times.
+    This behavior takes a lot of time and is not performance-friendly. Creating
+    one object through the web might be not a performance issue. But in several
+    parts of the GEVER-system we're creating a lot of content programatically
+    (i.e. dossiertemplates).
+
+    To improve the performence in such case, you can decativate the index-methods
```
deiferni: in that case/ in such a case

opengever/base/monkey/patches/cmf_catalog_aware.py
```diff
@@ -0,0 +1,88 @@
...
+    def __call__(self):
+
+        def _is_indexing_disabled():
+            return IDisableCatalogIndexing.providedBy(getRequest())
+
+        def indexObject(self):
+            if _is_indexing_disabled():
+                # do nothing if indexing is disabled
+                return
+
```
lukasgraf: These empty lines somehow bug me - can you please ✂️  them? 😉

opengever/base/monkey/patches/cmf_catalog_aware.py
lukasgraf: performence -> performance

opengever/base/monkey/patches/cmf_catalog_aware.py
```diff
@@ -0,0 +1,88 @@
...
+            return original_reindexObject(self, idxs)
+
+        from Products.CMFCore.CMFCatalogAware import CMFCatalogAware
+        original_indexObject = CMFCatalogAware.indexObject
+        original_unindexObject = CMFCatalogAware.unindexObject
+        original_reindexObject = CMFCatalogAware.reindexObject
```
lukasgraf: When keeping references to the original methods, you should also set `__patch_refs__ = False` in the the referencing scope - otherwise, `patch_refs` might also patch *those* references 😉  - see for example https://github.com/4teamwork/opengever.core/blob/master/opengever/base/monkey/patches/default_values.py#L243

opengever/base/monkey/patches/cmf_catalog_aware.py
elioschmutz: @deiferni

lukasgraf: @elioschmutz I'd actually leave this in here. It's still resuable even if it's in `opengever/base/monkey/patches/cmf_catalog_aware.py`, and that way one is more aware of the fact that this is feature based on a monkey patch when using it.

elioschmutz:
@phgross @lukasgraf @deiferni 👍  updated your requested changes. Please take another look.

lukasgraf:
@elioschmutz looks good, unfortunately you'll need to rebase again because a release happened in the meantime :trollface:

elioschmutz:
👍

elioschmutz:
@deiferni

PR #2434, opened 2016-12-16
Implement reusable loader for OGGBundles

lukasgraf:
This moves the logic for loading OGGBundles and yielding items from them into a reusable component `BundleLoader`.

As of now, responsibilities for the new loader are:
- Locating and reading JSON files from bundles on the filesystem
- Validating the JSON files against their respective JSON schemas
- Injecting the items' appropriate `portal_type` as `item['_type']`
- Returning items in order of *type* (repo roots before folders, folders before dossiers, etc..). Proper order in terms of "parents before children" is still handled by the `ResolveGUIDSection` for now.

@phgross


opengever/bundle/tests/helpers.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+def get_title(item):
+    title = item.get('title_de')
+    if not title:
+        title = item['title']
+    return title
+
+
+def get_pt(item):
```
phgross: I would rename this method, its hard to figure out, what `get_pt` does.

opengever/bundle/tests/test_bundle_loader.py
```diff
@@ -0,0 +1,45 @@
...
+             ('repositoryfolder', u'Organisation'),
+             ('businesscasedossier', u'Dossier Vreni Meier'),
+             ('businesscasedossier', u'Hanspeter M\xfcller'),
+             ('document', u'Bewerbung Hanspeter M\xfcller'),
+             ('document', u'Entlassung Hanspeter M\xfcller')],
+            [(get_pt(i), get_title(i)) for i in items])
+
+    def test_inserts_portal_type(self):
+        items = self.load_bundle()
+        self.assertEqual([
```
phgross: Extending the assets with a mail, would be nice test.

lukasgraf:
@phgross updated

PR #2439, opened 2016-12-20
Extend OGGBundle fileloader to handle mails correctly.

lukasgraf:
Extends the fileloader in the OGGBundle pipeline to properly deal with `*.eml` mails, and adds some additional tests.

@deiferni

### DIFF
--- a/opengever/setup/sections/fileloader.py
+++ b/opengever/setup/sections/fileloader.py
@@ -1,6 +1,7 @@
 from collective.transmogrifier.interfaces import ISection
 from collective.transmogrifier.interfaces import ISectionBlueprint
 from collective.transmogrifier.utils import defaultMatcher
+from ftw.mail.mail import IMail
 from mimetypes import guess_type
 from opengever.document.document import IDocumentSchema
 from opengever.document.subscribers import set_digitally_available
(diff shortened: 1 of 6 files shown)


opengever/setup/tests/test_section_constructor.py
```diff
@@ -16,6 +16,8 @@ def setUp(self):
 
         lang_tool = api.portal.get_tool('portal_languages')
         lang_tool.setDefaultLanguage('de-ch')
+        # Required to create mails, but not dossiers - no idea why
+        self.grant('Manager')
```
deiferni: Hmm that may have something to do with the mail not being created inside a dossier, i'd investigate in that direction.

lukasgraf: Already tried that, doesn't change a thing :-/

opengever/setup/sections/fileloader.py
```diff
@@ -88,12 +90,17 @@ def __iter__(self):
                     yield item
```
lukasgraf: 👍  - I will change this to
1) Determine the type (doc or mail) *once* at the beginning of the loop and
2) based on that, assign a `file_field` once and use that to `field(obj).set(value)` (assuming that works)

lukasgraf:
@deiferni updated

PR #2446, opened 2016-12-22
Install and configure plone.app.caching.

jone:
Install plone.app.caching and configure it with gzip compression.

This adds cachec-control-headers to most requests (css, js, images).
Primary documents are not cached.

-

Without p.a.caching, resources such as CSS and JS used be served from the browser cache, but without an explicit cache-control-header.
Therefore this change does not have a big impact for the user, but we now have a standard tool for implementing better caches and we serve resources with cache-control-headers.


lukasgraf:
Looks good to me! 👍

PR #2451, opened 2016-12-22
🚧 Implement UserTurboIndex.

jone:
🚧 _experimental_
When sorting a tabbed view by a user column, we had to sort the complete result set using the brain metadata in order to have it sorted by fullname rather than userid.
The problem is that this breaks lazyness and triggers brain builds for all items in the result set, leading to performance problems.

In order to fix that, we are adding a new UserTurboIndex, currently replacing the "Creator" and "responsible" indexes (proof of concept).

This new index allows to sort by another value than the indexed value, by providing a "documentToKeyMap".
The catalog does only use that though when the "limit" in the sort method is not None, we therefore have to patch this argument.


jone:
If I remember correctly, this change did not have a significant impact on the speed. 
Therefore I don't think we will do that now.

lukasgraf:
@jone I recently stumbled across this one again, and was thinking whether or not we should revive it.

In general I think post-sorting in Python that destroys lazyness should be avoided - but if you're saying performance improvements weren't that significant, then we should probably let the dead rest 😉  I'll dig it up later if necessary, so 👍  on not deleting the branch quite yet.

PR #2454, opened 2016-12-23
Remove ftw.footer, replace with static footer viewlet.

jone:
⚠️  the following PRs should be merged first
Styles: https://github.com/4teamwork/plonetheme.teamraum/pull/495
Required for upgrades: https://github.com/4teamwork/ftw.upgrade/pull/123

Replace ftw.footer because portlets are known to be slow and a simple
viewlet is easier to manage.

- 4teamwork address was changed
- opengever.core version is now shown

**Before**
bildschirmfoto 2016-12-23 um 12 34 00

**After**
screen shot 2017-01-26 at 14 50 25


opengever/base/viewlets/configure.zcml
```diff
@@ -79,4 +79,12 @@
       layer="opengever.base.interfaces.IOpengeverBaseLayer"
       />
 
+  <browser:viewlet
+      name="opengever.base.footer"
+      manager="plone.app.layout.viewlets.interfaces.IPortalFooter"
+      class=".footer.FooterViewlet"
+      permission="zope2.View"
+      layer="opengever.base.interfaces.IOpengeverBaseLayer"
```
deiferni: this is a new viewlet, could we omit the layer to allow policies to easily overwrite the layer? Currently only some on-site policies define their own browser-layer. Or would you follow another strategy to overwrite the layer in policies?

jone: @deiferni it is best practice to always set a layer for viewlets.
The reason for that is: viewlets are automatically "enabled" when registered, which lead to broken layouts on vanilla Plone sites when the ZCML of packages are loaded which provide viewlets without a layer.

But since it is not likely that we have non-GEVER sites in an installation which has `opengever.core` in the path we can cease applying that best practice in opengever.core.

I can remove the layer 😉

jone: @deiferni I've removed the layer.

deiferni:
I'll hijack this branch as it requires some coordination, we'll have to adapt the following policies as well:

- [ ] opengever.bs
- [ ] opengever.ai
- [ ] opengever.sg
- [ ] opengever.zug
- [ ] dev (4tw)
- [ ] test (4tw)
- [ ] nightly (4tw)
- [ ] lab (4tw)
- [ ] opengever.hpcdual
- [ ] opengever.nidau
- [ ] opengever.phvs
- [ ] opengever.phbern
- [ ] opengever.edk
- [ ] opengever.kgkoeniz
- [ ] opengever.duerrenaesch
- [ ] opengever.htwchur
- [ ] opengever.bistumbasel
- [ ] opengever.liesberg
- [ ] opengever.mumpf
- [ ] opengever.framsteg
- [ ] opengever.ftw

deiferni:
@jone i've investigated, the policies are still installable, even when `portlets.xml` contains invalid portlet assignments. I'd suggest to merge this PR and the update the policies successively. Since i have now amended some more commits i'd be glad if you could do a quick cross-review.

sources.cfg
```diff
@@ -9,6 +9,7 @@ development-packages =
   plone.restapi
   plone.rest
   plonetheme.teamraum
+  ftw.upgrade
```
jone: I have merged https://github.com/4teamwork/ftw.upgrade/pull/123
I suggest to release `ftw.upgrade` and add a minimum version requirement to `setup.py`, because the upgrade will not work when the method is missing.

jone:
I think it's good 👍 
Maybe CSS needs to be adjusted. Do you want to add that to the theme?

deiferni:
@jone yes i've already updated the CSS, see linked PR above 😉 , also i noticed that the screenshots were no longer up to date, now there is a new after screenshot that has proper styling.

jone:
ah, nice 👍

deiferni:
also i have tested the changes with a policy that has a custom footer, seems to work as expected 👍

Issue #2345, opened 2016-11-30
Disposition overview improvements


phgross:
- [x] group dossier listing by repositoryfolder

more Informations: https://basecamp.com/2768704/projects/12554551/todos/274414343

PR #2456, opened 2017-01-04
Disposition overview: display dossiers grouped by repositoryfolder

phgross:
Closes one part of #2345 
bildschirmfoto 2017-01-04 um 14 27 23

The PR includes an upgradestep which inserts the missing repository title to the dossier mapping on closed dispositions. It may be possible that some repository titles are be renamed since the disposition closing and the repository title is not really correct, but because the disposition feature is not used in production yet, thats not a problem.

Styling in a separate PR:


opengever/disposition/browser/overview.py
```diff
@@ -13,8 +14,17 @@ class DispositionOverview(BrowserView, GeverTabMixin):
 
     show_searchform = False
 
-    def get_dossiers(self):
-        return self.context.get_dossier_representations()
+    def get_dossiers_grouped_by_repository(self):
+        dossiers = self.context.get_dossier_representations()
+        grouped_dossiers = {}
+        for dossier in dossiers:
+            repo = dossier.repo_grouper()
```
deiferni: i would call this `key` and the method `get_grouping_key` or something like that

opengever/disposition/disposition.py
```diff
@@ -81,6 +92,12 @@ def __init__(self, dossier_mapping, disposition):
     def additional_metadata_available(self):
         return False
 
+    def repo_grouper(self):
+        return self.repository_title
+
+    def get_repository_title(self):
+        return None
```
deiferni: Couldn't this also return `self.repository_title`?

phgross:
@deiferni updated

deiferni:
_(i just rebased and force-pushed)_

PR #2457, opened 2017-01-04
Make transfer number only editable for archivists and managers.

### LINKED ISSUE #2343: Field transfer_number should onyl be editable by archivists

phgross:
Protect the field with a separate permission "Edit transfer number".

Closes #2343


opengever/disposition/upgrades/20170104162045_introduce_new_permission_edit_transfer_number/rolemap.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
+
+<rolemap>
```
deiferni: I think you don't need to redefine the whole `rolemap.xml`, but just use the following:

```xml
<rolemap>
  <permissions>
    <permission name="opengever.disposition: Edit transfer number" acquire="True">
      <role name="Manager"/>
    </permission>
  </permissions>
</rolemap>
```

phgross: @deiferni you're right, updated ...

phgross:
@deiferni updated

PR #2461, opened 2017-01-11
Hide dossiers offer and archive transition from the action menu.

### LINKED ISSUE #2423: Offer transition is available in the Dossier UI
The newly introduced transition `dossier-transition-offer` is visible for users with the role `Records Manager` in the UI. This should not the case, because the transition should only done during the disposition creation.

bildschirmfoto 2016-12-14 um 17 03 10

phgross:
This two transitions are only done via creating or archiving a disposition, but should not be visible in the UI for a user.

Fixes #2423


phgross:
@deiferni added tests ...

PR #2464, opened 2017-01-12
Drop security declaration for nonexistent method 'schedule'.

deiferni:
This gets rid of the last `WARNING` during instance bootup 🎉 .


lukasgraf:
Nice! 👍 🎉

PR #2472, opened 2017-01-16
Adjust dossiers retention expiration calculation.

### LINKED ISSUE #2425: Change retention period expire calculation
> Die Berechnung der Aufbewahrungsfrist hat - nach Rücksprache mit Zug - nach Geschäftsjahr zu erfolgen. Beispiel:
> - Dossierabschluss 03.01.2013 --> 10 Jahre Aufbewahrung
> --> Dossier kann ab 01.01.2024 ausgesondert werden.

> Im Moment des Angebots (also z.B. am 02.05.2024) würde dann das Feld "Anbietezeitpunkt" befüllt.

Further Information: https://basecamp.com/2768704/projects/12554551/todos/286197285

phgross:
The retention period of a dossier should expire at the beginning of the year after the retention period, see https://basecamp.com/2768704/projects/12554551/todos/286197285. This PR changes the calculation and reindex the expiration date for all resolved dossiers. Closes #2425.

Note: This PR also disables the calculation of the retention_expiration date for all resolved dossiers in a earlier upgradestep. This makes sense for all deployments which has not installed the newest GEVER release yet otherwise the expiration date would be calculate twice. For deployments where the retention_expiration is already installed the disabled calculation has no impact.

### DIFF
--- a/opengever/dossier/base.py
+++ b/opengever/dossier/base.py
@@ -2,7 +2,6 @@ from Acquisition import aq_inner
 from Acquisition import aq_parent
 from datetime import date
 from datetime import datetime
-from dateutil.relativedelta import relativedelta
 from five import grok
 from opengever.base.behaviors.lifecycle import ILifeCycle
 from opengever.base.interfaces import IReferenceNumber
@@ -49,6 +48,10 @@ DOSSIER_STATES_OFFERABLE = DOSSIER_STATES_CLOSED + ['dossier-state-offered']
 
 
 _marker = object()
+
+
+def is_january_the_first(value):
+    return value.month == 1 and value.day == 1
 
 
 class DossierContainer(Container):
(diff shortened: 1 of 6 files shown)


opengever/dossier/base.py
```diff
@@ -312,9 +311,16 @@ def get_reference_number(self):
         return IReferenceNumber(self).get_number()
 
     def get_retention_expiration_date(self):
+        """Returns the date when the expiration date expires:
+
+        The start of the next year (first january) after the retention period.
```
deiferni: the first of january

opengever/dossier/base.py
```diff
@@ -312,9 +311,16 @@ def get_reference_number(self):
...
+        """
         if IDossier(self).end:
-            return IDossier(self).end + relativedelta(
-                years=ILifeCycle(self).retention_period)
+            year = IDossier(self).end.year + ILifeCycle(self).retention_period
```
deiferni: `end` is not required so we should expect that it may not be set and handle that also

opengever/dossier/base.py
```diff
@@ -312,9 +311,16 @@ def get_reference_number(self):
...
+            if IDossier(self).end == date(IDossier(self).end.year, 1, 1):
```
deiferni: - use helper function `is_january_the_first` or similar
- this line was quite confusing to read, comparing `month` and `day` only would explain what you're after
- i'd also mention in the docstring that end-date includes jan. the first

phgross:
@deiferni updated ...

PR #2476, opened 2017-01-17
Include REST API in mainline GEVER

lukasgraf:
Includes the REST API in mainline GEVER.

This adds Python dependencies on `plone.restapi` and `plone.rest`, and installs the `plone.restapi:default` profile by default.

- Introduces a new role `APIUser`
- By default, only the `Manager` and `APIUser` roles have the `plone.restapi: Use REST API` permission
- The `APIUser` role isn't assigned to any users or groups by default

@deiferni @Rotonen

/cc @jone

Closes https://github.com/4teamwork/opengever.core/issues/2474


opengever/policy/base/upgrades/20170117132929_disallow_api_use_for_everyone_except_manager/upgrade.py
```diff
@@ -0,0 +1,9 @@
+from ftw.upgrade import UpgradeStep
+
+
+class DisallowAPIUseForEveryoneExceptManager(UpgradeStep):
+    """Disallow API use for everyone except Manager.
```
jone: + `and APIUser` ?

lukasgraf: Technically, this change / upgrade step really does disable it for everyone except `Manager`. One commit / upgrade step later, I introduce the new role `APIUser` that enables it again for users that have that role.

lukasgraf:
@deiferni damn, missed a test still using that layer, hence the failing tests - updated.

opengever/examplecontent/profiles/default/metadata.xml
```diff
@@ -1,7 +1,5 @@
 <metadata>
   <dependencies>
-    <dependency>profile-opengever.policy.base:default</dependency>
```
deiferni: Wouldn't do that yet, though theoretically correct that creates an examplecontent profile which is different from our SaaS profile, i created an issue for that problem: https://github.com/4teamwork/opengever.core/issues/2478

opengever/examplecontent/profiles/default/metadata.xml
```diff
@@ -1,7 +1,5 @@
...
     <dependency>profile-plonetheme.teamraum:gever</dependency>
-    <dependency>profile-opengever.policy.base:mimetype</dependency>
```
deiferni: same here, i'd leave that in for now

setup.py
```diff
@@ -121,6 +119,8 @@
         'plone.principalsource',
         'plone.protect >= 3.0.15',
         'plone.registry',
+        'plone.rest',
```
deiferni: not necessary, this is an indirect dependency only (`plone.restapi` already pulls that in)

Rotonen: We are also building API endpoints on top of naked plone.rest which will break if plone.rest would not get pulled in (just in case plone.restapi is ever replaced in the future).

lukasgraf:
@deiferni updated (dropped commit by @Rotonen that fixes the twice-installed policy profiles)

lukasgraf:
@deiferni rebased

PR #2481, opened 2017-01-19
OGGBundle constructor: Support setting local reference number part for dossiers (when provided by JSON item).

lukasgraf:
OGGBundle constructor: Support setting local reference number part for dossiers (when provided by JSON item).

If (for dossiers) a key `reference_number` is provided, this will be used to prevent automatically issuing the next free reference number (local part), and set the given local part value instead.

I also updated some existing tests so that we test the setting of the local parts (called "reference_prefixes") for repository folders (which already worked) with non-sequential numbers.

TODO (separately): Validate uniqueness of reference number local parts on their respective levels.

@deiferni

### DIFF
--- a/opengever/setup/sections/constructor.py
+++ b/opengever/setup/sections/constructor.py
@@ -2,10 +2,16 @@ from collective.transmogrifier.interfaces import ISection
 from collective.transmogrifier.interfaces import ISectionBlueprint
 from opengever.base.behaviors.translated_title import ITranslatedTitle
 from opengever.base.behaviors.translated_title import TRANSLATED_TITLE_NAMES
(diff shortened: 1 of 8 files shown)


opengever/setup/sections/constructor.py
```diff
@@ -56,8 +74,24 @@ def __iter__(self):
 
                 title_args = dict((key, item.get(key)) for key in title_keys)
 
-                obj = createContentInContainer(
-                    context, portal_type, **title_args)
+                with NoDossierReferenceNumbersIssued():
+                    # Create the object without automatically issuing a
+                    # reference number - we might want to set it explicitely
```
deiferni: typo: explicitly

PR #2482, opened 2017-01-19
OGGBundle loader: Strip extension for title if necessary.

lukasgraf:
⚠️ Builds upon #2481 - merge that one first ⚠️  

Removes any file extension from the title if necessary, so it doesn't end up in the final title.

### DIFF
--- a/opengever/bundle/loader.py
+++ b/opengever/bundle/loader.py
@@ -53,7 +53,22 @@ class BundleLoader(object):
             self.validate_schema(items, json_name)
             for item in items:
                 item['_type'] = self.determine_portal_type(json_name, item)
+                if json_name == 'documents.json':
+                    self.strip_extension_from_title(item)
                 self.items.append(item)
+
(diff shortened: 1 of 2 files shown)


lukasgraf:
@deiferni rebased

opengever/bundle/loader.py
```diff
@@ -53,8 +53,23 @@ def load_items(self):
             self.validate_schema(items, json_name)
             for item in items:
                 item['_type'] = self.determine_portal_type(json_name, item)
+                if json_name == 'documents.json':
```
deiferni: i'm sorry to ask such a fundamental question, but is this really the bundle-loaders responsibility? I'd argue that the correct title must be present in the bundle's data since we already have a separate `title` and `filepath` field.

lukasgraf: As discussed: Leave this as-is for now, since not having a file extension in the title is the default behavior in GEVER.

PR #2494, opened 2017-01-26
Extend policy template

jone:
- add bumblebee configuration when bumblebee feature is selected
- ask for deployment number, used in deployment file name and deployment number configuration


deiferni:
Thanks 👍

PR #2496, opened 2017-01-27
Officeconnector url scheme, attach to outlook

Rotonen:
* A new subproduct to handle generating Officeconnector URLs
  * Uses BrowserView traversals for feature availability on templates
  * Uses plone.rest for issuing URLs
  * Uses javascript on views to fetch a newstyle URL and then pass that onto the OS
* Checkout functionality is only equivalent to a dummy stub at this point

/cc @bierik for the javascript

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,14 @@ Changelog
 4.14.2 (unreleased)
 -------------------
 
+- Modified opengever.document to conditionally use the new Office Connector
+  functionality.
+  [Rotonen]
+
+- Added the preliminary release of the opengever.officeconnector subproduct,
+  which only provides the attach-to-outlook functionality for now.
+  [Rotonen]
+
 - Change pre appraisal handling and allow appraisal per repository.
   [phgross]
 
--- a/opengever/document/locales/de/LC_MESSAGES/opengever.document.po
+++ b/opengever/document/locales/de/LC_MESSAGES/opengever.document.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
-"POT-Creation-Date: 2016-11-14 11:29+0000\n"
+"POT-Creation-Date: 2017-01-30 15:48+0000\n"
 "PO-Revision-Date: 2013-03-27 10:44+0100\n"
 "Last-Translator: Jonas Baumann <j.baumann@4teamwork.ch>\n"
 "Language-Team: 4teamwork <info@4teamwork.ch>\n"
@@ -72,7 +72,7 @@ msgstr "Es ist nicht erlaubt, das Dokument ${title} auszuchecken."
 msgid "Could not check out object: ${title}, it is not a document"
 msgstr "${title} ist kein Dokument und konnte deshalb nicht ausgecheckt werden."
 
-#: ./opengever/document/browser/report.py:96
+#: ./opengever/document/browser/report.py:81
 msgid "Could not generate the report"
 msgstr "Report konnte nicht generiert werden"
 
@@ -80,7 +80,7 @@ msgstr "Report konnte nicht generiert werden"
 msgid "Create Task"
 msgstr "Aufgabe erstellen"
 
-#: ./opengever/document/browser/overview.py:200
+#: ./opengever/document/browser/overview.py:208
 msgid "Current version: ${version}"
 msgstr "Aktuelle Version: ${version}"
 
@@ -116,11 +116,11 @@ msgstr "Als E-Mail versenden"
 msgid "Submit additional documents"
 msgstr "Zusätzliche Anhänge einreichen"
 
-#: ./opengever/document/browser/overview.py:196
+#: ./opengever/document/browser/overview.py:204
 msgid "Submitted version: ${version}"
 msgstr "Eingereichte Version: ${version}"
 
-#: ./opengever/document/browser/overview.py:171
+#: ./opengever/document/browser/overview.py:174
 msgid "Submitted with"
 msgstr "Eingereicht bei"
 
@@ -178,34 +178,34 @@ msgid "description_download_confirmation"
 msgstr "Sie sind dabei, eine Kopie des Dokuments ${filename} herunterzuladen."
 
 #. Default: "You don't select a file and document is also not preserved in paper_form, please correct it."
-#: ./opengever/document/behaviors/metadata.py:182
+#: ./opengever/document/behaviors/metadata.py:194
 msgid "error_file_and_preserved_as_paper"
 msgstr "Sie haben weder eine Datei ausgewählt noch ist das Dokument in Papierform aufbewahrt, bitte korrigieren."
 
 #. Default: "It's not possible to add E-mails here, please '                    'send it to ${mailaddress} or drag it to the dossier '                    ' (Dragn'n'Drop)."
-#: ./opengever/document/document.py:99
+#: ./opengever/document/document.py:96
 msgid "error_mail_upload"
 msgstr "Es ist nicht erlaubt hier E-Mails anzufügen. Bitte senden Sie das E-Mail an die Addresse ${mailaddress} oder ziehen Sie es in das Dossier (Drag'n'Drop)."
 
 #. Default: "You have not selected any Items"
-#: ./opengever/document/browser/report.py:80
+#: ./opengever/document/browser/report.py:65
 msgid "error_no_items"
 msgstr "Sie haben keine Objekte ausgewählt."
 
 #. Default: "Either the title or the file is required."
-#: ./opengever/document/document.py:76
+#: ./opengever/document/document.py:73
 msgid "error_title_or_file_required"
 msgstr "Ein Titel oder eine Datei muss angegeben werden."
 
 #. Default: "Archive file"
-#: ./opengever/document/behaviors/metadata.py:46
+#: ./opengever/document/behaviors/metadata.py:58
 msgid "fieldset_archive_file"
 msgstr "Archiv Datei"
 
 #. Default: "Common"
 #: ./opengever/document/behaviors/metadata.py:39
 #: ./opengever/document/behaviors/related_docs.py:39
-#: ./opengever/document/document.py:52
+#: ./opengever/document/document.py:49
 msgid "fieldset_common"
 msgstr "Allgemein"
 
@@ -230,12 +230,12 @@ msgid "heading_checkin_comment_form"
 msgstr "Dokumente einchecken"
 
 #. Default: ""
-#: ./opengever/document/behaviors/metadata.py:137
+#: ./opengever/document/behaviors/metadata.py:144
 msgid "help_archival_file"
 msgstr "Archivtaugliche Version der Originaldatei"
 
 #. Default: "Surname firstname or a userid(would be automatically resolved to fullname)"
-#: ./opengever/document/behaviors/metadata.py:115
+#: ./opengever/document/behaviors/metadata.py:120
 msgid "help_author"
 msgstr "Nachname Vorname oder ein Benutzerkürzel (wird automatisch nach Nachname Vorname aufgelöst)."
 
@@ -244,35 +244,35 @@ msgstr "Nachname Vorname oder ein Benutzerkürzel (wird automatisch nach Nachnam
 msgid "help_checkin_journal_comment"
 msgstr "Was wurde am Dokument geändert?"
 
-#: ./opengever/document/behaviors/metadata.py:100
+#: ./opengever/document/behaviors/metadata.py:105
 msgid "help_delivery_date"
 msgstr "Datum, an dem das Dokument über den Korrespondenzweg versandt worden ist"
 
-#: ./opengever/document/behaviors/metadata.py:83
+#: ./opengever/document/behaviors/metadata.py:88
 msgid "help_document_date"
 msgstr "Datum des Dokuments"
 
 #. Default: ""
-#: ./opengever/document/document.py:69
+#: ./opengever/document/document.py:66
 msgid "help_file"
 msgstr "Datei, die zu einem Dossier hinzugefügt wird"
 
 #. Default: ""
-#: ./opengever/document/behaviors/metadata.py:75
+#: ./opengever/document/behaviors/metadata.py:80
 msgid "help_foreign_reference"
 msgstr "Referenz auf das entsprechende Dossier des Absenders"
 
 #. Default: ""
-#: ./opengever/document/behaviors/metadata.py:130
+#: ./opengever/document/behaviors/metadata.py:135
 msgid "help_preserved_as_paper"
 msgstr "In Papierform aufbewahrt"
 
 #. Default: ""
-#: ./opengever/document/behaviors/metadata.py:156
+#: ./opengever/document/behaviors/metadata.py:163
 msgid "help_preview"
 msgstr "Online Voransicht des Originaldokuments"
 
-#: ./opengever/document/behaviors/metadata.py:92
+#: ./opengever/document/behaviors/metadata.py:97
 msgid "help_receipt_date"
 msgstr "Datum, an dem das Dokument über den Korrespondenzweg angekommen ist"
 
@@ -287,19 +287,25 @@ msgid "label_actor"
 msgstr "Geändert von"
 
 #. Default: "Archival File"
-#: ./opengever/document/behaviors/metadata.py:136
-#: ./opengever/document/browser/overview.py:166
+#: ./opengever/document/behaviors/metadata.py:143
+#: ./opengever/document/browser/overview.py:169
 msgid "label_archival_file"
 msgstr "Archivdatei"
 
 #. Default: "Archival file state"
-#: ./opengever/document/behaviors/metadata.py:143
+#: ./opengever/document/behaviors/metadata.py:150
 msgid "label_archival_file_state"
 msgstr "Status Archivdatei"
 
+#. Default: "Attach to email"
+#: ./opengever/document/browser/templates/file.pt:54
+#: ./opengever/document/widgets/tooltip.pt:42
+msgid "label_attach_to_email"
+msgstr "Mit Mailprogramm versenden"
+
 #. Default: "Author"
-#: ./opengever/document/behaviors/metadata.py:114
-#: ./opengever/document/browser/report.py:46
+#: ./opengever/document/behaviors/metadata.py:119
+#: ./opengever/document/browser/report.py:32
 msgid "label_author"
 msgstr "Autor"
 
@@ -313,7 +319,7 @@ msgid "label_cancel"
 msgstr "Abbrechen"
 
 #. Default: "Change archival file"
-#: ./opengever/document/browser/archival_file_form.py:53
+#: ./opengever/document/browser/archival_file_form.py:49
 #: ./opengever/document/browser/templates/archiv_file.pt:24
 msgid "label_change_archival_file"
 msgstr "Archivdatei ändern"
@@ -329,8 +335,9 @@ msgstr "Ausgecheckt"
 msgid "label_checkin_journal_comment"
 msgstr "Journal-Kommentar"
 
-#. Default: "Edit Document"
-#: ./opengever/document/browser/templates/file.pt:36
+#. Default: "Checkout and edit"
+#: ./opengever/document/browser/templates/file.pt:38
+#: ./opengever/document/widgets/tooltip.pt:27
 msgid "label_checkout_and_edit"
 msgstr "Auschecken / Bearbeiten"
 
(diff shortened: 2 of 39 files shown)


opengever/officeconnector/helpers.py
```diff
@@ -0,0 +1,17 @@
+from plone import api
+from opengever.officeconnector.interfaces import IOfficeconnectorSettings
+
+
+def is_officeconnector_enabled():
+    # This will or feature checks in the future
```
lukasgraf: This isn't being used ATM. Is this really needed? Also, is it really possible to boil down any combination of specific settings (outlook attach enabled, direct checkout enabled) to a simple "office connector enabled as a whole"?

opengever/officeconnector/interfaces.py
```diff
@@ -0,0 +1,22 @@
...
+                default=u'OfficeConnector Outlook support'),
+        description=_(u'label_enable_officeconnector_description',
+                      default=(u'Enable attach to Outlook with the new style '
+                               u'OfficeConnector URLs')),
+        default=False)
+
+    direct_checkout_and_edit_enabled = schema.Bool(
+        title=_(u'label_enable_officeconnector_direct_checkout_and_edit',
+                default=u'OfficeConnector direct checkout suppport'),
+        description=_(u'label_enable_officeconnector_description',
```
lukasgraf: The message ID `label_enable_officeconnector_description` is shared with the setting above, with a different default 💥  (`bin/i18n-build` should complain about this, if I'm not mistaken).

opengever/officeconnector/service.py
```diff
@@ -0,0 +1,80 @@
...
+from Products.CMFCore.utils import getToolByName
+from Products.PluggableAuthService.interfaces.plugins import IAuthenticationPlugin  # noqa
+from zExceptions import NotFound
+
+import json
+
+
+class OfficeConnectorToken(Service):
+
+    def create_officeconnector_url(self, context, payload={}):
```
lukasgraf: Careful! Mutable keyword argument that is being mutated in this function! (L44 ff) - see http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

I'm assuming this hasn't been an issue because with a correct invocation, `payload` will always be supplied, so the default is basically unused?

opengever/officeconnector/service.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class OfficeConnectorAttachToken(OfficeConnectorToken):
+
+    def render(self):
+        # Feature disabled or used wrong
+        if not is_officeconnector_attach_feature_enabled():
+            return None
+        payload = {'action': 'attach'}
+        return super(OfficeConnectorAttachToken,
```
lukasgraf: The `super()` call isn't needed here - just  `self.create_officeconnector_url(...)` is enough. Basic MRO will locate the method on the superclass.

`super()` is only needed in the method definition, when *extending* a method inherited from a superclass, to refer to the superclass' implementation (or more precisely, to the one next in line in terms of MRO).

opengever/officeconnector/configure.zcml
```diff
@@ -0,0 +1,42 @@
+<configure
+    xmlns="http://namespaces.zope.org/zope"
+    xmlns:genericsetup="http://namespaces.zope.org/genericsetup"
+    xmlns:grok="http://namespaces.zope.org/grok"
+    xmlns:i18n="http://namespaces.zope.org/i18n"
+    xmlns:plone="http://namespaces.plone.org/plone"
+    xmlns:profilehook="http://namespaces.zope.org/profilehook"
```
lukasgraf: `grok` and `profilehook` namespaces are unused.

opengever/officeconnector/browser/feature_enabled.py
```diff
@@ -0,0 +1,15 @@
+from Products.Five import BrowserView
+from opengever.officeconnector.helpers import is_officeconnector_attach_feature_enabled  # noqa
+from opengever.officeconnector.helpers import is_officeconnector_checkout_feature_enabled  # noqa
+
+
+class OfficeconnectorAttachEnabledView(BrowserView):
```
lukasgraf: I think I'd rather stick these in a single browser view as methods, and allow the methods to be traversed.

So `document-1/@@officeconnector-settings/is_attach_enabled` vs. 
`document-1/@@is_officeconnector_attach_enabled`.

This can be done by allowing traversal to those methods by using `allowed_attributes` in the ZCML declaration and listing the two allowed method names explicitly, or by using an interface that describes all public methods, and whitelisting all of them at once like done in the [`@@gever-state` view](https://github.com/4teamwork/opengever.core/blob/master/opengever/base/browser/configure.zcml#L65-L71).

opengever/officeconnector/tests/test_api.py
```diff
@@ -0,0 +1,110 @@
...
+from opengever.api.testing import RelativeSession
+from opengever.core.testing import OPENGEVER_FUNCTIONAL_ZSERVER_TESTING
+from opengever.officeconnector.interfaces import IOfficeconnectorSettings
+from opengever.testing import FunctionalTestCase
+from plone import api
+from plone.app.testing import SITE_OWNER_NAME
+from plone.app.testing import SITE_OWNER_PASSWORD
+
+import json
+import jwt
```
lukasgraf: This should go in `test_requires` in `setup.py` - it's only being pulled in indirectly through `plone.restapi` as of now.

docs/HISTORY.txt
```diff
@@ -26,7 +35,6 @@ Changelog
   - Introduces a new role APIUser
   - By default, only the Manager and APIUser roles have the "plone.restapi: Use REST API" permission
   - The APIUser role isn't assigned to any users or groups by default
-
```
lukasgraf: This blank is actually needed for this nested bullet list to be valid ReStructuredText. `rst2html.py docs/HISTORY.txt > /dev/null` :wink:

opengever/officeconnector/helpers.py
Rotonen: This is indeed a remnant of a different approach from the past to have been used in ZPT.

Rotonen:
@lukasgraf round 2 🔔

docs/HISTORY.txt
```diff
@@ -5,7 +5,16 @@ Changelog
...
-- Remove ftw.footer, replace with static footer viewlet. [jone]
+- Modified opengever.document to conditinally use the new officeconnector
+  functionality
+  [Rotonen]
+
+- Added the preliminary release of the opengever.officeconnector subproduct,
+  which only provides the attach-to-outlook functionality for now.
+  [Rotonen]
+
+- Remove ftw.footer, replace with static footer viewlet.
```
lukasgraf: This is already in `master`, though with whitespace changes. Rebase fail?

opengever/officeconnector/browser/settings.py
```diff
@@ -0,0 +1,19 @@
...
+from opengever.officeconnector.helpers import is_officeconnector_checkout_feature_enabled  # noqa
+from opengever.officeconnector.interfaces import IOfficeconnectorSettings
+from plone.memoize.view import memoize_contextless
+from Products.Five import BrowserView
+from zope.interface import implements
+
+
+class SettingsView(BrowserView):
+
+    implements(IOfficeconnectorSettings)
```
lukasgraf: This implements the wrong interface, no?. Should be `IOfficeconnectorSettingsView`.

opengever/officeconnector/configure.zcml
```diff
@@ -0,0 +1,40 @@
...
+    <plone:service
+      method="GET"
+      for="opengever.document.document.IDocumentSchema"
+      accept="application/json"
+      factory=".service.OfficeConnectorAttachToken"
+      name="officeconnector_attach_token"
+      permission="zope2.View"
+      />
+
+      <plone:service
```
lukasgraf: Indentation is messed up here on the second `service` block. Also, I would tend to use 4 spaces of indentation for ZCML as well. We don't have a clear style guide for that, but it's the style more commonly in use.

opengever/officeconnector/service.py
lukasgraf: Still a `super` call left over?

Rotonen:
@lukasgraf round 3 🔔 

I'll consider the eventual HISTORY.txt rebase a TKO.

Rotonen:
@lukasgraf on top of the changelog rebase, we're still missing German translations at this point.

lukasgraf:
`label_checkout_and_edit` is easy - `Auschecken / Bearbeiten` (same as before).

The "attach" functionality is a bit more tricky, let me think about that.

Rotonen:
Also journalization:

1) Attaching to email does not have a separate journal entry as there will be an entry of the file download anyway

or

2) Generation of an attach officeconnector URL will leave a journal entry

or

3) There should be a separate view and a journal entry for download-to-attach.

or

4) Options 2 and 3 together

Currently we're rolling with option 1.

lukasgraf:
Re: Journalization:

I'd say in the end it would be desirable to have the specific action taken on the document journalized (not so much for accountability purposes, but for ease of debugging / clarity) (so 2. or 3., or possibly both).

For now however, it's IMHO enough if a document can't be retrieved from GEVER without leaving some sort of trace. -> 1. is enough.

Rotonen:
The future approach will require some forethought.

From my point of view 2. is a bit messy as we do not have a guarantee of anything having successfully happened at that point yet. Especially as this is done at client-side javascript and if that doesn't succesfully hand off to the OS layer this can create a massive burst of events as per user retries.

Also 3. is  a bit messy as that way lies a path of making the download views a bushy thicket of special views.

I'll make that a ticket once this lands.

Rotonen:
@lukasgraf round 4 🔔

Rotonen:
@deiferni @lukasgraf also unified file editing in templates - round 5 🔔

opengever/officeconnector/interfaces.py
```diff
@@ -0,0 +1,31 @@
...
+    direct_checkout_and_edit_enabled = schema.Bool(
+        title=_(u'label_enable_officeconnector_direct_checkout_and_edit',
+                default=u'OfficeConnector direct checkout suppport'),
+        description=_(u'label_enable_officeconnector_direct_checkout_and_edit_description', # noqa
+                      default=(u'Enable direct checkout and edit with the '
+                               u'new style OfficeConnector URLs')),
+        default=False)
+
+
+class IOfficeconnectorSettingsView(Interface):
```
deiferni: This should be proper camel-case: IOffice**C**onnectorSettingsView

opengever/document/locales/fr/LC_MESSAGES/opengever.document.po
```diff
@@ -323,8 +329,10 @@ msgstr "Avec checkout"
 msgid "label_checkin_journal_comment"
 msgstr "Commentaire de l'historique"
 
-#. Default: "Edit Document"
-#: ./opengever/document/browser/templates/file.pt:36
+#. Default: "Checkout and edit"
+#: ./opengever/document/browser/templates/file.pt:38
+#: ./opengever/document/widgets/tooltip.pt:27
+#, fuzzy
```
deiferni: there is a fuzzy translation in here, please double-check and remove the fuzzy comment

opengever/document/locales/de/LC_MESSAGES/opengever.document.po
```diff
@@ -329,8 +335,10 @@ msgstr "Ausgecheckt"
 msgid "label_checkin_journal_comment"
 msgstr "Journal-Kommentar"
 
-#. Default: "Edit Document"
-#: ./opengever/document/browser/templates/file.pt:36
+#. Default: "Checkout and edit"
+#: ./opengever/document/browser/templates/file.pt:38
+#: ./opengever/document/widgets/tooltip.pt:27
+#, fuzzy
```
deiferni: same here

opengever/document/static/officeconnector.js
```diff
@@ -0,0 +1,31 @@
+$(document).on('click', '#checkout_url', function(){
+  event.preventDefault();
+  // This relies on us always being able to figure out the document url
+  // via the download URL link.
+  document_download_url = $('#action-download')[0].href;
```
deiferni: could you also use data-attributes for the urls here, e.g. `data-officeconnector-checkout-token-url`?

opengever/document/static/officeconnector.js
```diff
@@ -0,0 +1,31 @@
...
+  document_url = document_download_url.substring(0, document_download_url.lastIndexOf('/'));
+  $.ajax({
```
deiferni: i think you could use `$.getJSON` here and then drop providing accept-header and JSON-parsing the response

opengever/officeconnector/profiles/default/registry.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0"?>
+<registry>
+    <records interface="opengever.officeconnector.interfaces.IOfficeconnectorSettings" />
```
deiferni: Same here, it is Office**C**onnector

opengever/officeconnector/service.py
```diff
@@ -0,0 +1,78 @@
...
+        # likelihood the Plone site keyring.
+        for id_, authenticator in authenticators:
+            if authenticator.meta_type == "JWT Authentication Plugin":
+                plugin = authenticator
+                break
+
+        if not plugin:
+            return None
+
+        payload['download'] = addTokenToUrl('/'.join([context.absolute_url(), 'download_file_version'])) # noqa
```
deiferni: or just `\n` after `absolute_url()` 😉

opengever/officeconnector/service.py
```diff
@@ -0,0 +1,78 @@
...
+        # Attach the file info if this is a file
+        try:
+            payload['filename'] = context.file.filename
```
deiferni: use `context.get_filename`, which returns `None` if there is no file

opengever/officeconnector/service.py
```diff
@@ -0,0 +1,78 @@
...
+            payload['content-type'] = context.file.contentType
+        except AttributeError:
```
deiferni: i wouldn't catch `AttributeError` if not absolutely necessary, that may shadow other `AttributeError`s you did not think of in that context, especially since `@property` is a thing, better explicitly check if the attribute is there

opengever/officeconnector/service.py
```diff
@@ -0,0 +1,78 @@
...
+
+        return json.dumps(response)
+
+
+class OfficeConnectorAttachToken(OfficeConnectorToken):
+
+    def render(self):
+        # Feature disabled or used wrong
+        if not is_officeconnector_attach_feature_enabled():
+            return None
```
deiferni: in bumblebee we `raise NotFound`, that exactly happens if you return None here? Also should we handle that in OfficeConnector (if not done so already)?

opengever/officeconnector/service.py
```diff
@@ -0,0 +1,78 @@
...
+        payload = {'action': 'attach'}
+        return self.create_officeconnector_url(self.context, payload=payload)
+
+
+class OfficeConnectorCheckoutToken(OfficeConnectorToken):
+
+    def render(self):
+        # Feature disabled or used wrong
+        if not is_officeconnector_checkout_feature_enabled():
+            return None
```
deiferni: Same here, maybe `raise NotFound`

opengever/officeconnector/service.py
```diff
@@ -0,0 +1,78 @@
...
+from Products.CMFCore.utils import getToolByName
+from Products.PluggableAuthService.interfaces.plugins import IAuthenticationPlugin  # noqa
+from zExceptions import NotFound
+
+import json
+
+
+class OfficeConnectorToken(Service):
+
+    def create_officeconnector_url(self, context, payload):
```
deiferni: you pass `context`, but also use `self.context`, i'd drop the context argument here

opengever/officeconnector/service.py
```diff
@@ -0,0 +1,78 @@
...
+            payload['filename'] = context.file.filename
+            payload['content-type'] = context.file.contentType
+        except AttributeError:
+            pass
+
+        user_id = api.user.get_current().getId()
+        response = {}
+        response['token'] = plugin.create_token(user_id, data=payload)
+
+        return json.dumps(response)
```
deiferni: there is a `JSONResponse` class that you could use here, e.g. `JSONResponse(self.request).data(response).dump()`

benefits:
- takes care of stuff you should do like setting the content-type headers
- clearly states what you're doing here

opengever/officeconnector/service.py
deiferni: ah i see, it seems to return `204`, but is that correct?

opengever/officeconnector/tests/test_feature.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class TestIsOfficeconnectorFeatureEnabled(FunctionalTestCase):
+
+    def test_if_registry_entries_are_false_per_default(self):
+        self.assertFalse(is_officeconnector_attach_feature_enabled())
+        self.assertFalse(is_officeconnector_checkout_feature_enabled())
+
+    def test_if_registry_entries_are_true(self):
+
```
deiferni: ✂️  whitespace

opengever/officeconnector/tests/test_feature.py
```diff
@@ -0,0 +1,95 @@
...
+        api.portal.set_registry_record(
+            'direct_checkout_and_edit_enabled',
+            True,
+            interface=IOfficeconnectorSettings)
+
+        self.assertTrue(is_officeconnector_attach_feature_enabled())
+        self.assertTrue(is_officeconnector_checkout_feature_enabled())
+
+    def test_if_registry_entries_are_false(self):
+
```
deiferni: ✂️  whitespace

opengever/officeconnector/tests/test_feature.py
```diff
@@ -0,0 +1,95 @@
...
+            interface=IOfficeconnectorSettings)
+
+        self.assertFalse(is_officeconnector_attach_feature_enabled())
+        self.assertFalse(is_officeconnector_checkout_feature_enabled())
+
+
+class TestIsOfficeconnectorFeatureEnabledView(FunctionalTestCase):
+
+    def test_if_registry_entries_are_false_per_default(self):
+
```
deiferni: ✂️  whitespace

opengever/officeconnector/tests/test_feature.py
```diff
@@ -0,0 +1,95 @@
...
+        self.assertFalse(
+            self.portal.restrictedTraverse(
+                '@@officeconnector_settings/is_attach_enabled')())
+
+        self.assertFalse(
+            self.portal.restrictedTraverse(
+                '@@officeconnector_settings/is_checkout_enabled')())
+
+    def test_if_registry_entry_is_true(self):
+
```
deiferni: ✂️  whitespace

opengever/officeconnector/tests/test_feature.py
```diff
@@ -0,0 +1,95 @@
...
+        self.assertTrue(
+            self.portal.restrictedTraverse(
+                '@@officeconnector_settings/is_attach_enabled')())
+
+        self.assertTrue(
+            self.portal.restrictedTraverse(
+                '@@officeconnector_settings/is_checkout_enabled')())
+
+    def test_false_if_registry_entry_is_false(self):
+
```
deiferni: ✂️  whitespace

opengever/document/static/officeconnector.js
Rotonen: No, `$.getJSON()` slaps on extra entries into the `Accept` header which do not work with `plone.rest` content negotiation strategies. Due to legacy Zope krakens just under the surface, there can only be one entry in the `Accept` header at this point.

Adding one header here is IMO neater than mangling / header blacklisting on the `$.getJSON()` call.

opengever/officeconnector/service.py
Rotonen: AFAIK `plone.rest.Service` handles the response side of this. I'm essentially just mirroring how authentication works in `plone.restapi`.

The `JSONResponse` class seems to be a Django thing?

opengever/officeconnector/service.py
Rotonen: In plone.restapi this was the prevalent pattern for the auth I'm mirroring so I went with this, but that might as well just be a way of thinking about auth thing, not a REST thing.

Will go with 404 here as that makes more sense to me too.

Rotonen: Hmm on a second note `raise NotFound` will make a full Zope-side HTML page render, which will quite confuse JSON clients?

Rotonen: I'm actually just incorrect, `plone.rest.Service` renders JSON here with a `raise NotFound`.

opengever/document/locales/fr/LC_MESSAGES/opengever.document.po
Rotonen: The label was mentioned twice in an older .pot file in-development and i18n-build reruns do not clean up the comments of the fuzzy flags unless they regenerate the whole entry.

Nice catch.

Is this a gettext bug or are we just not covering this corner in our gettext locale build process?

opengever/document/static/officeconnector.js
Rotonen: That is a whole feature of HTML5 I've actually just missed so far. Will have to read up on good patterns and reevaluate.

opengever/officeconnector/service.py
deiferni: ah yes of course, nvm, i assumed this was a `BrowserView`.

opengever/document/locales/fr/LC_MESSAGES/opengever.document.po
deiferni: No it is a feature 😉 , the user is required to cleanup the fuzzy flags manually. That makes sense since the fuzzy match may have been wrong.

Rotonen: Yes, the fuzzy comment itself is a feature and makes sense.

Is the fuzzy comment not getting cleared up on successive builds when the initial conditino which puts it into place is cleared a bug or simply undefined behaviour / user error?

deiferni: I think there is no way to know why the fuzzy comment got there in the first place, so it just stays (als long as the msgid exists) until manually cleared up.

Rotonen: Should we have a grep check in the CI tests for these?

There is currently one in disposition.

deiferni: Yeah, would make sense, we don't want them in a `.po` file, ever.

Rotonen: I verified the one in disposition is also a dud like this, taking the liberty of removing that as well.

Rotonen:
Rebased, ready for rereview, round 6 🔔

Rotonen:
Rebased.

opengever/document/static/officeconnector.js
```diff
@@ -0,0 +1,23 @@
+$(document).on('click', '#officeconnector-checkout-url', function(){
+  event.preventDefault();
```
Rotonen: I was under the impression there is an implicit event in .on() calls in jQuery.

So `$(document).on('click', '#officeconnector-checkout-url', function(event){`?

bierik: 👍  exactly -> http://www.w3schools.com/jquery/event_preventdefault.asp

Rotonen:
@bierik ready for rereview.

opengever/document/static/officeconnector.js
```diff
@@ -0,0 +1,22 @@
...
+    officeconnector_config.url = $('#officeconnector-attach-url').data('officeconnectorAttachTokenUrl');
+    $.ajax(officeconnector_config).done(openOfficeconnector);
+});
+
+$(document).on('click', '#officeconnector-checkout-url', function(event){
+    event.preventDefault();
+    officeconnector_config.url = $('#officeconnector-checkout-url').data('officeconnectorCheckoutTokenUrl')
+    $.ajax(officeconnector_config).done(openOfficeconnector);
+});
+config
```
bierik: ✂️

Rotonen:
Amended, rebased, ready for rereview.

Rotonen:
🔔 🔔 🔔 

@lukasgraf @deiferni

PR #2505, opened 2017-01-31
Fixed SIP package name, include disposition transfer number if exists.

phgross:
Replaced the placeholder `myRef` with the transfer number if it exits.

Fix according to the users Testfeedback.

### DIFF
--- a/opengever/disposition/tests/test_sippackage.py
+++ b/opengever/disposition/tests/test_sippackage.py
@@ -21,17 +21,23 @@ class TestSIPPackage(FunctionalTestCase):
 
         SIP_[Ablieferungsdatum]_[Name der ablieferenden Stelle]_[Referenz].
         """
-        dossier = create(Builder('dossier').within(self.folder))
+        dossier = create(Builder('dossier').within(self.folder).as_expired())
+        disposition = create(Builder('disposition')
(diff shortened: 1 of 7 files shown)


opengever/disposition/tests/test_sippackage.py
```diff
@@ -21,17 +21,30 @@ def test_sip_folder_name_correspond_to_ech0160_definition(self):
...
+            package = SIPPackage(disposition)
             self.assertEquals(
-                'SIP_20161106_PLONE_MyRef', package.get_folder_name())
+                'SIP_20161106_PLONE', package.get_folder_name())
+
+        disposition.transfer_number = u'10\xe434'
+        with freeze(datetime(2016, 11, 6)):
+            package = SIPPackage(disposition)
+            self.assertEquals(
+                u'SIP_20161106_PLONE_10\xe434', package.get_folder_name())
```
deiferni: could also you add a test that test without transfer number?

phgross: @deiferni see line 32 😉 ...

deiferni: 🙈  sorry i was looking for another test-method, nvm.

PR #2537, opened 2017-02-08
OGGBundle import: Implement setting workflow state for imported objects

lukasgraf:
This implements setting the workflow state for imported objects to the state given in the `review_state` property of the bundle item.

Largely based on https://github.com/4teamwork/opengever.bsmigration/blob/master/opengever/bsmigration/sections/workflow.py

@deiferni


opengever/bundle/sections/workflow.py
```diff
@@ -0,0 +1,84 @@
...
+                yield item
+                continue
+
+            if statekey not in item:  # no WF state given
+                yield item
+                continue
+
+            path, state = item[pathkey], item[statekey]
+
+            obj = traverse(self.context, path, None)
```
deiferni: Is there still a reference to the constructed object in `item`? in that case i'd drop the pathkey stuff and just grab the object from `item`.

opengever/bundle/sections/workflow.py
```diff
@@ -0,0 +1,84 @@
...
+                    "Cannot set workflow state for %s. "
+                    "Object doesn't exist" % path)
+                yield item
+                continue
+
+            try:
+                wf_ids = self.wftool.getChainFor(obj)
+                if wf_ids:
+                    wf_id = wf_ids[0]
+                    comment = 'Set dossier state upon import.'
```
deiferni: i think this is not limited to dossiers only, i'd adapt the message

opengever/base/schemadump/config.py
```diff
@@ -18,6 +18,36 @@
     'opengever.repository.repositoryroot': 'reporoot',
 }
 
+# Workflow states allowed in JSON schemas. The state that's listed first
+# indicates the initial workflow state (i.e. default state)
+ALLOWED_REVIEW_STATES = {
```
deiferni: could you add a short test that ensures these are kept in sync with the workflow definition?

opengever/bundle/sections/workflow.py
lukasgraf: So you also wouldn't log the `path` upon failure? Otherwise we can't drop the pathkey stuff

opengever/bundle/sections/workflow.py
lukasgraf: 👍

opengever/base/schemadump/config.py
lukasgraf: Then I basically already need to implement the fully dynamic export of available workflow states 😉

opengever/bundle/sections/workflow.py
deiferni: ah makes sense, the just drop the `traverse` stuff 😉

opengever/base/schemadump/config.py
deiferni: ok - as discussed this is out of scope now 👍

lukasgraf:
@deiferni updated

PR #2539, opened 2017-02-08
Improve Dossier/DossierTemplate by several features

maethu:
theming: https://github.com/4teamwork/plonetheme.teamraum/pull/502

- [x] Use ftw.keywordwidget

- [x] Restricted Keywords
restricted

- [x] Predefined Keywords
predefined

- [x] No predefined Keywords
not_predefined

- [x] No predefined Keywords and not restricted
not_predefined_not_restricted


versions.cfg
```diff
@@ -13,6 +13,7 @@ mr.developer = 1.33
 pyflakes = 0.9.1
 z3c.dependencychecker = 1.11
 zptlint = 0.2.4
+ftw.keywordwidget = 1.0.4
```
deiferni: Could you move this pinning to http://kgs.4teamwork.ch/release/opengever/develop

sources.cfg
```diff
@@ -5,8 +5,12 @@ extensions = mr.developer
 development-packages =
   opengever.maintenance
   collective.js.timeago
+  plonetheme.teamraum
 
 auto-checkout = ${buildout:development-packages}
 
+[branches]
+plontheme.teamraum = mle-select2-styles
```
deiferni: should be dropped before merge

maethu: @deiferni pls merge this one first. https://github.com/4teamwork/plonetheme.teamraum/pull/500

opengever/dossier/indexers.py
```diff
@@ -14,6 +14,13 @@
 
 
 @indexer(IDossierMarker)
+def SubjectIndexer(obj):
+    aobj = IDossier(obj)
+    return aobj.keywords
+grok.global_adapter(SubjectIndexer, name="Subject")
```
maethu: @deiferni The index is currently empty and not in use. Only plone default components are using this index. GEVER has no default component, which needs this index.

sources.cfg
maethu: https://github.com/4teamwork/plonetheme.teamraum/pull/502

versions.cfg
maethu: @deiferni done

opengever/dossier/indexers.py
maethu: I do no **re**using the index I just *using* it 😉

opengever/dossier/dossiertemplate/behaviors.py
```diff
@@ -20,9 +19,10 @@ class IDossierTemplateSchema(form.Schema):
 
     form.fieldset(
         u'common',
-        label=base_mf(u'fieldset_common', default=u'Common'),
```
maethu: @deiferni I removed this during rebase, since it's not necessary to define a label, if the fieldset with label is defined previously.

maethu: @deiferni ahhh no it's necessary... damn. hold on

maethu:
@deiferni Seems to be fine know :-) sry for confusing.

maethu:
@deiferni Some test cases are failing in several ways. Even if I fix the missing commas. Example: the test-case `test_show_only_whitelisted_schema_fields_in_edit_form` has been change generally an does not match the changes made in this PR. I'm working on it.

maethu:
@deiferni Elio was mean to me and refactored some test assertions I also changed. So... I had to change them again.

PR #2552, opened 2017-02-11
Implement opengever.theme to replace plonetheme.teamraum:gever

raphael-s:
🚧 🚧 🚧 

Adds `opengever.theme` which was previously implemented in [`plonetheme.gever`](https://github.com/4teamwork/plonetheme.gever). Since this theme is only used for GEVER we do not need a seperate package for it, we can just implement it in `opengever.core` itself.
This replaces `plonetheme.teamraum:gever` which we want to remove, because it doesn't have anything to do with GEVER per se. We want a theme which is only used for GEVER and not for other projects.

🚧 There are still issues over at [`plonetheme.gever`](https://github.com/4teamwork/plonetheme.gever) which we need to solve before merging this.

/cc @deiferni


deiferni:
Thanks 💛 

As discussed this is currently on hold.

PR #2557, opened 2017-02-13
🚧 POC layer isolation

deiferni:
🚧

### DIFF
--- a/opengever/core/testing.py
+++ b/opengever/core/testing.py
@@ -1,16 +1,19 @@
 from collective.taskqueue.interfaces import ITaskQueue
 from collective.transmogrifier import transmogrifier
+from ftw.builder import Builder
+from ftw.builder import create
 from ftw.builder import session
 from ftw.builder.testing import BUILDER_LAYER
 from ftw.builder.testing import set_builder_session_factory
 from ftw.bumblebee.tests.helpers import BumblebeeTestTaskQueue
+from ftw.dictstorage.sql import DictStorageModel
 from ftw.testing import ComponentRegistryLayer
 from ftw.testing.quickinstaller import snapshots
 from opengever.activity.interfaces import IActivitySettings
 from opengever.base import model
 from opengever.base.model import create_session
 from opengever.bumblebee.interfaces import IGeverBumblebeeSettings
-from opengever.dossier.dossiertemplate.interfaces import IDossierTemplateSettings # noqa
+from opengever.dossier.dossiertemplate.interfaces import IDossierTemplateSettings  # noqa
 from opengever.meeting.interfaces import IMeetingSettings
 from opengever.officeatwork.interfaces import IOfficeatworkSettings
 from opengever.ogds.base.setup import create_sql_tables
@@ -30,8 +33,15 @@ from plone.browserlayer.utils import unregister_layer
 from plone.dexterity.schema import SCHEMA_CACHE
 from plone.testing import Layer
 from plone.testing import z2
+from plone.testing.z2 import zopeApp
 from Products.CMFCore.utils import getToolByName
+from sqlalchemy import create_engine
+from sqlalchemy import MetaData
+from sqlalchemy.orm import scoped_session
+from sqlalchemy.orm import sessionmaker
 from sqlalchemy.pool import StaticPool
+from sqlalchemy_utils import create_database
+from sqlalchemy_utils import drop_database
 from Testing.ZopeTestCase.utils import setupCoreSessions
 from z3c.saconfig import EngineFactory
 from z3c.saconfig import GloballyScopedSession
@@ -40,9 +50,13 @@ from z3c.saconfig.interfaces import IScopedSession
 from zope.component import getSiteManager
 from zope.component import provideUtility
 from zope.configuration import xmlconfig
+from zope.globalrequest import setRequest
 from zope.sqlalchemy import datamanager
+
 import logging
 import os
+import random
+import string
 import sys
 import transaction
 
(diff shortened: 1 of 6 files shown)


test-plone-4.3.x.cfg
```diff
@@ -15,11 +15,14 @@ test-egg = opengever.core[api, tests]
...
 [test]
 arguments = ['-s', '${buildout:package-namespace}', '--exit-with-status', '--auto-color', '--auto-progress', '--xml', '--package-path', '${buildout:directory}/${buildout:package-namespace}', '${buildout:package-namespace}']
-
+eggs +=
+  psycopg2
 
 [test-jenkins]
 test-command = bin/mtest $@
+eggs +=
+  psycopg2
```
jone: I believe this is not necessary since there is no DB connection initiated from the `mtest` script, right?

opengever/core/testing.py
```diff
@@ -241,6 +244,40 @@ def allowAllTypes(self, portal):
         portal.portal_types['Plone Site'].filter_content_types = False
 
 
+OPENGEVER_FIXTURE_SQL = OpengeverFixtureSQL()
+
+
+class OpengeverFixture(PloneSandboxLayer):
```
jone: If I understand this correctly, the difference between `OpengeverFixtureSQL` and `OpengeverFixture`  is that `OpengeverFixture` _does_ setup the SQL, but `OpengeverFixtureSQL` _does not_ setup the SQL, right?

This sounds confusing for me, I would have named it the other way round.
Can you explain this to me?

opengever/core/testing.py
```diff
@@ -241,6 +244,40 @@ def allowAllTypes(self, portal):
...
+OPENGEVER_FIXTURE = OpengeverFixture()
+
+
+class APILayer(Layer):
+    """A layer that installs the plone.restapi:default generic setup profile.
+    """
+
+    def setUp(self):
+        with ploneSite() as site:
+            applyProfile(site, 'plone.restapi:default')
```
jone: The restapi is now installed by default since https://github.com/4teamwork/opengever.core/pull/2476 , so this can be removed, I guess.

opengever/core/testing.py
```diff
@@ -268,6 +305,16 @@ def functional_session_factory():
     return sess
 
 
+def functional_session_factory_sql():
+    sess = session.BuilderSession()
+    # Auto commit must be False to enable matroska rollbacks!
+    sess.auto_commit = False
```
jone: Given the test:
```python
@browsing
def test_car_colorization(self, browser):
    car = create(Builder('car').colored('red'))
    browser.open(car)
    self.assertEquals('red', browser('.car-color').first.text)
```

- I have to use functional tests when using a browser
- I expect the Builder to auto-commit, so I do not have to write `transaction.commit()`
- The transaction must be committed after creating the "car" so that the browser (using a new transaction) can access it.

By disabling auto-commit here, this no longer works.
How do you "fix" that? What are the thoughts about that?

opengever/core/testing.py
```diff
@@ -324,6 +374,164 @@ def inactivate_filing_number(portal):
...
+        # the underlying layers and further tests will unwantedly try to use
+        # whatever matroska engine we last used!
+        super(SQLTestLayer, self).tearDown()
+
+    def nextMatroska(self):
+        # Create a new database
+        self.current_matroska_engine = create_engine(
+            'postgresql+psycopg2:///opengever_test_{0}'
+            .format(''.join(
+                random.choice(string.ascii_lowercase + string.digits)
```
jone: What's the chance of having a conflict between multiple proceesses running in parallel?
Should we test for whether there already is a database?

opengever/core/testing.py
```diff
@@ -324,6 +374,164 @@ def inactivate_filing_number(portal):
...
+        model.Session.bind = self.current_matroska_engine
+
+    def seedData(self):
+        # Circumvent our improper use of ftw.builder
+        with zopeApp() as app:
+            session.current_session = session.factory()
+            setRequest(app.REQUEST)
+            setRoles(api.portal.get(),
+                     api.user.get_current().id,
+                     ['Contributor', 'Editor', 'Reader', 'Member'])
```
jone: If I understand this correctly, we now have more roles in the tests than we had before.
Is this intended?

opengever/testing/test_case_sql.py
```diff
@@ -0,0 +1,183 @@
...
+from Products.CMFCore.utils import getToolByName
+from zope.component import getUtility
+from zope.i18n import translate
+from zope.interface import alsoProvides
+
+import json
+import unittest2
+
+
+class TestCaseSQL(unittest2.TestCase):
```
jone: What is the idea here? Why `SQL`?

opengever/testing/test_case_sql.py
```diff
@@ -0,0 +1,183 @@
...
+        super(FunctionalTestCaseSQL, self).setUp()
+        self.portal = self.layer['portal']
+        self.app = self.layer['app']
+        self.request = self.app.REQUEST
+        alsoProvides(self.request, IOpengeverBaseLayer)
+        self.membership_tool = getToolByName(self.portal, 'portal_membership')
+
+        self.grant('Contributor', 'Editor', 'Reader')
+
+        # necessary to force tabbed-view into correct mode, otherwise it only
```
jone: This not a more `correct mode`, it is just "another" mode.
It makes the tabbedview render HTML-tables instead of initializing the ExtJS library (which does not work with the current testbrowser).

Actually the ExtJS-mode is more "correct" since it is the one actually used in production 😏

opengever/testing/test_case_sql.py
```diff
@@ -0,0 +1,183 @@
...
+
+import json
+import unittest2
+
+
+class TestCaseSQL(unittest2.TestCase):
+    pass
+
+
+class FunctionalTestCaseSQL(TestCaseSQL):
```
jone: I realized that this class is mostly a copy of `.test_case. FunctionalTestCase`.
What is the reason for not inheriting the code instead of copying it?

opengever/testing/test_case_sql.py
```diff
@@ -0,0 +1,183 @@
...
+        self.assertEquals(title, translate(action.get('title')))
+
+    """
+    Browser API
+    Deprecated, please don't extend this API and use
+    opengever.testing.browser whenever possible.
+    """
+
+    def assertPageContains(self, text):
+        self.assertIn(text, self.browser.contents)
```
jone: This is very bad: a failing assertion message contains all the HTML of the page, which will make me 😡 
Please do not do such things, but use selectors for being more specific.

opengever/testing/test_case_sql.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def assertPageContainsNot(self, text):
+        self.assertNotIn(text, self.browser.contents)
```
jone: This is very bad: a failing assertion message contains all the HTML of the page, which will make me 😡 
Please do not do such things, but use selectors for being more specific.

opengever/testing/tests/test_sql_layer_isolation.py
```diff
@@ -0,0 +1,67 @@
...
+from opengever.core.testing import OPENGEVER_SQL_TEST_LAYER
+from opengever.ogds.models.admin_unit import AdminUnit
+from opengever.testing import FunctionalTestCaseSQL
+
+import transaction
+
+
+class TestLayerIsolation(FunctionalTestCaseSQL):
+
+    layer = OPENGEVER_SQL_TEST_LAYER
```
jone: Why isn't the `FunctionalTestCaseSQL` already using this layer? Why is there a `OPENGEVER_SQL_TEST_LAYER` and a `OPENGEVER_FUNCTIONAL_TESTING_SQL` layer?

opengever/testing/test_case_sql.py
Rotonen: The original reason is to maximize rebaseability against master during initial development and to also allow more decoupled development of the concept itself.

opengever/testing/test_case_sql.py
Rotonen: Please split this into its own ticket as that is also in the normal test case then.

test-plone-4.3.x.cfg
Rotonen: There is a new db connection every time a new matroska is spawned.

opengever/core/testing.py
Rotonen: The namespacing is an artefact of this being 'let us use a real SQL server to allow for persistent fixtures'. Most of the real time in a test run is spent waiting for Builder to endlessly recreate objects into an in-memory SQLite.

All naming within the concept may be considered working titles at this point.

opengever/core/testing.py
Rotonen: Browser tests are not conceivable to move onto the persistent fixtures approach at this point. Neither is anything which commits in-test. We rely on the transaction manager clean state to be able to avoid spawning new matroskas. If there has been a commit a new matroska gets spawned as we cannot be sure of a cleanup.

This still leaves the majority of the test-run up for performance gains.

opengever/core/testing.py
Rotonen: Very high in anything which touches meetings.

opengever/core/testing.py
Rotonen: This indeed looks to be a copy/paste fail.

opengever/testing/test_case_sql.py
Rotonen: https://github.com/4teamwork/opengever.core/blob/e5834941ffd0a6551c9032edd11bc5ad5722a3a2/opengever/testing/test_case.py#L167

test-plone-4.3.x.cfg
jone: hmm, but there is no import in `bin/mtest`.
`bin/mtest` runs subprocesses with `bin/test` (which already has the `psycopg2` egg).
I still don't understand why this needs to be added to the `[test-jenkins]` eggs, as it is already in the `[test]` eggs.

opengever/core/testing.py
jone: Fair enough 👍 I think it did confuse me here and there 😉

opengever/core/testing.py
jone: So that means we cannot use the testbrowser with the new fixture, right?
Is there a plan or ideas to make that work?

At the moment there are 1284 of 3105 tests using the testbrowser. It would be great to improve those too.

opengever/core/testing.py
jone: Can we include the `ZSERVER_PORT` environment variable in the database name?
This should make it impossible that two test-processes conflict here, since they cannot be run in parallel with the same `ZSERVER_PORT` (that's `bin/mtest` and Jenkins Port Allocator business).

Rotonen: I do not oppose, but then we need to rely more on the developer environment for running tests. When weighing this we opted for random.

test-plone-4.3.x.cfg
Rotonen: I do not recall why I ended up with that anymore at this point. Remove and see what happens?

opengever/core/testing.py
Rotonen: This means we cannot use the test browser with the new fixture layers, yes.

There is a persistent plan to figure that stuff out at some point, but as far as we can see at this point this will require us to either majorly extend or reimplement the Zope transaction manager. We opted to take the 2/3 hit on the potential performance gain instead.

opengever/testing/test_case_sql.py
jone: Fair enough 👍
Would be grate to change that to subclassing before merging...

opengever/testing/test_case_sql.py
jone: => https://github.com/4teamwork/opengever.core/pull/2621

opengever/testing/test_case_sql.py
jone: 👍 I have realized now that this is a copy 😉 
=> Fixed in https://github.com/4teamwork/opengever.core/pull/2621

opengever/testing/test_case_sql.py
jone: 👍 => https://github.com/4teamwork/opengever.core/pull/2620

opengever/core/testing.py
jone: Ok 👍

opengever/core/testing.py
jone:
```python
port = int(os.environ.get('ZSERVER_PORT', 55001))
```
That's how its done in `plone.testing.z2`. This should work for all cases: locally, jenkins, with `bin/test` and with `bin/mtest`. I rely on that on other places such as in `ftw.testbrowser`.

Rotonen: We do still need a random component there for using nested layers with matroskas.

The grand idea is to have a base persistent fixture for all of GEVER and then build more specific ones on top of that one.

Rotonen: It could also be in theory SQL variant agnostic if we want to test certain stuff against multiple SQL vendors, but that is pending us to solve a few differences like the already solved SQLite vs. PostgreSQL sequence numbers.

deiferni:
@jone afaik there should be no need to keep this POC around longer, agree?

jone:
Sure 👍 There are also some branches with Postgres implementations around, but at the moment we do not require Postgres.

PR #2558, opened 2017-02-13
Defer creation of initial versions to end of OGGBundle pipeline

lukasgraf:
- Defer the creation of initial versions to the **end of the OGGBundle pipeline** by introducing a dedicated post-processing section
- Make sure the *automatic* creation of initial versions is actually disabled
- Use a **request layer** instead of monkey patching to disable automatic creation of initial versions
- Do **intermediate commits** during creation of initial versions. This addresses
  a **performance issue** where the time to write initial versions increases
  exponentially with the number of previous initial versions created in the
  same transaction.

Explicit commits of the transaction will be logged and indicated in the transaction note. Example how this will look in the ZMI Undo tab:

undo

@deiferni

### DIFF
--- a/opengever/bundle/sections/post_processing.py
+++ b/opengever/bundle/sections/post_processing.py
@@ -12,6 +12,7 @@ log.setLevel(logging.INFO)
 
 
 INTERMEDIATE_COMMIT_INTERVAL = 1000
+VERSIONABLE_TYPES = ('opengever.document.document', 'ftw.mail.mail')
 
 
 class PostProcessingSection(object):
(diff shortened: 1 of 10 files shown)


opengever/bundle/sections/post_processing.py
```diff
@@ -38,6 +45,20 @@ def __iter__(self):
...
+                    "Intermediate commit during OGGBundle post-processing. "
+                    "%s of %s items." % (count, len(items_to_post_process)))
+
+        self.commit_and_log(
+            "Final commit after OGGBundle post-processing. "
+            "%s of %s items." % (count, len(items_to_post_process)))
+
+    def commit_and_log(self, msg):
+        transaction.get().note(msg)
+        log.info(msg)
```
deiferni: i'd swap commit and log, that way the message is logged after the final commit and we know we're done for sure.

lukasgraf: 👍  - will update this accordingly in the next upcoming PR

PR #2561, opened 2017-02-14
OGGBundle pipeline: Implement basic reports for imported objects

lukasgraf:
This implements some basic reports for imported objects.

In addition to the summary at the start, right after reading the contents from the bundle, this will also produce reports at the end of the import that are based on what objects actually could be found in **Plone** (i.e. based on catalog queries).

This includes a very short summary of object counts in ASCII, and a much more detailed XLSX report.

@deiferni

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,11 @@ Changelog
 4.15.2 (unreleased)
 -------------------
 
+- OGGBundle pipeline: Implement basic reports for imported objects. After
+  a bundle import, a quick ASCII summary and a much more detailed XLSX
+  report will be generated.
+  [lgraf]
+
 - Fix send_documents view on the inbox.
   Allows the functionality to save sent files in the dossier
   only for IDossierMarker-types.
--- a/opengever/bundle/report.py
+++ b/opengever/bundle/report.py
@@ -1,0 +1,64 @@
+from collections import OrderedDict
+from opengever.bundle.loader import BUNDLE_JSON_TYPES
+from opengever.bundle.sections.constructor import BUNDLE_GUID_KEY
+from plone import api
+from zope.annotation import IAnnotations
+import logging
+
+
+log = logging.getLogger(__name__)
+log.setLevel(logging.INFO)
+
+
+class DataCollector(object):
+    """Collect report data from Plone.
+    """
+
+    def __init__(self, bundle):
+        self.bundle = bundle
+
+    def __call__(self):
+        data = {}
+        catalog = api.portal.get_tool('portal_catalog')
+
+        portal_types = BUNDLE_JSON_TYPES.values()
+        portal_types.append('ftw.mail.mail')
+
+        for portal_type in portal_types:
+            data[portal_type] = []
+            log.info("Collecting %s" % portal_type)
+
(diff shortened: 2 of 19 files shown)


docs/HISTORY.txt
```diff
@@ -5,6 +5,9 @@ Changelog
 4.15.2 (unreleased)
 -------------------
 
+- OGGBundle constructor: Store bundle GUID on Plone obj for later.
```
deiferni: could you revise the cl entry? i'd add a more general one about the added reporting.

opengever/bundle/report.py
```diff
@@ -0,0 +1,237 @@
...
+            [('guid', guid), ('path', path), ('title', title)])
+
+        file_field = self.get_file_field(obj)
+        if file_field:
+            item_info['file_size'] = file_field.getSize()
+
+        return item_info
+
+    def get_permissions(self, obj, guid):
+
```
deiferni: ✂️

opengever/bundle/report.py
```diff
@@ -0,0 +1,237 @@
...
+        local_roles = obj.get_local_roles()
+        inheritance_blocked = getattr(obj, '__ac_local_roles_block__', False)
+
+        # Alway include at least one row per object with the info whether
```
deiferni: typo: Alway**s**

opengever/bundle/sections/report.py
```diff
@@ -0,0 +1,58 @@
...
+        self.build_ascii_summary(report_data)
+        self.build_xlsx_report(report_data)
+
+    def build_ascii_summary(self, report_data):
+        summary = ASCIISummaryBuilder(report_data).build()
+        log.info('\n\n%s\n' % summary)
+
+    def build_xlsx_report(self, report_data):
+        try:
+            report_path = PathFinder().var + '/report.xlsx'
```
deiferni: i'd use `os.path.join` here

opengever/bundle/tests/__init__.py
```diff
@@ -5,6 +5,10 @@
 
 @implementer(IAttributeAnnotatable)
 class MockTransmogrifier(object):
+    """
```
deiferni: i'd just `pass` if there is no docstring, or provide a docstring 😉

docs/HISTORY.txt
lukasgraf: Yep, I forgot the overall changelog-entry completely. Will add one, and instead omit this implementation detail from the changelog

opengever/bundle/report.py
lukasgraf: 👍

opengever/bundle/report.py
lukasgraf: 👍

opengever/bundle/sections/report.py
lukasgraf: 👍

opengever/bundle/tests/__init__.py
lukasgraf: Personal preference, I think it looks slightly nicer to have an empty docstring instead of `pass` 😉  - but I'll change it 👍

lukasgraf:
@deiferni updated!

PR #2565, opened 2017-02-14
OGGBundle report: Make getting file size more safe

lukasgraf:
OGGBundle report: Make getting file size more safe to deal with the case of documents that don't have a file.

@deiferni


lukasgraf:
@deiferni as discussed, it's quite tricky and probably not worth it to write a test for this. Documents without file are currently not supported on a schema level (`filepath` is required, and must be of type `string` (not `null`)).

So the only way this could be provoked (which is also what actually lead to this failure) is if a file is not found or gets skipped because of business rules.

PR #2569, opened 2017-02-15
FileLoader: Add support for loading files from UNC paths.

lukasgraf:
This implements loading files that are referenced via **UNC paths**.

For UNC paths to work, the bundle import needs to be provided with some configuration that maps UNC mounts to respective POSIX paths:

```json
{
    "unc_mounts": {
        "\\\\host_xy\\mount": "/mnt/host_xy_fileshare"
    }
}
```

In order to supply this configuration, the concept of **Ingestion Settings** is introduced.
These are supposed to be stored in `~/.opengever/bundle_ingestion/settings.json` on the machine, because these are site-specific settings that depend on the environment where the bundle is being imported.

During tests, they need to be injected instead of attempting to read them from a file that won't exist on the CI server or developers machine.

*(The name "ingestion settings", which admittedly is a bit clunky, was chosen to avoid confusion with other concepts. `import_config.py` could easily be read as "import (verb) the configuration". And the name "import" is already heavily overloaded. "configuration" conflicts with the concept of the "bundle configuration", which is something else entirely)*

@deiferni


lukasgraf:
@deiferni phew 😉 🎉 - rebased *(tests still running)*

PR #2578, opened 2017-02-16
Allow add root content only for managers.

### LINKED ISSUE #2246: Fix root-content creation permissions for Administrator
Currently an `Administrator` can create root content (Template-Folder, Inbox, ...), this should be possible for `Manager` only.

elioschmutz:
This PR removes the permission to add root content for administrators.

GIF as Administartor:
screen2

closes #2246

### DIFF
--- a/opengever/dossier/tests/test_templatefolder.py
+++ b/opengever/dossier/tests/test_templatefolder.py
@@ -13,6 +13,7 @@ from opengever.core.testing import OPENGEVER_FUNCTIONAL_MEETING_LAYER
 from opengever.dossier.docprops import TemporaryDocFile
 from opengever.dossier.interfaces import ITemplateFolderProperties
 from opengever.dossier.templatefolder import get_template_folder
+from opengever.dossier.templatefolder.interfaces import ITemplateFolder
 from opengever.dossier.tests import OGDS_USER_ATTRIBUTES
 from opengever.journal.handlers import DOC_PROPERTIES_UPDATED
 from opengever.journal.tests.utils import get_journal_entry
@@ -476,6 +477,7 @@ class TestTemplateFolder(FunctionalTestCase):
 
     @browsing
     def test_adding(self, browser):
+        self.grant('Manager')
         add_languages(['de-ch'])
         browser.login().open(self.portal)
         factoriesmenu.add('Template Folder')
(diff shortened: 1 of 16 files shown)


opengever/meeting/profiles/default/rolemap.xml
```diff
@@ -15,8 +15,6 @@
 
     <permission name="opengever.meeting: Add Committee" acquire="True">
       <role name="Manager"/>
-      <role name="Contributor"/>
```
deiferni: the commitee must still be addable, the correct type would be commiteecontainer

elioschmutz: The committee was never addable on the Plone-Site.

There is no separate permission for the committeecontainer. They both uses the `opengever.meeting: Add Committee` permission.

The difference makes the workflow. While the committeecontainer is addable on a plonesite, the committee is only addable in a committeecontainer. On the plonesite, only a manager has the `opengever.meeting: Add Committee` permission which means, only the manager can add a committeecontainer on a plone siteroot. The committee container has its own workflow (https://github.com/4teamwork/opengever.core/blob/master/opengever/meeting/profiles/default/workflows/opengever_committeecontainer_workflow/definition.xml#L41) where the Committee is still addable for Contributors.

deiferni: ok, sorry, i feared that the content-types were somehow mixed up 😉  

I don't see a reason not to have an own permission for the container though, this is very confusing (and probably my bad 🙈 ). Could you add the permission `opengever.meeting: Add CommitteeContainer` as well? I'd say that would be in scope for this PR.

elioschmutz: 👍

elioschmutz:
@deiferni 👍 done!

opengever/dossier/tests/test_templatefolder.py
```diff
@@ -476,14 +477,35 @@ class TestTemplateFolder(FunctionalTestCase):
...
 
         browser.fill({'Title': 'Templates',
                       'Responsible': TEST_USER_ID}).save()
 
-        self.assertEquals('tabbed_view', plone.view())
+        self.assertTrue(ITemplateFolder.providedBy(browser.context))
+
+    @browsing
+    def test_is_only_addable_by_manager(self, browser):
+        add_languages(['de-ch'])
```
deiferni: setting the language should not be necessary here?

opengever/inbox/tests/test_inbox.py
```diff
@@ -2,19 +2,43 @@
...
 
 
 class TestInbox(FunctionalTestCase):
 
-    def setUp(self):
-        super(TestInbox, self).setUp()
+    @browsing
+    def test_adding(self, browser):
+        self.grant('Manager')
+        add_languages(['de-ch'])
```
deiferni: i think you don't need to set the language here as long as you don't test explicitly for the language related stuff, but i'm not 100% sure as the translated title-fields might actually rely on certain languages being available. could you verify/test whether it works without setting the title?

(for all the following occasions as well)

elioschmutz: If the language is not set explicitly, it will be `fr` and `de`. This means, there would be a german and a french title field which are both required. This blows up the testing-setup if we want to create content with the testbrowser.

This is a known pattern in the gever tests... I just reused it from other tests...

deiferni: Yes but it is also in some tests where you don't create content over the factories-menu, so setting it there may not be required. Could you double-check those, i'd like to avoid unnecessary setup in our tests whenever possible.

elioschmutz: Thats true 👍 

I removed the unused language-settings.

PR #2588, opened 2017-02-20
testing: install opengever.policy.base:default.

jone:
Change the test setup so that we install ``opengever.policy.base:default`` rather than its dependencies, so that we include the configuration from the policy profile.

The language is reset to English so that we do not have to change all the tests.

Because a bunch of tests rely on having ``filter_content_types=False`` configured for the ``Plone Site`` FTI, we still keep that for now. This should be awoided for future tests and can be elminiated step by step.

----

This change was initially planned to be part of the profile merge, but it makes sense as a standalone change too.

### DIFF
--- a/opengever/base/tests/test_pasting_allowed.py
+++ b/opengever/base/tests/test_pasting_allowed.py
@@ -18,7 +18,7 @@ class TestPastingAllowed(FunctionalTestCase):
 
         browser.open(contactfolder)
         actions = browser.css('#plone-contentmenu-actions li').text
-        self.assertSequenceEqual(['Export as Zip'], actions)
+        self.assertSequenceEqual([], actions)
 
     @browsing
     def test_paste_action_not_displayed_for_templatefolder(self, browser):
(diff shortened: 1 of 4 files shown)


opengever/dossier/tests/test_templatefolder.py
```diff
@@ -607,7 +607,7 @@ def test_receipt_delivery_and_subdossier_column_are_hidden_in_trash_tab(self, br
     def test_enabled_actions_are_limited_in_document_tab(self, browser):
         browser.login().open(self.templatefolder, view=DOCUMENT_TAB)
 
-        self.assertEqual(
+        self.assertItemsEqual(
```
deiferni: i think the order should always be guaranteed and should remain constant so i'd rather adapt the list this once than use `assertItemsEqual`.

jone: @deiferni I have realized that this is not needet at this point and therefore have reverted all changes in this file.

jone:
The tests fail because there already is a problem which will be avoided by installing og.policy.base.
The failure is solved in the PR #2592
Once rebased, I expect everything to pass again.

jone:
@deiferni @lukasgraf rebased and ready 😉

PR #2582, opened 2017-02-16
Reindex SearchableText (dossier) after changing the comment thru the overlay

maethu:
Closes #2574

PR #2589, opened 2017-02-20
Dossier Note Overlay: Enable CSRF protection by optimizing a test case.

maethu:
This is a fix of a feature introduced with #2582 
There's no changelog entry, since it's a bugfix of a feature in the same release.

### DIFF
--- a/opengever/dossier/tests/test_dossier_byline.py
+++ b/opengever/dossier/tests/test_dossier_byline.py
@@ -8,6 +8,7 @@ from opengever.core.testing import activate_filing_number
 from opengever.core.testing import inactivate_filing_number
 from opengever.dossier.behaviors.dossier import IDossier
 from opengever.testing import create_ogds_user
+from plone.protect import createToken
 from plone.registry.interfaces import IRegistry
 from zope.component import getUtility
 from zope.intid.interfaces import IIntIds
@@ -156,9 +157,12 @@ class TestDossierByline(TestBylineBase):
     @browsing
     def test_dossier_byline_save_comments_endpoint(self, browser):
         payload = '{"comments": "New comment"}'
-        browser.login().visit(self.dossier,
-                              view='save_comments',
-                              data={'data': payload})
+        browser.login().visit(self.dossier)
+
+        browser.visit(self.dossier,
+                      view='save_comments',
+                      data={'data': payload,
+                            '_authenticator': createToken()})
 
         dossier_data = IDossier(self.dossier)
         self.assertEquals("New comment", dossier_data.comments)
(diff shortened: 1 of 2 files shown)


opengever/dossier/tests/test_dossier_byline.py
```diff
@@ -156,9 +156,13 @@ def test_dossier_byline_comments_editlink_data(self, browser):
...
-        browser.login().visit(self.dossier,
-                              view='save_comments',
-                              data={'data': payload})
+        browser.login().visit(self.dossier)
+
+        token = browser.css('#editNoteOverlay [name="_authenticator"]').first
+        browser.visit(self.dossier,
+                      view='save_comments',
+                      data={'data': payload,
+                            '_authenticator': token.value})
```
deiferni: i think you could just use `createToken` from `plone.protect` instead.

maethu: @deiferni Good to know, thank you!!

BTW:
I still got this warning while running the tests:

```
error parsing dom, failure to add csrf token to response for url http://nohost/plone/dossier-1/save_comments
error parsing dom, failure to add csrf token to response for url http://nohost/plone/dossier-1/save_comments
error parsing dom, failure to add csrf token to response for url http://nohost/plone/dossier-1/save_comments
error parsing dom, failure to add csrf token to response for url http://nohost/plone/dossier-1/save_comments
error parsing dom, failure to add csrf token to response for url http://nohost/plone/dossier-1/save_comments
error parsing dom, failure to add csrf token to response for url http://nohost/plone/dossier-1/save_comments
```

The problem ist that I respond with a 204 (no content). So there's no content to add any token??

deiferni: yes the token can only be added to a response that contains HTML, otherwise the response not processed (same for JSON).

maethu: k

deiferni: @maethu we just discussed this, can you double check your response headers, i.e. make sure that the content-types are set correctly? It only attempts to transform when the header is set to be`text/html`-ish.

deiferni:
@maethu we just discussed this, can you double check your response headers, i.e. make sure that the content-types are set correctly? It only attempts to transform when the header is set to betext/html-ish.

opengever/dossier/tests/test_dossier_byline.py
lukasgraf: Simply adding a content type of `application/json` to the `204` response should do the trick.

maethu:
@deiferni Well... I only set the disable diazo header... Since it's 204 I assumed the response header (content type) does not matter. Apparently it does. I changed it to JSON.

PR #2590, opened 2017-02-20
Translate missing "open detail view" in bumblebee-overlay.

### LINKED ISSUE #2580: Missing translation "open detail view"
- open an overlay from a document listing or from the dossier overview
- missing translation for the link to the document-overview

screen shot 2017-02-16 at 11 57 50

elioschmutz:
bildschirmfoto 2017-02-20 um 16 32 55

closes #2580

### DIFF
--- a/opengever/bumblebee/browser/templates/bumblebeeoverlay.pt
+++ b/opengever/bumblebee/browser/templates/bumblebeeoverlay.pt
@@ -106,6 +106,7 @@
       <li tal:define="link view/overlay/get_detail_view_url" tal:condition="python:not view.on_detail_view and link">
         <a id="action-view"
             tal:attributes="href link"
+            i18n:domain="opengever.bumblebee"
             i18n:translate="label_open_detail_view">Open detail view</a>
       </li>
     </ul>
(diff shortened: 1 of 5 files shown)


opengever/bumblebee/browser/templates/bumblebeeoverlay.pt
```diff
@@ -106,6 +106,7 @@
       <li tal:define="link view/overlay/get_detail_view_url" tal:condition="python:not view.on_detail_view and link">
         <a id="action-view"
             tal:attributes="href link"
+            i18n:domain="opengever.bumblebee"
```
deiferni: i think it should not be necessary to define the `i18n:domain` here, the root-tag of the template should already have an `i18n:domain` declaration which should then also apply to its children (and if there is no such declaration there i'd add it on the root tag).

elioschmutz: It has, but the containing `ul` element defines a new domain `opengever.document` for its children. So we have to reset it again to `opengever.bumblebee`

deiferni: :man_facepalming: oh well

PR #2593, opened 2017-02-21
Allow Admins to view and edit all private content.

maethu:
This PR changes the permissions for `Administrator` on private content. They are now allowed to view and edit all private content. They are not allowed to create new content as in a publicly visible dossier since there is no use-case to perform agency actions for private content. Also this side-steps permission problems that would arise because the workflows are heavily based on the content's `Owner`.

- [x] In DEV environment every member  of `DEVELOPMENT_USERS_GROUP` was able to see private documents of other users. 
- [x] GEVER let `Admin` edit and view private content 
- [x] Change Docu: https://docs.onegovgever.ch/user-manual/meine_ablage/

Closes https://github.com/4teamwork/opengever.ftw/issues/10


deiferni:
I'll hijack this branch 😈

deiferni:
Tested locally, seems to work as epxected.

PR #2638, opened 2017-02-28
DefaultConstrainTypeDecider: read max dossier nesting levels from registry

lukasgraf:
The `DefaultConstrainTypeDecider` used for dossiers used to have the maximum nesting levels hardcoded.

This change makes it determine these values dynamically from the according registry setting instead.

Also includes a **test that failed against the previous implementation**.

@deiferni

### DIFF
--- a/opengever/dossier/base.py
+++ b/opengever/dossier/base.py
@@ -348,24 +348,40 @@ class DefaultConstrainTypeDecider(grok.MultiAdapter):
     grok.adapts(Interface, IDossierMarker, IDexterityFTI)
     grok.name('')
 
-    CONSTRAIN_CONFIGURATION = {
-        'opengever.dossier.businesscasedossier': {
-            'opengever.dossier.businesscasedossier': 2,
-            'opengever.dossier.projectdossier': 1,
-            },
-        'opengever.private.dossier': {
-            'opengever.private.dossier': 2
-        },
-        'opengever.dossier.projectdossier': {
-            'opengever.dossier.projectdossier': 1,
-            'opengever.dossier.businesscasedossier': 1,
-            },
-        }
-
     def __init__(self, request, context, fti):
         self.context = context
         self.request = request
         self.fti = fti
+
+    @property
+    def CONSTRAIN_CONFIGURATION(self):
+        """Return the dynamically computed default constrain configuration.
+
+        This previously was a class-level variable, and is being extended by
+        type-specific ConstrainTypeDeciders in policies (e.g. opengever.zug).
+        """
+        # The registry record 'maximum_dossier_depth' actually describes the
+        # maximum levels of *sub*dossiers allowed. That's why we increase this
+        # value by one below, in order to get the maximum allowed nesting
+        # level for dossiers in general.
+        max_subdossier_levels = api.portal.get_registry_record(
+            'maximum_dossier_depth', interface=IDossierContainerTypes)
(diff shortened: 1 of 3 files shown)


opengever/dossier/base.py
```diff
@@ -348,25 +348,41 @@ class DefaultConstrainTypeDecider(grok.MultiAdapter):
...
+        # The registry record 'maximum_dossier_depth' actually describes the
+        # maximum levels of *sub*dossiers allowed. That's why we increase this
+        # value by one below, in order to get the maximum allowed nesting
+        # level for dossiers in general.
+        max_subdossier_levels = api.portal.get_registry_record(
+            'maximum_dossier_depth', interface=IDossierContainerTypes)
+        max_dossier_levels = max_subdossier_levels + 1
+
+        config = {
+            'opengever.dossier.businesscasedossier': {
```
deiferni: could you refactor this in such a way that all dossiers that should respect the configured nesting limit can be defined as class-attribute? E.g. as a tuple

```python
constrained_dossier_types = ('opengever.dossier.businesscasedossier', 'opengever.private.dossier',)
```
this may simplify customising in policies see for an example: https://github.com/4teamwork/opengever.zug/blob/97b72ee00d61a5e6b31632850dd91991d699518a/opengever/zug/bdarp/menu.py#L45-L53

opengever/dossier/base.py
```diff
@@ -348,25 +348,41 @@ class DefaultConstrainTypeDecider(grok.MultiAdapter):
...
+                'opengever.dossier.businesscasedossier': max_dossier_levels,
+                'opengever.dossier.projectdossier': 1,
+            },
+            'opengever.private.dossier': {
+                'opengever.private.dossier': 2
```
deiferni: I think this should also be `max_dossier_levels`

lukasgraf: @deiferni Really? I think it might be exactly the use case that some organizations want to control nesting depth in the public repository, but leave it up to people to create their own mess in their private structures.

opengever/dossier/base.py
lukasgraf: @deiferni that wouldn't be a refactoring, that would be a behavioral change. Right now, the configuration of nesting depths for regular dossiers and all other dossier types are independent of each other. With that change we would be grouping together two dossier types that I'm not sure belong together at this point.

opengever/dossier/base.py
deiferni: Hm, if that is the case shouldn't we expose the private dossier nesting level as a separate configuration option for easier customization? (would be out of scope for this PR though).

opengever/dossier/base.py
deiferni: ok i see, as discussed it is actually desired to have the option to have different nesting levels per type. this comment is void 😉 .

lukasgraf:
@deiferni nope, no change to policies needed as far as I can tell. Also, I didn't see any redundancy that would need to be eliminated in policies.

deiferni:
i think we can drop

https://github.com/4teamwork/opengever.hpcdual/blob/master/opengever/hpcdual/dossier_restricition.py
https://github.com/4teamwork/opengever.edk/blob/master/opengever/edk/dossierrestriction.py

they just mirror the configuration value of `maximum_dossier_depth`. project-dossier is irrelevant as far as i know 😉 .

no other SaaS customers seem to have override this configuration, so for that part we should be safe.

PR #2645, opened 2017-03-01
Implement file upload form for documents

### LINKED ISSUE #2644: Create a form for replacing a file in a document
Design criterium:

* Cannot traverse to it unless the document is checked out by the current user
* Mandatory file field
  * Always replaces a file

Pointers:

https://github.com/4teamwork/opengever.core/blob/master/opengever/document/forms.py#L34

buchi:
Fixes #2644

### DIFF
--- a/opengever/document/forms.py
+++ b/opengever/document/forms.py
@@ -1,10 +1,16 @@
 from AccessControl import getSecurityManager
+from opengever.document.document import IDocumentSchema
 from opengever.document.interfaces import ICheckinCheckoutManager
 from opengever.document.interfaces import NO_DOWNLOAD_DISPLAY_MODE
 from opengever.document.interfaces import NO_DOWNLOAD_INPUT_MODE
 from plone.dexterity.browser.edit import DefaultEditForm
+from plone.dexterity.events import EditFinishedEvent
+from plone.dexterity.i18n import MessageFactory as _
+from plone.directives import form
 from plone.z3cform import layout
+from z3c.form import button
 from zope.component import queryMultiAdapter
+from zope.event import notify
 
 
 class DocumentEditForm(DefaultEditForm):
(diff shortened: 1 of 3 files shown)


opengever/document/forms.py
```diff
@@ -39,3 +45,42 @@ def updateWidgets(self):
...
+        data, errors = self.extractData()
+        if errors:
+            self.status = self.formErrorsMessage
+            self.request.response.setStatus(400)
+            return
+
+        current_user_id = getSecurityManager().getUser().getId()
+        manager = queryMultiAdapter((self.context, self.request),
+                                    ICheckinCheckoutManager)
+        if manager.get_checked_out_by() != current_user_id:
```
deiferni: `manager` already knows that, you can use `manager.is_checked_out_by_current_user()` here

opengever/document/forms.py
```diff
@@ -39,3 +45,42 @@ def updateWidgets(self):
...
+
+class IDocumentFileUploadForm(IDocumentSchema):
+    """Schema for the file upload form."""
+
+    form.omitted('title')
+
+
+class DocumentFileUploadForm(DefaultEditForm):
+    """A form for just uploading the file of a document without any metadata
+       fields. Used by Office Connector"""
```
deiferni: missing terminating dot 😉

opengever/document/forms.py
```diff
@@ -39,3 +45,42 @@ def updateWidgets(self):
...
+    schema = IDocumentFileUploadForm
+    additionalSchemata = ()
+    render_form = True
```
deiferni: as an alternative strategy you could be able to just select the desired fields here, i.e.:

```python
from z3c.form.field import Fields

    fields = Fields(ITheInterface).select(
        'file')
```

That would have the benefit of explicitly just select what you want and make that visible for other devs, also may be more stable in case we change fields in the future.

(untested, but something similar may work)

PR #2647, opened 2017-03-01
Improve disposal of inactive dossiers

phgross:
- Set end date to current date when deactivating a dossier, including an upgradestep which set the end date for inactive dossiers.
- Pre appraisal: Handle inactive dossier always as not archival worthy.
- Display inactive dossiers in a separate listing on the disposition overview.

bildschirmfoto 2017-03-01 um 17 33 48

### DIFF
--- a/opengever/disposition/browser/overview.py
+++ b/opengever/disposition/browser/overview.py
@@ -10,26 +10,49 @@ from Products.CMFPlone.CatalogTool import zero_fill
 from Products.Five.browser import BrowserView
 
 
+def sort_on_sortable_title(item):
+    if isinstance(item[0], unicode):
+        return num_sort_regex.sub(zero_fill, item[0])
+    return num_sort_regex.sub(zero_fill, item[0].Title())
+
+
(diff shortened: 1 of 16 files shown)


opengever/disposition/browser/overview.py
```diff
@@ -10,26 +10,52 @@
...
+        self.active_dossiers = sorted(
+            active_dossiers.items(), key=sort_on_sortable_title)
 
-        return sorted(grouped_dossiers.items(), key=sort_on_sortable_title)
+        self.inactive_dossiers = sorted(
+            inactive_dossiers.items(), key=sort_on_sortable_title)
+
+    def _add_to(self, mapping, dossier):
+        key = dossier.get_grouping_key()
+        if key not in mapping:
```
deiferni: you can write that as a one-liner:

```python
mapping.setdefault(key, []).append(dossier)
```

phgross:
@deiferni updated

deiferni:
lucky you, i'm gonna sneak that in before we do the xml reformat thing 😉

PR #2653, opened 2017-03-02
Fix off-by-one bumblebee preview image for document versions.

### LINKED ISSUE #2646: mismatch bumblebee preview for versions
the preview image for versions of a document is off by one (i.e. too old), see for details https://extranet.4teamwork.ch/support/gever-ai/tracker/233

deiferni:
The bumblebee checksum was not updated before saving the document to
portal_repository, thus version T of a document always displayed the
preview image for version T-1.

Fixes #2646.


opengever/bumblebee/tests/test_versions.py
```diff
@@ -0,0 +1,66 @@
...
+        self.assertEqual(
+            DOCX_CHECKSUM, IBumblebeeDocument(document).get_checksum())
+
+        manager = getMultiAdapter((document, self.portal.REQUEST),
+                                  ICheckinCheckoutManager)
+        manager.checkin()
+
+        # checksum has been updated
+        current_document_checksum = IBumblebeeDocument(document).get_checksum()
+        self.assertNotEqual(DOCX_CHECKSUM, current_document_checksum)
```
jone: Wouldn't it be better to assert the checksum of the new doc rather than verifying that it is not the initial?
I would define the checksum of `foo` above and use it here and on line 65.

deiferni:
@jone i just realised that the pr description misses that information, already updated seconds after your comment 😉

opengever/bumblebee/tests/test_versions.py
deiferni: not absolutetly necessary but nicer, you're right! i'll amend.

PR #2654, opened 2017-03-02
Update ftw.keywordwidget from 1.1.1 to 1.1.2 to fix an ie11 issue.

elioschmutz:
This PR uses the new ftw.keywordwidget version 1.1.2 to fix an ie11 issue.

See https://github.com/4teamwork/ftw.keywordwidget/pull/6 for more information.

Before:
screen2

After:
screen1

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,9 @@ Changelog
 4.15.2 (unreleased)
 -------------------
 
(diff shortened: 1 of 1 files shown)


deiferni:
We usually update http://kgs.4teamwork.ch/release/opengever/develop directly, i have already done so - you can drop the entry in `versions.cfg`. I'd keep the CL entry though that will be useful for the release notes.

PR #2669, opened 2017-03-08
Fix setting incorrect contentType for drag&drop uploaded *.msg mails.

### LINKED ISSUE #2567: Drag&Drop uploaded *.msg files are not registered with bumblebee
- d&d upload a *.msg to a dossier
- open dossier overview/overlay
- no preview image

the bumlebee checksum is there
probably something with events firing before the message has been converted to eml and then deciding that there is no file and thus not registering the document in bumblebee

\cc @bbuehlmann this the issue you reported yesterday, i can reproduce it

deiferni:
This PR fixes an issue with setting the correct message `contentType` for drag&drop uploaded  `*.msg` mails.

These mails are converted to `*.eml` during upload, but due to a problem with a temporarily wrong filename the `contentType` of their file ist not set correctly. This causes issue with bumblebee because it thinks that it does not support the content and thus does not render a preview image for these mails. This PR fixes that issue by setting the correct filename after conversion.

This PR also contains an upgrade step that checks all mails and has a look at all of them that
are not of the expected mime-type. It then sets the mime-type to what would
be set by `get_contenttype` while creating the mail's `NamedBlobFile`.

Fixes #2567.


deiferni:
yes i know, should not be called, working on it 🚧  😓

deiferni:
@lukasgraf test should pass now 🙏

lukasgraf:
@deiferni yay 👍 Will merge once green, ok?

PR #2670, opened 2017-03-08
Reset end date when reactivate a resolved or activate an inactive dossier.

### LINKED ISSUE #2667: Reset end date when reactivating a dossier
The `end` date of a dossier should be cleared when a dossier gets reactivated.

phgross:
Closes #2667

Additional I've added missing tests for the dossier reactivation.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,7 +5,7 @@ Changelog
 4.15.2 (unreleased)
 -------------------
 
-- Reset end date when reactivating a resolved dossier.
+- Reset end date when reactivating a resolved or activating an inactive dossier.
   [phgross]
 
 - Fix permission issue if user tries to trash a document in a private dossier.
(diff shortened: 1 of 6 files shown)


opengever/dossier/activate.py
```diff
@@ -33,12 +34,17 @@ def check_preconditions(self):
     def activate(self):
         # subdossiers
         for subdossier in self.context.get_subdossiers():
+            self.reset_end_date(subdossier.getObject())
```
deiferni: i would move everything that happens here to the dossier into an `activate` method, so that you can only tell it to do something here i.e. `self.context.activate()`.

docs/HISTORY.txt
```diff
@@ -5,6 +5,9 @@ Changelog
 4.15.2 (unreleased)
 -------------------
 
+- Reset end date when reactivate a resolved or activate an inactive dossier.
```
deiferni: reactivating

deiferni: activating

deiferni:
@phgross is this one ready for re-review?

phgross:
@deiferni yes

deiferni:
(rebased)

PR #2673, opened 2017-03-09
Fix permission issue if user tries to trash a document in a private dossier.

### LINKED ISSUE #2480: Papierkorb defekt in "Meine Ablage"
*Vorgehen*
- Meine Ablage öffnen
- Dossier dossier erstellen
- Dokument hochladen
- Im Dokumente-Tab des Dossiers das Dokument auswählen
- "In den Papierkorb verschieben"

*Falsches Resultat*
- Status message `Sie haben keine Objekte ausgewählt`
- Dokument nicht in Papierkorb verschoben

*Erwartet*
- Dokument im Papierkorb

elioschmutz:
This PR fixes a traversal permission issue.

Trashing a document will use restrictedTraverse to lookup the document (https://github.com/4teamwork/opengever.core/blob/master/opengever/trash/trash.py#L94)

Unfortunately, only the manager has the `Access Contents Information` permission on the private-root-folder. But the restrictedTraverse method needs the `Access Contents Information` permission for each object it traverses on.

To fix the issue, I allow `Access Contents Information` to `Members`.

screen2

closes #2480


deiferni:
We have investigated and `Member` is correct since the `users_group_id` is automatically assigned to the `Member` role during setup, see: https://github.com/4teamwork/opengever.core/blob/master/opengever/setup/creation/orgunit.py#L46-L51

elioschmutz:
@jone I adjusted the workflow. The user now has access to the root-folder itself, but he can't do anything on it. So I think it's ok.

Before:
screen1

After:
screen2

@jone Why should the `Access content information` and the `View` permission always be the same if there are two different permissions? I mean, in our case, the user don't need to view the content itself. He only has to access its properties.

jone:
@elioschmutz I like it 👍 
Having `Access content information` and `View` not in sync can lead to problems, such as a content is found in search / listings but cannot be accessed (no `Access content information`) or problems with relatiosn (no `View`). There are a lot of places where the on or the other is checked, but not both.
Years ago we have come to the solution that it is best to always keep them in sync. By following this thumb rule we do not need to completely understand all use cases. It's the safe path. 😉

elioschmutz:
ok, thanks @jone 🎉

PR #2716, opened 2017-03-13
Amend the JS redirector mechanism to support OC URLs

### LINKED ISSUE #2689: Also switch portal / item actions of documents to use new style OC URLs when configured
screen shot 2017-03-10 at 15 11 48

Currently at least the top action bar still does not change the checkout link type.

### LINKED ISSUE #2696: Dokument aus Vorlage erstellen: ZEM-Datei wird geliefert statt neues Format
Betrifft OfficeConnector:
dokausvorlage

Rotonen:
* Amend the OfficeConnector `plone.rest` URL services to allow one to init them in ~any `BrowserView`
* Amend the redirector to issue out an OC URL in case of the OC URL checkout feature being enabled
* Amend the JS redirector to handle `oc:foobar` URLs

Considerations:
* Diving deep into portal site_actions would not have squashed #2696
* Not amending the JS redirector trips up the popup blocker on Chrome
* Could avoid the JSON parsing within the redirector, but did not want to break up the service logic

Fixes #2689 
Fixes #2696

### DIFF
--- a/opengever/base/redirector.py
+++ b/opengever/base/redirector.py
@@ -1,7 +1,5 @@
 from five import grok
 from opengever.base.interfaces import IRedirector
-from persistent.dict import PersistentDict
-from persistent.list import PersistentList
 from plone.app.layout.viewlets.interfaces import IAboveContentTitle
 from zope.interface import Interface
 from zope.publisher.interfaces.browser import IBrowserRequest
@@ -29,7 +27,8 @@ class RedirectorCookie(object):
         return value
 
     def _get(self):
-        # Load from response cookie in case this request added / updated the cookie
+        # Load from response cookie in case this request added or updated the
+        # cookie
         cookie = self.request.response.cookies.get(REDIRECTOR_COOKIE_NAME, {})
         value = cookie.get('value', None)
         if value == 'deleted':
@@ -81,8 +80,7 @@ class Redirector(grok.Adapter):
 
 
 class RedirectorViewlet(grok.Viewlet):
-    """ Viewlet which adds the redirects for the IRedirector.
-    """
+    """Viewlet which adds the redirects for the IRedirector."""
 
     grok.name('redirector')
     grok.context(Interface)
--- a/opengever/base/redirector.py
+++ b/opengever/base/redirector.py
@@ -90,7 +90,11 @@ class RedirectorViewlet(grok.Viewlet):
     JS_TEMPLATE = '''
 <script type="text/javascript" class="redirector">
 $(function() {
+  if ('%(url)s'.split(':')[0] == 'oc') {
+    window.location = '%(url)s';
+  } else {
     window.setTimeout("window.open('%(url)s', '%(target)s');", %(timeout)s);
+  }
 });
 </script>
 '''
--- a/opengever/document/checkout/checkout.py
+++ b/opengever/document/checkout/checkout.py
@@ -3,6 +3,8 @@ from opengever.base.interfaces import IRedirector
 from opengever.document import _
 from opengever.document.document import IDocumentSchema
 from opengever.document.interfaces import ICheckinCheckoutManager
+from opengever.officeconnector.helpers import create_oc_url
+from opengever.officeconnector.helpers import is_officeconnector_checkout_feature_enabled  # noqa
 from Products.CMFCore.utils import getToolByName
 from Products.statusmessages.interfaces import IStatusMessage
 from zope.component import getMultiAdapter
@@ -24,7 +26,6 @@ class CheckoutDocuments(grok.View):
     grok.name('checkout_documents')
 
     def render(self):
-
         # check whether we have paths or not
         paths = self.request.get('paths')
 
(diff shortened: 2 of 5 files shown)


opengever/base/redirector.py
```diff
@@ -90,7 +90,13 @@ class RedirectorViewlet(grok.Viewlet):
     JS_TEMPLATE = '''
 <script type="text/javascript" class="redirector">
 $(function() {
+  if ('%(url)s'.split(':')[0] == 'http'
+      || '%(url)s'.split(':')[0] == 'https') {
     window.setTimeout("window.open('%(url)s', '%(target)s');", %(timeout)s);
+  }
+  else if ('%(url)s'.split(':')[0] == 'oc') {
```
jone: if the protocol happens to not be `http`, `https`or `oc` or there is no protocol (relative URL), nothing will happen, where before a `window.open` was tried.
In order to avoid changing the behavior of unexpected URLs (e.g. relative URLs) I propose to reverse the order of the conditions and make the previous behavior the default by putting it in an `else` without further condition:
```javascript
if ('%(url)s'.split(':')[0] == 'oc') {
  window.location = '%(url)s';
} else {
  window.setTimeout("window.open('%(url)s', '%(target)s');", %(timeout)s);
}
```

Rotonen:
@jone @deiferni ready for rereview.

opengever/officeconnector/service.py
```diff
@@ -17,6 +17,14 @@
 class OfficeConnectorURL(Service):
     """Create oc:<JWT> URLs for javascript to fetch and pass to the OS."""
 
+    def __init__(self, context=None, request=None):
+        # Handle cases where we jam this into unintended contexts
+        if context:
+            self.context = context
+        if request:
+            self.request = request
+        super(OfficeConnectorURL, self).__init__(context, request)
```
deiferni: i'm a bit confused as to why it is necessary to have optional arguments here? the code seems to actively rely on always having `self.request` and `self.context`, as far as i could see `Service` inherits from `object`, so we should always have to set those here? or am i missing some magic?

Rotonen: The `__init__` of `plone.rest.Service` is the `__init__` of `object`.

https://github.com/plone/plone.rest/blob/375c7c8306fd3f68f4ba8a327ee73950286ac453/src/plone/rest/service.py#L8

If we are to call it directly we need to be able to pass in a context and a request.

opengever/document/checkout/checkout.py
```diff
@@ -3,11 +3,15 @@
 from opengever.document import _
 from opengever.document.document import IDocumentSchema
 from opengever.document.interfaces import ICheckinCheckoutManager
+from opengever.officeconnector.helpers import is_officeconnector_checkout_feature_enabled  # noqa
+from opengever.officeconnector.service import OfficeConnectorCheckoutURL
 from Products.CMFCore.utils import getToolByName
 from Products.statusmessages.interfaces import IStatusMessage
 from zope.component import getMultiAdapter
 from zope.interface import Interface
```
Rotonen: It's actually the Plone isort ruleset and our only diff to it is the newline. The reason we do not use that is that ~any editor will be able to sort lines the very least with a bytewise sort, which will work for us.

https://docs.plone.org/develop/styleguide/python.html

OfficeConnector follows that so I've so far gone with it.

opengever/document/checkout/checkout.py
```diff
@@ -71,11 +74,15 @@ def render(self):
...
-
+            if not is_officeconnector_checkout_feature_enabled():
+                redirector.redirect(
+                    '%s/external_edit' % objects[0].absolute_url(),
+                    target='_self',
+                    timeout=1000)
+            else:
+                json_url = OfficeConnectorCheckoutURL(
+                    self.context, self.request).render()
+                redirector.redirect(json.loads(json_url)['url'])
```
Rotonen: I was mulling around how to split it up without messing up with the service error handling flow for the JS approach which avoids the page reload on clicking the checkout button on the file details templates (file, tooltip, bumblebee).

https://github.com/4teamwork/opengever.core/blob/55c656c4161731ea223cfce408f7f2c72a2b27b9/opengever/officeconnector/service.py#L17

One approach is to tear that down and use the redirector for all of them.

I'm open for suggestions.

opengever/document/checkout/checkout.py
deiferni: i'd follow gever conventions here for the sake of local consistency

opengever/officeconnector/service.py
deiferni: uhm yes, but then i would suggest to:

- don't make the parameters optional, since code relies on the attributes being set
- don't pass them to the `super` call since what won't work with `object`

opengever/document/checkout/checkout.py
deiferni: one idea, move payload/action to class level (quick sketch), 

- exposes action name as variable
- removes duplication of creating the payload
- you can define render once and then only configure

```
class OfficeConnectorURL(Service)
    action = None
    def render(self):
        # Feature disabled or used wrong
        if not is_officeconnector_checkout_feature_enabled():
            raise NotFound
        return self.create_officeconnector_url()

    def create_officeconnector_url(self):
        if not self.action:
            puke()
       ...
       payload = {'action': self.action}
       payload['url'] = self.get_the_url_with_a_better_name()

    def get_the_url_with_a_better_name(self):
        return '/'.join([
            api.portal.get().absolute_url(),
            'oc_' + self.action,
            api.content.get_uuid(self.context),
            ])

class OfficeConnectorAttachURL(OfficeConnectorURL):
    action = 'attach'
```

opengever/officeconnector/service.py
Rotonen: https://github.com/plone/plone.rest/blob/f711777060cdad4b1fa247af59cf8b4997218916/src/plone/rest/zcml.py#L108

It's a meta class of `BrowserView` under the hood, not an actual `object`.

Rotonen:
@jone @deiferni ready for rereview.

Issue #2724, opened 2017-03-13
OfficeConnector attach to email: BCC the parent dossier email


Rotonen:
Required for 4teamwork/OfficeConnector#132.

PR #2725, opened 2017-03-13
BCC dossier byline when attaching a document to emails via OC if dossier is open

Rotonen:
* Add the byline email of the parent dossier of a document as a BCC field to OfficeConnector email attach action payloads when said dossier is open.

Implements #2724

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,10 @@ Changelog
 4.15.2 (unreleased)
 -------------------
 
+- Add the mail-in address of the parent dossier of a document as a BCC field to
+  OfficeConnector email attach action payloads when said dossier is open.
+  [Rotonen]
+
 - Reset end date when reactivating a resolved or activating an inactive dossier.
   [phgross]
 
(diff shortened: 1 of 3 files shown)


opengever/officeconnector/service.py
```diff
@@ -110,7 +111,16 @@ class OfficeConnectorAttachPayload(OfficeConnectorPayload):
 
     def render(self):
         self.request.response.setHeader('Content-type', 'application/json')
-        return json.dumps(self.get_base_payload())
+
+        payload = self.get_base_payload()
+
+        # We assume the parent of a document is always a dossier
```
deiferni: unfortunately this assumption won't hold. documents can be inside tasks or proposals for example. you can use `self.document.get_parent_dossier()` though. be aware this may be `None` (which is valid) so you must handle that accordingly.

docs/HISTORY.txt
```diff
@@ -5,6 +5,10 @@ Changelog
 4.15.2 (unreleased)
 -------------------
 
+- Add the byline email of the parent dossier of a document as a BCC field to
```
deiferni: byline email -> mail-in address

Rotonen:
@deiferni ready for rereview.

PR #2695, opened 2017-03-10
Move dossier notes to dossier overview

bierik:
🚧  DO NOT MERGE YET!!!

⚠️  Depends on https://github.com/4teamwork/plonetheme.teamraum/pull/516

@deiferni I moved the note overlay markup to the dossier byline. Please make sure that this markup is not loaded in subdossiers. The javascript is working fine so far. Feel free to hijack my branch to proceed the work.

PR #2727, opened 2017-03-14
Move "Aktenotiz" out of the byline, add icon next to title and display text in overview

deiferni:
Styles in https://github.com/4teamwork/plonetheme.teamraum/pull/518

Display a sticky note icon next to the title and an edit link if there is a comment:

screen shot 2017-03-14 at 11 56 06

Display an add link in the overview if there is no comment yet:

screen shot 2017-03-14 at 11 56 22

Follow-up for https://github.com/4teamwork/opengever.core/pull/2695.

### DIFF
--- a/opengever/dossier/browser/overview.py
+++ b/opengever/dossier/browser/overview.py
@@ -11,6 +11,7 @@ from opengever.dossier.behaviors.dossier import IDossier
 from opengever.dossier.behaviors.dossier import IDossierMarker
 from opengever.dossier.behaviors.participation import IParticipationAware
 from opengever.dossier.dossiertemplate.behaviors import IDossierTemplateMarker
+from opengever.dossier.utils import truncate_ellipsis
 from opengever.globalindex.model.task import Task
 from opengever.ogds.base.actor import Actor
 from opengever.ogds.base.utils import get_current_admin_unit
@@ -26,6 +27,7 @@ from zope.security import checkPermission
 class DossierOverview(BoxesViewMixin, grok.View, GeverTabMixin):
 
     show_searchform = False
+    comments_max_length = 400
 
     grok.context(IDossierMarker)
     grok.name('tabbedview_view-overview')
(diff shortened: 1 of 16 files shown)


opengever/dossier/browser/overview.py
```diff
@@ -43,11 +45,21 @@ def catalog(self, types, showTrashed=False,
...
                  self.make_description_box()]]
 
+    def get_comments(self):
+        return truncate_ellipsis(
+            IDossier(self.context).comments, self.comments_max_length)
+
+    def make_comment_box(self, has_no_content_placeholder=True):
+        return dict(id='comments', content=self.get_comments(),
+                    label=_(u"label_comments", default="Comments"),
+                    has_no_content_placeholder=has_no_content_placeholder)
```
elioschmutz: The full `has_no_content_placeholder` implementation could be removed if you just return an empty space i.e. `' '` in the `get_comments` function if there are no comments.

If this is not an option, it would be great to change the default value of `has_no_content_placehoder` to `False` because it is only used once with this value. And this is the most common usecase for this function.

opengever/meeting/configure.zcml
```diff
@@ -34,4 +34,12 @@
       permission="zope2.View"
       />
 
+  <browser:viewlet
+      name="opengever.dossier.comment"
+      for="opengever.meeting.interfaces.IMeetingDossier"
+      manager="plone.app.layout.viewlets.interfaces.IBelowContentTitle"
+      class="opengever.dossier.viewlets.comment.BusinessCaseCommentViewlet"
+      permission="cmf.ModifyPortalContent"
+      />
```
elioschmutz: Users with only read-permission will see a broken layout if there is no comment:

bildschirmfoto 2017-03-14 um 16 25 14

elioschmutz:
Looks good and works great! Only two things i've seen while reviewing.

opengever/dossier/browser/overview.py
deiferni: using a string with one space seems a bit hackish and may break if someone decides to do a `strip()` for example. 

The default was chosen in such a way that other implementations don't have to be changed. I guess the variable-name is a bit confusing since, if set to `True` it will display a placeholder with the label "no content" (this is the current behavior and also the default), if set to `False` no such placeholder is displayed (this is new and only needed in one case).

I'll rename to `has_empty_marker` to avoid the negation in the variable name itself, i hope that helps to clarify 😉 .

opengever/meeting/configure.zcml
deiferni: ah indeed, nice catch. also they probably won't see the sticky note icon which may be confusing.

opengever/dossier/browser/overview.py
elioschmutz: you're right! 👍

PR #2730, opened 2017-03-14
Hide newest tasks box for private dossiers overview tabbedview view.

elioschmutz:
Hide newest tasks box for private dossiers overview tabbedview view because it's not possible to add tasks in private dossiers.

Before:
bildschirmfoto 2017-03-14 um 14 31 56

After:
bildschirmfoto 2017-03-14 um 14 31 29

closes #2730

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,10 @@ Changelog
 4.15.2 (unreleased)
 -------------------
 
+- Hide newest tasks box for private dossiers overview tabbedview view because
+  it's not possible to add tasks in private dossiers.
+  [elioschmutz]
+
 - Fixed role guards for dossier's offer transition.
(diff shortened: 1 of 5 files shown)


opengever/private/dossier.py
```diff
@@ -12,3 +12,6 @@ class PrivateDossier(DossierContainer):
     """A specific dossier type, only addable to a PrivateFolder in the
     private Area and only visible for the Owner and Managers.
     """
+
+    def has_task_support(self):
```
deiferni: this duplicates `allowedContentTypes` and portal type configuration, i would consider changing this to check whether task is in the dossiers `allowedContentTypes`.

elioschmutz: Thats right! I updated the function 👍

PR #2734, opened 2017-03-14
Fix quota usage recalculation to count correctly.

jone:
Fixes two problems in the quota usage recalculation method:

1. The usage was calculated for all documents on the site, but it should only be calculated for the documents within the quota container.
2. By clearing the usage map before recalculating we make sure that objects, which are no longer there, are no longer counted.


jone:
@deiferni I've added an upgrade step 🙈

PR #2744, opened 2017-03-15
Polish file action buttons

### LINKED ISSUE #2733: Allow attaching documents to email also from non-open dossiers
The templates currently wrongly use the condition of whether or not the document can be checked out or not for also deciding whether or not the file can be attached to an email client.

Rotonen:
File action buttons looking pretty depends on: 4teamwork/plonetheme.teamraum/pull/542

* Unify visible / active / disabled behaviour across file buttons
* Move logic needed by the file action buttons to the document base mixin
  * This will need review and refactoring - where should this live?
* Unify visible / active / disabled behaviour across all file manipulation templates
* Fix a few bad old tests (tooltips now actually check whether a file can be checked out or not)
* Amend OfficeConnector template tests as per file presence and per dossier workflow status

Closes #2733


deiferni:
Backref:
https://basecamp.com/2768704/projects/12554551/todos/301289539
https://basecamp.com/2768704/projects/12554551/todos/301289500
https://basecamp.com/2768704/projects/12554551/todos/301289524

Rotonen:
@deiferni Ready for rereview.

It is slightly confusing at first that if you are the one who has checked a file out, it'll still be possible to check out the file, but this is actually a convenience feature for the usability of OfficeConnector. The fix for this would be to eventually split the label for the action to separate between checkout and edit and just editing something you've already checked out, but we need to figure out conflict handling on the OfficeConnector side out better than currently (as in there should be user consent for overriding local changes from a 'new checkout action').

deiferni:
_hijacked :trollface:_

deiferni:
@Rotonen i've seen that you always introduce newlines in the import section between the `from …` and `import …` blocks. we don't do that at the moment, so for the sake of local consistency please follow our convention.

Rotonen:
@deiferni I'm using isort in my editor to warn me of wrongly sorted imports with the Plone isort config. I have not figured out how to get it to drop the newline for just a flat sort, but so far @lukasgraf was ok with it as the newline is the only difference to a flat sort.

```cat ~/.isort.cfg
[settings]
force_alphabetical_sort = True
force_single_line = True
lines_after_imports = 2
line_length = 200
not_skip = __init__.py```

I'm quite open to figuring out a suitable isort config.

deiferni:
@Rotonen i'm sure you'll figure that out somehow 😜

Rotonen:
@deiferni in this case I'll just drop my pre-commit hook for now.

phgross:
Tested locally:
 - [ ] `Metadaten bearbeiten` action is missing on a document which is checked out by another user (Tooltip).

deiferni:
@phgross as discussed, the edit metadata link is now displayed (more often) but inactive

deiferni:
@Rotonen i have amended a bit here, as discussed. if you have time to give it a quick scan again that  would be good.

Rotonen:
@deiferni don't really get why the reorder of buttons was a usability thing, but I'm not spotting anything else strange. You split the logic nicely into neat places for it to live in preparation of proper widgetization.

I'll go over the officeconnector template tests again early next week as there are still apparently cases we do not test for (checked out by me, checked out by someone else).

deiferni:
@Rotonen the motivation was to change the current button order as less as possible, and also to display the "open as pdf" stuff in similar positions (users will be affected when swapping from `opengever.pdfconverter` to `bumblebee`.

Rotonen:
@deiferni I could not parse in-process the totality of what goes where so I just tried to make it usable. Sticking to already existing ordering is sensible.

Issue #2746, opened 2017-03-16
ftw.copymovepatches Performance measurements. 


maethu:
**Test setup:**
- Local opengever.core installation
- ptftotext is installed
- 1 Dossier Containing 60 PDF's
- PDF-Size between several KB to 5MB
- At least 20 are over 1MB
- All pdf's are randomly select from my downloads folder to get a good average mix of pdfs. 
- User: philippe.gross
- All necessary objects are in the ram cache.

Measurement:
Time elapsed from cut until fully pasted. No rendering included. 

```diff 
diff --git a/opengever/dossier/move_items.py b/opengever/dossier/move_items.py
index 0459f36..3d0c849 100644
--- a/opengever/dossier/move_items.py
+++ b/opengever/dossier/move_items.py
@@ -102,8 +102,14 @@ class MoveItemsForm(form.Form):
 
                 try:
                     # Try to cut and paste object
+                    import datetime
+                    a = datetime.datetime.now()
                     clipboard = parent.manage_cutObjects(obj.id)
                     destination.manage_pasteObjects(clipboard)
+                    b = datetime.datetime.now()
+                    print "*" * 20
+                    print b-a
+                    print "*" * 20
                     copied_items += 1
 
                 except ResourceLockedError:

```
Test **with** ftw.copymovepatches:
Move whole dossier from one to another top level Ordnungsposition.
- 0:00:03.763331
- 0:00:03.158838
- 0:00:03.108819
- 0:00:03.316855

Test **without** ftw.copymovepatches:
- 0:00:08.512235
- 0:00:08.616335
- 0:00:07.890991
- 0:00:08.297138

Test **with** ftw.copymovepatches with 300 pdfs in one dossier:
Move whole dossier from one to another top level Ordnungsposition.
- 0:00:15.554797
- 0:00:15.629114
- 0:00:14.494641
- 0:00:15.954439

Test **without** ftw.copymovepatches  with 300 pdfs in one dossier:
- 0:01:26.327596
- 0:01:26.947109
- 0:01:25.216036
- 0:01:25.741105

**Conclusion**
This is needed in opengever, since it minimizes the conflict potential.
The time saving potential is better, the more files are moved.

**What the patch does**
Plone Default doest a `uncatalog` and catalog of a moved content.
[...]

PR #2747, opened 2017-03-16
Install ftw.copymovepatches for better move performance.

maethu:
Check #2746 for details / analysis


deiferni:
Hijacked  and rebased on your behalf 😉

PR #2745, opened 2017-03-15
Testing A1: Refactor sqlite testing setup

jone:
Change the setup of the sqlite database and the connection to the builder.
These changes make the sql builders actually commit instead of just flushing the session.

The ability to actually commit is important in order to move towards testing with PostgreSQL.

PR #2752, opened 2017-03-17
Testing A2: extract sqlite testing setup into separate module and layer

jone:
*~~Rebase and change target branch when #2745 (`Testing 1: Refactor sqlite testing setup.`) is merged.~~*

In preparation for moving slowly towards PostegreSQL, this change extracts the SQLite setup into a separate layer.
This is already prepared to be set up the same way as PostgreSQL will be.

Because of the order of execution, the `OpengeverFixture` now needs to be initialized with an sql fixture, which allows us to initialize the database after loading some ZCML but before setting up the Plone site.

### DIFF
--- a/opengever/core/sqlite_testing.py
+++ b/opengever/core/sqlite_testing.py
@@ -22,7 +22,7 @@ class SQLiteMemoryFixture(Layer):
     All threads use the same connection and transaction.
     """
 
-    def setup_sql_connection(self):
+    def setUp(self):
         setup_memory_database()
 
     def testSetUp(self):
--- a/opengever/core/sqlite_testing.py
+++ b/opengever/core/sqlite_testing.py
@@ -1,0 +1,99 @@
+from opengever.base import model
+from opengever.base.model import create_session
+from opengever.ogds.base.setup import create_sql_tables
+from opengever.ogds.models import BASE
+from plone.testing import Layer
+from sqlalchemy.pool import StaticPool
+from z3c.saconfig import EngineFactory
+from z3c.saconfig import GloballyScopedSession
+from z3c.saconfig.interfaces import IEngineFactory
+from z3c.saconfig.interfaces import IScopedSession
+from zope.component import provideUtility
+from zope.sqlalchemy import datamanager
+from zope.sqlalchemy import ZopeTransactionExtension
+import transaction
+
+
+class SQLiteMemoryFixture(Layer):
+    """The sqlite memory fixture provides an in-memory sqlite database
+    globally.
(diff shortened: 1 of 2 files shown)


jone:
Rebased onto `master`, changed target branch to `master`, ready for review.

opengever/core/sqlite_testing.py
```diff
@@ -0,0 +1,99 @@
...
+    # The session should be closed on test tear down.
+    scoped_session = GloballyScopedSession(
+        engine=u'opengever_db',
+        extension=ZopeTransactionExtension(keep_session=True))
+    provideUtility(
+        scoped_session, provides=IScopedSession, name=u'opengever')
+
+
+def create_tables():
+    """Creates the sql tables and makes sqlite specific configrations.
```
deiferni: configurations

opengever/core/testing.py
```diff
@@ -139,18 +105,14 @@ def setUp(self):
...
-    def testSetUp(self):
-        super(OpengeverFixture, self).testSetUp()
-        setup_sql_tables()
-
-    def testTearDown(self):
-        # Tear down the sql session because we use the keep_session flag.
-        model.Session.close_all()
-        truncate_sql_tables()
-        super(OpengeverFixture, self).testTearDown()
+    def __init__(self, sql_layer):
```
deiferni: I don't like that this introduces another way to add base layers and that we now have a mix of `Strategy` for setting up SQL and a `Layer`, but having failed to come up with a better idea 👍 😉

jone: Yes, I agree. The thing is that the layer order must be `PLONE_FIXTURE` => `BUILDER_LAYER` => `SQLITE_MEMORY_FIXTURE` => `OPENGEVER_FIXTURE`.
Since I want to make the sql fixture pluggable and replaceable, I need to set it as an additional  `bases` class.

In fact I'm using standard testing layer mechanisms for doing that. I changed the `__init__` to show that we an sql-layer is mandatory and I can make sure that the layer is correct.

In my last commit I have changed sql-layer-setup so that it is done in `setUp` rather than within the `OpengeverFixture`.
I could eliminate the `__init__` at this point but I would have to replicate the `defaultBases` classes wherever I'm instantiating `OpengeverFixture`, which like less than having a custom `__init__`.

jone:
Updated:
- rebased
- fixed typos
- slightly refactor how the sql is set up, so that the layers know less about each other

opengever/core/testing.py
deiferni: Yep, this is good enough at the moment 👍

PR #2753, opened 2017-03-17
🚧 Add PostgreSQL testing layer with basic isolation

jone:
🚧 *Rebase and change target branch when #2798 (`Isolate component registry between tests `) is merged.*

This change introduces a new PostgreSQL layer with the Matroska isolation concept adopted from the `POC layer isolation` PR #2557. But it does not yet provide any fixture as #2557 proposes; I want to do that at a later point.

This change introduces a new `OPENGEVER_FUNCTIONAL_TESTING_POSTGRES` testing layer which is not used at this point at all.

🚧 This is not yet finished.

### DIFF
--- a/opengever/core/postgresql_testing.py
+++ b/opengever/core/postgresql_testing.py
@@ -1,0 +1,199 @@
+from ftw.dictstorage.sql import DictStorageModel
+from opengever.base import model
+from opengever.ogds.models import BASE
+from plone.testing import Layer
+from sqlalchemy import create_engine
+from sqlalchemy import MetaData
+from sqlalchemy.orm import create_session
+from sqlalchemy_utils import create_database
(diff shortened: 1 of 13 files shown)


opengever/core/postgresql_testing.py
```diff
@@ -0,0 +1,194 @@
...
+            drop_database(self._disposed_engine)
+            self._disposed_engine = None
+
+        if self.parent is None:
+            # root generation
+            drop_database(self.engine.url)
+            return
+
+        if False:
+            # transaction rollback / abort
```
jone: Implement transaction rollback check using a safepoint as it is implemented in #2557.
The generation can be disposed for reusal if the transaction can sucessfully be rolled back.

PR #2754, opened 2017-03-17
Testing B1: do not tear down, isolate.

jone:
There is no need to tear down persistent objects as long as the database isolation works.

I would like to remove all teardown code because it will make the move towards PostgreSQL smoother.

### DIFF
--- a/opengever/core/testing.py
+++ b/opengever/core/testing.py
@@ -59,10 +59,6 @@ def toggle_feature(registry_interface, enabled=True):
     transaction.commit()
 
 
-def deactivate_meeting():
-    toggle_feature(IMeetingSettings, enabled=False)
-
-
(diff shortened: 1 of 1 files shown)


opengever/core/testing.py
```diff
@@ -357,9 +347,6 @@ class ActivityLayer(PloneSandboxLayer):
     def setUpPloneSite(self, portal):
         activate_activity_center()
 
-    def tearDownPloneSite(self, portal):
-        deactivate_activity_center()
```
jone: Is the `deactivate_activity_center` function still needed or can we remove it too?

jone: 👍 removed no longer used deactivate functions.

jone:
Updated: rebased onto master.

PR #2764, opened 2017-03-20
Get rid of unnecessary column label in extract attachments view.

lukasgraf:
Gets rid of a column header that isn't supposed to be shown. Apparently, in a recent version `ftw.tabbedview` falls back to the column ID if the `column_title` is falsy. 

**Before**:
extract_before

**After**:
extract_after

### DIFF
--- a/opengever/mail/browser/extract_attachments.py
+++ b/opengever/mail/browser/extract_attachments.py
@@ -10,12 +10,12 @@ from zope.component import getUtility
 import os.path
 
 
-def attachment_checkbox_helper(item, position):
+def attachment_checkbox_helper(item, value):
     attrs = {'type': 'checkbox',
              'class': 'noborder selectable',
              'name': 'attachments:list',
-             'id': 'attachment%s' % str(position),
-             'value': str(position)}
+             'id': 'attachment%s' % str(item['position']),
+             'value': str(item['position'])}
 
     return '<input %s />' % ' '.join(['%s="%s"' % (k, v)
                                       for k, v in attrs.items()])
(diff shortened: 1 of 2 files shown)


opengever/mail/browser/extract_attachments.py
```diff
@@ -86,7 +86,7 @@ def render_attachment_table(self):
 
         columns = (
             {'column': 'position',
-             'column_title': u'',
+             'column_title': u' ',
```
deiferni: As an alternative, providing an empty `column` and no `column_title` as done in the tasklistings should work, too:

https://github.com/4teamwork/opengever.core/blob/master/opengever/tabbedview/browser/tasklisting.py#L68-L72

that would make the whitespace-string "hack" unnecessary.

lukasgraf: @deiferni sounded good and so I tried that, but then the checkboxes aren't named properly anymore for the form submit, so the column *does* need an internal name at least.

deiferni: @lukasgraf sorry to nag you about this, couldn't you adapt `attachment_checkbox_helper` to do:

```python
    attrs = {'type': 'checkbox',
             'class': 'noborder selectable',
             'name': 'attachments:list',
             'id': 'attachment%s' % str(item.position),
             'value': str(item.position)}
```

as `task_id_checkbox_helper ` does: https://github.com/4teamwork/opengever.core/blob/master/opengever/tabbedview/helper.py#L31-L45

lukasgraf:
@deiferni rebased

lukasgraf:
@deiferni works as you suggested, updated

PR #2774, opened 2017-03-21
Disposition workflow guards

phgross:
Protect dispose transition with a guard, that checks if any of the dossier has to be stored in the longtime archive. Otherwise there is a new transition which allows to directly close a disposition after appraisal.

(See documentation change) for more information.

Styling separate PR plonetheme.teamraum


phgross:
@deiferni updated ...

PR #2779, opened 2017-03-22
Don't store versions for mails, ever.

deiferni:
Don't attempt to store versions for mails, mails should not have
versions ever. In some cases they do nevertheless and then iterating
over versions may fail utterly.

### DIFF
--- a/opengever/bumblebee/converter.py
+++ b/opengever/bumblebee/converter.py
@@ -1,0 +1,46 @@
+from ftw.bumblebee.converter import BumblebeeConverter
+from ftw.mail.mail import IMail
+
+
+class GeverBumblebeeConverter(BumblebeeConverter):
+
+    def _handle_store_batch(self, brains):
+        """This function stores the documents and their versions in bumblebee.
+
+        It decides if a document has to be updated  depending on its
+        conversion timestamp and updates the timestamp after storing the
+        document.
+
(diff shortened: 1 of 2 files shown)


deiferni:
manually tested in the dev environment, works as expected and stores successfully in an environment with mail versions.

opengever/bumblebee/converter.py
```diff
@@ -0,0 +1,46 @@
...
+
+            self._prepare_store_batch(obj, objs_to_store, batch_checksums)
+
+            # don't attempt to store versions for mails, mails should not have
+            # versions. in some cases they do nevertheless and then iterating
+            # over versions may fail utterly. we don't like that thus we
+            # avoid that it happens.
+            # XXX should be fixed properly by making sure mails don't have
+            # versions, ever.
+            if IMail.providedBy(obj):
```
lukasgraf: Move this up before `self._prepare_store_batch` to abort earlier?

deiferni: The first `_prepare_store_batch` stores the object itself, which is necessary also for mails since they support havin a bumblebee preview. We only want to avoid to iterate over the versions since that might 💥 .

lukasgraf: 👍

PR #2784, opened 2017-03-23
Allow commenting on tasks

### LINKED ISSUE #2666: Allow to add response without a state change to tasks
A user should be able to add a response to a task without changing the review state, for example to add a small update of the current status.

It should be visible in the process listing as all the other responses.

elioschmutz:
This PR adds the commentig feature for tasks.

screen1

related issue: https://github.com/4teamwork/plonetheme.teamraum/pull/523

closes #2666


lukasgraf:
@elioschmutz tests failing...

docs/HISTORY.txt
```diff
@@ -5,8 +5,8 @@ Changelog
 2017.1.1 (unreleased)
 ---------------------
 
-- Nothing changed yet.
-
+- Add functionality to comment a task.
+  [elioschmutz]
```
phgross: See https://github.com/4teamwork/opengever.core#updating-the-history-file

opengever/task/response.py
```diff
@@ -163,6 +164,63 @@ def can_delete_response(self):
...
+        return translate(
+            pmf('label_add_comment', default='Add Comment'),
+            context=self.request)
+
+    @button.buttonAndHandler(_(u'save', default='Save'),
+                             name='save')
+    def handleSubmit(self, action):
+        data, errors = self.extractData()
+        if errors:
+            errorMessage = '<ul>'
```
phgross: Is there a reason for handling the error message separately and manually, I guess this is handled correctly by the AddForm.

opengever/task/response.py
```diff
@@ -163,6 +164,63 @@ def can_delete_response(self):
...
+            self.status = errorMessage
+            return None
+
+        self.add_response(data)
+        self.request.RESPONSE.redirect(self.context.absolute_url())
+
+    @button.buttonAndHandler(_(u'cancel', default='Cancel'),
+                             name='cancel', )
+    def handleCancel(self, action):
+        return self.request.RESPONSE.redirect('.')
```
phgross: self.request.RESPONSE.redirect(self.context.absolute_url())

opengever/task/response.py
```diff
@@ -163,6 +164,63 @@ def can_delete_response(self):
...
+        units = ogds_service().assigned_org_units()
+        return get_current_org_unit() in units
+
+    def record_activity(self, response):
+        pass
+
+
+class TaskCommentResponseAddFormView(layout.FormWrapper, grok.View):
+    grok.context(ITask)
+    grok.name("addtaskcommentresponse")
```
phgross: I think the task in the `addtaskcommentresponse` is obsolete and on a inbox also a bit wrong, I think `addcommentresponse` is clear enough.

opengever/task/response.py
```diff
@@ -187,16 +187,28 @@ def handleSubmit(self, action):
...
     @button.buttonAndHandler(_(u'cancel', default='Cancel'),
                              name='cancel', )
     def handleCancel(self, action):
         return self.request.RESPONSE.redirect('.')
 
-    def add_response(self, data):
-        pass
+    def create_response(self, data):
+        response = Response(data.get('text'))
+        response.transition = "task-commented"
```
phgross: Define a constant `TASK_COMMENTED_TRANSITON`, or something similar with the string "task-commented" .

opengever/task/activities.py
```diff
@@ -22,6 +22,32 @@ def __init__(self, context, request, parent):
         self.parent = parent
 
 
+class TaskCommentedActivity(TaskActivity):
```
phgross: Is an separate Activity class really necessary? As far as I See the `TaskTransitionActivity` does exactly the same. Or is it more for a clear difference between those two activities?

opengever/task/tests/test_activities.py
```diff
@@ -95,6 +95,24 @@ def test_task_accepted(self, browser):
         self.assertEquals(u'Wird n\xe4chste Woche erledigt.', activity.description)
 
     @browsing
+    def test_task_commented(self, browser):
```
phgross: The notification settings for the new "activity kind" is missing. I would disable the mail notification for `comments` by default.

opengever/task/response.py
```diff
@@ -223,6 +224,22 @@ def is_user_assigned_to_current_org_unit(self):
...
+        current_related_ids = [item.to_id for item in self.context.relatedItems]
+        intids = getUtility(IIntIds)
+        for item in data.get('relatedItems', []):
+            to_id = intids.getId(item)
+
+            # relation allready exists
+            if to_id not in current_related_ids:
+                self.context.relatedItems.append(RelationValue(to_id))
+
+            response.add_change('relatedItems',
```
phgross: This adds the related_items change also when the relation has existed before ...

opengever/task/tests/test_response.py
```diff
@@ -40,6 +40,20 @@ def test_add_related_response_object_after_commenting(self, browser):
...
+        task = create(Builder('task').titled('Task 1').within(dossier))
+        document = create(Builder('document').titled('Document 1').within(dossier))
+
+        browser.login().visit(task, view="addtaskcommentresponse")
+        browser.fill({'Related Items': [document]}).find('Save').click()
+
+        self.assertEqual(
+            [document],
+            [item.to_object for item in task.relatedItems])
+
```
phgross: Also assert that the change is visible in the response...

opengever/task/comment_response.py
```diff
@@ -0,0 +1,22 @@
+from opengever.task.interfaces import ICommentResponseHandler
+from plone import api
+from zope.interface import implements
+
+
+class CommentResponseHandler(object):
```
phgross: Why not introducing a new Permission `CommentOnTask` or `AddTaskComment`. Which is managed by the Dossier workflow. So we could use plone's default workflow management instead of implementing our own permission/workflow handling ...

opengever/task/interfaces.py
```diff
@@ -147,7 +147,7 @@ def is_allowed():
         """Check if the current user is allowed to add a comment response
         """
 
-    def add_response():
+    def add_response(text):
```
phgross: Should be move to another commit 😉

docs/HISTORY.txt
elioschmutz: Done!

opengever/task/response.py
elioschmutz: The related items have been removed from the task-comments in an other release: https://github.com/4teamwork/opengever.core/pull/2784/commits/8d62408431a206be210c271a2a1ba63e77b0b6bf

opengever/task/response.py
elioschmutz: See commit: https://github.com/4teamwork/opengever.core/pull/2784/commits/a4a12027e7553d10e03cde17be0d047a36911831

opengever/task/response.py
elioschmutz: This is current implementation of the TaskTransitionResponseAddForm (see https://github.com/4teamwork/opengever.core/blob/master/opengever/task/response.py#L176)

I refactored the whole class in this commit: https://github.com/4teamwork/opengever.core/pull/2784/commits/6ac1b6cb0095d260c7c5ed988c2b11dbc58ca41d

opengever/task/response.py
elioschmutz: Fixed through commit https://github.com/4teamwork/opengever.core/pull/2784/commits/6ac1b6cb0095d260c7c5ed988c2b11dbc58ca41d

opengever/task/response.py
elioschmutz: Renamed in https://github.com/4teamwork/opengever.core/pull/2784/commits/ede566a356a7a47b42432af0128b3d403201fbc6

opengever/task/activities.py
elioschmutz: It is for clear difference because it is not a TaskTransitionActivity. We could improve it by using a better superclass (TaskActivity). But if we do this, we should refactor all the other subclasses too. I think, this should not be part of this PR

opengever/task/comment_response.py
elioschmutz: 👍  good point! I created a new permission `AddTaskComment` as you suggested. See commit https://github.com/4teamwork/opengever.core/pull/2784/commits/21cb87b5df3857331887e257f3129df24b81acad

opengever/task/tests/test_activities.py
elioschmutz: 👍  see commit https://github.com/4teamwork/opengever.core/pull/2784/commits/530bb822d8166da05b72c79b7a9cfe228d23389c

opengever/task/tests/test_response.py
elioschmutz: Related Items are no longer provided by task comments.

opengever/task/interfaces.py
elioschmutz: 👍

elioschmutz:
@phgross pls take another look!

opengever/task/comment_response.py
```diff
@@ -0,0 +1,46 @@
...
+    def _add_response_to_obj(self, obj, response):
+        container = IResponseContainer(obj)
+        container.add(response)
+
+        notify(ObjectModifiedEvent(obj))
+
+    def _record_activity(self, response):
+        TaskCommentedActivity(self.context,
+                              self.context.REQUEST,
+                              None,
```
deiferni: The class seems to require a parameter, which is always `None`? I'd remove the parameter, if it is required by a super-class define it in `TaskCommentedActivity`s `__init__`.

opengever/task/response.py
```diff
@@ -142,11 +153,55 @@ def can_delete_response(self):
...
+
+    fields = field.Fields(ITaskCommentResponseFormSchema)
+
+    @property
+    def label(self):
+        return translate(
+            pmf('label_add_comment', default='Add Comment'),
+            context=self.request)
+
+    @button.buttonAndHandler(_(u'save', default='Save'),
```
deiferni: i think this would fit on one line 😉

opengever/task/response.py
```diff
@@ -142,11 +153,55 @@ def can_delete_response(self):
...
+        return super(TaskCommentResponseAddForm, self).handleAdd(self, action)
+
+    def createAndAdd(self, data):
+        response_handler = ICommentResponseHandler(self.context)
+        return response_handler.add_response(data.get('text'))
+
+    def nextURL(self):
+        return self.context.absolute_url()
+
+    @button.buttonAndHandler(_(u'cancel', default='Cancel'),
```
deiferni: i think this would fit on one line 😉 , also there is a terinating comma

opengever/task/response.py
```diff
@@ -142,11 +153,55 @@ def can_delete_response(self):
...
+
+class TaskCommentResponseAddFormView(layout.FormWrapper, grok.View):
+    grok.context(ITask)
+    grok.name("addcommentresponse")
+    grok.require('opengever.task.AddTaskComment')
+
+    form = TaskCommentResponseAddForm
+
+    def __call__(self):
+        if not ICommentResponseHandler(self.context).is_allowed():
```
deiferni: Shouldn't this already be covered by the permission?

opengever/task/comment_response.py
elioschmutz: This would be against substitution principles...

deiferni: as just discussed, the hierarchy is off a bit and we'll adapt, so this comment is half-void 😉

deiferni:
Backref: https://basecamp.com/2768704/projects/12554551/todos/300701856

docs/HISTORY.txt
```diff
@@ -6,9 +6,9 @@ Changelog
 ---------------------
 
 - Enable keywordswidget for IDocumentMetadata behavior.  [elioschmutz]
+- Add functionality to comment a task. [elioschmutz]
 - Refactor: Use SQLFormSupport class for: Meeting and Member [elioschmutz]
 
-
```
phgross: 😏  ?

opengever/activity/upgrades/20170328142858_add_task_commented_notification_setting/upgrade.py
```diff
@@ -0,0 +1,28 @@
+from opengever.activity.model import NotificationDefault
```
phgross: You have to create your own table definition in schema migrations, instead of importing them. Otherwise there is potential for 💥 .

Examples https://github.com/4teamwork/opengever.core/blob/master/opengever/meeting/upgrades/20160218114243_add_title_column_to_meeting/upgrade.py#L9

docs/HISTORY.txt
elioschmutz: 😎

elioschmutz:
😅  ok! done! @phgross pls take another look to the history and the activity upgradestep.

phgross:
@elioschmutz Changes looks 👍 . Furthermore I've tested the new feature locally works fine on a single admin_unit installation.

But on a multi_admin_unit setup the comment responses are not synched to the foreign admin_unit. See 
bildschirmfoto 2017-04-05 um 11 39 17

If it's easier for you, to first 🚢  this PR and solve the multi-adminunit issue in an separate one, that would be ok for me.

PR #2787, opened 2017-03-24
Make sure qtip tooltips are destroyed on hide.

### LINKED ISSUE #2786: Make sure qtip tooltips are destroyed on hide
The OfficeConnector JS cannot distinguish between documents on the same DOM so it always just finds the first relevant data element. We need to have an explicit destroy on IE11 in order to achieve this.

Rotonen:
* Make sure qtip tooltips are removed from the DOM on hide on IE11.

Fixes #2786

### DIFF
--- a/opengever/base/browser/resources/tooltip.js
+++ b/opengever/base/browser/resources/tooltip.js
@@ -29,7 +29,8 @@
       fixed: true // Make sure the tooltip gets not hidden if mouse is over the tooltip
     },
     events: {
-      show: closeTooltips
+      show: closeTooltips,
+      hide: destroyTooltips,
     }
   };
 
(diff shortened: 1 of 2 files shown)


opengever/base/browser/resources/tooltip.js
```diff
@@ -29,7 +29,10 @@
       fixed: true // Make sure the tooltip gets not hidden if mouse is over the tooltip
     },
     events: {
-      show: closeTooltips
+      show: closeTooltips,
+      hide: function(event, api) {
```
bierik: How about a named separate function to keep the settings object clean:

```javascript
events: {
  show: closeTooltips,
  hide: destroyTooltips
}

...

function destroyTooltips(event, api) { api.destroy(true); }
```

Rotonen: Will do.

Rotonen:
@bierik ready for rereview.

PR #2799, opened 2017-03-28
i18n-build tripwire

Rotonen:
* Depends on `pcregrep`


jone:
I think the multiline message string has a valid format (emacs generates it like that, for example). So I think we need to support it 🤔

Rotonen:
https://www.gnu.org/software/gettext/manual/html_node/Normalizing.html

For now.

`msgstr ""\n\n` is the thing to match for to catch a missing translation.

jone:
Fair enough 👍 
Why is it currently failing? There is no list of missing translations, did you remove that?

Rotonen:
`pcregrep` (from `pcer-utils`) is not available on Centos 6. Will be available on Centos 7.

jone:
Ok. I think we should wait for jenkins2 then.

Rotonen:
Yep. Before then we need to make the translations complete so the tests will actually pass on that.

Rotonen:
https://ci.4teamwork.ch/builds/71915/tasks/103213#bottom

Here is a purposeful trip up for visualization purposes to aid discussion.

Rotonen:
Also FWIW `pcregrep` would be available in both Centos 5 and Centos 6, it's just under the `pcre` package - RHEL 7 split it off into `pcre-tools`.

opengever/advancedsearch/locales/de/LC_MESSAGES/opengever.advancedsearch.po
```diff
@@ -72,7 +72,7 @@ msgstr "Frist"
 
 #: ./opengever/advancedsearch/advanced_search.py:242
 msgid "label_deadline_2"
-msgstr ""
+msgstr "this seems to be a legit find"
```
deiferni: I think this is a special case where we don't want to have a label. This seems to be a date-range form that is styled to render two fields on the same level.
screen shot 2017-03-31 at 16 34 49
I'd remove the title `label_deadline_2` form the form field.

opengever/dossier/locales/de/LC_MESSAGES/opengever.dossier.po
```diff
@@ -438,7 +435,7 @@ msgstr "Es wurde ein Kommentar erfasst!"
 #. Default: "View note"
 #: ./opengever/dossier/viewlets/templates/note.pt:28
 msgid "help_text_view_note"
-msgstr ""
+msgstr "this seems to be a legit find"
```
deiferni: Kommentar anzeigen.

opengever/officeatwork/locales/de/LC_MESSAGES/opengever.officeatwork.po
```diff
@@ -24,5 +21,4 @@ msgstr "Mit officeatwork erstellen"
 
 #: ./opengever/officeatwork/browser/documentfromofficeatwork.py:54
 msgid "Creation with officeatwork initiated successfully"
-msgstr ""
-
+msgstr "this seems to be a legit find"
```
deiferni: Inhaltserstellung mit officeatwork erfolgreich gestartet

opengever/officeconnector/locales/de/LC_MESSAGES/opengever.officeconnector.po
```diff
@@ -9,28 +9,24 @@ msgstr ""
...
-"Language-Code: en\n"
-"Language-Name: English\n"
 "Preferred-Encodings: utf-8 latin1\n"
-"Domain: DOMAIN\n"
 
 #. Default: "OfficeConnector Outlook support"
 #: ./opengever/officeconnector/interfaces.py:9
 msgid "label_enable_officeconnector_attach_to_outlook"
-msgstr ""
+msgstr "now we also have to translate all admin panel stuff?"
```
deiferni: I'd remove using a messagefactory from the registry setting interface. As far as i've seen none of our registry interface field descriptions are translated.

test-i18n-de.cfg
```diff
@@ -0,0 +1,62 @@
...
+    i18n-build
+    bin-test-jenkins
+
+package-name = opengever.core
+package-namespace = opengever
+
+jenkins_python = $PYTHON27
+
+[versions]
+zope.security = 4.0.0
```
deiferni: could you add a comment why this pinning is necessary? or even better, find a way to remove it 😉 .

test-i18n-de.cfg
```diff
@@ -0,0 +1,62 @@
...
+cmds=cp i18n-build.in bin/i18n-build && chmod +x bin/i18n-build
+
+[bin-test-jenkins]
+recipe = collective.recipe.template
+input = inline:
+    #!/bin/bash
+    # Build locales
+    bin/i18n-build &> /dev/null
+    OUTPUT=""
+
```
deiferni: `set -euo pipefail`? or do you need different behavior in your script?

test-i18n-de.cfg
```diff
@@ -0,0 +1,62 @@
...
+        OUTPUT+=$(echo "" && echo "" && pcregrep -n -H -M 'msgstr ""\n\n' "$file")
+    done
+
+    if [ -z "$OUTPUT" ]
+    then
+        exit 0
+    fi
+
+    # Newlines for visual clarity
+    echo ""
```
deiferni: would `printf` be an option/alternative?

test-i18n-de.cfg
Rotonen: `pcregrep` returns 1 if nothing matches.

test-i18n-de.cfg
Rotonen: The newline escapes get relatively messy, IMO.

test-i18n-de.cfg
Rotonen: `i18ndude` does not install without this pinning due to `zope.security` being pinned here:
https://github.com/4teamwork/ftw-buildouts/blob/master/test-base.cfg#L356

I also tried to pin `i18ndude` as done here:
https://github.com/4teamwork/opengever.core/blob/master/versions.cfg#L11

Feedback is quite welcome in how to get rid of the pinning or why the pinning is necessary - I have no idea at this point.

Rotonen: Also you cannot uninstall instance if you just build this out. You need a fresh clone.

Documenting for convenience of testing:
`git clone --branch i18n-build-tripwire --depth 1 git@github.com:4teamwork/opengever.core.git`

deiferni: can you extend from `versions.cfg` after extending from ` https://raw.githubusercontent.com/4teamwork/ftw-buildouts/master/test-base.cfg` to include `develop` pinnings? i think they should work.

test-i18n-de.cfg
deiferni: fair enough 👍

test-i18n-de.cfg
deiferni: fair enough 👍

Rotonen:
@deiferni ready for rereview.

opengever/document/locales/fr/LC_MESSAGES/opengever.document.po
```diff
@@ -106,6 +106,10 @@ msgstr "Quitter GEVER"
 msgid "Reverted file to version ${version_id}"
 msgstr "Retour à la version ${version_id}  du document"
 
+#: ./opengever/document/forms.py:63
+msgid "Save"
+msgstr ""
```
deiferni: You seem to have lost a french translation here, i think that one should be `Sauvegarder` according to other translations. Even though the french translations cannot be guaranteed to be provided we should attempt to keep the ones we had 😉 .

opengever/document/locales/fr/LC_MESSAGES/opengever.document.po
```diff
@@ -413,17 +421,22 @@ msgstr "Télécharger une copie"
...
 #. Default: "Foreign Reference"
-#: ./opengever/document/behaviors/metadata.py:79
+#: ./opengever/document/behaviors/metadata.py:83
 msgid "label_foreign_reference"
 msgstr "Code externe"
 
+#. Default: "Initial version (document copied)"
+#: ./opengever/document/subscribers.py:28
+msgid "label_initial_version_copied"
+msgstr ""
```
deiferni: same here, we had this in french already

opengever/document/locales/fr/LC_MESSAGES/opengever.document.po
```diff
@@ -242,36 +246,40 @@ msgstr "Nom, prénom ou acronyme de l'utilisateur (sera automatiquement créé d
...
 
-#: ./opengever/document/behaviors/metadata.py:80
+#: ./opengever/document/behaviors/metadata.py:84
 msgid "help_foreign_reference"
 msgstr "Référence sur le dossier correspondant de l'expéditeur"
 
-#: ./opengever/document/behaviors/metadata.py:135
+#: ./opengever/document/behaviors/metadata.py:73
+msgid "help_keywords"
+msgstr ""
```
deiferni: same here

PR #2801, opened 2017-03-29
Fix max-height of tree portlet.

bierik:
Closes https://github.com/4teamwork/opengever.core/issues/2722

Depends on https://github.com/4teamwork/plonetheme.teamraum/pull/525

The height calculation using javascript on every pageresize is very
expensive and buggy. So use the native CSS `calc` function and relative `vh` unit
to limit the height of the tree portlet.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2017.1.2 (unreleased)
 ---------------------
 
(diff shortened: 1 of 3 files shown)


jone:
- Does all supported browsers support these functions?
- Can you add a screenshot?
- Could you add a changelog entry?

bierik:
@deiferni rebased

bierik:
@deiferni updated... sry I forgot to squash

PR #2828, opened 2017-04-06
Implement task comment syncing between admin units

### LINKED ISSUE #2822: Implement Task-Comment Syncing between Admin-Units
On a multi_admin_unit setup the comment responses are not synched to the foreign admin_unit. 

8845cce6-19f4-11e7-8e5c-9d49b2fa4665

Depends on the functionality implemented in #2666

elioschmutz:
This PR adds the functionality to sync the task-response-comments between tasks.

screen1

TODO (I will do this in the same PR in a new commit as soon as this PR has been reviewed):
- [x] Refactoring the`WorkflowStateSyncer` to use the new `BaseResponseSyncerSender`

closes #2822


opengever/task/tests/test_syncer.py
```diff
@@ -0,0 +1,283 @@
...
+        return self
+
+    def read(self):
+        return self.response
+
+
+class TestBaseResponseSyncerSender(FunctionalTestCase):
+
+    def _mock_dispatch_request(self, sender, response):
+        mock = MockDispatchRequest(response=response)
```
phgross: A cleaner way to patch the _dispatch_request method would be to create a class which inherits from BaseResponseSyncerSender, for example.

```
class TestBaseResponseSyncerSender(BaseResponseSyncerSender):
   
  requests = []

  def _dispatch_request(self, target_admin_unit_id, viewname, path, data):
        self.requests.append({
             'target_admin_unit_id': target_admin_unit_id,
            'viewname': viewname,
            'path': path,
            'data': data,
        })

```

opengever/task/syncer.py
```diff
@@ -0,0 +1,150 @@
...
+        self.request = request
+
+    def sync_related_tasks(self, transition, text, **kwargs):
+        tasks = self.get_related_tasks_to_sync()
+        for task in tasks:
+            self.sync_related_task(task, transition, text, **kwargs)
+
+        return tasks
+
+    def get_related_tasks_to_sync(self):
```
elioschmutz: 👍  I updated the docstring in the Interface.

opengever/task/tests/test_syncer.py
```diff
@@ -0,0 +1,283 @@
...
+                 'viewname': 'SYNC_TASK'},
+                {'data': {'text': 't\xc3\xa4st', 'transition': 'test-transition'},
+                 'path': u'task-3',
+                 'target_admin_unit_id': u'client1',
+                 'viewname': 'SYNC_TASK'}
+            ],
+            mock_request.requests
+        )
+
+    def test_sync_related_task_raises_an_exception_if_syncing_failed(self):
```
elioschmutz: True, but this are different methods...

- sync_related_task
- sync_related_tasks

So i have to test them in a separate test.

opengever/task/tests/test_syncer.py
elioschmutz: The current implementation is better for the specific use-case for the following reasons:

1. I mock only the foreign method which is not a part of the class I do test
2. The method needs to return the result of the request. Depending on the result, the tested functions behave differently. With the current implementation I can define my desired return value for each testcase and can test against it
3. I do not mock away an internal method (`_dispatch_request`)

So I prefer to not change it.

opengever/task/interfaces.py
```diff
@@ -96,6 +96,41 @@ def change_remote_tasks_workflow_state(transition, text):
         """
 
 
+class IResponseSyncerSender(Interface):
```
elioschmutz: This is the interface implemented by the `BaseResponseSyncerSender` which sends in its most basic implementation a task-response to a `BaseResponseSyncerReceiver`. So in general, its not a good idea to name it more general if the class is doing some specific tasks.

There are more specific interfaces like the `ICommentResponseSyncerSender` wich is implemented by the `CommentResponseSyncerSender` class.

Also the `WorkflowStateSyncer` is doing the same task as the `BaseResponseSyncerSender` in a more specific way. It sends a task-response to a receiver but includes a workflow change. It is called by a class named `TaskTransitionResponseAddForm`.

The same happens with the `DeadlineModifier` or `RemoteDeadlineModifier`. This is again an own implementation of sending responses to a receiver but in addition it changes the `deadline` of a task.

As I already written down in the PR-Intro message, I will refactor the `WorkflowStateSyncer` and the `DeadlineModifier` to use the new `BaseResponseSyncerSender` for sending task responses.

elioschmutz:
Ready to review the refactoring-commits

opengever/task/syncer.py
```diff
@@ -148,3 +178,43 @@ class CommentResponseSyncerReceiver(BaseResponseSyncerReceiver):
...
+        before = wftool.getInfoFor(self.context, 'review_state')
+        before = wftool.getTitleForStateOnType(before, self.context.Type())
+
+        wftool.doActionFor(self.context, transition)
+
+        after = wftool.getInfoFor(self.context, 'review_state')
+        after = wftool.getTitleForStateOnType(after, self.context.Type())
+
+        if responsible and responsible is not 'None':
+            # special handling for reassign
```
phgross: please move the special handling for reassing to a method called `updated_responsible`

opengever/task/tests/test_statesyncer.py
```diff
@@ -1,153 +0,0 @@
...
-        self.prepare_request(task, text=u'I am done!',
-                             transition='task-transition-reassign',
-                             responsible='hugo.boss',
-                             responsible_client='afi')
-        task.unrestrictedTraverse(remote_viewname)()
-
-        self.assertEquals('afi', task.responsible_client)
-        self.assertEquals('hugo.boss', task.responsible)
-
-    def test_adds_corresponding_response(self):
```
phgross: I think you dropped the Response checks, I can't find them anywhere else ...

opengever/task/tests/test_deadline_modification.py
```diff
@@ -16,9 +17,26 @@
 
 class TestDeadlineModificationForm(FunctionalTestCase):
 
+    def tearDown(self):
+        super(TestDeadlineModificationForm, self).tearDown()
+        self.enable_IInternalOpengeverRequestLayer()
+
+    def enable_IInternalOpengeverRequestLayer(self):
```
phgross: The `enable_IInternalOpengeverRequestLayer` and `disable_IInternalOpengeverRequestLayer` are confusing, please implement them the same way than in the previous commit (Workflowsyncer)

opengever/task/response_syncer/__init__.py
```diff
@@ -26,9 +19,20 @@ class ResponseSyncerSenderException(Exception):
     """
 
 
-def sync_task_response(context, request, syncer_name, transition, text, **kwargs):
+def sync_task_response(task, request, syncer_name, transition, text, **kwargs):
+    """Method to get the responsible IResponseSyncerSender adapter
+    and run the response syncing process for predecessors and successorts
```
phgross: successor*t*s

opengever/task/response_syncer/__init__.py
```diff
@@ -26,9 +19,20 @@ class ResponseSyncerSenderException(Exception):
...
+    of the current task.
+
+    :param task: An object providing the ITask interface
+    :param request: The current request
+    :param syncer_name: A name of a IResponseSyncerSender adapter name
+    :param transition: The transition-name for the task-response obj
```
phgross: The transition which has been done and should be synced now.

opengever/task/response_syncer/__init__.py
```diff
@@ -26,9 +19,20 @@ class ResponseSyncerSenderException(Exception):
...
+    :param text: The text for the task-response obj
```
phgross: The response text.

opengever/task/tests/test_deadline_modification.py
elioschmutz: I adjusted the implementation as it is in: https://github.com/4teamwork/opengever.core/blob/master/opengever/task/tests/test_assign.py#L87

Unfortunately I cannot use the Workflowsyncer implementation with browser-tests because the testbrowser generates a new request-obj on each page-request.

elioschmutz:
@phgross pls take another look!

opengever/task/tests/test_response_syncer.py
```diff
@@ -475,21 +475,21 @@ def test_sync_state_change_on_predecessor_to_successor(self):
                           api.content.get_state(successor))
 
     def test_adds_corresponding_response(self):
-        predecessor = create(Builder('task').in_state('task-state-in-progress'))
+        predecessor = create(Builder('task').in_state('task-state-resolved'))
```
phgross: Why do you have to change that test? The response syncher should still be able for the `task-state-resolved`

elioschmutz: The current implementation disallows syncing `task-state-in-progress-resolved` transitions.

See https://github.com/4teamwork/opengever.core/blob/master/opengever/task/statesyncer.py#L75

The current implemented test is only passing because it calls the receiver-view directly (https://github.com/4teamwork/opengever.core/blob/master/opengever/task/tests/test_statesyncer.py#L144) instead through the `WorkflowStateSyncer` which does the check (see https://github.com/4teamwork/opengever.core/blob/master/opengever/task/statesyncer.py#L22)

I think this is an issue. @phgross do you agree, that this is an issue and the syncer should allow `task-state-in-progress-resolved` transitions (https://github.com/4teamwork/opengever.core/blob/es-2822-implement-task-comment-syncing-between-admin-units/opengever/task/response_syncer/workflow.py#L27)?

elioschmutz:
@phgross tests are passing...

PR #2832, opened 2017-04-07
Meeting: add a "Word implementation" feature flag.

jone:
Part of #2829

The new feature flag "Word implementation" is used for switching the meeting implemetation to the new Word implementation.

The new word implementation allows to write protocols directly in word.
In order to develop it in parallel it needs to be feature flag.

When enabling the Word implementation feature flag no meeting objects
should exist, otherwise they migth lack a proper migration at this point.

### DIFF
--- a/opengever/meeting/__init__.py
+++ b/opengever/meeting/__init__.py
@@ -1,4 +1,5 @@
 from opengever.meeting.interfaces import IMeetingSettings
+from plone import api
 from plone.registry.interfaces import IRegistry
 from zope.component import getUtility
 from zope.i18nmessageid import MessageFactory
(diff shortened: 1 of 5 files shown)


opengever/meeting/__init__.py
```diff
@@ -14,3 +14,28 @@ def is_meeting_feature_enabled():
...
+
+    Switching from old to new by activating this flag should only happen when
+    there are no meeting objects, since they are currently not migrated
+    automatically.
+    """
+    if not is_meeting_feature_enabled():
+        return False
+
+    try:
+        registry = getUtility(IRegistry)
```
deiferni: could switch to `plone.api` here if you want

jone: 👍 `plone.api` helps to handle the "missing" use case in one line.

jone:
Updated and rebased.

PR #2835, opened 2017-04-07
Meeting: Tab für Antragsvorlagen in Vorlageordner

jone:
Im Vorlageordner wird ein neues Tab "Antragsvorlagen" hinzugefügt, welches die Antragsvorlagen (#2834) in diesem Vorlageordner anzeigen.
Das Tab verhält sich genau gleich wie das Tab für Sablon-Vorlagen: gewisse Spalten und Aktionen des normalen Dokumente-Tabs stehen werden entfernt.

Dies ist Teil des Auftrags #2829.

bildschirmfoto 2017-04-07 um 13 58 17

### DIFF
--- a/opengever/dossier/browser/tabbed.py
+++ b/opengever/dossier/browser/tabbed.py
@@ -2,6 +2,7 @@ from opengever.contact import is_contact_feature_enabled
 from opengever.dossier import _
 from opengever.dossier.dossiertemplate import is_dossier_template_feature_enabled
 from opengever.meeting import is_meeting_feature_enabled
+from opengever.meeting import is_word_meeting_implementation_enabled
 from opengever.tabbedview import GeverTabbedView
 
 
@@ -102,6 +103,14 @@ class TemplateFolderTabbedView(GeverTabbedView):
                     }
 
     @property
+    def proposal_templates_tab(self):
+        if is_word_meeting_implementation_enabled():
+            return {'id': 'proposaltemplates',
+                    'title': _(u'label_proposal_templates',
+                               default=u'Proposal Templates'),
+                    }
+
+    @property
     def dossiertemplate_tab(self):
         if is_dossier_template_feature_enabled():
             return {'id': 'dossiertemplates',
@@ -114,6 +123,7 @@ class TemplateFolderTabbedView(GeverTabbedView):
             self.template_tab,
             self.dossiertemplate_tab,
             self.sablon_tab,
+            self.proposal_templates_tab,
             self.tasktemplate_folders_tab
         ]
 
--- a/opengever/dossier/browser/tabbed.py
+++ b/opengever/dossier/browser/tabbed.py
@@ -42,43 +42,41 @@ class DossierTabbedView(GeverTabbedView):
     def subdossiers_tab(self):
         if self.context.show_subdossier:
             return {'id': 'subdossiers',
-                    'title': _(u'label_subdossiers', default=u'Subdossiers'),
-                    }
-
-        return None
+                    'title': _(u'label_subdossiers', default=u'Subdossiers')}
+        else:
+            return None
 
     @property
     def proposals_tab(self):
         if is_meeting_feature_enabled():
             return {'id': 'proposals',
-                    'title': _(u'label_proposals', default=u'Proposals'),
-                    }
-
-        return None
+                    'title': _(u'label_proposals', default=u'Proposals')}
+        else:
+            return None
 
     @property
     def participations_tab(self):
         if is_contact_feature_enabled():
             return {
                 'id': 'participations',
-                'title': _(u'label_participations', default=u'Participations'),
-                }
-
-        return {
-            'id': 'participants',
-            'title': _(u'label_participants', default=u'Participants'),
-            }
+                'title': _(u'label_participations', default=u'Participations')}
+        else:
+            return {
+                'id': 'participants',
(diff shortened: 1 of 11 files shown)


opengever/dossier/tests/test_templatefolder.py
```diff
@@ -569,6 +570,17 @@ def test_addable_types_with_meating_feature(self, browser):
              'Template Folder'],
             factoriesmenu.addable_types())
 
+    @browsing
+    def test_addable_types_with_meating_word_implementation(self, browser):
```
deiferni: meating 🍖  🍗  😄

opengever/dossier/browser/tabbed.py
```diff
@@ -102,6 +103,14 @@ def sablon_tab(self):
                     }
 
     @property
+    def proposal_templates_tab(self):
+        if is_word_meeting_implementation_enabled():
+            return {'id': 'proposaltemplates',
+                    'title': _(u'label_proposal_templates',
+                               default=u'Proposal Templates'),
+                    }
+
```
jone: @deiferni I prefer local consistency: there are already two properties looking exactly the same.
Since `None` is the default when there is no return statement I only would find it good to add a `return None` when it is combined with an abort condition:
```python
def bla():
    if bad:
        return None
    return 'good stuff'
```
I find putting a `return None` on the last line of a function only creates noise and does not really help.

opengever/dossier/tests/test_templatefolder.py
jone: I was probably hungry 😂

opengever/dossier/browser/tabbed.py
deiferni: now we may enter the domain of personal preference 😉 , but let me attempt to convince you:

i usually use an explicit `return None` to indicate that this return value that is relevant for the caller and shall be processes as `None`. this also makes clear at once that this is a valid code-path. in case of this method one might argue that it is simple enough and obvious, but while that is true now the method may grow and be rewritten.

i also find the guard-clause style with early returns good pratice since that distinguishes clearly between the usually more likely case of `good stuff` (not indented) and `bad` (indented). 

i'd suggest you change the method to:

```python

    def proposal_templates_tab(self):
        if not is_word_meeting_implementation_enabled():
            return None
       return {'id': 'proposaltemplates',
               'title': _(u'label_proposal_templates',
                          default=u'Proposal Templates'),
               }
```

for the sake of local consistency, which is also very dear to me, i'd suggest you follow the boy Scout Rule 😉  and change all the methods.

deiferni: (oh, i just found this nice SO answer http://stackoverflow.com/questions/15300550/python-return-return-none-and-no-return-at-all)

jone: Ok, I'm good with that, I have refactored the conditional tabs.
I used this style:
```python
    @property
    def proposal_templates_tab(self):
            return {'id': 'proposaltemplates',
                    'title': _(u'label_proposal_templates',
                               default=u'Proposal Templates')}
        else:
            return None
```
- I'm fine with that as long as an exiplicit if/else statement is used.
- Because the argument is that we check for `None` I have added filters which treat the `None` values.

jone:
Rebased, changed base to `master`.

jone:
@phgross kannst du dies anschauen? Die Tests passen jetzt.

opengever/dossier/locales/de/LC_MESSAGES/opengever.dossier.po
```diff
@@ -864,11 +869,12 @@ msgid "time_expired"
...
-#: ./opengever/dossier/resolve.py:192
+#: ./opengever/dossier/resolve.py:190
 msgid "title_dossier_journal"
 msgstr "Dossier Journal ${title}, ${timestamp}"
 
 #. Default: "You didn't select any participants."
 #: ./opengever/dossier/behaviors/participation.py:196
 msgid "warning_no_participants_selected"
-msgstr "Sie haben keine Beteiligungen ausgewählt."
+msgstr "Sie haben keine Beteiligungen ausgewählt.Participantslabel_participantsBeteiligungen"
```
phgross: ✂️  .Participantslabel_participantsBeteiligungen

jone: Opps, thanks 👍

jone:
@phgross I double checked that the grid state storage works for the proposal template tabs on the templates folder. Be aware that I already have rebased on to Elios changes.

opengever/dossier/locales/fr/LC_MESSAGES/opengever.dossier.po
```diff
@@ -860,12 +865,12 @@ msgid "time_expired"
...
-#: ./opengever/dossier/resolve.py:192
+#: ./opengever/dossier/resolve.py:190
 msgid "title_dossier_journal"
 msgstr ""
 
 #. Default: "You didn't select any participants."
 #: ./opengever/dossier/behaviors/participation.py:196
 msgid "warning_no_participants_selected"
-msgstr "Vous n'avez sélectionné aucune participation"
+msgstr "Vous n'avez sélectionné aucune participationParticipantslabel_participantsParticipants"
```
phgross: 🤔

jone:
@phgross It might be ok now 🙄

PR #2837, opened 2017-04-10
Convert existing keywords to utf-8 first before upgrading.

deiferni:
This PR ensures that keywords are enocded as `utf-8` before being reindexed.

At the Moment `ftw.keywordwidget` stores the keywords as `utf-8`, this seems to be required by `plone.app.vocabularies.Keywords`. Previous document keywords are stored as `unicode`, however. They keywords-vocabulary is global and shared by documents and dossiers. Thus when executing this upgrade step there will be issues with already stored dossier keywords when reindexing as you cannot mix `utf-8` and `unicode` in a `KeywordIndex`.

### DIFF
--- a/opengever/document/upgrades/20170321120536_enable_ftw_keywordwidget_for_document_metadata_behavior/upgrade.py
+++ b/opengever/document/upgrades/20170321120536_enable_ftw_keywordwidget_for_document_metadata_behavior/upgrade.py
@@ -1,13 +1,22 @@
 from ftw.upgrade import UpgradeStep
 from opengever.document.behaviors.metadata import IDocumentMetadata
+from Products.CMFPlone.utils import safe_unicode
 
 
 class EnableFtwKeywordwidgetForDocumentMetadataBehavior(UpgradeStep):
     """Enable ftw keywordwidget for document metadata behavior.
     """
 
+    query = {'object_provides': [IDocumentMetadata.__identifier__]}
+
     def __call__(self):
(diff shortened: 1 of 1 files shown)


opengever/document/upgrades/20170321120536_enable_ftw_keywordwidget_for_document_metadata_behavior/upgrade.py
```diff
@@ -1,13 +1,22 @@
 from ftw.upgrade import UpgradeStep
 from opengever.document.behaviors.metadata import IDocumentMetadata
+from Products.CMFPlone.utils import safe_unicode
 
 
 class EnableFtwKeywordwidgetForDocumentMetadataBehavior(UpgradeStep):
     """Enable ftw keywordwidget for document metadata behavior.
     """
 
+    query = {'object_provides': [IDocumentMetadata.__identifier__]}
```
phgross: @deiferni are only Documents affected by this problems or dossiers as well ?

deiferni: @phgross only documents are affected in this upgrade since we reindex documents only. Already looking into dossiers 😉 , fix will come in a separate PR, if necessary.

phgross: 👍

PR #2851, opened 2017-04-12
No longer show export as pdf actions on private dossiers.

### LINKED ISSUE #2810: Remove actions to print cover/details for private dossier
The action to print a dossier cover and dossier details should not be displayed for a private dossier. 
They "don't work" and they should not be available for private dossiers in any case.

screen shot 2017-04-03 at 15 48 53

Backref: https://basecamp.com/2768704/projects/12554551/todos/302471085

deiferni:
screen shot 2017-04-12 at 15 31 04

Closes #2810.

### DIFF
--- a/opengever/latex/upgrades/20170412152602_don_t_show_pdf_actions_on_private_dossier/actions.xml
+++ b/opengever/latex/upgrades/20170412152602_don_t_show_pdf_actions_on_private_dossier/actions.xml
@@ -1,0 +1,15 @@
+<object xmlns:i18n="http://xml.zope.org/namespaces/i18n" name="portal_actions" meta_type="Plone Actions Tool">
+  <object name="object_buttons" meta_type="CMF Action Category">
+
+    <object name="pdf_dossierdetails" meta_type="CMF Action" i18n:domain="opengever.latex">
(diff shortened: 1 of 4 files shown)


opengever/latex/upgrades/20170412152602_don_t_show_pdf_actions_on_private_dossier/actions.xml
```diff
@@ -0,0 +1,33 @@
+<object xmlns:i18n="http://xml.zope.org/namespaces/i18n" name="portal_actions" meta_type="Plone Actions Tool">
+
+  <object name="object_buttons" meta_type="CMF Action Category">
+
+    <object name="pdf_dossierdetails" meta_type="CMF Action" i18n:domain="opengever.latex">
+      <property name="title" i18n:translate="">Print details (PDF)</property>
```
phgross: I would only add the changing properties in the upgradestep, its easier to read/understand

deiferni: That works as well, updated.

PR #2862, opened 2017-04-20
Fix sortorder of repository favorites in the treeportlet.

### LINKED ISSUE #2853: Sorting of repository favorites in the treeportlet incorrect.
screenshot 4

- [x] fix sort order of favorites treeportlet

Extranet: https://extranet.4teamwork.ch/extern/opengever-kanton-zug/sprint-backlog/20344

elioschmutz:
Before:

bildschirmfoto 2017-04-20 um 12 16 50

After: 

bildschirmfoto 2017-04-20 um 12 14 54

closes #2853

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -12,6 +12,7 @@ Changelog
   numbers in order to avoid conflict errors. [lgraf]
 - Bundle import: Don't commit "before post-processing". [lgraf]
 - Allow get prefills for the CompleteSuccessorTaskForm. [elioschmutz]
+- Fix sortorder of repository favorites in the treeportlet.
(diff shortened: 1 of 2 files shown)


elioschmutz:
@phgross my implementation works with IE >= 11

Do we need to provide a solution for IE < 11?

phgross:
@elioschmutz We don't need to provide a solution for IE 10, but the fix should not break the javascript completly on IE 10. Means the treeportlet should still work in IE 10 but it doesn't matter if the favorites aren't sorted correctly.

elioschmutz:
@phgross The `localeCompare` method is available in IE10 but it does not provide the `numeric` property. So, there is no problem using this method in IE10. It just sorts the entries as before this fix.

PR #2865, opened 2017-04-20
Implement more generic gridstatestorage adapter

elioschmutz:
- [x] Release and pin ftw.tabbedview kgs

This PR implements a more generic gridstatestorage adapter. This is necessary because we need some functionality for other types. To prevent code-duplication we made it more generic.

For what do we need this PR explicitly?
The current implementation was for the Dossier and for the FrontPage. Now we need the same implementation for the Inbox. In general, we need the `-proxy` adjustment for the whole GEVER and we need the static portal-type adjustment only for the dossier.


elioschmutz:
@phgross 👍

PR #2869, opened 2017-04-24
Move the attach to email action to second position, after the copy action

### LINKED ISSUE #2864: Display OC attach to email action as 2nd action
- [x] The attach to outlook action replaces the old form based E-Mail action and should be displayed at the same position.

 
screen shot 2017-04-20 at 15 31 14

elioschmutz:
Before:
972c569e-25de-11e7-8661-11e07b63cb72

After:
bildschirmfoto 2017-04-24 um 16 53 53

closes #2864

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 ---------------------
 
 - Dossierdetails PDF: Fix indentation of dossier metadata table. [lgraf]
+- Move the attach to email action to second position, after the copy action. [elioschmutz]
 - OGGBundle import: Make sure persistent changes that re-enable LDAP are
   always committed. [lgraf]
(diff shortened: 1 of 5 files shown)


docs/HISTORY.txt
```diff
@@ -5,6 +5,7 @@ Changelog
 2017.1.2 (unreleased)
 ---------------------
 
+- Display officeconnector attach to email action after send as email action. [elioschmutz]
```
phgross: The send_as_email action will be replaced by the attach to email action, whenever the attach feature flag is enabled. So i would change the Changelogentry and the PR description to:
 - Move the attach to email action to second position, after the copy action.

elioschmutz: 👍  updated

PR #2880, opened 2017-04-26
Reduce mtest CPU use

Rotonen:
The `BufferedPrinter` threads were using 100% CPU in a busy loop situation. Adding a 200ms sleep to each of their business logic brings this down to sensibility and reading from the disk only every 200ms is a touch more sensible in any case.

We could even go as far as to consider bumping this up to multiple seconds, but 200ms preserves local interactivity on a human scale rather nicely for local test runs.


jone:
Oops, sorry for this one 🙈 and thanks for fixing!
👍

PR #2882, opened 2017-04-26
submitted proposal edit form: fix wrong class.

jone:
I believe the submited propsal edit form should have the content type class `SubmittedProposal`, not `Proposal`.
⚠️ I have no idea if this is on purpose or by accident, nor can I asses the impact of this ching. It just struck me as odd.


deiferni:
@jone / @phgross  you're correct 👍 .

_Also i have quickly verified that there should be no negative consequences of having used the incorrect class. As far as i could see we only call `partition_data` which is defined on `ModelContainer` and not overwritten by the relevant subclasses `Proposal` or `SubmittedProposal`. So we should be fine._

PR #2889, opened 2017-04-28
Reworked portlet initialization when setting up GEVER

### LINKED ISSUE #1214: Portlet assignments for Inbox, Templates and Contacts are missing on new setup
On a newly set up GEVER from `4.5-stable`, but probably also `master`, the portlet assignments for the `Inbox`, `Templates` and `Contacts` global tabs are missing.

Once fixed this should probably be backported to `4.5-stable`.

phgross:
With the newly introduced `PortletUpdater` in ftw.inflator (https://github.com/4teamwork/ftw.inflator/pull/33), it's now possible to define portlet assignments in the content json files. I replaced the current implementation hook and moved the portlet definitions to the content json files.

Closes #1214

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -14,6 +14,7 @@ Changelog
 
   [lgraf]
 
+- Reworked portlet initialization during setup. [phgross]
 - Meeting: add proposal template tab to templates folder. [jone]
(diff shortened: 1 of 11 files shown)


phgross:
As discussed, we prefer an automatic solution with eventhandlers, that configure the portlets when ever a content is created. I will do that in a separate PR ...

PR #2898, opened 2017-05-01
Tell the user what goes wrong in a large OC multiattach

### LINKED ISSUE #2897: Handle and localize when OC URLs are too long for IE11
Seems we forgot to implement user-side notifications for the case when one attaches too many documents at once.

The scaffolding is already mostly in place so we just need to i18n the message and use a JS alert for notifying the user about what happened.

https://github.com/4teamwork/opengever.core/blob/master/opengever/officeconnector/service.py#L27
https://github.com/4teamwork/opengever.core/blob/master/opengever/document/static/officeconnector.js#L44
https://github.com/4teamwork/opengever.core/blob/master/opengever/officeconnector/helpers.py#L112

Rotonen:
* Pull in ua-parser
  * Needs pinnings for it and its dependencies on KGS?
* Detect IE11 and set limit appropriately
* Return an error message for the case of generating a too long of an URL
* Display any errors from the OC URL API to the user with a JS `alert()`

We can now extend this by just adding more points from which we return an error message. I do not currently foresee a need, but officeatwork stuff might find this useful.

Closes #2897


opengever/officeconnector/helpers.py
```diff
@@ -106,10 +107,16 @@ def create_oc_url(request, context, payload):
     # arbitrary and inconsistent results of 506..509.
     #
     # For operational safety we've set the total url + separator + payload
-    # limit at 500 characters.
+    # limit at 500 or 2000 characters.
     url = 'oc:' + token
+    user_agent = ParseUserAgent(request.environ['HTTP_USER_AGENT'])
```
lukasgraf: -> `request.environ.get('HTTP_USER_AGENT', '')` - otherwise this will result in a `KeyError` for UAs that don't send a `User-Agent` header. This should also eliminate the need to hack in an empty UA string in the test above.

Rotonen: I do not mind giving a 5xx to a client who does not send an user agent, but the test cleanliness is a valid argument - will amend.

Rotonen:

lukasgraf:
@Rotonen I sneaked in a commit that updates the changelog - missed that during review.

Rotonen:
@lukasgraf thank you.

PR #2900, opened 2017-05-01
Avoid unauthorized redirect loops

jone:
The require_login script was customized in order to support MS Office
probing, which happens without the session (cookies).

The old implementation had issues:
- When the request had a payload, such as a POST request, the payload
  did not survive the redirect. When a ++add++ could be rendered but not
  posted, this caused the form to reload empty without an error.
- When the unauthorized exception did not happen during traversal but
  during publishing a redirect loop was created.

In order to fix this issues I rewrote the require_login script to be
more robust.
- No longer redirect when the request method is not GET.
- When redirecting, add a marker GET parameter for detecting loopbacks
  and breaking redirect loops.
- No longer catch all exceptions; catch only the Unauthorized exception.
- Remove GET-Params when converting to a path when probing.
- Crash on purpose when came_from is not within our site root.

---

Besides fixing a bunch of actual redirect loops in the application this change allows us to move on to a newer ftw.testbrowser later, without downgrading the HTTP error handling.

### DIFF
--- a/opengever/base/skins/opengever.base_scripts/require_login.py
+++ b/opengever/base/skins/opengever.base_scripts/require_login.py
@@ -8,21 +8,51 @@
 ##title=Login
 ##
 
-login = 'login'
+from AccessControl import Unauthorized
 
 portal = context.portal_url.getPortalObject()
-# if cookie crumbler did a traverse instead of a redirect,
-# this would be the way to get the value of came_from
-#url = portal.getCurrentUrl()
-#context.REQUEST.set('came_from', url)
+portal_url = portal.absolute_url().rstrip('/') + '/'
+request = context.REQUEST
 
 if context.portal_membership.isAnonymousUser():
-    return portal.restrictedTraverse(login)()
+    return portal.restrictedTraverse('login')()
+
+if request.method != 'GET':
+    # We should not try to redirect when we have a POST payload,
+    # because we cannot take along the payload.
+    return portal.restrictedTraverse('insufficient_privileges')()
+
+
+came_from_url = request.get('came_from')
+if 'require_login_avoid_loop=' in came_from_url.split('?')[-1]:
+    # We already sent the user back to the came_from url,
+    # but it looks that we have a redirect loop due to an Unauthorized
+    # exception while publishing the traversable endpoint.
+    return portal.restrictedTraverse('insufficient_privileges')()
 else:
-    expected_location = context.REQUEST.get('came_from')
-    try:
-        #XXX Attempt a traverse to the given path
-        portal.restrictedTraverse(expected_location.replace(portal.absolute_url()+'/',''))
-        container.REQUEST.RESPONSE.redirect(expected_location)
-    except:
-        return portal.restrictedTraverse('insufficient_privileges')()
+    # Attach a marker so that we can detect redirect loops.
+    if '?' in came_from_url:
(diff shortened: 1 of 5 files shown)


opengever/base/skins/opengever.base_scripts/require_login.py
```diff
@@ -8,21 +8,51 @@
...
+
+
+if not came_from_url.startswith(portal_url):
+    raise ValueError('URL {!r} not in portal {!r}'.format(
+        came_from_url, portal_url))
+
+
+came_from_path = came_from_url[len(portal_url):].split('?')[0].split('#')[0]
+try:
+    # Probe for an actualy Unauthorized exception while traversing.
```
lukasgraf: `actualy` -> `actual`

opengever/base/skins/opengever.base_scripts/require_login.py
```diff
@@ -8,21 +8,51 @@
...
+came_from_url = request.get('came_from')
+if 'require_login_awoid_loop=' in came_from_url.split('?')[-1]:
+    # We already sent the user back to the came_from url,
+    # but it looks that we have a redirect loop due to an Unauthorized
+    # exception while publishing the traversable endpoint.
+    return portal.restrictedTraverse('insufficient_privileges')()
+else:
+    # Attach a marker so that we can detect redirect loops.
+    if '?' in came_from_url:
+        came_from_url += '&require_login_awoid_loop=true'
```
lukasgraf: Hmm. I'm not too excited about this staying in user-visible URLs (apart from some URLs becoming a bit ugly). What if a user ends up on a document this way, and decides to copy & paste this link from the address bar into a word document? 😝 Not a big deal though.

lukasgraf:
I would suggest to deploy this to `dev.onegovgever.ch` quickly after merge - we should give this a thorough test with Virtual Hosting + CAS Portal as well.

(The CAS server URL (`https://dev.onegovgever.ch/portal`) for example is outside the Plone portal URL (`https://dev.onegovgever.ch/fd/`), though I couldn't think of a case where this would end up in `came_from`).

opengever/base/skins/opengever.base_scripts/require_login.py
```diff
@@ -8,21 +8,51 @@
...
+    return portal.restrictedTraverse('login')()
+
+if request.method != 'GET':
+    # We should not try to redirect when we have a POST payload,
+    # because we cannot take along the payload.
+    return portal.restrictedTraverse('insufficient_privileges')()
+
+
+came_from_url = request.get('came_from')
+if 'require_login_awoid_loop=' in came_from_url.split('?')[-1]:
```
jone: Oops

opengever/base/skins/opengever.base_scripts/require_login.py
jone: Yes, that is not nice, but I didn't find a better solution. Cookies wouldn't be a good idea because they are session-wide. I also thought about using an abbreviation, but it doesn't really make it better.

Do you have another idea how to detect redirect loops?

jone:
@lukasgraf thanks for the review; I amended according to your comments.

jone:
Be aware: the discussion about the GET parameter is now "outdated" because of a name change, but it is not really fixed.

PR #2908, opened 2017-05-03
Always render a task link for administrators or managers.

### LINKED ISSUE #2838: Task links aren't linked for Administrators or Managers
Currently only the local_roles are checked by the decision, if a link should be rendered or not (see https://github.com/4teamwork/opengever.core/blob/46456394c6853362e0edd851c8bc5a36608b998d/opengever/globalindex/model/task.py#L304). Therefore global roles as Administrator or Manager are ignored, although they are able to see the tasks, a link is not rendered.

phgross:
Skip access test when the current user has Administrator or Manager role.

Closes #2838

### DIFF
--- a/opengever/globalindex/model/task.py
+++ b/opengever/globalindex/model/task.py
@@ -274,6 +274,10 @@ class Task(Base):
         if not member:
             return False
 
+        roles = api.user.get_roles(user=member)
+        if 'Administrator' in roles or 'Manager' in roles:
+            return True
+
         principals = set(member.getGroups() + [member.getId()])
         allowed_principals = set(self.principals)
         return len(principals & allowed_principals) > 0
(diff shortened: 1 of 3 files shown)


opengever/globalindex/model/task.py
```diff
@@ -274,6 +274,10 @@ def has_access(self, member):
         if not member:
             return False
 
+        roles = api.user.get_roles(user=member)
```
deiferni: i'd better check for permissions (i.e. `cmf.ManagePortal`, ...)

phgross: There is no permission I could check for and actually would make sense. An administrator has no `cmf.ManagePortal` permission, so I can't use that one.

In general I would not reuse an existing permission for a different check/use case. Thats confusing and can be hard to detect when making feature changes.

Another option: Introducing a new Permission `opengever.task.GlobalTaskAccess` only managed by an rolemap.xml and check for that permission. 
@deiferni what do you think?

lukasgraf: 👍  `opengever.task.GlobalTaskAccess`

deiferni: yep, i'd go for that 👍

phgross:
@deiferni updated ...

phgross:
@deiferni ready to rereview ...

PR #2914, opened 2017-05-08
Tests: Manage monkey patching of PDFCONVERTER_AVAILABLE through a context manager

lukasgraf:
Previously, the `PDFCONVERTER_AVAILABLE` flag was being monkey patched in tests without being reset properly.

In one case, it was being "reset" to `False` in a hardcoded way. This was the default until we added `opengever.pdfconverter` as a test dependency. From that point on, the flag was being reset to an incorrect default value. However, this only started to flare up in tests once we shuffled around their execution order with parallelization improvements to `bin/mtest`.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2017.1.2 (unreleased)
 ---------------------
 
+- Tests: Manage monkey patching of PDFCONVERTER_AVAILABLE through a context manager. [lgraf]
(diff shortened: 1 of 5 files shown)


lukasgraf:
@Rotonen also put this on top of your branch to verify it fixes the build: https://ci.4teamwork.ch/builds/81063/tasks/117342

Rotonen:
@lukasgraf imperative or informative?

lukasgraf:
@Rotonen informative :P

PR #2925, opened 2017-05-12
Inter-Admin-Unit Requests: Extract response body into the local namespace

lukasgraf:
Inter-Admin-Unit Requests: Extract response body into the local namespace so Sentry can log it as a local in case of failure.

Example Sentry Event:
https://sentry.4teamwork.ch/sentry/onegov-gever/issues/3129/events/88432/

(See `response_body` in locals below code, above traceback ("Show more"))

This partly helps with #1541 (still some things left to do though)


maethu:
@lukasgraf I'll rebase..

maethu:
Followup of this PR is #2958

PR #2934, opened 2017-05-15
 Respect `maximum_dossier_depth` registry record in DefaultConstrainTypeDecider

### LINKED ISSUE #955: Subdossier-Tiefe: DefaultConstrainTypeDecider sollte Registry-Eintrag maximum_dossier_depth verwenden

maethu:
Changing `maximum_dossier_depth` is now implemented in the `DefaultConstrainTypeDecider`. 
The value is used for:
- opengever.dossier.businesscasedossier
- opengever.private.dossier

I slightly change the signature of the  adapter. CONSTRAIN_CONFIGURATION is no longer a `constant`, since it now depends on a configurable value. This may break customer integration of this adapter.

I also had to change the default value of `maximum_dossier_depth` from 1 to 2, since most of the tests depend on 2 levels.

Closes #955

### DIFF
--- a/opengever/dossier/base.py
+++ b/opengever/dossier/base.py
@@ -259,7 +259,7 @@ class DossierContainer(Container):
             'review_state': [
                 'dossier-state-active',
                 'dossier-state-resolved', ],
-            })
+        })
 
         end_dates = []
         # main dossier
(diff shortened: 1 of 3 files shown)


opengever/dossier/base.py
```diff
@@ -370,25 +370,31 @@ class DefaultConstrainTypeDecider(grok.MultiAdapter):
...
+
+        self.constrain_configuration = {
+            'opengever.dossier.businesscasedossier': {
+                'opengever.dossier.businesscasedossier': max_dossier_depth,
+                'opengever.dossier.projectdossier': 1,
+            },
+            'opengever.private.dossier': {
+                'opengever.private.dossier': max_dossier_depth
+            },
+            'opengever.dossier.projectdossier': {
```
phgross: `opengever.dossier.projectdossier` is obsolete see https://github.com/4teamwork/opengever.core/issues/2936.

maethu:
@phgross I removed the `opengever.dossier.projectdossier` parts. Please check again.

opengever/dossier/base.py
```diff
@@ -370,25 +370,26 @@ class DefaultConstrainTypeDecider(grok.MultiAdapter):
...
-            },
-        'opengever.private.dossier': {
-            'opengever.private.dossier': 2
-        },
-        'opengever.dossier.projectdossier': {
-            'opengever.dossier.projectdossier': 1,
-            'opengever.dossier.businesscasedossier': 1,
-            },
-        }
+    constrain_configuration = None
```
deiferni: the class attribute is not necessary, you'll set the instance variable anyway in init, so the fallback should not be needed

maethu:
@deiferni rebased and fixed the one thing you mentioned. 
I also opened an issue in opengever.zug to fix the adapters inheriting from DefaultConstrainTypeDecider

maethu:
@deiferni One more thing. I change the default value from `maximum_dossier_depth`. But if value us persisted in the registry with 1 for given GEVER installations. Should I migrate the value to 2 if 1 is in registry?

deiferni:
(as just discussed, we should add the missing increment)

maethu:
@deiferni done 👍

maethu:
@deiferni  I had to change the  `test_get_subdossier_depth_from_registry` test.

PR #2938, opened 2017-05-16
Allow the same Users to edit Tasktemplatefolders in active state as in inactive state

### LINKED ISSUE #1795: Titel eines Standardablaufs kann nicht mehr bearbeitet werden
Getestet auch mit zopemaster-User. Der Menüpunkt "Bearbeiten" wird nicht angezeigt. Konkret: Ich möchte den Titel des Standardablaufs umbenennen.

bildschirmfoto 2016-05-08 um 12 37 11

maethu:
Also use ftw.builder in tests and only use one test class. 

tmpl

Closes #1795


opengever/tasktemplates/tests/test_tasktemplate.py
```diff
@@ -101,3 +103,17 @@ def test_view(self, browser):
...
+    def test_tasktemplate_can_be_edited_when_activated(self, browser):
+        templatefolder = create(Builder('tasktemplatefolder')
+                                .titled(u'Task templates')
+                                .in_state('tasktemplatefolder-state-activ'))
+
+        browser.login().visit(templatefolder)
+        self.assertTrue(browser.css('#contentview-edit'),
+                        'Task template folder should be editable in '
+                        'active state')
+        browser.css('#contentview-edit a').first.click()
```
Rotonen: I'm against using indexing in browser tests which will break at layout changes.

maethu: @Rotonen OK I replaced it by going to the `edit` view directly. But I'm still doing the roundtrip with "submitting the form". Is this superfluous?

To submit the form I need to do it by label, since the are multiple forms on that view (livesearch).

Rotonen: Finding by label will only fail when the form itself is changed as the `z3c.form` actually chooses the form-side code path based on the submit value. That is fine.

PR #2940, opened 2017-05-16
Fix local roles string type

### LINKED ISSUE #2927: Probleme mit Umlauten in Gruppennamen
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 23: ordinal not in range(128)

https://sentry.4teamwork.ch/sentry/onegov-gever/issues/3160/

```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 23: ordinal not in range(128)
(9 additional frame(s) were not displayed)
...
  File "home/zope/eggs/Products.CMFPlone-4.3.2-py2.7.egg/Products/CMFPlone/CatalogTool.py", line 393, in searchResults
    return ZCatalog.searchResults(self, REQUEST, **kw)
  File "home/zope/eggs/Products.ZCatalog-3.0.2-py2.7.egg/Products/ZCatalog/ZCatalog.py", line 604, in searchResults
    return self._catalog.searchResults(REQUEST, used, **kw)
  File "home/zope/eggs/Products.ZCatalog-3.0.2-py2.7.egg/Products/ZCatalog/Catalog.py", line 1072, in searchResults
    return self.search(args, sort_indexes, reverse, sort_limit, _merge)
  File "home/zope/eggs/Products.ZCatalog-3.0.2-py2.7.egg/Products/ZCatalog/Catalog.py", line 549, in search
    r = _apply_index(query, rs)
  File "home/zope/eggs/Products.ZCatalog-3.0.2-py2.7.egg/Products/PluginIndexes/common/UnIndex.py", line 426, in _apply_index
    s = index.get(k, None)

UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 23: ordinal not in range(128)
```

buchi:
Fixes #2927

### DIFF
--- a/opengever/meeting/upgrades/20170516133120_fix_local_roles_string_type/upgrade.py
+++ b/opengever/meeting/upgrades/20170516133120_fix_local_roles_string_type/upgrade.py
@@ -1,0 +1,22 @@
+from ftw.upgrade import UpgradeStep
+
+
+class FixLocalRolesStringType(UpgradeStep):
+    """Fix local roles string type.
+    """
+
+    def __call__(self):
+        needs_reindex = False
+        for obj in self.objects({'portal_type': 'opengever.meeting.committee'},
+                                'Fix local roles string type'):
+            for principal, roles in obj.get_local_roles():
+                if isinstance(principal, unicode):
+                    principal = principal.encode('utf8')
+                    obj.manage_delLocalRoles([principal])
+                    obj.manage_setLocalRoles(principal, list(roles))
+                    needs_reindex = True
+
+        if needs_reindex:
+            catalog = self.getToolByName('portal_catalog')
+            catalog.clearIndex('allowedRolesAndUsers')
+            self.catalog_rebuild_index('allowedRolesAndUsers')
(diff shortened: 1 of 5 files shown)


opengever/meeting/upgrades/20170516133120_fix_local_roles_string_type/upgrade.py
```diff
@@ -0,0 +1,19 @@
+from ftw.upgrade import UpgradeStep
+
+
+class FixLocalRolesStringType(UpgradeStep):
+    """Fix local roles string type.
+    """
+
+    def __call__(self):
+        for obj in self.objects({'portal_type': 'opengever.meeting.committee'},
```
deiferni: I'd prefer the option o inspect the index for sanity and only rebuild when we need to. or could we say that this is only an issue when `opengever.meeting` is active or when we find committees? i assume that clearing and rebuilding the index is quite expensive, so if there is a safe way to skip doing that it under certain conditions i'd find a way to do so.

PR #2952, opened 2017-05-17
Fix translation for the "Open as PDF" action.

### LINKED ISSUE #2951: Missing translation for Open as PDF document action.
bildschirmfoto_2017-05-17_um_13_34_07

phgross:
Fixes #2951 

Because the action is now part of the document subdirectory, i moved the translation to the og.document domain.

Backport to: 2017.2.x

### DIFF
--- a/opengever/document/locales/de/LC_MESSAGES/opengever.document.po
+++ b/opengever/document/locales/de/LC_MESSAGES/opengever.document.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
-"POT-Creation-Date: 2017-04-27 11:21+0000\n"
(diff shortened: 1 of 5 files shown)


opengever/document/locales/de/LC_MESSAGES/opengever.document.po
```diff
@@ -114,10 +114,6 @@ msgstr "GEVER verlassen"
 msgid "Reverted file to version ${version_id}."
 msgstr "Datei des Dokuments auf Version ${version_id} zurückgesetzt."
 
-#: ./opengever/document/forms.py:65
-msgid "Save"
```
deiferni: this translation is dropped somehow, is this valid?

phgross: @deiferni Yes I double checked it, its no longer a translated string.

PR #2953, opened 2017-05-17
Fix string type in bundle loader

buchi:
With this fix all items injected into the transmogrifier pipeline contain utf8 encoded byte strings instead of unicode strings. This fixes an issue with unicode strings in local roles (#2927).

It may introduce new issues if something in the ogg bundle pipeline relies on unicode strings. But in general items in the transmogrifier pipeline should always contain byte strings. Various sections will break with unicode strings.

/cc @lukasgraf

### DIFF
--- a/opengever/bundle/loader.py
+++ b/opengever/bundle/loader.py
@@ -1,3 +1,5 @@
+from collections import Iterable
+from collections import Mapping
 from collections import OrderedDict
 from datetime import datetime
 from jsonschema import FormatChecker
@@ -117,7 +119,7 @@ class BundleLoader(object):
     def _load_items(self):
         self._items = []
         for json_name, portal_type in BUNDLE_JSON_TYPES.items():
-            items = self._load_json(json_name)
+            items = unicode2bytes(self._load_json(json_name))
             if items is None:
                 continue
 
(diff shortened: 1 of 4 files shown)


opengever/bundle/loader.py
```diff
@@ -244,3 +246,17 @@ def __call__(self):
...
+
+
+def unicode2bytes(data):
+    """Converts unicode strings to byte strings."""
+    if isinstance(data, unicode):
+        return data.encode('utf8')
+    elif isinstance(data, str):
+        return data
+    elif isinstance(data, Mapping):
+        return dict(map(unicode2bytes, data.iteritems()))
```
deiferni: should we also attempt preserve the mapping-type here?

opengever/bundle/loader.py
```diff
@@ -244,3 +246,17 @@ def __call__(self):
             log.info('No ingestion settings found at %s' % settings_path)
             return {}
         return settings
+
+
+def unicode2bytes(data):
```
deiferni: could you provide some unit-tests for this helper function?

buchi: I've added some unit tests, though this is rather trivial code.

opengever/bundle/loader.py
buchi: done.

opengever/bundle/loader.py
deiferni: 😄 💛

buchi:
The DX schema updater handles byte strings: https://github.com/collective/transmogrify.dexterity/blob/master/transmogrify/dexterity/converters.py#L449

As already said above, it may introduce new issues. But relying on unicode strings in the transmogrifier pipeline is wrong.

PR #2958, opened 2017-05-18
Implement and use a safe_call decorator for all remote endpoints

maethu:
The goal is to improve the error handling and debugging ability for requests to other gever sites.
See #1541 for more details.

This PR implements:
- A `safe_call` decorator, which wraps the `__call__` method of an endpoint in a try/except. In case of an exception the traceback is returned as string.
- The safe_call decorator also tries to notify sentry if possible, to not swallow the exception silently.
- On the client site the `dispatch_request` method is now able to  report the traceback of the failed remote call as exception message.
- ConflictError and KeyBoardInterrupt are not catched.
- The parameter to_raise allows you to add more Exceptions, which should not be handled.

**The result:**
If a remote call fails there are 2 entries in sentry. https://sentry.4teamwork.ch/sentry/exceptiontest/
one from the client (actually raised to the user) and one from the remote site (manually send to sentry).

screen shot 2017-05-18 at 09 09 51

### DIFF
--- a/opengever/base/request.py
+++ b/opengever/base/request.py
@@ -159,7 +159,7 @@ def _remote_request(target_admin_unit_id, viewname, path, data, headers):
     response = opener.open(request)
     content = response.read().strip()
 
-    if response.headers.type == 'text/traceback':
+    if response.headers.type == 'text/x.traceback':
         raise RemoteRequestFailed(url, content)
     else:
         return StringIO(content)
@@ -189,7 +189,7 @@ def tracebackify(*args, **kwargs):
                     raise
                 except Exception:
                     self.request.response.setHeader("Content-type",
-                                                    "text/traceback")
+                                                    "text/x.traceback")
                     e_type, e_value, tb = sys.exc_info()
 
                     if HAS_RAVEN:
--- a/opengever/base/request.py
+++ b/opengever/base/request.py
@@ -33,7 +33,7 @@ class RemoteRequestFailed(Exception):
         self.tb = tb
 
     def __str__(self):
-        return 'Remote request to "{url}": Traceback: {tb}'.format(
+        return 'Remote request to "{url}" failed: Traceback: {tb}'.format(
             **{'url': self.url, 'tb': self.tb})
 
 
@@ -156,21 +156,13 @@ def _remote_request(target_admin_unit_id, viewname, path, data, headers):
                               urllib.urlencode(data),
                               headers)
 
-    content = opener.open(request).read().strip()
+    response = opener.open(request)
+    content = response.read().strip()
 
-    try:
-        json.loads(content)
-    except ValueError:
-        is_json = False
-    else:
-        is_json = True
-
-    if content == 'OK' or is_json:
-        # Other code depends on a io obj.
-        return StringIO(content)
-    else:
-        # Otherwise expect a traceback as response from the remote server
+    if response.headers.type == 'text/traceback':
         raise RemoteRequestFailed(url, content)
+    else:
+        return StringIO(content)
 
 
 def tracebackify(*args, **kwargs):
@@ -197,7 +189,7 @@ def tracebackify(*args, **kwargs):
                     raise
                 except Exception:
                     self.request.response.setHeader("Content-type",
-                                                    "text/plain")
+                                                    "text/traceback")
                     e_type, e_value, tb = sys.exc_info()
 
                     if HAS_RAVEN:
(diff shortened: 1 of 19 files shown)


maethu:
hmmm.. working on testing

maethu:
@deiferni 

- Added some test cases
- Add parameter to ignore more exceptions.

opengever/base/request.py
```diff
@@ -129,4 +143,64 @@ def _remote_request(target_admin_unit_id, viewname, path, data, headers):
...
+
+    if content == 'OK' or is_json:
+        return response
+    else:
+        # Otherwise expect a traceback as response from the remote server
+        raise Exception('The remote request failed', response)
+
+
+def safe_call(*args, **kwargs):
+    """Decorator for remote enpoints
```
deiferni: typo: enpoints

opengever/base/request.py
```diff
@@ -129,4 +143,64 @@ def _remote_request(target_admin_unit_id, viewname, path, data, headers):
...
+        is_json = True
+
+    if content == 'OK' or is_json:
+        return response
+    else:
+        # Otherwise expect a traceback as response from the remote server
+        raise Exception('The remote request failed', response)
+
+
+def safe_call(*args, **kwargs):
```
deiferni: naming suggestion: should we call this `remote_view` instead? using the decorator makes only sense for remote stuff so it would make sense to include that information in the decorator name already. also we decorate the class, i.e. the view.

maethu: @deiferni I called  it `safe_call`, because it makes the __call__ method somehow safer. 
The decorator works for every view, doesn't matters if it is a remote call, or local traversing. 

At least `call`, should be in the name.

imho the current name is suitable for the usecases.

deiferni: hmm, what i'm interested in is that it should be obvious to the programmer that the decorator is on the receiving side of every `dispatch_json_request` and `dispatch_request`. i would find it useful if that was obvious from the decorator name, to me `safe_call` does not contain that information. naming is hard 🤔 .

opengever/base/request.py
```diff
@@ -129,4 +143,64 @@ def _remote_request(target_admin_unit_id, viewname, path, data, headers):
...
+
+    if not isinstance(to_raise, list):
+        to_raise = [to_raise]
+
+    def wrapper(Cls):
+
+        class SafeCall(Cls):
+            def __call__(self, *args, **kwargs):
+
+                try:
```
maethu: @deiferni done.

opengever/base/request.py
maethu: `make_remote_request_safe_by_wrapping_the_call_method_decorator_of_browserviews` :rofl:

maethu: @deiferni How about just: `endpoint`

deiferni: as discussed we settled for `tracebackify`

maethu:
@deiferni ready!!!! :-)

opengever/base/request.py
```diff
@@ -129,4 +155,63 @@ def _remote_request(target_admin_unit_id, viewname, path, data, headers):
     request = urllib2.Request(url,
                               urllib.urlencode(data),
                               headers)
-    return opener.open(request)
+
+    content = opener.open(request).read().strip()
```
deiferni: we've been discussing a custom `content-type` header for the exceptions to detect whether we have an exception response. would that still be a valid alternative for traceback-detection here?

maethu: ahh missed that -.-

maethu:
@deiferni again ;-) `text/traceback` content-type introduced.

opengever/base/request.py
```diff
@@ -129,4 +155,55 @@ def _remote_request(target_admin_unit_id, viewname, path, data, headers):
...
+
+                try:
+                    return super(SafeCall, self).__call__(*args, **kwargs)
+                except (ConflictError, KeyboardInterrupt):
+                    raise
+                except tuple(to_re_raise):
+                    raise
+                except Exception:
+                    self.request.response.setHeader("Content-type",
+                                                    "text/traceback")
```
deiferni: should we use text/x.traceback here to indicate that this is not an official type registered with IANA,  as indicated by http://www.restpatterns.org/Glossary/MIME_Type#Custom_MIME_Types ?

maethu: Wasn't aware of that. According to restpatterns.org there needs to be a `x` as prefix

maethu: @deiferni updated.

PR #2966, opened 2017-05-18
Disable ldap authentication with cas

### LINKED ISSUE #1657: login_form bei Installationen mit Portalseite nicht zugänglich machen
Es soll nur noch der Login über die Portalseite möglich sein. 

Zusätzlich kann bei jeder solchen Installation beim entsprechenden LDAP-Plugin die Funktion "Authentication" deaktiviert werden.

buchi:
Fixes #1657

### DIFF
--- a/opengever/setup/upgrades/configure.zcml
+++ b/opengever/setup/upgrades/configure.zcml
@@ -22,7 +22,7 @@
         />
 
     <upgrade-step:directory
-        profile="opengever.setup:default"
+        profile="opengever.setup:casauth"
         directory="."
         />
 
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2017.2.2 (unreleased)
 ---------------------
 
+- Disable LDAP authentication in CAS setups. [buchi]
 - Make sure that pasting is allowed before pasting. [deiferni]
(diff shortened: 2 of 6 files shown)


opengever/setup/upgrades/configure.zcml
```diff
@@ -22,7 +22,7 @@
         />
 
     <upgrade-step:directory
-        profile="opengever.setup:default"
```
deiferni: oh i just noticed that this repurposes the `upgrades` directory, by convention this always points to the `default` profile. also it may cause conflicts for the upcoming profile merge. could you provide a separate `upgrades_cas` directory?

deiferni:
@buchi `ping`
what's the state here? in the meantime we have merged all our generic setup profiles (see https://github.com/4teamwork/opengever.core/pull/2570) so this PR needs an update.

phgross:
@buchi kannst du diesen PR noch aktualisieren(siehe Kommentar von deif)

deiferni:
@buchi what is the state here? could we merge this? or are there still unresolved issues wit the API?

lukasgraf:
This is obsolete - we already deactivate the `IAuthentication` capability for the `ldap` plugin by default for non-development setups: 

https://github.com/4teamwork/opengever.core/blob/d2dfc7636f2c0505fc1318565e03ecbf28ba936c/opengever/setup/deploy.py#L121-L124

Issue #1244, opened 2015-10-09
Make batching in journal tab lazy


lukasgraf:
It looks like the current journal implementation doesn't do lazy batching - it loads all the entries at once, even though only the first `${batch_size}` items need to be displayed.

Particularly on contexts with huge journals this results in a significant performance hit.

Rotonen:
As far as I can tell, we'll need to add a new lazily queryable storage backend to `ftw.journal` and migrate all of our journal entries to that.

phgross:
That issue is not urgent yet, we already discussed the use of `ftw.activity` instead of the current `ftw.journal`. But we would need to write a OGIP and etc. I move the issue to inbox.

Discussion:  https://basecamp.com/2768704/projects/13482340/todos/315085591

lukasgraf:
Specifically, the storage in `ftw.journal` would need to be switched to a BTree instead of a `PersistentList`, and existing journal entries be migrated accordingly.

PR #2969, opened 2017-05-19
Optimize journal queries

Rotonen:
The actual root cause of #1244 was tackled in the lack of actual understanding of how to build the whole thing properly up from `ftw.table` -> `ftw.tabbedview` -> `opengever.core`. This is doable, but not within the time frame of the sprint now. Our current implementation is a hack which goes against the grain of how one should use these tools.

The actual end result query should eventually end up relying on `plone.batching`.

In the course of mucking around a technique to optimize the slowest part, the deepcopy within the abusely-built-query, was discovered. It turned out pickling and unpickling the objects with cPickle is faster than making a deepcopy. This is a hack, but a valid one at least partially addressing the issue at hand?

I've locally metrified a gain in the journal of a dossier, but further testing by others on this metric is most welcome as this seems to be a bit hardware dependent. If we can find a situation this ends up being slower for this particular query, we should not proceed as this essentially trades a bit of RAM (the serialization overhead vs. the deepcopy object in-memroy) for less CPU cycles.

`/fd/ordnungssystem/ressourcen-und-support/finanzen/planung/finanzplanung/dossier-5#journal`

```
Entry count: 7
Deepcopy took 0.0014500617981s
cPickle took 0.000730037689209s
cPickle / Deepcopy = 0.503452811575
Deepcopy / cPickle = 1.98628347485
```

```
Entry count: 107
Deepcopy took 0.0270109176636s
cPickle took 0.0104432106018s
cPickle / Deepcopy = 0.386629241253
Deepcopy / cPickle = 2.5864572394
```

```
Entry count: 1107
Deepcopy took 0.623929023743s
cPickle took 0.126353025436s
cPickle / Deepcopy = 0.202511857324
Deepcopy / cPickle = 4.93798246293
```

```
Entry count: 11107
Deepcopy took 6.77266812325s
cPickle took 2.10805010796s
cPickle / Deepcopy = 0.311258439007
Deepcopy / cPickle = 3.21276429706
```

```
Entry count: 111107
Deepcopy took 85.6392359734s
cPickle took 28.9555430412s
cPickle / Deepcopy = 0.338110711897
Deepcopy / cPickle = 2.95761111617
```

The results above are averages of 5 runs each wherein discarding the best and worst results as outliers. I had no patience for more runs of the 100k set. No hard guarantees, but better than nothing.

One can also easily get in-situ in-context metrification (or journal entry generation) going on in a pdb session when one pops into a local context proper interpreter via `!import code; code.interact(local=vars())`.

Where to loop to generate entries:
https://github.com/4teamwork/opengever.core/blob/60b6df7c4079b14c34d75e02d0043031853ce3d9/opengever/journal/form.py#L67

```
for x in xrange(n):
    if x % 100 == 0:
        print x
    entry.save()
```

Where to metrify the gains:
https://github.com/4teamwork/opengever.core/blob/60b6df7c4079b14c34d75e02d0043031853ce3d9/opengever/journal/tab.py#L142

```
from copy import deepcopy
from timeit import default_timer

import cPickle

print "Entry count: {}".format(len(data))

begin = default_timer()
deepcopy(data)
end = default_timer()
dcresult = end - begin

print "Deepcopy took {}s".format(dcresult)

begin = default_timer()
cPickle.loads(cPickle.dumps(data))
end = default_timer()
cpresult = end - begin

print "cPickle took {}s".format(cpresult)

print "cPickle / Deepcopy = {}".format(cpresult / dcresult)
print "Deepcopy / cPickle = {}".format(dcresult / cpresult)
```

### DIFF
--- a/opengever/journal/tab.py
+++ b/opengever/journal/tab.py
@@ -1,5 +1,4 @@
 from BeautifulSoup import BeautifulSoup
-from copy import deepcopy
 from five import grok
 from ftw.journal.config import JOURNAL_ENTRIES_ANNOTATIONS_KEY
 from ftw.journal.interfaces import IAnnotationsJournalizable
@@ -21,6 +20,8 @@ from zope.globalrequest import getRequest
 from zope.i18n import translate
 from zope.interface import implements
 from zope.interface import Interface
+
+import cPickle
 
 
 def tooltip_helper(item, value):
@@ -139,6 +140,8 @@ class JournalTableSource(GeverTableSource):
               'title': title_column_sorter,
               'actor': actor_column_sorter}
 
+    batching_enabled = True
+
     def validate_base_query(self, query):
         context = self.config.context
         if IAnnotationsJournalizable.providedBy(context):
@@ -147,7 +150,9 @@ class JournalTableSource(GeverTableSource):
         elif IWorkflowHistoryJournalizable.providedBy(context):
             raise NotImplemented
 
-        data = deepcopy(data)
+        # XXX - a performance hack to replace deepcopy(data)
+        # This only works as persistent objects are guaranteed to be picklable
+        data = cPickle.loads(cPickle.dumps(data))
         return data
 
     def extend_query_with_ordering(self, results):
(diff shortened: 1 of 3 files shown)


opengever/journal/tab.py
```diff
@@ -139,7 +150,7 @@ def validate_base_query(self, query):
         elif IWorkflowHistoryJournalizable.providedBy(context):
             raise NotImplemented
 
-        data = deepcopy(data)
+        data = cPickle.loads(cPickle.dumps(data))
```
deiferni: on second thought this may be worth a comment with a short hint that this is a performance hack

Rotonen:
@deiferni ready for rereview.

PR #2998, opened 2017-06-01
First steps towards project shared QA tool settings

Rotonen:
The intention here:

"You are not forced to use these tools on this project, but if you do use these tools, you are forced to use these commonly agreed upon settings on this project."

I'm myself on pycodestyle + pydocstyle + pyflakes + isort.

The expectation is discussion and changes on the rules will slowly naturally happen over time.

The big discussion will be pylint as it's very verbose and quite opinionated. If we go for pylint, I would start by ignoring everything and starting to whitelist checks one by one - IMO the useful checks are mostly import level issues pyflakes would not catch.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,19 @@
+[isort]
+force_alphabetical_sort = True
+force_single_line = True
+lines_after_imports = 2
+line_length = 200
+not_skip = __init__.py
+
+[mccabe]
+
+[pycodestyle]
+
+[pydocstyle]
+ignore = D100,D102,D103,D105,D200,D203,D205,D212,D213,D400,D401,D404
+
+[pyflakes]
+
+[pylint]
+
+[radon]


setup.cfg
```diff
@@ -0,0 +1,19 @@
...
+lines_after_imports = 2
+line_length = 200
+not_skip = __init__.py
+
+[mccabe]
+
+[pycodestyle]
+
+[pydocstyle]
+ignore = D100,D102,D103,D105,D200,D203,D205,D212,D213,D400,D401,D404
```
deiferni: could also add
- D101 (imho not always a good idea, especially for test-classes)
- D104 (we rarely do so, some packages are self-explanatory)

just ouf of curiosity, how did you come up with that list 😄 ?

Rotonen: This list is essentially from @lukasgraf.

AFAIK we can amend `#noqa: D101` to the test classes. I'd prefer this.

I've yet to encounter D104 in anything opened so far.

deiferni: i personally am not a fan of spreading the `#noqa` comments everywhere in the code, it is fine for the imports but if you have to write a `#noqa` comment instead of a docstring then what's the point 🤔 . may be personal preference though … but i would prefer to add the ignore rules for most docstrings and then require/improve docstrings in reviews. Otherwise i fear we end up with meaningless docstrings that just repeat the method name/class name.

Rotonen: As an alternative setup.cfg would allow ignores for directories. I'm not sure if that has glob support in it. We could just ignore pycodestyle on `./**/test_*.py,./**/__init__.py` and thus avoid the discussion.

setup.cfg uses the buildout-familiar ini syntax:
```
ignores =
    # Ignore docstring warnings for tests
    ./**/test_*.py
    # Ignore docstring warnings for modules
    ./**/__init__.py
```

I've not tested this yet, but as a concept would this work for you?

deiferni: @Rotonen sounds tempting 😉

lukasgraf:
It's not obvious to me why these changes need to happen in `opengever.core` instead of `ftw-buildouts`, where there's already some efforts to provide common tooling configurations:
https://github.com/4teamwork/ftw-buildouts/blob/master/pep8.cfg
https://github.com/4teamwork/ftw-buildouts/blob/master/pylintrc

We don't want divergent style guides for `ftw.*` and `opengever.core`, so can't we address this on a company wide level?

Rotonen:
@lukasgraf I was hoping to limit the scope of discussion to our team first and later move it out to be company wide after a wider discussion.

PR #3004, opened 2017-06-06
Hard-code base profiles to be installed for every deployment.

deiferni:
The profiles

- `opengever.core:default`
- `plonetheme.teamraum:gever`
- `opengever.policy.base:mimetype`

must always be installed in that exact order and are always the first profiles that were listed in a policy's `metadata.xml`. This is an unnecessary duplication, so with this PR we hard-code the to be installed profiles. Also remove the option `base_profile` for the policy directives.

When this PR is merged it is no longer necessary to include the profiles manually in each policy. Also policies will most likely fail if they have incorrect dependencies sine https://github.com/4teamwork/opengever.core/pull/2570 introduced a check to avoid duplicate profile installation.

### DIFF
--- a/opengever/setup/deploy.py
+++ b/opengever/setup/deploy.py
@@ -30,8 +30,6 @@ BASE_PROFILES = (
     'opengever.policy.base:mimetype',
 )
 
-MIMETYPE_FIX_PROFILE = 'profile-opengever.policy.base:mimetype'
-
 
 class GeverDeployment(object):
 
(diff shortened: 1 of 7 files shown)


opengever/setup/deploy.py
```diff
@@ -145,8 +143,6 @@ def install_policy_profile(self):
         stool = getToolByName(self.site, 'portal_setup')
         stool.runAllImportStepsFromProfile(
             'profile-{}'.format(policy_profile))
-        # fix mime-type definitions by overriding temaraum-theme mimetypes
-        stool.runAllImportStepsFromProfile(MIMETYPE_FIX_PROFILE)
```
phgross: If the policy has a dependency to `plonetheme.teamraum`, which they actually all have, the plonetheme.teamraum profile is installed after the `MIMETYPE_FIX_PROFILE` and the mimetypes aren't fixed ..

phgross: Ah, now I see that you'll remove all the `plonetheme.teamraum` dependencies in the policy, as you do in og.hatchery. So that change should be ok.

deiferni: exactly. now that i think about it, we also should guard against installing plonetheme.teamraum twice. ill amend!

deiferni:
@phgross amended

PR #3008, opened 2017-06-06
Use ProcessPoolExecutor to execute the batches.

deiferni:
Use `ProcessPoolExecutor` instead of `ThreadPoolExecutor`. It seems to be able to isolate environment variables correctly whereas `ThreadPoolExecutor` seems to have problems.

I have verified that with a modified `bin/test` that just logged the ports in its env, see https://github.com/4teamwork/opengever.core/tree/deif-TMP-debug-ports. After switching to ProcessPoolExecutor duplicate port assignments could not be observed in the log any more, before switching duplicate port assignments could be observed in every run.

### DIFF
--- a/bin/mtest
+++ b/bin/mtest
@@ -85,7 +85,7 @@ def main():
         ]
 
     print("Running tests with {} processors.".format(PROCESSORS), flush=True)
-    with concurrent.futures.ThreadPoolExecutor(max_workers=PROCESSORS) as executor:  # noqa
+    with concurrent.futures.ProcessPoolExecutor(max_workers=PROCESSORS) as executor:  # noqa
         outputs = executor.map(run_tests, batches)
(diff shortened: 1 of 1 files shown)


Rotonen:
@deiferni this is how `concurrent.futures` does it under the hood:

https://github.com/python/cpython/blob/master/Lib/concurrent/futures/process.py#L430
https://github.com/python/cpython/blob/master/Lib/multiprocessing/process.py#L105

But as per:
https://github.com/python/cpython/blob/master/Lib/multiprocessing/process.py#L68

I do not understand what is going on there either. Somehow that subclasses multiprocessing and redefines the `Popen()` of the `BaseProcess`, but cannot see where. Would have liked to see an example of how to do it as we need as that'd be useful in general.

PR #3015, opened 2017-06-07
Update translations

phgross:
- Move translation of strings profile XMLs (actions.xml, ftis etc.) which has been moved to the opengever.core-profile recently, to a separate domain `opengever.core`
- Update all translations files
- Improve /fix some translations strings according to our translators feedback.


opengever/core/locales/de/LC_MESSAGES/opengever.core.po
```diff
@@ -0,0 +1,252 @@
...
+msgid "Create Disposition"
+msgstr "Aussonderungsangebot erstellen"
+
+#: ./opengever/core/profiles/default/actions.xml
+msgid "Create Task"
+msgstr "Aufgabe erstellen"
+
+#: ./opengever/core/profiles/default/actions.xml
+msgid "Delete"
+msgstr "Supprimer"
```
deiferni: this is french 🤔

opengever/core/locales/de/LC_MESSAGES/opengever.core.po
```diff
@@ -0,0 +1,252 @@
...
+msgid "Submit additional documents"
+msgstr "Zusätzliche Anhänge einreichen"
+
+#: ./opengever/core/profiles/default/types/opengever.meeting.submittedproposal.xml
+msgid "Submitted Proposal"
+msgstr "Eingereichten Antrag"
+
+#: ./opengever/core/profiles/default/types/opengever.task.task.xml
+msgid "Task"
+msgstr "Tâche"
```
deiferni: this is also french.

opengever/core/locales/de/LC_MESSAGES/opengever.core.po
```diff
@@ -0,0 +1,252 @@
...
+msgid "untrashed"
+msgstr "Reaktivieren"
+
+#: ./opengever/core/profiles/default/actions.xml
+msgid "with comment"
+msgstr "Mit Kommentar"
+
+#: ./opengever/core/profiles/default/actions.xml
+msgid "without comment"
+msgstr "Ohne Kommentar"
```
deiferni: shouldn't this string be deleted from another file?

phgross:
Some of the message strings are still used in other packages, therefore they're not removed in other files. 

Because the bin/i18n-build checks only for translations inside the subpackage, I would let the duplicated for now.

phgross:
@deiferni updated ...

Issue #2825, opened 2017-04-06
The tooltip has JS performance issues with unexpected AJAX returns


Rotonen:
When accessing a tooltip view of a document, the javascript throws a hissy fit if something unexpected, like the login prompt, is returned instead.

```at XMLHttpRequest.d (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (eval at <anonymous> (++resource++plone.app.jquery.js:2))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))))), <anonymous>:4:18318)```

This trashes the CPU enough to grind my development machine down to a halt.

[...]

PR #3020, opened 2017-06-08
Handle unknown asynchronous tooltip response.

bierik:
Closes https://github.com/4teamwork/opengever.core/issues/2825

The tooltip may call the content with a logged out user. This leads
to an unwanted rendering of the whole login page in the tooltip.

So add a response header to make sure the response is a true tooltip response
and redirect the user to the login page with a simple reload.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 
 2017.4.1 (unreleased)
 ---------------------
+- Handle unknown asynchronous tooltip response. [Kevin Bieri]
 - Update ftw.keyowordwidget to 1.3.3, to fix a problem while select/search new keywords. [mathias.leimgruber]
 - Upgrade ftw.zopemaster which uses the new ftw.slacker for slack notifications. [elioschmutz]
 - Refactor the function: earliest_possible_end_date. [elioschmutz]
--- a/opengever/base/browser/resources/tooltip.js
+++ b/opengever/base/browser/resources/tooltip.js
@@ -12,10 +12,10 @@
       my: "center left",
       adjust: {
         mouse: false, // Do not track the mouse
-        x: 20, // Slighly shift the location in x direction
-        y: 20 // Slighly shift the location in y direction
+        x: 20, // Slightly shift the location in x direction
+        y: 20 // Slightly shift the location in y direction
       },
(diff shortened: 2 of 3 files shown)


bierik:
@phgross We need a way to provide a translated string for rendering the error message.

bierik:
@phgross ping

bierik:
@phgross pingpong

bierik:
As discussed with @phgross we implement a browserview which returns a json containing the translated labels.

Rotonen:
@phgross @bierik why not a `plone.rest` service? That'd be neater here?

bierik:
👍 Of course

deiferni:
@bierik is this thing ready - a bit confused about the PR history here 🤔 ?

bierik:
@phgross updated

opengever/base/configure.zcml
```diff
@@ -110,6 +110,15 @@
       permission="zope2.View"
       />
 
+  <!-- JSON endpoint for javascript translations -->
+  <plone:service
+      method="GET"
+      for="*"
+      factory=".service.Internationalization"
+      name="@i18n"
+      permission="zope.Public"
```
Rotonen: Rather use `zope2.View`? Or the REST API permission?

opengever/base/service.py
```diff
@@ -1,17 +1,17 @@
+from frontend_translations import get_translations
 from opengever.document.interfaces import ICheckinCheckoutManager
-from plone.rest import Service
+from plone.restapi.services import Service
```
Rotonen: What's going on here?

bierik: Right I should put that in a separate commit. Here I was not able to make the `i18n` endpoint working with the `plone.rest` `Service`. So I used the `plone.restapi.services`. This was also @lukasgraf suggestion.

opengever/base/configure.zcml
bierik: Anonymous users need to have access to this endpoint as well. What `REST API` permission should I use here?

bierik:
@Rotonen @phgross This PR is just about fixing unknown asynchronous tooltip response.
The implementation for the i18n on client side has been moved to https://github.com/4teamwork/opengever.core/pull/3020

PR #3028, opened 2017-06-13
Added missing cancel button for the manual journal entry form.

phgross:
Because z3c.form could not handle the inherited add and the new cancel button, also the add button hat do be defined by the `ManualJournalEntryAddForm` itself.

Besides that I've also fixed the styling of the add button.

### DIFF
--- a/opengever/journal/form.py
+++ b/opengever/journal/form.py
@@ -83,6 +83,10 @@ class ManualJournalEntryAddForm(AddForm):
         return self.request.RESPONSE.redirect(
             '{}#journal'.format(self.context.absolute_url()))
 
+    def updateActions(self):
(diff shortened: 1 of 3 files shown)


opengever/journal/form.py
```diff
@@ -64,6 +67,26 @@ class ManualJournalEntryAddForm(AddForm):
         async=True
     )
 
+    @button.buttonAndHandler(z3c_mf('Add'), name='add')
+    def handleAdd(self, action):
+        data, errors = self.extractData()
```
deiferni: can't you just do a super-call here? would prefer that to copying the code.

phgross:
@deiferni updated ...

phgross:
@elioschmutz tests are failing ...

PR #3035, opened 2017-06-16
Show userid instead of E-Mail address on responsible field of dossier.

### LINKED ISSUE #3022: Dossier Federführung zeigt E-Mail Adresse anstatt Benutzername in Klammern an.
dev.onegovgever.ch
bildschirmfoto_2017-06-09_um_11_12_15

test.onegovgever.ch
bildschirmfoto_2017-06-09_um_11_13_32

Ich vermute das diese Änderung mit den Anpassungen rund um OGIP 15 reingerutscht ist, @maethu kannst du das korrigieren und auch bei anderen Benutzer-Auswahl Felder kontrollieren?

maethu:
screen shot 2017-06-16 at 08 45 28

Fixes #3022

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -7,6 +7,7 @@ Changelog
 
 - Include missing permission definitions from Products.CMFCore. [phgross]
 - Fix PrivateFolders Title encoding. [phgross]
+- Show userid instead of E-Mail address on all sources (except EMailSource). [mathias.leimgruber]
 - Fix "leave GEVER" on logout overlay. [mathias.leimgruber]
 - Disbale swfobject.js - no longer load Flash Fallback for multiuploads. [mathias.leimgruber]
 - Add missing contacts only sources, used by customer special dossiers. [phgross]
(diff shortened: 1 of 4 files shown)


docs/HISTORY.txt
```diff
@@ -5,6 +5,7 @@ Changelog
 2017.3.0 (unreleased)
 ---------------------
 
+- Show userid instead of E-Mail address on responsible field of dossier. [mathias.leimgruber]
```
phgross: Does this change only has inpact to the dossiers responsible field or does now all user selections this way?

Because all user selections, with the ecxeption of the Email-Address Source, should contain the userid not the email address, see https://test.onegovgever.ch/fd/ordnungssystem/soziale-sicherheit/ahv-iv-und-zusatzleistungen/allgemeines/dossier-23/++add++opengever.task.task for an example of the previous release.

maethu: @phgross It only affects the dossier responsible field.

phgross: @maethu as is already mentioned all sources should contain the userid not the e-mail address in parentheses, except the sources who contains a e-Mail address as value, could you change that behavior ?

maethu: ahh `all`, no I did not checked that sry...

maethu:
@phgross I also changed `UsersContactsInboxesSource` and `AllUsersAndInboxesSource`.

PR #3039, opened 2017-06-19
Catch up tests with ftw.testbrowser bugfix.

deiferni:
In https://github.com/4teamwork/ftw.testbrowser/pull/55 a bug has been fixed where the view would be exposed as `browser.context`. Unfortunately a test actually relied on that behaviour 😂 . This PR rewrites the test to only test against the model.

_CL should not be required i'd say_


lukasgraf:
Thanks! 💚  💜  🎉 

Agreed, no CL needed. 👍

PR #3046, opened 2017-06-21
Add integration testing layer with fixture.

jone:
The integration testing layer sets up a fixture with commonly used objects.
By setting up a fixture instead of creating all the objects in each test we want to speed up the test suite.

Since we cannot rollback commits in SQL, we use integration testing and avoid unwanted commits so that we can achieve isolation between tests by rolling back to savepoints.

This change provides the infrastructure for building the fixture and gradually switching tests to the new integration testing.

It is a conscious decision to call the new testing "integration testing" and the old testing "functional testing" in order to have a simple language to communicate between developers.

### DIFF
--- a/opengever/core/testing.py
+++ b/opengever/core/testing.py
@@ -4,6 +4,7 @@ from ftw.builder import session
 from ftw.builder.testing import BUILDER_LAYER
 from ftw.builder.testing import set_builder_session_factory
 from ftw.bumblebee.tests.helpers import BumblebeeTestTaskQueue
+from ftw.testbrowser import TRAVERSAL_BROWSER_FIXTURE
 from ftw.testing import ComponentRegistryLayer
 from ftw.testing import TransactionInterceptor
 from ftw.testing.layer import COMPONENT_REGISTRY_ISOLATION
@@ -493,5 +494,5 @@ class GEVERIntegrationTesting(IntegrationTesting):
 OPENGEVER_INTEGRATION_TESTING = GEVERIntegrationTesting(
     # Warning: do not try to base other layers on ContentFixtureLayer.
     # See docstring of ContentFixtureLayer.
-    bases=(ContentFixtureLayer(), ),
+    bases=(ContentFixtureLayer(), TRAVERSAL_BROWSER_FIXTURE),
     name="opengever.core:integration")
--- a/opengever/core/testing.py
+++ b/opengever/core/testing.py
@@ -5,6 +5,7 @@ from ftw.builder.testing import BUILDER_LAYER
 from ftw.builder.testing import set_builder_session_factory
 from ftw.bumblebee.tests.helpers import BumblebeeTestTaskQueue
 from ftw.testing import ComponentRegistryLayer
+from ftw.testing import TransactionInterceptor
 from ftw.testing.layer import COMPONENT_REGISTRY_ISOLATION
 from ftw.testing.quickinstaller import snapshots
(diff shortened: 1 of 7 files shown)


opengever/core/testing.py
```diff
@@ -380,3 +384,114 @@ def setUpPloneSite(self, portal):
...
+        in the correct spot easily by using the fixture objects as parents.
+        Therefore we do not allow to create objects at the portal.
+        """
+        pass
+
+
+class GEVERIntegrationTesting(IntegrationTesting):
+    """Custom integration testing extension:
+    1. Make savepoint at test begin and rollback after the test.
+    2. Prevent all interaction with transactions, primarely in order to avoid
```
deiferni: primarily

opengever/core/tests/test_integration_testing_setup.py
```diff
@@ -0,0 +1,51 @@
+from ftw.builder import Builder
+from ftw.builder import create
+from opengever.ogds.base.utils import ogds_service
+from opengever.testing import IntegrationTestCase
+import transaction
+
+
+class TestIntegrationTestSetup(IntegrationTestCase):
+
```
deiferni: you could add a quick description why you alias the methods, in case it should not be obvious for someone

opengever/testing/fixtures.py
```diff
@@ -0,0 +1,43 @@
...
+    @staticuid('repotree')
+    def create_repository_tree(self):
+        create(Builder('repository_root').titled('Repository Root'))
+
+    @contextmanager
+    def freeze_at_hour(self, hour):
+        """Freeze the time when creating content with builders, so that
+        we can rely on consistent creation times.
+        Since we can sort consistently when all objects have the exact same
+        creation times we need to move the clock forward whenevery things are
```
deiferni: whenever

opengever/testing/fixtures.py
```diff
@@ -0,0 +1,43 @@
...
+        created, using ftw.builder's ticking creator in combination with a
+        freezed clock.
```
deiferni: frozen

opengever/testing/fixtures.py
```diff
@@ -0,0 +1,43 @@
...
+
+        In order to be able to insert new objects in the fixture without
+        scrambling up all timestamps, we group the builders and let each group
```
deiferni: scrambling -> messing/screwing?

opengever/core/testing.py
```diff
@@ -380,3 +384,114 @@ def setUpPloneSite(self, portal):
...
+    a new transaction.
+    """
+
+
+    defaultBases = (COMPONENT_REGISTRY_ISOLATION, )
+
+    def __init__(self):
+        # By super-super-calling the __init__ we remove the SQL bases,
+        # since we are implementing SQL setup in this layer directly.
+        PloneSandboxLayer.__init__(self)
```
jone: Background:
The primary reason for extracting SQL into separate layer was to make it possible to run the tests against PostgreSQL, so that we can isolate between tests, which we believed was not possible with sqlite.
But I've now realized that we can achieve test isolation with the in-memory sqlite database too, and therefore have no need to switch to PostgreSQL.

But we might want to switch to PostgreSQL because of other reasons, such as using a database used in production.

I'm not sure at this point how we will proceed and it is not relevant for this pull request. That's why I decided to add this workaround and postpone this discussion.

PR #3047, opened 2017-06-21
Reworked contact syncer functionality to reduce duration time

phgross:
The new implementation does no longer use the ORM but use SQLAlchemy core functionality and  bulk inserts/updates. This leads to a essential performance win.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -12,6 +12,7 @@ Changelog
   do its job, and reduces growth of memory usage during import. [lgraf]
 - Bundle import: Display current memory usage (RSS) in progress logger. [lgraf]
 - Word meeting: show proposal files in meeting view. [jone]
+- Reworked contact syncer functionality, to improve speed with bulk insert and updates. [phgross]
 - Format line breaks in task descriptions. [tarnap]
 - SPV word: checkout proposal document after creating proposal. [jone]
 
--- a/opengever/contact/syncer/object_syncer.py
+++ b/opengever/contact/syncer/object_syncer.py
@@ -147,6 +147,7 @@ class OrganizationSyncer(SQLObjectSyncer):
     model = Organization
     default_values = {'contact_type': 'organization'}
     attributes = {'name': 'name',
+                  'description': 'description',
                   'former_contact_id': 'former_contact_id',
                   'is_active': lambda row: bool(getattr(row, 'is_active'))}
 
@@ -179,6 +180,7 @@ class PersonSyncer(SQLObjectSyncer):
                   'academic_title': 'title',
                   'firstname': 'firstname',
                   'lastname': 'lastname',
+                  'description': 'description',
                   'former_contact_id': 'former_contact_id',
                   'is_active': lambda row: bool(getattr(row, 'is_active'))}
 
--- a/opengever/contact/syncer/object_syncer.py
+++ b/opengever/contact/syncer/object_syncer.py
@@ -2,16 +2,21 @@ from AccessControl.SecurityInfo import ClassSecurityInformation
 from ftw.upgrade import ProgressLogger
 from opengever.base.model import create_session
 from opengever.contact.models import Address
-from opengever.contact.models import Contact
 from opengever.contact.models import MailAddress
 from opengever.contact.models import Organization
 from opengever.contact.models import OrgRole
 from opengever.contact.models import Person
 from opengever.contact.models import PhoneNumber
 from opengever.contact.models import URL
+from Products.CMFPlone.utils import safe_unicode
+from sqlalchemy.orm import aliased
+from sqlalchemy.sql import select
+from sqlalchemy.sql.expression import column
+from sqlalchemy.sql.expression import join
+from sqlalchemy.sql.expression import table
 from urlparse import urlparse
+from zope.sqlalchemy.datamanager import mark_changed
 import logging
-import transaction
 
 
 logger = logging.getLogger('opengever.contact')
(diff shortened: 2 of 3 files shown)


opengever/contact/syncer/object_syncer.py
```diff
@@ -44,321 +49,317 @@ def __call__(self):
...
+        if to_insert or to_update:
+            mark_changed(self.db_session)
 
-        logger.info("Syncing of {} succeeded".format(self.type_name))
-        self.log_statistic()
+    def get_identifier(self, row):
+        return getattr(row, self.unique_key)
 
-        transaction.commit()
+    def preparing_values(self, result, existing):
```
deiferni: prepare (should be imperative)

docs/HISTORY.txt
```diff
@@ -78,6 +79,7 @@ Changelog
 - Add reference column to document listings. [tarnap]
 - Do not display templates without an assigned file in the CreateDocumentFromTemplate form. [tarnap]
 - Add dossier and subdossier link in task listings. [elioschmutz]
+- Include missing permission definitions from Products.CMFCore. [phgross]
```
jone: Should this be part of this PR or is it a rebase artefact?

opengever/contact/syncer/object_syncer.py
```diff
@@ -2,23 +2,28 @@
...
 import logging
-import transaction
 
 
 logger = logging.getLogger('opengever.contact')
 logger.setLevel(logging.INFO)
 
 
-class ObjectSyncerProgressLogger(ProgressLogger):
+class SQLObjectSyncerProgressLogger(ProgressLogger):
```
jone: What does `SQL` mean in this context?
If I understand it correctly, this class is just a progress logger which does not know the amount of items. Actually, it cannot log any progress, it just logs items 🤔 
I just think it does not know about SQL or the object syncer.

opengever/contact/syncer/object_syncer.py
```diff
@@ -43,322 +48,350 @@ def __call__(self):
         self.logger.info('%s: %s' % (self._counter, self.message))
 
 
-class ObjectSyncer(object):
-    """Use the Objectsyncer to sync an sql-table with data form separate DB.
+class SQLObjectSyncer(object):
+    """The SQLObjectSyncer supports syncing an sql-table with data form
```
jone: `form` => `from`

opengever/contact/syncer/object_syncer.py
```diff
@@ -43,322 +48,350 @@ def __call__(self):
...
-
-    type_name = ""
-
-    total_items = 0
-
-    internal_object = None
-    remote_object = None
+    # Needs to be defined by the subclass
+    # ORM Mapper class
+    mapper = None
```
jone: Is `mapper` the correct term here?
I was confused about this name until I realized that it contains the model object.
How about naming it `model`?

opengever/contact/syncer/object_syncer.py
```diff
@@ -43,322 +48,350 @@ def __call__(self):
...
-            else:
-                self.add_object(remote_object)
-
-        logger.info("Syncing of {} succeeded".format(self.type_name))
-        self.log_statistic()
+        to_insert, to_update = self.prepare_values(result, existing)
 
-        transaction.commit()
+        self.db_session.bulk_insert_mappings(
+            self.mapper, to_insert, return_defaults=True)
```
jone: Why is `return_defaults` necessary?
It's told that it reduces performance and does only make sense when multiple tables are involved.
But I'm no SQLAlchemy expert 😏 
http://docs.sqlalchemy.org/en/latest/orm/session_api.html#sqlalchemy.orm.session.Session.bulk_insert_mappings.params.return_defaults

opengever/contact/syncer/object_syncer.py
```diff
@@ -43,322 +48,350 @@ def __call__(self):
...
+    # Needs to be defined by the subclass
+    # ORM Mapper class
+    mapper = None
+    # attributename/value pairs which are set for all objects.
+    fixed_values = {}
+    # columnname/value pair of all attributes to update/add
+    attributes = {}
+    # column_name used destinct object selection
+    unique_key = None
+    primary_key = None
```
jone: I realized that I'm confused by the names `unique_key` and `primary_key`. If I understand it correctly, I would describe those as:
`unique_key` => `source_id_column`: the name of the column where the ID of the source system is stored
`primary_key` => `gever_id_column`: the name of the column where our ID for the row is stored.

Is this correct or am I understanding it wrong?

opengever/contact/syncer/object_syncer.py
```diff
@@ -43,322 +48,350 @@ def __call__(self):
...
-        """Starts the sync-process
-        """
-        logger.info("Syncing {}...".format(self.type_name))
-        self.reset_statistic()
-
-        self.initalize_existing_mapping()
-        self.initalize_contact_id_mapping()
+        logger.info('Start syncing {}'.format(self.mapper.__tablename__))
 
+        existing = self.get_existing_mapping()
```
jone: Maybe `existing_id_lookup`?

opengever/contact/syncer/object_syncer.py
```diff
@@ -43,322 +48,350 @@ def __call__(self):
...
-
-    total_items = 0
-
-    internal_object = None
-    remote_object = None
+    # Needs to be defined by the subclass
+    # ORM Mapper class
+    mapper = None
+    # attributename/value pairs which are set for all objects.
+    fixed_values = {}
```
jone: I'd rename that to `default_values`: those values can actually be overwritten by keys defined `attribues` below.

opengever/contact/syncer/object_syncer.py
```diff
@@ -43,322 +48,350 @@ def __call__(self):
...
-        the current row.
-        """
-        raise NotImplementedError()
-
-    def get_remote_object(self, row):
-        """Creates and returns a new object (without adding it
-        to the db) instance with the given attributes in the row.
-        """
-        raise NotImplementedError()
+    def get_identifier(self, row):
```
jone: Because we work with source-rows and with our rows, I'd be more specific in the naming so that the reader intuitively knows which kind of data we talk of.
Maybe `source_row` and `gever_row`?

opengever/contact/syncer/object_syncer.py
```diff
@@ -43,322 +48,350 @@ def __call__(self):
...
+    def finalize_insert_data(self, data, row):
+        return data
 
-        return value
+    def is_existing(self, row, existing):
+        return self.get_identifier(row) in existing
 
+    def get_existing_mapping(self):
+        """Returns a dict with `primary_key` `unique_key` pairs for each
+        currently existing row.
```
jone: Think not the _keys_ are returned, but a lookup table of the _values_.
See also the discussion above about the name of `primary_key` and `unique_key`.

I'd probably describe it something like:
> Returns a lookup-table of all rows currently stored in gever.
> The lookup table is a dict where the key is the source-id and the value is the gever-id.

opengever/contact/syncer/object_syncer.py
```diff
@@ -43,322 +48,350 @@ def __call__(self):
...
+            data, row, existing)
 
-class MailSyncer(ObjectSyncer):
+        # Because Person is a inherited Model we also need to pass in the
+        # id as the person primary key.
+        data['person_id'] = existing[self.get_identifier(row)]
+        return data
 
-    type_name = "Mails"
+class ContactAdditionsSyncer(SQLObjectSyncer):
```
jone: I'd be great to have a docstring here, telling what an additions syncer is and when / for which purpose it is used.

opengever/contact/syncer/object_syncer.py
phgross: Because of the way we currently define defaults and autoincrements, the `return_defaults` is necessary. This is a consequence of the behavior that we must support multiple backend types. 😏

phgross:
@jone I've updated the PR according to your comments, please have another look ...

PR #3048, opened 2017-06-21
Implement and enable redirector eTag adapter.

### LINKED ISSUE #3026: Prepared downloads via redirector should flush cache
When a document gets checked out and the *.zem download is prepared via [Redirector](https://github.com/4teamwork/opengever.core/blob/master/opengever/base/redirector.py#L56), the cache has to be flushed, otherwise the file is not downloaded until the user opens a not cached site. 

Solution:
 - [ ] Include the redirector cookie in the cachekey, with a separate ETag Adapter.

phgross:
Flush cache when preparing a download via Redirector adapter.

Fixes #3026

### DIFF
--- a/opengever/base/redirector.py
+++ b/opengever/base/redirector.py
@@ -1,8 +1,12 @@
 from five import grok
 from opengever.base.interfaces import IRedirector
+from plone.app.caching.interfaces import IETagValue
 from plone.app.layout.viewlets.interfaces import IAboveContentTitle
+from zope.component import adapts
+from zope.interface import implements
 from zope.interface import Interface
 from zope.publisher.interfaces.browser import IBrowserRequest
+import hashlib
 import json
 
 
@@ -26,7 +30,7 @@ class RedirectorCookie(object):
             self._set([])
         return value
 
-    def _get(self):
+    def get_value(self):
         # Load from response cookie in case this request added or updated the
         # cookie
         cookie = self.request.response.cookies.get(REDIRECTOR_COOKIE_NAME, {})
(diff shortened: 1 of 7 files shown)


opengever/base/redirector.py
```diff
@@ -122,3 +131,22 @@ def render(self):
...
+
+    def compute_md5hash(self, value):
+        m = hashlib.md5()
+        m.update(value.encode('utf-8'))
+        return m.hexdigest()
+
+    def __call__(self):
+        value = RedirectorCookie(self.request)._get_value()
+        if value:
+            return self.compute_md5hash(value)
```
deiferni: i'd add explicit returns for all paths for readability, so either add `return None` at the end or implement guard-clause style (which i'd prefer):

```python
value = RedirectorCookie(self.request)._get_value()
if not value:
    return None
return self.compute_md5hash(value)
```

opengever/base/redirector.py
```diff
@@ -26,7 +30,7 @@ def read(self, remove=True):
             self._set([])
         return value
 
-    def _get(self):
+    def _get_value(self):
```
deiferni: this is still a protected method, but now it is used publicly. i'd drop the leading `_`.

opengever/base/tests/test_redirector.py
```diff
@@ -98,3 +102,30 @@ def test_viewlet_only_redirects_once(self):
...
+
+        self.assertIsNone(self.get_etag_value_for(document))
+
+        # register redirect
+        redirector = IRedirector(self.request)
+        redirector.redirect(
+            'http://nohost/plone/document-1/externaledit',
+            target='named-window')
+
+        self.assertEquals('bcda738beecd660241c93d98dc016a9c',
```
deiferni: think it is a bit brittle to assert against the cookie hash here, may be good enough to assert that we actually get a non-None value?

phgross:
@deiferni updated ...

PR #3051, opened 2017-06-22
Create document from template using office connector

### LINKED ISSUE #3044: Dokument aus Vorlage öffnet Office Connector mit dem alten Protokoll
Wenn ein Dokument aus einer Vorlage erstellt wird, so wird dies anschliessend mit dem alten office Connector Protokoll geöffnet. Es kann somit nicht direkt eingecheckt werden nach dem Beenden.

tarnap:
I'm not totally certain about the second commit, which refactors some code:

https://github.com/4teamwork/opengever.core/commit/b7fe2f1778eef5aeabedec660045eb4a5f898e8e#diff-f6f40726c157ab858647bea6cd2a2b37L77
https://github.com/4teamwork/opengever.core/commit/b7fe2f1778eef5aeabedec660045eb4a5f898e8e#diff-f6f40726c157ab858647bea6cd2a2b37L83
and
https://github.com/4teamwork/opengever.core/commit/b7fe2f1778eef5aeabedec660045eb4a5f898e8e#diff-f6f40726c157ab858647bea6cd2a2b37R74

I dislike the fact that we're using `objects[0]` and `self.context` first and use only `objects[0]` afterwards... is it the same object?
Maybe @Rotonen can explain what's going on there.

Resolves #3044

### DIFF
--- a/opengever/document/document.py
+++ b/opengever/document/document.py
@@ -7,6 +7,7 @@ from collective import dexteritytextindexer
 from five import grok
 from ftw.mail.interfaces import IEmailAddress
 from ftw.tabbedview.interfaces import ITabbedviewUploadable
+from opengever.base.interfaces import IRedirector
 from opengever.document import _
 from opengever.document.base import BaseDocumentMixin
 from opengever.document.behaviors.related_docs import IRelatedDocuments
@@ -14,6 +15,8 @@ from opengever.document.interfaces import ICheckinCheckoutManager
 from opengever.dossier.behaviors.dossier import IDossierMarker
 from opengever.meeting.proposal import IProposal
 from opengever.meeting.proposal import ISubmittedProposal
+from opengever.officeconnector.helpers import is_officeconnector_checkout_feature_enabled  # noqa
+from opengever.officeconnector.helpers import create_oc_url
 from opengever.task.task import ITask
 from plone import api
 from plone.autoform import directives as form_directives
@@ -258,3 +261,17 @@ class Document(Item, BaseDocumentMixin):
                 self, transition='document-transition-initialize')
 
         return response
+
+    def setup_external_edit_redirect(self, request):
+        redirector = IRedirector(request)
+        if is_officeconnector_checkout_feature_enabled():
+            redirector.redirect(create_oc_url(
+                request,
+                self,
+                dict(action='checkout'),
+            ))
+        else:
+            redirector.redirect(
+                '%s/external_edit' % self.absolute_url(),
+                target='_self',
+                timeout=1000)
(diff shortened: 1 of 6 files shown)


tarnap:
Forgot to update change log, doing it right now.

tarnap:
After discussing with @Rotonen and @lukasgraf, it is safe to replace `self.context` with `objects[0]`, since `objects[0]` is set to `self.context`.

opengever/officeconnector/helpers.py
```diff
@@ -23,7 +23,8 @@ def is_officeconnector_checkout_feature_enabled():
 def parse_documents(request, context):
     documents = []
 
-    if request['REQUEST_METHOD'] == 'GET':
+    if (request['REQUEST_METHOD'] == 'GET' or
+        request['REQUEST_METHOD'] == 'POST' and 'BODY' not in request):
```
Rotonen: Indent the visual indent here one extra level to avoid a collision between a visual indent of the condition and the functional indent of the conditional block.

opengever/dossier/tests/test_templatefolder.py
```diff
@@ -499,6 +501,26 @@ def test_templates_without_a_file_are_not_listed(self, browser):
...
+        template_word = create(Builder('document')
+                               .titled('Word Docx template')
+                               .within(self.templatefolder)
+                               .with_dummy_content())
+
+        browser.login().open(self.dossier, view='document_with_template')
+        browser.fill({'form.widgets.template': _make_token(template_word),
+                      'Title': 'Test OfficeConnector'}).save()
+
+        self.assertNotEquals(None, re.search(
```
Rotonen: Would `self.assertIn('\'oc:', browser.css('script.redirector').first.text)` work for you here?

At least for me the regex obsfucates what is going on here. Inlining and / or nesting too much into returns or asserts impairs readability - breaking that out for legibility is a fine route.

opengever/document/document.py
```diff
@@ -258,3 +261,17 @@ def UNLOCK(self, REQUEST, RESPONSE):
                 self, transition='document-transition-initialize')
 
         return response
+
+    def setup_redirection_for_external_editing(self, request):
```
Rotonen: `setup_external_edit_redirect()` would fit our existing naming better, but our existing naming is a deep rabbit hole on its own.

opengever/dossier/tests/test_templatefolder.py
tarnap: Is "'oc:" enough?
I can change this, of course :)

Rotonen: Yes, as `:` is not allowed in the [base64 variant we use here](https://docs.python.org/2/library/base64.html#base64.urlsafe_b64encode) as the hash.

tarnap:
@Rotonen could you re-review my pull request?

Rotonen:
@tarnap what's the 'bodyless POST' case for in the OC helper?

tarnap:
@Rotonen Do you mean this one?
 https://github.com/4teamwork/opengever.core/pull/3051/files#diff-649accc1cbd5c56db491463dfc67706dR27

That one is required for this:
use officeconnector

As you can see, there's a form submission (therefore a POST request), but there is no 'BODY' in it, but rather a field containing the object to be opened with office connector.

PR #3052, opened 2017-06-23
Update to latests plone 4.3 version - 4.3.14

### LINKED ISSUE #3030: Update to latest plone 4.3.x version
- [ ] Plone 4.3.14
- [ ] Update pinnings for third party packages
...

phgross:
⚠️  The PR currently contains a separate develop-versions.cfg, I'll remove that and copy it to opengever/develop kgs, as soon the PR is approved.   

The following adjustments had to be made (more informations in the commit messages):
 - Added missing value for our own description `schema.Text` fields, so we have the same behavior than the IBasic behavior from opengever.base and plone.app.dexterity.
 - Also handler default fallback title `no_subject` in the OGMail title initalization.
 - Adjust XLS Reporter to let it work with newest `openpyxl` version (Footer, imports)
 - Remove doubled KB in file widgets. The newer version of plone.formwidget.namedfile the method file_size already returns the size with indication of size.
 - Some additional adjustments on tests, without any change on the functionality.

Closes #3030

### DIFF
--- a/opengever/base/quickupload.py
+++ b/opengever/base/quickupload.py
@@ -73,20 +73,17 @@ class OGQuickUploadCapableFileFactory(grok.Adapter):
     def __init__(self, context):
         self.context = aq_inner(context)
 
-    def __call__(self,
-                 filename,
-                 title,
-                 description,
-                 content_type,
+    def __call__(self, filename, title, description, content_type,
                  data, portal_type):
-
+        """Quickupload description inputs are hidden in gever
+        therefore we skip the description.
+        """
         if self.is_email_upload(filename):
             command = CreateEmailCommand(
-                self.context, filename, data, description=description,
+                self.context, filename, data,
                 message_source=MESSAGE_SOURCE_DRAG_DROP_UPLOAD)
         else:
-            command = CreateDocumentCommand(
-                self.context, filename, data, description=description)
+            command = CreateDocumentCommand(self.context, filename, data)
 
         try:
             obj = command.execute()
--- a/opengever/base/tests/test_quickupload.py
+++ b/opengever/base/tests/test_quickupload.py
@@ -33,7 +33,7 @@ class TestOGQuickupload(FunctionalTestCase):
 
         self.assertEquals('document', content.Title())
         self.assertEquals('text', content.file.data)
-        self.assertIsNone(content.description)
+        self.assertEquals(u'', content.description)
 
     def test_expect_one_journal_entry_after_upload(self):
         content = create(Builder('quickuploaded_document')
--- a/opengever/disposition/disposition.py
+++ b/opengever/disposition/disposition.py
@@ -249,7 +249,11 @@ class Disposition(Container):
                 obj=relation.to_object, transition='dossier-transition-archive')
 
     def destroy_dossiers(self):
+        if not self.dossiers:
+            return
+
         alsoProvides(getRequest(), IDuringDossierDestruction)
+
         dossiers = [relation.to_object for relation in self.dossiers]
         self.set_destroyed_dossiers(dossiers)
         self.check_destroy_permission(dossiers)
--- a/opengever/document/tests/test_overview.py
+++ b/opengever/document/tests/test_overview.py
@@ -307,7 +307,7 @@ class TestDocumentOverview(FunctionalTestCase):
                           archival_file_row.css('th').first.text)
         self.assertEquals('icon-pdf',
                           archival_file_row.css('td span')[0].get('class'))
-        self.assertEquals(u'test.pdf \u2014 0 KB',
+        self.assertEquals(u'test.pdf \u2014 1 KB',
                           archival_file_row.css('td span')[1].text)
 
 
--- a/opengever/base/behaviors/configure.zcml
+++ b/opengever/base/behaviors/configure.zcml
@@ -73,7 +73,6 @@
       title="SequenceNumber"
       description="Adds the sequence number indexer"
       provides=".sequence.ISequenceNumberBehavior"
-      for="plone.dexterity.interfaces.IDexterityContent"
       />
 
   <plone:behavior
--- a/opengever/base/reporter.py
+++ b/opengever/base/reporter.py
@@ -77,7 +77,7 @@ class XLSReporter(object):
         workbook = Workbook()
         sheet = workbook.active
         sheet.title = self.sheet_title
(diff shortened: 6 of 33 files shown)


lukasgraf:
Warnings on instance startup:
```
2017-06-26 09:48:17 WARNING plone.behavior Specifying 'for' in behavior 'SequenceNumber' if no 'factory' is given has no effect and is superfluous.
2017-06-26 09:48:18 WARNING plone.behavior Specifying 'for' in behavior 'Restrict addable dossier templates' if no 'factory' is given has no effect and is superfluous.
2017-06-26 09:48:18 WARNING plone.behavior Specifying 'for' in behavior 'Restricted dossier' if no 'factory' is given has no effect and is superfluous.
2017-06-26 09:48:19 WARNING plone.behavior Specifying 'for' in behavior 'Primary field quota subject' if no 'factory' is given has no effect and is superfluous.

2017-06-26 09:48:19 WARNING plone.behavior Specifying 'for' in behavior 'OpenGever Standard Sharing' if no 'factory' is given has no effect and is superfluous.
2017-06-26 09:48:19 WARNING plone.behavior Specifying 'for' in behavior 'OpenGever Dossier Sharing' if no 'factory' is given has no effect and is superfluous.
2017-06-26 09:48:20 WARNING plone.behavior Specifying 'for' in behavior 'Related items' if no 'factory' is given has no effect and is superfluous.
2017-06-26 09:48:20 WARNING plone.behavior Specifying 'for' in behavior 'Test behavior for keywordwidget Use-Cases' if no 'factory' is given has no effect and is superfluous.
```

phgross:
@lukasgraf I've checked the warnings and removed the superfluous `for` attribute on the behavior configurations (see #26f8613 and https://github.com/4teamwork/ftw.keywordwidget/pull/20).

The warning for the 'Related items' behavior happens because of the behavior configuration in `plone.app.relationfield` and has been removed in a later commit (https://github.com/plone/plone.app.relationfield/commit/8911ec519c1601524004bd19d728f342bb08d5c2) but I'm not sure if we should updated `plone.app.relationfield` from `1.2.3` to a `1.3.x` version, what do you think?

lukasgraf:
@phgross stick with the `1.2.3` version of `plone.app.relationfield` - just addressing these issues in our own codebase is enough for now.

opengever/base/quickupload.py
```diff
@@ -80,13 +80,14 @@ def __call__(self,
                  content_type,
                  data, portal_type):
 
+        # Quickupload description inputs are hidden in gever
```
deiferni: i'd move this info to the method docstring (or in this case, create a method docstring 😉 )

opengever/base/tests/test_quickupload.py
```diff
@@ -33,7 +33,7 @@ def test_set_default_values(self):
 
         self.assertEquals('document', content.Title())
         self.assertEquals('text', content.file.data)
-        self.assertIsNone(content.description)
+        self.assertEquals(u'', content.description)
```
deiferni: As an alternative strategy we could consider to always pass `None` as the description. That way the behaviour would remain exactly the same.

opengever/document/tests/test_overview.py
```diff
@@ -307,7 +307,7 @@ def test_archival_file_is_extended_with_mimetype_class(self, browser):
                           archival_file_row.css('th').first.text)
         self.assertEquals('icon-pdf',
                           archival_file_row.css('td span')[0].get('class'))
-        self.assertEquals(u'test.pdf \u2014 0 KB',
+        self.assertEquals(u'test.pdf \u2014 1 KB',
```
deiferni: and rounding is done different in that case?

lukasgraf:
Warning on each request:
```
WARNING plone.app.caching Could not find value adapter for ETag component redirector
```

opengever/disposition/disposition.py
```diff
@@ -250,6 +250,9 @@ def mark_dossiers_as_archived(self):
 
     def destroy_dossiers(self):
         alsoProvides(getRequest(), IDuringDossierDestruction)
+        if not self.dossiers:
```
deiferni: i would return before you let the request provide `IDuringDossierDestruction`

lukasgraf:
@phgross I've released `ftw.chameleon = 1.1.2`, could you bump that as well please?

opengever/base/tests/test_quickupload.py
phgross: As I wrote in the commit message I would go for an empty string. On the one hand its how all the other descriptions fields (og.base.IBasic and plone IBasic) work and on the other hand now all different ways to create an object (quickupload, api, invokeFactory, createContenInContainer) works exactly the same.

lukasgraf:
@phgross we should also consider bumping `plone.app.jquery` to `1.11.2` - probably better in a separate PR though, and after testing the plain Plone 4.3 stuff.

develop-versions.cfg
```diff
@@ -0,0 +1,200 @@
...
+sablon = 0.0.21
+
+[versions]
+# buildout / setuptools
+setuptools = 33.1.1
+zc.buildout = 2.5.0
+
+# Plone overrides
+
+# was 3.0.17 needed by plone.app.discussion ?
```
lukasgraf: Nope. `3.0.17` just was the most recent one the last time we bumped it - `3.0.19` should be perfect for us.

opengever/disposition/disposition.py
```diff
@@ -250,6 +250,9 @@ def mark_dossiers_as_archived(self):
 
     def destroy_dossiers(self):
         alsoProvides(getRequest(), IDuringDossierDestruction)
```
lukasgraf: Commit message: `descruction` -> `destruction`

opengever/document/tests/test_overview.py
phgross: Yes, but a file size of 0 KB, even if the file contains only some bytes, is confusing. Therefore that change makes sense to me.

phgross: See `zope.size.byteDisplay` for more information.

phgross:
@deiferni @lukasgraf I've updated the PR. As soon you approve it, I'll move the pinnings to the KGS and drop the pinning commits in this PR.

maethu:
@phgross IMHO this change would deserve a changelog entry 😉

PR #3055, opened 2017-06-23
Add related documents method to document overview

tarnap:
... also improve overview's `get_css_class` method.

When adding / editing a document (let's say A), related documents
(sat B and C) can be selected. These are listed under "related
documents" in the document overview (A lists B and C)..
Those related objects should list the documents which are related to
them as well (B and C should display A as well).

related documents

TODO: Add CSS assets to style the list of documents

Resolves: #1498 
See also: https://github.com/4teamwork/plonetheme.teamraum/pull/559

### DIFF
--- a/opengever/document/browser/overview.py
+++ b/opengever/document/browser/overview.py
@@ -1,5 +1,4 @@
 from AccessControl import getSecurityManager
-from Acquisition import aq_inner
 from five import grok
 from ftw import bumblebee
 from opengever.base import _ as ogbmf
@@ -24,11 +23,9 @@ from plone.i18n.normalizer.interfaces import IIDNormalizer
 from Products.CMFCore.utils import getToolByName
 from Products.MimetypesRegistry.common import MimeTypeException
 from z3c.form.browser.checkbox import SingleCheckBoxWidget
-from zc.relation.interfaces import ICatalog
 from zope.browserpage.viewpagetemplatefile import ViewPageTemplateFile
 from zope.component import getUtility
 from zope.component import queryMultiAdapter
-from zope.intid.interfaces import IIntIds
 
 
 class BaseRow(object):
@@ -201,19 +198,15 @@ class Overview(DisplayForm, GeverTabMixin, ActionButtonRendererMixin):
 
     def linked_documents(self):
         """Returns a list documents related to the context document."""
-        catalog = getUtility(ICatalog)
-        doc_id = getUtility(IIntIds).getId(aq_inner(self.context))
-        related_objects = self.context.related_items() + [
-            r.from_object
-            for r in catalog.findRelations(
-                {'to_id': doc_id, 'from_attribute': 'relatedItems'}
-            )
-        ]
+
         return [{
             'class': self.get_css_class(obj),
             'title': obj.Title(),
             'url': obj.absolute_url(),
-        } for obj in sorted(related_objects, key=lambda obj: obj.Title())]
+        } for obj in sorted(
+            self.context.related_items(bidirectional=True),
+            key=lambda obj: obj.Title()
+        )]
 
     def is_outdated(self, submitted_document):
         return not submitted_document.is_up_to_date(self.context)
--- a/opengever/document/browser/overview.py
+++ b/opengever/document/browser/overview.py
@@ -1,4 +1,5 @@
 from AccessControl import getSecurityManager
+from Acquisition import aq_inner
 from five import grok
 from ftw import bumblebee
 from opengever.base import _ as ogbmf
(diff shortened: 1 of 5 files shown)


tarnap:
https://docs.plone.org/adapt-and-extend/theming/templates_css/css.html

opengever/document/browser/templates/related_documents.pt
```diff
@@ -0,0 +1,12 @@
+<tal:block xmlns:i18n="http://xml.zope.org/namespaces/i18n"
+    i18n:domain="opengever.document">
+
+    <ul class="related_documents">
+        <li tal:repeat="doc view/linked_documents">
+            <a tal:attributes="href doc/absolute_url;
+                               class python:view.get_css_class(doc)"
```
jone: You could do that in `linked_documents` already, so that you have no method calls from within the page template.
Trying to avoid `python:`-statements in page templates is a good habit (when reasonalby achievable).

opengever/document/browser/overview.py
```diff
@@ -193,6 +199,21 @@ def get_update_document_url(self, submitted_document):
                     submitted_document.document_id
                     )
 
+    def linked_documents(self):
+        """Returns a list of dicts representing incoming and outgoing
+        references to/from other dossiers.
+        """
+
+        self.context.related_items()
```
jone: What's the purpose of that?

opengever/document/tests/test_overview.py
```diff
@@ -36,6 +36,24 @@ def tearDown(self):
...
+                            .relate_to(self.doc_a))
+        self.doc_c = create(Builder('document')
+                            .having(title=u'C')
+                            .relate_to(self.doc_b))
+
+        browser.login().open(self.doc_b, view='tabbedview_view-overview')
+
+        self.assertEquals(
+            sorted([self.doc_a.title, self.doc_c.title]),
+            sorted([e.text for e in browser.css('ul.related_documents a')])
```
jone:
```diff
-            sorted([e.text for e in browser.css('ul.related_documents a')])
+            sorted(browser.css('ul.related_documents a').text)
```

jone: Is the sort order really random? Why is it random?

opengever/document/browser/overview.py
tarnap: It has no purpose, it's leftover code.

opengever/document/browser/templates/related_documents.pt
tarnap: Ok, thanks!

opengever/document/browser/overview.py
```diff
@@ -193,6 +199,22 @@ def get_update_document_url(self, submitted_document):
                     submitted_document.document_id
                     )
 
+    def linked_documents(self):
+        """Returns a list documents related to the context document."""
```
deiferni: ↩️

opengever/document/browser/overview.py
```diff
@@ -193,6 +199,22 @@ def get_update_document_url(self, submitted_document):
...
+        catalog = getUtility(ICatalog)
+        doc_id = getUtility(IIntIds).getId(aq_inner(self.context))
+        related_objects = self.context.related_items() + [
```
deiferni: this could be placed on the model, i.e. amended to the `related_items` method on documents/mails/....

opengever/document/tests/test_overview.py
```diff
@@ -36,6 +36,24 @@ def tearDown(self):
         super(TestDocumentOverview, self).tearDown()
 
     @browsing
+    def test_overview_displays_related_documents(self, browser):
+        self.doc_a = create(Builder('document')
+                            .having(title=u'A'))
```
deiferni: i'd use titles with umlauts or other non-ascii characters here

opengever/document/browser/overview.py
tarnap: `doc.related_items()` is used in `opengever/trash/remover.py`.

If this logic is implemented in the document model, it might have an unwanted impact in remover's functionality.

As far as I remember @phgross told me to implement this in the view.
What was the reason for this?

I'm improving this and will add it in a further commit in the same pull request. I will link it to this comment as soon as its done to discuss it.

opengever/document/browser/overview.py
tarnap: The empty line was added :)

opengever/document/tests/test_overview.py
tarnap: The related documents are  now sorted in `linked_documents`.
The `sorted` calls are not necessary here anymore.

tarnap: Thanks :+1:

tarnap:
After styling:
related_documents

opengever/document/browser/overview.py
deiferni: adding a keyword argument that does not change the current (default) behaviour seems viable to me 👍

tarnap:
@deiferni @jone could you re-review my changes?

deiferni:
@tarnap tests are red 😢

tarnap:
@deiferni they're not anymore (merci @phgross)

tarnap:
@jone could you re-review my PR?

(btw, removed the comments of the red tests which weren't red after a while)

opengever/document/browser/overview.py
jone: @tarnap I think the idea was to insert the newline before the closing quotes:
```python
    def linked_documents(self):
        """Returns a list documents related to the context document.
        """
        return ...
```

```
"""Use three quotes
for multiline strings"""

"Use one quote for single line strings"
```

tarnap:
Ok, got it! Will do!

PR #2204, opened 2016-09-30
Contact view improvements 

phgross:
- Change Person label to `Name Surname`: According to the user label, also the person label should use the format `Name Surname` as every where else in GEVER. Closes #2190.
- Sort participations alphabetically on participant's title, closes #2189.
- Sort persons alphabetically on lastname, firstname in the organization detail view.
- Group related persons by active state in the organization detail view closes #2191

PR #3057, opened 2017-06-26
Sort meeting participants alphabetically.

deiferni:
The meeting participants are inserted into the protocol unsorted at the moment. This PR changes that behaviour and sorts the meeting participants by their name in the protocols participant list. 

It also swaps firstname/lastname of Members so that they now display the name as `Lastname Firstname` to be consistent with https://github.com/4teamwork/opengever.core/pull/2204 and thus the rest of GEVER.


opengever/meeting/tests/test_protocol_json_data.py
```diff
@@ -96,12 +96,13 @@ def test_add_members_handles_participants_are_no_longer_committee_memberships(se
         create_session().delete(self.membership_peter)
 
         self.assertEquals(
-            {'members': [{'fullname': u'Peter M\xfcller',
+            {'members': [{'fullname': u'Franz M\xfcller',
```
phgross: Does the test really fail, when sorting is removed?

Sorting tests with only two items is a bit fragile, I would add third member.

deiferni: it is different from the insertion order, so i'm quite sure it works now. it may break when the fixture is changed so i'll amend a third member for safety 😉 .

deiferni:
@phgross updated

PR #3063, opened 2017-06-27
Update testing documentation in readme

jone:
I have updated the testing documentation in the README:

- Remove the old gever-testbrowser documentation; `ftw.testbrowser` shall be used.
- Add a section about functional and integration testing.
- Add an example integration test case.
- Add some best practices.

@4teamwork/gever please have a look.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -353,6 +353,9 @@ Once a new policy has been generated the following things need to be added manua
 Tests
 -----
 
+Parallelisation
+~~~~~~~~~~~~~~~
+
 Use ``bin/mtest`` for running all test in multiple processes. Alternatively ``bin/test`` runs the tests in sequence.
 The multi process script distributes the packages (e.g. ``opengever.task``, ``opengever.base``, etc) into multiple processes,
 trying to balance the amount of test suites, so that it speeds up the test run.
--- a/README.rst
+++ b/README.rst
@@ -368,6 +368,210 @@ The ``bin/mtest`` script can be configured with environment variables:
   The colorization is useful for the visual separation of the output of the various processes,
   but it is not useful in a environment without color support.
 
+Functional or integration testing?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+We are shifting the tests from the older functional testing layer to the newer
+integration testing layer.
+
+**Integration testing:**
+
+- Should be used for new tests!
+- Comes with a preinstalled `testing fixtures`_.
+- Transactions are disabled for isolation purposes: `transaction.commit` is not allowed in tests.
+- Uses ``ftw.testbrowser``'s ``TraversalDriver``.
+
+**Functional testing:**
+
+- Should *not be used* for new tests, when possible.
+- Is factory-based, using ``ftw.builder``.
+- Uses transactions.
+- Limited / slow database isolation: a fresh setup is necessary for each test.
+
+
+Example integration test with browser:
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code:: python
+
+   from ftw.testbrowser import browsing
+   from ftw.testbrowser.pages import statusmessages
+   from opengever.testing import IntegrationTestCase
+
+   class TestExampleView(IntegrationTestCase):
+
+       @browsing
+       def test_example_view(self, browser):
+           self.login(self.dossier_responsible)
+           browser.login(self.dossier_responsible)
+           browser.open(self.dossier, view='example_view')
+           statusmessages.assert_no_error_messages()
+
+
+Best practices
+~~~~~~~~~~~~~~
+
+These best practices apply to the new **integration testing** layer.
+
+Do not commit the transaction
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Committing the transaction will break isolation.
+The testing layer will prevent you from interacting with the transaction.
+
+Use the fixture objects
+^^^^^^^^^^^^^^^^^^^^^^^
+
+The `testing fixtures`_ create content objects, users, groups and client
+configurations (admin units, org units) which are available for all tests.
+They can and should be modified to the needs of the test.
+
+Avoid creating objects (with ``ftw.builder``)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Creating objects with ``ftw.builder`` or with ``ftw.testbrowser`` is expensive
+because it takes a moment to index the object.
+Therefore we want to avoid creating unnecessary objects within the tests
+so that the tests are faster overall.
+
+Tests which have the job to test object creation (e.g. through the browser)
+obviously need to actually create an object, all other tests should try to
+reuse objects from the fixture and modify them as needed.
+
+Use users from fixture
+^^^^^^^^^^^^^^^^^^^^^^
+
+The fixture provides a set of standard users which should be used in tests.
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -365,63 +368,221 @@ The ``bin/mtest`` script can be configured with environment variables:
...
+``Manager``-user, the ``plone.app.testing``'s site owner may be used.
+
+Loggin users.
+^^^^^^^^^^^^^
+
+Integration testing tests start with a low-level security.
+Use the ``self.login`` method for setting up the needed security within the
+test code.
+
+TODO: Document browser login behavior https://4teamwork.slack.com/archives/C060FEYKA/p1498490241485990
```
jone: 🚧

README.rst
```diff
@@ -365,63 +368,221 @@ The ``bin/mtest`` script can be configured with environment variables:
...
 
-The center of the `Browser API` is the ``OGBrowser`` class. It's a
-simple subclass of ``plone.testing.z2.Browser`` and the easiest way to
-use it is to extend ``opengever.testing.FunctionalTestCase``:
+Use ``tearDown`` carefully.
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Do not tear down changes which are taken care of by some kind of isolation:
+
+- Do *not* tear down ZODB changes: the ZDOB is isolated by ``plone.app.testing``.
```
lukasgraf: `ZDOB` -> `ZODB`

README.rst
```diff
@@ -365,63 +368,221 @@ The ``bin/mtest`` script can be configured with environment variables:
...
+- Do *not* tear down SQL changes: we take care of that in the SQL testing layer
+  with savepoints / rollbacks.
+- Do *not* tear down component registry changes (e.g. new adapters, utilities,
+  event handlers) as this is taken care of by the
+  `COMPONENT_REGISTRY_ISOLATION`_ layer.
+- *Do* tear down modifications in evnironment variables (``os.environ``).
```
lukasgraf: `evnironment` -> `environment`

README.rst
```diff
@@ -365,63 +368,221 @@ The ``bin/mtest`` script can be configured with environment variables:
...
+.. code:: python
+
+    def test_closing_dossier(self):
+        self.assertTrue(self.dossier.is_open(),
+                        'Precondition: assumed dossier to be open')
+        self.dossier.close()
+        self.assertFalse(self.dossier.is_open())
+
+If the ``self.dossier`` is changed to be not open by default anymore, the failure
+should tell us that a procondition was no longer met rather than the ``close()``
```
lukasgraf: `procondition` -> `precondition`, `rather than` -> `rather than implying that`

README.rst
```diff
@@ -365,63 +368,221 @@ The ``bin/mtest`` script can be configured with environment variables:
...
+
+.. code:: python
+
+    def test_title_is_journalized_on_action(self):
+        self.dossier.title = u'The dossier'
+        action(self.dossier)
+        self.assertEquals(u'The dossier',
+                          last_journal_entry(self.dossier).title)
+
+Activating feature flags.
```
lukasgraf: I would leave the period at the end of those sections off (most of them are not full sentences). </nitpick>

README.rst
```diff
@@ -365,63 +368,235 @@ The ``bin/mtest`` script can be configured with environment variables:
...
+The fixture provides a set of standard users which should be used in tests.
+Do not use ``plone.app.testing``'s test user with global roles as it does
+not reflect properly how the security model of GEVER works.
+In order to test features which can only be executed by the system or by a
+``Manager``-user, the ``plone.app.testing``'s site owner may be used.
+
+Login
+^^^^^
+
+Integration testing tests start with a low-level security.
```
jone: > Im not 100% sure what you mean by this sentence. "Security" as in Security Manager as in.. User? I.e. `low-level security == user with minimal privileges`? Doesn't make sense to me though, because by default (without `login()`) it would be `Anonymous`, right?

Current implementation:
- in the test-code, the security manager is authorized as "self.reguler_user": a user who simply is in the users group of the client, no more responsibilities (see [test case setUp](https://github.com/4teamwork/opengever.core/blob/master/opengever/testing/integration_test_case.py#L25) and [fixtures](https://github.com/4teamwork/opengever.core/blob/master/opengever/testing/fixtures.py#L66-L67)),
- but the browser is not logged in at all (anonymous), as it is default in ``ftw.testbrowser``

I would like to change the behavior so that no user is logged in at all when a tests begins, because:
- as a test reader, I want to see for each test with which user it runs (as in BBD `As a dossier responsible, In order to .., I want to ..`)
- from a security / QA perspective I want to lead devs to making an active deicision for who this test should be
- as a developer I would like to have consistency between test-code and browser-code regarding who is the user under test.

🚧 I will come up with a PR separately which changes the code

lukasgraf: Ah, I see, thanks for clarifying! 👍  (and I agree with the proposed future changes).
I would rephrase that paragraph to something like that then:

> Integration tests start with logging in / assuming the security context of a user with minimal privileges (`self.regular_user`) on the server side. On the client side (browser), no user is logged in by default (Anonymous).
>
> Use the ``self.login`` method for setting up the needed security context within the test code.
[...]
```

Comment (jone) [inline README.rst]: Ok, I like your suggestion and I've amended it.
I'll work on changing the defaults as explained above and then change this part again in order to make this PR mergeable already.

File: README.rst [hunk 7]
```diff
@@ -365,63 +368,235 @@ The ``bin/mtest`` script can be configured with environment variables:
...
+   from ftw.testbrowser import browsing
+   from ftw.testbrowser.pages import statusmessages
+   from opengever.testing import IntegrationTestCase
 
+   class TestExampleView(IntegrationTestCase):
 
-Then you can use the ``Builder`` function in your test cases:
+       @browsing
+       def test_example_view(self, browser):
+           self.login(self.dossier_responsible)
```
deiferni: 😄 
```python
self.login(self.dossier_responsible, browser)
```

README.rst
```diff
@@ -365,63 +368,235 @@ The ``bin/mtest`` script can be configured with environment variables:
...
+reuse objects from the fixture and modify them as needed.
+
+Use users from fixture
+^^^^^^^^^^^^^^^^^^^^^^
+
+The fixture provides a set of standard users which should be used in tests.
+Do not use ``plone.app.testing``'s test user with global roles as it does
+not reflect properly how the security model of GEVER works.
+In order to test features which can only be executed by the system or by a
+``Manager``-user, the ``plone.app.testing``'s site owner may be used.
```
deiferni: should we alias the site-owner as self.manager? should we actively prevent that the discouraged test-user is used?

README.rst
jone: @deiferni The feature, which logs in the browser here, is not yet implemented. I will amend the documentation in the PR implementing the feature.

README.rst
jone: > should we alias the site-owner as self.manager?
I can do that, but I'd like to do that in a separate PR. 😎 

> should we actively prevent that the discouraged test-user is used?
The user has not enough priveleges to do anything by default and `self.grant` is not implemented, so the user is not really usable.
If a dev is still really serious about using test user, we should see and handle that in the review.
I therefore think that we should not spend more time on building guards regarding that 😉

README.rst
jone: @jone this is wrong, it is actually already implemented in `master`.
But it is corrected anyway in the next PR..

jone: => #3072

README.rst
jone: => #3073

PR #3064, opened 2017-06-27
Rename repository to repofolder in fixtures.

jone:
"repository" refers more to the whole tree, including the root. Therefore the term "repository folder" or short "repofolder" is more accurate.


jone:
I would wait with renaming the builder. When moving towards integration testing I expect the builder to be used less and less. At the end, we should only need to change it in the fixture and a handful of tests.

deiferni:
ok i agree, makes perfect sense 👍 .

Issue #3058, opened 2017-06-27
Add copy action to objects which can be copied


phgross:
Currently its only possible to copy a copyable object in its listings, it should also be possible to copy an object on its detail view:

Example for documents:
![bildschirmfoto_2017-06-27_um_08_47_13](https://user-images.githubusercontent.com/485755/27574109-4902871a-5b15-11e7-8f20-8d43fd8f10f4.png)

tarnap:
It's not as easy as it looks!
The standard Plone copy action does not work for us, see:
![copy action](https://user-images.githubusercontent.com/194114/27591694-bcfacc26-5b52-11e7-9e09-72cff4d37fb2.gif)

PR #3068, opened 2017-06-28
Copy single documents

tarnap:
Please take a look at the tests and let me know if more tests are needed.
The second commit is a mere "translation" from `grok.View` to `BrowserView`.

copy paste

Resolves: #3058

### DIFF
--- a/opengever/base/browser/configure.zcml
+++ b/opengever/base/browser/configure.zcml
@@ -108,4 +108,11 @@
       class=".breadcrumb_by_uid.ResolveUIDToBreadcrumb"
       />
 
+  <browser:page
+      name="copy_item"
+      for="plone.dexterity.interfaces.IDexterityContent"
+      class=".copy_items.CopyItemView"
+      permission="zope2.CopyOrMove"
+      />
+
 </configure>
--- a/opengever/base/browser/copy_items.py
+++ b/opengever/base/browser/copy_items.py
@@ -4,6 +4,7 @@ from opengever.base.clipboard import Clipboard
 from plone import api
 from plone.dexterity.interfaces import IDexterityContainer
 from plone.z3cform import layout
+from Products.Five.browser import BrowserView
 
 
 class CopyItemsFormView(layout.FormWrapper, grok.View):
@@ -59,3 +60,22 @@ class CopyItemsFormView(layout.FormWrapper, grok.View):
 
     def are_copyable(self, objs):
         return all(obj.cb_isCopyable() for obj in objs)
+
+
+class CopyItemView(BrowserView):
+
+    def __call__(self):
+        if self.context.cb_isCopyable():
+            Clipboard(self.request).set_objs([self.context])
+            msg = _(u'msg_successfuly_copied',
+                    default=u"Selected objects successfully copied.")
+            msg_type = 'info'
+        else:
+            msg = _(u'error_not_copyable',
+                    default=u"The item you selected cannot be copied.")
+            msg_type = 'error'
+
+        api.portal.show_message(
+            message=msg, request=self.request, type=msg_type)
+
+        return self.request.RESPONSE.redirect(self.context.absolute_url())
--- a/opengever/base/tests/test_copy_paste.py
+++ b/opengever/base/tests/test_copy_paste.py
@@ -29,6 +29,59 @@ class TestCopyItems(FunctionalTestCase):
         self.assertEqual(['Selected objects successfully copied.'], info_messages())
 
 
+class TestCopyItem(FunctionalTestCase):
+
+    def setUp(self):
+        super(TestCopyItem, self).setUp()
+        self.repo, self.repo_folder = create(Builder('repository_tree'))
+        self.dossier = create(Builder('dossier')
+                              .within(self.repo_folder))
+        self.document = create(Builder('document')
+                               .within(self.dossier))
+        self.mail = create(Builder('mail')
+                           .within(self.dossier))
+
+    @browsing
+    def test_statusmessage_if_copy_document_success(self, browser):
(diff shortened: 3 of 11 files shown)


opengever/base/browser/configure.zcml
```diff
@@ -108,4 +108,11 @@
       class=".breadcrumb_by_uid.ResolveUIDToBreadcrumb"
       />
 
+  <browser:page
+      name="copy_item"
+      for="opengever.document.document.IDocumentSchema"
```
deiferni: if we want to also support copying mails (which is probably true 😉 ) this view should be registered for `IBaseDocument` instead.

opengever/base/browser/copy_items.py
```diff
@@ -59,3 +60,23 @@ def extract_selected_objects(self):
 
     def are_copyable(self, objs):
         return all(obj.cb_isCopyable() for obj in objs)
+
+
+class CopyItemView(BrowserView):
+
+    def __call__(self):
+
```
deiferni: ✂️

opengever/base/tests/test_copy_paste.py
```diff
@@ -29,6 +29,36 @@ def test_redirects_back_and_show_statusmessage_if_copy_success(self, browser):
...
+    def setUp(self):
+        super(TestCopyItem, self).setUp()
+
+        self.dossier = create(Builder('dossier'))
+        self.document = create(Builder('document')
+                               .within(self.dossier))
+
+    @browsing
+    def test_statusmessage_if_copy_success(self, browser):
+
```
deiferni: ✂️

opengever/base/tests/test_copy_paste.py
```diff
@@ -29,6 +29,36 @@ def test_redirects_back_and_show_statusmessage_if_copy_success(self, browser):
...
+        browser.login().open(self.document, view='copy_item')
+
+        self.assertEqual(self.document.absolute_url(), browser.url)
+        self.assertEqual(['Selected objects successfully copied.'], info_messages())
+
+    @browsing
+    def test_statusmessage_if_paste_success(self, browser):
+
```
deiferni: ✂️

opengever/core/upgrades/20170628144821_add_copy_action_for_single_documents/actions.xml
```diff
@@ -0,0 +1,21 @@
...
+
+  <!-- ###################################################### /-->
+  <object name="object_buttons" meta_type="CMF Action Category">
+
+    <object name="copy_item" meta_type="CMF Action" i18n:domain="opengever.core">
+      <property name="title" i18n:translate="">Copy Item</property>
+      <property name="description" i18n:translate="" />
+      <property name="url_expr">string:${globals_view/getCurrentObjectUrl}/copy_item</property>
+      <property name="icon_expr" />
+      <property name="available_expr">object.restrictedTraverse("@@plone_interface_info").provides('opengever.document.document.IDocumentSchema')</property>
```
deiferni: this does not match what's in `actions.xml`, have you run the upgrade 😓 ?

opengever/base/browser/configure.zcml
```diff
@@ -108,4 +108,11 @@
...
+      class=".copy_items.CopyItemView"
+      permission="zope2.View"
```
deiferni: there is a copy or move permission, i'd say you should use it here

opengever/core/upgrades/20170628144821_add_copy_action_for_single_documents/actions.xml
tarnap: @deiferni actually, we (@maethu, @tarnap) did.

deiferni: i think it may now work, it misses the `python:` prefix 🤔

maethu: What exactly does not match? The expression seems to be the same?

maethu: haha I delete my prev. comment... it does not match... seems like @tarnap failed on copy&paste 😄

tarnap:
@deiferni is this fine now?

opengever/base/tests/test_copy_paste.py
```diff
@@ -29,6 +29,78 @@ def test_redirects_back_and_show_statusmessage_if_copy_success(self, browser):
         self.assertEqual(['Selected objects successfully copied.'], info_messages())
 
 
+class TestCopyItem(FunctionalTestCase):
+
+    def setUp(self):
+        super(TestCopyItem, self).setUp()
+
+        self.dossier = create(Builder('dossier'))
```
deiferni: i'd create the dossier in a repository and the repository in a repository root. it is good practice to nest content in tests the way it is nested in gever. currently the tests allow nesting that is not allowed otherwise, but we're about  to change that. you can do:

```python
self.repo, self.repo_folder = create(Builder('repository_tree'))
```

tarnap: @deiferni check again? (well... let's see if the tests pass... they do on my machine)

phgross:
@tarnap I'll hijack your branch ...

phgross:
I've add an additional expression to the copy action, so that it's only shown on documents and mails. 

We don't want to provide a copy action for dossiers. Therefore the [mentioned Problem](https://github.com/4teamwork/opengever.core/pull/3068#pullrequestreview-48379849), is not longer a problem for this change. 

The permission on the repositoryfolder has to be checked, when revising the workflows.

PR #3071, opened 2017-06-28
Fix an issue where creating a dossier from template ended in an exception.

elioschmutz:
While creating a dossier from a template, the user fills out fields on different wizard steps. Each step requires a new request. In the first step, the user chooses which dossiertemplate he wants to add, in the second step he fills out the dossier-fields prefilled from the choosen dossiertemplate and then he can save the dossier. We use an object-storage to retrive the dossietemplate in the next request. In some cases, this object looses its acquisitionwrapper which ends in an exception.

To fix this issue, we changed the implementation to use a path-storage and we retrieve the object in each request through traversing.

Issuer: https://extranet.4teamwork.ch/support/kunde-stadt-nidau/tracker/54
See related sentry issue: https://sentry.4teamwork.ch/sentry/onegov-gever/issues/3436/

### DIFF
--- a/opengever/dossier/dossiertemplate/form.py
+++ b/opengever/dossier/dossiertemplate/form.py
@@ -82,6 +82,19 @@ def get_wizard_storage(context):
     return storage.get_data(get_wizard_storage_key(context))
 
 
+def save_template_obj(context, template):
+    set_wizard_storage(context,
+                       {'template_path': '/'.join(template.getPhysicalPath())})
+
+
+def get_saved_template_obj(context):
+    template_path = get_wizard_storage(context).get('template_path')
+    if not template_path:
+        return None
+
+    return api.portal.get().restrictedTraverse(template_path, default=None)
+
+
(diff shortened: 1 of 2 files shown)


opengever/dossier/dossiertemplate/form.py
```diff
@@ -82,6 +82,23 @@ def get_wizard_storage(context):
...
+
+
+def get_saved_template_obj(context):
+    template_path = get_wizard_storage(context).get('template_path')
+    if not template_path:
+        return None
+
+    try:
+        return api.portal.get().restrictedTraverse(template_path)
+    except KeyError:
```
deiferni: this is not necessary, there is a default argument to `restrictedTraverse`:

```python
api.portal.get().restrictedTraverse(template_path, default=None)
```

elioschmutz: 👍

elioschmutz:
@deiferni Done! Please take another look!

PR #3072, opened 2017-06-28
Integration testing: start tests anonymously.

jone:
Tests should start anonymously by default, because:

- as a test reader, I want to see for each test with which user it runs  (as in BBD `As a dossier responsible, In order to .., I want to ..`)
- from a security / QA perspective I want to lead devopers to making an active decision for who this test should be
- as a developer I would like to have consistency between test-code and browser-code regarding who is the user under test.

``self.login`` is now also a context manager which resets to the prior security. This makes it easier to do authenticated preparations in the ``setUp`` method without the side effect the the user is now authorized in the test method.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -450,14 +450,17 @@ In order to test features which can only be executed by the system or by a
 Login
 ^^^^^
 
-Integration tests start with logging in / assuming the security context of a
-user with minimal privileges (``self.regular_user``) on the server side.
-On the client side (browser), no user is logged in by default (Anonymous).
+Integration tests start with *no user logged in*.
+The first thing each test should do, is to log in the user with the fewest
+privileges required for doing the task under test.
 
-Use the ``self.login`` method for setting up the needed security context
-within the test code.
-As a shorthand we can pass the browser instance into ``self.login``, which
-will then log in the browser too.
+The login command should *not* be moved to the ``setUp`` method; it should be
+clearly visible at the beginning of each test, so that a reader has the necessary
+context without scrolling to the top of the file.
+
+When authenticated preparations are required in the ``setUp`` method, use
+``self.login`` as a context manager in order to cleanup the authentication
+on exit, so that the tests still start anonymously.
 
 .. code:: python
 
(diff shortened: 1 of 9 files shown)


jone:
Rebased.

README.rst
```diff
@@ -450,14 +450,17 @@ In order to test features which can only be executed by the system or by a
 Login
 ^^^^^
 
-Integration tests start with logging in / assuming the security context of a
-user with minimal privileges (``self.regular_user``) on the server side.
-On the client side (browser), no user is logged in by default (Anonymous).
+Integration tests start with *no user logged in*.
+The first thing each test does is logging in the user with the fewest privileges
```
lukasgraf: `each test does is logging in ...` -> `each test should do, is to log in ...`

(Make it more clear that this isn't something that's being done automatically by the IntegrationTest layer, but that the developer writing the test should do)

opengever/testing/integration_test_case.py
```diff
@@ -32,20 +33,41 @@ def login(self, user, browser=None):
         >>> self.login(self.dossier_responsible, browser)
         Log in only security manager:
         >>> self.login(self.dossier_responsible)
+
+        The method may also used as context manager, ensuring that after leaving
```
lukasgraf: `also used` -> `also be used`

opengever/testing/integration_test_case.py
```diff
@@ -32,20 +33,41 @@ def login(self, user, browser=None):
...
+        @contextmanager
+        def login_context_manager():
+            try:
+                yield
+            finally:
+                setSecurityManager(security_manager)
+                if browser is not None:
+                    browser.clear_request_header('Authorization')
+                    [browser.append_request_header(name, value)
+                     for (name, value) in browser_auth_headers]
```
lukasgraf: I'm assuming you deliberately wrote this in this way because of recent feedback 😉 , but in this case, I would actually go for something like `map(browser.append_request_header, *browser_auth_headers)` 😉

jone: Because `append_request_header` requires two values it is not easy to use map here.
Your example will result in a `TypeError` or do the wrong thing 😉

PR #3073, opened 2017-06-28
Testing: provide self.manager as site owner.

jone:
Provide the site owner as ``self.manager`` in integration test cases.


jone:
Rebased.

PR #3074, opened 2017-06-29
Fix schema migration EnlargeProfileidInMigrationTrackingTable for oracle backend.

phgross:
Because Oracle is not able to change a Integer column to a BigInteger, when it contains data. I've add a separate implementation for oracle backends to  the `EnlargeProfileidInMigrationTrackingTable` upgradestep. Instead of changing the column type, it now renames the column, add the new one, copies the data and drops the old_column.

### DIFF
--- a/opengever/ogds/base/upgrades/20160211150813_enlarge_profileid_in_migration_tracking_table/upgrade.py
+++ b/opengever/ogds/base/upgrades/20160211150813_enlarge_profileid_in_migration_tracking_table/upgrade.py
@@ -1,5 +1,12 @@
 from opengever.core.upgrade import SchemaMigration
 from sqlalchemy import BigInteger
+from sqlalchemy import Column
+from sqlalchemy import Integer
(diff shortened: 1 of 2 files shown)


opengever/ogds/base/upgrades/20160211150813_enlarge_profileid_in_migration_tracking_table/upgrade.py
```diff
@@ -12,9 +19,68 @@ class EnlargeProfileidInMigrationTrackingTable(SchemaMigration):
     """
 
     def migrate(self):
+        if self.is_oracle:
+            self.migrate_oracle()
+
+        else:
+            self.op.alter_column(
+                table_name='opengever_upgrade_version',
```
deiferni: use the constant here, too

phgross: @deiferni updated ...

PR #3081, opened 2017-06-29
Remove dossier inheritance of templatefolder

### LINKED ISSUE #3078: NoResultFound: No row was found for one()
Der Fehler tritt beim Bearbeiten des Vorlagenordners auf.

https://sentry.4teamwork.ch/sentry/onegov-gever/issues/3486/

```
NoResultFound: No row was found for one()
(18 additional frame(s) were not displayed)
...
  File "home/zope/eggs/z3c.form-3.2.11-py2.7.egg/z3c/form/converter.py", line 279, in toWidgetValue
    return [terms.getTerm(value).token]
  File "home/zope/eggs/z3c.form-3.2.11-py2.7.egg/z3c/form/term.py", line 140, in getTerm
    return super(MissingTermsMixin, self).getTerm(value)
  File "home/zope/eggs/z3c.form-3.2.11-py2.7.egg/z3c/form/term.py", line 35, in getTerm
    return self.terms.getTerm(value)
  File "home/zope/theta.4teamwork.ch/01-dev.onegovgever.ch-fd/src/opengever.core/opengever/ogds/base/sources.py", line 327, in getTerm
    user = self.base_query.filter(User.userid == value).one()
  File "sqlalchemy/orm/query.py", line 2820, in one
    raise orm_exc.NoResultFound("No row was found for one()")

NoResultFound: No row was found for one()
```

- [x] Fix error

elioschmutz:
A templatefolder is a contenttype who is displaying a tabbedview... but it have been implemented as a Dossier.

That means, the templatefolder inherits the whole behavior of a dossier and also its schema.
But we don't want the schema of a dossier and we don't want the behavior of a dossier. We just want a templatefolder with templates.

Why this change?
Initially, the responsible field of a templatefolder was prefilled with
a dummy-user to be able to create an initial templatefolder while
installing opengever.
This was a workaround because the responsible field is required and the IDossier interface defines it and the templatefolder is a dossier :-)

Editing the templatefolder raised an exception because the responsible user
could not be found. One solution would have been to adjust the widget to allow
this deleted/not existing entries but deny from saving not-existing entries. That means,
if you edit the templatefolder, you have to set a new responsible user. But why?

The templatefolder should have had never a responsible field nor all the
other businesscase-dossier-specific fields because it is not a businesscase-dossier.

So in this PR I remove the Dossier-inheritance of a templatefolder. Now, everything seems to be easier...

closes #3078

### DIFF
--- a/opengever/core/profiles/default/registry.xml
+++ b/opengever/core/profiles/default/registry.xml
@@ -106,6 +106,7 @@
   <records interface="ftw.zipexport.interfaces.IZipExportSettings">
     <value key="enabled_dotted_names">
       <element>opengever.dossier.behaviors.dossier.IDossierMarker</element>
+      <element>opengever.dossier.templatefolder.interfaces.ITemplateFolder</element>
       <element>Products.CMFPlone.interfaces.siteroot.IPloneSiteRoot</element>
       <element>opengever.task.task.ITask</element>
     </value>
--- a/opengever/dossier/templatefolder/templatefolder.py
+++ b/opengever/dossier/templatefolder/templatefolder.py
@@ -1,11 +1,14 @@
+from Acquisition import aq_inner
+from Acquisition import aq_parent
 from five import grok
 from opengever.base.behaviors.translated_title import TranslatedTitleMixin
 from opengever.base.browser.translated_title import TranslatedTitleAddForm
 from opengever.base.browser.translated_title import TranslatedTitleEditForm
+from opengever.dossier.dossiertemplate import is_dossier_template_feature_enabled
 from opengever.dossier.templatefolder import ITemplateFolder
+from opengever.meeting import is_meeting_feature_enabled
+from opengever.meeting import is_word_meeting_implementation_enabled
 from plone.dexterity.content import Container
-from Acquisition import aq_parent
-from Acquisition import aq_inner
 
 
 class TemplateFolderAddForm(TranslatedTitleAddForm):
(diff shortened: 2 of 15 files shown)


elioschmutz:
@phgross Ready to review. Please see commit messages to see why I did some changes on tests.

opengever/dossier/templatefolder/templatefolder.py
```diff
@@ -1,33 +1,26 @@
...
     grok.context(ITemplateFolder)
 
 
-class TemplateFolder(DossierContainer, TranslatedTitleMixin):
+class TemplateFolder(Container, TranslatedTitleMixin):
     """Base class for template folder."""
 
     Title = TranslatedTitleMixin.Title
+
+    def is_subdossier(self):
```
phgross: `is_subdossier` is a confusing in that context, i would add a comment and describe who uses a `is_subdossier` method.

opengever/core/profiles/default/registry.xml
```diff
@@ -106,6 +106,7 @@
   <records interface="ftw.zipexport.interfaces.IZipExportSettings">
```
phgross: I'm not happy about the commit message.

You didn't not fixed the tests you enabled zip export for tempaltedossiers:

```
Enable zipexport for Templatedossiers. 

The zip export support for templatedossiers has been disabled since the templatefolder is no longer dossier.
```

elioschmutz: 👍

opengever/dossier/templatefolder/templatefolder.py
elioschmutz: this method is only used in a templatefolder specific event-handler. So I just changed the method-name to `is_subtemplatefolder`.

elioschmutz:
@phgross Please take another look!

deiferni:
When we drop `IDossier` from templatfolder it will also no longer provide `IDossierMarker`. That may lead to unexpected consequences with functionality that is registered for `IDossierMarker` only (after a quick search that seems to be quite a lot), so we should double-check carefully. (As far as i could see this has already happened though …)

elioschmutz:
@deiferni Thats true, it's quite a lot. But the most `IDossierMarker` implementations are for businesscasedossiers and are not necessary for the templatefolder.

The few places where it was necessary are already updated.

elioschmutz:
@phgross could you take another look to this PR? I need it for the PR #3108

phgross:
@elioschmutz FYI: I have to create the 2017.3 release first, before merging this PR ...

PR #3082, opened 2017-06-30
Add version number to journal's download entries

tarnap:
It's not possible to determine which file version the user downloaded from the journal entries.
This is improved by adding the version number (or `current version` if the downloaded version was the working copy) to the title.

version in journal

Resolves: #3075

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2017.3.1 (unreleased)
 ---------------------
 
+- Add version of downloaded file to journal's entry. [tarnap]
 - Expand current subdossier in subdossier tree. [Kevin Bieri]
 - Always download MSG file if avaiable. [mathias.leimgruber]
 - Do not sync comments to a predecessor forwarding. [phgross]
--- a/opengever/document/browser/download.py
+++ b/opengever/document/browser/download.py
@@ -160,5 +160,7 @@ class DocumentDownloadFileVersion(DownloadFileVersion):
 
         self._init_version_file()
         if self.version_file:
-            notify(FileCopyDownloadedEvent(self.context))
+            notify(FileCopyDownloadedEvent(
+                self.context,
+                getattr(self.request, 'version_id', None)))
         return super(DocumentDownloadFileVersion, self).render()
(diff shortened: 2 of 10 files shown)


tarnap:
Need help on the tests:
Where should I put the tests?

Rotonen:
@tarnap https://github.com/4teamwork/opengever.core/blob/master/opengever/document/tests/test_download.py.

tarnap:
@Rotonen isn't it a feature of the journal?

tarnap:
@Rotonen what about mails?

Rotonen:
@tarnap the journal tests are semantically for testing the journaling itself works. If something should leave a journal entry, we test for that there.

https://github.com/4teamwork/opengever.core/blob/master/opengever/mail/tests/test_mail_download_copy.py

tarnap:
@Rotonen thanks!
In that case I'll write tests for documents and mails

tarnap:
@Rotonen @deiferni @phgross @lukasgraf @jone this is ready for review!
Please double-check commits https://github.com/4teamwork/opengever.core/pull/3082/commits/0c30dd5f0419c26b4878bdd0f7a7e3360531687f
https://github.com/4teamwork/opengever.core/pull/3082/commits/60226011edcd7c6c47a9dd18fbdc20470df44e20

As far as I understood, the commit with the comments needs to be dropped.

Rotonen:
@tarnap Could still usability-wise indicate the version number in the current version, like `Aktuelle Version (4)`? I'd find it a bit annoying to do the arithmetic per hand if I would ever need to audit what it was which was actually downloaded.

tarnap:
@Rotonen discuss this with @phgross

jone:
There are quite a lot of commits listed in the history section of the PR because you have a `Resolves: #3082` in the commit message.
This confuses me a little, because the commit is itself part of `#3082`. This means that it cannot not resolve it, kind of 🙄 
(I usually prefer to not use issue numbers in commit messages because it has this effect whenever the commit is rewritten).

tarnap:
@jone: @deiferni told me already about it.
I mistakenly referenced the Pull-Request instead of the Issue in the commit.

tarnap:
@deiferni @jone I decided to keep referencing the issue in the commit, but leaving the hashtag. This way I'll still have the number of the issue when creating the pull request, yet there will be no reference... and I only need to add the hashtag when creating the pull request.

tarnap:
@4teamwork/gever this is ready for review

PR #3085, opened 2017-07-03
Integration testing support for plone.protects auto CSRF protection

jone:
The problem is that we use the same connection / transaction for preparation as for performing a request with `ftw.testbrowser`.
This means that we may already have changed objects on the connection, but the change is not from within the request.

We fix that by marking all objects which are already marked as changed on the current as safe for CSRF. This also means that the auto protection does no longer trigger within the test for the followed requests.

This PR also activates the transaction interceptors savepoint simulation. This is necessary because `plone.protect` calls `transaction.abort()` when the CSRF validation fails. We want the state to be rollback to before the request in this state. The savepoint simulation does exactly that.

Based on https://github.com/4teamwork/ftw.testing/pull/28

### DIFF
--- a/opengever/core/testing.py
+++ b/opengever/core/testing.py
@@ -488,9 +488,12 @@ class GEVERIntegrationTesting(IntegrationTesting):
         self.interceptor.intercept(self.interceptor.BEGIN
                                    | self.interceptor.COMMIT
                                    | self.interceptor.ABORT)
+        self.interceptor.begin_savepoint_simulation()
+        self.interceptor.begin()
         logout()
 
     def testTearDown(self):
+        self.interceptor.stop_savepoint_simulation()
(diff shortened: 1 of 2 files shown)


Rotonen:
This means we can try to go for tests which do not bleed to STDOUT during a test run?

jone:
@Rotonen no, this is actually caused by https://github.com/plone/plone.protect/blob/master/plone/protect/auto.py#L279 
`traceback.print_stack()` does actually print and not return 😞 
I'm gonna make a PR.

jone:
@Rotonen See the PR for not printing to stdout: https://github.com/plone/plone.protect/pull/63

jone:
The tests are now passing because of the new `ftw.testing` release.
Ready to merge, I think 😉

PR #3037, opened 2017-06-19
Move Proposal/SubmittedProposal fields from SQL to plone objects.

deiferni:
Closes #3025

for the duration of the migration from sql to plone we will purposefully break multi-admin-unit support.

- copy proposal text attributes to `Proposal` plone objects
- also copy attributes to `SubmittedProposal`  plone objects
- drop copied columns in SQL
- copy object content on proposal submission, partially re-use `Transporter`
- get rid of confusing submitted proposal builder, which does not build plone objects and replace with a clean builder for `SubmittedProposal`
- improve builders, add mixin to transparently load model for `ModelContainer` and prepare support for builders handling sql/plone transparently

Also closes #1357.

PR #3091, opened 2017-07-05
Proposals: migrate SQL based history to plone

### LINKED ISSUE #3042: Move sql based proposal history back to plone objects
goal: 

Move sql based proposal history back to plone objects (`Proposal` / `SubmittedProposal`)

requirements/thoughts:

- history entries should have a timestamp
- history entries should be ordered based on their timestamp
- admin-units can receive history entries from other admin-units for linked content (proposals for now, tasks etc. in the future)
- an objects state may change/be updated based on a history entry
- storing history entries should be fault-tolerant and not depend on insertion order (it should be possible that history entries are inserted in the "wrong" order, i.e. newer before older but this should not lead to incorrect display order or incorrect object state)
- one should be able to store history entries from "the future" (i.e. from another admin-unit with a newer gever version)

suggested implementation: 

- store history entries in objects annotation, don't expose it as a schema field to avoid exposing the field directly over the API
- implement a custom API endpoint to append history entries (can be done later, but should be kept in mind)
- store history entries on OOBtree, `datetime` as key, `PersistentDict` as value
- use `PersistentDict` instead of `Persistent` subclasses as value to allow storing history entries from "the future", see above
- use `datetime` key to always guarantee chronological display order (also independent from sorting order)
- generate `uuid4` id for each history entry

deiferni:
This PR replaces the SQL based history implementation for proposals/submitted-proposals with a plone-based one. History entries are moved from SQL to the content type's annotations.

- store history records in object annotations
- we only store the record data in a `PersistentMapping` **not** the record itself as a `Persistent` subclass, this allows us to store records that the system does not know yet (has only theoretical value for now but may be relevant once we want to support multiple GEVER versions per cluster)
- history record are ordered based on their timestamp
- admin-units can receive history records from other admin-units for linked content (proposals for now)
- storing history records is fault-tolerant and does not depend on insertion order (has only theoretical value for now, see above comment)
- each history record is also identified by a `UUID` (random `uuid4`), we use the same `UUID` for duplicated records, this identifies the same record across admin-units. That way the records could be merged again should duplication prove to be the wrong approach.

Closes #3042.

- [x] Manually test upgrade script with many permutations


deiferni:
That seems to be a rebase artefact which you can ignore. It is a commit from https://github.com/4teamwork/opengever.core/pull/3037. I'll amend/remove once that PR is merged.

opengever/core/upgrades/20170704112234_migrate_sql_proposal_history_to_plone/upgrade.py
```diff
@@ -0,0 +1,196 @@
...
+from sqlalchemy.sql.expression import join
+from sqlalchemy.sql.expression import table
+from uuid import uuid4
+from zope.annotation.interfaces import IAnnotations
+from zope.component import getUtility
+from zope.intid.interfaces import IIntIds
+import logging
+
+
+logger = logging.getLogger('ftw.upgrade')
```
jone: This shouldn't be `ftw.upgrade` here, maybe `opengever.core.upgrades`?

opengever/core/upgrades/20170704112234_migrate_sql_proposal_history_to_plone/upgrade.py
```diff
@@ -0,0 +1,196 @@
...
+                data['submitted_version'] = submitted_version
+
+        return data
+
+    def migrate_history_for_proposal(self, data):
+        """Append record to proposals history."""
+
+        proposal = self.get_proposal()
+        if not proposal:
+            return
```
jone: What may have happened that the proposal is missing here?

opengever/meeting/browser/receiveproposalhistory.py
```diff
@@ -0,0 +1,28 @@
...
+@tracebackify
+class ReceiveProposalHistory(BrowserView):
+    """Receive remote history record data and store locally.
+
+    Receive history from a submitted proposal and store/duplicate it into the
+    local proposals history.
+
+    This view is only available for internal requests.
+    """
+    def __call__(self):
```
jone: + empty line?

opengever/meeting/proposalhistory.py
```diff
@@ -0,0 +1,343 @@
...
+                self.context.get_source_admin_unit_id(),
+                '@@receive-proposal-history',
+                path=path,
+                data=request_data,)
+
+        return record
+
+    def receive_record(self, timestamp, data):
+        history = self._get_history_for_writing()
+        history[timestamp] = data
```
jone: Would it be wise to ensure here that we have a persistent mapping?

opengever/meeting/command.py
```diff
@@ -380,26 +386,29 @@ def __init__(self, proposal, document, submitted_document):
...
+
+        record = IHistory(self.proposal).append_record(
+            'document_updated',
+            document_title=self.document.title,
+            submitted_version=submitted_version,
+        )
+        history_data = json.dumps({
+            'submitted_version': submitted_version,
+            'uuid': record.uuid,
+            }, cls=AdvancedJSONEncoder)
```
jone: Would it make sense to have our own `dumps`- and `loads`-functions, where we inject our encoder and decoder? There are quite a lot of places where we use `cls=AdvancedJSONEncoder`.

opengever/meeting/proposalhistory.py
deiferni: We're currently doing input validation in the `ReceiveProposalHistory`: https://github.com/4teamwork/opengever.core/blob/cd27db1a2e95f64c25f2f38785aa694e5d502bda/opengever/meeting/browser/receiveproposalhistory.py#L19. (Also the UUID is converted there already.) I believe that input validation should happen at the application boundaries, i.e. as early as possible. Repeating the conversion to `PersistentMapping` should not be necessary on the model level, but i'd say we could add a check and throw a `TypeError` if input does not meet the expectations 🤔 . Could do the same for `append_record`. I'll amend.

jone: Fair enough. I think validating here and throwing an error would give us security. If we store non-persistent objects here, maybe because of a bug, it could get messy. Let's awoid that 😉

opengever/meeting/proposalhistory.py
```diff
@@ -0,0 +1,343 @@
...
+    """Basic implementation of a history record.
+
+    Contains basic set of required data and abstract implementation. Can
+    re-populate iself from data by invoking the __new__ method directly and
+    then re-populating the attributeson the instance instead of calling
+    __init__.
+
+    """
+    name = None
+    needs_syncing = False
```
deiferni: 👍  i have added documentation in the class docstring for both configuration options; `name` and `needs_syncing`.

opengever/meeting/proposalhistory.py
```diff
@@ -0,0 +1,343 @@
...
+            uuid = uuid4()
+        elif isinstance(uuid, basestring):
+            uuid = UUID(uuid)
+
+        self.context = context
+        self.timestamp = timestamp
+        self.data = PersistentMapping(
+            created=timestamp,
+            userid=api.user.get_current().getId(),
+            name=self.name,
```
deiferni: Hmm i have always had the combo `[Base]HistoryRecord.name` in my mind, which should be self-explanatory. But you're right, for someone just inspecting the data it may not be clear enough. I'll rename both the config-option and the dict-key to `history_type`.

opengever/meeting/proposalhistory.py
```diff
@@ -0,0 +1,343 @@
...
+            uuid=uuid)
+
+    def append_to(self, history):
+        if self.timestamp in history:
+            return  # XXX raise?
```
deiferni: i agree, amended to raise hard 👍 .

opengever/meeting/proposalhistory.py
```diff
@@ -0,0 +1,343 @@
...
+    """A Proposal has been created."""
+
+    name = 'created'
+
+    def message(self):
+        return _(u'proposal_history_label_created',
+                 u'Created by ${user}',
+                 mapping={'user': self.get_actor_link()})
+
+ProposalHistory.register(ProposalCreated)
```
deiferni: beautiful 💛

opengever/meeting/tabs/submittedproposallisting.py
```diff
@@ -42,4 +42,5 @@ class SubmittedProposalTableSource(SqlTableSource):
     grok.implements(ITableSource)
     grok.adapts(SubmittedProposalListingTab, Interface)
 
+    # XXX make title searchable
```
deiferni: This is a rebase artefact and will go away with my next rebase 😉 .

opengever/base/tests/test_json.py
```diff
@@ -0,0 +1,70 @@
+from datetime import datetime
+from opengever.base.jsondecoder import AdvancedJSONDecoder
+from opengever.base.jsonencoder import AdvancedJSONEncoder
```
deiferni: The thing is what we currently have asymmetric implementations. We don't load `UUID` and `PersistentMapping` from `json`, for example, but we only know how to serialize those datatypes. The idea here is that the specific endpoints (`receive-proposal-history` in our case) should restore/validate the input-data to their needs and with domain-knowledge. If we would want to be more generic we would have to annotate the datatypes in `json` somehow, maybe similar to `Transporter`s type-prefixes in the keys. But we currently only use this implementation in one place, i'd say it is too early to make it generic. I'd propose to not address this for now but keep it in mind for later. Ok? Alternatives?

Also i have already failed hard by calling the module `json` 😂 .

opengever/meeting/command.py
deiferni: We currently use `json.dumps` in 84 different places, 5 of wich use the `AdvancedJSONEncoder`. It is unclear to me whether we could/should switch to using the custom encoder in all those places. I'd say for the moment it would be useful when enabling the custom encoder is a deliberate action and thus it also should be visible in the code-segment that uses it. 

If we want to switch we should address that in another PR. Ok for you?

opengever/base/tests/test_json.py
jone: I think the hole point of doing such a encoder / decoder composition is to support more data types (such as datetime) in a quite generic way.
Although we will not be able to restore things like `UUID`s, the `datetime` example shows that encoder and decoder relate.

By moving it to one file we could motivate future developers to try to enhance encoder and decoder simultaneously.

I agree that domain-specific data should be processed outside of json. The same argument could also be applied to non-restorable values such as a `UUID`-object.

> Also i have already failed hard by calling the module json 😂 .

Ah yes, naming it `json` will be tricky 😂

opengever/meeting/command.py
jone: Couldn't we just provide a new advanced-dumps-function and use it in all the new places (those added in this pr)?

opengever/core/upgrades/20170704112234_migrate_sql_proposal_history_to_plone/upgrade.py
deiferni: It could have been deleted/removed from plone without removing the corresponding sql entries. this is unlikely, but not impossible and for the upgrade i'd like to be extra careful.

opengever/base/tests/test_json.py
deiferni: hmm got a good idea for the module name? brainstorming:
- `advancedjson`
- `j_son`
- `geverjson`
- `jayson` 😂 

meh, i'd probably go with `advancedjson`, agree? better ideas?

jone: I thought about `ogjson`, but `jayson` is quite awesome too 😂 
`advancedjson` is fine too 👍

opengever/meeting/command.py
deiferni: ok i'll put that into the new `advancedjson` module 👍 .

opengever/base/jsonencoder.py
```diff
@@ -0,0 +1,26 @@
...
+    def default(self, obj):
+        if isinstance(obj, datetime):
+            return obj.isoformat()
+        elif isinstance(obj, UUID):
+            return str(obj)
+        elif isinstance(obj, set):
+            return list(obj)
+        elif isinstance(obj, PersistentMapping):
+            return dict(obj)
+        else:
```
deiferni: As discussed, for now we don't allow bytestrings but raise hard when one is put into an encoder.

opengever/base/tests/test_json.py
```diff
@@ -0,0 +1,70 @@
...
+import pytz
+
+
+class TestAdvancedJSONEncoder(TestCase):
+
+    def test_serializes_naive_datetime(self):
+        encoder = AdvancedJSONEncoder()
+        dt = datetime(2014, 12, 31, 15, 45, 30, 999)
+        self.assertEquals('["2014-12-31T15:45:30.000999"]',
+                          encoder.encode([dt]))
```
deiferni: good point, i have added a roundtrip test for all supported data-types.

deiferni:
@jone updated with your feedback, please have another look

opengever/meeting/proposalhistory.py
```diff
@@ -0,0 +1,355 @@
...
+    Contains basic set of required data and abstract implementation. Can
+    re-populate iself from data by invoking the __new__ method directly and
+    then re-populating the attributes on the instance instead of calling
+    __init__.
+
+    Each record must have a unique `history_type` from which it can be built
+    with IHistory.append_record.
+
+    If `needs_syncing` is `True` a records that is created on the
+    `SubmittedProposal` side is automatically added to its corresponding
```
jone: Do we need to send history records the other way round too at some point? How would we name that?

opengever/base/advancedjson.py
```diff
@@ -0,0 +1,136 @@
...
+        return super(AdvancedJSONEncoder, self).encode(obj)
+
+    def ensure_unicode_values(self, obj):
+        """Be defensive about encoding issues and only allow unicode strings as
+        input value for now.
+        """
+
+        if isinstance(obj, basestring):
+            if not isinstance(obj, unicode):
+                raise ValueError('Expecting uncicode value {}'.format(obj))
```
jone: - `uncicode`
- maybe `'Expecting uncicode value, got {!r}'.`? this would include the `repr` of the object.

opengever/base/advancedjson.py
```diff
@@ -0,0 +1,136 @@
...
+        self.ensure_unicode_values(obj)
+        return super(AdvancedJSONEncoder, self).encode(obj)
+
+    def ensure_unicode_values(self, obj):
+        """Be defensive about encoding issues and only allow unicode strings as
+        input value for now.
+        """
+
+        if isinstance(obj, basestring):
+            if not isinstance(obj, unicode):
```
jone: why not just `isinstance(obj, str)` instead of the two conditions?

opengever/base/advancedjson.py
```diff
@@ -0,0 +1,136 @@
...
+                raise ValueError('Expecting uncicode value {}'.format(obj))
+        # it looks like a dict
+        # minimal method interface for dict includes keys and __getitem__, so
+        # we use those here, according to:
+        # https://docs.python.org/2/library/userdict.html#UserDict.DictMixin
+        elif hasattr(obj, 'keys'):
+            for key in obj.keys():
+                self.ensure_unicode_values(obj[key])
```
jone: We should probably verify the key too:
```python
                self.ensure_unicode_values(key)
                self.ensure_unicode_values(obj[key])
```

opengever/base/advancedjson.py
```diff
@@ -0,0 +1,136 @@
...
+        # it looks like an iterable
+        elif hasattr(obj, '__iter__'):
+            for item in obj:
+                self.ensure_unicode_values(item)
```
jone: A map would look quite good here 😏 🙄 
```python
elif hasattr(obj, '__iter__'):
    map(self.ensure_unicode_values, obj)
```

opengever/base/advancedjson.py
```diff
@@ -0,0 +1,136 @@
...
+    def default(self, obj):
+        if isinstance(obj, datetime):
+            return {
+                TYPE_KEY: u'datetime',
+                VALUE_KEY: obj.isoformat(),
+            }
+        elif isinstance(obj, UUID):
+            return {
+                TYPE_KEY: u'UUID',
+                VALUE_KEY: str(obj),
```
jone: Should we verify that `default` uses `unicode`, too?

opengever/meeting/proposalhistory.py
deiferni: At the moment there is no use-case to do so for proposals. If one comes up we'll have to think about naming again 😉 .

opengever/base/advancedjson.py
deiferni: 😂  😄

opengever/meeting/proposalhistory.py
jone: Fair enough 👍

opengever/base/advancedjson.py
jone: I tried 😂

opengever/base/advancedjson.py
deiferni: hmm yeah i thought about that as discussed, but i wouldn't do it since we can't deal with dictionaries constructed with keywords arguments that way `dict(foo=u'qux')`

jone: So much on "let's use unicode everywhere" 😂 
The risk may not be that big, that we have keys with user-input-strings which may contain umlauts.

deiferni: i consider the dict keys more of an attribute, i really can't see where they would be provided by the user 🤔 , so i'd say the assumption that we have ascii only here should be safe.

jone: Fine for me 👍

opengever/meeting/browser/receiveproposalhistory.py
```diff
@@ -0,0 +1,29 @@
...
+        data = json.loads(self.request.get('data'), cls=AdvancedJSONDecoder)
+        timestamp = data['timestamp']
+
+        data = data['data']
+        data['id'] = UUID(data['id'])
+        data = PersistentMapping(data)
+        IHistory(self.context).receive_record(timestamp, data)
+
+        return 'OK'
```
phgross: Use `from opengever.base.utils import ok_response` here

opengever/meeting/browser/submitdocuments.py
```diff
@@ -224,6 +225,17 @@ class RecieveSubmittedDocumentView(PrivilegedReceiveObject):
...
+                                  cls=AdvancedJSONDecoder)
+
+        with elevated_privileges():
+            IHistory(self.context).append_record(
+                'document_submitted',
+                document_title=document.title,
+                submitted_version=history_data['submitted_version'],
+                uuid=history_data['uuid']
+            )
+            ILockable(document).lock(MEETING_SUBMITTED_LOCK)
```
phgross: 🤔  why does this has to be done with `elevated_privileges` ? 

And specially creating a Lock with the `elevated_privileges`-User could be dangerous, or nor?

opengever/meeting/tests/test_proposalhistory.py
```diff
@@ -53,35 +53,64 @@ def submit_proposal(self):
         self.proposal.execute_transition('pending-submitted')
         transaction.commit()  # also make change visible in browser
 
+    def assert_proposal_history_records(self, expected_records, browser,
```
phgross: from ftw.testbrowser import browser as default_browser

browser=default_browser

opengever/meeting/browser/submitdocuments.py
jone: I'm not sure why `elevated_privileges` is used 🤔 I don't think that there are permission checks happening within `append_record`.

I don't think that creating a lock with `elevated_privileges` should be a problem: the context manager creates a security manager with the same userid, but different permissions. For the lock, only the userid should be relevant.

phgross: OK. I was not sure how elevated_privileges/ILockable stuff works. As long it doesn't create a lock as a "different user" thats fine.

jone: @phgross I've change the lock to no longer be created within the elevated privileges context manager.

jone:
@phgross could you do a rereview here?

jone:
@phgross rebased and ready 👍

PR #3111, opened 2017-07-12
Cleanup rolemap, don't redefine plone roles.

deiferni:
It is not necessary to refedine roles alredy defined by plone.
Also remove comments from profile-merge to allow sorting the roles alphabetically.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -7,6 +7,7 @@ Changelog
 
 - Move data for most text-fields of Proposal (SQL) to their corresponding plone-objects Proposal/SubmittedProposal. [deiferni]
 - Introduce versioning for Proposal/SubmittedProposal. [deiferni]
+- Cleanup rolemap, don't redefine plone roles. [deiferni]
 
 
 2017.3.0 (2017-07-12)
--- a/opengever/core/profiles/default/rolemap.xml
+++ b/opengever/core/profiles/default/rolemap.xml
@@ -1,32 +1,12 @@
 <rolemap>
   <roles>
-
-    <!--migrated from opengever/sharing/profiles/default/rolemap.xml-->
     <role name="Administrator" />
-    <role name="Anonymous" />
-    <role name="Authenticated" />
-    <role name="Contributor" />
-    <role name="Editor" />
-    <role name="Manager" />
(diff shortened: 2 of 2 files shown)


deiferni:
@jone hm. why 🤔 ?

jone:
I use `sort-lines` in emacs for sorting everything (imports, lists, ..) and it is configured case-sensitive.
When I'm sorting the list of roles I have this change:
```diff
diff --git a/opengever/core/profiles/default/rolemap.xml b/opengever/core/profiles/default/rolemap.xml
index 656f9d1..9ad29fb 100644
--- a/opengever/core/profiles/default/rolemap.xml
+++ b/opengever/core/profiles/default/rolemap.xml
@@ -1,7 +1,7 @@
 <rolemap>
   <roles>
-    <role name="Administrator" />
     <role name="APIUser" />
+    <role name="Administrator" />
     <role name="Archivist" />
     <role name="CommitteeGroupMember" />
     <role name="Publisher" />
```

At some point we've agreed to sort imports case sensitive, but that does not apply necessarily to other things.

deiferni:
@jone afaik we don't really have guidelines regarding sorting 🤔 . Wouldn't it make sense to also sort this list case-insensitive i.e. the same as the imports? Having different rules per location/file may be a bit confusing for the developers.

jone:
@deiferni We've had the discussion about casing from time to time but I couldn't find it documented. I think we'll need to document it in devdocs.
Let's stick with the case-insensitive sorting for now. 👍

Rotonen:
👍  on not taking a stance on sorting here - the issue is hairy.

Would we also do the i18n on the sorting? Under the hood it boils down to `LC_COLLATE`. In our case we should ensure we use the appropriate `de_CH.UTF-8` and `fr_CH.UTF-8` for sorting when we use those for translations.

jone:
@Rotonen Yes, sorting content is a whole separate story. We usually just strip the diacritics ([example](https://github.com/4teamwork/ftw.labels/blob/master/ftw/labels/utils.py)), which is usually good enough for German. I think indexes such as `sortable_title` work in a compareable way.

In code we usually only have ascii-characters to sort (imports, dependencies, ..), that's easier.

PR #3113, opened 2017-07-12
List only documents as related_documents in the documents overview.

### LINKED ISSUE #3112: AttributeError: get_css_class when accessing a document reference by a task
https://sentry.4teamwork.ch/sentry/onegov-gever/issues/3550/

```
  File "home/zope/eggs/five.pt-2.2.4-py2.7.egg/five/pt/expressions.py", line 161, in __call__
    return base()
  File "home/zope/01-gever-stasg-test/src/opengever.core/opengever/document/browser/overview.py", line 212, in linked_documents
    key=lambda obj: obj.Title()
  File "home/zope/01-gever-stasg-test/src/opengever.core/opengever/document/browser/overview.py", line 243, in get_css_class
    return get_css_class(context or self.context)
  File "home/zope/01-gever-stasg-test/src/opengever.core/opengever/base/browser/helper.py", line 34, in get_css_class
    return _get_task_css_class(item)
  File "home/zope/01-gever-stasg-test/src/opengever.core/opengever/base/browser/helper.py", line 17, in _get_task_css_class
    return task.get_css_class()
AttributeError: get_css_class
```

phgross:
Skip proposal and task from the back relations list. Closes #3112.

A more specific relation catalog query is not possible, because the relation field on the task and the proposal schema have the same name than the one in the `IRelatedDocuments` behavior. Therefore we have to filter the list afterwards.

### DIFF
--- a/opengever/document/tests/test_overview.py
+++ b/opengever/document/tests/test_overview.py
@@ -36,15 +36,26 @@ class TestDocumentOverview(FunctionalTestCase):
         super(TestDocumentOverview, self).tearDown()
 
     @browsing
-    def test_overview_displays_related_documents(self, browser):
+    def test_overview_displays_related_documents_but_only_documents(self, browser):
         self.doc_a = create(Builder('document')
+                            .within(self.dossier)
                             .having(title=u'A\xf6'))
         self.doc_b = create(Builder('document')
+                            .within(self.dossier)
                             .having(title=u'B\xf6')
                             .relate_to(self.doc_a))
         self.doc_c = create(Builder('document')
+                            .within(self.dossier)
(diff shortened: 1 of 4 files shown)


opengever/document/tests/test_overview.py
```diff
@@ -45,6 +45,9 @@ def test_overview_displays_related_documents(self, browser):
         self.doc_c = create(Builder('document')
                             .having(title=u'C\xf6')
                             .relate_to(self.doc_b))
+        self.task = create(Builder('task')
```
deiferni: the pr mentions proposals as well, should you also test for that?

PR #3116, opened 2017-07-12
Amend the OfficeConnector JS to only refresh after a LOCK is acquired

Rotonen:
Now that OfficeConnector (since 1.2.1) also does a WebDAV LOCK , change the javascript to check for that instead.

The document status JSON endpoint now also includes the document lock status.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -12,6 +12,7 @@ Changelog
 - Introduce versioning for Proposal/SubmittedProposal. [deiferni]
 - Cleanup rolemap, don't redefine plone roles. [deiferni]
 - Use OGMail original_message-file for bumblebee-conversion if available. [elioschmutz]
+- Amend the OfficeConnector javascript to refresh the page based on the lock status. [Rotonen]
 - Add the document lock status to the document status end point. [Rotonen]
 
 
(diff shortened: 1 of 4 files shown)


opengever/document/static/officeconnector.js
```diff
@@ -41,7 +41,7 @@
 
     function isDocumentCheckedOut(documentURL) {
```
bierik: Since we're checking for the lock the function should be named like `isDocumentLocked`

Rotonen:
@bierik ready for rereview.

docs/HISTORY.txt
```diff
@@ -8,6 +8,9 @@ Changelog
 - Move data for most text-fields of Proposal (SQL) to their corresponding plone-objects Proposal/SubmittedProposal. [deiferni]
 - Introduce versioning for Proposal/SubmittedProposal. [deiferni]
 - Cleanup rolemap, don't redefine plone roles. [deiferni]
+- Nothing changed yet.
```
Rotonen: Damn, that's the only case the union rebase auto merge does not handle - a default entry which should go away.

Rotonen:
@phgross ready for rereview.

PR #3135, opened 2017-07-14
Let long tests fail.

jone:
We need to make sure that we optimize tests so that they are fast. Most tests shouldn't take that long when implemented correctly. Therefore we want to fail whenever a test takes longer than the threshold.
Exceptions can be made with a decorator where the new time limit and a reason must be provided.
For providing tooling for making tests faster I have implemented a [flamegraph decorator](https://github.com/4teamwork/ftw.flamegraph/pull/1) and I added `ftw.flamegraph` to the tests dependencies, providing a hint in the failing assertion.

Discussion:
- Is this a good idea?
- What is a good time limit?
- ~~The time really depends on the machine. Would it make sense to use e.g. CPU-time only? How can I do that in a simple way?~~ => `time.clock()`


sources.cfg
```diff
@@ -9,9 +9,13 @@ development-packages =
 # Please cleanup your pull-request so that you are not adding ftw.*-packages
 # to the development-packages. It is your responsibility to release the
 # dependencies first, then update the versions.cfg in your pull-request.
+# https://github.com/4teamwork/ftw.flamegraph/pull/1
+  ftw.flamegraph
```
jone: 🚧 https://github.com/4teamwork/ftw.flamegraph/pull/1

sources.cfg
```diff
@@ -9,9 +9,13 @@ development-packages =
...
 
 auto-checkout = ${buildout:development-packages}
 
 [branches]
 Products.LDAPUserFolder = ftw
 Products.LDAPMultiPlugins = ftw
+# https://github.com/4teamwork/ftw.flamegraph/pull/1
+ftw.flamegraph = jone-decorator
```
jone: 🚧 https://github.com/4teamwork/ftw.flamegraph/pull/1

opengever/testing/integration_test_case.py
```diff
@@ -24,6 +26,42 @@ def setUp(self):
         self.portal = self.layer['portal']
         self.request = self.layer['request']
         map(self.activate_feature, self.features)
+        self._start_time = time()
```
buchi: I would use `time.clock()` instead of `time.time()`. This should make measurement more robust when the test server ist busy.

jone: Thanks, that's what I was looking for 👍

opengever/testing/integration_test_case.py
```diff
@@ -24,6 +26,42 @@ def setUp(self):
...
+        self._start_time = clock()
+        self._max_duration = 3000
+
+    def tearDown(self):
+        super(IntegrationTestCase, self).tearDown()
+        duration = (clock() - self._start_time) * 1000
+        self.assertLess(
+            duration, self._max_duration,
+            'The test took to long. It should not take longer'' than {} ms.'
+            ' Use the @ftw.flamegraph.flamegraph(True) decorator for investigating'
```
jone: I thought about making a shorthand for the flamegraph, e.g.:
```
@IntegrationTestCase.flamegraph(True)
def test_xy(self):
    pass
```

Since `IntegrationTestCase` is already imported this should work without an additional import and is quite easy to remember.
Is this a good idea?

phgross: 👍  i think so ... but passing in `True` looks quite redundant.


```python

Comment (jone) [inline opengever/testing/integration_test_case.py]: If we pass `True`, the decorator will open the resulting svg with `open [path]`. But this is a Mac OS command, that's we the option is `False` by default: https://github.com/4teamwork/ftw.flamegraph/pull/1#discussion_r127586450

But we could make a `@IntegrationTestCase.open_flamegraph` decorator: this would obviously default the `open_svg=True`. We wouldn't commit those decorator anyway, since we are only using it for development in gever. The platform issue can be handled by the developer..

**Comment by jone** [comment]
As discussed, I've added a flamegraph decorator:
```python
    @IntegrationTestCase.open_flamegraph
    def test_xy(self):
        pass
```
This decorator builds a flamegraph of the decorated test and opens it with `open result.svg`.

PR #3144, opened 2017-07-18
Fix unicode error in committee group vocabulary.

jone:
The token of a vocabulary term must be a bytestring because it will be sent over HTTP. The VocabularyTerm class casts tokens to bytestring in it's __init__, which will throw an encoding error when the token is a unicode containing non-ASCII characters.

Simply encoding the unicode as utf-8-bytestring will not work because the z3cform validation will fail and not allow to select the term.

A good token does only contain ASCII characters while beeing reproducable and hopefully not conflicting with other terms. Therefore I changed the token to be a normalized string generated from the group ID.

It seems that groups usually have no title. We should not fallback to the normalized string then but to the original group id.

---

Without this fix, the encoding problem can be produced by editing a any comittee with a `Manager`-user on an environment where the OGDS group table contains groups with non-ASCII characters.
This is the case with a default local development installation, as `zopemaster`, at http://localhost:8080/fd/sitzungen/committee-3/edit .

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -12,6 +12,7 @@ Changelog
 - Fix: remove ftw.showroom CSS on fresh installations too. [jone]
 - Define cookie path when setting the current orgunit id in the cookie. [phgross]
 - Make sure unicode is stored in proposal sql. [jone]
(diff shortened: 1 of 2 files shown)


jone:
Rebased.

PR #3148, opened 2017-07-18
SPV word: checkout proposal document after creating proposal.

jone:
Add a option for checking out and external-editing the proposal document
of newly created proposals. The option is enabled by default.

As discussed in #2829

bildschirmfoto 2017-07-18 um 16 58 00

Not sure why there is a CSS bug; I'm using a regular z3cform checkbox widget as it is used in other places. Probably needs to be fixed in the theme.

### DIFF
--- a/opengever/meeting/browser/proposalforms.py
+++ b/opengever/meeting/browser/proposalforms.py
@@ -2,6 +2,7 @@ from five import grok
 from ftw.table import helper
 from opengever.base.schema import TableChoice
 from opengever.base.widgets import TrixFieldWidget
+from opengever.document.interfaces import ICheckinCheckoutManager
 from opengever.dossier.templatefolder import get_template_folder
 from opengever.meeting import _
 from opengever.meeting import is_word_meeting_implementation_enabled
@@ -11,14 +12,19 @@ from opengever.meeting.proposal import IProposal
 from opengever.meeting.proposal import ISubmittedProposal
 from opengever.meeting.proposal import Proposal
 from opengever.meeting.proposal import SubmittedProposal
+from opengever.officeconnector.helpers import is_officeconnector_checkout_feature_enabled  # noqa
 from opengever.tabbedview.helper import document_with_icon
 from plone import api
 from plone.directives import dexterity
+from plone.directives.form import widget
 from plone.uuid.interfaces import IUUID
 from plone.z3cform.fieldsets.utils import move
 from z3c.form import field
+from z3c.form.browser.checkbox import SingleCheckBoxFieldWidget
 from z3c.form.interfaces import HIDDEN_MODE
+from zope.component import getMultiAdapter
 from zope.interface import provider
+from zope.schema import Bool
 from zope.schema.interfaces import IContextSourceBinder
 from zope.schema.vocabulary import SimpleVocabulary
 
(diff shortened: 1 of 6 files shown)


opengever/meeting/browser/proposalforms.py
```diff
@@ -117,6 +120,12 @@ class IAddProposal(IProposal):
              'column_title': _(u'label_modified', default=u'Modified'),
              'transform': helper.readable_date}))
 
+    widget(edit_after_creation=SingleCheckBoxFieldWidget)
+    edit_after_creation = Bool(
+        title=_(u'label_edit_after_creation', default='Edit after creation'),
```
phgross: default of a msgid should be unicode:
_(u'label_edit_after_creation', default=**u**'Edit after creation'),

opengever/meeting/browser/proposalforms.py
```diff
@@ -156,8 +165,17 @@ def createAndAdd(self, data):
...
-        proposal.create_proposal_document(proposal_template.file)
+        proposal_doc = proposal.create_proposal_document(proposal_template.file)
+        if edit_after_creation:
+            self.checkout_and_external_edit(proposal_doc)
         return proposal
+
+    def checkout_and_external_edit(self, document):
+        """Checkout the document and invoke an external editing session.
+        """
+        self.context.restrictedTraverse('checkout_documents').checkout(document)
```
phgross: I'm not 100% sure, but those this handle the two different checkout protocolls (new and old office connector plugin)?

jone: I copied that from the [`form.py` of the templatefolder](https://github.com/4teamwork/opengever.core/blob/321ad1278c087717223d482348a71ea78fd16b4d/opengever/dossier/templatefolder/form.py#L128-L137).

The line below does a `document.setup_external_edit_redirect(self.request)`, which works [differently when `officeconnector_checkout_feature_enabled`](https://github.com/4teamwork/opengever.core/blob/321ad1278c087717223d482348a71ea78fd16b4d/opengever/document/document.py#L288-L293).

But I'm not sure how this should be used when the feature is enabled. Should I not checkout the document at this point with the feature?

@Rotonen what is the correct way to do a checkout at this point?

Rotonen: @jone as far as I can see everything landing in the redirector should just pick the right thing to do after @tarnap fixed it. It fires a global JS function you can see fail in your console if things went right and you do not have a dev build of OfficeConnector for Mac currently running.

https://github.com/4teamwork/opengever.core/blob/master/opengever/base/redirector.py#L102

The check is rather crude, it just checks the URL being passed in begins with `oc:`.

jone: I have updated the `checkout_and_external_edit` method; it now works with and without OC-checkout feature flag.

When the OC-checkout feature flag is enabled, the generated [OC-link will contain the instruction to checkout the document](https://github.com/4teamwork/opengever.core/blob/321ad1278c087717223d482348a71ea78fd16b4d/opengever/document/document.py#L288-L293) and OC will checkout before starting the editing. Therefore we should not checkout the document already on proposal creation time.

When the feature flag is disabled we checkout the document while creating the proposal and then trigger `/external_edit`.

jone:
I've rebased and updated the checkout part. Please rereview.

jone:
@phgross motivational rebase 😎

PR #3152, opened 2017-07-19
Disable reference number column in the document listing of the inbox.

phgross:
Because documents inside the inbox does not have a meaningful reference number, the recently introduced column should be removed for the document listing of inboxes.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 ---------------------
 
 - Move handlebar templates to template files. [jone]
+- Disable reference number column in the document listing of the inbox. [phgross]
 - Fix: remove ftw.showroom CSS on fresh installations too. [jone]
 - Define cookie path when setting the current orgunit id in the cookie. [phgross]
 - Add version of downloaded file to journal's entry. [tarnap]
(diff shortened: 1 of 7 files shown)


docs/HISTORY.txt
```diff
@@ -5,6 +5,7 @@ Changelog
 2017.4.0 (unreleased)
 ---------------------
 
+- Disable reference nubmer column in inboxs document listing. [phgross]
```
jone:
```diff
- Disable reference nubmer column in inboxs document listing.
+ Disable reference number column in inbox's document listing. 
```

Rotonen: `in the document listing of the inbox`?

phgross:
@jone, @Rotonen updated the PR. I add also a commit which separates the translation msgid for the new reference column, this makes it easier to customize it in policies (see https://github.com/4teamwork/opengever.sg/issues/190)

opengever/tabbedview/browser/tabs.py
```diff
@@ -175,7 +175,8 @@ class Documents(BaseCatalogListingTab):
          'transform': translate_public_trial_options},
 
         {'column': 'reference',
-         'column_title': _(u'label_reference', default=u'Reference Number')},
+         'column_title': _(u'label_document_reference',
```
jone: Should this change appear in the changelog?

phgross: No, I don't think so, it does not change anything for the user, the translation in og.core stays the same.

jone: Fair enough 👍

PR #3155, opened 2017-07-19
Meeting word: add proposal document to meeting view

jone:
~⚠️ _requires https://github.com/4teamwork/plonetheme.teamraum/pull/564_~
ℹ️ _part of #2829_ 

Update the meeting view so that the proposal document is listed and can directly be edited.

Changes:
- Show proposal document right beneath proposal; the tooltip allows for fast checkout and edit.
- Make proposal title text bold, because it is the important part of the proposal.
- Prefix attachments with a "Attachments"-label; this was not clear enough.
- Move Attachment-Toggle from a separate "Documents" column to the left side (into the number column), because a toggle should be at the left side from a UX viewpoint and because the documents column did consume too much space.
- These changes are feature-flagged with the word-changes-feature.

bildschirmfoto 2017-07-19 um 16 46 09

### DIFF
--- a/sources.cfg
+++ b/sources.cfg
@@ -8,6 +8,8 @@ extensions += mr.developer
 development-packages =
   opengever.maintenance
   collective.js.timeago
+# https://github.com/4teamwork/plonetheme.teamraum/pull/564
+  plonetheme.teamraum
 
 # Please cleanup your pull-request so that you are not adding ftw.*-packages
 # to the development-packages. It is your responsibility to release the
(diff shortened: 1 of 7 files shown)


jone:
Rebased.

jone:
Rebased.

sources.cfg
```diff
@@ -18,3 +20,5 @@ auto-checkout = ${buildout:development-packages}
 [branches]
 Products.LDAPUserFolder = ftw
 Products.LDAPMultiPlugins = ftw
+# https://github.com/4teamwork/plonetheme.teamraum/pull/564
```
phgross: @jone I've merged the Theme PR, could you remove this line that I'll merge this as well.

PR #3158, opened 2017-07-20
Fix original message (*.msg) download

phgross:
- Fix mail download: Convert LF to CRLF only for EML mails.
- Enable *.msg download for all download copy links, also in the tooltip or bumblebeeoverlay

### DIFF
--- a/opengever/mail/browser/download.py
+++ b/opengever/mail/browser/download.py
@@ -1,5 +1,6 @@
 from opengever.document.browser.download import DocumentishDownload
 from plone.namedfile.interfaces import HAVE_BLOBS
+from opengever.mail.mail import IOGMail
 
 if HAVE_BLOBS:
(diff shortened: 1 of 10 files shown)


opengever/mail/browser/download.py
```diff
@@ -25,6 +25,9 @@ def convert_line_endings(self, filename):
         return ''.join(lines)
 
     def stream_data(self, named_file):
+        if self.fieldname == 'original_message':
```
lukasgraf: I would rather import the actual field here and compare against `field.getName()` - otherwise this will break silently if we ever rename this field.

I.e.
```python
from opengever.mail.mail import IOGMail

# ...

if self.fieldname == IOGMail.original_message.getName():
```

phgross: @lukasgraf good point, I've updated the PR.

PR #3165, opened 2017-07-20
Add upgradestep to cleanup no longer existing interfaces from zc.relation catalog.

phgross:
With #2949 the `ICreator` and `ICreatorAware` interfaces has been removed. But because the zc.relation cataog persists provided interfaces, the removed interfaces has been removed manually from the zc.relation catalog.

Closes #3164

PR #3166, opened 2017-07-21
Add test to detect interface removals without cleaning up the relation catalog

phgross:
Because the zc.relation cataog persists provided interfaces, the removed interfaces has been cleaned up manually from the zc.relation catalog. This test should help to not forget the manual  cleanup, when removing an interface.

See #3165 for more informations.

### DIFF
--- a/opengever/core/tests/test_relations.py
+++ b/opengever/core/tests/test_relations.py
@@ -1,0 +1,248 @@
+from ftw.builder import Builder
+from ftw.builder import create
+from opengever.testing import FunctionalTestCase
+from plone import api
+from plone.dexterity.utils import iterSchemataForType
+from z3c.relationfield.interfaces import IRelationChoice
+from z3c.relationfield.interfaces import IRelationList
+from zc.relation.interfaces import ICatalog
+from zope.component import getUtility
+
+
+# Update this list when removing or adding a interface, for example a behavior.
+# Becuase the zc.relation catalog persists provided interfaces, the catalog has
+# to be cleaned up when removing an interface from this list.
+# See oc.core.upgrades.CleanupICreatorInterfacesFromRelationCatalog for a howto
+# example
+EXPECTED_INTERFACES = [
+    'AccessControl.interfaces.IOwned',
+    'AccessControl.interfaces.IPermissionMappingSupport',
+    'AccessControl.interfaces.IRoleManager',
+    'Acquisition.interfaces.IAcquirer',
+    'App.interfaces.INavigation',
+    'App.interfaces.IPersistentExtra',
+    'App.interfaces.IUndoSupport',
+    'OFS.interfaces.ICopyContainer',
+    'OFS.interfaces.ICopySource',
+    'OFS.interfaces.IFTPAccess',
+    'OFS.interfaces.IFindSupport',
+    'OFS.interfaces.IFolder',
+    'OFS.interfaces.IItem',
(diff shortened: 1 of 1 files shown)


opengever/core/tests/test_relations.py
```diff
@@ -0,0 +1,241 @@
...
+from plone.dexterity.utils import iterSchemataForType
+from z3c.relationfield.interfaces import IRelationChoice
+from z3c.relationfield.interfaces import IRelationList
+from zc.relation.interfaces import ICatalog
+from zope.component import getUtility
+
+
+# Update this list when removing or adding a interface, for example a behavior.
+# Becuase the zc.relation catalog persists provided interfaces, the catalog has
+# to be cleanedup when removing an interface from this list.
```
jone: `cleanedup` => `cleaned up`

opengever/core/tests/test_relations.py
```diff
@@ -0,0 +1,241 @@
...
+
+        self.maxDiff = None
+        self.assertItemsEqual(
+            EXPECTED_INTERFACES, self.get_all_persisted_interfaces())
+
+    def test_all_types_with_relations_are_handled(self):
+        types_with_relations = filter(
+            has_relation_fields, list(api.portal.get_tool('portal_types')))
+
+        self.assertEquals(EXPECTED_TYPES_WITH_RELATIONS, types_with_relations)
```
jone: I'd add a assertion message here.
If I understand it correctly, the idea of this assertion is to fail when there is a new content type with a relation field: the dev must update the `EXPECTED_TYPES_WITH_RELATIONS` list AND update the builders in the other test, right?
It would be great to have an assertion message where it is explained how to "fix" a failure.

opengever/core/tests/test_relations.py
```diff
@@ -0,0 +1,241 @@
...
+            .relate_to(document_a)
+            .with_submitted())
+
+        dossier_expired = create(Builder('dossier').as_expired())
+        self.grant('Records Manager')
+        create(Builder('disposition').having(dossiers=[dossier_expired]))
+
+        self.maxDiff = None
+        self.assertItemsEqual(
+            EXPECTED_INTERFACES, self.get_all_persisted_interfaces())
```
jone: I'd add an assertion message here, telling the developer what to do in order to fix the issue.

opengever/core/tests/test_relations.py
phgross: Yes, @jone updated ...

opengever/core/tests/test_relations.py
```diff
@@ -0,0 +1,248 @@
...
+            'catalog is cleaned up and adjust the EXPECTED_INTERFACES list.')
+
+    def test_all_types_with_relations_are_handled(self):
+        types_with_relations = filter(
+            has_relation_fields, list(api.portal.get_tool('portal_types')))
+
+        self.assertEquals(
+            EXPECTED_TYPES_WITH_RELATIONS, types_with_relations,
+            'A new type with a relation field has been introduced, please'
+            'extend the content setup with this type and adjust the expected'
```
jone: `'extend` => `' extend`
Otherwise it will print `introduced, pleaseextend the content`.
I usually add the space at the new line in order to indicate that this is a multiline string.

phgross: @jone you're right, updated ...

PR #3168, opened 2017-07-21
Implement i18n REST endpoint.

bierik:
Use RESP API endpoint to have translations on client side.
So refactor the favorite tree portlet using the new i18n javascript function.

### DIFF
--- a/opengever/base/locales/de/LC_MESSAGES/opengever.base.po
+++ b/opengever/base/locales/de/LC_MESSAGES/opengever.base.po
@@ -4,7 +4,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
-"POT-Creation-Date: 2017-06-07 14:47+0000\n"
+"POT-Creation-Date: 2017-07-21 13:13+0000\n"
 "PO-Revision-Date: 2014-09-04 01:03+0200\n"
 "Last-Translator: Jonas Baumann <j.baumann@4teamwork.ch>\n"
 "Language-Team: Jonas Baumann <j.baumann@4teamwork.ch>\n"
@@ -73,17 +73,17 @@ msgid "desc_confirm_action"
 msgstr "Bestätigen Sie bitte, dass Sie diese Aktion durchführen möchten."
 
 #. Default: "You have not selected any Items."
-#: ./opengever/base/browser/copy_items.py:22
+#: ./opengever/base/browser/copy_items.py:23
 msgid "error_no_items"
 msgstr "Sie haben keine Objekte ausgewählt"
 
 #. Default: "The item you selected cannot be copied."
-#: ./opengever/base/browser/copy_items.py:29
+#: ./opengever/base/browser/copy_items.py:30
 msgid "error_not_copyable"
 msgstr "Der gewählte Inhalt kann nicht kopiert werden"
 
 #. Default: "Classification"
-#: ./opengever/base/behaviors/classification.py:61
+#: ./opengever/base/behaviors/classification.py:62
 msgid "fieldset_classification"
 msgstr "Sichtbarkeit"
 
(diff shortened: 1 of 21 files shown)


opengever/base/service.py
```diff
@@ -30,3 +30,10 @@ def render(self):
             payload['locked_by'] = None
 
         return json.dumps(payload)
+
+
+class Internationalization(Service):
+    """Provide translations in json format to use them on the frontend"""
+
+    def reply(self):
+        return json.dumps(get_translations(self.request))
```
lukasgraf: This doesn't fit with inheriting from the `plone.restapi` `Service` base class. Either use `reply()` and let the `Service` base class handle the JSON serializiation, or use `render()` and handle it yourself, but not both.

opengever/base/service.py
```diff
@@ -1,17 +1,17 @@
...
 from zope.component import queryMultiAdapter
-
 import json
 
 
 class DocumentStatus(Service):
     """Provide information on the current status of a document."""
 
-    def render(self):
+    def reply(self):
```
lukasgraf: No. 👎  `render()` is supposed to return a string, and `reply()` is supposed tu return a Python data structure that gets serialized to JSON and pretty printed by the `plone.restapi` `Service` base class. You can't just change this without adapting the method result! Why are you even changing OfficeConnector related services?

opengever/base/tests/test_i18n_service.py
```diff
@@ -0,0 +1,53 @@
...
+import transaction
+
+
+class Testi18nService(FunctionalTestCase):
+
+    def setUp(self):
+        super(Testi18nService, self).setUp()
+        lang_tool = api.portal.get_tool('portal_languages')
+        lang_tool.supported_langs = ['en', 'de-ch', 'fr-ch']
+        self.portal.manage_permission(UseRESTAPI, roles=['Authenticated'])
```
lukasgraf: This would require us to give the "Use REST API" permission to all customers on all deployments. We don't currently want to do that yet. I'd suggest to overwrite `check_permission` so that the i18n endpoint it doesn't require the `Use REST API" permission.

opengever/base/locales/de/LC_MESSAGES/opengever.base.po
```diff
@@ -93,53 +93,58 @@ msgid "fieldset_common"
...
 
-#: ./opengever/base/behaviors/lifecycle.py:44
+#: ./opengever/base/behaviors/lifecycle.py:45
 msgid "help_retention_period"
 msgstr "Zeitraum zwischen dem jüngsten Dokumentdatum eines in einem Dossier enthaltenen Dokuments und dem Zeitpunkt, an dem dieses für die Geschäftstätigkeit der Verwaltungseinheit nicht mehr benötigt wird."
 
+#. Default: "Add to favorites"
+#: ./opengever/base/frontend_translations.py:6
+msgid "label_add_to_favorites"
+msgstr "Als Favorit markieren"
```
lukasgraf: Why are you adding new translations in this commit?

bierik: I need some test translations.
Do you have an other idea testing the endpoint without preconfigured translations?

lukasgraf: Hmm, I see.
The PR body doesn't contain any description about what this endpoint's API looks like, and how it's supposed to be used. Or what the bigger purpose of this change is. Is the favorite portlet being used as a reference implementation for the new endpoint? If so, why can't the existing translations be removed if they're now only used in the frontend?

bierik: I tried to document this is the readme: https://github.com/4teamwork/opengever.core/pull/3168/files#diff-88b99bb28683bd5b7e3a204826ead112R326

Yes the favourite portlet is a reference implementation.
I also removed the previous implementation of the favourite portlet: https://github.com/4teamwork/opengever.core/pull/3168/files#diff-8c5fa85f611ec32605dd51d0ee463282L359

lukasgraf: But then why aren't the backend translations from `portlets/tree/locales/de/LC_MESSAGES/opengever.portlets.tree.po:msgid "label_remove_from_favorites"` removed? Are those still being used somewhere?

lukasgraf:
Obsolete - superseded by #3177

PR #3172, opened 2017-07-24
Improve linebreak display in tasks

tarnap:
Converts web intelligent text to html for line breaks and clickable links.

linebreaks_in_tasks

URLs embedded at the end of sentences remain an issue.

Resolves: #3054
See also: https://github.com/plone/plone.intelligenttext/issues/4

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,7 +5,7 @@ Changelog
 2017.4.1 (unreleased)
 ---------------------
 
-- Nothing changed yet.
+- Format line breaks in task descriptions. [tarnap]
 
 
 2017.4.0 (2017-07-26)
--- a/opengever/task/tests/test_overview.py
+++ b/opengever/task/tests/test_overview.py
@@ -9,6 +9,7 @@ from plone import api
 from plone.app.testing import TEST_USER_ID
 from StringIO import StringIO
 import transaction
+import unittest
 
 
 class TestTaskOverview(FunctionalTestCase):
@@ -221,6 +222,11 @@ class TestTaskOverview(FunctionalTestCase):
 
         self.assertEqual(expected, result)
 
+    @unittest.skip(
+        """Fails because we expect URLs which are embedded at the end of the
+        sentences should not contain the punctuation mark.
+        This is issued in
+        https://github.com/plone/plone.intelligenttext/issues/4""")
     @browsing
     def test_task_text_urls_in_sentences_are_transformed(self, browser):
         dossier = create(Builder('dossier').titled(u'Dossier'))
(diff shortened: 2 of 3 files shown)


opengever/task/tests/test_overview.py
```diff
@@ -178,6 +179,11 @@ def test_task_overview_displays_task_information(self, browser):
             table_data
         )
 
+    @unittest.skip(
```
phgross: Please, only skip the tests which is not supported via `plone.intelligenttext` but the functionality itself should be testet. Maybe its the easiest way to split those test to multiple tests

versions.cfg
```diff
@@ -159,6 +159,7 @@ plone.app.transmogrifier = 1.4
 plone.app.versioningbehavior = 1.3.1
 plone.batching = 1.0.8
 plone.formwidget.autocomplete = 1.2.5
+plone.intelligenttext = 2.1.0
```
phgross: Is not necessary already pinned via the Plone KGS see line 6.

phgross:
@tarnap https://github.com/4teamwork/opengever.core/pull/3172#pullrequestreview-51784713

opengever/task/tests/test_overview.py
```diff
@@ -179,6 +180,145 @@ def test_task_overview_displays_task_information(self, browser):
...
+        result = (browser.css('table.listing').find('Text').first.row
+                         .css('td').first.innerHTML)
+
+        self.assertEqual(expected, result)
+
+    @unittest.skip(
+        """Fails because we expect URLs which are embedded at the end of the
+        sentences should not contain the punctuation mark.
+        This is issued in
+        https://github.com/plone/plone.intelligenttext/issues/4""")
```
jone: @tarnap can you fix this in Plone?

Is it a blocker that this does not work yet?
If yes, we should include a patch in this PR.

tarnap:
@phgross Added screenshot and changelog entry

docs/HISTORY.txt
```diff
@@ -15,6 +15,7 @@ Changelog
 - Export msg in OGmail zip represenation if the original_message is available. [mathias.leimgruber]
 - Add original_message support to bundle import. [phgross]
 - Rename current orgunit cookie, to invalidate existing cookies. [phgross]
+- Format line breaks in task descriptions. [tarnap]
```
phgross: Changelog entry is in the wrong release section. Notice 2017.4.0 has been released today ...

opengever/task/tests/test_overview.py
tarnap: @jone: yes I can, and I did it!
https://github.com/plone/plone.intelligenttext/issues/4
https://github.com/plone/plone.intelligenttext/pull/5

PR #3176, opened 2017-07-24
Refactor the function: earliest_possible_end_date.

### LINKED ISSUE #3107: Resolve end date does not respect first level document dates
While archiving a dossier, the earliest possible end date does not respect documents stored on dossier-root-level.

Documents may be on dossier-root when:
- there is no subdossier
- the resolving is configured to be `lenient` (https://github.com/4teamwork/opengever.core/pull/2972)

elioschmutz:
This PR refactors the function: `earliest_possible_end_date`.

Beside an overall code-refactoring, the following changes where made on code-optimizations:

**Objects of dossiers**
Objects of Dossier-brains are not necessary because the start and end-date are in the catalog metadata

**Convert datetime objects into date objects if necessary**
All dates are already date-objects. Otherwise, there is something wrong in consistency

**Check if dossier start-date is available or not**
We filter out None-values and get the max-date. There is no need to do this check anymore.

**Sort end-dates**
Thats not necessary because the max-function does not require a sorted list.

closes #3107

### DIFF
--- a/opengever/dossier/base.py
+++ b/opengever/dossier/base.py
@@ -248,9 +248,12 @@ class DossierContainer(Container):
         return True
 
     def earliest_possible_end_date(self):
-
+        """The earliest possible end-date must be lather than all document dates and
+        all dossier start and end dates.
+        """
+        dates = []
         catalog = getToolByName(self, 'portal_catalog')
-        subdossiers = catalog({
+        dossier_brains = catalog({
             'path': '/'.join(self.getPhysicalPath()),
             'object_provides': [
                 'opengever.dossier.behaviors.dossier.IDossierMarker', ],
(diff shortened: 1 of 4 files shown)


opengever/dossier/base.py
```diff
@@ -248,9 +248,12 @@ def has_valid_enddate(self):
         return True
 
     def earliest_possible_end_date(self):
-
+        """The earliest possible end-date must be after all document dates and
```
jone: I'd probably change `after all document dates` to `lather than all document dates` 🤔

opengever/dossier/base.py
```diff
@@ -248,9 +248,12 @@ def has_valid_enddate(self):
...
+        all dossier start and end dates.
+        """
+        end_dates = []
         catalog = getToolByName(self, 'portal_catalog')
-        subdossiers = catalog({
+        dossier_brains = catalog({
```
jone: I think we should chang ethis to `catalog.unrestrictedSearchResults({`:
Since we are constructing a validation, we want to take all contents into account, no matter if the user can see it or not.
As long as we are not trying to get the objects from the brains nor leak any protected information (such as the title), we can do an unrestricted catalog query.

opengever/dossier/base.py
```diff
@@ -259,39 +262,21 @@ def earliest_possible_end_date(self):
...
+            end_dates.append(dossier_brain.start)
+
+        document_brains = catalog({
+            'path': '/'.join(self.getPhysicalPath()),
+            'object_provides': [
+                'opengever.document.behaviors.IBaseDocument'],
+        })
+
+        end_dates.extend(
+            [document_brain.document_date for document_brain in document_brains])
```
jone: Because of consistency, I'd either also use `for`-loop here, as with the dossiers, or use extend with a list comprehension in the dossier calculation above, too (once with start, once with end).

opengever/dossier/base.py
phgross: Hmm rather no. Specially for documents it would include trashed or even "removed" documents, and those document should definitely not be part of the calculation.

The case that a user can't see all document or all subdossier, but has still the permission to resolve it, seems unrealistic for me.

An option would be to use `catalog.unrestrictedSearchResults` and filter the `trashed` and `removed` documents manually. 
@jone what you think.

opengever/dossier/base.py
elioschmutz: @jone what do you want to say with `lather` 😄

opengever/dossier/base.py
elioschmutz: @phgross 👍  I don't change anything!

opengever/dossier/base.py
```diff
@@ -259,39 +262,21 @@ def earliest_possible_end_date(self):
...
-                        end_dates.append(doc.document_date.date())
-                    else:
-                        end_dates.append(doc.document_date)
-
-        if end_dates:
-            end_dates.sort()
-            return max(end_dates)
-        return None
+        for dossier_brain in dossier_brains:
+            end_dates.append(dossier_brain.end)
```
elioschmutz: 👍

opengever/dossier/base.py
jone: @phgross fair enough, you're right.

I wouldn't use `trashed` and `removed` with `catalog.unrestrictedSearchResults` as it problably makes more harm than good.

I'd therefore keep it as it is with a restricted search.

opengever/dossier/base.py
jone: 🙈 I meant `later than all document dates`

elioschmutz:
@jone what happened with this PR? Why did you closed it without merging?

jone:
:flushed: I have absolutely no idea. I've either accidentally clicked the close button, or it may have been referenced by another issue with a "close" statement, or it may be a side effect of merging base-pr #3184 and deleting the base-branch 🙈

jone:
I've changed the base to `master` and reopen the PR. I rebase may be needed.

elioschmutz:
😂  easy! I will do a rebase.

elioschmutz:
@jone @phgross reminder...

PR #3178, opened 2017-07-24
OGGBundle: Support *.msg import for mails

### LINKED ISSUE #3094: Support *.msg import form mails in OGGBundle
With #2948 GEVER supports now to store the *.msg in an additional field. The OGGBundle import should support this feature as well.

phgross:
The PR adds support to OGGBundle mail import for *.msg files to the `original_message` field. Therefore a new key `original_message_path` can be used in the `documents.json`.
 
Closes #3094.

This PR updates also the schemadumps. Some changes had to be done because of changes in the last releases which has been made without updating the schemadumps:
 - Add the two new mail fields `original_message` and `message_source`.
 - Add IMail.message_source field to IGNORED_FIELDS
 - Add fields `allow_add_businesscase_dossier` and `addable_dossier_templates`
 - Default of some `description` fields.


opengever/bundle/sections/fileloader.py
```diff
@@ -58,12 +59,11 @@ def __init__(self, transmogrifier, name, options, previous):
     def __iter__(self):
         for item in self.previous:
             guid = item['guid']
-            file_field = self.get_file_field(item)
 
-            try:
-                abs_filepath, path = self.get_abs_filepath(item)
-                self.validate_filepath(item, abs_filepath, path)
-            except FileLoadingFailed:
+            pathkey = self.pathkey(*item.keys())[0]
```
lukasgraf: I think this should use `self.get_supported_fields(item)` insead of `self.pathkey` instead? And `self.key` can be deleted as well.

opengever/bundle/sections/fileloader.py
```diff
@@ -74,16 +74,30 @@ def __iter__(self):
...
+                    continue
+
+                try:
+                    self.add_namedblob_file(abs_filepath, file_field, obj)
+                except EnvironmentError as e:
+                    # TODO: Check for this in OGGBundle validation
+                    logger.warning(
+                        "Can't open file %s. %s." % (abs_filepath, str(e)))
+                    error = (guid, abs_filepath, str(e), path)
+                    self.bundle.errors['files_io_errors'].append(error)
```
lukasgraf: Missing a `continue` here? Or should at least set `file_added` to `False`?

opengever/base/schemadump/field.py
```diff
@@ -123,22 +123,25 @@ def _get_vocabulary(self, field):
...
             elif field.vocabulary:
                 if IVocabularyFactory.providedBy(field.vocabulary):
-                    vf = field.vocabulary
+                    vocabulary = field.vocabulary(None)
                 # we (sometimes) use IContextSourceBinder independent of
                 # context - for fields where we don't we should define an
                 # override
                 elif IContextSourceBinder.providedBy(field.vocabulary):
-                    vf = field.vocabulary
+                    vocabulary = field.vocabulary(None)
```
phgross: Hmm good point, need to check that.

opengever/bundle/schemas/documents.schema.json
```diff
@@ -194,15 +195,6 @@
                     "description": "Fortlaufend gez\u00e4hlte Nummer eines Dokumentes.",
                     "title": "Laufnummer"
                 },
-                "creators": {
```
phgross: See #2949 and #1131

opengever/bundle/sections/fileloader.py
phgross: No, the pathkey is the path to the plone object not the file path, usually `_path`

lukasgraf: Ah, right. 👍

opengever/base/schemadump/field.py
phgross: @lukasgraf As we discussed together, even if the vocabulary is set with the `vocabulary` property, those case must be handled by the schema dumper.

However I changed the message_source from `source=` to `vocabulary=`.

opengever/bundle/sections/fileloader.py
phgross: I moved the `file_added` property inside the try method.

A `continue` is not necessary, because it should also try to import other fields.

PR #3179, opened 2017-07-24
Implement REST API endpoints for document checkout/checkin

buchi:


opengever/api/tests/test_checkout.py
```diff
@@ -0,0 +1,98 @@
...
+class TestCheckoutAPI(IntegrationTestCase):
+
+    def setUp(self):
+        super(TestCheckoutAPI, self).setUp()
+        setRoles(
+            self.portal, self.administrator.getId(),
+            self.administrator.getRoles() + ['APIUser'])
+        self.login(self.administrator)
+        self.document = create(Builder('document')
+                               .within(self.dossier))
```
jone: In my "Integration tests for opengever.dossier" PR I've [added a `self.document`](https://github.com/4teamwork/opengever.core/pull/3184/files#diff-80aa9c702cc0bd4985609181b4d9873bR147) to the default fixture.
I think this builder could be removed then.

opengever/api/tests/test_checkout.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @browsing
+    def test_checkout_document(self, browser):
+        browser.login(self.administrator.getId())
```
jone: The gever team has agreed to use `self.login` for logging in the test thread as well as the browser with `self.login` in integration tests, e.g.:
```
def test_checkout_document(self, browser):
    self.login(self. administrator, browser)
```
(An example is also in the readme https://github.com/4teamwork/opengever.core/#example-integration-test-with-browser)
Discussion: https://4teamwork.slack.com/archives/C060FEYKA/p1498490241485990

buchi: I've replaced `browser.login()` with `self.login()`.

opengever/api/tests/test_checkout.py
buchi: Ok. What's the time frame for #3184? I will need checkout/checkin functionality next week on dev or lab.

jone: #3184 is ready for review. @phgross what do you think, should we review / merge the dossier integration tests first, or ship this and change the dossier integration tests?
Both is quite easy, depends on your workload (review) and priorities.

phgross: I've reviewed #3184 should be merged soon. If it's ok I would ship #3184 first.

opengever/api/tests/test_checkout.py
```diff
@@ -0,0 +1,98 @@
...
+from zope.component import getMultiAdapter
+
+
+class TestCheckoutAPI(IntegrationTestCase):
+
+    def setUp(self):
+        super(TestCheckoutAPI, self).setUp()
+        setRoles(
+            self.portal, self.administrator.getId(),
+            self.administrator.getRoles() + ['APIUser'])
```
jone: @buchi Would it make sense to support that in the fixture, or have a helper?
Ideas:
1) Let `self.administrator` have the `APIUser`-role by default globally.
2) Add `self.api_user` with this role. It would be unclear what other permissions the user, though.
3) Add a `self.grant(self.administrator, 'APIUser')` to `IntegrationTestCase`.
4) Add a `self.enable_api_for(self.administrator)` to `IntegrationTestCase`.

I'd probably combine 1 and 4 🤔

opengever/api/tests/test_checkout.py
phgross: @buchi @jone #3184 is merged ...

buchi: I've removed the document builder.

opengever/api/tests/test_checkout.py
buchi: I would prefer 1. 
Sooner or later any user will need the "use api" permission if we want to use it for in the front end. Thus I'm not sure how long this role will stay...

jone: @buchi 1. is good; could you update the role [in the fixture](https://github.com/4teamwork/opengever.core/blob/be1aa3be36ed7a00a1a8e8cc10fea4726e3f6a58/opengever/testing/fixtures.py#L73) and remove it here?

buchi: @jone APIUser role is now granted in the fixture

PR #3182, opened 2017-07-24
Rename current orgunit cookie, to invalidate existing cookies.

phgross:
Otherwise the fixes in #3151 has no effect for users with an existing cookie, because the current orgunit cookie has a validity of 30 days.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -7,6 +7,7 @@ Changelog
 
 - Export msg in OGmail zip represenation if the original_message is available. [mathias.leimgruber]
 - Add original_message support to bundle import. [phgross]
+- Rename current orgunit cookie, to invalidate existing cookies. [phgross]
 - Move handlebar templates to template files. [jone]
 - Disable reference number column in the document listing of the inbox. [phgross]
 - Enable \*.msg download for all download copy links. [phgross]
--- a/opengever/ogds/base/ou_selector.py
+++ b/opengever/ogds/base/ou_selector.py
@@ -2,7 +2,7 @@ from collections import OrderedDict
 from Products.CMFPlone import PloneMessageFactory as pmf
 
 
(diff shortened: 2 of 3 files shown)


docs/HISTORY.txt
```diff
@@ -6,6 +6,8 @@ Changelog
 ---------------------
 
 - Export msg in OGmail zip represenation if the original_message is available. [mathias.leimgruber]
+- Rename current orgunit cookie, to invalidate existing cookies. [phgross]
+- Add upgradestep to cleanup no longer existing interfaces from zc.relation catalog. [phgross]
```
jone: ✂️

opengever/ogds/base/ou_selector.py
```diff
@@ -2,7 +2,7 @@
 from Products.CMFPlone import PloneMessageFactory as pmf
 
 
-CURRENT_ORG_UNIT_KEY = 'current_org_unit'
+CURRENT_ORG_UNIT_KEY = 'current-org-unit'
```
jone: I think in the case that someone must debug that later it is quite unclear why this changed without looking in the commit.
We could invent a versioning of the string: `current_org_unit_v2` 🤔

jone: Why does this invalidate the cookie?
I could only see this key used as internal storage key 🤔

phgross: You're right, changed it to `current_org_unit_v2`

phgross: Thats how the `CookieStorage` works: 

See:
https://github.com/4teamwork/opengever.core/blob/master/opengever/ogds/base/utils.py#L81-L100
https://github.com/4teamwork/opengever.core/blob/master/opengever/ogds/base/utils.py#L64-L65

jone: Ah, didn't read it properly 🙈 
Thanks for the hint 👍

Issue #3024, opened 2017-06-12
Speed up tests by improving testing isolation


jone:
We need to speed up tests by improving testing isolation. Although this is a general issue, it is important to have that for improving the development speed of the SPV, as the SPV is the slowest module at the moment.

**Goal:**
We want to have fixtures (layer) for a basic content setup and only use factories (`ftw.builder`) special use cases.

**Strategy:**
Because the SQL isolation is not that easy we need to stop commiting transactions in tests.
In order to stop commiting when doing browser requests we implement a new `ftw.testbrowser` request driver which supports making requests with thee need to commit.

jone:
Integration tests with a fixture and database isolation (including SQL) is implemented and used with a new, non-comitting testbrowser driver (traversal driver).

PR #3184, opened 2017-07-25
Integration tests for opengever.dossier

jone:
I've started rewriting the `opengever.dossier` tests with integration testing.
Part of #3024

I'd like to have this PR merged as soon as possible because we need the fixture for basing other tests on.

### Changes: 

**Enhance the integration test case base class:**
- catalog helpers
- workflow state helpers
- property for accesing all dossier tasks
- support for generic-setup-profile based feature flags; add filing number feature
- `IntegrationTestCase.clock` decorator

**Fixture enhancements**
- an empty dossier, which can be used for setting up a specific testing scenario
- a "secretariat user", who can archive / unarchive dossiers
- new subdossiers
- tasks and subtasks
- documents in main dossiers, subdossier and task
- proposal
- use UTC timezone as base for everything, since we should always use timezones

**Rewrite dossier tests**
8 test modules were rewritten.

### DIFF
--- a/opengever/dossier/tests/test_activate.py
+++ b/opengever/dossier/tests/test_activate.py
@@ -1,76 +1,51 @@
 from datetime import date
-from ftw.builder import Builder
-from ftw.builder import create
 from ftw.testbrowser import browsing
-from ftw.testbrowser.pages.statusmessages import error_messages
-from ftw.testbrowser.pages.statusmessages import info_messages
+from ftw.testbrowser.pages import editbar
+from ftw.testbrowser.pages import statusmessages
 from opengever.dossier.behaviors.dossier import IDossier
-from opengever.testing import FunctionalTestCase
-from plone import api
-from plone.protect import createToken
+from opengever.testing import IntegrationTestCase
 
 
-class TestDossierActivation(FunctionalTestCase):
-
-    def setUp(self):
-        super(TestDossierActivation, self).setUp()
-        self.grant('Editor', 'Publisher')
-        self.dossier = create(Builder('dossier')
-                              .in_state('dossier-state-inactive'))
+class TestDossierActivation(IntegrationTestCase):
 
     @browsing
     def test_recursively_activates_subdossier(self, browser):
-        subdossier = create(Builder('dossier')
-                            .within(self.dossier)
-                            .in_state('dossier-state-inactive'))
-        subsubdossier = create(Builder('dossier')
-                               .within(subdossier)
-                               .in_state('dossier-state-inactive'))
+        self.login(self.secretariat_user, browser)
+        self.set_workflow_state('dossier-state-inactive',
+                                self.dossier, self.subdossier)
 
-        browser.login().open(self.dossier, view='transition-activate',
-                             data={'_authenticator': createToken()})
+        browser.open(self.dossier)
+        editbar.menu_option('Actions', 'dossier-transition-activate').click()
+        statusmessages.assert_message('The Dossier has been activated')
 
-        self.assertEqual('dossier-state-active',
-                         api.content.get_state(self.dossier))
-        self.assertEqual('dossier-state-active',
-                         api.content.get_state(subdossier))
-        self.assertEqual('dossier-state-active',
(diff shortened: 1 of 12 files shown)


opengever/testing/integration_test_case.py
```diff
@@ -175,3 +176,59 @@ def get_catalog_metadata(self, obj):
...
+            'catalog index': workflow_state_id,
+            'catalog metadata': workflow_state_id}
+
+        got = {
+            'object': api.content.get_state(obj),
+            'catalog index': self.get_catalog_indexdata(obj)['review_state'],
+            'catalog metadata': self.get_catalog_metadata(obj)['review_state']}
+
+        self.assertEqual(
+            expected, got,
```
phgross: Nice ✨ , it happens all the time, that the brain is not up to date ...

opengever/dossier/tests/test_activate.py
```diff
@@ -2,75 +2,53 @@
...
 
-        self.assertEqual('The Dossier has been activated', info_messages()[0])
+        self.assert_workflow_state('dossier-state-active', self.dossier)
+        self.assert_workflow_state('dossier-state-active', self.subdossier)
 
     @browsing
-    def test_activating_a_subdossier_is_disallowed_when_main_dossier_is_inactive(self, browser):
-        subdossier = create(Builder('dossier')
-                            .within(self.dossier)
+    def test_activate_subdossier_is_disallowed_when_main_dossier_is_inactive2(
```
phgross: Typo: inactive2 -> inactive

opengever/dossier/tests/test_activate.py
```diff
@@ -2,75 +2,53 @@
...
+            self, browser):
+        self.login(self.secretariat_user, browser)
+        self.set_workflow_state('dossier-state-inactive', self.empty_dossier)
+        subdossier = create(Builder('dossier').within(self.empty_dossier)
```
phgross: Why does this test can't use self.dossier and self.subdossier?

opengever/testing/fixtures.py
```diff
@@ -141,8 +141,48 @@ def create_treaty_dossiers(self):
...
                     start=date(2016, 1, 1),
                     responsible='hugo.boss')))
-        self.register('subdossier',
-                      create(Builder('dossier').within(self.dossier).titled(u'2016')))
+
+        document = self.register('document', create(
+            Builder('document').within(self.dossier)
+            .titled(u'Vertr\xe4gsentwurf')
+            .attach_file_containing('Word dummy content',
+                                    u'vertr\xe4g sentwurf.docx')))
```
phgross: 🤔  not sure if we should non-ascii characters here. The filename of a document is normalized in any case. See https://github.com/4teamwork/opengever.core/blob/master/opengever/document/subscribers.py#L60

opengever/testing/fixtures.py
```diff
@@ -141,8 +141,48 @@ def create_treaty_dossiers(self):
...
+            .having(responsible_client=self.org_unit.id(),
+                    responsible=self.regular_user.getId(),
+                    issuer=self.dossier_responsible.getId(),
+                    task_type='correction',
+                    deadline=date(2016, 11, 1))
+            .in_state('task-state-resolved')
+            .relate_to(document)))
+
+        self.register('taskdocument', create(
+            Builder('document').within(self.dossier)
```
phgross: Should the `taskdocument` not be inside the task?

opengever/dossier/tests/test_activate.py
jone: You are absolutely right, we can use `self.subdossier` here.
I do not remember why I thought that I couldn't.
✅ updated

opengever/testing/fixtures.py
jone: Ah, good catch, makes sense.
✅ updated to `vertragsentwurf.docx`

opengever/testing/fixtures.py
jone: Good catch, updated. ✅

Issue #3205, opened 2017-08-02
Remove grok


phgross:
This issue is meant as an overview for the ongoing process to replace grok with zcml registrations. 

:white_large_square: = todo
:pencil2:  = in progress
:ballot_box_with_check:  = PR in review
:white_check_mark:  = PR merged

:white_check_mark: activity, @elioschmutz (#3244)
:white_check_mark: advancedsearch, @phgross
:white_check_mark: api ()
:white_check_mark: base, @elioschmutz (#3521)
:white_check_mark: bumblebee, @elioschmutz (#3248)
:white_check_mark: bundle ()
:white_check_mark: contact, @elioschmutz (#3269)
:white_check_mark:   core, @elioschmutz (#3528)
:white_check_mark: debug, @elioschmutz (#3255)
:white_check_mark: disposition, @elioschmutz  (#3267)
:white_check_mark: document, @elioschmutz (#3503, #3522)
:white_check_mark: dossier, @elioschmutz (#3508)
:white_check_mark: examplecontent ()
:white_check_mark: globalindex, @elioschmutz (#3272)
:white_check_mark: inbox, @elioschmutz (#3472)
:white_check_mark: journal, @elioschmutz (#3276)
:white_check_mark: latex, @elioschmutz (#3288)
:white_check_mark: locking, @elioschmutz (#3280)
:white_check_mark: mail, @elioschmutz (#3281)
:white_check_mark:  meeting, @elioschmutz (#3524)
:white_check_mark: officeatwork, @elioschmutz (#3499)
:white_check_mark: officeconnector ()
:white_check_mark: ogds.base, @lukasgraf (#3277)
:white_check_mark: policy ()
:white_check_mark:  policytemplates, @elioschmutz (#3525)
:white_check_mark: portlets ()
:white_check_mark: private, @elioschmutz (#3481)
:white_check_mark: quota ()
:white_check_mark: repository, @elioschmutz (no PR)
:white_check_mark: setup ()
:white_check_mark: sharing, @elioschmutz (#3284)
:white_check_mark: sphinx ()
:white_check_mark: tabbedview, @phgross
:white_check_mark: tabbedview fetch views entfernen, @elioschmutz (#3246)
:white_check_mark: task, @elioschmutz (#3497)
:white_check_mark: tasktemplates, @elioschmutz (#3286)
:white_check_mark: testing ()
:white_check_mark: trash, @phgross (#3315)
:white_check_mark: usermigration ()
[...]

PR #3207, opened 2017-08-02
Ungrok opengever.advancedsearch.

phgross:
Use ZCML instead of grok for registration (part of #3205).

### DIFF
--- a/opengever/advancedsearch/advanced_search.py
+++ b/opengever/advancedsearch/advanced_search.py
@@ -1,6 +1,5 @@
 from datetime import date
 from datetime import timedelta
-from five import grok
 from ftw.datepicker.widget import DatePickerFieldWidget
 from ftw.keywordwidget.widget import KeywordWidget
 from opengever.advancedsearch import _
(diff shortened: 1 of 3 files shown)


opengever/advancedsearch/advanced_search.py
```diff
@@ -34,8 +32,9 @@ def get_possible_dossier_states(context):
         states.append(SimpleVocabulary.createTerm(state, state, PMF(state)))
     return SimpleVocabulary(states)
 
+directlyProvides(get_possible_dossier_states, IContextSourceBinder)
```
jone: I prefer using decorators:
```python
from zope.interface import provider

@provider(IContextSourceBinder)
def get_possible_dossier_states(context):
    ...
```
Does this work too?

I prefer decorators because they are more error-prone to refactorings than module-level function calls, which can basically be at any point in the module or even in anther module.

phgross: Yes, I prefer decorators too. PR updated ...

opengever/advancedsearch/advanced_search.py
```diff
@@ -70,6 +69,7 @@ def get_types(context):
     return SimpleVocabulary(types)
 
 
+
```
jone: ✂️ 😉

PR #3209, opened 2017-08-02
Fix sorting by sqlalchemy column

### LINKED ISSUE #3161: Membership-listing: sorting on member column does not work.
The membership listing (Tab `Mitgliedschaften`) on a committee does not sort the member column correctly alphabetically.

- [x] Fix sorting b ymember column in membership-listing

elioschmutz:
This PR allows to sort by sqlalchemy columns and activates sorting on memberships-listing.

screen2

closes #3161

### DIFF
--- a/opengever/tabbedview/sqlsource.py
+++ b/opengever/tabbedview/sqlsource.py
@@ -40,6 +40,16 @@ class SqlTableSource(GeverTableSource):
 
         if self.config.sort_on:
             sort_on = self.config.sort_on
+
+            # sqlalchemy_sort_indexes is a dict on the TableSourceConfig which
+            # defines a mapping between default sort_on attributes based on
+            # strings and sqlalchemy sort-indexes based on an sqlalchemy column.
+            # This allows us to sort by joined table columns.
+            if hasattr(self.config, 'sqlalchemy_sort_indexes'):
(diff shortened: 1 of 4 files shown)


opengever/tabbedview/sqlsource.py
```diff
@@ -40,6 +40,16 @@ def extend_query_with_ordering(self, query):
 
         if self.config.sort_on:
             sort_on = self.config.sort_on
+
+            # sqlalchemy_sort_indexes is a dict on the TableSourceConfig which
+            # defines a mapping between default sort_on attributes based on
+            # strings and sqlalchemy sort-indexes based on an sqlalchemy column.
+            # This allows us to sort by joined table columns.
+            if hasattr(self.config, 'sqlalchemy_sort_indexes'):
```
phgross: it seems that `ftw.table` provides an additional attribute `sort_index` could we use that instead of creating a separate configuration ?

elioschmutz: No, unfortunately not. This attribute is just to separate sorting and displaying.

The problem is, that the sort_on parameter needs to be a string because it will be used as a query-parameter in a sort-request. So we have to do such a mapping...

phgross: ok.

PR #3211, opened 2017-08-03
Fix ArchiveForm, when resolving a dossier.

phgross:
Since #2972 the dossier resolver is an named adapter, and should be
gotten with the `get_resolver`. But the use of the DossierResolver in
the ArchiveForm has not been updated and failed therefore. This PR
fixes this issue and also add tests for the ArchiveForm.

### DIFF
--- a/opengever/dossier/tests/test_archiver.py
+++ b/opengever/dossier/tests/test_archiver.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from ftw.builder import Builder
 from ftw.builder import create
 from ftw.testbrowser import browsing
+from ftw.testbrowser.pages import statusmessages
 from ftw.testing import MockTestCase
 from grokcore.component.testing import grok
 from opengever.core.testing import ANNOTATION_LAYER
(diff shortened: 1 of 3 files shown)


opengever/dossier/tests/test_archiver.py
```diff
@@ -267,3 +268,99 @@ def test_enddate_may_be_latest_dossier_end_date(self, browser):
...
+        browser.open(self.empty_dossier, view='transition-archive')
+
+        browser.fill({'filing prefix': 'Government',
+                      'filing Year': u'2017',
+                      'Action': 'resolve and set filing no'})
+        browser.click_on('Archive')
+
+        self.assert_workflow_state(
+            'dossier-state-resolved', self.empty_dossier)
+        self.assertEquals(['The Dossier has been resolved'], info_messages())
```
jone: The `statusmessages` page object has an assertion function:
```python
from ftw.testbrowser.pages import statusmessages

statusmessages.assert_message('The Dossier has been resolved')
```
The assertion message is optimized there 😉

opengever/dossier/tests/test_archiver.py
```diff
@@ -3,6 +3,7 @@
 from ftw.builder import Builder
 from ftw.builder import create
 from ftw.testbrowser import browsing
+from ftw.testbrowser.pages.statusmessages import info_messages
```
jone: I usually import the `statusmessages` module, not the functions, so that I have the context in the test:

```python
assert_message('xy')  #  what message?
statusmessages.assert_message('xy')  # ah, a statusmessage its
```

opengever/dossier/tests/test_archiver.py
phgross: Ah good to know. PR updated

jone: By the way, that's a good one too:
```python
statusmessages.assert_no_error_messages()
```
I usually put this after modifying something, such as creating an object.
It can help finding a problem much faster, because it fails early with the error status message.

PR #3217, opened 2017-08-03
Add an additional batching_container for tabbedview-listings at the end of the table.

### LINKED ISSUE #840: Show paging navigation also at the bottom of table listings
C.f. http://feedback.onegovgever.ch/t/navigationselemente-fuer-seitenwechsel-auch-unterhalb-einer-auflistung-zur-verfuegung-stellen/66

elioschmutz:
Add an additional batching_container for tabbedview-listings at the end of the table.

bildschirmfoto 2017-08-03 um 14 38 46
bildschirmfoto 2017-08-03 um 14 43 42

closes #840


phgross:
bildschirmfoto 2017-08-04 um 16 45 26
looks confusing, the tabbedview actions are bit "lost".
 
@bierik @elioschmutz any ideas 🤔

elioschmutz:
We could add a divider:

bildschirmfoto 2017-08-07 um 09 28 50

elioschmutz:
@phgross what do you think?
@bierik do you have another idea (without reinvent the whole theme 😉 )

deiferni:
@elioschmutz what is the state here? this PR looks almost ready.

i would not add a divider since the actions operate on the items that are selected by checkbox in the table. the divider suggests that the operations are independent of the selection.

the order of the containers is consistent with the table heading which displays the page navigation closer to the table than the links that help selecting rows (i.e. ticking a rows checkbox) or unselecting all rows.

screen shot 2018-01-16 at 09 19 58

so the table footer should do the same, which you have done already.

elioschmutz:
Yes, the PR is ready if you agree my implementation.

opengever/tabbedview/browser/generic.pt
```diff
@@ -0,0 +1,66 @@
...
+            tal:attributes="action context/absolute_url">
+        <input type="hidden" name="orig_template" tal:attributes="value string:${context/absolute_url}/#${view/view_name}" />
+        <span tal:replace="structure context/@@authenticator/authenticator"/>
+
+        <tal:extjs condition="view/extjs_enabled">
+          <div id="listing_container"
+               class="ftwtable +extjs tab_container"><!--iefix--></div>
+
+          <!-- CUSTOM -->
+          <div id="batching_container" class="ftwtable"><!--iefix--></div>
```
deiferni: the `batching_container` id is duplicated, could you use another approach here?

opengever/tabbedview/browser/generic_with_bumblebee_viewchooser.pt
```diff
@@ -43,6 +43,10 @@
         <tal:extjs condition="view/extjs_enabled">
           <div id="listing_container"
                class="ftwtable +extjs tab_container"><!--iefix--></div>
+
+          <!-- CUSTOM -->
+          <div id="batching_container" class="ftwtable"><!--iefix--></div>
```
deiferni: same here i think

opengever/tabbedview/browser/generic_with_filters.pt
```diff
@@ -27,6 +27,10 @@
         <tal:extjs condition="view/extjs_enabled">
           <div id="listing_container"
                class="ftwtable +extjs tab_container"><!--iefix--></div>
+
+          <!-- CUSTOM -->
+          <div id="batching_container" class="ftwtable"><!--iefix--></div>
```
deiferni: and here

opengever/tabbedview/browser/generic.pt
elioschmutz: 🤔 thats true! I could, but:

`ftw.table` is hardcoded for using the IDs: https://github.com/4teamwork/ftw.table/blob/master/ftw/table/browser/ftwtable.extjs.js#L405

we could register a new static config for i.e. `batching_bottom` with the batching template if we hook in the function: https://github.com/4teamwork/ftw.tabbedview/blob/master/ftw/tabbedview/browser/listing.py#L255 which is after the config have been updated.

This would solve the duplicated ID for the batching-container.

Another problem would be, that the template of the batching_container itself contains elements with IDs, too: https://github.com/4teamwork/opengever.core/blob/master/opengever/tabbedview/browser/batching.pt#L70
This IDs are used in stylings, in JS-selectors for bumblebee and for the batching itself in `ftw.tabbedview`

I think, we will ran into several unpredictable problems if we try to change this...

But I can do it if it's really worth doing this effort.

deiferni: the thing is that an id **must** be unique within an html element. everything else is not valid and also does not meet developer expectations. could you use a `class` instead?

elioschmutz: Yes, but what I mean is:

```js
$('[id]').each(function(){
  var ids = $('[id="'+this.id+'"]');
  if(ids.length>1 && ids[0]==this)
    console.warn('Multiple IDs #'+this.id);
});
VM754:4 Multiple IDs #notificationTemplate
VM754:4 Multiple IDs #portal-logo-right
VM754:4 Multiple IDs #batching_container
VM754:4 Multiple IDs #message_no_contents
VM754:4 Multiple IDs #protect-script
```

the ID must be unique by definition, but browsers do allow it. We already have different duplicated IDs.

Thats the reason for my sentence:

`But I can do it if it's really worth doing this effort.`

As I wrote already, we get chain-f***d because the whole implementation is based on IDs. I would have to change `ftw.table`, `ftw.tabbedview`, `opengever.core` in python, JS and CSS to do this change.

I think, for this special case it's a matter of time and budget if I should do this change or not.

What do you think?

deiferni: I know we unfortunately already have some bad implementations. This does not justify to add a new one though 😉 .

Duplicating ids will cause issues, for example when accessing these elements programmatically with `getElementById` or the `#` css selector only the first one ist returned since the assumption is that they only may exist once. So even if browser do allow it, it will lead to problems and does not meet the programmers expectation also. Example:

```javascript

$('#notificationTemplate').length
>> 1

```
vs.
```javascript

$('[id="notificationTemplate"]').length
>> 2

```
But you're right, this increases the effort needed for this issue since you'd need to do additional work. Could you update the issue with our findings here and also provide an estimate via labels so we can prfioritize? I fear implementing it for the current milestone is no longer realistic 😢 .

elioschmutz: 👍 i just updated the description of this PR and added the estimated-label.

Implementing it for the next release (2018.1) should be possible. I can do it, just tell it

elioschmutz:
⚠️ TODO ⚠️ 

This PR adds an additional batching_container-element at the bottom of a listing-table.

Unfortunately, the batching-implementation is based on tag-ID's instead of classes. Means, we get duplicated ID's which are not allowed in valid HTML.

To solve this issue, we have to do several adjustments. Here an excerpt from an answer below:

> `ftw.table` is hardcoded for using the IDs: https://github.com/4teamwork/ftw.table/blob/master/ftw/table/browser/ftwtable.extjs.js#L405
> 
> we could register a new static config for i.e. `batching_bottom` with the batching template if we hook in the function: https://github.com/4teamwork/ftw.tabbedview/blob/master/ftw/tabbedview/browser/listing.py#L255 which is after the config have been updated.
> 
> This would solve the duplicated ID for the batching-container.
> 
> Another problem would be, that the template of the batching_container itself contains elements with IDs, too: https://github.com/4teamwork/opengever.core/blob/master/opengever/tabbedview/browser/batching.pt#L70
> This IDs are used in stylings, in JS-selectors for bumblebee and for the batching itself in `ftw.tabbedview`
> 

It seems, that the following packages have to be updated: `ftw.table`, `ftw.tabbedview`, `opengever.core`

- [ ] Implement the batching-template class based instead ID-based

⚠️ END TODO ⚠️

elioschmutz:
@deiferni here the new commit to solve the issue with duplicated IDs:
-
 https://github.com/4teamwork/opengever.core/pull/3217/commits/5a87cdb7df29083d2a55297aef40669036adb4ea

and here the related PRs of the other packages:
- https://github.com/4teamwork/ftw.table/pull/47
- https://github.com/4teamwork/ftw.tabbedview/pull/81

deiferni:
@elioschmutz the two PRs are merged. Can you create a new **major** for the two packages and provide the pinnings in this PR? After that it should be ready to 🚢.

elioschmutz:
@deiferni 👍 releases created.

The new implementation of `ftw.table` is not a braking change. So i just released a new minor version.

docs/HISTORY.txt
```diff
@@ -13,6 +13,8 @@ Changelog
 - Add tasks tab to repository folder. [jone]
 - Add documents tab to repository folder. [jone]
 - OGIP 17: Implement invitation storage for workspace invitations. [mathias.leimgruber]
+- Use excerpt header template as excerpt master document. [deiferni]
```
deiferni: the changelog has been updated for a bugfix release, this line does not belong here 😉 .

elioschmutz:
@deiferni 👍

PR #3218, opened 2017-08-03
Improve warning when opening an old version of a document

tarnap:
Resolves: #2913

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2017.4.1 (unreleased)
 ---------------------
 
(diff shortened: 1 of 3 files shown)


docs/HISTORY.txt
```diff
@@ -5,6 +5,7 @@ Changelog
 2017.4.1 (unreleased)
 ---------------------
 
+- Improve warning when opening an old document. [tarnap]
```
jone: not an old document, an old _version_ of a document

tarnap:
@jone I changed the changelog entry

PR #3219, opened 2017-08-03
Highlight search terms after fetching new search results per ajax.

### LINKED ISSUE #3077: Search results markup varies, depending on the action
# Problem
The titles of the listing in the search results vary depending on the search used.
Sometimes the titles are wrapped in spans, sometimes they're not.

Wrapped:
spanned

Unwrapped:
unspanned

# Further research

The titles are wrapped using the "upper form" or directly calling the URL (ergo, passing the text to be searched as GET parameters):
form_top_rendered
form_top

The titles are not wrapped using the form in the "content":
form_content_rendered
form_content

# Screen recording

Here's the whole behaviour:
funny2

elioschmutz:
screen2

closes #3077

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 ---------------------
 
 - Improve warning when opening an old version of a document. [tarnap]
+- Highlight search terms after fetching new search results per ajax. [elioschmutz]
 - Fix encoding error when syncing proposals with SQL. [jone]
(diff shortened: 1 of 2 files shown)


elioschmutz:
It is a plone issue. The same error occurs in a default plone.

See also my comment here https://github.com/4teamwork/opengever.core/pull/3219/files#diff-6a576c84a648cfff4d5b35c4e6f9e8dd

It's the same approach like in this PR: https://github.com/4teamwork/opengever.core/pull/1876/files

PR #3223, opened 2017-08-04
Fixed bug in search when sorting on relevance.

### LINKED ISSUE #3220: Unknown sort_on index: 'relevance'
Since (I think) the Plone 4.3.14 upgrade, we get a lot of these:

```python
CatalogError: Unknown sort_on index: 'relevance'
```

See:
https://sentry.4teamwork.ch/sentry/onegov-gever/?query=relevance

**Steps to reproduce:**
- Search for a common term (with more than 25 results) in the SearchBox and visit the full search results page
- Try to navigate to batch page 2

phgross:
The filter_query, which is customized by GEVER to handle some query parameters separate, has been refactored in the plone release. The previous method `filter_query` has been renamed to `_filter_query` and the method `filter_query` is now a wrapper of the `_filter_query`
method. Therefore the fix is only a rename of our customized `filter_query` method (see https://github.com/plone/plone.app.search/pull/18).

I've also checked, if the customization is still necessary and it is (date range queries, subject queries, empty SearchableText handling)

Closes #3220

### DIFF
--- a/opengever/base/browser/search.py
+++ b/opengever/base/browser/search.py
@@ -113,12 +113,15 @@ class OpengeverSearch(Search):
             'range': usage
         }
 
-    def filter_query(self, query):
-        """The filter query of the standard search view (plone.app.search)
+    def _filter_query(self, query):
+        """The _filter_query of the standard search view (plone.app.search)
         cancel the query generation if not SearchableText is given.
         In some case (for example in opengever.advancedsearch), we generate
         also searches without a searchabletext. So we temporarily fake
         the SearchableText.
+
+        Besides that we also handle date range queries and subject queries
+        separately.
 
         XXX This method should be removed, after the solr integration
         meta-XXX should it? now it contains custom stuff.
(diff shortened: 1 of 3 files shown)


opengever/base/browser/search.py
```diff
@@ -113,13 +113,16 @@ def handle_date_query_filter_value(self, usage, query, key, value):
             'range': usage
         }
 
-    def filter_query(self, query):
-        """The filter query of the standard search view (plone.app.search)
+    def _filter_query(self, query):
```
deiferni: This method duplicates a lot of its parents `_filter_query`. Could we consider to eliminate the duplication and do a `super` call instead?

lukasgraf: I would agree in principle, but this doesn't need to happen in this PR IMHO. The PR currently is minimally-invasive, and fixes the problem at hand.

lukasgraf:
@phgross looks good, please rebase 👍

PR #3250, opened 2017-08-08
Remove letting long tests fail.

jone:
The experience of the last days have shown that it is not wise to let long tests fail.

The main problem is that the duration can vary quite a bit, leaving us with too many flaky tests. While rewriting more and more tests with integration testing a trend to having more flaky tests is becoming apparent.

Those flaky tests are annoying and there is not enough room for fixing the issues: some actions just take that long in GEVER and optimizing those is often out of scope of a specific change.

The idea for starting that in the first place was to motivate a developer, who just writes a tests, to try to make it fast and think about the impact of certain test setups.

But at this point I see no way to achieve that without having constantly flaking tests on the CI server or foreign failing tests locally while developing.

It was an experiment, but all in all there is too much trouble and too few gain.

### DIFF
--- a/opengever/meeting/tests/test_proposal_word.py
+++ b/opengever/meeting/tests/test_proposal_word.py
@@ -13,8 +13,6 @@ from plone import api
 class TestProposalWithWord(IntegrationTestCase):
     features = ('meeting', 'word-meeting')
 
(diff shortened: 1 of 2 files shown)


jone:
Rebased in order to trigger CI.

PR #3253, opened 2017-08-09
Word meeting: Improve edit-document-button behavior in meeting view.

jone:
This pull-request improves the behavior and appeareance of the
edit-document button in the meeting view. The button is available per agenda item. When clicked, the proposal document is checked out and opened with office connector.

New behavior of the button:
- ~~The button is available when the agenda item is in the "pending"
  state or in the "revision" state.~~
- The button is only visible when the user has the permission to edit the document.
- The button is active as long as I can checkout and edit.
- The button is inactive when the document is checkout out by somebody
  else in order to communicate to the user that he could edit the
  document if it wouldn't be checked out.

bildschirmfoto 2017-08-09 um 10 18 05
bildschirmfoto 2017-08-09 um 10 18 38
bildschirmfoto 2017-08-09 um 10 19 07

### DIFF
--- a/opengever/meeting/browser/meetings/agendaitem.py
+++ b/opengever/meeting/browser/meetings/agendaitem.py
@@ -139,6 +139,30 @@ class AgendaItemsView(BrowserView):
         docs = IContentListing(submitted_proposal.get_excerpts())
         return [doc.render_link() for doc in docs]
 
+    @require_word_meeting_feature
+    def _get_edit_document_options(self, meeting, agenda_item):
+        """Return the agenda item options for the template regarding
+        editing the document and its lock.
+        """
+        proposal = agenda_item.proposal.submitted_oguid.resolve_object()
+        document = proposal.get_proposal_document()
+        checkout_manager = getMultiAdapter((document, self.request),
+                                           ICheckinCheckoutManager)
+
+        button = {}
+        button['visible'] = bool(checkout_manager.check_permission(
+            'Modify portal content'))
+        button['active'] = button['visible'] and (
+            checkout_manager.is_checkout_allowed() or
+            checkout_manager.is_checked_out_by_current_user())
+        button['url'] = meeting.get_url(
+            view='agenda_items/{}/edit_document'.format(
+                agenda_item.agenda_item_id))
+        return {
+            'proposal_document_checked_out': bool(
+                checkout_manager.get_checked_out_by()),
+            'edit_proposal_document_button': button}
+
     def _get_agenda_items(self):
         meeting = self.context.model
         agenda_items = []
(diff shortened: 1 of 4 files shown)


opengever/meeting/browser/meetings/agendaitem.py
```diff
@@ -139,6 +139,31 @@ def _serialize_excerpts(self, item):
...
+        """Return the agenda item options for the template regarding
+        editing the document and its lock.
+        """
+        proposal = agenda_item.proposal.submitted_oguid.resolve_object()
+        document = proposal.get_proposal_document()
+        checkout_manager = getMultiAdapter((document, self.request),
+                                           ICheckinCheckoutManager)
+
+        button = {}
+        button['visible'] = agenda_item.get_state() in (
```
deiferni: i would move this check to the agenda item, maybe in an `is_editable` method?

opengever/meeting/tests/test_agendaitem_word.py
```diff
@@ -40,26 +40,90 @@ def test_proposal_document_in_meeting_item_data(self, browser):
...
+                 'visible': True,
+                 'active': True,
+                 'url': self.agenda_item_url(agenda_item, 'edit_document')}},
             item_data)
 
-        document = self.submitted_word_proposal.get_proposal_document()
-        self.checkout_document(document)
-        browser.reload()
+    @browsing
+    def test_edit_document_possible_when_I_have_checked_it_out(self, browser):
```
deiferni: I -> i

opengever/meeting/browser/meetings/agendaitem.py
jone: My first intention was the same: having an `is_editable` method.
But I then realized that I want to distinguish "it is possible to edit the document at this point" from "I can actually edit the document":
When it is possible to edit the document in general, the button should be visible.
But when the document is checked out by someone else, the button should be deactivated.

I probably would expect from the name `is_editable` that it also verifies if it is checked out by someone else and that it does permission checks.
But if it does that I cannot have the behavior in the view that I want.
That's why I didn't extract it for now.

If we extract it we must come up with better names indicating what actually happens 🤔

opengever/meeting/tests/test_agendaitem_word.py
jone: A long time ago i've learned in school that i should always capitalize `I` 😜 
https://en.wikipedia.org/wiki/Capitalization#Pronouns
Do you think that does not apply for method names?

opengever/meeting/browser/meetings/agendaitem.py
deiferni: `is_editable_by_current_user()`
`is_editable_by(api.user.get_current())`

🤔 ?

opengever/meeting/tests/test_agendaitem_word.py
deiferni: yes 😉 
at least when following the PEPs there is no capitalization in method names. given a good reason i could live with it though 😉 .

jone: fair enough, updated 👍

opengever/meeting/browser/meetings/agendaitem.py
jone: The correct place for deciding whether the document can be edited is actually the document, not the agenda item (otherwise I could still edit it in the popup).

This means we need to submitted-proposal-workflow to control whether the proposal document is editable or not. This means we need to add workflow states and execute transitions when the agenda item state changes.

@deiferni thoughts about that?

deiferni: uhm. also on the document itself the available actions should be the same as in the overlay and the same as those that are available directly from the agenda item list. not sure if we want to add workflow states for checkin/checkout though.

maybe defer a "clean" implementation until we have converted agenda items to plone-objects with their own workflow and see if that helps?

jone: Checkout/checkin-states wouldn't be necessary, but workflow states on the parent submitted-proposal: a "decided"-state should prevent from modification.

I think we'll have to do that eventually, but I'd like to do it when rewriting the proposal-workflows with lawgiver, therefore deferring it to later would be best.

So we'll need to change it here anways. Do we want to do something about it yet or just ship it like this and improve it later? 🤔

deiferni: hmm ... maybe omit the point:

- The button is available when the agenda item is in the "pending" state or in the "revision" state.

completely? then although not 100% correct regarding agenda-item state the available actions would at least be consistent, right?

jone: I can do that; although this PR will then only be a refactoring and not change the behavior, as far as I see, because one can still edit the document in all states. 🤔 

I think I'll update it so that I check for a permission on the document instead; then it will automatically adapt to a new workflow.

deiferni: sounds good 👍

jone:
@deiferni I've updated the change according to our inline-discussion:
I have removed the test `test_edit_document_not_possible_when_agenda_item_decided` because I'm not yet implementing it.
I have changed the visibility to be solely based on checking for the `Modify portal content`-permission on the proposal document. It will adapt to a workflow change later and the button.

PR #3266, opened 2017-08-10
Add bumblebee gallery view for sablontemplate.

### LINKED ISSUE #3259: Tab "Sablon Vorlagen": Bumblebee funktioniert nicht
Bei dem Tab "Sablon Vorlagen" im Vorlagenordner geht die Bumblebee-Ansicht nicht.
Genau gleiches Problem wie in #3251 beschrieben.

@elioschmutz könntest du dies auch noch gleich anschauen, wenn du schon grad dran bist? 😜

elioschmutz:
This PR adds the bumblebee-gallery view to the sablontemplate tab.

bildschirmfoto 2017-08-10 um 11 10 50

closes #3259

### DIFF
--- a/opengever/core/upgrades/20170810105607_make_sablon_templates_i_bumblebeeable/upgrade.py
+++ b/opengever/core/upgrades/20170810105607_make_sablon_templates_i_bumblebeeable/upgrade.py
@@ -1,0 +1,24 @@
+from ftw.bumblebee import get_service_v3
+from ftw.bumblebee.interfaces import IBumblebeeDocument
+from ftw.upgrade import UpgradeStep
+from opengever.bumblebee import is_bumblebee_feature_enabled
+from plone import api
+
+
+class MakeSablonTemplatesIBumblebeeable(UpgradeStep):
+    """Make sablon templates i bumblebeeable.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
+        catalog = api.portal.get_tool('portal_catalog')
+        query = {'portal_type': 'opengever.meeting.sablontemplate'}
+        msg = 'Add bumblebee-previews for sablontemplates.'
+
+        for obj in self.objects(query, msg):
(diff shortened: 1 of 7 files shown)


opengever/core/upgrades/20170810105607_make_sablon_templates_i_bumblebeeable/upgrade.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def __call__(self):
+        self.install_upgrade_profile()
+        catalog = api.portal.get_tool('portal_catalog')
+        query = {'portal_type': 'opengever.meeting.sablontemplate'}
+        msg = 'Reindex object_provides for sablontemplates.'
+
+        for obj in self.objects(query, msg):
+            catalog.reindexObject(obj, idxs=['object_provides'],
+                                  update_metadata=False)
```
jone: I was not aware that bumblebee was not activated at all for sablon templates.
I think we'd need to do things such as updating the checksum in order for bumblebee to work 🤔 

As this is bigger than I thought: @deiferni what do you think about enabling bumblebee for sablon in general?

(Alternatively we could remove the bumblebee action from the tabbed view if we do not want bumblebee support here).

deiferni: So far it has not been deemed necessary since only we can edit the sablon templates and we're fading them out asap in any case.

From a technical perspective sablon templates are word files so it should work. Maybe have a look at how the converter performs with the many merge-fields first, though? Also as already mentioned please don't forget to calculate the checksums (and store if the bumblebee feature is enabled). I would assume this can be done in an upgrade step since the amount of bumblebee templates is very low 🤔 .

elioschmutz:
@deiferni @jone 👍  I Updated the PR with storing and checksum calculation ( 😞  )

PR #3270, opened 2017-08-10
Bundle Import: Reduce memory footprint

lukasgraf:
These changes aim to reduce the memory (RSS) footprint of a bundle import, particularly the high-water-mark.

While I was able to clean up some leaking references in earlier pull requests (x, y), that wasn't enough to significantly bring down max. memory usage.

At some point I discovered that re-setting the Plone site using the `getSite()` hook finally manages to get rid of those leaking references, and allows the Python garbage collector to do its job. In addition, the `cPickleCache` also needs to be garbage collected periodically.

So, in regular intervals the following things need to happen:

- Creating transaction savepoints
- Re-setting the Plone site using setSite()
- Garbage collecting the cPickleCache
- Python garbage collection (optional, doesn't effect high-water-mark)

I can't yet say what exact impact the interval has, and whether this is the optimal order, but at least the first three need to happen in order to get a significant reduction in the memory high-water-mark.

memory

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -12,6 +12,8 @@ Changelog
 - Add add bumblebee gallery view for proposaltemplate. [elioschmutz]
 - Word meeting: Improve edit-document-button behavior in meeting view. [jone]
 - Fix typo in contact detail view. [elioschmutz]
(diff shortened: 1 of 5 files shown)


lukasgraf:
@phgross thise one is now ready, and tested again.

PR #3271, opened 2017-08-14
Fixture: improve documentation and assert readme updated.

jone:
1) Fix the tree in the fixture documentation, so that it works with real Plone objects as well as SQL wrappers. The problem was that some were prefixed with the Plone site path, others were relative to the Plone site. Since the IntegrationTestCase traverses on the Plone site it is correct to make all paths relative to the Plone site.

2) Let the test fail if the readme was udpated when running the test. This does only fail once and then leave an uncommitted, changed readme behind, which must be committed be the developer. On the CI-server it will fail when a PR changes the fixture but misses the readme update.

bildschirmfoto 2017-08-14 um 10 06 12

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -481,36 +481,36 @@ Objects
 
 .. code::
 
-  - self.templates
-    - self.proposal_template
-    - self.sablon_template
+  - self.committee_container
+    - self.committee
+      - self.meeting
+      - self.submitted_proposal
+      - self.submitted_word_proposal
+    - self.empty_committee
   - self.repository_root
     - self.branch_repofolder
       - self.leaf_repofolder
         - self.archive_dossier
-        - self.meeting_dossier
         - self.dossier
-          - self.mail_msg
-          - self.mail_eml
           - self.document
-          - self.proposal
-          - self.task
-            - self.taskdocument
-            - self.subtask
-          - self.word_proposal
           - self.draft_proposal
           - self.draft_word_proposal
-          - self.subdossier2
+          - self.mail_eml
+          - self.mail_msg
+          - self.proposal
           - self.subdossier
             - self.subdocument
+          - self.subdossier2
+          - self.task
+            - self.subtask
+            - self.taskdocument
+          - self.word_proposal
         - self.empty_dossier
+        - self.meeting_dossier
     - self.empty_repofolder
-  - self.committee_container
-    - self.empty_committee
-    - self.committee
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -481,36 +481,36 @@ Objects
...
   - self.repository_root
-    - self.empty_repofolder
     - self.branch_repofolder
       - self.leaf_repofolder
-        - self.meeting_dossier
         - self.archive_dossier
-        - self.empty_dossier
+        - self.meeting_dossier
         - self.dossier
+          - self.mail_msg
```
deiferni: could we sort these entries alphabetically for better readability?

opengever/testing/tests/test_dump_fixture_structure.py
```diff
@@ -32,6 +38,10 @@ def test_dump_fixure(self):
                 self._render_tree(self._build_tree(data['object']),
                                   level=1)))
 
+        md5_after = self._readme_md5()
+        self.assertEquals(md5_before, md5_after,
+                          'Fixture has changed and readme was not updated.')
```
deiferni: to be extra nice you could also add a hint and say how to fix this

jone:
@deiferni updated according to your suggestions.

```
Failure in test test_dump_fixure (opengever.testing.tests.test_dump_fixture_structure.TestDumpFixtureStrutureToReadme)
  File "/Users/jone/projects/python/cache/eggs/unittest2-0.5.1-py2.7.egg/unittest2/case.py", line 340, in run
    testMethod()
  File "/Users/jone/projects/packages/opengever.core/opengever/testing/tests/test_dump_fixture_structure.py", line 45, in test_dump_fixure
    'Fixture has changed and readme was not updated.\n'
  File "/Users/jone/projects/python/cache/eggs/unittest2-0.5.1-py2.7.egg/unittest2/case.py", line 521, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/Users/jone/projects/python/cache/eggs/unittest2-0.5.1-py2.7.egg/unittest2/case.py", line 514, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: '05cc523af852f327249c882fb0ee3a28' != 'f467a6f01ddb09181a90b376a5c5e84e' : Fixture has changed and readme was not updated.
run:
 $ bin/test -t TestDumpFixtureStrutureToReadme
then commit the changes in the README.rst.
```

PR #3278, opened 2017-08-15
Add bin/test-cached helper script to execute a chached test-run.

deiferni:
This PR adds a simple wrapper script to enable cached testing on a per test-run basis. Motivation:

- i always forget the env-variable name and have to spend time figuring it out again 😢 
- i'd like to have access to a script that just remembers the name for me in an auto-completable, quick way

objections/suggestions?

### DIFF
--- a/development.cfg
+++ b/development.cfg
@@ -12,6 +12,7 @@ ogds-db-name = opengever
 
 development-parts +=
     ${buildout:sphinx-parts}
+    test-cached
 
 # this re-adds parts that would be dropped otherwise since buildout cannot really deal
 # with our complicated inheritance hierarchy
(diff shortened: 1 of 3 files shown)


development.cfg
```diff
@@ -50,3 +51,11 @@ recipe = collective.recipe.cmd
...
 on_update=true
 cmds=cp i18n-build.in bin/i18n-build && chmod +x bin/i18n-build
+
+[test-cached]
+recipe = collective.recipe.template
+output = ${buildout:directory}/bin/test-cached
+mode = 0755
+input = inline:
+    #!/usr/bin/env sh
+    GEVER_CACHE_TEST_DB=true bin/test "$@"
```
jone: Why do we need to rebuild that on buildout time? It has no variable parts 🤔 
I'd prefer commit the resulting `bin/test-cached` script and un-ignore it in `.gitignore`

Each added buildout part will take a moment of our time when we run buildout. If we can awoid with no costs, we should.

deiferni: yes the thing is i don't want to use it permanently yet. 

also - you're right, we've done that for some of the other scripts already. now the PR has been merged already,  so i'll crate another one that reverts.

PR #3279, opened 2017-08-15
Add bin/test-cached helper script to execute a chached test-run.

deiferni:
This PR adds a simple wrapper script to enable cached testing on a per test-run basis. Motivation:

- i always forget the env-variable name and have to spend time figuring it out again 😢
- i'd like to have access to a script that just remembers the name for me in an auto-completable, quick way

Also revert the already merged dynamically buildout-ed script in favour of one that is under version control directly, see https://github.com/4teamwork/opengever.core/pull/3278.

### DIFF
--- a/bin/test-cached
+++ b/bin/test-cached
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+#
+# USAGE:
+# bin/test-cached [any options/arguments supplied to bin/test]
+#
+set -euo pipefail
+
+GEVER_CACHE_TEST_DB=true bin/test "$@"
--- a/README.rst
+++ b/README.rst
@@ -776,6 +776,12 @@ You can enable the feature by setting an environment variable:
 
     GEVER_CACHE_TEST_DB=true bin/test -m opengever.dossier.tests.test_activate
 
+There is also a binary which does that for you for just one run for convenience:
+
+.. code:: sh
+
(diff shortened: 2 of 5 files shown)


bin/test-cached
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+#
+# USAGE:
+# bin/test-cached [any options/arguments supplied to bin/test]
+#
+set -euo pipefail
+
+GEVER_CACHE_TEST_DB=true bin/test "$@"
```
jone: I think it should exit with the exit code of `bin/test`.
e.g.:
```sh
GEVER_CACHE_TEST_DB=true bin/test "$@"
exit $?
```
or 
```sh
GEVER_CACHE_TEST_DB=true bin/test "$@" || exit $?
```

The exit code is not that important when executed locally, but it would be consistent and correct.

Rotonen: As long as we're not planning on integrating into this I fail to see the point.

deiferni: afaik the default behaviour of such a script is to exit with the exit code of the last command. so it already does what you want the way it is 😉 .

jone: @Rotonen for example pairing it with `git bisect` is quite useful; then the exit code matters.

@deiferni You're absolutely right; no need to change it 👍

PR #3282, opened 2017-08-15
Add a CI script to check the files changed in your branch with pyflakes

Rotonen:
* Finds out which files have changed since the branching point
* Ignores `__init__.py` files
* Cleans up pyflakes errors from the project

Example output from a failed run:
https://ci.4teamwork.ch/builds/102528/tasks/153309

Example output from a succesful run:
https://ci.4teamwork.ch/builds/102529/tasks/153315


Rotonen:
@phgross I'm requesting the right of way for this pull request as rebasing this can get painful quick over time.

opengever/base/monkey/patches/cmf_catalog_aware.py
```diff
@@ -77,7 +77,6 @@ def reindexObject(self, idxs=[]):
             return original_reindexObject(self, idxs)
 
         from Products.CMFCore.CMFCatalogAware import CMFCatalogAware
-        __patch_refs__ = False
```
jone: That's actually used in the monkey patcher: https://github.com/4teamwork/opengever.core/blob/master/opengever/base/monkey/patching.py#L58

I propose to change that to:
```python
locals()['__patch_refs__'] = False
```
This has the same affect without popping up in pyflakes.

Same for the other places.

opengever/globalindex/tests/test_query.py
```diff
@@ -42,7 +42,6 @@ def task(self, intid, current_admin_unit, sequence_number=1, **kw):
     def test_in_pending_state_returns_only_pending_tasks(self):
         task1 = self.task(1, 'xx', review_state='task-state-open')
         task2 = self.task(2, 'yy', review_state='task-state-resolved')
-        task3 = self.task(3, 'zz', review_state='task-state-cancelled')
```
jone: A non-pending task is needed for test to proof that it does _not_ appear in the assertion below.
Therefore I'd change that to creating a task, but not assigning it to an (unused) variable.

Rotonen: Thank you:
https://github.com/4teamwork/opengever.core/pull/3282/commits/ffe5a7e20312e8b8e5d0d37d27345324e0b10f92#diff-584095255ebaa1a91180a554b4ca24a4R45

opengever/base/monkey/patches/cmf_catalog_aware.py
Rotonen: Thank you:

https://github.com/4teamwork/opengever.core/pull/3282/commits/60aeb7dac9b6ae7be22f8684f18e3f3aedc3b9aa#diff-2c05c3981688af00d8bf40f8367e8808R80

https://github.com/4teamwork/opengever.core/pull/3282/commits/60aeb7dac9b6ae7be22f8684f18e3f3aedc3b9aa#diff-5c41c4347364f407d23c44d7f6f4d9beR243

https://github.com/4teamwork/opengever.core/pull/3282/commits/60aeb7dac9b6ae7be22f8684f18e3f3aedc3b9aa#diff-5c41c4347364f407d23c44d7f6f4d9beR260

https://github.com/4teamwork/opengever.core/pull/3282/commits/60aeb7dac9b6ae7be22f8684f18e3f3aedc3b9aa#diff-5c41c4347364f407d23c44d7f6f4d9beR291

opengever/globalindex/tests/test_query.py
```diff
@@ -177,7 +177,6 @@ def test_by_container_handles_similar_paths_exactly(self):
 
         dossier_11 = create(Builder('dossier').titled(u'Dossier 11'))
         dossier_11 = api.content.rename(obj=dossier_11, new_id='dossier-11')
-        task2 = create(Builder('task').within(dossier_11))
```
phgross: @Rotonen same here as above (see comment from jones).

`task2 = create(Builder('task').within(dossier_11))` -> `create(Builder('task').within(dossier_11))`

opengever/ech0147/bindings/ech0010.py
```diff
@@ -11,7 +11,6 @@
 import io
```
phgross: As far as I know the eCH binding files are generated automatically ... not sure if we should "pyflake" them or maybe doublecheck how there are generated.

test-pyflakes.cfg
```diff
@@ -0,0 +1,53 @@
...
+[bin-test-jenkins]
+recipe = collective.recipe.template
+input = inline:
+    #!/bin/bash
+    fork_point=$(git merge-base --fork-point origin/master)
+    files=$(git diff --name-only "$fork_point" \
+                | egrep '.py$' \
+                | grep -v __init__.py \
+                | grep -v bootstrap.py \
+                | grep -ve '*/skins/*' \
```
deiferni: i would suggest to also ignore everything in a `bindings` folder. that would include xml-binding files like https://github.com/4teamwork/opengever.core/blob/master/opengever/ech0147/bindings/ech0010.py and https://github.com/4teamwork/opengever.core/blob/master/opengever/disposition/ech0160/bindings/dokument.py. also it would help us to establish the convention to always put xml bindings in a bindings package and nowhere else.

opengever/disposition/ech0160/bindings/ablieferung.py
```diff
@@ -146,7 +145,7 @@ class datumTypB (pyxb.binding.basis.STD_union):
         * MM indicates the month
         * DD indicates the day
         Hinweis: Alle Komponenten werden benötigt!
-
```
deiferni: oh, now i've seen that you changed the xml bindings. these files are auto-generated with `PyXB`. i don't see a point editing them and would honestly advise against doing so. should we need to regenerate them the changes will be lost.

Rotonen:
@deiferni, @phgross, I discussed the bindings with @jone and we concluded you only end up in the bindings in the worst of debugging scenarios and the improved readability will be appreciated then.

The effort to manually clean up the generated bindings is a matter of a about half an hour (even accounting for for pep8 and pep257). I'm also planning on tackling the generator itself the next time we get these (or more of these( generated. I'm also assuming these to be relatively stable and not requiring constant regeneration.

opengever/globalindex/tests/test_query.py
Rotonen: Thank you:
https://github.com/4teamwork/opengever.core/pull/3282/commits/f9eb0e8752e4e9e7e44cc6d19f5019e9f0e426a7#diff-584095255ebaa1a91180a554b4ca24a4R180

Rotonen:
I will then just not touch anything from the bindings and add an exception to the CI runner to not flag further incoming bindings.

Rotonen:
@deiferni ready for rereview.

opengever/examplecontent/hooks.py
```diff
@@ -12,6 +12,9 @@
 from plone import api
 import pytz
 
+# Please pyflakes
+builders
```
deiferni: i'm also a bit unhappy to have unnecessary lines like these just to please the QA tool 😞 . have we discussed using `flake8` instead, afaik that could handle the `# noqa` comments that we already use.

_just for the record though, we can discuss this after merging this PR_

Rotonen: Not happy about it either. We're contemplating between flake8 or pylama as a wrapper if this gets too much out of hand. They do come with their very own bundles of associated integration trouble, so we'd rather avoid the wrappers if at all possible at a minimal cost of such nonsense. So far the only thing encountered are the test factory builders in `opengever.core` which do module import time magic without getting used in the files themselves.

PR #2852, opened 2017-04-13
OGIP 15 (Sammelaufgaben)

maethu:
Goal:
It should be possible to add multiple task with the add task form. For every responsible user there should be one separate task.

[Link to OGIP]

Tasks:
- [x] Add New `All` option to the add/edit task form.
- [x] New default value for the client field while adding tasks is `All`.
- [x] Make sure the new `All` option  is only available on tasks (add/edit)
- [x] Add multiple tasks if there's more than one user. Use plone dexterity form api.
- [x] Implement a new select2 based widget. #2910
- [x] Replace other gever specific autocomplete selection fields #2932 
- [x] Changelog

Styling: https://github.com/4teamwork/plonetheme.teamraum/pull/550

PR #3290, opened 2017-08-17
Filter results from disabled orgunits.

### LINKED ISSUE #3273: Orgunit's Inactive flag not respected by the new responsible fields
The new vocabulary sources used in ITask.responsible etc. (OGIP 15), provides also entries for inactive OrgUnits. Users from an inactive OrgUnit should newer be selectable.

maethu:
The problem was introduced with the OGIP 15 (https://github.com/4teamwork/opengever.core/pull/2852), which turned all vocabularies (IVocabulary) into sources (ISource) to support a new kind of selection widget provided by ftw.keywordwidget.

With this change filtering for disabled OrgUnits went lost.
This PR brings back this feature.

I also slightly refactored the source a little bit, otherwise this change would introduced a lot of duplicated code. 

I also hat to fix a side effect on the advanced search form.
A test-odgs user was not properly configured, since a OrgUnit was missing.
This change basically no longer allows user without an OrgUnit, since it is required to build the query.

Fixes #3273

### DIFF
--- a/opengever/ogds/base/sources.py
+++ b/opengever/ogds/base/sources.py
@@ -53,7 +53,7 @@ class AllUsersAndInboxesSource(object):
         query = create_session().query(*models) \
                                 .join(OrgUnit.users_group) \
                                 .join(Group.users)
-        query = query.filter(OrgUnit.enabled)
+        query = query.filter(OrgUnit.enabled == True)
 
         if self.only_current_orgunit:
             query = query.filter(OrgUnit.unit_id == self.client_id)
@@ -165,7 +165,7 @@ class AllUsersAndInboxesSource(object):
                               text_filters)
 
         query = OrgUnit.query
-        query = query.filter(OrgUnit.enabled)
+        query = query.filter(OrgUnit.enabled == True)
 
         if self.only_current_inbox:
             query = query.filter(OrgUnit.unit_id == self.client_id)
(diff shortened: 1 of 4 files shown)


maethu:
I'm struggling, sry...

Seems like my change has some more implications than I thought.

maethu:
@4teamwork/gever 
Well, I kicked the job again an it passes now 🤔 
I guess this is now ready to review

opengever/ogds/base/sources.py
```diff
@@ -40,6 +40,7 @@ def base_query(self):
         query = create_session().query(User, OrgUnit) \
                                 .join(OrgUnit.users_group) \
                                 .join(Group.users)
+        query = query.filter(OrgUnit.enabled)
```
lukasgraf: This relies on implicit conversion of this expresion to boolean.

[`Query.filter()`](http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.filter) doesn't show an example of a single-value expression without comparison operators - so I think SQLAlchemy is magically doing `bool(OrgUnit.enabled) == True` behind the scenes for your case.

But I'd rather see an explicit `filter(OrgUnit.enabled == True)` here (and the other comparisons), relying on SQLAlchemy expression magic too much can go wrong and cause very inefficient SQL queries.

maethu: @lukasgraf thx good catch, was not aware of that. 👍

PR #3291, opened 2017-08-18
Robusitfy the pyflakes CI script

Rotonen:
I had an errant placment of an echo statement there.

It also now handles the case of 'no unignored python files changed'.


Rotonen:
And also actually works as advertised:
https://ci.4teamwork.ch/builds/103167/tasks/154472

PR #3297, opened 2017-08-18
Only run the XML format check for changed xml files

Rotonen:
Include XML, ZCML. Exclude `definition.xml`, `*/upgrades/*`.

### DIFF
--- a/test-xml-convention.cfg
+++ b/test-xml-convention.cfg
@@ -5,3 +5,41 @@ extends =
 
 package-name = opengever.core
 package-namespace = opengever
+
+[test-jenkins]
+recipe = collective.recipe.template
+output = ${buildout:bin-directory}/test-jenkins
+mode = 755
+input = inline:
+    #!/bin/bash
+
+    fork_point=$(git merge-base --fork-point origin/master)
+    files=$(git diff --name-only "$fork_point" \
+                | egrep '(.xml|.zcml)$' \
+                | grep -v definition.xml \
+                | grep -ve '*/upgrades/*' \
+                | tr '\r\n' ' ')
+
+
+    if [[ "$files" = *[!\ ]* ]]
+    then
+        echo 'Detected changes in this branch in the following Python files:'
+        echo "$files" | tr ' ' '\n'
+        echo 'Running bin/format-xml --check:'
+        eval ${buildout:bin-directory}/format-xml --check "$files"
+        result=$?
+    fi
+
+    if [[ "$result" == 0 ]]
+    then
+       echo 'All XMLs formatted correctly.'
+    elif [ -z "$result" ]
+    then
+        echo ''
+        echo 'No XML files changed.'
+        echo ''
+        exit 0
+    fi
+    echo ''
+    echo 'The files above failed the format check!'
+    exit $result


phgross:
@Rotonen whats the reason for ignoring workflow definitions (`definition.xml`) and upgradesteps ?

test-xml-convention.cfg
```diff
@@ -5,3 +5,41 @@ extends =
...
+recipe = collective.recipe.template
+output = ${buildout:bin-directory}/test-jenkins
+mode = 755
+input = inline:
+    #!/bin/bash
+
+    fork_point=$(git merge-base --fork-point origin/master)
+    files=$(git diff --name-only "$fork_point" \
+                | egrep '(.xml|.zcml)$' \
+                | grep -v definition.xml \
```
deiferni: question: why the excludes? we reformatted all xmls some time ago. imho we could enforce that all of them remain formatted.

Rotonen:
@phgross @deiferni no idea - I'm mirroring existing behaviour.

https://github.com/4teamwork/ftw-buildouts/blob/8abbc163554af8c9b5d297f055399a31d7c7cf7a/format-xml.cfg#L44

Rotonen:
This should not be merged before being refactored into the qa test run.

test-xml-convention.cfg
jone: https://github.com/4teamwork/ftw-buildouts/commit/da958dd5a78c34586168686ed08dbd8b9f1aa4ab

The reason is that workflow definitions are in general generated with lawgiver. They are never edited manually and do not comply to those standards.

GEVER is the only product where we still take the risk of crafting workflows manually. But ignoring the `definition.xml` comes from ftw-buildouts, which is company-wide.
We also already have some lawgiver workflows in GEVER, so there is the problem.

I will not change `ftw.lawgiver` so that the generated workflows comply to our conventions, because:
- it would require to update all projects using `ftw.lawgiver` (~180 deployments)
- there is no python XML library which is able to produce a valid XML document; it can only be achived by using regexp
- the output is correct from an XML point of view

---

Since we want to move towards lawgiver in GEVER and thus will no longer edit workflow definitions manually in the far future, I recommend to keep `definition.xml` files ignored.
If we agree to not do so it must be "fixed" in GEVER and must not make it needlessly complicated to edit the existing lawgiver workflows.

test-xml-convention.cfg
```diff
@@ -5,3 +5,41 @@ extends =
...
+                | grep -ve '*/upgrades/*' \
+                | tr '\r\n' ' ')
+
+
+    if [[ "$files" = *[!\ ]* ]]
+    then
+        echo 'Detected changes in this branch in the following Python files:'
```
jone: `the following Python files:`
I believe those should be XML files, not python.

jone:
What's the reason to fork ftw-buildouts part instead of improving it in ftw-buildouts?

test-xml-convention.cfg
deiferni: makes perfect sense 😄 👍  linting auto-generated files is usually not worth the hassle

PR #3301, opened 2017-08-21
Fix and split up the pyflakes CI run script

### LINKED ISSUE #3299: Pyflakes check can't deal with deleted files
There seems to be an issue with the check in combination with deleted files in a PR, see https://ci.4teamwork.ch/builds/103672/tasks/155330 for a failing example.

Rotonen:
* Split the git fork point finding onto its own script
* Make a `bin/pyflakes-ci` script to be able to write unrestricted-by-buildout-internals bash
* Wrap that script with a trivial `bin/test-jenkins` via `collective.recipe.template`
* Fail per default
* Split up the parametres in a way which can be useful for future QA script code sharing

Closes #3299.

### DIFF
--- a/bin/.gitignore
+++ b/bin/.gitignore
@@ -3,7 +3,9 @@
 !cleanup-versions-cfg
 !create-schema-migration
 !create-upgrade
+!git-fork-point
 !mtest
 !pinchecker
 !remove-test-cache
 !test-cached
+!pyflakes-ci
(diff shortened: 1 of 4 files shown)


Rotonen:
Once this is merged, I'll port the other QA tools to use this as well. Unifying them under one interface is still out of scope for now, but the scaffolding towards that is mostly done.

PR #3302, opened 2017-08-21
Fix pyflakes-ci script. Correctly call pyflakes with file arguments.

deiferni:
@Rotonen just fixed this on my machine and i'll steal the contribution 😈 .


Rotonen:
@deiferni and should finally work now.

PR #3303, opened 2017-08-21
Point source-code link to 4teamwork/opengever.core.

deiferni:
As discussed in https://4teamwork.slack.com/archives/C060FEYKA/p1502960325000375 we'd like to update the footer link and point it to this repo instead of https://github.com/OneGov/onegov.gever.

What about the link to CI?
screen shot 2017-08-21 at 17 54 18
It currently points to a view in Jenkins, but that view contains outdated and manually configured dependencies. I propose to:
- remove the link to CI in the footer
- also remove the view, i don't think that we use it anymore since CI-Governor has take over, also the list is fairly outdated

objections (@phabegger, @phgross , @lukasgraf, @jone)?

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 ---------------------
 
 - Add upgrade step which fixes potential missing titles on opengever.private.root objects. [mathias.leimgruber]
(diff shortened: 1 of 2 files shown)


deiferni:
at first i didn't write one but it can't hurt so there it its. also link to CI is dropped.

PR #3304, opened 2017-08-22
Add upgrade step which fixes potential missing titles on opengever.private.root objects

### LINKED ISSUE #2911: Manually created private repositories missing title
Private repository containers (`/fd/private`) that have been created manually on deployments where they weren't created as part of the setup are missing a title.

For example on dev.onegovgever.ch and test.onegovgever.ch this is the case. This leads to errors, for example when trying to edit the navigation Portlet settings on the inbox (https://dev.onegovgever.ch/fd/eingangskorb/++contextportlets++plone.leftcolumn/navigation/edit?)

An upgrade step needs to be created that sets the title for these private repository containers.

_(I've fixed it manually on https://dev.onegovgever.ch/rk because I needed editing the nav portlet to work)_

maethu:
On my local machine I was only able to reproduce this state by setting manually `title_fr` and `title_de` to a empty string.
private

Hope I'm not missing anything. 

Closes #2911

### DIFF
--- a/opengever/core/upgrades/20170822083925_set_missing_titles_on_private_root/upgrade.py
+++ b/opengever/core/upgrades/20170822083925_set_missing_titles_on_private_root/upgrade.py
@@ -1,0 +1,22 @@
+from ftw.upgrade import UpgradeStep
+from opengever.base.behaviors.translated_title import ITranslatedTitle
+
+
+class SetMissingTitlesOnPrivateRoot(UpgradeStep):
+    """Set missing titles on private root.
+    """
+
+    def __call__(self):
+        query = {'portal_type': ['opengever.private.root']}
+        for folder in self.objects(query, 'Set title on private root folders'):
+            idxs = []
+
+            if not ITranslatedTitle(folder).title_de:
+                ITranslatedTitle(folder).title_de = u'Meine Ablage'
+                idxs.append('title_de')
+            if not ITranslatedTitle(folder).title_fr:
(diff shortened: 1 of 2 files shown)


opengever/core/upgrades/20170822083925_set_missing_titles_on_private_root/upgrade.py
```diff
@@ -0,0 +1,22 @@
...
+class SetMissingTitlesOnPrivateRoot(UpgradeStep):
+    """Set missing titles on private root.
+    """
+
+    def __call__(self):
+        query = {'portal_type': ['opengever.private.root']}
+        for folder in self.objects(query, 'Set title on private root folders'):
+            idxs = []
+
+            if not ITranslatedTitle(folder).title_de:
```
deiferni: could you run into an attribute error here with incorrectly initialised objects (could maybe be the case for stuff setup with `ftw.inflator`) or do default values take care of that and return `None`?

maethu: @deiferni As long as `title_de` and `title_fr` are defined in the schema, this will work.
Do you think I should check for the attribute first?

deiferni: well … i was not completely sure about the behaviour here, but if your certain it will work without the attribute check is not necessary in my opinion 😉

Issue #3124, opened 2017-07-13
Speed up checkin / checkout


jone:
The checkin-checkout-manager executes [`self.document.reindexObject()`](https://github.com/4teamwork/opengever.core/blob/master/opengever/document/checkout/manager.py#L153) on changes, which also reindexes the searchable text etc.

It think we could only reindex the relevant indexes, probably `self.document.reindexObject(idxs=['checked_out'])`.

- Find out which indexes actually change
- Only index those indexes
- Update all 3 places in the checkin/checkout-manager.

phgross:
Sounds reasonable. Deos Office Connector also updates the SearchableText when uploading the changed document 🤔 ? Otherwise the reindex of SearchableText is necessary when checking in a Document.

phgross:
I've checked doublechecked the current behavior.
 - Office Connector updates the searchable Text also when replacing the file via the edit form.
 - But replacing the document via D'n'D does not updated the SearchableText.

phgross:
Solved via #3311

PR #3311, opened 2017-08-23
Optimize reindexing for document checkout/checkin cycles

Rotonen:
* I've plugged the D'n'D upload reindexing hole
* Only reindex the `checked_out` index

### DIFF
--- a/opengever/document/checkout/manager.py
+++ b/opengever/document/checkout/manager.py
@@ -89,7 +89,14 @@ class CheckinCheckoutManager(object):
         self.annotations[CHECKIN_CHECKOUT_ANNOTATIONS_KEY] = user_id
 
         # finally, reindex the object
-        self.context.reindexObject()
+        catalog = api.portal.get_tool('portal_catalog')
+        catalog.reindexObject(
+            self.context,
+            idxs=(
+                'checked_out',
+                ),
+            update_metadata=True,
+            )
 
         # fire the event
         notify(ObjectCheckedOutEvent(self.context, ''))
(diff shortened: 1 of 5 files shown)


opengever/document/checkout/manager.py
```diff
@@ -89,7 +89,9 @@ def checkout(self):
         self.annotations[CHECKIN_CHECKOUT_ANNOTATIONS_KEY] = user_id
 
         # finally, reindex the object
-        self.context.reindexObject()
+        self.context.reindexObject(idxs=(
+            'checked_out',
+            ))
```
jone: Should we update metadata here?

I'd prefer to have an explicit `update_metadata=True or False` here in order to indicate that we thought about it and decided for one way or the other.

Rotonen: AFAICS the metadata edits are handled by a form so I'll just amend an explicit False here.

Rotonen: @jone `object.reindexObject(idxs=[])` does not have a keyword argument for updating metadata - should I also port this to use `portal_catalog.reindexObject(obj, idxs=[], update_metadata=False`?

jone: @Rotonen yes, makes sense 👍

Rotonen: @jone actually we do need to update the metadata as per other tests failing when this is not done and this is thus apparently always implicitly done when one uses `object.reindexObject()`.

jone: Ah, there is a `checked_out` metadata. I pitty that it is not possible to only update selected metadata..

phgross:
@Rotonen still WIP or ready to merge ?

Rotonen:
@jone did I miss anything here? I think it's better than it was, but I'm not sure if this actually closes #3124 .

jone:
Looks good 👍 it solves what's described in #3124.
A changelog entry would be nice.

Rotonen:
@jone ready for rereview.

phgross:
@Rotonen please rebase ...

Rotonen:
@phgross rebased.

PR #3312, opened 2017-08-24
Word meeting: Show decision number in meeting view.

jone:
When agenda items are decided, the decision number should be shown in the meeting view, at the right side of the agenda item title.

bildschirmfoto 2017-08-24 um 09 55 07
bildschirmfoto 2017-08-24 um 09 55 00

Part of #2829 
Requires https://github.com/4teamwork/plonetheme.teamraum/pull/575

### DIFF
--- a/opengever/meeting/browser/meetings/agendaitem.py
+++ b/opengever/meeting/browser/meetings/agendaitem.py
@@ -175,6 +175,7 @@ class AgendaItemsView(BrowserView):
                 view='agenda_items/{}/delete'.format(item.agenda_item_id))
             data['edit_link'] = meeting.get_url(
                 view='agenda_items/{}/edit'.format(item.agenda_item_id))
+            data['decision_number'] = item.decision_number
             if item.is_decide_possible():
                 data['decide_link'] = meeting.get_url(
                     view='agenda_items/{}/decide'.format(item.agenda_item_id))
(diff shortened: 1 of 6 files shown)


sources.cfg
```diff
@@ -26,4 +27,4 @@ auto-checkout = ${buildout:development-packages}
 [branches]
 Products.LDAPUserFolder = ftw
 Products.LDAPMultiPlugins = ftw
-plonetheme.teamraum = jone-meeting-word-view-2
+plonetheme.teamraum = jone-meeting-word-show-decision-number
```
jone: 🚧 https://github.com/4teamwork/plonetheme.teamraum/pull/575

opengever/meeting/browser/meetings/agendaitem.py
```diff
@@ -197,6 +197,9 @@ def _get_agenda_items(self):
                         view='agenda_items/{}/generate_excerpt'.format(
                             item.agenda_item_id))
 
+            if is_word_meeting_implementation_enabled():
```
deiferni: can't you add this  in any case? it should not matter but just not be displayed.

jone: Sure 👍

jone: Updated.

PR #3323, opened 2017-08-28
Fix textfilter for sql table listings when using oracle backends.

phgross:
Currently all expressions contain a string cast statement, even if the field is already a string column. This is unnecessary and leads to an issue with oracle backends (see traceback below). The new
implementation ads casting only for not stringy fields.

`DatabaseError: (cx_Oracle.DatabaseError) ORA-00906: missing left parenthesis`

Fixes https://sentry.4teamwork.ch/sentry/onegov-gever/issues/3749/

### DIFF
--- a/opengever/tabbedview/sqlsource.py
+++ b/opengever/tabbedview/sqlsource.py
@@ -89,9 +89,12 @@ class SqlTableSource(GeverTableSource):
                 # Issue #759
                 query.session
 
-                query = query.filter(or_(
-                    *[cast(field, String).ilike(term)
-                      for field in self.searchable_columns]))
+                expressions = []
+                for field in self.searchable_columns:
+                    if not issubclass(field.type.python_type, basestring):
+                        field = cast(field, String)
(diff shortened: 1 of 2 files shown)


opengever/tabbedview/sqlsource.py
```diff
@@ -89,9 +89,12 @@ def extend_query_with_textfilter(self, query, text):
                 # Issue #759
                 query.session
 
-                query = query.filter(or_(
-                    *[cast(field, String).ilike(term)
-                      for field in self.searchable_columns]))
+                expressions = []
+                for field in self.searchable_columns:
+                    if not issubclass(field.type.python_type, basestring):
```
deiferni: so this won't cast for oracle backends in that case?

phgross: It only cast when it's necessary (for not stringy types), but that handles Oracle correctly.

deiferni: ok :shipit: !

PR #3329, opened 2017-08-30
Hide contact field in the manual journal entry form, 

phgross:
when contact feature is disabled.

Because the implemented feature to add a reference to contacts in the manual entry form only works for the "new contact type", we decided to hide the contact field for installations with a disabled feature flag.


phgross:
@deiferni it's handled correctly

PR #3332, opened 2017-08-30
Fix Drag'n'Drop replacing for documents inside a resolved task.

phgross:
The former implementation also calls the baseclass checks, but those
are wrong for the drag'n'drop replacing use case. A check for `add
portal content` permission makes no sense.

### DIFF
--- a/opengever/document/tests/test_tabbed.py
+++ b/opengever/document/tests/test_tabbed.py
@@ -2,6 +2,7 @@ from ftw.builder import Builder
 from ftw.builder import create
 from ftw.testbrowser import browsing
 from opengever.testing import FunctionalTestCase
+from plone.app.testing import TEST_USER_ID
 from plone.locking.interfaces import IRefreshableLockable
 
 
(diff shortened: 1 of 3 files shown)


opengever/document/tests/test_tabbed.py
```diff
@@ -35,4 +36,25 @@ def test_upload_box_is_shown_when_document_is_checked_out_and_not_locked(self, b
                           .checked_out())
 
         browser.login().open(document)
-        self.assertIsNotNone(browser.css('#uploadbox'))
+        self.assertNotEquals(0, len(browser.css('#uploadbox')),
```
deiferni: use `assertGreater` here

opengever/document/tests/test_tabbed.py
```diff
@@ -35,4 +36,25 @@ def test_upload_box_is_shown_when_document_is_checked_out_and_not_locked(self, b
...
+                             'Uploadbox is not displayed, but should.')
+
+    @browsing
+    def test_upload_box_is_also_shown_in_a_resolved_task(self, browser):
```
deiferni: could you implement this as an integration test? i'd rather not add new functional tests to help us migrate towards integration tests quicker.

phgross: @deiferni I've convert the tests to an integration test.

opengever/document/tests/test_tabbed.py
```diff
@@ -1,38 +1,59 @@
...
-                          .checked_out())
+        self.login(self.regular_user, browser)
+        browser.open(self.document)
+
+        manager = queryMultiAdapter(
+            (self.document, self.request), ICheckinCheckoutManager)
+        manager.checkout()
+
+        browser.open(self.document)
+        self.assertNotEquals(0, len(browser.css('#uploadbox')),
```
deiferni: use `assertGreater` here

opengever/document/tests/test_tabbed.py
```diff
@@ -1,38 +1,59 @@
...
+
+        manager = queryMultiAdapter(
+            (self.taskdocument, self.request), ICheckinCheckoutManager)
+        manager.checkout()
+
+        browser.open(self.taskdocument)
 
-        browser.login().open(document)
-        self.assertIsNotNone(browser.css('#uploadbox'))
+        self.assertNotEquals(0, len(browser.css('#uploadbox')),
```
deiferni: same here, `assertGreater` would be better

phgross:
@deiferni ah missed that, updated ...

PR #3336, opened 2017-08-31
Word-meeting: Create document from configured template for ad-hoc agenda items.

deiferni:
This PR adds support for ad-hoc agenda items for the word-meeting feature. The ad-hoc agenda-items have the same user-interface as proposal base agenda items:

screen shot 2017-09-04 at 13 54 33

They require an ad-hoc proposal template to be configured on the committee or committee-
container.

screen shot 2017-09-04 at 13 57 17

When the user creates an ad-hoc agenda item the proposal template is copied into the meeting dossier and the copy is linked to the agenda-item. That file can then be edited the same way as the document of a proposal. Currently when the agenda-item is deleted the document will remain in the meeting dossier.

2do:
- [x] configure an ad-hoc template on committee-container
- [x] configure an (optional) ad-hoc template on committee
- [x] create a copy of that template into the meeting-dossier and reference from agenda-item
- [x] make quick checkout available
- [x] put into protocol
- ~~[ ] delete from dossier when agenda-item is deleted?~~

### DIFF
--- a/opengever/meeting/browser/meetings/agendaitem.py
+++ b/opengever/meeting/browser/meetings/agendaitem.py
@@ -10,6 +10,7 @@ from opengever.meeting.service import meeting_service
 from plone import api
 from plone.app.contentlisting.interfaces import IContentListing
 from plone.app.contentlisting.interfaces import IContentListingObject
+from Products.CMFPlone.utils import safe_unicode
 from Products.Five.browser import BrowserView
 from zExceptions import Forbidden
 from zExceptions import NotFound
@@ -187,33 +188,24 @@ class AgendaItemsView(BrowserView):
                     view='agenda_items/{}/revise'.format(item.agenda_item_id))
 
             if is_word_meeting_implementation_enabled():
-                document = None
-                if item.proposal:
-                    proposal = item.proposal.submitted_oguid.resolve_object()
-                    document = proposal.get_proposal_document()
-                elif item.has_document:
-                    document = item.resolve_document()
-
+                document = item.resolve_document()
                 if document:
                     data['document_link'] = (
                         IContentListingObject(document).render_link())
                     data.update(self._get_edit_document_options(
                         document, item, meeting))
 
-                data['excerpts'] = self._serialize_excerpts(item)
-                if self.can_generate_excerpt(item):
-                    data['generate_excerpt_link'] = meeting.get_url(
-                        view='agenda_items/{}/generate_excerpt'.format(
-                            item.agenda_item_id))
+                if item.has_proposal:
+                    data['excerpts'] = self._serialize_excerpts(item)
+                    if self.can_generate_excerpt(item):
+                        data['generate_excerpt_link'] = meeting.get_url(
+                            view='agenda_items/{}/generate_excerpt'.format(
+                                item.agenda_item_id))
 
             agenda_items.append(data)
         return agenda_items
 
     def has_documents(self, item):
-        if is_word_meeting_implementation_enabled():
-            if item.has_document:
-                return True
-
         if not item.has_proposal:
             return False
 
@@ -407,7 +399,7 @@ class AgendaItemsView(BrowserView):
         meeting.
         """
         self.check_editable()
-        title = self.request.get('title')
+        title = safe_unicode(self.request.get('title'))
         if not title:
             return JSONResponse(self.request).error(
                     _('empty_proposal', default=u"Proposal must not be empty.")
--- a/opengever/meeting/browser/meetings/agendaitem.py
+++ b/opengever/meeting/browser/meetings/agendaitem.py
@@ -162,7 +162,7 @@ class AgendaItemsView(BrowserView):
         return {
             'document_checked_out': bool(
                 checkout_manager.get_checked_out_by()),
-            'edit_proposal_document_button': button}
+            'edit_document_button': button}
 
     def _get_agenda_items(self):
         meeting = self.context.model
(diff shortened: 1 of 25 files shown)


opengever/meeting/browser/meetings/agendaitem.py
```diff
@@ -187,33 +188,24 @@ def _get_agenda_items(self):
...
                         IContentListingObject(document).render_link())
                     data.update(self._get_edit_document_options(
                         document, item, meeting))
 
-                data['excerpts'] = self._serialize_excerpts(item)
-                if self.can_generate_excerpt(item):
-                    data['generate_excerpt_link'] = meeting.get_url(
-                        view='agenda_items/{}/generate_excerpt'.format(
-                            item.agenda_item_id))
+                if item.has_proposal:
```
phgross: Whats the reason for this if. An Ad hoc agendaitem should/could also have excerpts or not ?

opengever/meeting/locales/de/LC_MESSAGES/opengever.meeting.po
```diff
@@ -666,30 +665,36 @@ msgid "help_select_dossier"
...
-#: ./opengever/meeting/model/committee.py:31
+#: ./opengever/meeting/model/committee.py:30
 msgid "inactive"
 msgstr "Inaktiv"
 
+#. Default: "Ad hoc agenda item template"
+#: ./opengever/meeting/committee.py:94
+#: ./opengever/meeting/committeecontainer.py:44
+msgid "label_ad_hoc_template"
+msgstr ""
```
phgross: Missing.

opengever/meeting/locales/de/LC_MESSAGES/opengever.meeting.po
```diff
@@ -1229,6 +1235,11 @@ msgstr "Die Änderungen konnten nicht gespeichert werden, das Protokoll wurde vo
 msgid "message_write_conflict"
 msgstr "Die Änderungen konnten nicht gespeichert werden, das Protokoll wurde in der Zwischenzeit modifiziert."
 
+#. Default: "No ad-hoc agenda-item template has been configured."
+#: ./opengever/meeting/browser/meetings/agendaitem.py:417
+msgid "missing_ad_hoc_template"
+msgstr ""
```
phgross: Missing.

opengever/meeting/model/agendaitem.py
```diff
@@ -265,6 +265,20 @@ def has_proposal(self):
         return self.proposal is not None
 
     @property
+    def has_document(self):
+        return self.document_int_id is not None
```
phgross: The fact that an agendaitem of a proposal returns False for `has_document` is wrong or at least confusing. Rename to `has_ad_hoc_document`? And also the model should be renamed  `ad_hoc_document` instead of `document`.

opengever/meeting/command.py
```diff
@@ -317,17 +317,16 @@ def generate_file_data(self):
             builder = DocumentBuilder(Document(master_path))
 
             for index, item in enumerate(self.meeting.agenda_items):
-                if not item.has_proposal:
+                if not item.has_document:
```
phgross: This handling is duplicated in many places. IMO it would make sense to move this handling to the agenda_item model. A method get_document which returns the ad_hoc_document or the proposal document.

opengever/meeting/model/meeting.py
```diff
@@ -344,6 +348,35 @@ def schedule_text(self, title, is_paragraph=False):
...
+        ad_hoc_template = committee.get_ad_hoc_template()
+        if not ad_hoc_template:
+            raise MissingAdHocTemplate
+
+        meeting_dossier = self.get_dossier()
+        if not api.user.get_current().checkPermission(
+            'opengever.document: Add document', meeting_dossier):
+            raise MissingMeetingDossierPermissions
+
+        document_title = _(u'title_ad_hoc_document',
```
phgross: Why should the document title be prefixed with `Ad Hoc Traktandum`? 

As a gever user I don't care if the document has been created as an ad hoc agendaitem or with an proposal.
`Genehmigung Protokoll vom 23. September 2017` has more value than `Ad Hoc Traktandum: Genehmigung Protokoll vom 23. September 2017`

opengever/meeting/browser/meetings/agendaitem.py
deiferni: At the moment, no (for the word implementation). since also apparently there can't be an official "decision" for such an item. it is an open question though and unclear to me, see https://github.com/4teamwork/opengever.core/issues/2829.
If we want excerpts they will come later in a separate PR.

opengever/meeting/locales/de/LC_MESSAGES/opengever.meeting.po
deiferni: this has been added in a later commit 😉

opengever/meeting/locales/de/LC_MESSAGES/opengever.meeting.po
deiferni: same here.

opengever/meeting/command.py
deiferni: i completely agree, the responsibilities are a mess at the moment. it is not only the document method but also other agenda-item list methods that violate encapsulation and directly access `agenda_item.proposal` or other attributes. it is already an item in https://github.com/4teamwork/opengever.core/issues/2829, if you don't mind i'd like to tackle this in separate refactoring PRs to fix the issues throughout the whole codebase and not only in the files touched in this PR.

opengever/meeting/model/meeting.py
deiferni: That behavior is consistent with prefixing the proposal documents "Antragsdokument". IMHO both of them should go away. I've added an item to the to-do list https://github.com/4teamwork/opengever.core/issues/2829.

opengever/meeting/model/agendaitem.py
deiferni: this has already been rewritten in a later commit 😉 . an agenda-item has a document when
- there is a proposal
- an ad-hoc-document is attached to the agenda-item

also i'll rename the attribute to `ad_hoc_document` to clarify, as suggested 👍

deiferni:
@phgross changed as requested/commented, please have another look

opengever/meeting/browser/meetings/agendaitem.py
phgross: Ok.

PR #3337, opened 2017-08-31
Add configuration option to hide the mail preview tab.

phgross:
Because the bumblebee-preview and the preview tab, doesn't show the same results in some cases, some customers want to disable the preview `tab`. This is now possible with registry setting. But by default the preview tab is still visible.

Besides that:
 - Replaced tabbedview FTI actions with a MailTabbedview
 - Convert TestPreviewTab tests to integration tests

### DIFF
--- a/opengever/mail/interfaces.py
+++ b/opengever/mail/interfaces.py
@@ -36,3 +36,11 @@ class IAttachmentsDeletedEvent(IObjectModifiedEvent):
     """
 
     attachments = Attribute("List of attachments that have been removed")
+
+
+class IMailTabbedviewSettings(Interface):
+
+    preview_tab_visible = schema.Bool(
+        title=u'Is the preview tab in the mail tabbedview visible',
+        default=True,
+    )
--- a/opengever/mail/tests/test_mail_previewtab.py
+++ b/opengever/mail/tests/test_mail_previewtab.py
@@ -1,17 +1,17 @@
-from ftw.builder import Builder
-from ftw.builder import create
 from ftw.testbrowser import browsing
 from opengever.mail.tests import MAIL_DATA
-from opengever.testing import FunctionalTestCase
+from opengever.testing import IntegrationTestCase
 from pkg_resources import resource_string
 
 
-class TestPreview(FunctionalTestCase):
+class TestPreviewTab(IntegrationTestCase):
 
     @browsing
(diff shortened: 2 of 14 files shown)


opengever/mail/interfaces.py
```diff
@@ -36,3 +36,11 @@ class IAttachmentsDeletedEvent(IObjectModifiedEvent):
     """
 
     attachments = Attribute("List of attachments that have been removed")
+
+
+class IMailTabbedviewConf(Interface):
```
deiferni: we ususally use `Settings` in such a name, so i'd prefer `IMailTabbedviewSettings`

opengever/mail/tests/test_mail_previewtab.py
```diff
@@ -31,3 +33,21 @@ def test_preview_tab_can_handle_attachment_with_wrong_mimetype(self, browser):
...
+    @browsing
+    def test_preview_tab_can_be_disabled_by_registry_flag(self, browser):
+        self.login(self.regular_user, browser)
+        browser.open(self.mail, view='tabbed_view')
+
+        self.assertEquals(
+            ['Overview', 'Preview', 'Journal', 'Sharing'],
+            browser.css('.formTab').text)
+
+        api.portal.set_registry_record(
```
deiferni: this may be registered as a feature in `IntegrationTestCase` and then toggled with `self.activate_feature`

phgross: hmm 🤔  the implemented registry entry is more just a setting than a feature flag, it should only enable/disable the preview tab, no other thing should be implemented according to this option. So this test will be the only one which it should be necessary to enable or disable the setting. 

Therefore I would not register this as a feature, @deiferni what you think?

deiferni: it is a flag that is set based on what the customer wants, which then enables/disables stuff. i would still call that a feature 😉  even though it just toggles one tab for one content-type.

the advantage of registering it as a feature would be to increase its visibility to other developers and also consistency in how to handle features during tests.

deiferni: well ok after having a thorough look here it may be better to not make it a feature after all. 👍

phgross:
@deiferni renamed the setting interface ...

Issue #3224, opened 2017-08-04
is_subdossier index is not updated when moving a subdossier to a repositoryfolder


phgross:
Steps to reproduce:
 - Create a dossier with a subdossier
 - Move  the subdossier to a repositoryfolder 
 - The subdossier, which is no longer a subdossier, is not visible, because the `is_subdossier` index is not updated.

PR #3348, opened 2017-09-04
Replace is_subdossier FieldIndex with an BooleanIndex.

### LINKED ISSUE #3345: Replace and reindex is_subdossier index
As a follow up of this fix https://github.com/4teamwork/opengever.core/pull/3235, the `is_subdossier` index should also be reindexed by an upgradestep.

Because we have to recalculate the index anyways it would be the right time to also change the index type from a `FieldIndex` to a `BooleanIndex`

phgross:
Closes #3345

The upgradestep also reindexes the value for all dossiers, which fixes some dossiers with a potential wrong value because of the bug #3224.

### DIFF
--- a/opengever/core/upgrades/20170904162218_replace_is_subdossier_field_index_with_a_boolean_index/upgrade.py
+++ b/opengever/core/upgrades/20170904162218_replace_is_subdossier_field_index_with_a_boolean_index/upgrade.py
@@ -1,0 +1,21 @@
+from ftw.upgrade import UpgradeStep
+from opengever.dossier.behaviors.dossier import IDossierMarker
+
+
+INDEX_NAME = 'is_subdossier'
+
+
+class ReplaceIsSubdossierFieldIndexWithABooleanIndex(UpgradeStep):
+    """Replace is_subdossier field index with a boolean index.
+    """
+
+    def __call__(self):
+        self.catalog_remove_index(INDEX_NAME)
+        self.catalog_add_index(INDEX_NAME, 'BooleanIndex')
+        catalog = self.getToolByName('portal_catalog')
+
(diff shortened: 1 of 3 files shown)


opengever/core/upgrades/20170904162218_replace_is_subdossier_field_index_with_a_boolean_index/upgrade.py
```diff
@@ -0,0 +1,19 @@
...
+
+INDEX_NAME = 'is_subdossier'
+
+
+class ReplaceIsSubdossierFieldIndexWithABooleanIndex(UpgradeStep):
+    """Replace is_subdossier field index with a boolean index.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
```
lukasgraf: This upgrade step doesn't have an upgrade profile?

opengever/core/upgrades/20170904162218_replace_is_subdossier_field_index_with_a_boolean_index/upgrade.py
```diff
@@ -0,0 +1,19 @@
...
+        self.catalog_remove_index(INDEX_NAME)
+        self.catalog_add_index(INDEX_NAME, 'BooleanIndex')
+
+        for obj in self.objects({'object_provides': IDossierMarker.__identifier__},
+                                'Reindex is_subdossier index'):
+            obj.reindexObject(idxs=[INDEX_NAME])
```
lukasgraf: Should be `catalog.reindexObject(obj, idxs=[INDEX_NAME], update_metadata=False)`

phgross:
@deiferni @lukasgraf updated ..

PR #3357, opened 2017-09-07
Fix a potential issue with excerpts.

deiferni:
🚧  ⚠️  Excerpts were not stored in a PersistenList, also there were reported but not reproducable issue of excerpts showing up/not showing up unexceptedly in a proposal.

_Not sure if this is necessary though, the list items are instances of `RelationValue`._


deiferni:
⚠️ 🚧  hold please, i think this PR should not be merged as is - i just discovered the `excerpts`are part of the `ISubmittedProposal` schema. We should access them in a different manner though. I'll amend.

deiferni:
The `RelationList` is initialised `ISubmittedProposal.excerpts` in the functionality we attempt to change here. I think it is being done correctly and should not be the source of aforementioned issues.

PR #3364, opened 2017-09-11
Format line breaks and links in task responses.

### LINKED ISSUE #3363: Linebreaks and URLs not well displayed in task responses
Since #3172 the task description has been transformed with the webintelligent transform, but the responses are still not transformed 😞 . 

Bsp. https://dev.onegovgever.ch/fd/ordnungssystem/fuehrung/gemeinderecht/uebergeordnete-erlasse/dossier-426/task-101#overview

phgross:
The responses are now displayed in the same manner than task descriptions (see #3172)

Closes #3363

bildschirmfoto 2017-09-11 um 16 47 48
bildschirmfoto 2017-09-11 um 16 47 34

### DIFF
--- a/opengever/task/viewlets/response_templates/responseview.pt
+++ b/opengever/task/viewlets/response_templates/responseview.pt
@@ -29,7 +29,7 @@
               Response Header
             </h3>
 
-            <div class="text" tal:content="response/text"></div>
+            <div class="text" tal:content="structure response_info/text"></div>
 
           </div>
         </div>
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 ---------------------
 
 - Fixed custom sort for catalog listings. [phgross]
+- Format line breaks and links in task responses. [phgross]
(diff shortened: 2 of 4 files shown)


opengever/task/viewlets/response_templates/responseview.pt
```diff
@@ -29,7 +29,7 @@
               Response Header
             </h3>
 
-            <div class="text" tal:content="response/text"></div>
+            <div class="text" tal:content="structure response_info/text"></div>
```
lukasgraf: Could you maybe add a test for XSS? (Like `<script>` tags should be filtered out). This is prone to XSS, and while I believe the `x-web-intelligent` transform does the right thing, we might change in the future how the `response_info` text is preprocessed.

phgross: Should be part of the `x-web-intelligent` test IMO. I've doublechecked but could not find any tests there. I'll add one here ...

phgross: @lukasgraf updated ...

PR #3367, opened 2017-09-12
Fix shared default value for excerpt lists.

deiferni:
This PR fixes an issue with mutable default arguments in combination with a `RelationList`.

When creating a new excerpt for a submitted proposal and setting the excerpt list attribute/field on the submitted proposal the default value from the `ISubmittedProposal` schema was used instead of a new list instance. This happened by using `getattr` which already returns
the schema default. This lead to the default value being shared between multiple instances of submitted proposals.

We attempt to fix this with an upgrade step. Excerpts are stored in the meeting-dossier. We fix shared excerpt lists by creating a new list instance and filtering out the excerpts that come from foreign meeting dossiers should they be present in the excerpt list.

We can't completely fix relations within the same meeting, so it may happen that multiple submitted proposals point to the same excerpt within the same meeting dossier. Since this issue never entered production that should be rare and unproblematic though.

See also #3368 for a proposal/discussion how to address the problem in the future.

### DIFF
--- a/opengever/core/upgrades/20170912122143_fix_shared_excerpt_list_reference/upgrade.py
+++ b/opengever/core/upgrades/20170912122143_fix_shared_excerpt_list_reference/upgrade.py
@@ -1,0 +1,49 @@
+from ftw.upgrade import UpgradeStep
+
+
+class FixSharedExcerptListReference(UpgradeStep):
+    """Fix shared excerpt list reference.
+
+    Excerpts are stored in the meeting-dossier. We fix shared excerpt lists by
+    creating a new list instance and filtering out the excerpts that come from
+    foreign meeting dossiers should they be present in the excerpt list.
+
+    We can't completely fix relations within the same meeting, so it may happen
+    that multiple submitted proposals point to the same excerpt within the same
+    meeting dossier.
+    """
+    def __call__(self):
+        msg = 'Fix shared excerpt list reference.'
+        query = {'portal_type': 'opengever.meeting.submittedproposal'}
+
+        for submitted_proposal in self.objects(query, msg):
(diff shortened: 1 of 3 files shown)


jone:
Should we also set the field default value to `None` in order to avoid this problem in the future by making sure that we cannot append to it?

opengever/core/upgrades/20170912122143_fix_shared_excerpt_list_reference/upgrade.py
```diff
@@ -0,0 +1,49 @@
...
+        # cause unexpected errors we add this safeguard here
+        excerpt_relations = getattr(submitted_proposal, 'excerpts', None)
+        if not excerpt_relations:
+            # empty lists will be correctly intialized/replaced with a new
+            # instance when the first excerpt is created, we don't need to
+            # set the attribute in that case
+            return
+
+        meeting_dossier = self._get_meeting_dossier(submitted_proposal)
+        valid_excerpts = set(meeting_dossier.listFolderContents())
```
jone: Does `listFolderContents` return brains or objects? I'm just not sure 😉

deiferni: objects 😉

deiferni:
@jone afaik the widget trips up if we use `None` as a value since it attempts to iterate unconditionally. Also i can't use the immutable `tuple` due to a type-check in `RelationList` or one of its superclasses. I think we should address the problem and come up with a solution that can be used in all `RelationList`s, see https://github.com/4teamwork/opengever.core/issues/3368.

deiferni:
uhm, so 🚢 ?

PR #3369, opened 2017-09-12
OGGBundle: Fix title encoding in constructor section.

phgross:
Currently the construtor section does not ensure the encoding for the title attribute, which must be unicode. This can lead in to broken `zope.i18nmessageid.message.Message` objects in the journal of an object, which raises an `UnicodeDecodeError` (see https://sentry.4teamwork.ch/sentry/onegov-gever/issues/3941/) when trying to translate the Message by `zope.i18n.translate`.

Therefore the journal of those object can not be rendered or converted in to a PDF when resolving the Dossier.

### DIFF
--- a/opengever/bundle/tests/test_section_constructor.py
+++ b/opengever/bundle/tests/test_section_constructor.py
@@ -92,6 +92,22 @@ class TestConstructor(FunctionalTestCase):
         self.assertFalse(hasattr(content, 'title_de'))
         self.assertFalse(hasattr(content, 'title_Fr'))
 
+    def test_title_is_unicode(self):
+        item = {
+            u"guid": "12345xy",
+            u"_type": u"opengever.document.document",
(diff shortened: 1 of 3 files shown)


opengever/bundle/tests/test_section_constructor.py
```diff
@@ -92,6 +92,21 @@ def test_creates_simple_title_content(self):
...
+            u"title": u'Bewerbung Hanspeter M\xfcller'.encode('utf-8'),
+        }
+
+        section = self.setup_section(previous=[item])
+        list(section)
+
+        portal = api.portal.get()
+        content = portal.restrictedTraverse(item['_path'])
+
+        self.assertTrue(isinstance(content.title, unicode))
```
deiferni: you could also add an `assertEqual` to make sure the title is the correct string

PR #3384, opened 2017-09-18
SPV word: New edit meeting form.

jone:
**Background:**
In the older SPV implementation, the meeting details were edited through the "protocol"-view, which included the meeting details form.
But with the word feature we are no longer using the "protocol"-view because the protocol is now written directly in the word files, triggered through the meeting view.
Request: https://github.com/4teamwork/gever/issues/65

**Changes:**
- The meeting details form is extracted from the "protocol"-view into a separate edit form.
- The "edit"-action is moved from the left sidebar to Plone's editbar (`object`-action), because it is the plan to get rid of the sidebar step by step.

**Implementation details:**
- The meeting-type has no `FTI` => the action can only be registred publicly.
- The published meeting wrapper does not allow unsecure access to attributes, therefore the edit-condition-check is implemented on the view.
- The fixture is updated to define explicit membership periods, so that the members get available in the edit form.
- A `deactivate_feature`-method is added to the test case base class. It makes it easier to test action availability.
- Locking and conflict detection is copied along from the protocol view.
- Because of consistency, the look-and-feel should go towards Plone / rest of GEVER.
- The status button in the sidebar is now visually broken. It will be moved to a separate place and thus fixing is not worth the effort.

bildschirmfoto 2017-09-18 um 11 07 54
bildschirmfoto 2017-09-18 um 11 07 49

### DIFF
--- a/opengever/meeting/browser/meetings/edit_meeting.py
+++ b/opengever/meeting/browser/meetings/edit_meeting.py
@@ -1,0 +1,164 @@
+from calendar import timegm
+from opengever.base import _ as _base
+from opengever.base.browser.modelforms import ModelEditForm
+from opengever.base.date_time import as_utc
+from opengever.meeting import _
+from opengever.meeting import is_word_meeting_implementation_enabled
+from opengever.meeting import require_word_meeting_feature
+from opengever.meeting.browser.meetings.protocol import IMeetingMetadata
+from opengever.meeting.model import Meeting
+from opengever.ogds.base.actor import Actor
+from plone.locking.interfaces import ILockable
+from Products.Five.browser.pagetemplatefile import ViewPageTemplateFile
+from z3c.form import button
+from zExceptions import Redirect
+
+
+class EditMeetingView(ModelEditForm):
+    """The edit meeting view provides a form for editing the meeting details.
+
+    This view is used when the word meeting feature is enabled.
+    The older implementation used the protocol view for editing these fields.
+    This view was copied from the protocol view.
+    """
+
+    ignoreContext = True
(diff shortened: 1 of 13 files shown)


jone:
Rebased because of changelog conflict.

opengever/meeting/browser/meetings/edit_meeting.py
```diff
@@ -0,0 +1,157 @@
...
+from opengever.meeting.browser.meetings.protocol import IMeetingMetadata
+from opengever.meeting.model import Meeting
+from opengever.ogds.base.actor import Actor
+from plone.locking.interfaces import ILockable
+from Products.Five.browser.pagetemplatefile import ViewPageTemplateFile
+from z3c.form import button
+from zExceptions import Redirect
+
+
+class EditMeetingView(ModelEditForm):
```
deiferni: please add a short docstring here mentioning that this is currently a copy of the protocol edit class and why

jone:
Updated according to your suggestions.
(Be aware that I've renamed the object-action according to the renamed view. If you did run the upgrade step from the old implementation you'll end up with a wrong action `edit_meeting` and a correct action `edit-meeting`.)

opengever/meeting/browser/meetings/edit_meeting.py
```diff
@@ -0,0 +1,164 @@
...
+    """The edit meeting view provides a form for editing the meeting details.
+
+    This view is used when the edit-meeting feature is enabled.
```
deiferni: edit-meeting -> word meeting?

jone:
Updated 👍

PR #3389, opened 2017-09-19
Word-Meeting: Remove sidebar from meeting view

jone:
**Goal:**
When the word-meeting feature is enabled, we want to get rid of the sidebar in the meeting view in order to create room for a new navigation.
Requests: https://github.com/4teamwork/gever/issues/94, https://github.com/4teamwork/gever/issues/86

**Changes:**
- Move ZIP-Export action to the actions menu.
- Move meeting status from the sidebar to the meeting details table.
  - Requires CSS changes https://github.com/4teamwork/plonetheme.teamraum/pull/584
- Move workflow transition buttons to the actions menu in order to be consistent with the rest of GEVER.
  - Dialog-boxes are kept since we keep the same HTML ID, onto which the JS click handler is bound.

**Implementation Details:**
- Actions are registered globally since a meeting has no FTI where actions could be registered.
- Workflow transitions are added by extending the already customized workflow menu.

bildschirmfoto 2017-09-18 um 15 24 27
bildschirmfoto 2017-09-18 um 15 51 05
bildschirmfoto 2017-09-19 um 08 40 48
bildschirmfoto 2017-09-19 um 08 41 00


sources.cfg
```diff
@@ -18,3 +20,5 @@ auto-checkout = ${buildout:development-packages}
 [branches]
 Products.LDAPUserFolder = ftw
 Products.LDAPMultiPlugins = ftw
+# https://github.com/4teamwork/plonetheme.teamraum/pull/584
+plonetheme.teamraum = jone-spv-remove-sidebar
```
jone: 🚧 https://github.com/4teamwork/plonetheme.teamraum/pull/584

opengever/core/profiles/default/actions.xml
```diff
@@ -612,6 +612,18 @@
       <property name="visible">True</property>
     </object>
 
+    <object name="export_pdf" meta_type="CMF Action" i18n:domain="opengever.core">
+      <property name="title" i18n:translate="">Export as Zip</property>
+      <property name="description" i18n:translate="" />
+      <property name="url_expr">string:${object_url}/zipexport</property>
+      <property name="icon_expr" />
+      <property name="available_expr">here/@@zipexport/visible_in_actions_menu|nothing</property>
```
deiferni: it would be a time to rename this action to:
- either `zip_export` to cleanly override the default implementation
- or `zip_export_meeting` to indicate that this is a meeting specific implementation

i would probably go for `zip_export_meeting` in this case.

opengever/meeting/browser/meetings/transitions.py
```diff
@@ -25,7 +25,7 @@ def __call__(self):
                 default='Transition ${transition} executed',
                 mapping={'transition': self.model.workflow.transitions.get(transition).title})
 
-        return JSONResponse(self.request).redirect(self.model.get_url()).info(msg).dump();
+        return JSONResponse(self.request).redirect(self.model.get_url()).info(msg).dump()
```
deiferni: looks like someone was in JS-mode 😂

opengever/core/profiles/default/actions.xml
jone: Fair enough.
I've changed it to `export-meeting-zip` because of personal preferences 😏 
This affects the non-word implementation though.

PR #3391, opened 2017-09-19
Register an is_subdossier indexer for dossiertemplates.

phgross:
We querying dossiertemplates with an `is_subdossier=False` query in the dossiertemplate listing, but there is no indexer for dossiertemplates.

With a former change which has changed the `is_subdossier` index from a FieldIndex to a BooleanIndex, this leads to a confusing behavior when switching for the dossiertemplate listing tab to the template document tab and back.

Note: Lukas will take a further look in to that confusing behavior, but for now the additional indexer and reindexing all dossiertemplates "solves" the problem.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2017.5.1 (unreleased)
 ---------------------
 
+- Fix is_subdossier replacement upgradestep, reindex all objects. [phgross]
 - Register an is_subdossier indexer for dossiertemplates. [phgross]
 - Add missing french translations. [phgross]
 
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2017.5.1 (unreleased)
 ---------------------
 
+- Register an is_subdossier indexer for dossiertemplates. [phgross]
 - Add missing french translations. [phgross]
 
 
--- a/opengever/core/upgrades/20170919144003_replace_is_subdossier_field_index_with_a_boolean_index/upgrade.py
+++ b/opengever/core/upgrades/20170919144003_replace_is_subdossier_field_index_with_a_boolean_index/upgrade.py
@@ -12,10 +12,4 @@ class ReplaceIsSubdossierFieldIndexWithABooleanIndex(UpgradeStep):
     def __call__(self):
         self.catalog_remove_index(INDEX_NAME)
         self.catalog_add_index(INDEX_NAME, 'BooleanIndex')
-        catalog = self.getToolByName('portal_catalog')
-
-        for obj in self.objects({'object_provides': IDossierMarker.__identifier__},
-                                'Reindex is_subdossier index'):
-
-            catalog.reindexObject(
-                obj, idxs=[INDEX_NAME], update_metadata=False)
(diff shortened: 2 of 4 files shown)


opengever/core/upgrades/20170919085051_reindex_is_subdossier_for_dossiertemplates/upgrade.py
```diff
@@ -0,0 +1,18 @@
+from opengever.dossier.dossiertemplate.behaviors import IDossierTemplateMarker
+from ftw.upgrade import UpgradeStep
+
+
+class ReindexIsSubdossierForDossiertemplates(UpgradeStep):
+    """Reindex is_subdossier for dossiertemplates.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
```
lukasgraf: ✂️

phgross:
@lukasgraf updated ...

lukasgraf:
I analyzed this problem some more, and I think I found the reason(s) for the buggy behavior:

The `BooleanIndex` uses some highly optimized internal data structures and strategies:

- **Fact**: Its inverted index (`self._index`) contains a set of RIDs that have the value (`True` or `False) that is currently being tracked by the index. The `bool` value that's being tracked by the index may change though: When updating the index, [it checks whether the value being tracked falls into the smaller side of a 60% split](https://github.com/zopefoundation/Products.ZCatalog/blob/3.0.2/src/Products/PluginIndexes/BooleanIndex/BooleanIndex.py#L124-L128), and flips the value being tracked so that it's always the one with the least entries.

-  **Fact**: In its `_apply_index` method, the `BooleanIndex` [makes a distinction](https://github.com/zopefoundation/Products.ZCatalog/blob/3.0.2/src/Products/PluginIndexes/BooleanIndex/BooleanIndex.py#L223-L232) between three different cases:

    - Queried value is equal to the one being currently tracked
    - Queried value is different from the one being tracked, and resultset is non-empty
    - Queried value is different from the one being tracked, and resultset is `None`
  Based on those distinctions, it applies different set operations to different data structures.

- **Guess**: Indexes get passed into their `_apply_index()` method
  - either a non-empty resultset, which they're supposed to remove from (filter out non-matching records)
  - a resultset of `None`, which they're supposed to add to (i.e., construct the initial resultset that's supposed to be filtered down by indexes applied later)

- **Guess**: In the [branch where the `BooleanIndex` consults its `_unindex`](https://github.com/zopefoundation/Products.ZCatalog/blob/3.0.2/src/Products/PluginIndexes/BooleanIndex/BooleanIndex.py#L229), it basically builds a complement of the indexed values it has. E.g. "the records that are `True` must be the ones where I don't have `False` stored". This might give incorrect counts if not all records have an indexed value.

- **Guess**: The switching behavior we're seeing is caused by the query planner getting confused. The query planner might decide to apply the indexes in a different order based on past queries. Based on the above behavior, this might lead to unexpected results that don't provide the optimization the query planner was "hoping" for, leading it to revert to chose a different plan for the next query.

As a consequence, I think for a `BooleanIndex`, all objects must be indexed (whether or not an indexer is registered).

phgross:
I'll create a new PR which solves the #3392 issue completely

PR #3396, opened 2017-09-19
Fix ordering in bumblebee gallery.

jone:
Content listings should be ordered in a way the user can interpret. Currently we order by the catalog's internal values, such as the rid, which is confusing for users and considered a bug.

It's best to order by modified date by now, in order to show the latest changes, so that users can find recently used documents easily.

In the future we will hopefully implement an ordering menu, where the user can decide how to order. But modified date is a good fix for now.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -8,6 +8,7 @@ Changelog
 - SPV word: Move workflow transitions to actions menu in meeting view. [jone]
 - SPV word: Move meeting status in meeting view. [jone]
 - SPV word: Move ZIP-export action to actions menu. [jone]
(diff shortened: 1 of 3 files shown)


jone:
This is more complex: actually the table source will not sort by date indexes. I will provide a better fix.

jone:
This change requires https://github.com/4teamwork/opengever.core/pull/3399 in order to work correctly and the implementation must be changed to update the default sorting of the catalog source config instead of changing the result.

jone:
Becaus #3399 is more complex as I thought, I have updated this change to no longer base on it and implemented a simple, shipable fix for the problem.

We now just add the `sort_on` to the query when it was not provided by the table source.

PR #3399, opened 2017-09-20
🚧 Tabbedview: let the catalog sort date indexes.

jone:
ftw.table does not sort by date indexes in order to circumvent the problem that date indexes with falsy values remove result from the result set.

ftw.table instead sorts the results afterwards, breaking the lazyness and retrieving the brains. This is approach is years old and not a good solution.

The better solution to avoid such problems is to make sure that a date index never contains falsy values. For optional dates this could easily achieved by using an early (1900) or late (2199) default date in the indexer. The catalog index will decide whether to use an early or late default date because it does not have this competence.

**Proposal**
1. Make sure that we dont have falsy values in catalog date indexes.
2. Disable ftw.table's custom sort index workaround.

**Advantags**
1. Loading listings with many results when ordered by date indexes will be much faster since the catalog will do the sorting and we can access only the page while keeping a lazy result set.
2. We have a solution which applies to all catalog queries, not just tables.
3. It's easier to do sorting in other components using the ftw.table sources, such as preview galleries.

**Ensuring non-falsy values**
No action is needed at the moment for ensuring that date index are non-falsy, since they already seem to be correct.
1. A test ensures that we don't have falsy values when the fixture is created. This could protect us a bit from future errors.
2. An [opengever.maintenance](https://github.com/4teamwork/opengever.maintenance/pull/92) script can be executed on productive installations for verifying that the indexes are correct. No bad installation was found up to now.

**Changing ftw.table**
I've decided to not change ftw.table, since we cannot be sure that all other installations using it have correct indexes. They may rely on this behavior. It's a legacy we can't get rid of that easily in all projects.

### DIFF
--- a/opengever/tabbedview/tests/test_table_source_configs.py
+++ b/opengever/tabbedview/tests/test_table_source_configs.py
@@ -1,0 +1,51 @@
+from itertools import chain
+from opengever.tabbedview.interfaces import IGeverCatalogTableSourceConfig
+from opengever.testing import IntegrationTestCase
+from plone.testing.zca import _REGISTRIES
+from Products.CMFCore.utils import getToolByName
+from Products.PluginIndexes.DateIndex.DateIndex import DateIndex
+
+
+class TestTableSourceConfigs(IntegrationTestCase):
+
+    def test_no_date_indexes_with_falsy_values(self):
+        """When sorting a catalog query by a date index containing falsy values,
+        it will remove brains with falsy values from the result set.
+        It cannot do a correct job in this case because the date index does not
+        know whether falsy items need to be inserted at the top or at the bottom.
+
+        The correct approach is to write date indexers which always return a
+        datetime object. It may be in the far-future or in the far-past when the
+        actual value is not set.
+        """
+        catalog = getToolByName(self.portal, 'portal_catalog')
+        indexes_with_falsy_values = []
+
+        for index in catalog._catalog.indexes.values():
+            if not isinstance(index, DateIndex):
+                continue
+            if filter(lambda value: not value, index.uniqueValues()):
+                indexes_with_falsy_values.append(index)
+
+        self.assertEquals(
+            [], indexes_with_falsy_values,
+            'Date indexes are not allowed to contain falsy values.'
+            ' Change the indexer so that always returns a datetime object.')
+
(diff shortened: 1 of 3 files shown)


opengever/tabbedview/tests/test_table_source_configs.py
```diff
@@ -0,0 +1,51 @@
...
+            'Date indexes are not allowed to contain falsy values.'
+            ' Change the indexer so that always returns a datetime object.')
+
+    def test_date_indexes_are_not_custom_sort_indexes(self):
+        """ftw.table's catalog source does not let the catalog sort by date
+        indexes because it may remove results when the index contains falsy
+        values.
+
+        We make sure that we don't have falsy values in our date indexes.
+        Therefore we can remove this protection by.
```
deiferni: ~~by~~

jone:
🚧 There is a problem when sorting by receipt / delivery dates, which are empty 🤔 
An assumption is wrong. Investigating.

jone:
Ok, the point is that date indexes never index falsy values, probably because of its internal data structure. So the assumption that GEVER is already save, is wrong.
On a sample with 315 objects in the catalog I have these stats:
```python
(Pdb) from Products.PluginIndexes.DateIndex.DateIndex import DateIndex
(Pdb) pp {index.__name__: len(index._unindex) for index in catalog._catalog.indexes.values() if isinstance(index, DateIndex)}
{'Date': 315,
 'created': 315,
 'date_of_completion': 1,
 'deadline': 1,
 'delivery_date': 0,
 'document_date': 126,
 'effective': 315,
 'end': 0,
 'expires': 315,
 'modified': 315,
 'receipt_date': 0,
 'retention_expiration': 0,
 'start': 148}
```

Proposed actions:
- Update or remove `src/opengever.maintenance/opengever/maintenance/scripts/check_datetime_indexes_for_falsy_values.py` because of wrong assumptions in docstring and code.
- Change the test to count the unindex length, let it fail.
- Write indexers for all date indexes, which ensure that always a datetime is returned.

Benefits:
We can sort catalog queries by date index without the fear of loosing brains.
This is the only correct fix I know of at this point: datetime indexers must not return falsy values.

jone:
I'm closing this, because it doesn't really solve a problem. See discussion.

PR #3402, opened 2017-09-20
SPV word: Add proposal documents to meeting ZIP export.

jone:
When the word-meeting feature flag is enabled, the ZIP export of a meeting should also contain the proposal documents.

Closes https://github.com/4teamwork/gever/issues/95


jone:
Rebased.

PR #3404, opened 2017-09-21
SPV: Fix dialog text when closing a meeting.

jone:
At any point in the process, the committee responsible must be able to go back so that all documents can be edited.

bildschirmfoto 2017-09-21 um 10 32 38
bildschirmfoto 2017-09-21 um 10 07 47

This is changed in both, the old meeting implementation and the new word implementation.


jone:
rebased 👍

PR #3405, opened 2017-09-21
SPV word: let meeting be reopened.

jone:
**Background:**
Users make errors. Therefore a secretary of a committee / meeting should always be able to fix typos and errors in documents.

**Requirement:**
In order to be able to fix problems in documents, a user must be able to reopen an already closed meeting. A new "reopen" workflow transition of the meeting workflow allows to do so.

**Problems:**
When _closing_ a meeting, it overrides the protocol and removes the lock, allowing the user to edit the document.
By adding a "reopen" transition we create the problem that a user may have made changes to the generated protocol, but by reopening and closing again the protocol would have been overwritten.
This is addressed by 
1. not overriding protocols which are no longer locked
2. not removing the lock when reopening the meeting.

**Screenshots:**
bildschirmfoto 2017-09-21 um 10 06 46
bildschirmfoto 2017-09-21 um 12 07 27
bildschirmfoto 2017-09-21 um 12 07 35

Request: https://github.com/4teamwork/gever/issues/97

### DIFF
--- a/opengever/meeting/browser/meetings/meeting.py
+++ b/opengever/meeting/browser/meetings/meeting.py
@@ -288,6 +288,20 @@ class MeetingView(BrowserView):
         if self.has_protocol_document:
             return self.model.protocol_document.get_download_url()
 
+    def will_closing_regenerate_protocol(self):
+        """This method can tell whether we will regenerate the protocol when
+        closing the meething.
+
+        The protocol will be generated while it is locked or not yet existing.
+        When the user reopens the meeting the document will be left unlocked
+        because the user may have made changes at this point.
+        The protocol is no longer regenerated automatically when it
+        may have been changed.
+        """
+        if not self.has_protocol_document():
+            return True
+        return self.model.protocol_document.is_locked()
+
     def url_download_agendaitem_list(self):
         if self.has_agendaitem_list_document:
             return self.model.agendaitem_list_document.get_download_url()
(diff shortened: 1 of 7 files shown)


opengever/meeting/browser/meetings/meeting.py
```diff
@@ -288,6 +288,20 @@ def url_download_protocol(self):
         if self.has_protocol_document:
             return self.model.protocol_document.get_download_url()
 
+    def will_closing_regenerate_protocol(self):
+        """This method can tell whether we will regenerate the protocol when
+        when closing the meething.
```
phgross: `when when`

opengever/meeting/browser/resources/meeting.js
```diff
@@ -45,6 +54,14 @@
       },
       {
         method: "click",
+        target: "#closed-held",
```
phgross: Why is an JS handler necessary? The action visible in your screenshot is an editbar action, wich works without JS, isn't it?

jone: This is actually executed with JavaScript historically:
- The workflow is not (yet) a Plone workflow since meetings are not Plone objects.
- The transition actions used to be placed in the lefthand sidebar and executed with AJAX, since they may require confirmation through a dialog modal (Are you sure to close the meeting?).
- There are no regular transition endpoints (e.g. doing redirects on success), but only AJAX endpoints returning JSON.
- I moved the actions to the actions menu in #3389 (custom menu adapter).

I would like to have a meeting Plone object (or meeting dossier) with a regular lawgiver workflow at some point, which will solve that. Until then we do it in JavaScript.

phgross: ok

PR #3408, opened 2017-09-21
Fix caching problems in JSON responses with IE 11.

jone:
The JSONResponse component is used for generating standardized JSON responses of read-only (GET) as well as writing (POST) requests.

The responses were not equipped with caching headers (such as cache-control). If a response does not have a cache-control header, some browsers, such as IE 11, may cache the response, at least of GET requests.

This component is primarely used for AJAX requests. We usually want to avoid caching AJAX responses as their content may change over time.
We especially do not want to let the browser be in control of cache invalidation.

Therefore we use no-cache headers on all AJAX responses by default.

Implementation details:
- `Cache-Control: no-store`: Neither browser nor intermediate components are allowed to keep a copy in the cache.
- `Pragma: no-cache`: No-cache command for older browser / proxies.
- `Expires: 0`: Indicates that the content has already expired and may have changed.

Bug-Report: https://github.com/4teamwork/gever/issues/61

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -19,6 +19,7 @@ Changelog
 - SPV word: Add proposal documents to meeting ZIP export. [jone]
 - SPV word: Prefix decision numbers with year. [jone]
 - SPV: Fix dialog text when closing a meeting. [jone]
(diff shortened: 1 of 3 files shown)


jone:
Updated 👍

PR #3428, opened 2017-09-26
Improve remove / delete labels for agenda items

tarnap:
There are several types of agenda items, amongst them:
- ad hoc agenda items
- proposed agenda items

Ad hoc agenda items should be deleted.
Proposed agenda items should only be removed from schedule.

rename_detraktandieren


tarnap:
I bi bir screen-capture nid igschlafe, nur churz abglänkt worde ;)

tarnap:
@jone history entry added

PR #3429, opened 2017-09-26
Trash document of deleted ad hoc agenda items

tarnap:
When a responsible of a committee creates an agenda item ad hoc, a
document is created to contain the ad hoc item's proposal / decision.
If the ad hoc item is removed, the related document should be trashed.

If the agenda item is created from a proposal (handed in by a proposer),
then the proposal should remain until a related agenda item has been
scheduled.
Therefore removing the related agenda item should keep the document
untrashed.

Resolves https://github.com/4teamwork/gever/issues/76

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -7,6 +7,7 @@ Changelog
 
 - Fix private folder creation for userids containing dashes. [phgross]
 - SPV word: Set "remove" and "delete" translations correctly. [tarnap]
+- SPV word: trash documents of removed ad hoc agenda items. [tarnap]
 - SPV word: Represent paragraphs in the generated protocol. [jone]
 - SPV word: Allow to set custom excerpt titles. [tarnap]
 - OGDS update: Truncate purely descriptive user fields. [lgraf]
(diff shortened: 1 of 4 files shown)


lukasgraf:
> *No description provided.*

spock

tarnap:
There's no review request either ;)

opengever/meeting/tests/test_agendaitem.py
```diff
@@ -231,6 +232,18 @@ def test_delete_agenda_item(self, browser):
                           browser.json.get('messages'))
 
     @browsing
+    def test_delete_agenda_item_does_not_trash_proposal(self, browser):
+        item = create(Builder('agenda_item').having(
+            title=u'foo', meeting=self.meeting))
```
jone: You could move this test to `test_agendaitem_word.py`, because:
- it actually is a word-feature, since the non-word implementation does not have a document to trash
- you could use the new integration testing there
- you could use the fixture (`self.schedule_proposal(self.word_proposal)`)

tarnap: :+1: will do!

tarnap:
@jone: the tests passed now :)

PR #3437, opened 2017-09-27
Remove excerpt template from forms when word feature enabled

tarnap:
The excerpt template is not required when the word feature is enabled.
This also improves some existing tests.

Resolves https://github.com/4teamwork/gever/issues/102

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 ---------------------
 
 - SPV word: Remove proposal document's title prefix. [tarnap]
+- SPV word: Hide excerpt template from form when word feature enabled. [tarnap]
 - SPV word: Remove "Ad-hoc" from the GUI. [tarnap]
 - Fix private folder creation for userids containing dashes. [phgross]
 - SPV word: Set "remove" and "delete" translations correctly. [tarnap]
--- a/opengever/meeting/committeecontainer.py
+++ b/opengever/meeting/committeecontainer.py
@@ -1,5 +1,6 @@
 from opengever.base.behaviors.translated_title import TranslatedTitleMixin
 from opengever.meeting import _
+from opengever.meeting import is_word_meeting_implementation_enabled
 from opengever.meeting import require_word_meeting_feature
 from opengever.meeting.model import Member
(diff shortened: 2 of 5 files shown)


docs/HISTORY.txt
```diff
@@ -6,6 +6,7 @@ Changelog
 ---------------------
 
 - SPV word: Remove proposal document's title prefix. [tarnap]
+- SPV word: Remove excerpt template frcom form when word feature enabled. [tarnap]
```
jone: `frcom` ✂️

jone: Maybe `Hide` instead of `Remove`?

opengever/meeting/committee.py
```diff
@@ -3,6 +3,7 @@
 from five import grok
 from opengever.base.validators import BaseRepositoryfolderValidator
 from opengever.meeting import _
+from opengever.meeting import is_word_meeting_implementation_enabled
```
jone: ✂️ ?

opengever/meeting/committeecontainer.py
```diff
@@ -8,6 +9,13 @@
 from plone.dexterity.content import Container
 from plone.directives import form
 from z3c.relationfield.schema import RelationChoice
+from zope.interface import Invalid
+
+
+def excerpt_template_constraint(value):
+    if is_word_meeting_implementation_enabled() or bool(value):
```
jone: I'd probably separate the conditions and add a comment, so that the future reader understands it better:
```python
    # The excerpt template is not available when the word-meetin feature enabled,
    # thus the validator should not complain.
    return True
if not value:
    raise Invalid()
return True
```
What do you think?

docs/HISTORY.txt
tarnap: Makes sense

tarnap: Ah nei... Wurstfingeralert!

opengever/meeting/committeecontainer.py
tarnap: Ok, I'll add the comment as well to the commit message.

tarnap:
@jone: made the changes as requested, also improved the title and description of the commit.

PR #3438, opened 2017-09-27
Fix flaky tests

jone:
Fix some flaky tests:

1) Document listing tests (tabbedview): test is rewritten with integration testing and is modified to only assert that one expected item is in the list of documents. In the fixture we actually have many documents, since word-based proposals also contain a document. It is more practical / less brittle here to not assert the complete list but only verify that one document is there.

2) Bumblebee gallery tests: use a ticking creator for moving the time forward so that we dont have two objects created at the same time[1] and thus the ordering is not consistent.

[1] same time means in the same minute, since the catalog's sort-index is minutes based (for better internal tree balancing, I guess).

### DIFF
--- a/opengever/tabbedview/tests/test_bumblebee_gallery.py
+++ b/opengever/tabbedview/tests/test_bumblebee_gallery.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 from ftw.builder import Builder
 from ftw.builder import create
+from ftw.builder import ticking_creator
 from ftw.tabbedview.interfaces import ITabbedView
 from ftw.testbrowser import browsing
 from ftw.testing import freeze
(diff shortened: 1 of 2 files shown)


opengever/tabbedview/tests/test_bumblebee_gallery.py
```diff
@@ -325,35 +326,39 @@ def test_returns_streamed_dict_representations_of_brains(self):
...
+            view = getMultiAdapter(
+                (dossier, self.request), name="tabbedview_view-documents-gallery")
 
-        self.request.set('documentPointer', 1)
+            self.request.set('documentPointer', 1)
 
-        create(Builder('document').within(dossier))
-        document1 = create(Builder('document').within(dossier))
-        document2 = create(Builder('document').within(dossier))
+            create(Builder('document').within(dossier))
```
phgross: this document is not part of the previews? 🤔

phgross: Ah now i got it, I would rename `document1` and `document2` to `document2`, `document3`.

PR #3442, opened 2017-09-27
Translation files: remove lines from comments.

jone:
**Motivation:**
The translation files often have a lot of changes when moving code around.  These changes are very noisy in the diffs.

**Change:**
Let bin/i18n-build remove the line numbers from comments in the generated pot files. These changes are then synced to the po files. This only applies to automatically generated pot files (primary package domain).

### DIFF
--- a/bin/i18n-build
+++ b/bin/i18n-build
@@ -55,6 +55,7 @@ do
     build_cmd="bin/i18ndude rebuild-pot --pot $potfile --create $pkgname $merge_cmd ./${pkg_relpath}"
     echo $build_cmd
     eval $build_cmd
+    sed -i.bak 's/\(#: .*\):.*/\1/' $potfile && rm $potfile.bak
 
     # Sync all lanuguages defined in $LANGUAGES
     for lang in $LANGUAGES
--- a/opengever/activity/locales/de/LC_MESSAGES/opengever.activity.po
+++ b/opengever/activity/locales/de/LC_MESSAGES/opengever.activity.po
@@ -4,7 +4,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
-"POT-Creation-Date: 2017-04-27 11:20+0000\n"
+"POT-Creation-Date: 2017-10-02 12:51+0000\n"
 "PO-Revision-Date: 2015-02-24 08:40+0100\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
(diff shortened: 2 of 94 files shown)


lukasgraf:
@jone that would mean we lose the functionality to link to the source of the message string on Weblate:

Example:

https://translations.onegovgever.ch/translate/onegov-gever/opengever-activity/fr/?sid=f31a2514-a507-11e7-b409-5cf3fc97d25c&offset=1

weblate_msg_source

jone:
@lukasgraf how bad is that? Who uses that?

I've also discussed with @Rotonen : his editor resolves those lines as links. But he said that the line number usually isn't accurate and it is usuall some lines before or after the line in the comment. Therefore came to the conclusion that not having a line number but only the filename is as good as having a wrong line number, and it is not lying.

My problem is not with the file numbers, but with the fact that the numbers are always wrong and if I change translations I have a huge diff and I'm in fact cleaning up after the last editor (who didn't update the line numbers).

If we decide to not remove the line numbers I'd like to discuss options to enforce up-to-date line numbers, so that they are part of each change which affect the line numbers.

Rotonen:
@jone having thought about it a bit more today after telling @tarnap otherwise - with the i18n qa check in place we'll actually over time clean up the translations to the point where these unclean update states are gone and going forwards you'll only have a diff of what you actually touched.

Rotonen:
@jone or `bin/i18n-build` the whole project in this pull request and as per the QA refusing unclean i18n states we'll have everything in sync starting from then?

jone:
@Rotonen QA does not cover line numbers at the moment, only missing German translations. I'm not sure whether and how we want to enforce updating line numbers, as this could get really annoying. 🤔

Rotonen:
The how is easier as I want to bundle the i18n checks into the up and coming qa script anyway. We already have pcregrep as a testing dependency so we can build the required rule for catching the (mis)match.

The if is tricky - I would not mind this happening as the build failure comes quick upon a push and the required fixup commit is both trivial and quick to do on the developer side.

The discussion is more around where do we want to have the noise around this if we keep the line numbers - with the line numbers some part of our current process will be noisy no matter what.

lukasgraf:
@jone as discussed 🗣  over 🍻  - line numbers in Weblate are expendable, a simple reference to the file as a whole should be enough to cover the few cases where this feature is useful.

Regarding enforcement of always-current line numbers: The cost of translation line number noise in diffs **where only Python / Template lines change their position** (but no actual translations are touched) far outweighs the benefits for me.

So 👍  for ditching translation line numbers.

jone:
I have rebased onto master.
Shall we merge that?
Does anybody want to do a review?

bin/i18n-build
```diff
@@ -55,6 +55,7 @@ do
     build_cmd="bin/i18ndude rebuild-pot --pot $potfile --create $pkgname $merge_cmd ./${pkg_relpath}"
     echo $build_cmd
     eval $build_cmd
+    sed -i.bak 's/\(#: .*\):.*/\1/' $potfile && rm $potfile.bak
```
lukasgraf: @jone the `-i.bak .. && rm $potfile.bak` hack is to achieve portability between GNU sed and OS X BSD-style sed I assume?

jone: @lukasgraf yes: `sed -i ...` does not work on OSX.

PR #3448, opened 2017-09-28
Ungrok proposal add form.

jone:
Ungrok proposal add form in order to make subclassing easier.


jone:
@phgross could you take a look at this one?

PR #3452, opened 2017-09-29
Update plone.restapi to 1.0a21

buchi:
Version 1.0a21 provides support for locking.


Rotonen:
What should we do with this for OfficeConnector payloads? Add (un)lock URL data and if that is present it'll use the API instead of WebDAV?

buchi:
In OC we should switch to REST API endpoints for locking/unlocking and checkout/checkin.
However this requires the `plone.restapi: Use REST API` permission which is currently not granted to all GEVER users. In the longer term every GEVER user will need that permission if we start building our frontend on top of plone.restapi.

PR #3454, opened 2017-09-29
Display other participants

tarnap:
Displays "other participants" in the meeting view (if any available).

Adding some "other participants":
weitere_teilnehmer_1

Displaying "other participants":
weitere_teilnehmer_without_labels

Resolves https://github.com/4teamwork/gever/issues/85

### DIFF
--- a/opengever/meeting/locales/de/LC_MESSAGES/opengever.meeting.po
+++ b/opengever/meeting/locales/de/LC_MESSAGES/opengever.meeting.po
@@ -1650,7 +1650,7 @@ msgstr "Status:"
 #. Default: "Other participants:"
 #: ./opengever/meeting/browser/meetings/templates/meeting-word.pt:200
 msgid "view_label_other_participants"
-msgstr ""
+msgstr "Weitere Teilnehmende"
 
 #. Default: "Participants:"
 #: ./opengever/meeting/browser/meetings/templates/meeting-word.pt
--- a/opengever/meeting/locales/de/LC_MESSAGES/opengever.meeting.po
+++ b/opengever/meeting/locales/de/LC_MESSAGES/opengever.meeting.po
@@ -1647,6 +1647,11 @@ msgstr "Beginn:"
 msgid "view_label_meeting_status"
 msgstr "Status:"
 
+#. Default: "Other participants:"
+#: ./opengever/meeting/browser/meetings/templates/meeting-word.pt:200
+msgid "view_label_other_participants"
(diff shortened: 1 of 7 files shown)


jone:
@tarnap can you add a screenshot so that I can see how it looks? 😉

tarnap:
@jone: screenshots added :)

opengever/meeting/model/meeting.py
```diff
@@ -147,6 +147,13 @@ class Meeting(Base, SQLFormSupport):
     excerpt_documents = relationship('GeneratedExcerpt',
                                      secondary=meeting_excerpts,)
 
+    def get_other_participants_list(self):
+        if self.other_participants is not None:
+            return filter(len, map(lambda value: value.strip(),
+                                   self.other_participants.split(',')))
```
jone: Since there is a textarea, I'd have expected that the participants are line separated, not comma separated. This is consistent with some Plone behavior.
I'd therefore change it here to split by `'\n'` instead.

tarnap: :+1: will do!

tarnap:
@jone: other participants are now added by line! I also updated the screenshots :)

opengever/meeting/locales/de/LC_MESSAGES/opengever.meeting.po
```diff
@@ -1647,6 +1647,11 @@ msgstr "Beginn:"
 msgid "view_label_meeting_status"
 msgstr "Status:"
 
+#. Default: "Other participants:"
+#: ./opengever/meeting/browser/meetings/templates/meeting-word.pt:200
+msgid "view_label_other_participants"
+msgstr "Weitere Teilnehmende"
```
jone: Just realized: the other labels end with a colon.
I'm also not yet happy that it breaks the words 🤔 

How about changing that to `Weitere:`? Because we already have `Teilnehmer:` on the line before.
What do you think?

tarnap: In general, I'd replace "Teilnehmer" with "Teilnehmende" because of the gender police.

Social issues

@jone: What if we dropped the label in the view completely and used it only in the edit form?

jone: @tarnap +1 for dropping the label.

tarnap:
@jone: the label was dropped, the other tests adapted.

PR #3461, opened 2017-10-02
SPV: Create successor proposals

jone:
### Goal
The goal is to make it possible to create successor proposal based on decided proposals.
On one hand it should be visible in the proposals that they depend on each other,
on the other hand it should make it easier for the user to reuse / base on a proposal document.

Closes https://github.com/4teamwork/gever/issues/99

### Implementation Details
- A `Create successor proposal` button appears on decided proposals (but not on a submitted proposal).
- The idea is that a business case is in one dossier, therefore all its proposal shall be in the same proposals.
- The button leads to the proposal addform, which has prefille defaults:
  - The fields are filled with the defaults from the predecessor.
  - The predecessor proposal document can be used as template for the successor proposal; a copy is created.
  - The excerpt of the predecessor is proposed to be used as attachment; can be deselected.
  - The original attachments are proposed to be attached as well.

### Screenshots

bildschirmfoto 2017-10-02 um 16 58 54
bildschirmfoto 2017-10-02 um 17 00 01
bildschirmfoto 2017-10-02 um 17 11 08
bildschirmfoto 2017-10-02 um 17 12 20

### DIFF
--- a/opengever/meeting/browser/proposalforms.py
+++ b/opengever/meeting/browser/proposalforms.py
@@ -5,6 +5,7 @@ from opengever.base.widgets import TrixFieldWidget
 from opengever.document.interfaces import ICheckinCheckoutManager
 from opengever.meeting import _
 from opengever.meeting import is_word_meeting_implementation_enabled
+from opengever.meeting import require_word_meeting_feature
 from opengever.meeting.form import ModelProxyAddForm
 from opengever.meeting.form import ModelProxyEditForm
 from opengever.meeting.proposal import IProposal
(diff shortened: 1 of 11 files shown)


opengever/meeting/browser/proposalforms.py
```diff
@@ -140,6 +149,44 @@ def updateWidgets(self):
         if len(ltool.getSupportedLanguages()) <= 1:
             self.widgets['language'].mode = HIDDEN_MODE
 
+    @require_word_meeting_feature
+    def prefillPredecessorDefaults(self):
+        """When we create a successor proposal, the defaults change to
+        be based on the predecessor.
+        Since the relation widgets do not support changing values we must
+        prefill the values in the requ
```
phgross: requ**est**

jone: 👍

PR #3464, opened 2017-10-04
SPV word: Register excerpt relations properly in the relation catalog.

jone:
The relation from the submitted proposal to the excerpt were not registred properly in the relation catalog.

- Make sure that `addRelations` is called, which registers all relations in the relation catalog.
- Do the same for existing submitted proposals in an upgrade step.

The `excerpts` relation list is only available in submitted proposals.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -7,6 +7,7 @@ Changelog
 
 - SPV word: Fix protocol zip export. [tarnap]
 - Bundle import: Fix deactivation of LDAP plugin during import. [lgraf]
+- SPV word: Register excerpt relations properly in the relation catalog. [jone]
 - Update Plone version to 4.3.15. [lgraf]
 - SPV: Fix proposal history. [tarnap]
 - SPV: Fix deleting ad-hoc agenda items. [tarnap]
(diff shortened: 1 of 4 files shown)


opengever/core/upgrades/20171004082226_update_relation_catalog_for_proposal_excerpts/upgrade.py
```diff
@@ -0,0 +1,13 @@
...
+from z3c.relationfield.event import addRelations
+
+
+class UpdateRelationCatalogForProposalExcerpts(UpgradeStep):
+    """Register excerpt relations of submitted proposals in relation catalog.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
+        query = {'object_provides': 'opengever.meeting.proposal.ISubmittedProposal'}
```
phgross: Ist this upgradestep necessary on installation wich has the spv-word featureflag disabled? Otherwise it would make sense to add a check here...

jone: You're right: it's not necessary (but also not harmful).
I've added a condition.

jone:
Rebased and updated.

PR #3465, opened 2017-10-04
SPV word: provide meeting docproperties in proposal document and excerpt.

jone:
When a document belongs to a proposal in any way, meeting docproperties are provided for this document.
Example:
```python
{'ogg.meeting.decision_number': '2016 / 2',
 'ogg.meeting.agenda_item_number': '1.',
 'ogg.meeting.proposal_title': '\xc3\x84nderungen am Personalreglement',
 'ogg.meeting.proposal_state': 'Decided'}
```

**The docproperties are provided for these documents:**

- Proposal documents which are stored within the case dossier.
- Submitted proposal documents which are stored within the committee.
- Excerpt documents which are created in the committed (and stored within the meeting dossier).
- Excerpt documents which were sent back to the proposer and thus stored in the case dossier.

Resolves https://github.com/4teamwork/gever/issues/77

### DIFF
--- a/opengever/document/tests/test_document_word.py
+++ b/opengever/document/tests/test_document_word.py
@@ -1,0 +1,19 @@
+from opengever.testing import IntegrationTestCase
+
+
+class TestDocumentProposal(IntegrationTestCase):
+    features = ('meeting', 'word-meeting')
+
+    def test_get_proposal_returns_None_for_regular_document(self):
+        self.login(self.dossier_responsible)
+        self.assertIsNone(self.document.get_proposal())
+
+    def test_get_proposal_of_proposal_document_in_case_dossier(self):
+        self.login(self.dossier_responsible)
+        proposal = self.word_proposal
+        self.assertEquals(proposal, proposal.get_proposal_document().get_proposal())
+
+    def test_get_proposal_of_proposal_document_in_committee(self):
+        self.login(self.committee_responsible)
+        proposal = self.submitted_word_proposal
+        self.assertEquals(proposal, proposal.get_proposal_document().get_proposal())
--- a/opengever/document/tests/test_document_word.py
+++ b/opengever/document/tests/test_document_word.py
@@ -17,3 +17,29 @@ class TestDocumentProposal(IntegrationTestCase):
         self.login(self.committee_responsible)
         proposal = self.submitted_word_proposal
         self.assertEquals(proposal, proposal.get_proposal_document().get_proposal())
+
+    def test_get_proposal_of_excerpt_document_in_committee(self):
+        self.login(self.committee_responsible)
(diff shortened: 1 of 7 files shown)


jone:
Rebased and ready for review.

opengever/meeting/tests/test_docproperties.py
```diff
@@ -47,13 +50,15 @@ def test_scheduled_proposal_document(self):
         with self.login(self.dossier_responsible):
             self.assertEquals('scheduled', self.word_proposal.get_state().title)
             self.assertEquals(
-                {'ogg.meeting.decision_number': ''},
+                {'ogg.meeting.decision_number': '',
+                 'ogg.meeting.agenda_item_number': '1.'},
```
phgross: Where does the ending point came from? Does it make sense to include it in the docproperties.

opengever/meeting/docprops.py
```diff
@@ -18,6 +19,8 @@ def get_meeting_properties(self):
             'decision_number': '',
             'agenda_item_number': '',
             'proposal_title': proposal.Title(),
+            'proposal_state': translate(proposal.get_state().title,
+                                        context=self.context.REQUEST),
```
phgross: Does `self.context` have `REQUEST` defined in any case?

opengever/document/tests/test_document_word.py
```diff
@@ -17,3 +19,29 @@ def test_get_proposal_of_proposal_document_in_committee(self):
         self.login(self.committee_responsible)
         proposal = self.submitted_word_proposal
         self.assertEquals(proposal, proposal.get_proposal_document().get_proposal())
+
+    def test_get_proposal_of_excerpt_document_in_committe(self):
```
phgross: test_get_proposal_of_excerpt_document_in_committe*e*

opengever/meeting/tests/test_docproperties.py
jone: The agenda item number has this format (as in first agenda item, second agenda item, ..).
I'd leave it as it is for consistency with the rest of the GUI.
bildschirmfoto 2017-10-10 um 09 04 08

phgross: Make sense.

opengever/meeting/docprops.py
jone: I think so: the context is always a `IProposal` dexterity object.
Dexterity objects are acquisition wrapped and I expect that there always is a REQUEST in the acquisition change (actually the it is the topmost wrapper).
A lot would break when the request is missing 😉

opengever/document/tests/test_document_word.py
jone: 👍 thanks

jone:
@phgross thanks for the review.
I've answered your questions, changed the test name and rebased onto master.

PR #3466, opened 2017-10-04
Use new protocol command if word feature enabled

tarnap:
There are two command classes:

`CreateGeneratedDocumentCommand` - used when the word feature is disabled
`MergeDocxProtocolCommand` - used when the word feature is enabled.

This affects the content of the protocol word file when exported in a ZIP.

Resolves: https://github.com/4teamwork/gever/issues/112

### Testing
We don't have (yet) testing infrastructure for exact comparison of the content of word files. Therefore this change cannot be reliably tested at this point.

### Manual testing
When a meeting doesn't yet have a protocol, and I export the ZIP file, the protocol generated in the ZIP should use the new protocol generating mechanism. This can be verified by making sure that the agenda item document / proposal document is merged into the protocol. We didn't do that in the non-word implementation.

Steps in order to test it manually:
- Create a meeting.
- Create an agenda item (proposal or ad-hoc, doesn't matter).
- Edit the proposal document of a the agenda item and add distinctive content.
- The protocol of the meeting must NOT be generated (otherwise the ZIP would just use that and not go to the affected code part).
- Export the ZIP file.
- Ensure that the protocol word file contains the content of the proposal document.

bildschirmfoto 2017-10-04 um 11 57 00
bildschirmfoto 2017-10-04 um 11 57 56

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2017.5.2 (unreleased)
 ---------------------
 
+- SPV word: Fix protocol zip export. [tarnap]
 - Bundle import: Fix deactivation of LDAP plugin during import. [lgraf]
 - Update Plone version to 4.3.15. [lgraf]
 - SPV: Fix proposal history. [tarnap]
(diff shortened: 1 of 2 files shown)


jone:
@tarnap I've updated the PR description with a testing note and screenshots, as discussed.

tarnap:
@jone: Merci :) <3

PR #3476, opened 2017-10-05
Fixed default_value for hidden responsible_client field on tasktemplates

phgross:
Without grokking the subpackage the `plone.directives.form.default_value` decorator is not working. Therefore the creation of a tasktemplate failed with a validation error for a
hidden field (see https://github.com/4teamwork/opengever.ftw/issues/46). We now register the default_value adapter directly without the help of the decorator. (see https://community.plone.org/t/howto-default-value-for-datetime-field-in-plone5-without-using-plone-directives/2014/6 for further information)

Closes https://github.com/4teamwork/opengever.ftw/issues/46

Backport to `2017.5-stable` needed.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -528,6 +528,8 @@ Objects
   - self.templates
     - self.proposal_template
     - self.sablon_template
+    - self.tasktemplatefolder
+      - self.tasktemplate
 
 .. </fixture:objects>
 
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 ---------------------
 
 - Fix default_documents_as_links default value for opengever.mail. [elioschmutz]
+- Fix responsible_client default value for tasktemplates. [phgross]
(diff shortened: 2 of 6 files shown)


elioschmutz:
Why not using the `defaultFactory`-pattern?

```python
responsible_client = schema.Choice(
    title=_(u'label_resonsible_client',
        default=u'Responsible Client'),
    defaultFactory= default_responsible_client,
    ...
    )

def responsible_client_default_value():
    return get_current_org_unit().id()
```

this is the default way of setting a computed field values.

phgross:
@elioschmutz thats right `defaultFactory` is the way to go for general default_values of a type, wich should also work when creating objects via API. But this default should only work on the form. Because the responsible select combines the two values `responsible_client` and `responsible` therefore the `responsible_client` is hidden.
But when creating a tasktemplate via API there should not be a default_value for the `responsible_client` field.

elioschmutz:
ah, ok 👍

PR #3481, opened 2017-10-06
Remove grok from opengever.private

elioschmutz:
related to #3205


opengever/meeting/browser/meetingdossier_forms.py
```diff
@@ -0,0 +1,6 @@
+from opengever.dossier.browser.forms import DossierAddView
+
+
+class MeetingDossierAddView(DossierAddView):
+    """Add view for opengever.private.dossier
```
phgross: opengever.meeting.dossier

elioschmutz:
@phgross 👍

PR #3484, opened 2017-10-10
Do not fill end date when start is selected...

tarnap:
... yet keep the allowed min date and min time.

This also refactors the datetimepicker resource:
The earliest possible end date is set when the widget is shown.

end_time

Resolves https://github.com/4teamwork/gever/issues/91

### DIFF
--- a/opengever/meeting/browser/resources/datetimepicker.js
+++ b/opengever/meeting/browser/resources/datetimepicker.js
@@ -34,15 +34,30 @@
       this.start.val(this.startwidget.str(roundedStartDateTime));
     }
 
+    self.getMinTime = function () {
+      var start = self.startwidget.getCurrentTime();
+      var end = self.endwidget.getCurrentTime();
+      if (end.getDayOfYear() <= start.getDayOfYear()) {
+        return start.format2();
+      } else {
+        return false;
+      }
+    }
+
     // make sure no date is selected before the start of the range by looking
     // up the earliest valid time in the startwidget
     self.end.datetimepicker({
       onShow: function (current_time, input) {
                 this.setOptions({
                   minDate: self.startwidget.getCurrentTime(),
-                  minTime: self.startwidget.getCurrentTime()
+                  minTime: self.getMinTime()
                 });
-              }
+              },
+      onSelectDate: function(current_time, input) {
+                      this.setOptions({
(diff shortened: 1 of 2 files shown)


opengever/meeting/browser/resources/datetimepicker.js
```diff
@@ -29,38 +29,24 @@
...
-        updateSelectableDates();
-      });
+
+    // make sure no date is selected before the start of the range by looking
+    // up the earliest valid time in the startwidget
+    self.end.datetimepicker({
+      onShow: function (current_time, input) {
+                this.setOptions({
+                  minDate: self.startwidget.getCurrentTime(),
+                  minTime: self.startwidget.getCurrentTime()
```
jone: I don't think that this does the right thing. I think `minTime` is not bound to `minDate`.

I have strange effects:
- When the start is "today", the behavior is different than when it's in the future or past.
- When I select "[today] at 22:00" as start, I cannot select "[tomorrow] at 02:00" as end.
- When I select "[yesterday] at 12:00" as start, I can select "[yesterday] at 10:00" as end.
- When I select "[tomorrow] at 12:00" as start, I can not select any end time.

The behavior is very confusing. 
I suggest that we dont set `minTime` at all, as it seems not do what we want.

tarnap:
@jone: Like this?
end_time_ii

opengever/meeting/browser/resources/datetimepicker.js
```diff
@@ -29,38 +29,39 @@
...
-        self.startwidget.setCurrentTime(date);
-        updateDate(date);
-        updateSelectableDates();
-      });
+
+    self.getMinTime = function () {
+      var start = self.startwidget.getCurrentTime();
+      var end = self.endwidget.getCurrentTime();
+      if (end.getDayOfYear() <= start.getDayOfYear()) {
+        return start.format2();
```
jone: What is `fromat2`? Where does it come from? Just curious, I don't know it.

tarnap: @jone I don't know where it comes from either. I guess it's the widget's internal function. I noticed that there are 3 format functions: `format0`, `format1` and `format2`. `format0` returns a string containing `format1` and `format2`.

PR #3487, opened 2017-10-10
SPV: Update and fix proposal tabs.

jone:
The proposal tab configuration had these problems:
- Sorting by various columns, such as title, did not work.
- Grouping does not make sense when sorting does not work.
- The default width of columns was not configured well.
- The Proposal-ID column did not help and therefore we want to get rid
of it.

Changes:
- Remove Proposal-ID column.
- Disable sorting and grouping on generated columns.
- Configure sane default widths.

bildschirmfoto 2017-10-10 um 12 21 59

Fixes https://github.com/4teamwork/gever/issues/137, where sorting by title did not work.
Withs are optimized for my 15" display.

### DIFF
--- a/opengever/meeting/browser/documents/proposalstab.py
+++ b/opengever/meeting/browser/documents/proposalstab.py
@@ -74,30 +74,37 @@ class ProposalListingTab(FilteredListingTab):
     @property
     def columns(self):
         return (
-            {'column': 'proposal_id',
-             'column_title': _(u'label_proposal_id',
-                               default=u'Reference Number')},
-
             {'column': 'decision_number',
              'column_title': _(u'label_decision_number',
                                default=u'Decision number'),
-             'transform': lambda item, value: item.get_decision_number()},
+             'transform': lambda item, value: item.get_decision_number(),
+             'sortable': False,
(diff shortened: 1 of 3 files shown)


opengever/meeting/browser/documents/proposalstab.py
```diff
@@ -74,30 +74,37 @@ class ProposalListingTab(FilteredListingTab):
...
+             'transform': lambda item, value: item.get_decision_number(),
+             'sortable': False,
+             'groupable': False,
+             'width': 80},
 
             {'column': 'title',
              'column_title': _(u'column_title', default=u'Title'),
-             'transform': proposal_link},
+             'transform': proposal_link,
+             'sortable': False,
```
phgross: @jone how are the listings sorted now? No column expect the review_state is sortable. But i think thats not a good default sorting.

jone: Good question; it's actually not sorted by a specific column. I think there is not much in the table where sorting makes sense. 🤔 
I think we have to add the title to the column at some point.
But for now I just want to avoid errors.

PR #3498, opened 2017-10-10
Remove link to dossier if permission is missing

tarnap:
If a user does not have the permission to view the dossier a link to the
dossier is not reasonable.
Therefore when the dossiers are listed, the view permissions are checked
and if the user is missing the permission to view the dossier only the
title of the dossier is rendered.

dossier_title_without_link_when_no_permission

Resolves https://github.com/4teamwork/gever/issues/125

### DIFF
--- a/opengever/meeting/tabs/meetinglisting.py
+++ b/opengever/meeting/tabs/meetinglisting.py
@@ -6,6 +6,8 @@ from opengever.meeting import _
 from opengever.meeting.model import Meeting
 from opengever.tabbedview import BaseListingTab
 from opengever.tabbedview import SqlTableSource
+from plone import api
+from Products.CMFPlone.utils import safe_unicode
 from zope.component import adapter
 from zope.globalrequest import getRequest
 from zope.i18n import translate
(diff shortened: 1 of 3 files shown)


tarnap:
@4teamwork/spv: We probably need a test for this... do we need a test for this?

jone:
👍 test would be great. You can just call the helper function directly and assert the result.

For the underline: I agree, we do not want an underline when its not a link.

jone:
Underline was removed here: https://github.com/4teamwork/plonetheme.teamraum/pull/589

opengever/meeting/tests/test_meeting_listings.py
```diff
@@ -0,0 +1,44 @@
...
+    features = ('meeting',)
+
+    maxDiff = None
+
+    @browsing
+    def test_span_appears_if_no_view_permission_on_dossier(self, browser):
+        self.login(self.dossier_responsible, browser)
+
+        self.meeting_dossier.manage_setLocalRoles(self.meeting_user.getId(),
+                                          ['This role does not exist'])
```
jone: Is this necessary? I'd expect that just blocking the inheritance (`__ac_local_roles_block__`) is enough. 🤔

opengever/meeting/tabs/meetinglisting.py
```diff
@@ -26,11 +27,16 @@ def translated_state(item, value):
...
 
-def dossier_link(item, value):
+def dossier_link_or_title(item, value):
     dossier = item.get_dossier()
     if not dossier:
         return
 
+    if not api.user.has_permission('View', obj=dossier):
+        span = u'<span title="{0}" class="{1}">{0}</span>'.format(
+            escape_html(dossier.title), get_css_class(dossier))
```
jone: If `dossier` is an actual document, it would be better to use the official accessor `dossier.Title()` for accessing the title.
But this will return a bytestring, therefore we need to decode it. 
So:
```python
escape_html(safe_unicode(dossier.Title()))
```

opengever/meeting/tests/test_meeting_listings.py
tarnap: You are right!
I needed to do so first because I was using the `self.dossier` instead of the `self.meeting_dossier`.

opengever/meeting/tabs/meetinglisting.py
tarnap: improvement made!

tarnap:
@jone: was fehlt hier?

PR #3501, opened 2017-10-11
 List team-tasks in team member's MyTask tab of the personal overview

### LINKED ISSUE #3415: Teamaufgaben in Auflistungen behandeln.
- [ ] In der persönlichen Auflistung `Meine Aufgabe` sollen auch Teamaufgaben aufgelistet werden, sofern der aktuelle Benutzer Mitglied dieses Teams ist.
 - [ ] Teamaufgaben sollen speziell markiert werden.

phgross:
The MyTask tab in the personal overview now shows also tasks, which are assigned to a team, where the current user is a team member.

Besides that, the `responsible` and `issuer` column in tasklistings is now linked and prefixed with a icon to differ the different actors (user, inbox, team).

bildschirmfoto 2017-10-11 um 11 57 28

Styling in a separate PR: https://github.com/4teamwork/plonetheme.teamraum/pull/590

Closes #3415

### DIFF
--- a/opengever/tabbedview/browser/tasklisting.py
+++ b/opengever/tabbedview/browser/tasklisting.py
@@ -2,6 +2,7 @@ from ftw.table import helper
 from ftw.table.interfaces import ITableSource, ITableSourceConfig
 from opengever.globalindex.model.task import Task
 from opengever.globalindex.utils import indexed_task_link_helper
+from opengever.ogds.base.actor import Actor
 from opengever.ogds.base.utils import get_current_admin_unit
 from opengever.tabbedview import _
 from opengever.tabbedview import BaseListingTab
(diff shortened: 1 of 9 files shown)


opengever/tabbedview/browser/tasklisting.py
```diff
@@ -25,6 +27,17 @@
 from zope.interface import Interface
 
 
+@ram.cache(author_cache_key)
+def linked_task_actor(item, actor):
+    if not isinstance(actor, unicode):
+        if actor is not None:
+            actor = actor.decode('utf-8')
```
jone: I think you could do that shorter:
```python
from Products.CMFPlone.utils import safe_unicode
actor = safe_unicode(actor) or ''
```

phgross: Right, updated ...

opengever/tabbedview/browser/tasklisting.py
```diff
@@ -25,6 +28,12 @@
 from zope.interface import Interface
 
 
+@ram.cache(author_cache_key)
+def linked_task_actor(item, actor):
+    actor = safe_unicode(actor) or ''
```
jone: Actually I think it should be `u''` instead of `''`, right? We want to have unicode..

phgross: yep 🙈

PR #3504, opened 2017-10-12
Refactor JSON schema generation and dumping

lukasgraf:
This PR refactors the generation of JSON schemas so that the code is more DRY, and is easier to maintain and test.

- Introduces a new `JSONSchema` class that handles building and manipulating JSON schemas
- Makes sure sort order of dumped VDEX vocabularies is stable
- Adds tests for the JSON schema generation
- Adds tests that verify that the checked in schema-dumps are up-to-date
- Fixes use of `anyOf` constraint: The current implementation only worked for a single `anyOf` constraint. Changed so the require-anyOf constraints are nested in a global `allO` constraint

I already changed the `parent_guid` property to use `require_any_of()`, even though in this version it's just a single property. Future parent pointers like `parent_reference_number` can simply be added [here](https://github.com/4teamwork/opengever.core/commit/b1049635799f10777351d8b980ff4a051203a911#diff-25ce786e16a8ce901ebbe8698d253e56R331).

---

📣 **Review notes**:
Unfortunately, the one big refactoring commit is a bit of a mess, and probably pretty hard to review. Probably best to look at the final code instead of the diff.

But I started with adding tests first, so for every subsequent commit, if the tests are green for that commit, the schema dumps didn't change (unless there's schema-dump updates *in* that commit).

Therefore I split the changes and updates to the schema dumps into separate commits, and pushed them to jenkins seperately, so every commit is tested.


opengever/base/schemadump/schema.py
```diff
@@ -171,218 +172,223 @@ def dump(self, sql_type):
...
+
+        desc = SEQUENCE_NUMBER_LABELS[self.portal_type]
+        self.ct_schema.add_property('sequence_number', {
+            'type': 'integer',
+            'title': u'Laufnummer',
+            'description': desc,
+        })
+
+    def _add_mail_properties(self):
+        if self.portal_type == 'opengever.document.document':
```
phgross: I would add a comment here `add_mail_properties` and checking for `opengever.document.document` is confusing.

lukasgraf:
@phgross updated 👍  - added comment regarding mails vs. documents.

PR #3511, opened 2017-10-13
OGGBundle parent reference number support

phgross:
This PR adds an additional property `parent_ref_tuple` instead of referencing the parent with the `parent_guid` property, it's now possible to reference an already existing container as a parent. The `parent_ref_tuple` property expects a nested list with all reference number parts, grouped by type. For example the subdossier with the reference number `Client 2.4.12.3 / 143.3` ->  would be referenced with `[[2, 4, 12, 3], [143, 3]]`.

This change is the first part of #3435.


opengever/base/reference_formatter.py
```diff
@@ -87,6 +88,39 @@ def sorter(self, brain_or_value):
             parts.append(part)
         return parts
 
+    def get_portal_part(self):
+        """Returns the reference number part of the portal, the adminunit's
+        abbreviation.
+        """
+        return get_current_admin_unit().abbreviation
+
+    def list_to_string(self, reference_list):
```
lukasgraf: This basically duplicates most of the logic from `complete_number()` - couldn't this method just turn a `reference_list` into the `numbers` format that `complete_number()` expects, and then delegate the formatting logic to `complete_number()`?

```python
[[1, 4, 5], [452, 4], [135]]
```
->
```
{
    'site': ['Client'],
    'repository': ['1', '4', '5'],
    'dossier': ['452', '4'],
    'document': ['135'],
}
```

opengever/bundle/sections/resolveguid.py
```diff
@@ -72,6 +95,27 @@ def register_items(self):
 
             self.bundle.item_by_guid[guid] = item
 
+            if 'parent_ref_tuple' in item:
+                reference_number = self.get_formatter().list_to_string(
+                    item['parent_ref_tuple'])
+                item['parent_ref_number'] = reference_number
+                used_ref_numbers.append(reference_number)
+
+        self.bundle.path_by_reference_number = self.get_reference_mapping(
```
lukasgraf: Since this probably could take a while, I would add some logging around this.

opengever/bundle/sections/resolveguid.py
```diff
@@ -72,6 +95,27 @@ def register_items(self):
...
+            used_ref_numbers)
+
+    def get_relative_path(self, brain):
+        """Returns the path relative to the plone site for the given brain.
+        """
+        return '/'.join(brain.getPath().split('/')[2:])
+
+    def get_reference_mapping(self, reference_numbers):
```
lukasgraf: `get` -> `build`

opengever/bundle/sections/resolveguid.py
```diff
@@ -72,6 +95,27 @@ def register_items(self):
 
             self.bundle.item_by_guid[guid] = item
 
+            if 'parent_ref_tuple' in item:
```
lukasgraf: I'm not terribly happy with the name `parent_ref_tuple` as a public-facing name in the bundle spec. I understand that we need to distinguish it from the formatted reference number (string), but since the OGGBundle is sort of a public API, I think it's important that we choose good names.

How about `parent_reference` for the tuple-style, and `formatted_refnum` for the formatted string (in the bundle code)?

opengever/bundle/schemas/documents.schema.json
```diff
@@ -197,6 +197,12 @@
                         "string"
                     ]
                 },
+                "parent_ref_tuple": {
```
lukasgraf: Could you please also update the OGGBundle specification (*section "Abbildung von Verschachtelung (containment)"*):
https://github.com/4teamwork/opengever.core/blob/master/docs/public/dev-manual/oggbundle/index.rst#abbildung-von-verschachtelung-containment

I would suggest to reword that section to something like this:

```
Da die Daten in den JSON-Dateien nicht verschachtelt abgelegt werden, ist es nötig diese Verschachtelung während dem Import aufzulösen. Diese Verschachtelung wird mittels global eindeutiger ID (GUID) und einem Pointer von Children auf das enthaltende Parent abgebildet. Dazu hat muss jedes Objekt über eine GUID verfügen. Diese muss im Attribut **guid** gespeichert werden. Die Verschachtelung wird mittels einer Referenz auf das Parent hergestellt, dazu muss jedes Objekt, das ein Parent besitzt, das Attribut **parent\_guid** definieren, und damit auf das Parent referenzieren:

  {
  "guid": "7777-0000-0000-0000",
  ...
  },
  {
  "guid": "9999-0000-0000-0000",
  "parent_guid": "7777-0000-0000-0000",
  ...
  }

Alternativ zur (im Bundle selbst definierten) GUID kann auch das Akzenzeichen eines Objekts als eindeutige Referenz auf das Parent verwendet werden. Die Verwendung des Aktenzeichens als Parent-Pointer erlaubt es, bereits existierende Objekte (die also nicht im referenzierenden Bundle mitgeliefert werden) zu referenzieren, und ermöglicht so partielle Importe. So ist z.B. das importieren von Dokumenten in ein bestehendes Dossier möglich, indem dieses Dossier über sein Aktenzeichen referenziert wird.

Wird zur Referenzierung das Aktenzeichen verwendet, muss dazu das Attribut **parent\_reference** (statt **parent\_guid**) gesetzt werden. Das Aktenzeichen in diesem Attribut wird als verschachtelte Arrays von Integern erwartet, welche die einzelnen Komponenten des Aktenzeichens (ohne Formatierung) abbilden. Beispiel: `[[1, 3, 5], [472, 9]` entspricht dem Aktenzeichen `1.3.5 / 472.9` (Position 1.3.5, Dossier 472, Subdossier 9):

  {
  "guid": "9999-0000-0000-0000",
  "parent_reference": [[1, 3, 5], [472, 9],
  ...
  }
```

lukasgraf:
Just as a reminder / xref: In Issue #1561 we track some long standing changes to the reference number implementation that we still need to fix.

opengever/bundle/schemas/documents.schema.json
lukasgraf: @phgross BTW, before you push, please also add a changelog entry in the OGGBunde specification docs (and I would suggest to bump the version to `1.0`)

phgross:
@lukasgraf updated as discussed

PR #3512, opened 2017-10-13
Update testing libs

jone:
Update to newest `ftw.testing`
- Switch to `FTWIntegrationTesting`, which includes the transaction interceptor setup.
- We still do SQL integration (mark session)

Update to newest `ftw.testbrowser`
- Handles auto CSRF automatically.


opengever/core/testing.py
```diff
@@ -491,63 +493,17 @@ def allowAllTypes(self, portal):
...
-            self.handlePubAfterTraversal, [IPubAfterTraversal])
-
-    def tearDown(self):
-        self.interceptor.uninstall()
-        super(GEVERIntegrationTesting, self).tearDown()
+class GEVERIntegrationTesting(FTWIntegrationTesting):
 
-    def testSetUp(self):
-        super(GEVERIntegrationTesting, self).testSetUp()
+    def makeSavepoint(self):
```
Rotonen: Can you elaborate?

Rotonen:
Does this now squelch CSRF warnings on empty responses in the tests or what's that with the CSRF, just that we do not have to manually amend the token in the tests?

opengever/core/testing.py
jone: We're extending the superclass here: https://github.com/4teamwork/ftw.testing/blob/f8645fb30b45f43b5b6f1c8258c0e55c21304181/ftw/testing/integration_testing.py#L55
See the comment a line below:
```
         # In order to let the SQL transaction manager make a savepoint of no
         # changes we need to mark the session as changed first. 
```

jone:
@Rotonen The problem is that in this setup we use one connection and never commit changes. This means that new requests will operate on a "dirty" connection in tests. Those dirty objects will trigger auto CSRF.
There `handlePubAfterTraversal` method, which I'm removing here, did mark all dirty objects on the connection before publishing a request, so that the auto CSRF only tracks new changes, kind of.
This behavior was now moved to ftw.testbrowser, because it now supports Plone 5, which has auto CSRF: https://github.com/4teamwork/ftw.testbrowser/pull/74/commits/c6aa24ee41a9cfe59366ed0a6ae1373c2a0cc0f2

The problem with warnings on empty responses is that those empty responses say that they are `text/html`, but it is not (it has no `<html>`, for example). So those are bugs in our code, not a testing issue.

opengever/core/testing.py
Rotonen: Previously savepoints did not work with SQLite. Did you figure something new out in `ftw.testing` to enable this?

jone: No, they work well, even when nested.
it just isn't possible to roll back a savpoint when the transaction is committed.
So the point of our integration testing is that it does not commit.
We use the [transaction interceptor](https://github.com/4teamwork/ftw.testing#transaction-interceptor) to make sure that it is not possible to commit the transaction. That's the trick 😉

PR #3513, opened 2017-10-13
Do not render document link without View permission.

jone:
When a user does not have the permission to view a document, the DocumentLinkWidget should not render a link (which will lead to unsufficient privileges), but render a text instead.

This does not affect regular listings, where the users doesnt see such documents anyway, but explicit relations where the user must know that a document exists, but may not have access to the document.

bildschirmfoto 2017-10-13 um 14 30 38

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2017.5.2 (unreleased)
 ---------------------
 
+- Do not render document link without View permission. [jone]
 - Refactor JSON schema generation and dumping, add tests. [lgraf]
 - Ungrok opengever.document. [elioschmutz]
 - SPV word: Selectable proposal templates can be configured per committee. [jone]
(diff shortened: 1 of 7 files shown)


Rotonen:
Could it also have a no-access icon next to the document icon so users do not end up thinking link generation has broken?

jone:
This could look like this:
bildschirmfoto 2017-10-13 um 14 51 12

I have mixed feelings.
We already have a tooltip telling the problem.

Opinions?

Rotonen:
I'm not trusting people to get what is going on from the tooltip alone, but I have more faith in people to hover their mouse over it once there is the icon in place.

Can one easily overlay the icon on top of the document icon via CSS?

jone:
Hmm, I can't overlay it so that it looks nice 🤔 impressions:
bildschirmfoto 2017-10-13 um 15 00 13
bildschirmfoto 2017-10-13 um 15 00 36

Rotonen:
Yeah, that'd require way too much fiddly detail work and potentially dynamic colour stuff for other icons to look nice.

Anyone else?

jone:
I have added a `no-access` class to the span; this allows for such stylings. Sine the styling itselfs belongs to `plonetheme.teamraum` anway, I'll make a separate PR for the layout.
So we can continue here 😉

PR #3516, opened 2017-10-16
Extract bumblebee preview generation into standalone component.

jone:
**Incentive:**
The bumblebee preview listing is currently implemented in the bumblebee browser view mixin.
This makes it impossible to reuse it in other places than tabbed views.
In the current meeting development, we will use the listing in other places (e.g. meetings, proposal views, etc), so there is a need for a more flexible component for rendering those previews.

**Change:**
This pull request extracts the functionality to render a bumblebee preview listing, such as in bumblebee tabs, into a new, standalone component.
It replaces the same functionality in the existing bumblebee preview listing tab mixin in order to proof its functionality and avoid duplicates.

**Implementation Details:**
The new `PreviewListing` is instantiated with a view (in order to have context and request for page template rendering) and is configured by chaining methods. It supports objects and brains and allows for easy extension when we want to support other things (such as SQL objects) in the future.
A fetch URL must be provided. The view must make sure that the URL works (`allowed_attributes`).

**Example:**
```python
listing = (PreviewListing(view)
           .for_objects(list_of_objects)
           .with_fetch_url(fetch_url)
           .with_batchsize(10))
listing.render()
```


jone:
@phgross Yes I did verify that fetching works. 
Also: there are already existing, untouched tests, verifying that the view/fetch endpoint works.

PR #3517, opened 2017-10-17
Fix send as attachment action

### LINKED ISSUE #3443: Attach to outlook action not working on inbox documents listing
bildschirmfoto 2017-09-28 um 09 16 01

The action raises an JS error.

tarnap:
The action "send as attachment" in the inbox raised a javascript error,
because the service, which is needed to fill the attributes of the email
was not available.
This commit adds a default service which provides an empty set of values for the
resulting email.

no_error_message

Resolves #3443

### DIFF
--- a/opengever/base/service.py
+++ b/opengever/base/service.py
@@ -30,3 +30,10 @@ class DocumentStatus(Service):
             payload['locked_by'] = None
 
         return json.dumps(payload)
+
+
+class DefaultEmailAttributes(Service):
+    """Prevent action send as attachment to fail."""
+
+    def render(self):
+        return json.dumps({})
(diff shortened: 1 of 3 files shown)


opengever/base/service.py
```diff
@@ -30,3 +30,10 @@ def render(self):
             payload['locked_by'] = None
 
         return json.dumps(payload)
+
+
+class DefaultEmailAttributes(Service):
+    """Prevent action send as attachment to fail."""
+
+    def render(self):
+        return json.dumps({})
```
Rotonen: No test?

tarnap:
This pull request is being replaced by...

PR #3519, opened 2017-10-17
OGGBundle spec: Link to schemas in og.core instead of duplicating them for docs.

lukasgraf:
This removes duplication of schema dumps used for the OGGBundle spec.

Also contains some smaller drive-by fixes.


phgross:
Docs published ...

PR #3521, opened 2017-10-17
Ungrok plone directives

elioschmutz:
This PR removes groked `plone.directives`. It also checks for the `form.default_value`

related to #3205


elioschmutz:
@phgross

PR #3522, opened 2017-10-18
Fix missing response-variable due to ungrok opengever.document.

elioschmutz:
This PR accesses the response-variable the default-way instead the grok-way.

Related to #3205


elioschmutz:
Yes, I checked all the other packages.

lukasgraf:
https://sentry.4teamwork.ch/sentry/onegov-gever/issues/4277/

PR #3527, opened 2017-10-18
Fix wrong default value in filingnumber-adjustment view.

elioschmutz:
This PR fixes the wrong default value in filingnumber-adjustment view.

closes #3526

PR #3528, opened 2017-10-18
Ungrok opengever core

elioschmutz:
This PR ungroks `opengever.core`.

related to #3205

⚠️ @phgross beside of the og.maintenance dependency there is one dependency with a grok-dependency. The `collective.z3cform.datagridfield`. Unfortunately there is no newer version for plone4.3 without grok. Should I try to update the ungrokked version of `collective.z3cform.datagridfield` to run with plone4.3? Or doesn't it matter if it's grokked or not?


Rotonen:
As the latest master is still tested against Plone 4.3 https://github.com/collective/collective.z3cform.datagridfield/blob/master/test-4.3.x.cfg should be worth a shot.

elioschmutz:
https://travis-ci.org/collective/collective.z3cform.datagridfield/branches :joy:

@Rotonen do you think it's a good idea?

Anyway: I installed it from master without success. I would have to spend more time into debugging if it is desired.

Rotonen:
I'd say yes to working towards cutting a new workable-for-us grok-free release of `collective.z3cform.datagridfield` upstream, but indeed @phgross should make the call.

phgross:
Hmm I'll take a look into `collective.z3cform.datagridfield` later this day, but as I see, the datagridwidget is only used by the filingnumber form, a maintenance form which is use very rarely ...

elioschmutz:
@phgross the form is broken anyway. I already made a PR to fix default values: (https://github.com/4teamwork/opengever.core/pull/3527). But the datagrid witdget would be still broken. If you try to add/remove rows, you'll get JS errors. So I think, if this form have ever been used once, then just for changing an existing number. All the other usecases haven't been possible due to software-issues.

So +1 for remove the whole view 😉

phgross:
@elioschmutz: I've discussed it with @lukasgraf, please remove the form and with that also the dependency to `collective.z3cform.datagridfield`.

elioschmutz:
@phgross 👍 tests are passing...

lukasgraf:
:tada:

PR #3529, opened 2017-10-18
Fix send as attachment

### LINKED ISSUE #3443: Attach to outlook action not working on inbox documents listing

tarnap:
The action "send as attachment" in the inbox raised a javascript error,
because the service, which is needed to fill the attributes of the email
was not available.
This commit adds a default service which provides an empty set of values for the
resulting email.

no_error_message

Also add inbox fixture and add missing reindex after changing roles.

Resolves #3443

### DIFF
--- a/opengever/base/tests/test_email_attributes_service.py
+++ b/opengever/base/tests/test_email_attributes_service.py
@@ -1,0 +1,13 @@
+from ftw.testbrowser import browsing
+from opengever.testing import IntegrationTestCase
+
+
+class TestEmailAttributesService(IntegrationTestCase):
+
+    @browsing
+    def test_service_is_available_for_inbox(self, browser):
+        self.login(self.secretariat_user, browser)
+        browser.open(self.inbox,
+                     view='attributes',
+                     headers={'Accept': 'application/json'})
+        self.assertEquals({}, browser.json)
--- a/opengever/testing/fixtures.py
+++ b/opengever/testing/fixtures.py
@@ -50,6 +50,7 @@ class OpengeverContentFixture(object):
                 self.create_templates()
                 with self.features('meeting'):
                     self.create_committees()
+                self.create_inbox()
 
         with self.freeze_at_hour(14):
             with self.login(self.dossier_responsible):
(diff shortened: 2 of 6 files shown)


opengever/testing/fixtures.py
```diff
@@ -301,6 +302,18 @@ def create_committees(self):
...
 
     @staticuid()
+    def create_inbox(self):
+        self.inbox = self.register('inbox', create(
+            Builder('inbox')
+            .titled(u'Eingangsk\xf6rbli')
+            .having(id='eingangskorb',
+                    responsible_org_unit='fa',
+                    inbox_group=self.org_unit.inbox_group)))
+        self.inbox.manage_setLocalRoles(
```
Rotonen: Should this rather live as a method in the builder instead so you just pass the user to the builder in the fixture setup and it does the reindex under the hood in the builder?

opengever/base/tests/test_email_attributes_service.py
```diff
@@ -0,0 +1,13 @@
+from ftw.testbrowser import browsing
+from opengever.testing import IntegrationTestCase
+
+
+class TestEmailAttributesService(IntegrationTestCase):
+
+    @browsing
+    def test_service_is_available_for_inbox(self, browser):
```
Rotonen: Just for reference:

For other OfficeConnector user stories we have play-throughs in the API testing.

https://github.com/4teamwork/opengever.core/blob/master/opengever/officeconnector/tests/test_api.py

I'll add this user story later to the tests when I port them to be integration tests.

opengever/testing/fixtures.py
tarnap: @Rotonen: [I created an issue for this in the builder](https://github.com/4teamwork/ftw.builder/issues/57).
It is out of the scope of this pull request.

Rotonen: Negative, the builder in question is on the `opengever.core` side:

https://github.com/4teamwork/opengever.core/blob/master/opengever/testing/builders/dx.py#L67

jone: @Rotonen The inbox builder is not the right place to do that, because local roles is not an inbox feature, its a Plone feature. So if we refactor that, it should go into [`ftw.builder.builder.PloneObjectBuilder`](https://github.com/4teamwork/ftw.builder/blob/4f654c6827dc9eabc0d04eb6fa9176a95c31356d/ftw/builder/builder.py#L79). This is the common superclass of object-builders supporting local roles. It would allow us to benefit outside of GEVER too.

But anyway: it is out of scope here. This pull request is not about fixtures, it is about the "send as attachment" feature. It is necessary to have an inbox in order for the test to pass, but it is not relevant how the roles are set. So refactoring / extracting functionality to `ftw.builder` should go into a separate pull request and a separate discussion 😉

tarnap: @jone: merci :)

tarnap:
@jone: Could you review this pull request?

PR #3530, opened 2017-10-19
Do not link meeting dossier if no view permission

tarnap:
If a user tries to access a view he has no permission to, the
application raises an error.
There should be no link to a view to which a user has no permission,
therefore the link should only be displayed if the user has the required
permission.

dossier_title_without_link_when_no_permission_2

Resolves https://github.com/4teamwork/gever/issues/126


jone:
Sorry, I didn't plan this well:
In https://github.com/4teamwork/opengever.core/pull/3543 I have moved this link to the byline, so the code touched here is no longer there.
It looks like this:
bildschirmfoto 2017-10-24 um 08 20 51

Therefore #3543 replaces this pull request.

PR #3534, opened 2017-10-19
Add dossiertemplates configuration flag `respect_dossier_depth`.

phgross:
Currently dossiertemplates are not respecting the maximum dossier depth configuration option. The current behavior is the desired one, for most of the customers, but the behavior should be
configurable. Therefore we add an additional registry option `respect_dossier_depth` to the dossiertemplate settings. Which is disabled by default.

See https://basecamp.com/2768704/projects/14549453/todos/326699123 for
more information.

### DIFF
--- a/opengever/dossier/dossiertemplate/dossiertemplate.py
+++ b/opengever/dossier/dossiertemplate/dossiertemplate.py
@@ -5,6 +5,8 @@ from opengever.dossier import _
 from opengever.dossier.behaviors.dossier import IDossier
 from opengever.dossier.dossiertemplate.behaviors import IDossierTemplateMarker
 from opengever.dossier.dossiertemplate.behaviors import IDossierTemplateSchema
+from opengever.dossier.dossiertemplate.interfaces import IDossierTemplateSettings
+from opengever.dossier.interfaces import IDossierContainerTypes
 from opengever.dossier.utils import truncate_ellipsis
 from opengever.ogds.base.actor import Actor
 from plone import api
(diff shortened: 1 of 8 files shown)


opengever/dossier/dossiertemplate/dossiertemplate.py
```diff
@@ -75,6 +77,32 @@ def label(self):
...
+        return types
+
+    def is_subdossier_allowed(self):
+        if self.is_respect_max_dossier_depth:
+            max_dossier_depth = api.portal.get_registry_record(
+                'maximum_dossier_depth', interface=IDossierContainerTypes)
+
+            current_levels = len([item for item in self.aq_chain
+                                  if IDossierTemplateSchema.providedBy(item)])
+            return current_levels <= max_dossier_depth
```
jone: I have the suspicion that there is an off-by-one error here:
Given the `max_dossier_depth = 2` 
and the current context is a sub-dossier-template
`aq_chain` is something like: `[<subdossiertemplate>, <dossiertemplate>, <templates>, <site>, <app>]`,
therefore `current_levels = 2`
therefore `current_levels <= max_dossier_depth` is `2 <= 2` is `True`

But if the max_depth is 2, I should be able to add a sub-sub-dossier, because this would result in a depth of 3, right?

Or am I getting this wrong?

phgross: @jone I double checked it local: A `max_dossier_depth =2` means that two subdossiers levels are allowed, so actually 3 levels of dossiers. So your example above is right, but thats how it should be. 

The naming/counting is definitely confusing, but I won't change that in this PR.

jone: Ah, well, then it's correct 👍 
I've just interpreted that the primary dossier is counted in too.

PR #3537, opened 2017-10-19
Fix fetch new showroom items in serach-view

### LINKED ISSUE #3535: Fetch new showroom items in advanced-search will fail after second batch.
The `ftw.showroom` fetches new items as soon as it reaches the end of its current batch.

The advanced-search implementation of the fetch-method is the following:

https://github.com/4teamwork/opengever.core/blob/master/opengever/bumblebee/browser/resources/bumblebee.js#L112

It grabs the next-button of the advanced-search batch-view and calls this link per ajax to request the new items. Unfortunately this link does not change after fetching new items. The advanced-search-view will be still on the first batch. So reaching the end of the second batch in the showroom will call the same fetch-link again. You'll get the same items once again.

elioschmutz:
The showroom search-view implementation for fetching new objects will use the default
next/previous button urls of the searchform to request new objects based on the current search
params.

If we load only the first next or previous batch-slice, this will work well. But if we
reach the end/beginning of the second batch-slice, it will reuse the same fetch-link as for
the first request, because the search-view itself will not update its buttons after fetching
new objects through the showroom.

To solve this issue, we implement a new class for handling the current request-urls for the
tail/head function of the showroom.

There is a side-issue (not required for this PR!): https://github.com/4teamwork/ftw.showroom/issues/45 - The issue describes a bad behavior where it is possible, that the next/previous batch-slice will be swapped with the after-next/previous batch slice. 

closes #3535

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -9,6 +9,7 @@ Changelog
 - Remove the broken and unused fillingnumber-adjustment view to get rid of the grokked
   dependency collective.z3cform.datagridfield. [elioschmutz]
 - Reenable action "move items" for inbox. [tarnap]
+- Fix batch-handling in the showroom overlay for the search-view. [elioschmutz]
 - OGGBundle: Fix tracking of item counts (actual vs. raw). [lgraf]
 - OGGBundle: Support delta imports using GUID. [lgraf]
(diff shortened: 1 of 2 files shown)


elioschmutz:
@phgross ready to review...

elioschmutz:
👍  rebased and PR description updated

PR #3538, opened 2017-10-20
OGGBundle: Support delta imports using GUID:

lukasgraf:
This allows items in bundles to reference a GUID via `parent_guid` that refers to an object that has previously been imported into Plone.

If an object with the same GUID exists both in Plone as a JSON item in the bundle, the one from Plone is used as the parent container for any items that refer to that GUID.

---

The **report logic** had to change slightly to accommodate this change:
In order to generate the report, we now track the acutal constructed GUIDs in `bundle.constructed_guids`. For generating the report, we simply query the catalog for all these GUIDs.

### DIFF
--- a/opengever/bundle/sections/resolveguid.py
+++ b/opengever/bundle/sections/resolveguid.py
@@ -32,6 +32,10 @@ class ReferenceNumberNotFound(Exception):
     pass
 
 
+class MissingParentPointer(Exception):
+    pass
+
+
 class ResolveGUIDSection(object):
     """Resolve and validate GUIDs.
 
@@ -62,6 +66,8 @@ class ResolveGUIDSection(object):
 
         # Table of formatted refnums that exist in Plone
         self.bundle.existing_refnums = ()
+        # Table of bundle GUIDs that exist in Plone
+        self.bundle.existing_guids = ()
 
         # Current reference number formatter
         self.formatter = None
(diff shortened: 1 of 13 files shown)


opengever/bundle/sections/resolveguid.py
```diff
@@ -91,38 +97,59 @@ def get_existing_refnums(self):
...
-                    "Couldn't find container with reference number %s "
-                    "(referenced as parent by item by GUID %s )" % (
-                        formatted_refnum, guid))
+        # Verify parent pointers - all referenced items/containers must exist
+        for guid, item in self.bundle.item_by_guid.items():
+            parent_guid = item.get('parent_guid')
+            parent_reference = item.get('parent_reference')
+
+            if parent_guid is not None:
+                if not (guid in self.bundle.item_by_guid or
```
phgross: `guid in self.bundle.item_by_guid` is obsolete because your looping over `self.bundle.item_by_guid`

lukasgraf: @phgross nice catch! The idea here was to check whether the **`parent_guid`** is in either of those mappings, instead of the `guid`. So this check was pretty wrong 🙈  (Tests didn't fail because later down the line in `build_tree()` the error is caught, and we *also* raise `MissingParent` there.

lukasgraf:
@phgross updated

PR #3542, opened 2017-10-23
Avoid id conflicts when setting up a repository.

phgross:
Actually this should be done by the constructor section in general, but for now I add this fix to the
`PathFromReferenceNumberSection` which generates the path and id for all repositories.


opengever/setup/sections/reference.py
```diff
@@ -115,6 +121,14 @@ def normalize(self, item, max_length):
         normalized_id = self.normalizer.normalize(title, max_length=max_length)
         normalized_id = self.id_normalizer.normalize(
             normalized_id, max_length=max_length)
+
+        # Avoid id conflicts
+        parent = traverse(self.context, parent_path)
+        chooser = INameChooser(parent)
+        if not chooser.checkName(title.encode('utf-8'), parent):
+            normalized_id = INameChooser(parent).chooseName(
+                normalized_id, parent)
```
jone: The second paramenter of `chooseName` should be the new object, not the parent.
See: https://github.com/plone/plone.app.content/blob/396468bb70e9236af30a9c534ac3c81243e60172/plone/app/content/namechooser.py#L31
But since we pass in the `normalized_id`, the new object is only used for getting the request.

Did you pass the parent on purpose here?

phgross: @jone No not really, the new object does not exists at this point of the transmogrifier pipeline, should I switch to the self.context (the plonesite) instead?

jone: Well, then it does not matter whether we pass in `parent` or `self.context`.
👍

PR #3545, opened 2017-10-24
Remove unpermitted link to excerpt document

tarnap:
When an excerpt was returned, the user sees a link to the excerpt
document in the view of the proposal.
If the user has no permission to view the excerpt document, the link
should not be displayed.

Old:

New (left: with permission - right: without permission):
protokollauszug_disabled

Resolves https://github.com/4teamwork/gever/issues/132

### DIFF
--- a/opengever/meeting/browser/proposaloverview.py
+++ b/opengever/meeting/browser/proposaloverview.py
@@ -2,6 +2,7 @@ from ftw import bumblebee
 from opengever.base.browser.helper import get_css_class
 from opengever.bumblebee import is_bumblebee_feature_enabled
 from opengever.document import _ as document_mf
+from opengever.document.widgets.document_link import DocumentLinkWidget
 from opengever.meeting import is_word_meeting_implementation_enabled
 from opengever.meeting.browser.proposaltransitions import ProposalTransitionController
 from opengever.meeting.interfaces import IHistory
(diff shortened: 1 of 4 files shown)


opengever/meeting/browser/proposaloverview.py
```diff
@@ -31,15 +34,21 @@ def get_main_attributes(self):
     def get_css_class(self, item):
         """Return the sprite-css-class for the given object.
         """
-        css = get_css_class(item)
-        return '{} {}'.format("rollover-breadcrumb", css)
+        classes = ['rollover-breadcrumb', get_css_class(item)]
+        if not self.has_dossier_view_permission():
```
jone: The name `get_css_class` sounds quite generic, but we are testing here for the permission on the excerpt.
If we'd pass in an attachment for example, it could get confusing; and the name `get_css_class` will allow us to do that in the future. Also the name `has_dossier_view_permission` is not accurate, as we do not test the dossier permission.

I suggest that you assess if we can use the [`DocumentLinkWidget`](https://github.com/4teamwork/opengever.core/blob/master/opengever/document/widgets/document_link.py#L7) for rendering the link.

The `DocumentLinkWidget` already checks the view permission. It also renders a link with a bumblebee overlay (that seems to be missing at the moment).

@tarnap could you try that?

PR #3549, opened 2017-10-25
Fix zip export when agendaitemlisttemplate missing

### LINKED ISSUE #3338: Meeting zip export of a meeting fails if there is no agenda item list template set on the committee container
This could be handled gracefully in the zip exporter.

It is likely this issue has many near cousins as well so some exploratory work to find and fix those as well is welcome.

tarnap:
The creation of the zip file fails when there is no agenda item list
template.
The export itself should work when there is no agenda item list
template, the agenda item list should not be included in the zip file in
that case.
Since getting the agenda item list template is obfuscated, a try-except
is used to prevent the server error.

Resolves #3338

### DIFF
--- a/opengever/meeting/tests/test_zipexport.py
+++ b/opengever/meeting/tests/test_zipexport.py
@@ -1,0 +1,29 @@
+from ftw.testbrowser import browsing
+from ftw.testbrowser.pages import statusmessages
+from opengever.testing import IntegrationTestCase
+
+
+class TestZipExportWithWord(IntegrationTestCase):
+
+    features = ('meeting', 'word-meeting')
+
+    @browsing
+    def test_meeting_can_be_exported_to_zip_when_agenda_item_list_template_is_missing(self, browser):
+        self.login(self.committee_responsible, browser)
+        self.committee.agendaitem_list_template = None
+        self.committee_container.agendaitem_list_template = None
+        browser.open(self.meeting, view='export-meeting-zip')
+        statusmessages.assert_no_error_messages()
+
+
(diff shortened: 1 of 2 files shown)


opengever/meeting/tests/test_zipexport.py
```diff
@@ -0,0 +1,29 @@
...
+            self, browser):
+        self.login(self.committee_responsible, browser)
+        self.committee.agendaitem_list_template = None
+        self.committee_container.agendaitem_list_template = None
+        browser.open(self.meeting, view='export-meeting-zip')
+        statusmessages.assert_no_error_messages()
+
+
+class TestZipExportWithoutWord(IntegrationTestCase):
+
```
jone:
```
    features = ('meeting',) 
```
Shouldn't the meeting feature be activated here?

opengever/meeting/tests/test_zipexport.py
```diff
@@ -0,0 +1,29 @@
...
+from opengever.testing import IntegrationTestCase
+
+
+class TestZipExportWithWord(IntegrationTestCase):
+
+    features = ('meeting', 'word-meeting')
+
+    @browsing
+    def test_meeting_can_be_exported_to_zip_when_agenda_item_list_template_is_missing(  # noqa
+            self, browser):
```
jone: Our `max-line-length` is [currently `150`](https://github.com/4teamwork/ftw-buildouts/blob/master/pycodestyle.cfg#L11), so I'd prefer having the arguments on the same line. 🤔

tarnap:
@jone: the feature has been added :)

PR #3552, opened 2017-10-26
Prevent editing agenda item list when it should not be allowed.

### LINKED ISSUE #3551: New agenda items can be added in meeting held state
The item won't have a decision number in that case. The state-check in the add-view is not correct.

screen shot 2017-10-26 at 08 46 24

deiferni:
No longer allow the following actions when the meeting has been held:

- schedule paragraphs
- schedule agenda items
- edit agenda item title
- delete agenda items

Fixes #3551.

### DIFF
--- a/opengever/meeting/browser/meetings/agendaitem.py
+++ b/opengever/meeting/browser/meetings/agendaitem.py
@@ -245,8 +245,7 @@ class AgendaItemsView(BrowserView):
         """Updates the order of the agendaitems. The new sortOrder is expected
         in the request parameter `sortOrder`.
         """
-        if not self.context.model.is_editable():
-            raise Unauthorized("Editing is not allowed")
+        self.check_agendalist_editable()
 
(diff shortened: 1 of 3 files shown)


opengever/meeting/browser/meetings/agendaitem.py
```diff
@@ -459,6 +456,10 @@ def check_editable(self):
         if not self.meeting.is_editable():
             raise Unauthorized("Editing is not allowed")
 
+    def check_agendalist_editable(self):
```
jone: I'd prefer `require` over `check`.
`check` could also be interpreted as returning a boolean, that's why I don't like this word in method names.

deiferni: there is already a `check_editable` method, i followed the convention 😉 . but i agree, require also indicates the purpose a bit better, so i'll rename both.

deiferni:
@jone updated according to your suggestions

PR #3553, opened 2017-10-26
Display error message to client when state has changed in meantime

### LINKED ISSUE #2436: Prevent duplicate submission of reopening/revise
Investigate if ie11 behaves differently

https://sentry.4teamwork.ch/sentry/onegov-gever/issues/1370/
https://sentry.4teamwork.ch/sentry/onegov-gever/issues/1547/

```
AssertionError: 
(2 additional frame(s) were not displayed)
...
  File "home/zope/eggs/opengever.core-4.14.1-py2.7.egg/opengever/meeting/browser/meetings/agendaitem.py", line 249, in reopen
    self.agenda_item.reopen()
  File "home/zope/eggs/opengever.core-4.14.1-py2.7.egg/opengever/meeting/model/agendaitem.py", line 327, in reopen
    self.workflow.execute_transition(None, self, 'decided-revision')
  File "home/zope/eggs/opengever.core-4.14.1-py2.7.egg/opengever/meeting/workflow.py", line 88, in execute_transition
    transition.execute(obj, model)
  File "home/zope/eggs/opengever.core-4.14.1-py2.7.egg/opengever/meeting/workflow.py", line 47, in execute
    assert self.can_execute(model)

AssertionError
```

deiferni:
This PR changes the behaviour of the agenda item list view to also display error messages to users when an ajax request failed because the servers state has changed in the meantime. We also trigger a reload of the agenda-item list for some of these cases to display the current state.

screen shot 2017-10-26 at 15 28 13

Changes include:
- throw exceptions in model instead of `assert` statements
- catch exceptions in controller/view and display messages to users
- also use correct status codes in json responses instead of 200 everywhere
- always display the error messages in a json response instead of just for succeeded responses

Closes #2436.

### DIFF
--- a/opengever/meeting/browser/meetings/agendaitem.py
+++ b/opengever/meeting/browser/meetings/agendaitem.py
@@ -185,7 +185,7 @@ class AgendaItemsView(BrowserView):
             'edit_document_button': button}
 
     def _get_agenda_items(self):
-        meeting = self.context.model
+        meeting = self.meeting
         agenda_items = []
         for item in meeting.agenda_items:
             data = item.serialize()
@@ -248,7 +248,7 @@ class AgendaItemsView(BrowserView):
         """
         self.require_agendalist_editable()
 
-        self.context.model.reorder_agenda_items(
+        self.meeting.reorder_agenda_items(
             json.loads(self.request.get('sortOrder')))
 
         return JSONResponse(self.request).info(
@@ -305,7 +305,7 @@ class AgendaItemsView(BrowserView):
         """
         meeting_state = self.meeting.get_state()
 
-        if not self.context.model.is_editable():
+        if not self.meeting.is_editable():
             raise Unauthorized("Editing is not allowed")
 
         error_response = self._checkin_proposal_document_before_deciding()
@@ -360,7 +360,7 @@ class AgendaItemsView(BrowserView):
         """Reopen the current agendaitem. Set the workflow state to revision
         to indicate that editing is possible again.
         """
-        if not self.context.model.is_editable():
+        if not self.meeting.is_editable():
             raise Unauthorized("Editing is not allowed")
 
         self.agenda_item.reopen()
@@ -373,7 +373,7 @@ class AgendaItemsView(BrowserView):
         """Revise the current agendaitem. Set the workflow state to decided
         to indicate that editing is no longer possible.
         """
-        if not self.context.model.is_editable():
+        if not self.meeting.is_editable():
             raise Unauthorized("Editing is not allowed")
 
         try:
@@ -471,7 +471,7 @@ class AgendaItemsView(BrowserView):
         """Generate an excerpt of an agenda item and store it in
         the meeting dossier.
         """
-        if not self.context.model.is_editable():
+        if not self.meeting.is_editable():
             raise Unauthorized("Editing is not allowed")
 
         if not self.agenda_item.can_generate_excerpt():
--- a/opengever/meeting/browser/meetings/agendaitem.py
+++ b/opengever/meeting/browser/meetings/agendaitem.py
@@ -103,12 +103,14 @@ def return_jsonified_exceptions(func):
             return JSONResponse(getRequest()).error(
                 _(u'invalid_agenda_item_state',
                   default=u'The agenda item is in an invalid state for '
-                           'this action.')).dump()
+                           'this action.'),
+                status=401).dump()
 
         except Unauthorized:
             return JSONResponse(getRequest()).error(
                 _(u'editing_not_allowed',
-                  default=u'Editing is not allowed.')).dump()
+                  default=u'Editing is not allowed.'),
+                status=401).dump()
 
         except MissingMeetingDossierPermissions:
             return JSONResponse(getRequest()).error(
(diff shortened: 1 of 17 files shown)


opengever/meeting/locales/de/LC_MESSAGES/opengever.meeting.po
```diff
@@ -791,6 +791,11 @@ msgstr "Sitzung durchführen"
 msgid "inactive"
 msgstr "Inaktiv"
 
+#. Default: "The agenda item is in an invalid state for this action."
+#: ./opengever/meeting/browser/meetings/agendaitem.py
+msgid "invalid_agenda_item_state"
+msgstr "Diese Aktion kann im aktuellen Status des Traktandums nicht durchgeführt werden."
```
jone: 🤔 "Status" könnte mit dem Review-State verwechselt werden. Wäre hier allenfalls "Zustand" besser?

opengever/meeting/browser/meetings/agendaitem.py
```diff
@@ -89,6 +91,43 @@ def return_excerpt():
...
+                _(u'invalid_agenda_item_state',
+                  default=u'The agenda item is in an invalid state for '
+                           'this action.'),
+                status=401).dump()
+
+        except Unauthorized:
+            return JSONResponse(getRequest()).error(
+                _(u'editing_not_allowed',
+                  default=u'Editing is not allowed.'),
+                status=401).dump()
```
jone: I would always use `.remain()` (see original `MissingMeetingDossierPermissions` handling).
Some controller actions decide based on `.remain()` / `.proceed()` what should happen on response.
We usually want to indicate that whetever the request was, it did not work. The API for that is to use `.remain()` .

I think it is a good pattern to do that for all exceptions. It does not harm and leads to consistency.

deiferni: I double-checked this functionality an `remain` is not used by the frontend in those cases, that's why i removed it.

Also i slightly disagree with you here 😉 , i think we should only add the remain/proceed flag  to the response when it is actually needed by the frontend. Otherwise i don't see the purpose of having a separate method to add the flag, we could add `{'proceed': False}` to any `error` response. Also it would be better to rely on HTTP status-codes in combination with the jquery `Deferred` object's `done` and `fail` functions instead of using a custom flag here. Keeping the `remain` only where it is used would allow us to successively rewrite the frontend and switch to status codes, and then to remove the `remain` method once all has been rewritten. What do you think?

deiferni:
@jone updated the one suggestion i agree with 😉

opengever/meeting/browser/meetings/agendaitem.py
jone: My thought was that an `Unauthorized` exception may happen unexpectedly on any action. We would like to prevent the frontend from continuing as if it had worked.

But your argument about response codes is an interesting input:
I was actually confused why error handling happens in `done`, but did not realize that this is because all responses were `200 OK`, even when erroneous.

You've added some response codes. This means `done` is not even called, so `proceed` / `remain` is in deed not really necessary here.

So: I agree; I'd also add response status codes to the other errors though, so that all errors are not responsed with 200.

opengever/meeting/browser/meetings/agendaitem.py
```diff
@@ -89,6 +91,43 @@ def return_excerpt():
...
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+
+        except (WrongAgendaItemState, CannotExecuteTransition):
+            return JSONResponse(getRequest()).error(
+                _(u'invalid_agenda_item_state',
+                  default=u'The agenda item is in an invalid state for '
+                           'this action.'),
+                status=401).dump()
```
jone: I'd go with `403` Forbidden here.
`401` is in practice usually used for anonymous requests to indicate that the user should authenticate.

opengever/meeting/browser/meetings/agendaitem.py
```diff
@@ -89,6 +91,43 @@ def return_excerpt():
...
+            return JSONResponse(getRequest()).error(
+                _(u'editing_not_allowed',
+                  default=u'Editing is not allowed.'),
+                status=401).dump()
+
+        except MissingMeetingDossierPermissions:
+            return JSONResponse(getRequest()).error(
+                _('error_no_permission_to_add_document',
+                  default=u'Insufficient privileges to add a '
+                          u'document to the meeting dossier.')).dump()
```
jone: I'd set the status to `403` here.

opengever/meeting/browser/meetings/agendaitem.py
```diff
@@ -89,6 +91,43 @@ def return_excerpt():
...
+
+        except MissingAdHocTemplate:
+            return JSONResponse(getRequest()).error(
+                    _('missing_ad_hoc_template',
+                      default=u"No ad-hoc agenda-item template has been "
+                              u"configured.")
+                ).remain().dump()
```
jone: This is a configuration problem on the server side, therefor the status code must be `5xx`.
I propose to use `501 Not Implemented`.

jone: Did you verify if the `.remain()` is really required here? 🤔

opengever/meeting/browser/meetings/agendaitem.py
deiferni: we've been using `Unauthorized` incorrectly for some time now, you're right, good time to change it!

opengever/meeting/browser/meetings/agendaitem.py
deiferni: nice catch, after switching to properly using status codes we can remove the `remain` here, too.

deiferni:
@jone updated as we discussed, please re-review.

PR #3556, opened 2017-10-27
Sort committees in committee vocabulary

tarnap:
This vocabulary is used to display the committees in the form to add
proposals.
These committees should be sorted by their title.

Before:
bildschirmfoto 2017-10-19 um 11 24 43

Now:
sorted_committees_in_proposal_add_form_word

I split the PR into two commits, the first one contains the test (which fails), the second one contains the changes letting the test pass.
As soon as this PR is approved, the two commits are going to be squashed.

Resolves https://github.com/4teamwork/gever/issues/141

### DIFF
--- a/opengever/meeting/tests/test_committee_vocabulary.py
+++ b/opengever/meeting/tests/test_committee_vocabulary.py
@@ -1,0 +1,17 @@
+from opengever.meeting.model import Committee
+from opengever.meeting.vocabulary import CommitteeVocabulary
+from opengever.testing import IntegrationTestCase
+
+
+class TestCommitteeVocabulary(IntegrationTestCase):
+
+    def test_get_committees_returns_list_sorted_by_title(self):
+        simple_vocabulary = CommitteeVocabulary()
+        titles_from_vocabulary = [term.title
+                                  for term in simple_vocabulary(context=None)]
+        unsorted_titles = [committee.title for committee in Committee.query.all()]
+        sorted_titles = [u'Kommission f\xfcr Verkehr',
+                         u'Rechnungspr\xfcfungskommission']
+
+        self.assertNotEquals(unsorted_titles, titles_from_vocabulary)
+        self.assertEquals(sorted_titles, titles_from_vocabulary)
--- a/opengever/meeting/vocabulary.py
+++ b/opengever/meeting/vocabulary.py
@@ -19,16 +19,15 @@ from zope.schema.vocabulary import SimpleVocabulary
 class CommitteeVocabulary(object):
 
     def __call__(self, context):
-        terms = []
-
-        for committee in self.get_committees():
-            terms.append(SimpleTerm(value=committee,
-                                    token=committee.committee_id,
-                                    title=committee.title))
-        return SimpleVocabulary(terms)
+        return SimpleVocabulary([
+            SimpleTerm(value=committee,
+                       token=committee.committee_id,
+                       title=committee.title)
+            for committee in self.get_committees()
+        ])
 
     def get_committees(self):
-        return Committee.query.all()
+        return Committee.query.order_by('title').all()
 
 
 class ActiveCommitteeVocabulary(CommitteeVocabulary):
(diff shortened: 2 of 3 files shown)


opengever/meeting/vocabulary.py
```diff
@@ -19,13 +19,15 @@
...
 
     def __call__(self, context):
-        terms = []
-
-        for committee in self.get_committees():
-            terms.append(SimpleTerm(value=committee,
-                                    token=committee.committee_id,
-                                    title=committee.title))
-        return SimpleVocabulary(terms)
+        sorted_committees = sorted(
```
deiferni: The correct approach here would be to sort in `SQL` already, you can to that by adding an `order_by` to the query in `get_committees`.

opengever/meeting/tests/test_committee_vocabulary.py
```diff
@@ -0,0 +1,10 @@
+from opengever.meeting.vocabulary import CommitteeVocabulary
+from opengever.testing import IntegrationTestCase
+
+
+class TestCommitteeVocabulary(IntegrationTestCase):
+
+    def test_committees_are_sorted_by_committee_title(self):
+        simple_vocabulary = CommitteeVocabulary()
+        titles = [term.title for term in simple_vocabulary(context=None)]
+        self.assertEquals(sorted(titles), titles)
```
deiferni: how many committees are in the fixture? i think explicitly listing the expected titles would be better here.

tarnap: There are only two.
I will list their titles.

opengever/meeting/vocabulary.py
tarnap: ah! ok! Will do!

opengever/meeting/vocabulary.py
```diff
@@ -19,13 +19,15 @@
...
+            self.get_committees(),
+            cmp=lambda c1, c2: cmp(c1.title, c2.title))
```
jone: I'm aware that this will be replaced by SQL, but here my input anway:
Using a sort key lambda would be cleaner here:
```python
sorted([], key=lambda committee: committee.title)
```
Or use an attrgetter:
```python
from operator import attrgetter
sorted([], attrgetter('title'))
```
😉

opengever/meeting/tests/test_committee_vocabulary.py
deiferni: ok, i would also recommend to add an assertion at the beginning of the test that documents that the order in the vocabulary is different from the default database result

opengever/meeting/vocabulary.py
tarnap: operator!

tarnap:
@jone, @deiferni: this is ready for re-review!

deiferni:
as discussed:
- use string literals in tests
- black-box test the whole vocabulary

tarnap:
@deiferni:
- string literals in tests Ahoi!
- black-box test the whole vocabulary Ahoi!

opengever/meeting/vocabulary.py
```diff
@@ -19,16 +19,15 @@
...
+        return SimpleVocabulary([
+            SimpleTerm(value=committee,
+                       token=committee.committee_id,
+                       title=committee.title)
+            for committee in self.get_committees()
+        ])
 
     def get_committees(self):
-        return Committee.query.all()
+        return Committee.query.order_by('title').all()
```
Rotonen: Should we add an index on `committees.title` since this is written to less than it is read from in order to avoid the constant sequential scans over the titles the ordering requires?

deiferni: @Rotonen fair point, there is indeed no index for the title yet. we can expect that there will be "thousands" of proposals in a system so an index should be justified 🤔 . @tarnap could you also add an upgrade-step to enable an index to the title column?

tarnap: @Rotonen: Good idea! This is called every time the proposals are listed, and there are a  few views listing proposals.

PR #3561, opened 2017-10-30
Allow filtering/sorting by proposal title.

deiferni:
Allow to filter and sort proposals by their title/submitted_title in proposal listings. The following listings are improved:

- my proposals
- dossier proposal listing
- committee submitted proposal listing

This is implemented by once again syncing the proposal and submitted proposal titles to sql. The sql table should be thought of as a read-only index though, the primary data remains in plone.

The migration consists of two steps, first the model is changed then data is migrated.
_Theoretically this would run on multiple plone sites even though in practice it is not used that way :smile:._

Closes https://github.com/4teamwork/gever/issues/66.

⚠️  Unlocks https://github.com/4teamwork/gever/issues/120.

### DIFF
--- a/opengever/base/sqlsyncer.py
+++ b/opengever/base/sqlsyncer.py
@@ -1,0 +1,58 @@
+from Acquisition import aq_base
+from Products.CMFPlone.interfaces import IPloneSiteRoot
+from zope.component import getUtility
+from zope.intid.interfaces import IIntIds
+from zope.lifecycleevent.interfaces import IObjectRemovedEvent
+import logging
+
+
+logger = logging.getLogger('opengever.sqlsyncer')
+
+
+class SqlSyncer(object):
+    """Base class to sync plone objects to an sql index."""
+
+    def __init__(self, obj, event):
+        self.obj = obj
+        self.event = event
+
+        if not self.is_uninstalling_plone():
+            self.obj_id = self.get_object_id()
+
+    def is_uninstalling_plone(self):
+        return IObjectRemovedEvent.providedBy(self.event) \
+            and IPloneSiteRoot.providedBy(self.event.object)
+
+    def get_object_id(self):
+        intids = getUtility(IIntIds)
+        try:
+            return intids.getId(self.obj)
+        except KeyError:
+            try:
+                # In some case (remote task updating etc)
+                # only the base_object provides an intid.
+                int_id = intids.getId(aq_base(self.obj))
+                logger.info('Used aq_base(obj) fallback to get intid.')
+                return int_id
+            except KeyError:
+                # The intid event handler didn't create an intid for this
+                # object yet. The event will be fired again after creating the
+                # id.
+                return None
+
+    def sync(self):
+        if self.is_uninstalling_plone():
+            logger.warn(
+                'Synchronisation of object {!r} skipped, '
+                'plone site is being uninstalled.'.format(self.obj))
+            return
+
+        if self.obj_id is None:
+            logger.warn('Synchronisation of object {!r} skipped, '
+                        'no intid exists for this object.'.format(self.obj))
+            return
+
+        return self.sync_with_sql()
+
+    def sync_with_sql(self):
+        raise NotImplementedError()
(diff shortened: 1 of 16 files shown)


deiferni:
(dependency: https://github.com/4teamwork/gever/issues/120)

opengever/core/upgrades/20171027135623_add_title_columns_to_proposal_sql_model/upgrade.py
```diff
@@ -0,0 +1,19 @@
...
+    """Add title columns to proposal sql model.
+    """
+
+    def migrate(self):
+        self.op.add_column(
+            'proposals', Column(
+                'title', String(MAX_TITLE_LENGTH), index=True))
+        self.op.add_column(
+            'proposals', Column(
+                'submitted_title', String(MAX_TITLE_LENGTH), index=True))
```
jone: Shouldn't usually the title for proposal and submitted proposal be the same? 🤔 although I see the problem..

opengever/globalindex/handlers/task.py
```diff
@@ -1,75 +1,33 @@
...
 from opengever.ogds.base.utils import get_current_org_unit
-from Products.CMFPlone.interfaces import IPloneSiteRoot
-from zope.component import getUtility
-from zope.intid.interfaces import IIntIds
-from zope.lifecycleevent.interfaces import IObjectRemovedEvent
 import logging
 
 
-logger = logging.getLogger('opengever.globalindex')
+logger = logging.getLogger('opengever.sqlsyncer')
```
jone: This is still in the `globalindex` package 🤔

opengever/core/upgrades/20171027135623_add_title_columns_to_proposal_sql_model/upgrade.py
deiferni: Yes, it should, but also at the moment we offer the option to edit the submitted proposals, so it could differ. That's why i chose to add two columns at the moment.

opengever/base/sqlsyncer.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def get_object_id(self):
+        intids = getUtility(IIntIds)
+        try:
+            int_id = intids.getId(self.obj)
+        except KeyError:
+            try:
+                # In some case (remote task updating etc)
+                # only the base_object provides an intid.
+                int_id = intids.getId(aq_base(self.obj))
```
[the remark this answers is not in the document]
deiferni: I asked myself the same questions, but this is legacy code, i extracted it only into its own class but left the rest unchanged. I'd rather not start to refactor task-related stuff here, if that is fine for you.

opengever/core/upgrades/20171027135623_add_title_columns_to_proposal_sql_model/upgrade.py
jone: Fair enough 👍

opengever/base/sqlsyncer.py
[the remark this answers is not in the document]
jone: At the point I wrote this I actually wasn't aware that you just have moved it; but I thought I leave the comment so that we can discuss.

I agree with leaving it as it is 👍

opengever/globalindex/handlers/task.py
deiferni: True, will revert.

deiferni: True, will revert.

opengever/base/sqlsyncer.py
```diff
@@ -0,0 +1,57 @@
...
+            self.obj_id = self.get_object_id()
+
+    def is_uninstalling_plone(self):
+        return IObjectRemovedEvent.providedBy(self.event) \
+            and IPloneSiteRoot.providedBy(self.event.object)
+
+    def get_object_id(self):
+        intids = getUtility(IIntIds)
+        try:
+            int_id = intids.getId(self.obj)
```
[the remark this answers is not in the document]
deiferni: hmm well, a minor refactoring should be ok 😉 , will adapt.

deiferni:
@jone updated as you suggested

PR #3573, opened 2017-10-31
Allow adding excerpts headers and suffixes

tarnap:
A sablon template can be defined on committee or committee container
which will be added (if available) to the created excerpts.
If the committee has no templates available, the template of the
committee container is used.

Since the new excerpts do not only contain the content of the agenda
item document, the tests which previously compared the content of the
generated excerpts with the agenda item documents need to be changed.

The import of the MergeDocxExcerptCommand inside the method of the
AgendaItem is required to get rid of circular imports.

vorlagen excerpts

Resolves https://github.com/4teamwork/gever/issues/117

### DIFF
--- a/opengever/meeting/command.py
+++ b/opengever/meeting/command.py
@@ -48,7 +48,7 @@ class MergeDocxExcerptCommand(CreateDocumentCommand):
     """Create or update a merged excerpt word file.
     """
 
-    def __init__(self, context, agenda_item, filename, title, lock_document_after_creation=False):
+    def __init__(self, context, agenda_item, filename, title):
         self.agenda_item = agenda_item
         self.excerpt_protocol_data = ExcerptProtocolData(
             self.agenda_item.meeting,
--- a/opengever/meeting/command.py
+++ b/opengever/meeting/command.py
@@ -61,6 +61,12 @@ class MergeDocxExcerptCommand(CreateDocumentCommand):
             title=title)
 
     def generate_file_data(self):
+        header_template = self.agenda_item.get_excerpt_header_template()
+        suffix_template = self.agenda_item.get_excerpt_suffix_template()
+
+        if header_template is None and suffix_template is None:
+            return self.agenda_item.resolve_document().file.data
+
         with DocxMergeTool() as merge_tool:
             if self.agenda_item.get_excerpt_header_template() is not None:
                 merge_tool.add_sablon(self.get_header_sablon())
--- a/opengever/meeting/command.py
+++ b/opengever/meeting/command.py
@@ -44,6 +44,52 @@ import json
 MIME_DOCX = 'application/vnd.openxmlformats-officedocument.wordprocessingml.document'
 
 
+class MergeDocxExcerptCommand(CreateDocumentCommand):
+    """Create or update a merged excerpt word file.
+    """
+
+    def __init__(self, context, agenda_item, filename, title, lock_document_after_creation=False):
+        self.agenda_item = agenda_item
+        self.excerpt_protocol_data = ExcerptProtocolData(
+            self.agenda_item.meeting,
+            [self.agenda_item])
+        self.filename = filename
+        super(MergeDocxExcerptCommand, self).__init__(
+            context=context,
+            filename=filename,
+            data=None,
+            title=title)
+
+    def generate_file_data(self):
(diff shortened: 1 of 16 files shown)


opengever/meeting/command.py
```diff
@@ -44,6 +44,52 @@
 MIME_DOCX = 'application/vnd.openxmlformats-officedocument.wordprocessingml.document'
 
 
+class MergeDocxExcerptCommand(CreateDocumentCommand):
+    """Create or update a merged excerpt word file.
+    """
+
+    def __init__(self, context, agenda_item, filename, title, lock_document_after_creation=False):
```
deiferni: the `lock_document_after_creation` argument looks unused. i would remove it, we don't want to lock the excerpts when the word meeting feature is enabled.

opengever/meeting/command.py
```diff
@@ -44,6 +44,52 @@
...
+        self.filename = filename
+        super(MergeDocxExcerptCommand, self).__init__(
+            context=context,
+            filename=filename,
+            data=None,
+            title=title)
+
+    def generate_file_data(self):
+        with DocxMergeTool() as merge_tool:
+            if self.agenda_item.get_excerpt_header_template() is not None:
```
deiferni: this will also use the merge tool if there is only one document. may this have any negative side-effects?

opengever/meeting/command.py
```diff
@@ -44,6 +44,52 @@
...
+                merge_tool.add_sablon(self.get_suffix_sablon())
+
+            return merge_tool()
+
+    def get_header_sablon(self):
+        return Sablon(self.agenda_item.get_excerpt_header_template()).process(
+            self.excerpt_protocol_data.as_json())
+
+    def get_suffix_sablon(self):
+        return Sablon(self.agenda_item.get_excerpt_suffix_template()).process(
```
deiferni: you could rewrite these two and have only one method that gets the sablon template passed in. that would have the advantage of not having to resolve the template twice. up to you though 😉 .

opengever/meeting/tests/test_committee_container_word.py
```diff
@@ -65,6 +69,8 @@ def test_can_add_with_paragraph_template(self, browser):
         browser.fill({'Title': u'Sitzungen',
                       'Protocol header template': self.sablon_template,
                       'Protocol suffix template': self.sablon_template,
+                      'Excerpt header template': self.sablon_template,
```
deiferni: i'm a bit confused by these `can_add_with_…` test methods. they `fill` multiple templates, but only `assert` for one of them. i think this may have grown in the wrong direction 😄 .
maybe it would be better to have one test that creates content with all the fields and then asserts for all of them? i do not really see the purpose of testing each field individually in its separate test. also it is a bit expensive for the value it provides as creating a lot of contents makes the tests slower 😢 .

opengever/meeting/command.py
tarnap: You are right! We didn't need that variable, therefore we don't need it as argument either!

opengever/meeting/command.py
tarnap: The `DocxMergeTool` is self explanatory:
It creates a temprorary directory into which the files are copied, merged and the resulting document's data is returned.
Finally the created directory is removed.

What would you do? See if there is a header or suffix template and use the `DocxMergeTool` and if there isn't return the `agenda_item.resolve_document().data`?

opengever/meeting/command.py
tarnap: fair'nough! Done!

opengever/meeting/command.py
deiferni: just a review question 😉 . we don't need to change the code when you're certain it will also work with just the one filed. maybe you could add a test for that if there is none yet? if there are issues what you suggested is exactly what i would do: just use the merge tool if there are two or more files.

opengever/meeting/tests/test_committee_container_word.py
tarnap: fair'nough!
will do!

tarnap:
@deiferni I did all the changes but the last one. Vilech chumi morn no derzue

tarnap:
@deiferni: I refactored the test you asked me to. Please take a look at it!

PR #3574, opened 2017-10-31
Exclude the latest version in the bumblebee versioning warning logic

### LINKED ISSUE #3572: Korrektur Hinweis „Sie sehen sich eine alte Version an“
bildschirmfoto 2017-10-31 um 15 57 24
Der Hinweis „Sie sehen sich eine alte Version an“ im Bumblebee Overlay sollte nicht anzeigt werden, wenn es die aktuelle Version ist.

Siehe https://basecamp.com/2768704/projects/14549453/todos/324768462

Rotonen:
* Exclude the latest version in the bumblebee versioning warning logic

Closes #3572

### DIFF
--- a/opengever/bumblebee/browser/overlay.py
+++ b/opengever/bumblebee/browser/overlay.py
@@ -43,6 +43,7 @@ import os
 class BumblebeeBaseDocumentOverlay(ActionButtonRendererMixin):
     """Bumblebee overlay for base documents.
     """
+
     version_id = None
 
     def __init__(self, context, request):
@@ -141,7 +142,7 @@ class BumblebeeBaseDocumentOverlay(ActionButtonRendererMixin):
             # Bumblebee will use a placeholder filename
             return None
 
-        filename, extenstion = os.path.splitext(self.get_file().filename)
+        filename = os.path.splitext(self.get_file().filename)[0]
         return u'{}.pdf'.format(filename)
 
     def _is_checkout_and_edit_available(self):
--- a/opengever/bumblebee/browser/overlay.py
+++ b/opengever/bumblebee/browser/overlay.py
@@ -50,6 +50,10 @@ class BumblebeeBaseDocumentOverlay(ActionButtonRendererMixin):
         self.context = context
         self.request = request
 
+    def is_latest_version(self):
+        """A documentish without versions is always the latest version."""
+        return True
+
     def get_preview_pdf_url(self):
         return bumblebee.get_service_v3().get_representation_url(
             self.context, 'preview')
@@ -176,6 +180,10 @@ class BumblebeeMailOverlay(BumblebeeBaseDocumentOverlay):
     """Bumblebee overlay for base mails.
     """
 
+    def is_latest_version(self):
+        """Mails are not versionable."""
+        return True
+
     def get_open_as_pdf_url(self):
         return u'{}/bumblebee-open-pdf?filename={}'.format(
             self.context.absolute_url(),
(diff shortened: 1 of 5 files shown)


phgross:
@Rotonen the test are not happy yet

Rotonen:
@phgross sorry about that, deemed ~1h without a notification was good to do a pull request. For some reason the failure notification only came at 10am today.

Some fun with when the overlay view flags or does not flag something with `IVersionedContextMarker`, looking into it.

Rotonen:
@phgross ready for rereview.

opengever/bumblebee/browser/overlay.py
```diff
@@ -141,7 +142,7 @@ def _get_pdf_filename(self):
             # Bumblebee will use a placeholder filename
             return None
 
-        filename, extenstion = os.path.splitext(self.get_file().filename)
+        filename, _ = os.path.splitext(self.get_file().filename)
```
deiferni: why `_`? this is what we usually use as name for the message-factories. i would not use if for unused variables here. could you rename?

opengever/bumblebee/browser/overlay.py
```diff
@@ -241,7 +242,9 @@ def __call__(self):
             raise NotFound
 
         overlay_context = self.context
-        version_id = self._get_version_id()
+
+        request_version_id = self._get_version_id()
+        version_id = None if self._is_latest_version() and not request_version_id else request_version_id
```
deiferni: why is it necessary to omit adding the `IVersionedContextMarker` here 🤔 ?

opengever/bumblebee/browser/templates/bumblebeeoverlay.pt
```diff
@@ -34,7 +34,7 @@
     <div class="info-viewlets">
       <div tal:replace="structure view/overlay/render_checked_out_viewlet" />
       <div tal:replace="structure view/overlay/render_lock_info_viewlet" />
-      <div tal:condition="view/is_versioned">
+      <div tal:condition="not: view/_is_latest_version">
```
deiferni: i'd make this non-protected when used in the template

opengever/bumblebee/browser/overlay.py
Rotonen: Squelching an unused variable warning. That is an unfortunate collision.

The extension is not used for anything and that is clearer than slapping `[0]` at the end.

opengever/bumblebee/browser/overlay.py
Rotonen: I have no greater insight into the history here: are there bumblebee-displayables which cannot be versioned (in an inbox or somesuch)?

deiferni: yes, mails for example. but i also do not know what `IVersionedContextMarker` is used for … so you'd best double-check?

Rotonen: So, completely removing `IVersionedContextMarker` results in only two test failures:

https://github.com/4teamwork/opengever.core/blob/04f1f640f4f434bd587837049afaab4d4375d074/opengever/bumblebee/tests/test_overlay_view.py#L151

and

https://github.com/4teamwork/opengever.core/blob/04f1f640f4f434bd587837049afaab4d4375d074/opengever/bumblebee/tests/test_overlay_view.py#L283

now both have the file action button for opening the document as PDF.

Is that even wrong? If not, we could just chuck away the whole adapter?

Rotonen: Found the root cause after a while: documents no longer have an initial version unless edited with an editor or otherwise manipulated so the premises of the logic of both the tests and the logic being tested here are wrong.

Drawing board time.

Rotonen: Moved the warning display logic to happen per adapter.

PR #3575, opened 2017-11-01
Fix textfilter in sqlsource table listings for oracle backends.

phgross:
Oracle raises an "ORA-00906: missing left parenthesis" erorr when
casting a integer column to a string and querying it with ilike.

Therefore I add a special handling for oracle backends, which solves
the issue https://sentry.4teamwork.ch/sentry/onegov-gever/issues/3749/
when filtering in task listings like the MyTasks tab.

Not valid Query 
```
SELECT tasks.id, tasks.admin_unit_id [...]
FROM tasks
WHERE tasks.responsible IN ('hugo.boss') AND (lower(CAST(tasks.sequence_number AS VARCHAR)) LIKE lower('%phip%')) ORDER BY created DESC
```

Valid query
```
SELECT tasks.id, tasks.admin_unit_id [...]
FROM tasks
WHERE tasks.responsible IN ('hugo.boss') AND (tasks.sequence_number LIKE lower('%phip%')) ORDER BY created DESC
```

### DIFF
--- a/opengever/tabbedview/sqlsource.py
+++ b/opengever/tabbedview/sqlsource.py
@@ -1,4 +1,5 @@
 from ftw.table.interfaces import ITableSource
+from opengever.base.model import is_oracle
 from opengever.tabbedview import GeverTableSource
 from opengever.tabbedview.interfaces import IGeverTableSourceConfig
 from sqlalchemy import or_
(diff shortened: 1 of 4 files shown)


opengever/tabbedview/sqlsource.py
```diff
@@ -90,13 +91,21 @@ def extend_query_with_textfilter(self, query, text):
 
                 expressions = []
                 for field in self.searchable_columns:
-                    if not issubclass(field.type.python_type, basestring):
+                    if self.needs_cast(field):
                         field = cast(field, String)
                     expressions.append(field.ilike(term))
```
deiferni: you could extract this and the `needs_cast` method into a `cast_to_string` function and then just write:

`expressions.append(cast_to_string(field).ilike(term))`

`needs_cast` may be misleading since it returns whether a cast to `string` is necessary. bonus points for adding some unit tests for `cast_to_string`.

phgross:
@deiferni updated ...

PR #3583, opened 2017-11-02
Remove submitted proposal noword fields

tarnap:
Before:
submitted_proposal_with_noword_fields

And now:
submitted_proposal_without_noword_fields

Resolves https://github.com/4teamwork/gever/issues/144

### DIFF
--- a/opengever/meeting/browser/proposalforms.py
+++ b/opengever/meeting/browser/proposalforms.py
@@ -79,6 +79,11 @@ class SubmittedProposalEditForm(FieldConfigurationMixin,
     fields = field.Fields(SubmittedProposal.model_schema, ignoreContext=True)
     content_type = SubmittedProposal
 
+    def updateFields(self):
+        super(SubmittedProposalEditForm, self).updateFields()
+        if is_word_meeting_implementation_enabled():
+            self.fields = self.fields.omit('considerations', 'excerpts')
+
     def updateWidgets(self):
         super(SubmittedProposalEditForm, self).updateWidgets()
         self.widgets['relatedItems'].mode = HIDDEN_MODE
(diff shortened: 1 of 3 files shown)


opengever/meeting/browser/proposalforms.py
```diff
@@ -83,6 +83,11 @@ def updateWidgets(self):
         super(SubmittedProposalEditForm, self).updateWidgets()
         self.widgets['relatedItems'].mode = HIDDEN_MODE
 
+    def updateFields(self):
+        super(SubmittedProposalEditForm, self).updateFields()
+        if is_word_meeting_implementation_enabled():
+            self.fields = self.fields.omit('considerations', 'excerpts')
```
deiferni: couldn't you add these fields in `FieldConfigurationMixin`?

tarnap:
@deiferni: Changes made as requested.

opengever/meeting/browser/proposalforms.py
```diff
@@ -50,6 +50,7 @@ def updateFields(self):
         if is_word_meeting_implementation_enabled():
             map(self.omit_field, textfields)
             move(self, 'title', before='*')
+            self.fields = self.fields.omit('considerations', 'excerpts')
```
deiferni: but … why not add it to `textfields`? everything in that tuple is omitted already 😉 .

tarnap:
@deiferni: changes made as requested: jitz schiessts! wieso?

opengever/meeting/browser/proposalforms.py
tarnap: This seemed to work well... The tests fail since i added 'considerations' and 'excerpts' to the list of textfields.

opengever/meeting/browser/proposalforms.py
jone: Are those fields only in submitted proposals, not in (regular) proposals?
If so, this is the reason for the test failures: `FieldConfigurationMixin` is applied to both, so it should be put there but in the actual form.

deiferni:
as discussed, adding it to `textfields` does not seem to be a good idea since these fields are submitted-proposal specific.

tarnap:
@deiferni: I added a second commit, which makes the "considerations" field a trix-field.

PR #3584, opened 2017-11-02
Automatically update gever version in versions.cfg.

deiferni:
With this PR we automatically update the released version in `versions.cfg` while creating a release with `zest.releaer`.

- We use https://github.com/zestsoftware/zest.releaser `prerelease` hook to pin the to be released gever version in `versions.cfg` at the same time `setup.py` and `HISTORY.txt` are updated, i.e. when creating the `Preparing release [num]` commit.
- We use the `postrelease` hook to remove the pinning again, i.e. when creating the `Back to development: [num++]` commit.

Example commits:
- prerelease:  https://github.com/4teamwork/opengever.core/commit/23977156bbcff405ebcaae15dafcddee616821f5
- postrelease: https://github.com/4teamwork/opengever.core/commit/935299224b379e2a00c4cd5ee1bb56328b94f2a4

⚠️  todo once approved: 
- [x] update wiki/internal docs

### DIFF
--- a/opengever/core/releaser.py
+++ b/opengever/core/releaser.py
@@ -1,6 +1,7 @@
 import logging
 import os.path
 import sys
+import re
 
 
 logger = logging.getLogger("opengever.releaser")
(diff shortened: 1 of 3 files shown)


opengever/core/releaser.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def _make_versions_file_path(data):
+    """Return versions file path. Validate file exists."""
+
+    versions_file_path = os.path.join(data["reporoot"], VERSIONS_FILE_NAME)
+
+    if not os.path.exists(versions_file_path):
+        logger.critical("Could not find 'versions.cfg' at '{}'.".format(
+            versions_file_path))
```
jone: If I understand this correctly, this would print:
`Could not find 'versions.cfg' at '/Users/jone/projects/packages/opengever.core/versions.cfg'.`
So: either print filename and directory or print fullpath, but not both.

deiferni: `True`, changed to only print dir and filename.

deiferni:
@jone updated, also added CL entry.

PR #3586, opened 2017-11-06
Squelch an SQLAlchemy stderr bleed from a test

Rotonen:
I'm intending on moving to asserting on unclean stderr from a test batch in the CI runs and this is the last one standing in the way.

https://ci.4teamwork.ch/builds/121918/tasks/187756

Querying on an empty notification role target group spits out (retouched for readability):

```
eggs/SQLAlchemy-1.1.10-py2.7-linux-x86_64.egg/sqlalchemy/sql/default_comparator.py:161: 
SAWarning:
The IN-predicate on "notifications.userid" was invoked with an empty sequence.
This results in a contradiction, which nonetheless can be expensive to evaluate.
Consider alternative strategies for improved performance.
```

This breaks the test assertion (to explicitly test for no one having being notified in case no one has registered). Suggestions?

### DIFF
--- a/opengever/activity/model/query.py
+++ b/opengever/activity/model/query.py
@@ -8,12 +8,14 @@ from opengever.ogds.models.query import BaseQuery
 
 
 class ActivityQuery(BaseQuery):
-    pass
+    """Provide accessors to activity."""
+
 
 Activity.query_cls = ActivityQuery
 
 
 class NotificationQuery(BaseQuery):
+    """Provide accessors to notifications."""
 
     def by_user(self, userid):
         return self.filter_by(userid=userid)
@@ -32,14 +34,17 @@ Notification.query_cls = NotificationQuery
 
 
 class ResourceQuery(BaseQuery):
+    """Provide accessors to resources."""
 
     def get_by_oguid(self, oguid):
         return self.filter_by(oguid=oguid).first()
+
 
 Resource.query_cls = ResourceQuery
 
 
 class NotificationDefaultQuery(BaseQuery):
+    """Provide a default accessors to activities."""
 
     def is_dispatch_needed(self, dispatch_setting, kind):
         setting = self.filter_by(kind=kind).first()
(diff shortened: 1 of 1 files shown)


opengever/activity/tests/test_model.py
```diff
@@ -59,7 +62,7 @@ def test_by_subscription_role_query(self):
 
         notifications = Notification.query.by_subscription_roles(
```
phgross: Does  the following assertion raises the same warning?
```
self.assertEquals(
    0,
    Notification.query.by_subscription_roles([TASK_RESPONSIBLE_ROLE], activity).count())
```

Rotonen: No, as the issue is `Notification.query.by_subscription_roles()` having been passed in a role with no registered watchers.

deiferni: i would attempt to avoid issuing such queries at all, this could mean that in this case we just return an empty list in `by_subscription_roles` if we don't have any `subscriptions`. 

we should only rely that the thing being returned by that method is an `iterable`, so the change should be fine 🤔

Rotonen: It returns an empty list, that is not the issue at hand. It is just that the query generated in that case has an empty `IN` clause by the way SQLAlchemy works. So either we figure out the testing or we actually fix the underlying query for that case.

deiferni: that is exactly what i'm saying though 😄 , if you return when no `subscriptions` have been returned the following code with the `in_` won't be called, right?

```python
        return self.filter_by(activity_id=activity.id).filter(
            Notification.userid.in_(user_ids))
```

deiferni: oh `by_resource_and_role` has also an `in_` … maybe abort even earlier then.

Rotonen:
Addressed the queries. Ready for rereview.

PR #3587, opened 2017-11-07
Fix ie11 action buttons, don't put title into action link.

deiferni:
This allows us to not hide the link on ie11 which caused issues with displaying the fa-icon.

⚠️  requires https://github.com/4teamwork/plonetheme.teamraum/pull/596.

now correctly displays the document actions on ie11:
screen shot 2017-11-07 at 10 51 25

Fixes https://github.com/4teamwork/gever/issues/146.


deiferni:
yeah unfortunately that is a trade-off for now. there may be a better way to do this of course …

PR #3588, opened 2017-11-07
Port mtest to multiprocessing and split out layers

Rotonen:
Current test timings on my laptop with this layer split:

```
 0.033%                     policy               02 seconds
 0.037%                     sphinx               02 seconds
 0.037%            policytemplates               02 seconds
 0.144%                    testing               09 seconds
 0.144%                        api               09 seconds
 0.150%                    ech0147               10 seconds
 0.182%               contentstats               12 seconds
 0.460%                      debug               30 seconds
 0.616%               officeatwork               40 seconds
 0.654%                   portlets               43 seconds
 0.670%                    sharing               44 seconds
 0.691%                 repository               45 seconds
 0.727%                     bundle               48 seconds
 0.735%                    locking               48 seconds
 0.737%                       core               48 seconds
 0.759%                      quota               50 seconds
 0.766%             advancedsearch               50 seconds
 0.796%                      setup               52 seconds
 0.840%                globalindex               55 seconds
 0.938%                       ogds    01 minutes 02 seconds
 1.023%              usermigration    01 minutes 07 seconds
 1.128%                   activity    01 minutes 14 seconds
 1.302%            officeconnector    01 minutes 26 seconds
 1.310%                    journal    01 minutes 26 seconds
 1.352%              tasktemplates    01 minutes 29 seconds
 1.363%                    private    01 minutes 29 seconds
 1.414%                      latex    01 minutes 33 seconds
 1.478%                 tabbedview    01 minutes 37 seconds
 1.500%                      trash    01 minutes 38 seconds
 1.685%             examplecontent    01 minutes 51 seconds
 1.929%                    contact    02 minutes 07 seconds
 2.459%                      inbox    02 minutes 41 seconds
 2.652%                  bumblebee    02 minutes 54 seconds
 2.948%                    meeting    03 minutes 14 seconds
 3.377%                       mail    03 minutes 42 seconds
 3.796%    core:functional:zserver    04 minutes 09 seconds
 5.123%                disposition    05 minutes 36 seconds
 7.255%                       base    07 minutes 56 seconds
 7.311%                       task    07 minutes 60 seconds
 7.987%                    dossier    08 minutes 44 seconds
 8.434%                   document    09 minutes 14 seconds
 9.170%           core:integration    10 minutes 02 seconds
13.886%               MeetingLayer    15 minutes 12 seconds

Total:     01 hours 49 minutes 25 seconds
Wallclock:          38 minutes 43 seconds
```

What:
* No longer depends on Python 3
* Splits out the slow layers to run as separate processes
  * This thins down the slowest modules to their leanest
  * Reduces on duplicated layer setup time across workers
* Always run the ZServer layer separate of everything else
  * Lets us get rid of the port management for the most part
  * The layer separation is a greedy match (`functional` would match `functional:zserver`)
* Negates these layers from the other runs
* Changes the list of known slow modules
  * We also now have only one tier of slow modules instead of two
  * Splits off the slow modules onto their own processes
* Clumps the remaining modules onto a single batch
  * This is now the batch to finish last
* The meeting layer is now the second slowest batch
* Logs
  * Nicer output
* Has a debug mode for local running
  * More information
  * Early failure prints
* Concurrency control
  * Per default uses (up to) `system threads // 2 - 1` processes
  * Can be overridden by `MTEST_PROCESSORS`
  * Both can be overridden with the new `-j` flag
* Has rough CLI documentation built in
  * `-h`

Gains:
* 12min runs
* Everything ported off the functional tests directly lands us wallclock time savings
  * After a bit this will be true of the MeetingLayer tests instead
  * We can then decide whether to tackle that or change the module list to accommodate
* When the integration tests are the longest run it can easily be chunked over multiple workers
* Also trips up on an unclean stderr
* 5min .. 8min runs should be possible
  * Pushing any further would require the ability to chunk tests within a layer

### DIFF
--- a/bin/mtest
+++ b/bin/mtest
@@ -1,146 +1,322 @@
-#!/usr/bin/env python3
-
-# Runs the tests in configurably many subprocesses.
-
+#!/usr/bin/env python
+"""A concurrent wrapper for timing xmltestrunner tests for opengever.core."""
+from __future__ import print_function
+from argparse import ArgumentParser
 from copy import copy
 from fnmatch import fnmatch
+from logging import DEBUG
+from logging import Formatter
+from logging import getLogger
+from logging import INFO
+from logging import StreamHandler
+from logging.handlers import MemoryHandler
+from multiprocessing import cpu_count
+from multiprocessing import Pool
+from os import environ
+from os import killpg
+from os import path
+from os import setpgrp
+from os import unlink
+from os import walk
+from signal import SIGINT
+from signal import SIGKILL
+from signal import signal
+from subprocess import PIPE
+from subprocess import Popen
+from time import time
 
-import concurrent.futures
-import os
-import subprocess
-import sys
+
+def humanize_time(seconds):
+    """Humanize a seconds based delta time.
+
+    Only handles time spans up to weeks for simplicity.
+    """
+    minutes, seconds = divmod(seconds, 60)
+    hours, minutes = divmod(minutes, 60)
+    days, hours = divmod(hours, 24)
+    weeks, days = divmod(days, 7)
+
+    seconds = int(round(seconds))
+    minutes = int(round(minutes))
+    hours = int(round(hours))
+    days = int(round(days))
+    weeks = int(round(weeks))
+
+    output = []
+
+    if weeks:
+        output.append('{:02d} weeks'.format(weeks))
+    if days:
+        output.append('{:02d} days'.format(days))
+    if hours:
+        output.append('{:02d} hours'.format(hours))
+    if minutes:
+        output.append('{:02d} minutes'.format(minutes))
+
+    output.append('{:02d} seconds'.format(seconds))
+
+    return ' '.join(output)
 
 
-os.environ['PYTHONDONTWRITEBYTECODE'] = '1'
+def setup_termination():
+    # Set the group flag so that subprocesses will be in the same group.
+    setpgrp()
 
-PROCESSORS = int(os.environ.get('MTEST_PROCESSORS', '4'))
+    def terminate(signum, frame):
+        # Kill the group (including main process) on terminal signal.
+        killpg(0, SIGKILL)
 
-BUILDOUT_PATH = os.path.abspath(os.path.join(__file__, '..', '..'))
-OPENGEVER_PATH = os.path.join(BUILDOUT_PATH, 'opengever')
+    signal(SIGINT, terminate)
+
+
+def get_concurrency():
+    # Counter system congestion and hyperthreading, FWIW
+    concurrency = cpu_count() // 2 - 1
+
+    # Make sure we use at least one core
+    if concurrency < 1:
+        concurrency = 1
+
+    return concurrency
+
+
+def create_test_run_params():
+    test_run_params = []
+
+    # Manually ordered per metrification
+    standalone_modules = (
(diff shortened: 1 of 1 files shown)


Rotonen:
We're green. Ready for review.

I recommend having a local test run or two and also verifying it catches some intentional test failure you throw at it.

It does not propagate SIGINT properly / xmltestrunner does not always listen for `SIGINT` / xmltestrunner has race conditions getting stuck upon a `SIGINT` so one will have to `SIGTERM`, `SIGKILL` or `killall python` to reliably get rid of the children as well. CI Governor handles this and nothing gets stuck on the CI server when one terminates a test run.

Rotonen:
@jone yes, I also tried to achieve that, but that is actually tougher than it superficially seems as xmltestrunner does not always listen to `SIGINT` / it does in an unclean way and gets stuck (especially on the ZServer layer worker). Also the main runner becomes a bit confusing looking as you have to use `map_async`, a hack with `SIG_IGN` to detach and reattach `SIGINT` at the pool creation time and also to use a callback to collect the results. I rather like how clean and simple it looks like currently.

As `SIGTERM` seems to work better, we could just let it suicide upon a `SIGINT` with `kill -TERM -- -$PID` (minus PID means GPID means kill group) as the main process does nothing sensitive, but thats also quite dirty.

I recognize it'd be nice, but I'll gladly leave this for later as `killall python` (or any eleganter equivalent) is fine on a developer machine and the CI kills the whole process group upon terminating a job.

deiferni:
so can we merge? anything speeding up tests is very welcome!

jone:
@Rotonen how about just killing the process group on terminal signal? Seems to work quite well:
```diff
diff --git a/bin/mtest b/bin/mtest
index 5f25bda70..a7b39f164 100755
--- a/bin/mtest
+++ b/bin/mtest
@@ -246,6 +246,19 @@ def main():
     handle_results(results)
 
 
+def setup_termination():
+    import os
+    import signal
+    # Set the group flag so that subprocesses will be in the same group:
+    os.setpgrp()
+
+    def terminate(signum, frame):
+        # Kill the group (including main process) on terminal signal.
+        os.killpg(0, signal.SIGKILL)
+
+    signal.signal(signal.SIGINT, terminate)
+
+
 # Having the __main__ guard is necessary for multiprocessing.Pool().
 if __name__ == '__main__':
     # Globals
@@ -302,4 +315,5 @@ if __name__ == '__main__':
 
     logger.addHandler(memory_handler)
 
+    setup_termination()
     main()
```

I find it quite annoying when Ctrl-C does not work 😜

Rotonen:
@jone this is what I said earlier as the dirty option, `kill -TERM -- -$PID`, but I'd rather properly propagate `SIGINT` across the tree, timeout, propagate `SIGTERM` across the tree, timeout, propagate `SIGKILL` across the tree.

I suppose there are no on-disk fragments or partials which can bite us in the arse here so I've just implemented your swift murder scheme as xmltestrunner seems inherently misbehaving throughout the Plone stack (`bin/alltests` on `buildout.coredev` also does not handle `SIGINT` very well or at all when you use the `-j` parametre, but the ones on ZODB and Zope do behave).

Rotonen:
So the last remaining hurdle is whether or not the local usability is to the liking of @lukasgraf ?

PR #3593, opened 2017-11-08
Notification badge as a separate channel

phgross:
This ist the first part of  refactorings on the activity modul to support personal notification settings (see OGIP 29 fore more details).

- Move NotificationDispatcher base class to a modul dispatcher.py
- Remove boolean `mail_notification` column from defaults table. (Currently the mail notification can be configured in two ways. With the boolean flag `mail_notification` and per roles with
`mail_notification_roles`, which makes the whole notification default configuration unnecessary complicated. Therefore we drop the boolean column. All configuration possibilities can be still done with the `mail_notification_roles` setting.
- Add badge boolean column to notifications
- Add badge notifications settings including an upgradestep which enables badge notification for all activities.
- Add badge dispatcher wich sets the badge flag.
- Only display notification with enabled badge flag in the badge viewlet.

All these changes does not have impacts on the current functionality


opengever/activity/center.py
```diff
@@ -155,14 +155,17 @@ def get_notification(self, notification_id):
...
+                           badge_only=False):
 
         order = desc if sort_reverse else asc
         query = Notification.query
         if userid:
             query = query.by_user(userid)
 
         query = query.join(Notification.activity)
+        if badge_only:
+            query = query.filter(Notification.badge == True)  # noqa
```
deiferni: should `NotificationTableSource` also only display badged notifications?

opengever/activity/center.py
```diff
@@ -131,15 +131,15 @@ def create_notifications(self, activity):
     def get_users_notifications(self, userid, only_unread=False, limit=None):
         query = Notification.query.by_user(userid)
         if only_unread:
-            query = query.filter(Notification.is_read == False)
+            query = query.filter(Notification.is_read == False)  # noqa
```
deiferni: i think we should use Notification.is_read.is_(False) here

opengever/activity/center.py
```diff
@@ -131,15 +131,15 @@ def create_notifications(self, activity):
...
 
         query = query.join(
             Notification.activity).order_by(desc(Activity.created))
         return query.limit(limit).all()
 
     def count_users_unread_notifications(self, userid):
         query = Notification.query.by_user(userid)
-        return query.filter(Notification.is_read == False).count()
+        return query.filter(Notification.is_read == False).count()  # noqa
```
deiferni: same here, we should use `is_`

opengever/activity/center.py
phgross: Hmm no i don't think so. The notification listing should be a list of all notifications/activities doesn't matter over which channel the users has been notified.

opengever/activity/dispatcher.py
```diff
@@ -0,0 +1,55 @@
...
+        for notification in notifications:
+            try:
+                self.dispatch_notification(notification)
+            except ConflictError:
+                raise
+
+            except Exception:
+                not_dispatched.append(notifications)
+                tcb = ''.join(traceback.format_exception(*sys.exc_info()))
+                logger.error('Exception while dispatch activity '
```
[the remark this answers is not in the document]
phgross: Good point, but I'll fix that in a separate PR (I just moved the file).I opened a Issue (#3602)

opengever/activity/model/notification.py
```diff
@@ -21,6 +21,7 @@ class Notification(Base):
     activity = relationship("Activity", backref="notifications")
 
     is_read = Column(Boolean, default=False, nullable=False)
+    badge = Column(Boolean, default=False, nullable=False)
```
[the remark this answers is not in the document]
phgross: @deiferni  Hmm `is_visible` is to unspecific in my opinion. What I plan is something like that 
bildschirmfoto 2017-11-13 um 17 29 16

What about `notified_by_badge` ?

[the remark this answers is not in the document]
deiferni: acutally after having read the whole PR it is pretty clear to me what this is 😉 , counter-proposal: `is_badge`

opengever/activity/dispatcher.py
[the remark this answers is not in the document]
deiferni: i hate when i'm blamed for code i just moved 😄 , separate issue is 👍  ✨

opengever/activity/model/notification.py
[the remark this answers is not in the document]
phgross: 😄 ok, I'll rename it to `is_badge`

opengever/activity/center.py
deiferni: What i thought is, we display a link to "All Notifications" from the notification icon dropdown in the toolbar. The dropdown only lists (the last N) *badge*-notifications. When i click the link i would expect that it behaves like a `more` link, i.e. that i see all *badge*-notifications. I don't expect to see notifications that were dispatched e.g. via mail, or at least not in the default view.

Also (please correct me if i don't remember the implementation correctly) i think we always create the `Notification`, then check if we should dispatch it. With the upcoming pull-request https://github.com/4teamwork/opengever.core/pull/3594 where i can user-define what notifications i want to see would this mean that i can see unsubscribed notifications showing up in the listing? If that is the case i think that may be confusing.

We may have to rethink how/when a `Notification` database entry is created? 🤔 ☕️ ?

phgross: You're right, for now it makes more sense to display only the notifications which are displayed in the badge.

phgross:
@deiferni updated according to your Feedback.

PR #3594, opened 2017-11-08
Allow personal notification settings

phgross:
Add new model NotificationSetting, which stores the personal notification settings, in the same manner than we have stored the settings for the complete installation. The NotificationDispatchers checks first for a personal setting and uses the notification_defaults as a fallback when deciding if a notification has to be dispatched ... 

The form, where a user can add and edit those personal settings will follow in a separate PR. 

This is a part of OGIP 29

### DIFF
--- a/opengever/activity/dispatcher.py
+++ b/opengever/activity/dispatcher.py
@@ -1,8 +1,12 @@
+from opengever.activity.model import NotificationDefault
+from opengever.activity.model import Subscription
+from opengever.activity.model import Watcher
+from opengever.activity.model.settings import NotificationSetting
+from opengever.base.model import create_session
+from ZODB.POSException import ConflictError
 import logging
 import sys
 import traceback
-from opengever.activity.model import NotificationDefault
-from ZODB.POSException import ConflictError
 
 
 logger = logging.getLogger('opengever.activity')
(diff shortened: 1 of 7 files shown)


deiferni:
for reviewing purposes i have changed the base branch of this PR to `pg_badge_as_a_separate_channel`. this must be changed back to master once the base is merged.

opengever/activity/model/settings.py
```diff
@@ -1,24 +1,19 @@
 from opengever.base.model import Base
+from opengever.ogds.models import USER_ID_LENGTH
+from opengever.ogds.models.user import User  # noqa
```
deiferni: i think you can omit the `# noqa` here?

opengever/activity/model/settings.py
```diff
@@ -40,3 +35,39 @@ def badge_notification_roles(self):
...
+
+    notification_default_id = Column('id', Integer,
+                                     Sequence('notification_defaults_id_seq'),
+                                     primary_key=True)
+    kind = Column(String(50), nullable=False, unique=True)
+
+    def __init__(self, kind, mail_notification_roles=[],
+                 badge_notification_roles=[]):
+        self.kind = kind
+        self.set_mail_notification_roles(mail_notification_roles)
```
deiferni: if you make `set_mail_notification_roles` and `set_badge_notification_roles` a `property.setter` the default constructor can take care of setting these attributes. you should then be able to remove the duplicated constructor.

opengever/activity/dispatcher.py
```diff
@@ -13,22 +15,39 @@ class NotificationDispatcher(object):
...
+            return False
+
+        dispatched_roles = getattr(setting, self.roles_key)
+        subscriptions = Subscription.query.filter_by(
+            resource=notification.activity.resource)
 
-        return getattr(setting, self.roles_key)
+        for subscription in subscriptions:
+            if subscription.role in dispatched_roles:
+                if userid in subscription.watcher.get_user_ids():
```
deiferni: this may lazy-load the watcher attribute and issue many, many queries 🤔

opengever/activity/dispatcher.py
```diff
@@ -13,22 +15,39 @@ class NotificationDispatcher(object):
...
         if not setting:
-            return []
+            return False
+
+        dispatched_roles = getattr(setting, self.roles_key)
+        subscriptions = Subscription.query.filter_by(
+            resource=notification.activity.resource)
 
-        return getattr(setting, self.roles_key)
+        for subscription in subscriptions:
```
deiferni: could you do this with a query? all you want out of this is a boolean. i think loading all the ORM stuff here may be a bit overkill and expensive, especially since this may happen many times.

phgross:
@deiferni updated according to your feedback.

opengever/activity/dispatcher.py
phgross: I've improved the query. But because the watcher model does not stores the user id, but the actor id, the check can not be done within query.

PR #3595, opened 2017-11-09
Add short description about updating policies.

deiferni:
See also https://github.com/4teamwork/opengever-deployments/pull/8


deiferni:
@fguyer das gibts sogar schon https://github.com/4teamwork/opengever.core/blob/1d45769adc825bfbdca33063636b0924329ffa47/docs/intern/operations/saas.rst#release-einspielen 😉

PR #3596, opened 2017-11-09
Fix agenda items rendering on non-word meeting.

deiferni:
The code expected an element to be present all the time whereas it is only
there when the word-meeting feature flag has been enabled.


opengever/meeting/browser/resources/meeting.js
```diff
@@ -369,6 +372,10 @@
     };
 
     this.renderNavigation = function(data) {
+      if (this.navigationTemplate === undefined) {
```
bierik: `typeof this.navigationTemplate === 'undefined'`

deiferni:
@bierik updated

PR #3599, opened 2017-11-13
Adds own field for dossier manager

elioschmutz:
Add new field dossier_manager to the IProtectDossier behavior to set the dossier-manager manually.

The field "dossier_manager" is not read-only as described in the requirements. A manager or a dossier-manager can change this field. This is required if you need to add i.e. a group instead only a user as a dossier-manager or if the manager wants to change the dossier-manager.

screen1

### DIFF
--- a/docs/schema-dumps/opengever.dossier.businesscasedossier.schema.json
+++ b/docs/schema-dumps/opengever.dossier.businesscasedossier.schema.json
@@ -234,6 +234,13 @@
             "description": "W\u00e4hlen Sie Benutzer und Gruppen aus, welche im Dossier lesend und schreibend Zugriff erhalten sollen. Die Benutzer dieser Gruppe werden dazu berechtigt, das Dossier zu bearbeiten, Inhalte zu erstellen und zu bearbeiten.",
             "_zope_schema_type": "List",
             "default": null
+        },
+        "dossier_manager": {
+            "type": "string",
+            "title": "Dossier-Verwalter",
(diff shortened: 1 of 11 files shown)


docs/schema-dumps/opengever.dossier.businesscasedossier.schema.json
```diff
@@ -234,6 +234,13 @@
             "description": "W\u00e4hlen Sie Benutzer und Gruppen aus, welche im Dossier lesend und schreibend Zugriff erhalten sollen. Die Benutzer dieser Gruppe werden dazu berechtigt, das Dossier zu bearbeiten, Inhalte zu erstellen und zu bearbeiten.",
             "_zope_schema_type": "List",
             "default": null
+        },
```
phgross: Because the fields are only handled by the forms and would not have any effect when creating an object by API, it makes no sense to add them to the API schema documentation. I would ignore them all

opengever/dossier/tests/test_protect_dossier.py
```diff
@@ -92,6 +92,7 @@ def test_dossier_manager_can_set_protect_dossier_fields(self, browser):
         form = browser.find_form_by_field('Reading')
         form.find_widget('Reading').fill('projekt_a')
         form.find_widget('Reading and writing').fill('projekt_b')
+        form.find_widget('Dossier manager').fill(self.dossier_manager.getId())
```
phgross: There is no test which asserts the default value ...

elioschmutz:
@phgross 👍 updated the PR with two new commits for your request.

PR #3610, opened 2017-11-14
Fix syncing the submitted proposal title to sql.

### LINKED ISSUE #3607: Missing proposal title in committee proposal listings.
The proposal title is no longer visible in the committee submitted proposal listing when the word meeting feature is not enabled. however when viewing the submitted proposal object, the title is visible.

screen shot 2017-11-14 at 13 42 48
screen shot 2017-11-14 at 13 54 25

deiferni:
The title and other proposal-model attributes should only be synced after
they have been set from the attribute data provided as JSON in the request.

We have not noticed this is a problem so far because the test tests with
attachments which cause an object modified event to be fired later. Same for
the word implementation since creating the word document fires a modified
event, too.

Also contains a critical fix for `SQLUpgradeStep` that ensures statements always are picked up as changes by the zope datamanager.

_Should be backported as a bugfix level release._

Fixes #3607.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2017.6.5 (unreleased)
 ---------------------
 
+- Fix syncing the submitted proposal title to sql. [deiferni]
 - SPV: Fix member links in committee overview. [tarnap]
 - SPV: Make the considerations field a trix field. [tarnap]
 - Don't escape description in overview twice for dossier templates. [deiferni]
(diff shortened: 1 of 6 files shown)


phgross:
@deiferni changes are looking 👍, but I don't really understand the argument to not create an upgradestep. An upgradestep would not disallow shipping the bugfix without downtime in my opinion. Let's discuss that tomorrow

deiferni:
@phgross yes, i was thinking that there is a chance of running into conflicts while upgrading, so for simplicity it is easier to avoid the upgrade. we can just observe the upgrade and re-run in case of conflicts though.

phgross:
@deiferni yes or run the upgradestep later, thats possible too ...

deiferni:
@phgross updated with
- upgrade step to re-sync
- fix for sql-upgrade steps that don't touch plone objects

PR #3616, opened 2017-11-16
Add concluded meetings to committee overview tab

tarnap:
Old:
status_master

New:
ready_for_pr

Resolves https://github.com/4teamwork/gever/issues/158

### DIFF
--- a/opengever/meeting/model/query.py
+++ b/opengever/meeting/model/query.py
@@ -184,7 +184,7 @@ class MeetingQuery(BaseQuery):
     def _past_meetings(self, committee):
         query = self._committee_meetings(committee)
         query = query.filter(Meeting.start < utcnow_tz_aware())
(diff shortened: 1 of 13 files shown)


opengever/meeting/model/query.py
```diff
@@ -181,6 +181,12 @@ def _upcoming_meetings(self, committee):
         query = query.order_by(Meeting.start)
         return query
 
+    def _closed_meetings(self, committee):
```
deiferni: if you look at the method right below this one, does that look similar? 😂

tarnap: ahhh!!

tarnap:
@deiferni: The last commit removes the duplicated query method and uses the existing one. I'll squash the commits as soon as you approve this PR

PR #3617, opened 2017-11-16
Update bumblebee checksum when modifying document doc-properties.

deiferni:
Update bumblebee checksum when modifying a document's doc-properties.

Fire a modified event when we modify the document. Bumblebee listens to those so we can be sure that the cached document checksum will always be correct after adding doc-properties to a docx file.

This fixes an issue where the bumblebee checksum is not updated after creating/modifying a document and thus the bumblebee preview image or overlay could not be rendered.

As discussed there is no upgrade step here since fixing this would include re-calculating the checksum for every document. We will provide a button with which an `Administrator` and `Manager` can attempt to re-render the bumblebee preview, see https://github.com/4teamwork/opengever.core/issues/3619 for a suggestion.

Fixes https://github.com/4teamwork/gever/issues/156.

### DIFF
--- a/opengever/testing/integration_test_case.py
+++ b/opengever/testing/integration_test_case.py
@@ -41,6 +41,7 @@ FEATURE_FLAGS = {
     'bumblebee-open-pdf-new-tab': 'opengever.bumblebee.interfaces.IGeverBumblebeeSettings.open_pdf_in_a_new_window',
     'activity': 'opengever.activity.interfaces.IActivitySettings.is_feature_enabled',
     'contact': 'opengever.contact.interfaces.IContactSettings.is_feature_enabled',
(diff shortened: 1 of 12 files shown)


opengever/testing/integration_test_case.py
```diff
@@ -41,6 +41,7 @@
     'bumblebee-open-pdf-new-tab': 'opengever.bumblebee.interfaces.IGeverBumblebeeSettings.open_pdf_in_a_new_window',
     'activity': 'opengever.activity.interfaces.IActivitySettings.is_feature_enabled',
     'contact': 'opengever.contact.interfaces.IContactSettings.is_feature_enabled',
+    'update-doc-properties': 'opengever.dossier.interfaces.ITemplateFolderProperties.create_doc_properties'
```
phgross: I would name the feature just `doc-properties` similar than the other features

deiferni:
amended with your suggestions.

Issue #3354, opened 2017-09-06
Schema-migration tracking table can't deal with orphaned upgrade


deiferni:
The SchemaMigration tracking table `opengever_upgrade_version` can't handle `ftw.upgrade`s orphaned upgrade steps. It only tracks the most recent version of a profile's schema migrations. Problematic situation:

- SchemaMigration A, version 1
- SchemaMigration B, version 2

The SchemaMigration B Pull-Request is merged first. We execute the upgrades, the tracking-table version is 2. 
Then we merge Pull-Request for Schemamigration A. The upgrade is discovered by `ftw.upgrade`  but since the tracking-table version is at 2 already, the upgrade won't be installed.

Proposed changes:
Also let the tracking-table track each installed upgrade-step and not only the last version for each profile. That way it behaves the same way as `ftw.upgrade`.

phgross:
@jone could you take a look here in the next weeks, would be great to solve this issue with the january release.

PR #3620, opened 2017-11-16
Form for personal activity settings

phgross:
The form is implemented as a small JS form with a handlebars template.

notification settings 2

Styling in an separate PR: https://github.com/4teamwork/plonetheme.teamraum/pull/601

Currently the Endpoints are not plone.REST endpoints. But I'll not change this in this PR, because I wont some testing feedback, as soon as possible. 

The NotificationSettingsForm is not directly registered in opengever.activity or available via a separate URL, instead the PersonalPreferences View inherits the `NotificationSettingsForm`. The reason for that workaround is that we plan that the personal-preferences view provides some more personal settings and the notification form will only be a part of that. But it makes no sense to build a "complex form inheritance" for a possible feature change.

### DIFF
--- a/opengever/activity/browser/resources/settings.js
+++ b/opengever/activity/browser/resources/settings.js
@@ -54,7 +54,9 @@
           activities[activity.kind] = new SettingRow(activity);
         });
       }
-      return this.template({ activities: Object.values(activities), translations: data.translations });
+
+      var values = Object.keys(activities).map(function(e) { return activities[e] })
+      return this.template({ activities: values, translations: data.translations });
     };
 
     this.save_setting = function(target, event) {
--- a/opengever/activity/browser/resources/settings.js
+++ b/opengever/activity/browser/resources/settings.js
@@ -1,0 +1,147 @@
+(function(global, $, Controller, EditboxController, Pin, HBS) {
+
+  "use strict";
+
+  HBS.registerHelper({
+    or: function (v1, v2) {return v1 || v2;}
+  });
+
+  HBS.registerHelper('equals', function(arg1, arg2, options) {
+    return (arg1 === arg2) ? options.fn(this) : options.inverse(this);
+  });
+
+  HBS.registerHelper('translate', function(messageString, options) {
+    return options.data.root.translations[messageString];
+  });
+
+  function SettingRow(data) {
+    this.kind = data.kind;
+    this.type_id = data.type_id;
+    this.kind_title = data.kind_title;
+    this.css_class = data.css_class;
+    this.mail = data.mail;
+    this.badge = data.badge;
+    this.setting_type = data.setting_type;
+    this.changed = false;
+  }
+
+
+  function SettingsFormController(options) {
+    global.Controller.call(
+      this,
+      $('#settingsForm').html(),
+      $('#notification-settings-form'));
+
+    this.fetch = function(){
+      return $.get($('#notification-settings-form').data('list-url'));
+    };
+
+    var activities;
+
+    this.render = function(data) {
+      if (activities) {
+        // Rendering after update
+        data.activities.forEach(function(activity) {
+          if (!activities[activity.kind].changed) {
+            activities[activity.kind] = new SettingRow(activity);
+          }
+        });
+      } else {
+        // Initial rendering
+        activities = {};
+
+        data.activities.forEach(function(activity) {
+          activities[activity.kind] = new SettingRow(activity);
+        });
+      }
+      return this.template({ activities: Object.values(activities), translations: data.translations });
+    };
+
+    this.save_setting = function(target, event) {
+      var row = target.parents('tr');
+      var mail = row.find("ul.mail input:checkbox:checked").map(function(){ return $(this).val(); }).get();
+      var badge = row.find("ul.badge input:checkbox:checked").map(function(){ return $(this).val(); }).get();
+
+      return this.request($('#notification-settings-form').data('save-url'), {
+        method: "POST",
+        data: { kind: row.data('kind'),
+                mail: JSON.stringify(mail),
+                badge: JSON.stringify(badge)}
+      }).done(function() {
+        activities[row.data('kind')].changed = false;
+      });
+    };
+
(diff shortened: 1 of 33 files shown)


opengever/activity/browser/resources/settings.js
```diff
@@ -0,0 +1,146 @@
+(function(global, $, Controller, EditboxController, Pin, HBS) {
+
+  "use strict";
+
+  HBS.registerHelper({
+    or: function (v1, v2) {return v1 || v2;}
+  });
+
+  HBS.registerHelper('equals', function(arg1, arg2, options) {
+    return (arg1 == arg2) ? options.fn(this) : options.inverse(this);
```
deiferni: you may wanna use `===` here

opengever/activity/browser/settings.py
```diff
@@ -0,0 +1,187 @@
...
+from zope.i18n import translate
+import json
+
+
+TEMPLATES_DIR = Path(__file__).joinpath('..', 'templates').abspath()
+
+
+# The following list contains all necessary informations about the activity
+# groups which should be exposed in the notification settings form.
+ACTIVITY_GROUPS = [
```
deiferni: this seems to duplicate a lot of what is already in here:https://github.com/4teamwork/opengever.core/blob/c81c0369cab47d16aa22119afc994d2f948aab9c/opengever/activity/hooks.py#L8

do you see a way to reduce that duplication?

opengever/activity/browser/settings.py
```diff
@@ -0,0 +1,187 @@
...
+        kind = self.request.form['kind']
+        mail = json.loads(self.request.form['mail'])
+        badge = json.loads(self.request.form['badge'])
+
+        # todo check roles before save
+        setting = self.get_or_create_setting(kind)
+        setting.mail_notification_roles = mail
+        setting.badge_notification_roles = badge
+
+        return JSONResponse(self.request).proceed().dump()
```
deiferni: may be nice to have a short confirmation for the user with `.info()`?

opengever/activity/browser/settings.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def reset(self):
+        """Reset a personal setting
+        """
+        kind = self.request.form['kind']
+        setting = self.get_setting(kind)
+        create_session().delete(setting)
+        return JSONResponse(self.request).proceed().dump()
```
deiferni: may be nice to have a short confirmation for the user with `.info()`?

opengever/activity/browser/settings.py
phgross: In my opinion the user gets enough information because the save and cancel button will disappear. An additional success message will be more disruptive that it would help.

opengever/activity/browser/settings.py
phgross: Same here ...

opengever/activity/browser/settings.py
phgross: I've discussed this issue last friday with lukas. To combine all informations in the same list, would make the information completely unreadable. So we would not change this for now. I did the same for the kind translations.

phgross:
@deiferni please take another look, i made some improvements according to yours, pascals and lukas feedback.

opengever/activity/browser/settings.py
deiferni: Yes could maybe make some full-blown objects here, but not a blocker for this PR if we're not there yet :wink:.

My main concern for not duplicating is that changes could be made in one place and not in the other one. But i've seen you have added `test_list_all_settings`, and if i understand correctly it ensures that the lists are in sync.

deiferni:
⚠️  i think i had an orphaned schema migration locally, could you quickly verify before merging that there are no orhpans? currently orphaned schema migrations don't get executed, see https://github.com/4teamwork/opengever.core/issues/3354.

deiferni:
Please also create and link `plonetheme.teamraum`-`pg_notification_settings_form` PR.

deiferni:
I can produce a button that has no visible effect, to reproduce:

screen shot 2017-11-20 at 14 55 23

- In the row `Aufgabe hinzugefügt` check `Auftraggeber` (or any checkbox i guess)
- Click save
- Uncheck `Auftraggeber`
- `Auf Standard zurücksetzen` has no visible effect and actually fails with `UnmappedInstanceError` which is something to look into, too 🤔 .

opengever/base/browser/personalpreferences.py
```diff
@@ -0,0 +1,9 @@
+from opengever.activity.browser.settings import NotificationSettingsForm
+
+
+class PersonalPreferences(NotificationSettingsForm):
```
deiferni: why have two separate forms/views? do we use `NotificationSettingsForm` somewhere else (couldn't find where)?

opengever/core/upgrades/20171117133906_enable_personal_preferences_action/actions.xml
```diff
@@ -0,0 +1,22 @@
...
+
+    <object name="preferences" meta_type="CMF Action" i18n:domain="plone">
+      <property name="title" i18n:translate="">Preferences</property>
+      <property name="description" i18n:translate="" />
+      <property name="url_expr">string:${globals_view/navigationRootUrl}/@@personal-preferences</property>
+      <property name="link_target" />
+      <property name="icon_expr" />
+      <property name="available_expr">python:member is not None</property>
+      <property name="permissions">
+        <element value="View" />
```
deiferni: the form itself is protected with the `cmf.SetOwnProperties` permissions, should we use that one here, too?

phgross: The original plone action does not, but it definitely makes sense, updated ...

opengever/base/browser/personalpreferences.py
phgross: See updated PR description (sorry forgot to describe that in the PR)

phgross:
@deiferni updated

PR #3622, opened 2017-11-17
Use localstorage for tracking expanded_uids instead of cookie.

bierik:
Closes https://github.com/4teamwork/opengever.core/issues/2830


lukasgraf:
🎉  💚  💜  🍰  ❗️

PR #3627, opened 2017-11-17
Add field description for keyword fields

njohner:
I added the new description of the keyword fields in the translations.
I included a line-break to make it more readable.
I also added a french translation.
screen shot 2017-11-17 at 15 41 57
screen shot 2017-11-17 at 15 44 47

### DIFF
--- a/opengever/document/locales/de/LC_MESSAGES/opengever.document.po
+++ b/opengever/document/locales/de/LC_MESSAGES/opengever.document.po
@@ -248,7 +248,7 @@ msgstr "Referenz auf das entsprechende Dossier des Absenders"
 
 #: ./opengever/document/behaviors/metadata.py
 msgid "help_keywords"
-msgstr "Schlagwörter zur Umschreibung eines Dossiers. Nicht zu verwechseln mit der Ordnungsposition"
+msgstr "Schlagwörter zur Umschreibung eines Dokuments. Nicht zu verwechseln mit der Ordnungsposition. <br>ACHTUNG: Beachten Sie bei der Verwendung von Schlagworten die Datenschutzvorgaben (z.B. keine Eigennamen)."
 
 #. Default: ""
 #: ./opengever/document/behaviors/metadata.py
--- a/docs/schema-dumps/ftw.mail.mail.schema.json
+++ b/docs/schema-dumps/ftw.mail.mail.schema.json
@@ -63,7 +63,7 @@
         "keywords": {
             "type": "array",
             "title": "Schlagworte",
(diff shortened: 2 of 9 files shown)


opengever/document/locales/de/LC_MESSAGES/opengever.document.po
```diff
@@ -248,7 +248,7 @@ msgstr "Referenz auf das entsprechende Dossier des Absenders"
 
 #: ./opengever/document/behaviors/metadata.py
 msgid "help_keywords"
-msgstr "Schlagwörter zur Umschreibung eines Dossiers. Nicht zu verwechseln mit der Ordnungsposition"
+msgstr "Schlagwörter zur Umschreibung eines Dossiers. Nicht zu verwechseln mit der Ordnungsposition. <br>ACHTUNG: Beachten Sie bei der Verwendung von Schlagworten die Datenschutzvorgaben (z.B. keine Eigennamen)."
```
deiferni: I suspect the translation here should be `… Umschreibung eines Dokuments.`.

If it is a document field (i think so since it is in the `opengever.document` domain and in the document sub-package) the translation should also refer to the document 🤔 .

njohner: That makes sense 😄...

tarnap:
@deiferni: Is html-code (the line break) in the translations really :+1:?

njohner:
I wasn't sure for the html line break, that's why I pointed it out...

tarnap:
@njohner: I'm not sure either

deiferni:
i would say it is discouraged 🤔  (especially when we control the template), but it will work.

in this case we don't have a separate template or custom widget but just render the field, so i think it is ok. we also use it in other places.

Rotonen:
Markup-from-translation is an XSS vector. Rather break it up into two labels.

tarnap:
> we also use it in other places

:cry:

njohner:
Ok so not sure what should be done here. I can't break it up into two labels, because we don't have control over the form and just render the description field. So leave it as is? Take out the linebreak?

deiferni:
@Rotonen the thing here is that this is a zope schema field in a plone contenttype form. So in order to avoid markup in that case i think we'd have to create a custom field and/or a custom widget. This is a bit overkill IMHO. Also we are in "complete" control of the translations.

In general i agree though, i also think we should discourage to put markup into the translation files.

deiferni:
_btw: the tests seem to be failing, but somehow our CI does not tell github._

phgross:
@njohner FYI: hijacked your branch and updated the schema dumps (was the reason for the failing tests).

njohner:
closes #3625

PR #3628, opened 2017-11-17
Fix redirector after sablon template creation

### LINKED ISSUE #3609: Wrong redirection when creating new sablon templates
An error occurs when adding a new sablon template.

sablom bug

tarnap:
Previously: sablom bug

And now: 
sablom bug_fixed

Resolves #3609

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 ---------------------
 
 - Show filterlist also on empty tabbedview listings. [phgross]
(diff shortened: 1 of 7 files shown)


deiferni:
could you squash the commits after rebase though? i personally find it useful to have these three things in one commit (new functionality, tests for new functionality and upgrade that installs it).

tarnap:
@deiferni: Of course. I split things into several commits to make the review easier.

PR #3631, opened 2017-11-20
Fixed and convert TaskTemplateTrigger tests to integration tests.

phgross:
This PR also fixes the currently broken master ...

### DIFF
--- a/opengever/tasktemplates/tests/test_trigger.py
+++ b/opengever/tasktemplates/tests/test_trigger.py
@@ -4,64 +4,38 @@ from ftw.builder import Builder
 from ftw.builder import create
 from ftw.testbrowser import browsing
 from ftw.testbrowser.pages.statusmessages import error_messages
+from opengever.tasktemplates.content.tasktemplate import ITaskTemplate
 from opengever.tasktemplates.interfaces import IFromTasktemplateGenerated
-from opengever.testing import FunctionalTestCase
+from opengever.testing import IntegrationTestCase
 from plone import api
-from plone.app.testing import TEST_USER_ID
-import transaction
 
 
-class TestTriggeringTaskTemplate(FunctionalTestCase):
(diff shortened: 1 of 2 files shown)


opengever/tasktemplates/tests/test_trigger.py
```diff
@@ -4,202 +4,205 @@
...
 
 
-class TestTriggeringTaskTemplate(FunctionalTestCase):
+class TestTriggeringTaskTemplate(IntegrationTestCase):
 
-    def setUp(self):
-        super(TestTriggeringTaskTemplate, self).setUp()
+    def trigger_tasktemplatefolder(
+            self, browser, folder=u'Verfahren Neuanstellung',
+            templates=[], documents=[]):
```
Rotonen: Default to `None` and set to `[]` based on whether or not something came in.

deiferni: careful wit mutable defaults!

opengever/tasktemplates/tests/test_trigger.py
```diff
@@ -4,202 +4,205 @@
...
 
         for subtask in main_task.listFolderContents():
             self.assertTrue(IFromTasktemplateGenerated.providedBy(subtask))
 
     @browsing
     def test_creates_a_subtask_for_each_selected_template(self, browser):
+        self.login(self.regular_user, browser=browser)
+        create(Builder('tasktemplate')
+               .titled(u'Notebook einrichten.')
+               .having(**{'issuer': 'responsible',
```
deiferni: i find it a bit weird to use a dict here for the keyword arguments, especially since the dict is also just defined in-line and not a variable, so why not:
```python
.having(issuer='responsible', …)
```

phgross:
@deiferni @Rotonen updated ...

PR #3634, opened 2017-11-20
Set excel file name to dossier title when creating a disposition report.

njohner:
Dossier title seemed to be the most appropriate basis for the report filename.

Should this kind of change be marked in the HISTORY.txt file?

### DIFF
--- a/opengever/disposition/browser/excel_export.py
+++ b/opengever/disposition/browser/excel_export.py
@@ -8,7 +8,9 @@ from opengever.base.reporter import XLSReporter
 from opengever.disposition import _
 from opengever.dossier import _ as dossier_mf
 from plone import api
+from plone.i18n.normalizer.interfaces import IIDNormalizer
 from Products.Five.browser import BrowserView
+from zope.component import getUtility
 from zope.globalrequest import getRequest
 from zope.i18n import translate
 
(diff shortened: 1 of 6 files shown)


opengever/disposition/browser/excel_export.py
```diff
@@ -81,10 +81,19 @@ def __call__(self):
...
+            self.request, self.get_file_name())
 
         return data
 
+    def get_file_name(self):
+        """Returns the filename for the excel spreadsheet.
+        """
+        name_basis = _(u'appraisal_list', default=u'appraisal_list')
+        file_name = "{0}_{1}.xlsx".format(
+            self.context.translate(name_basis),
```
phgross: self.context.translate is outdated, please use

```
translate(name_basis, context=self.request)
```

opengever/disposition/browser/excel_export.py
```diff
@@ -81,10 +81,19 @@ def __call__(self):
...
+            self.context.title_or_id().replace(" ", "_"))
```
phgross: instead of only replacing spaces i would use the ID normalizer, so unicode chars etc. are handled too.
```
from plone.i18n.normalizer.interfaces import IIDNormalizer
normalizer = getUtility(IIDNormalizer)

normalizer.normalize(self.context.title_or_id())
```

njohner: Ok thanks, didn't know its existence.

opengever/disposition/browser/excel_export.py
njohner: Ok changed it

njohner:
* changes done according to comments
* Wrote a test for the filename.

Do translations work in the test environment? I tried to switch language in the testing without success (I tried `view.context.plone.language="de"` and `view.request['LANGUAGE']="de"`)

njohner:
Adapted the test to use the german translation.

njohner:
rebased do I merge?

phgross:
@njohner I'll as soon the builds are green. Because of release timing etc. its easier if one of the GEVER team merges the PR.

njohner:
perfekt

njohner:
closes #2675

PR #3638, opened 2017-11-21
Add group title option to the OGDS Synchronisation.

phgross:
Add an registry configuration option to define an LDAP attribute name, which should be used when syncing LDAP groups into the OGDS.

This allow us to use the group description as group title on some installations, thats a part of the OGIP 23 changes.

### DIFF
--- a/opengever/ogds/base/sync/ogds_updater.py
+++ b/opengever/ogds/base/sync/ogds_updater.py
@@ -3,12 +3,14 @@ from logging.handlers import TimedRotatingFileHandler
 from opengever.base.model import create_session
 from opengever.base.pathfinder import PathFinder
 from opengever.ogds.base.interfaces import ILDAPSearch
+from opengever.ogds.base.interfaces import IOGDSSyncConfiguration
 from opengever.ogds.base.interfaces import IOGDSUpdater
 from opengever.ogds.base.sync.import_stamp import set_remote_import_stamp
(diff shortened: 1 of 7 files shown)


opengever/ogds/base/sync/ogds_updater.py
```diff
@@ -297,6 +304,11 @@ def import_groups(self):
 
                     setattr(group, col.name, value)
 
+                # Sync group title
+                title_attribute = self.get_group_title_ldap_attribute()
+                if title_attribute and info.get(title_attribute):
+                    setattr(group, 'title', info.get(title_attribute))
```
deiferni: you don't need setattr here, this should work fine:

```python
group.title = info.get(title_attribute)
```

opengever/ogds/base/sync/ogds_updater.py
```diff
@@ -297,6 +304,11 @@ def import_groups(self):
 
                     setattr(group, col.name, value)
 
+                # Sync group title
+                title_attribute = self.get_group_title_ldap_attribute()
+                if title_attribute and info.get(title_attribute):
```
deiferni: or:
```python
… title_attribute in info
```
?

phgross:
@deiferni updated ...

phgross:
@deiferni updated and added tests.

PR #3639, opened 2017-11-21
Fix renaming of agenda items with proposal

tarnap:
change_agenda_item_title_fixed

Resolves https://github.com/4teamwork/gever/issues/160

### DIFF
--- a/opengever/meeting/model/agendaitem.py
+++ b/opengever/meeting/model/agendaitem.py
@@ -211,6 +211,7 @@ class AgendaItem(Base):
     def set_title(self, title):
         if self.has_proposal:
             self.submitted_proposal.title = title
+            self.submitted_proposal.sync_model()
         else:
             self.title = title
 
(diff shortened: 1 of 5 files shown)


opengever/meeting/model/agendaitem.py
```diff
@@ -211,6 +211,7 @@ def get_title(self, include_number=False):
     def set_title(self, title):
         if self.has_proposal:
             self.submitted_proposal.title = title
+            self.submitted_proposal.sync_model()
```
deiferni: could you add a test for this, or extend an existing test and make sure the title is set correctly on sql and plone?

tarnap: Of course! :)

tarnap:
@deiferni: This is probably the kind of test you wanted... but, could you suggest a better title for the created `agenda_item`? https://github.com/4teamwork/opengever.core/pull/3639/commits/4d7f0afe01e70841ad57156bfcf33656bf68ea61

tarnap:
@deiferni: shall I squash the commits together before merging?

deiferni:
@tarnap i'd say so, looks small enough for one commit.

PR #3655, opened 2017-11-22
Replace doc-properties in protocol, use correct master doc for excerpt.

deiferni:
This PR improves/fixes doc-property handling in `opengever.meeting`.

- It ensures that doc-properties are updated in the protocol when merging all the submitted-proposal documents together. It also makes sure that then the doc-properties are replaced with their text-value in the generated `docx` file for the protocol. This curcumvents conflicts that would otherwise be caused by the submitted-proposal documents since they all contain the same set of gever doc properties, but with different values.

- It does not replace doc-properties when creating excerpts but uses the submitted proposals document as master document, thus keeping all its doc-properties. 

I removed knowledge of `Sablon` and `Document` from `DocxMergeTool` and decoupled it a bit from opengever, it now only receives data of a docx document and it is the callers responsibility to extract the data from whatever object he is currently handling.

Closes https://github.com/4teamwork/gever/issues/136.


opengever/meeting/mergetool.py
```diff
@@ -1,50 +1,70 @@
...
-        map(composer.append, map(Document, self._file_paths[1:]))
         result_path = self._tempdir_path.joinpath('result.docx')
-        composer.save(result_path)
+        self._composer.save(result_path)
         return result_path.bytes()
 
-    def _get_file(self):
-        path = self._tempdir_path.joinpath('{0}.docx'.format(len(self._file_paths)))
-        self._file_paths.append(path)
+    def add(self, bytes):
```
jone: `bytes` is a built-in type, please choose a different name.

opengever/meeting/mergetool.py
```diff
@@ -1,50 +1,70 @@
...
+    It is used as a context manager and accepts bytes of documents.
     Calling the DocxMergeTool object merges the documents and returns the resulting
     bytes.
-    The files are merged in the order they were added.
+
+    The files are merged in the order they were added or inserted as as
+    specified by inserts index.
+
     """
+    def __init__(self, master, remove_property_fields=True):
```
jone: pep8 would require a blank line before this one, right?

deiferni: hm no pep 8 does not really take a detailed stance on docstrings, maybe in  https://www.python.org/dev/peps/pep-0257/ one could find a recommendation 🤔 .

but out of personal preference i don't like trailing newlines inside the docstrings so i moved it between docstring and method definition.

opengever/meeting/mergetool.py
deiferni: 😱  changed.

deiferni:
@jone updated with your suggestions. also tests are now green since the package has finally propagated through the CDN 😄 .

opengever/meeting/mergetool.py
jone: Ah, I wasn't refering to the empty line within the docstring but the one separating docstring and method. So you actually exactly did what I menat 😂

deiferni: 🙃

PR #3660, opened 2017-11-22
Add mimetype for MS OneNote.

phgross:
The icon will be available with https://github.com/4teamwork/plonetheme.teamraum/pull/598

### DIFF
--- a/opengever/policy/base/profiles/mimetype/mimetypes.xml
+++ b/opengever/policy/base/profiles/mimetype/mimetypes.xml
@@ -83,6 +83,16 @@
       mimetypes="application/vnd.ms-outlook"
       />
 
+  <!-- MS OneNoe -->
+  <mimetype
+      name="Microsoft OneNote"
+      binary="True"
+      extensions="one"
+      globs="*.one"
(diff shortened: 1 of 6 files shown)


opengever/policy/base/profiles/mimetype/mimetypes.xml
```diff
@@ -83,6 +83,16 @@
       mimetypes="application/vnd.ms-outlook"
       />
 
+  <!-- MS OneNoe -->
+  <mimetype
+      name="Microsoft OneNote"
+      binary="True"
+      extensions="one"
+      globs="*.one"
+      icon_path="icon_dokument_onenote.gif"
```
deiferni: so, just to understand how the internals work here, i assume that icon_path is necessary even though we don't use it? and that is why it is set?

phgross: exactly, out of `icon_dokument_onenote.gif` it generates the css_class `icon-dokument_onenote` which is handled by plonetheme.teamraum (see the helper method `get_css_class` for more details).

PR #3667, opened 2017-11-23
Show subdossier depending on max dossier depth.

phgross:
- Only show subdossier tab on dossiers where the max depth allows to add subdossiers.
- Reworked and renamed the `is_subdossier_addable` method.
- Always show the subdossier tab, when subdossiers exists, even when there not allowed by the max_depth.

Backport needed: 2017.5.x

### DIFF
--- a/opengever/dossier/browser/tabbed.py
+++ b/opengever/dossier/browser/tabbed.py
@@ -1,6 +1,7 @@
 from opengever.contact import is_contact_feature_enabled
 from opengever.dossier import _
 from opengever.dossier.dossiertemplate import is_dossier_template_feature_enabled
+from opengever.dossier.dossiertemplate.interfaces import IDossierTemplateSettings
 from opengever.meeting import is_meeting_feature_enabled
 from opengever.meeting import is_word_meeting_implementation_enabled
 from opengever.tabbedview import GeverTabbedView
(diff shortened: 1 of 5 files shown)


opengever/dossier/browser/tabbed.py
```diff
@@ -40,9 +41,23 @@ class DossierTabbedView(GeverTabbedView):
         'title': _(u'label_info', default=u'Info'),
         }
 
+    def is_subdossier_visible(self):
```
deiferni: i can see a problem here in the following situation:
- max-depth is enforced also for dossier-templates
- there are legacy dossiers that exceed max-depth (this may be the case for migrations, for example)

i think in in those cases the subdossier tab would be removed too early, thus "hiding" some sub-dossiers? shouldn't we instead just check if there are more subdossiers, and in that case always display the tab?

phgross: Yes, but the max-depth is only enforced for dossier-templates when the `respect_max_depth` is enabled. Deployments with with migrated content, which does not respect the dossiers max_depth, should not enable the `repect_max_depth`, because that would just be incorrect. 😉 

I don't like the fact about your proposal that in some dossiers the subdossiers tab will be shown and in some dossiers not ...

deiferni: we could have the following situation:

- a customer wants that dossier templates respect max depth (`respect_max_depth = True`)
- a customer could have (possibly migrated) legacy dossiers where `max-depth` was not enforced, or the desired `max-depth` could change due o a policy change on the customer side

wouldn't this then hide the subdossier tab for those legacy dossiers where `max-depth` was not enforced?

phgross:
@deiferni updated as discussed ...

Issue #3653, opened 2017-11-22
Prefix items with icons in the Admin tab "Geschützte Objekte"


phgross:
With #3613 a new tab "Blocked local roles" has been introduced which lists repositoryfolders and dossiers with blocked local roles. To have a better possibility to diff the objects also the links should be prefixed with the corresponding contenttype icon.

Hints:
- The view you'll find in opengever.sharing.browser.
- To get the correct contenttype class you can use the method `get_css_class` (`opengever.base.browser.helper.get_css_class`)

![bildschirmfoto_2017-11-22_um_09_58_04](https://user-images.githubusercontent.com/485755/33123933-d52394fc-cf7b-11e7-9b54-7e20b5308952.png)

phgross:
Closed via #2812 3670

PR #3670, opened 2017-11-23
Show icons in BlockedLocalRolesList

njohner:
I also changed the naming of the `children` variable which was redefined in a loop over `children`. While that works, it seemed to be a rather unfortunate choice 😄 
solves #3653 
screen shot 2017-11-23 at 17 12 49

### DIFF
--- a/opengever/sharing/tests/test_blocked_local_roles_listing.py
+++ b/opengever/sharing/tests/test_blocked_local_roles_listing.py
@@ -205,6 +205,13 @@ class TestBlockedLocalRolesListing(IntegrationTestCase):
 
         self.assertFalse(browser.css('.level3 a').text)
 
+        nodes = browser.css('.blocked-local-roles-link')
+        repo = 'contenttype-opengever-repository-repositoryfolder'
+        dossier = 'contenttype-opengever-dossier-businesscasedossier'
+        self.assertIn(repo, nodes[0].classes)
+        self.assertIn(repo, nodes[1].classes)
+        self.assertIn(dossier, nodes[2].classes)
+
     @browsing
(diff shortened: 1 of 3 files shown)


njohner:
wops, indeed broke all the tests...

njohner:
Do we want to keep the `blocked-local-roles-link` class for these blocks? I put it back for now, which also fixes all the tests that rely on the class to find the blocks. From what I've seen the class is not used in any css of `opengever.core`, but I guess it doesn't hurt to have it there and maybe it's used in some clients'installations? I could also take it out and adapt the tests.
LMKWYT

njohner:
I realise I hadn't put anyone as reviewer on this...

Rotonen:
@njohner I originally added the class in there to future-proof any styling choices specific to this listing - if there is some specifying class there we only need to change `plonetheme.teamraum` at that time and we're good to go.

opengever/sharing/tests/test_blocked_local_roles_listing.py
```diff
@@ -205,6 +205,13 @@ def test_blocked_role_tab_tree_rendering_of_dossier(self, browser):
 
         self.assertFalse(browser.css('.level3 a').text)
 
+        nodes = browser.css('.blocked-local-roles-link')
```
Rotonen: Amend these to the repofolder tree rendering test just below as well to future-proof for incoming changes.

I predict at some point there will be a wish to somehow highlight the original results and this can break either of these.

njohner: Ok done.

njohner:
Ok rebased, squashed and corrected changelog

PR #3680, opened 2017-11-24
Fix the repository root tab registration upgrade step

Rotonen:
My bad, copy-paste fail.

We've not yet rolled this out to anywhere so we can just retrofix this like so here?


phgross:
We already updated DEV systems (our own some customer dev systems). So please touch the upgradestep.

bin/upgrade touch "path to the upgradestep"

PR #3688, opened 2017-11-27
Fix bug for unknown mimetypes

### LINKED ISSUE #3637: UnboundLocalError when extracting mail attachments
to reproduce:

- use mail from https://extranet.4teamwork.ch/support/staatsarchiv-kanton-basel-stadt/tracker/40
- Aktionen -> Anhänge speichern
- 💥

njohner:
One of the jpg has `image/jpg` as content type which is not a registered mimetype. I tried to add it as a new mimetype but could not make it work yet.

For now, simply fixing the missing condition will avoid crash and simply give a missing icon in the table, see below. That safeguard should anyway be there.

screen shot 2017-11-27 at 16 54 56
fixes #3637

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -7,6 +7,7 @@ Changelog
 
 - Fix bux when accepting a multi admin unit team task with option participate. [phgross]
 - Fix an untranslated OC error message. [Rotonen]
+- Fix bug for unknown mimetypes [njohner]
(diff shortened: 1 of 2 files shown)


njohner:
Yes when I use the same image and insert it in a different E-mail, it does get assigne `image/jpeg` as content-type. This could be set by hand to `image/jpg` in any E-mail to recreate the error for a test case.

PR #3690, opened 2017-11-28
German translation for notification badge

phgross:
There is no corresponding german translation therefore we use `GEVER Benachrichtigung`.

additional:
- Replace the term `Benachrichtigungsmelder` with `Benachrichtigungsanzeige` in documentation
- Update schema .inc file in api documentation


docs/public/dev-manual/api/schemas/ftw.mail.mail.inc
```diff
@@ -70,7 +70,7 @@
        :Datentyp: ``Tuple``
        
        
-       :Beschreibung: Schlagwörter zur Umschreibung eines Dossiers. Nicht zu verwechseln mit der Ordnungsposition
+       :Beschreibung: Schlagwörter zur Umschreibung eines Dokuments. Nicht zu verwechseln mit der Ordnungsposition. <br>ACHTUNG: Beachten Sie bei der Verwendung von Schlagworten die Datenschutzvorgaben (z.B. keine Eigennamen).
```
deiferni: this contains inline `br`s, can `rst` deal with this?

opengever/activity/locales/de/LC_MESSAGES/opengever.activity.po
```diff
@@ -105,7 +105,7 @@ msgstr "Aktivität"
 #. Default: "Badge"
 #: ./opengever/activity/browser/settings.py
 msgid "label_badge"
-msgstr "Badge"
+msgstr "GEVER Benachrichtigung"
```
deiferni: why also GEVER? shouldn't that be obvious to the user 😄 ?

docs/public/dev-manual/api/schemas/ftw.mail.mail.inc
phgross: It displays the <br> tag as text. The <br> tag is part of the field description, I just updated documentation. This update was missing in the previous PR.

opengever/activity/locales/de/LC_MESSAGES/opengever.activity.po
phgross: Yes and No. GEVER stands for Application. So its clearer that the In-Application Notifications are meant.

deiferni: i see, works for me 😄 .

PR #3691, opened 2017-11-28
Increase group title length.

phgross:
The limit of 50 makes no sense and bring problems with the new group title sync option (see https://github.com/4teamwork/opengever.core/pull/3638).

This PR contains only the schema migration, schema change in https://github.com/4teamwork/opengever.ogds.models/pull/50.

### DIFF
--- a/opengever/core/upgrades/20171128092302_increase_group_title_length/upgrade.py
+++ b/opengever/core/upgrades/20171128092302_increase_group_title_length/upgrade.py
@@ -1,0 +1,18 @@
+from opengever.core.upgrade import SchemaMigration
+from sqlalchemy import String
+
+
+GROUP_TITLE_LENGTH = 255
+
+
+class IncreaseGroupTitleLength(SchemaMigration):
+    """Increase group title length.
+    """
+
+    def migrate(self):
+        self.op.alter_column(
+            'groups',
+            'title',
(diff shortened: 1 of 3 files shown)


opengever/core/upgrades/20171128092302_increase_group_title_length/upgrade.py
```diff
@@ -0,0 +1,18 @@
...
+    """Increase group title length.
+    """
+
+    def migrate(self):
+        self.op.alter_column(
+            'groups',
+            'title',
+            type_=String(GROUP_TITLE_LENGTH),
+            existing_nullable=True,
+            existing_type=String(50))
```
deiferni: this is required by mysql when the type is not changed, http://alembic.zzzcomputing.com/en/latest/ops.html#alembic.operations.Operations.alter_column.params.existing_type. but we change the type. so i assume this won't cause trouble being here 🤔 ?

phgross: No I don't thinks so.  We had the same ugpradestep for other columns, see https://github.com/4teamwork/opengever.core/blob/033e68ae4de64908e89cd7726cf516f26ec5ce8b/opengever/ogds/base/upgrades/to4301.py#L32-L36 for example.

PR #3694, opened 2017-11-28
OGDS sync: handle multivalued group titles.

phgross:
The group title attribute is configurable by the IOGDSSyncConfiguration group_title_ldap_attribute configuration flag and can be a multivalued attribute.

### DIFF
--- a/opengever/ogds/base/sync/ogds_updater.py
+++ b/opengever/ogds/base/sync/ogds_updater.py
@@ -12,6 +12,7 @@ from opengever.ogds.models.group import Group
 from opengever.ogds.models.user import User
 from plone import api
 from Products.CMFPlone.interfaces import IPloneSiteRoot
+from Products.CMFPlone.utils import safe_unicode
 from Products.LDAPMultiPlugins.interfaces import ILDAPMultiPlugin
 from sqlalchemy import String
 from sqlalchemy.orm.exc import MultipleResultsFound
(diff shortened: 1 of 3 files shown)


opengever/ogds/base/sync/ogds_updater.py
```diff
@@ -133,6 +133,22 @@ def _get_uid_attr(self, ldap_userfolder):
...
+    def _convert_value(self, value):
+        """Joins multivalued (list or tuples) in to a single string and make
+        sure its an unicode string.
+        """
+
+        # We can't store sequences in SQL columns. So if we do get
+        # a multi-valued field to be stored directly in OGDS, we
+        # treat it as a multi-line string and join it.
+        if isinstance(value, list) or isinstance(value, tuple):
+            value = ' '.join([str(v) for v in value])
```
deiferni: if `v` could be `unicode` or `str` i would convert here, otherwise `' '.join()` may fail for non-ascii chars.

opengever/ogds/base/tests/test_ogds_updater.py
```diff
@@ -79,6 +79,26 @@ def test_uses_title_attribute_for_group_title_when_set(self):
         group = ogds_service().fetch_group('og_mandant1_users')
         self.assertEquals(u'OG Mandant1 users', group.title)
 
+    def test_handles_multivalues_group_titles(self):
+        FAKE_LDAP_USERFOLDER.groups = [
+            create(Builder('ldapgroup')
+                   .having(description=[u'OG Mandant1 users', 'Addition'])
```
deiferni: use umlauts, use unicode 😉

phgross:
@deiferni updated ...

PR #3697, opened 2017-11-28
Remove checkout/edit and cancel actions

### LINKED ISSUE #2905: Disable obsolete checkout action.
The obsolete checkout action in the header section should no longer be activated.
screen shot 2017-05-03 at 14 22 50

njohner:
* Removed the *checkout/edit* portal_type action.
* Removed the *checkout cancel* portal_type action and made a document-action.

**User checked out the document, manager can cancel the checkout**
manager

**Manager checked out document, user can't cancel**
user

resolves #2905


Rotonen:
Users out there AFAIK primarily use the portal action bar for checking a document out instead of the file action buttons.

Can we get some user feedback from real users on this?

opengever/core/upgrades/20171124163904_remove_edit_checkout_and_cancel_actions/types/opengever.document.document.xml
```diff
@@ -0,0 +1,14 @@
+<object xmlns:i18n="http://xml.zope.org/namespaces/i18n" name="opengever.document.document" meta_type="Dexterity FTI" i18n:domain="opengever.core">
+
+  <action
+      action_id="checkout_edit_document"
+      remove="true">
+  </action>
+
+  <action
+      action_id="cancel_document_checkout"
```
deiferni: does this actually work? i see some redundancy with `actions_remove_type_action`, i think only either `actions_remove_type_action` or the changes in this xml are necessary. could you double check?

_(the fact that we have an `actions_remove_type_action` method indicates that the removing actions via `xml` may not actually work.)_

njohner: I first had only the `xml` remove and it did not work. So I added the `actions_remove_type_action` and then it worked. I'll test again with only the `actions_remove_type_action`, but without the `xml`...

deiferni: 👍  i strongly suspect the `xml` is not necessary 😉

njohner: Yes you were right of course! xml remove does not seem to work, and the `actions_remove_type_action` alone is just fine

deiferni:
also if we merge this as is the documentation must be updated, see https://docs.onegovgever.ch/user-manual/dokumente/bearbeiten/#dokument-auschecken-und-bearbeiten

phgross:
@njohner some questions, but changes looks 👍 :

 - Are the revert icon already implemented in `plonetheme.teamraum`? if not please open a PR there as well and paste the link in to the PR description.
 - Is force checkin still possible for administrators and managers?

njohner:
I don't think the revert icon is in `plonetheme.teamraum`, so I'll create an issue there.
Force checkin still works for `Administrator` or `Verwalten` rights.

njohner:
That was a misunderstanding on my side, we don't need to do anything in `plonetheme.teamraum` here

deiferni:
@njohner i am looking at the open PRs. What is the state here? Is it ready for re-review or are you still working on this?

njohner:
@deiferni 
Just the documentation still needs to be updated. Should I do that?
For the rest it's ready

deiferni:
@njohner yes, please.

njohner:
@deiferni 
Ok I updated the docs, so this is ready for review. Thanks @siegy22 for local bumblebee installation!
Here are 2 snapshots from the user-manual showing the updated images and text.

**Only one button remaining for checkout:**
screen shot 2018-01-15 at 16 48 00

**Moved the cancel checkout button**
screen shot 2018-01-15 at 16 47 16

njohner:
@deiferni @Rotonen 
Just had to adapt one more test (https://github.com/4teamwork/opengever.core/pull/3697/commits/62da6a1897e1ecd8b8efe98a17b030998680767d), which I think was ported by @Rotonen to integration layer since I first made this pull request. A one line change, but I'll wait for a thumbs up before merging.

opengever/document/locales/de/LC_MESSAGES/opengever.document.po
```diff
@@ -23,6 +23,10 @@ msgstr "Mit Mailprogramm versenden"
 msgid "Can't edit the document at moment, beacuse it's locked."
 msgstr "Das Dokument kann momentan nicht bearbeitet werden, es ist gesperrt."
 
+#: ./opengever/document/browser/templates/macros.pt
+msgid "Cancel checkout"
```
Rotonen: Base already has a translation for this and there are similar ones within document as well - how do we end up with a new one here?

deiferni: Good point. The translation has been moved to a page template. The previous implementation used actions, which have to be manually translated in the `plone` domain.
@njohner i think you could drop the manual translations for the checkout related stuff since the actions are dropped, too. You'll finde the `*.pot` file here, 
https://github.com/4teamwork/opengever.core/blob/master/opengever/base/locales/plone.pot, please also update the corresponding `*.po` files.

njohner:
@Rotonen @deiferni 
Ok I removed the translations from the `opengever/base/locales/plone.pot`. I rebased and as soon as the tests pass this should be ready.

njohner:
Oh noooo! It was obviously meant to be squashed 👎

njohner:
@deiferni Ok done! Sorry!

PR #3698, opened 2017-11-28
Create initial content for meeting word feature

### LINKED ISSUE #3603: Create initial content for meeting word feature
We don't have initial content for the meeting-word feature yet. So whenever a developer creates a new plone site for development she has to create that content manually. 

We should create word-meeting intitial content in `opengever.examplecontent`. 
The following is a non-exhaustive list of what should be amended:

- at least one proposal template to create new proposals
- committee container has required templates (proposal templates and sablon templates) that are needed, create and link those
- some word-proposals
- maybe re-purpose an existing committee for the word feature and put some submitted word proposals into that committee
- create at least one meeting in aforementioned commitee and add proposal based and also ad-hoc agenda items
- …

tarnap:
Add some initial content for the SPV

with_meeting_word_default_content

Resolves #3603


tarnap:
@deiferni: I left the commits unsquashed for easier review.

opengever/examplecontent/hooks.py
```diff
@@ -1,213 +1,13 @@
-from datetime import date
-from datetime import datetime
-from datetime import time
-from datetime import timedelta
-from ftw.builder import Builder
-from ftw.builder import create
-from ftw.builder import session
-from opengever.base.model import create_session
-from opengever.examplecontent.contacts import ExampleContactCreator
+from opengever.examplecontent.contacts import ExampleContentCreator as ContactExampleContentCreator
```
deiferni: you redefine the names here anyway, why the rename then 🤔 ? wouldn't it make more sense to use `ContactExampleContentCreator` everywhere, also as class name?

opengever/examplecontent/meeting.py
```diff
@@ -0,0 +1,226 @@
...
+    u'Der Rat stellt der Versammlung den Antrag, der Anstellung von Hans '
+    u'Baumann als Sachbearbeiter mit einem Bescha\u0308ftigungsgrad von 90%, '
+    u'Amtsantritt 01.05.2015, zuzustimmen.'
+)
+
+INITIAL_POSITION_2 = u'Der Vorbericht zum Finanzplan wurde erarbeitet und liegt vor.'
+
+PROPOSED_ACTION_2 = u'Die Versammlung nimmt vom Finanzplan 2015 - 2019 Kenntnis.'
+
+INITIAL_POSITION_3 = (
```
deiferni: 💛

opengever/examplecontent/meeting.py
```diff
@@ -180,6 +208,27 @@ def create_meeting(self, delta):
...
+        title = u"Kommission f\xfcr Rechtsfragen, {}".format(
+            api.portal.get_localized_time(datetime=start))
+
+        dossier = create(Builder('meeting_dossier')
+                         .having(title=u'Meeting {}'.format(
+                             api.portal.get_localized_time(start)),)
+                         .within(self.repository_folder_meeting_word))
+        meeting = create(Builder('meeting')
+                         .having(title=title,
+                                 committee=self.committee_law_model,
```
deiferni: please use a different committee here than for the non-word meetings, otherwise it will be hard for us to navigate to the correct meeting. i think it would help to indicate the feature in the committee title, e.g. "Kommission MIT word" or something like that. Same thing for the non-word committee.

opengever/examplecontent/profiles/default/registry.xml
```diff
@@ -6,6 +6,13 @@
     <value>True</value>
   </record>
 
+  <record
+      interface="opengever.meeting.interfaces.IMeetingSettings"
```
deiferni: somehow we use a different way to specify the interface here, i've seen that in policies too. could you group this with the other entry that refers to the same interface, like so:

```
  <records interface="opengever.meeting.interfaces.IMeetingSettings">
    <value key="is_feature_enabled">True</value>
    <value key="is_word_implementation_enabled">True</value>
  </records>
```

opengever/examplecontent/tests/test_hooks.py
```diff
@@ -2,22 +2,14 @@
...
 from opengever.meeting.model import Meeting
 from opengever.meeting.model import Proposal
-from opengever.testing import FunctionalTestCase
+from opengever.testing import IntegrationTestCase
 from plone import api
 from plone.app.testing import applyProfile
 
 
-class TestHooks(FunctionalTestCase):
+class TestHooks(IntegrationTestCase):
```
deiferni: hmm this looks fishy, could you verify that this test is actually executed? i'm almost certain that this test should fail due to changed numbers in the queries …

opengever/examplecontent/tests/test_hooks.py
```diff
@@ -20,6 +20,6 @@ def test_execute_examplecontent_hooks(self):
         applyProfile(api.portal.get(), 'opengever.examplecontent:municipality_content')
         applyProfile(api.portal.get(), 'opengever.examplecontent:init')
 
-        self.assertEqual(4, Proposal.query.count())
-        self.assertEqual(4, AgendaItem.query.count())
-        self.assertEqual(5, Meeting.query.count())
+        self.assertEqual(11, Proposal.query.count())
```
deiferni: ah. forget my previous comment about the test.

tarnap:
@deiferni: I will change the code by the start of the next week. Thanks for your review :)

opengever/examplecontent/meeting.py
tarnap: @deiferni: What if we just rename both committees?
I'm using `committee_assembly_model` for the no-word content and `committee_law_model` for the word content.

tarnap:
@deiferni: Added commits containing some of the changes you requested.
Let's discuss the change requests I haven't accomplished yet after lunch.

opengever/examplecontent/meeting.py
deiferni: yeah i agree, changing the committee title should be enough, if that is what you meant.

tarnap:
@deiferni: what commits shall i / would you squash together?
and... shouldn't we replace the template with the ones we previously created for gever?

deiferni:
@tarnap i'd squad new template commits, tests + functionality and refactorings of same file.
not sure what you mean about replacing the templates 🤔 ?

tarnap:
@deiferni: I'm referencing these files:
https://github.com/4teamwork/opengever.core/pull/3698/files#diff-482bac0eff9a39c7e34115f5d09fde70
https://github.com/4teamwork/opengever.core/pull/3698/files#diff-d0ff67ed6327eb324f0a2b86242758f6
https://github.com/4teamwork/opengever.core/pull/3698/files#diff-cefd877eec5e20950914fe3ae57595cc
https://github.com/4teamwork/opengever.core/pull/3698/files#diff-ff86bee13dae01f72bb768c1cd6050f4
https://github.com/4teamwork/opengever.core/pull/3698/files#diff-bc8c7e0b7f19b290bcea48bd3563f9be
https://github.com/4teamwork/opengever.core/pull/3698/files#diff-d78b6d9d858cf7ff3a1febcd4ff728c8

They're not real templates but rather copy-pastes from the other existing .docx files.
We created some real sablon templates the previous week... which probably belong in here!

deiferni:
@tarnap yes, absolutely! the templates in examplecontent should be our demo/example templates. did not have a look at the templates yet, will have glance once you consider the PR ready.

tarnap:
@deiferni: please re-re-re-review

tarnap:
@deiferni: There was an error in the "zwischentitel.docx".
The template referenced a field which didn't exist.
Please rerererereview

deiferni:
👍  looks good now.

🚧  but as discussed there seems to be an issue with github.com/4teamwork/docxcompose still, we'll wait with the until that one is resolved. otherwise we have default content which produces un-openable protocols and excerpts which may be very annoying.

deiferni:
fix is imminent. we'll merge this now.

PR #3702, opened 2017-11-29
Change repository sharing tests to IntegrationTestCase

njohner:
Not sure why sharing the repository for reading also gives `DossierManager` rights.

### DIFF
--- a/opengever/repository/tests/test_sharing.py
+++ b/opengever/repository/tests/test_sharing.py
@@ -1,38 +1,36 @@
-from ftw.builder import Builder
-from ftw.builder import create
 from ftw.testbrowser import browsing
 from ftw.testbrowser.pages.statusmessages import info_messages
-from opengever.testing import FunctionalTestCase
+from opengever.testing import IntegrationTestCase
 
 
-class TestSharingOnRepositoryFolders(FunctionalTestCase):
-
-    def setUp(self):
-        super(TestSharingOnRepositoryFolders, self).setUp()
-        self.grant('Administrator')
-        self.repo_folder = create(Builder('repository'))
+class TestSharingOnRepositoryFolders(IntegrationTestCase):
 
     @browsing
     def test_add_local_roles_to_user(self, browser):
-        # search for test user
-        browser.login().open(self.repo_folder, view='sharing')
-        browser.fill({'search_term': 'test'})
+        # because of how form fill works combined with the testbrowser
+        # boxes that were already checked for on user will get checked
+        # for all the others when submitting the form. We therefore block
+        # permission inheritance on branch_repofolder then test the local
+        # roleaddition in the leaf_repofolder, to be sure we have a clean
+        # sheet.
+        self.login(self.administrator, browser)
+        self.branch_repofolder.__ac_local_roles_block__ = True
+        browser.open(self.leaf_repofolder, view='sharing')
+        self.assert_local_roles(
+            tuple(), self.regular_user.getId(), self.leaf_repofolder)
+        browser.fill({'search_term': self.regular_user.getId()})
         browser.css('#sharing-save-button').first.click()
-
-        # change local roles and save
-        browser.fill({'entries.role_Reviewer:records': True})
+        browser.fill({'entries.role_Reader:records': True})
         browser.click_on('Save')
-
-        self.assertEquals(['Changes saved.'], info_messages())
-        self.assertEquals(
-            (('test_user_1_', ('Owner', u'Reviewer')),),
-            self.repo_folder.get_local_roles())
+        self.assert_local_roles(
+            (u'Reader',), self.regular_user.getId(), self.leaf_repofolder)
 
     @browsing
     def test_break_permission_inheritance(self, browser):
-        browser.login().open(self.repo_folder, view='sharing')
+        self.login(self.administrator, browser)
+        browser.open(self.branch_repofolder, view='sharing')
         browser.fill({'Inherit permissions from higher levels': False})
         browser.click_on('Save')
 
         self.assertEquals(['Changes saved.'], info_messages())
-        self.assertTrue(self.repo_folder.__ac_local_roles_block__)
+        self.assertTrue(self.branch_repofolder.__ac_local_roles_block__)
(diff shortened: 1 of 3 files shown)


opengever/repository/tests/test_sharing.py
```diff
@@ -1,38 +1,59 @@
...
-    def setUp(self):
-        super(TestSharingOnRepositoryFolders, self).setUp()
-        self.grant('Administrator')
-        self.repo_folder = create(Builder('repository'))
+from opengever.testing import IntegrationTestCase
+
+
+class TestSharingOnRepositoryFolders(IntegrationTestCase):
+
+    def assert_local_roles(self, expected_roles, username, context, method="Equal"):
```
Rotonen: This should go to the `IntegrationTestCase` class as it can be useful to use elsewhere as well.

opengever/repository/tests/test_sharing.py
```diff
@@ -1,38 +1,59 @@
...
+                expected_roles, current_roles,
+                "The user '{}' should have the roles {} on context {}. "
+                "But he has {}".format(username, expected_roles, context, current_roles))
+        elif method == "In":
+            for expected_role in expected_roles:
+                self.assertIn(
+                    expected_role, current_roles,
+                    "The user '{}' should have at least the roles {} on context {}. "
+                    "But he has {}".format(username, expected_roles, context, current_roles))
+        elif method == "NotIn":
```
jone: I'm not yet conviced of this approach with the method. 
This implementation has currently the problem that when I make a typo and, for example, use the method `"notin"`, it will always pass.

It would be awesome if everything could be sovled with an "Equal" method. Wouldn't that be possible?

opengever/repository/tests/test_sharing.py
```diff
@@ -1,38 +1,59 @@
...
-        self.grant('Administrator')
-        self.repo_folder = create(Builder('repository'))
+from opengever.testing import IntegrationTestCase
+
+
+class TestSharingOnRepositoryFolders(IntegrationTestCase):
+
+    def assert_local_roles(self, expected_roles, username, context, method="Equal"):
+        current_roles = []
+        for participant, roles in context.get_local_roles():
```
jone: You could use `dict(context.get_local_roles()).get(username)` instead of having a loop here.

opengever/repository/tests/test_sharing.py
njohner: Ok let me think about it!
Also I don't understand why setting a local role does not change the permissions the user has in the repository.

jone: It is the job of the workflow to decide which role has which permission in which workflow state. Thus adding a role may not have an effect for a specific workflow state, or even for a type at all. It could have an effect on its children though. Certain roles are not relevant at all.

njohner:
* moved the assert_local_roles to the IntegrationTestCase
* Used Jone's workaround  to have a single local role added to the regular_user in the test
Please have another look

njohner:
Oh sorry, just saw I missed one of Jone's comments, I'll quickly integrate his suggestion and let you know when I'm done

njohner:
Ok all done.

opengever/testing/integration_test_case.py
```diff
@@ -375,6 +375,13 @@ def assert_has_not_permissions(self, permissions, obj, msg=None):
                 obj,
                 msg or ''))
 
+    def assert_local_roles(self, expected_roles, username, context):
+        current_roles = dict(context.get_local_roles()).get(username) or []
```
jone: You could use the default argument of the `get` method here:
```diff
-        current_roles = dict(context.get_local_roles()).get(username, [])
+        current_roles = dict(context.get_local_roles()).get(username) or []
```
Just cosmetics. I personally like it better...

opengever/testing/integration_test_case.py
```diff
@@ -375,6 +375,13 @@ def assert_has_not_permissions(self, permissions, obj, msg=None):
...
+        self.assertItemsEqual(
+            expected_roles, current_roles,
+            "The user '{}' should have the roles {} on context {}. "
```
jone: I'd use a repr of the context and roles:
```diff
-            "The user '{}' should have the roles {} on context {}. "
+            "The user '{}' should have the roles {!r} on context {!r}. "
```

opengever/testing/integration_test_case.py
```diff
@@ -375,6 +375,13 @@ def assert_has_not_permissions(self, permissions, obj, msg=None):
                 obj,
                 msg or ''))
 
+    def assert_local_roles(self, expected_roles, username, context):
```
jone: It would be more consistent with other test methods (e.g. `login`) to expect a user object or a user id here:
```diff
-    def assert_local_roles(self, expected_roles, username, context):
+    def assert_local_roles(self, expected_roles, user, context):
+        if hasattr(user, 'getId'):
+            userid = user.getId()
+        else:
+            userid = user
```

opengever/repository/tests/test_sharing.py
```diff
@@ -1,38 +1,36 @@
...
-        browser.login().open(self.repo_folder, view='sharing')
-        browser.fill({'search_term': 'test'})
+        # because of how form fill works combined with the testbrowser
+        # boxes that were already checked for on user will get checked
+        # for all the others when submitting the form. We therefore block
+        # permission inheritance on branch_repofolder then test the local
+        # roleaddition in the leaf_repofolder, to be sure we have a clean
+        # sheet.
+        self.login(self.administrator, browser)
+        self.branch_repofolder.__ac_local_roles_block__=True
```
jone: pep8: spaces around `=` :wink

Rotonen: Seems the QA script does not work in all scenarios - will have to dig back into the 'find branching point' script:
https://ci.4teamwork.ch/builds/128256/tasks/199851

opengever/testing/integration_test_case.py
njohner: Ah I learned something here, didn't know that formatting...

Rotonen: https://pyformat.info/

njohner:
Ok all done

njohner:
@phgross 
Ok checked that all comments had been integrated and rebased.

PR #3703, opened 2017-11-29
Extend OGGBundle with the DossierManager role.

### LINKED ISSUE #3672: Extend OGGBundle with additional property for `dossier_manager` permission
Similar we did for the excel import (see #3652) also the OGGBundle should be extended for repositoryfolders and dossiers.

elioschmutz:
This PR extends the OGGBundler with the `DossierManager` role.

closes #3672

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2017.7.0 (unreleased)
 ---------------------
 
+- Extend OGGBundle with the DossierManager role. [elioschmutz]
 - OGDS sync: handle multivalued group titles. [phgross]
 - Increase group title length. [phgross]
 - Fix bux when accepting a multi admin unit team task with option participate. [phgross]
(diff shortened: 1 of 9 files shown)


elioschmutz:
@phgross I changed the mapping from required to optional (https://github.com/4teamwork/opengever.core/pull/3703/files#diff-90519493faa58540b0bd9761558f1ce3R63). 

Is there a reason, why it was required?

phgross:
@elioschmutz not sure why the mapping was required, @lukasgraf do you remember?

phgross:
@lukasgraf ?

docs/HISTORY.txt
```diff
@@ -5,6 +5,7 @@ Changelog
 2017.7.0 (unreleased)
 ---------------------
 
+- Extend OGGBundler with the DossierManager role. [elioschmutz]
```
lukasgraf: `OGGBundler` -> `OGGBundle`

lukasgraf:
@elioschmutz @phgross no particular reason why `rolemap[role_shortname]` was used. It's BTW not about whether the mapping is optional or required, the behavior before was that *if* a `_permissions` mapping was provided, it needed to list every single permission (`add`, `edit`, ...). But that is unnecessary in my opinion, your change to `rolemap.get(role_shortname, [])` looks good 👍

elioschmutz:
@lukasgraf 👍 updated PR

PR #3704, opened 2017-11-29
Change documents_tab test to IntegrationTestCase

njohner:
The `link.click()` makes the first test slow, but it calls the `redirect_to_parent_dossier` view, which I don't think can be tested simply from the URLs.

### DIFF
--- a/opengever/dossier/tests/test_documents_tab.py
+++ b/opengever/dossier/tests/test_documents_tab.py
@@ -1,42 +1,27 @@
-from ftw.builder import Builder
-from ftw.builder import create
 from ftw.testbrowser import browsing
-from opengever.testing import FunctionalTestCase
+from opengever.testing import IntegrationTestCase
+from opengever.base.behaviors.base import IOpenGeverBase
 
 
-class TestDocumentsTab(FunctionalTestCase):
-
-    def setUp(self):
-        super(TestDocumentsTab, self).setUp()
-
-        self.root = create(Builder('repository_root'))
-        self.repo = create(Builder('repository').within(self.root))
-
-        self.dossier = create(Builder('dossier'))
-        self.subdossier = create(
-            Builder('dossier')
-            .within(self.dossier)
-            .titled(u'S\xfcbdossier <Foo> Bar'))
-        self.document = create(Builder('document').within(self.subdossier))
+class TestDocumentsTab(IntegrationTestCase):
 
     @browsing
     def test_containing_subdossiers_are_linked(self, browser):
-        browser.login().visit(self.dossier, view='tabbedview_view-documents')
-
-        browser.css('table.listing').first.css('a.subdossierLink')
+        self.login(self.regular_user, browser)
+        IOpenGeverBase(self.subdossier).title = u'S\xfcbdossier <Foo> Bar'
+        self.subdocument.reindexObject()
+        browser.open(self.dossier, view='tabbedview_view-documents')
         link = browser.css('table.listing').first.css('a.subdossierLink').first
-        # Title should be HTML escaped and encoded properly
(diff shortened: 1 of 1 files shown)


opengever/dossier/tests/test_documents_tab.py
```diff
@@ -1,42 +1,24 @@
...
-        browser.login().visit(self.dossier, view='tabbedview_view-documents')
-
-        browser.css('table.listing').first.css('a.subdossierLink')
+        self.login(self.regular_user, browser)
+        browser.open(self.dossier, view='tabbedview_view-documents')
         link = browser.css('table.listing').first.css('a.subdossierLink').first
-        # Title should be HTML escaped and encoded properly
-        self.assertEqual(u'S\xfcbdossier &lt;Foo&gt; Bar', link.innerHTML)
-
+        self.assertEqual(u"2016", link.innerHTML)
```
Rotonen: Try `.css().text` instead of `.innerHTML` here.

There is a chance the original test predates `.text`.

njohner:
Ok done.

opengever/dossier/tests/test_documents_tab.py
```diff
@@ -1,42 +1,24 @@
...
-        browser.login().visit(self.dossier, view='tabbedview_view-documents')
-
-        browser.css('table.listing').first.css('a.subdossierLink')
+        self.login(self.regular_user, browser)
+        browser.open(self.dossier, view='tabbedview_view-documents')
         link = browser.css('table.listing').first.css('a.subdossierLink').first
-        # Title should be HTML escaped and encoded properly
-        self.assertEqual(u'S\xfcbdossier &lt;Foo&gt; Bar', link.innerHTML)
-
+        self.assertEqual(u"2016", link.text)
```
phgross: The former tests also assert that the title is e HTML escaped and encoded properly, but your new one does not do this any more. 

change the title of self.subdossier before you open the view:
```
IDossier(self.subdossier).title = u'S\xfcbdossier <Foo> Bar'
browser.open(self.dossier, view='tabbedview_view-documents')
[...]
```

opengever/dossier/tests/test_documents_tab.py
```diff
@@ -1,42 +1,27 @@
...
-        self.document = create(Builder('document').within(self.subdossier))
+class TestDocumentsTab(IntegrationTestCase):
 
     @browsing
     def test_containing_subdossiers_are_linked(self, browser):
-        browser.login().visit(self.dossier, view='tabbedview_view-documents')
-
-        browser.css('table.listing').first.css('a.subdossierLink')
+        self.login(self.regular_user, browser)
+        model.Schema(self.subdossier).setTitle(u'S\xfcbdossier <Foo> Bar')
```
phgross: @njohner Because the title field is part of the IOpenGeverBase behavior which uses a own factory, you have to adapt the object before setting the title. 

The following should work:
```
from opengever.base.behaviors.base import IOpenGeverBase

    @browsing
    def test_containing_subdossiers_are_linked(self, browser):

        IOpenGeverBase(self.subdossier).title = u'S\xfcbdossier <Foo> Bar'
        self.subdocument.reindexObject()

        browser.open(self.dossier, view='tabbedview_view-documents')
        link = browser.css('table.listing').first.css('a.subdossierLink').first
        self.assertEqual(u'S\xfcbdossier &lt;Foo&gt; Bar', link.innerHTML)
        link.click()
        self.assertEqual(browser.url, self.subdossier.absolute_url())

```

njohner: @phgross 
Thanks that also works. I just rebased and everything should be ok.

PR #3707, opened 2017-11-29
Fix manual journal entry link on documents journal.

phgross:
Make sure the manual entry add form is called on the current context, so that the entry is added to the right journal.

Backport-needed: `2017.6-stable`

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 ---------------------
 
 - Remove ZIP export action on plonesite. [phgross]
+- Fix manual journal entry link on documents journal. [phgross]
 - Make sure PDF Preview link is only available for documents not for mails. [phgross]
 - Don't display nochange/remove radio buttons for file in document add-form. [deiferni]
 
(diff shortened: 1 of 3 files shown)


opengever/journal/tab.py
```diff
@@ -104,10 +105,13 @@ def get_base_query(self):
         return None
 
     def manual_entry_allowed(self):
-        return api.user.has_permission('Modify portal content',
+        if IDossierMarker.providedBy(self.context):
```
deiferni: so, what was the issue here? this will only disable the link right? is the functionality to add journal entries still exposed somehow? or did we just render the link in a place where it did not work?

phgross: The link is displayed on the documents journal but links to the journal_entry_form on the dossier, wich adds an entry to the dossier journal. Thats really confusing for the user.

deiferni:
ok dug a bit into this:
- what about repository folders, according to description they should not show the link but they do?
- i can still add journal entries to dossiers by url-guessing since the view is registered for `IJournalizable`.

phgross:
@deiferni updated the PR, as we discussed yesterday. Please have another look

PR #3709, opened 2017-11-30
Explicitly set globstar for the i18n tripwire script

### LINKED ISSUE #3458: i18n tripwire glob doesn't cover packages with three components
This glob doesn't cover package structures deeper than two packages deep:

https://github.com/4teamwork/opengever.core/blob/e8a532fe812fb066b15f03a5967d6cc27ab13909/test-i18n-de.cfg#L49-L50

So packages like `opengever/ogds/base` or `opengever/portlets/tree` aren't matched.

Rotonen:
Shell defaults and implementation details have varied over time and this is the most portable way to ensure the behaviour we seek.

Closes #3458

### DIFF
--- a/test-i18n-de.cfg
+++ b/test-i18n-de.cfg
@@ -18,6 +18,8 @@ recipe = collective.recipe.template
 input = inline:
     #!/bin/bash
 
+    shopt -s globstar
+
     OUTPUT=""
 
     # Test building locales - split into 2 passes for different context lengths


Rotonen:
My bad on the copy paste.

I'm not willing to support bash 3.

Developers who want to run this locally can install bash 4 easily enough.

lukasgraf:
As discussed: WFM 👍

PR #3710, opened 2017-11-30
Add debug view to download docxcompose raw files.

deiferni:
The view offers a zip that contains the following files for a meeting, if available:

- protocol header filled sablon template
- all agenda item documents (ad-hoc or from proposal)
- all paragraph filled sablon templates
- protocol footer filled sablon template

This will help to quickly acquire the documents necessary to debug possible docxcompose issues.

### DIFF
--- a/opengever/meeting/browser/meetings/debug.py
+++ b/opengever/meeting/browser/meetings/debug.py
@@ -1,0 +1,93 @@
+from ftw.zipexport.generation import ZipGenerator
+from ftw.zipexport.utils import normalize_path
+from opengever.meeting import is_word_meeting_implementation_enabled
+from opengever.meeting.command import ProtocolOperations
+from opengever.meeting.protocol import ProtocolData
+from opengever.meeting.sablon import Sablon
+from Products.CMFPlone.utils import safe_unicode
+from Products.Five.browser import BrowserView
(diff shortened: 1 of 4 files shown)


opengever/meeting/browser/meetings/debug.py
```diff
@@ -0,0 +1,93 @@
...
+            self.add_header_sablon(generator)
+            for index, agenda_item in enumerate(self.meeting.agenda_items, 1):
+                self.add_agenda_item(index, agenda_item, generator)
+            self.add_suffix_sablon(index, generator)
+
+            # Return zip
+            response = self.request.response
+            zip_file = generator.generate()
+            filename = '{}.zip'.format(normalize_path(self.meeting.title))
+            response.setHeader(
```
phgross: Just something, maybe for a separate Issue, a general method which sets the headers correctly and returns a filestream_iterator would be nice. There are a lot of those spread around opengever.core.

phgross:
@deiferni please add a short hint to the intern documentation about this view.

opengever/meeting/browser/meetings/debug.py
deiferni: indeed!

Issue #3388, opened 2017-09-18
Unify table selects on the extract attachment view.


deiferni:
For consistency's sake the table-selects of the view to extract attachments from a mail should be normalised to look like the other selects. 
- add comma
- add total count

should look like:
![screen shot 2017-09-18 at 17 12 37](https://user-images.githubusercontent.com/736583/30549231-8d873280-9c94-11e7-94d6-78c085c0c2dd.png)

looks like:
![screen shot 2017-09-18 at 17 12 27](https://user-images.githubusercontent.com/736583/30549230-8d85aa3c-9c94-11e7-885e-a71c5ab57544.png)

phgross:
Closed via #3715

PR #3715, opened 2017-12-01
Normalize appearance of 'all' and 'none' shortcuts.

njohner:
Done!

screen shot 2017-12-01 at 12 59 33

solves #3388

### DIFF
--- a/opengever/mail/browser/templates/extract_attachments.pt
+++ b/opengever/mail/browser/templates/extract_attachments.pt
@@ -51,8 +51,10 @@
               </div>
               <br />
               <p i18n:domain="opengever.tabbedview" >
-                  <span i18n:translate="Choose:">Choose</span>:
-                  <a i18n:translate="all" href="#" id="all">All</a>
+                  <span i18n:translate="Choose">Choose</span>
+                  <a href="#" id="all">
+                    <span i18n:translate="label_tabbedview_filter_all"/>
+                    (<span tal:replace="view/get_number_of_attachments"></span>)</a>,
                   <a i18n:translate="none" href="#" id="none">None</a>
               </p>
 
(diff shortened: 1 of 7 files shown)


opengever/mail/browser/templates/extract_attachments.pt
```diff
@@ -51,8 +51,10 @@
               </div>
               <br />
               <p i18n:domain="opengever.tabbedview" >
-                  <span i18n:translate="Choose:">Choose</span>:
-                  <a i18n:translate="all" href="#" id="all">All</a>
+                  <span i18n:translate="Choose:">Choose</span>
```
phgross: Remove the `:` also from the translation label (do not forget to update the po files 😉 )

njohner: Ah I thought it was there on purpose. At some point someone here (can't remember who), told me that they included things like ":" in the translations, because conventions might be different in one language or another.

phgross: Yes, there could be some translated strings with `:`, but your translation does not longer contain a `:`, therefore  it should also go away from the translation.

njohner:
Ok done.
I also changed **opengever/tabbedview/locales/opengever.tabbedview-manual.pot**, but I'm not sure what that file is used for?

deiferni:
@njohner the translation is placed in the `opengever.tabbedview` domain. the file which contains the translation is in the `opengever.mail` package. our `i18n-build` script will not pick up this translation automatically since it is in a foreign package, that is what the `[domain]-manual.pot` files are for. every msgid in the `[domain]-manual.pot` file is combined with the automatically extracted/managed translations in `[domain].pot`.

njohner:
@deiferni Ok thanks!

PR #3720, opened 2017-12-01
Fix inheritance in test_dossier_byline

njohner:
Another inheritance that made us run 18 instead of 10 tests.


njohner:
Depending on how fast we rewrite all the tests to the IntegrationLayer 😄

PR #3721, opened 2017-12-04
Fix width of dossier and title column in meeting tab

tarnap:
column_width_meeting_tab

Resolves https://github.com/4teamwork/gever/issues/159

### DIFF
--- a/opengever/meeting/tabs/meetinglisting.py
+++ b/opengever/meeting/tabs/meetinglisting.py
@@ -55,10 +55,15 @@ class MeetingListingTab(BaseListingTab):
     columns = (
         {'column': 'committee_id',
          'column_title': _(u'column_title', default=u'Title'),
-         'transform': lambda item, value: item.get_link()},
+         'transform': lambda item, value: item.get_link(),
+         'sortable': False,
+         'groupable': False,
+         'width': 300},
 
         {'column': 'workflow_state',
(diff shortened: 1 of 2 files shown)


deiferni:
_(squash the commits plz 😉 )_

opengever/meeting/tabs/meetinglisting.py
```diff
@@ -81,7 +82,7 @@ class MeetingListingTab(BaseListingTab):
         {'column': 'dossier',
          'column_title': _(u'dossier', default=u'Dossier'),
          'transform': dossier_link_or_title,
-         'sortable': False},
```
phgross: you drop the `sortable:false` here, why`

tarnap: @phgross: do I really drop it?

phgross: Ah you re add it in a later PR ...

tarnap:
@phgross: As far as I was told, the custom settings will remain. These newly added widths are default values (if no custom value was set so far)

deiferni:
@phgross no that is the idea, actually. if a new user opens that table for the first time to column widths are almost certainly too small. we attempt to find a better setting, which may just work and which will give a better first impression 😉 .

PR #3722, opened 2017-12-04
Allow to add proposal documents to tasks

tarnap:
This makes only sense when used with the feature flag 'meeting-word'.
When used without the feature flag, the proposals appear in the listing,
but no documents are associated to it (and cannot be selected).

There is no test, since the only possibility to test this is by testing
the endpoint.
The endpoint was capable of adding proposal documents to tasks already,
and only the listing in the widget was required.

add_proposal_document_to_task

Resolves https://github.com/4teamwork/gever/issues/147


tarnap:
@jone: merci :)

PR #3723, opened 2017-12-04
Port Office Connector API tests to integration tests

Rotonen:
I'm expecting a few rounds of polishing on the exact form of how to name and do things within the tests.

This is laying the groundwork for expanding the Office Connector tests in the near future. Currently this just maintains the current set of things we're testing for.


opengever/officeconnector/testing.py
```diff
@@ -0,0 +1,478 @@
...
+from zope.component import getMultiAdapter
+import json
+import jwt
+
+
+class OCIntegrationTestCase(IntegrationTestCase):
+    """Extend IntegrationTestCase with Office Connector specific helpers."""
+
+    def fetch_document_attach_oc_url(self, browser, document):
+        browser.raise_http_errors = False
```
jone: What's the reason for not raising HTTP errors?

opengever/officeconnector/testing.py
```diff
@@ -0,0 +1,478 @@
...
+        self.assertEquals('download', download)
+
+        filename = payload.get('filename', None)
+        self.assertTrue(filename)
+        self.assertEquals(document.get_filename(), filename)
+
+        title = payload.get('title', None)
+        self.assertEquals(title, document.title_or_id())
+
+        parent_dossier = document.getParentNode()
```
jone: What is `getParentNode`?
For walking up the acquisition chain we use `aq_parent` from `Acquisition`, the official function.

opengever/officeconnector/testing.py
```diff
@@ -0,0 +1,478 @@
...
+        csrf_token = payload.get('csrf-token', None)
+        self.assertTrue(csrf_token)
+
+        document_url = payload.get('document-url', None)
+        self.assertTrue(document_url)
+        self.assertEquals(document.absolute_url(), document_url)
+
+        download = payload.get('download', None)
+        self.assertTrue(download)
+        self.assertEquals('download', download)
```
jone: In which case is `download == 'download'` but not `assertTrue(download)`?
I don't think you need both.

opengever/officeconnector/testing.py
```diff
@@ -0,0 +1,478 @@
...
+
+        documents = token.get('documents', [])
+        self.assertEquals(1, len(documents))
+        self.assertEquals(api.content.get_uuid(self.document), documents[0])
+
+        user = token.get('sub', None)
+        self.assertEquals(self.regular_user.id, user)
+
+        expiry = int(token.get('exp', 0))
+        self.assertTrue(int(time()) < expiry)
```
jone: `self.assertLess` has a better assertion message.

opengever/officeconnector/testing.py
```diff
@@ -0,0 +1,478 @@
...
+            'token': token,
+            }
+
+        return tokens
+
+    def fetch_document_checkout_payloads(self, browser, tokens):
+        browser.raise_http_errors = False
+
+        current_user = api.user.get_current()
+        browser.logout()
```
jone: I don't understand that: why is it necessary to logout / login?
It seems you are just working with some lists an dicts (in `token`).

opengever/officeconnector/testing.py
```diff
@@ -0,0 +1,478 @@
...
+            data=data,
+            headers=headers,
+            view='oc_checkout',
+            ).json
+
+        return payloads
+
+    def validate_checkout_payload(self, payload, document):
+        checkin_with_comment = payload.get('checkin-with-comment', None)
+        self.assertTrue(checkin_with_comment)
```
jone: If this fails, it tells us `None is not true`.
If you remove this line, you'll get `AssertionError: '@@checkin_document' != None`, which gives us a lot more context.

opengever/testing/integration_test_case.py
```diff
@@ -375,6 +377,12 @@ def assert_has_not_permissions(self, permissions, obj, msg=None):
                 obj,
                 msg or ''))
 
+    def assert_journal_entry(self, obj, type, title, entry=-1):
+        entry = get_journal_entry(obj, entry)
+        action = entry.get('action')
+        self.assertEquals(type, action.get('type'))
```
jone: `type` is a builtin and should not be used as variable name.

opengever/officeconnector/testing.py
```diff
@@ -0,0 +1,478 @@
...
+        self.assertTrue(lock_manager.is_locked())
+
+        try:
+            lock_token = ET.fromstring(browser.contents).find(
+                './d:lockdiscovery/d:activelock/d:locktoken/',
+                {'d': 'DAV:'},
+                ).text
+        except ET.ParseError:
+            lock_token = None
+        self.assertTrue(lock_token)
```
jone: You are converting an `ET.ParseError` to an `AssertionError` here: the assertion error will always be raised when an `ET.ParseError` is raised. But why catch the `ET.ParseError` at all?
Now you remove information, so that you only have the info `AssertionError: None is not true` when it fails.

I would not catch the `ET.ParseError` at all and let it bubble up to the test output.
For example:
```python
self.assertTrue(ET.fromstring(...))
```

opengever/officeconnector/testing.py
```diff
@@ -0,0 +1,478 @@
...
+
+        self.login(current_user, browser)
+
+        return file_contents
+
+    def upload_document(self, browser, tokens, payload, document, new_file):
+        browser.raise_http_errors = False
+
+        current_user = api.user.get_current()
+        browser.logout()
```
jone: This pattern for isolating / restoring the security is repead multiple times in this file.
We could easily move this into a context manager:
```python
def upload_document(self, browser, ...):
  with self.isolate_security(browser):
    browser.logout()
    ...
```
or use a decorator:
```python
@isolated_security
def upload_document(self, browser, ...):
    browser.logout()
    ...
```
We could just extract the code from the context manager in `self.login`: https://github.com/4teamwork/opengever.core/blob/master/opengever/testing/integration_test_case.py#L136

opengever/officeconnector/tests/test_api.py
```diff
@@ -1,748 +1,687 @@
...
+
+class TestOfficeconnectorAPIDisabled(OCIntegrationTestCase):
+    @browsing
+    def test_attach_to_email_open_without_file(self, browser):
+        self.login(self.regular_user, browser)
+        self.document.file = None
+
+        oc_url = self.fetch_document_attach_oc_url(browser, self.document)
+        self.assertIsNone(oc_url)
+        self.assertEquals(404, browser.status_code)
```
jone: If `raise_http_errors` haven't been set to `False`, you could do this:
```python
with browser.expect_http_error(404):
    self.fetch_document_attach_oc_url(browser, self.document)
```
That's how the testbrowser is designed to be used for such tests 😉

Rotonen:
@jone ready for rereview.

Most things were artefacts of early development convenience having been left into place before copy pasting the whole scaffold around.

I went with a `as_officeconnector(browser)` context manager as there were a few bits where stuff happens in both contexts.

opengever/officeconnector/testing.py
```diff
@@ -0,0 +1,442 @@
...
+    def as_officeconnector(self, browser):
+        current_user = api.user.get_current()
+        try:
+            yield browser.logout()
+        finally:
+            self.login(current_user, browser)
+
+    def fetch_document_attach_oc_url(self, browser, document):
+        headers = {
+            'Accept': 'application/json',
```
jone: If it makes it simpler, you could set this header for the browser session. See http://ftwtestbrowser.readthedocs.io/en/stable/api_browser.html#ftw.testbrowser.drivers.requestsdriver.RequestsDriver.append_request_header

If you need the browser with and without this header at the same time, you could think about using a clone: http://ftwtestbrowser.readthedocs.io/en/stable/api_browser.html#ftw.testbrowser.core.Browser.clone

Maybe this could be used in `as_officeconnector`?

opengever/officeconnector/tests/test_api.py
```diff
@@ -1,748 +1,802 @@
...
+            self.assertIsNone(oc_url)
+
+    @browsing
+    def test_attach_to_email_open_with_file(self, browser):
+        self.login(self.regular_user, browser)
+
+        oc_url = self.fetch_document_attach_oc_url(browser, self.document)
+
+        self.assertTrue(oc_url)
+        self.assertEquals(200, browser.status_code)
```
jone: If the idea of theese is that you want a `2xx`, you don't need to do that: the testbrowser throws errors for `4xx` and `5xx`, see: http://ftwtestbrowser.readthedocs.io/en/stable/userdoc.html#error-handling
If you want to verify that it is exactly `200`, and not `201` or `204` or such, it should be kept.

opengever/officeconnector/testing.py
Rotonen: I thought about that originally and decided to go with this as there is the checkout/checking stuff with `text/html` until we start using the API for that.

There is also WebDAV in the mix until we do the lock/unlock cycles over the API.

I like the current explicit clarity this is giving me.

opengever/officeconnector/tests/test_api.py
Rotonen: This is indeed to guard against unwarranted 20x as OC does different things on those in some cases.

I'm starting to treat `officeconnector/tests/test_api.py` as the authoritative source of what Office Connector **should** do.

opengever/officeconnector/testing.py
Rotonen: That said when the reality of everything-over-API lands, I'll go for that, but that'll be sometime after TUS support lands into `plone.restapi`.

jone: Fair 👍

Rotonen:
We ready to merge or someone else still want to put more eyes on this?

Rotonen:
Merging now as this is in my way and I'll come back to the nuances thereof soon enough.

PR #3725, opened 2017-12-05
Correct dossier overview tests

njohner:
Tests for the `MeetingDossier` overview were not executed anymore since we moved them to `IntegrationTestCase`. Instead we executed the tests twice for the `Dossier` overview. This test inherits from the `dossier.tests.test_overview.TestOverview`, and used to (when still on the `FunctionalLayer`) redefine the `setUp` method to work on a `MeetingDossier` instead of a `Dossier`. This is not so easily done in the `IntegrationLayer` and had not been taken care of.

Several elements were missing in the fixture to allow testing of the `MeetingDossier` overview. To reuse the test of the `Dossier` for the `MeetingDossier` I needed a similar structure in both the `dossier` and `meeting_dossier`. The `OpengeverContentFixture` was modified as follows:
* Add a document `meeting_document` in the `meeting_dossier`
* Add a task `meeting_task` in the `meeting_dossier`
* Add a sub-task in `meeting_task`
* Related the `closed_meeting_dossier` to both the `dossier` and `meeting_dossier`
* Removed relation between `meeting_dossier` and `dossier`

To run the same tests once on the `dossier` and once on the `meeting_dossier` of the `OpengeverContentFixture`, we defined the tested elements and expected results in the `setUp` method. As we're not allowed to access the `ordnungssystem` in this context, we define them as `str` and then use the `eval` function in the tests. This is a bit of a hack and if you have a cleaner solution let me know!

Of course I also had to adapt a few other tests to account for the changes in the `OpengeverContentFixture`.

### DIFF
--- a/opengever/contentstats/tests/test_content_stats_integration.py
+++ b/opengever/contentstats/tests/test_content_stats_integration.py
@@ -178,11 +178,10 @@ class TestContentStatsIntegration(IntegrationTestCase):
             IStatsProvider, name='file_mimetypes')
 
         self.assertEqual({
-            '': 1,
             'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet': 1,
             'application/vnd.openxmlformats-officedocument.wordprocessingml.document': 11,
             'message/rfc822': 2,
-            'text/plain': 1},
+            'text/plain': 2},
             stats_provider.get_raw_stats())
 
     @browsing
@@ -193,11 +192,10 @@ class TestContentStatsIntegration(IntegrationTestCase):
         table = browser.css('#content-stats-file_mimetypes').first
 
         self.assertEquals([
-            ['', '', '1'],
             ['', 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet', '1'],
             ['', 'application/vnd.openxmlformats-officedocument.wordprocessingml.document', '11'],
             ['', 'message/rfc822', '2'],
-            ['', 'text/plain', '1']],
+            ['', 'text/plain', '2']],
             table.lists())
 
     @browsing
--- a/opengever/contentstats/tests/test_content_stats_integration.py
+++ b/opengever/contentstats/tests/test_content_stats_integration.py
@@ -161,14 +161,14 @@ class TestContentStatsIntegration(IntegrationTestCase):
             (self.portal, self.portal.REQUEST),
             IStatsProvider, name='checked_out_docs')
 
-        self.assertEqual({'checked_out': 0, 'checked_in': 17},
+        self.assertEqual({'checked_out': 0, 'checked_in': 18},
                          stats_provider.get_raw_stats())
 
         # Check out a document
         getMultiAdapter((self.document, self.document.REQUEST),
                         ICheckinCheckoutManager).checkout()
 
-        self.assertEqual({'checked_out': 1, 'checked_in': 16},
+        self.assertEqual({'checked_out': 1, 'checked_in': 17},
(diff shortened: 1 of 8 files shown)


opengever/testing/fixtures.py
```diff
@@ -574,13 +574,29 @@ def transform(self, data):
 
     @staticuid()
     def create_meetings(self):
-        meeting_dossier = self.register('meeting_dossier', create(
+        self.meeting_dossier = self.register('meeting_dossier', create(
```
Rotonen: Should you split adding stuff into the fixtures (+ fixing the resulting failing tests) onto its own commit for a cleaner history (especially clearer for the failing tests having been touched)?

njohner:
Yes that should be clearer.

opengever/contentstats/tests/test_content_stats_integration.py
```diff
@@ -178,6 +178,7 @@ def test_file_mimetypes_provider(self):
             IStatsProvider, name='file_mimetypes')
 
         self.assertEqual({
+            '': 1,
```
deiferni: what is going on here? looks like you created a document without a file. this is valid, but may not have been on purpose.

opengever/dossier/tests/test_overview.py
```diff
@@ -64,74 +77,84 @@ def test_task_box_items_are_filtered_by_admin_unit(self, browser):
...
         self.assertEqual(
             [u'B\xfchler Josef', 'Meier AG'],
             browser.css('#participationsBox li:not(.moreLink) a').text)
 
     @browsing
     def test_task_link_is_safe_html_transformed(self, browser):
         self.login(self.regular_user, browser=browser)
+        test_title = u"Foo <script>alert('foo')</script>"
+        test_title_safe = 'Foo &lt;script&gt;alert(\'foo\')&lt;/script&gt'
+        eval(self.tested_task).get_sql_object().title = test_title
```
deiferni: you say in the PR description you don't have permissions here. have you tried another user, e.g. `self.dossier_responsible`? sorry to say but the `eval` must go away, i'm sure we'll find a way 😉 .

opengever/dossier/tests/test_overview.py
```diff
@@ -17,20 +17,33 @@
 
 class TestOverview(IntegrationTestCase):
 
+    def setUp(self):
+        super(TestOverview, self).setUp()
+        self.tested_dossier = "self.dossier"
```
deiferni: ah now i see what you are doing 😄 . please don't use `eval` though (there is almost certainly a better way to do what you wanna do than `eval`) … this stackoverflow post sums up some important points about that: https://stackoverflow.com/questions/1832940/why-is-using-eval-a-bad-practice.

in your case you could define a `property` that returns the dossiers to test, e.g.

```python
@property
def tested_dossier(self):
  return self.dossier
```

the property can the be overriden in a subclass.

opengever/dossier/tests/test_overview.py
njohner: Yes I just realized I'm stupid. I'll correct that

deiferni: it may also be that the currently used user does not have the permissions required. we have a large set of test-users: https://github.com/4teamwork/opengever.core#users
it may not always be obvious which one to choose best though 😉 .

njohner: Anyway, your solution with defining properties is much cleaner. They get defined just once, whereas the setUp is run for every test...

PR #3735, opened 2017-12-06
Move test_dossier_workflow to IntegrationLayer

njohner:
I think I would need some input here. All 3 tests tested were testing only negative results, which is not very robust. I added a positive test for `test_deleting_dossier_is_only_allowed_for_managers`. For the two others I could not find any situation where these two workflow transitions appear in the action menu.


njohner:
@phgross 
All done. I made the changes according to @deiferni's input and rebased.

PR #3762, opened 2017-12-11
Fix force-checkin for administrators.

### LINKED ISSUE #3756: Force checkin not possible for administrators
For user which only has the roles `Administrator` and `Member` is not possible to force checkin a document. The Checkin action is indeed visible but when trying to checkin a document `Unauthorized Message` is shown.

Analyse:
An administrator does not have a `CMFEditions save new version` permission, which is necessary for creating a new version.

phgross:
Give administrators the `CMFEditions save new version` permission, which is necessary to checkin a document and creating a new version.

Closes #3756.

### DIFF
--- a/opengever/core/profiles/default/rolemap.xml
+++ b/opengever/core/profiles/default/rolemap.xml
@@ -37,6 +37,14 @@
       <role name="Manager" />
     </permission>
 
+    <permission name="CMFEditions: Save new version" acquire="True">
+      <role name="Administrator" />
+      <role name="Manager" />
+      <role name="Contributor" />
+      <role name="Editor" />
+      <role name="Reviewer" />
+    </permission>
+
     <!--migrated from opengever/base/profiles/default/rolemap.xml-->
     <permission name="Remove GEVER content" acquire="True">
       <role name="Manager" />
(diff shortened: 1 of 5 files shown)


opengever/core/profiles/default/rolemap.xml
```diff
@@ -37,6 +37,14 @@
       <role name="Manager" />
     </permission>
 
+    <permission name="CMFEditions: Save new version" acquire="True">
+      <role name="Administrator" />
+      <role name="Manager" />
+      <role name="Contributor" />
```
deiferni: the PR description mentions `Administrator`. i get why we also have `Manager`, but why `Contributor`, `Editor` and `Reviewer`? Is this a copy paste thing? If not, could we add a description why this is necessary?

phgross: The current settings (default by Products.CMFEditions) is:
- Manager
- Site Administrator
- Owner
- Editor
- Reviewer
- Contributor

Do not break existing functionality I just dropped the two roles `Site Administrator` (not used by GEVER at all) and `Owner` (owner should not permits to permissions) and add the missing role `Administrator`. 

The `Editor` role makes sense, he is able to edit a document and checkin/create a new version. Same for the Contributor, he is able to create a document which could be also to create an initial version.

The `Reviewer` role is questionable, should I also drop it?

deiferni: ah i see. you copied this from another package to customize a bit. so this is more or less the configuration we have been using all along. in that case i'd not touch it too much 😉 .

Issue #3726, opened 2017-12-05
Port content stats, indexer tests etc. prone-for-fixture-failures tests to functional tests


Rotonen:
Any tests which break as an unintended side effect of volatility of touching the the integration test fixtures should be ported to funcional tests to stay out of the way.

There are at least a few content stats tests and a few indexer tests peppered around which would be nice to catch.

Ref. https://github.com/4teamwork/opengever.core/pull/3725 for a fine example.

Rotonen:
@phgross FYI I took the liberty of initiative here before @njohner gets frustrated with this particular one.

njohner:
Two other cases that should be made more robust against fixture changes. 
Thanks @Rotonen  

https://github.com/4teamwork/opengever.core/blob/master/opengever/bumblebee/tests/test_overlay_adapter_mail.py#L68

https://github.com/4teamwork/opengever.core/blob/master/opengever/mail/tests/test_indexers.py#L41

Rotonen:
@phgross @njohner am I to coach / collaborate / hijack the branch here?

njohner:
I am just working on other things these days, as this did not look too urgent to me...

Rotonen:
So I'll just hijack this sometime in January.

njohner:
@Rotonen 
I'm jobless just now so I'll start porting these

njohner:
I think the fixture is pretty stable now, and we've ported most tests to the integration layer, so do we still want to port these tests back to the functional layer?

Rotonen:
Once we get the sequential runs and sequential runs of all modules to work again, I have a helper script for shortcutting this locally a bit.

```
function list-gever-modules {
	for package in opengever plonetheme
	do
		for module in "$package"/*/
		do
			echo "$package"."$(basename "$module")"
		done
	done
}

function amend-gever-fixture {
	for module in $(list-gever-modules)
	do
		until bin/test -x --layer integration -m "$module"
		do
			read -r -p 'Fix the failure above and press Enter.'
		done
	done
}

function test-gever-modules {
	for module in $(list-gever-modules)
	do
		until bin/test -x -m "$module"
		do
			read -r -p 'Fix the failure above and press Enter.'
		done
	done
}
```

So yes, this one is obsolete at this point.

PR #3763, opened 2017-12-12
Moved test_content_stats_integration to FunctionalLayer.

njohner:
This test should be on the FunctionalLayer as it is particularly sensitive to changes in the fixture.

I wrote a small "fixture" just for this test and grouped the tests so that the fixture gets executed just once. This makes the test 4 times faster, at the cost of not being able to execute each test separately.

This is part of #3726

### DIFF
--- a/opengever/contentstats/tests/test_content_stats_integration.py
+++ b/opengever/contentstats/tests/test_content_stats_integration.py
@@ -1,16 +1,20 @@
+from ftw.builder import Builder
+from ftw.builder import create
 from ftw.contentstats.interfaces import IStatsKeyFilter
 from ftw.contentstats.interfaces import IStatsProvider
 from ftw.testbrowser import browsing
 from opengever.document.interfaces import ICheckinCheckoutManager
-from opengever.testing import IntegrationTestCase
+from opengever.testing import FunctionalTestCase
+from opengever.mail.tests import MAIL_DATA
+from plone.app.testing import SITE_OWNER_NAME
 from plone import api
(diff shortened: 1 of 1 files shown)


opengever/contentstats/tests/test_content_stats_integration.py
```diff
@@ -155,67 +158,105 @@ def test_review_states_filter(self):
             'explicit assertions for the following workflow states:\n'
             '%r' % (all_possible_workflow_states - covered_states))
 
-    def test_checked_out_docs_stats_provider(self):
-        self.login(self.regular_user)
+
+class TestContentStatsIntegrationWithFixture(FunctionalTestCase):
+
+    def fixture(self):
```
deiferni: this should be in `setUp` so it is called automatically before each test

opengever/contentstats/tests/test_content_stats_integration.py
```diff
@@ -155,67 +158,105 @@ def test_review_states_filter(self):
...
+                           .titled(u'Dokument im Eingangsk\xf6rbli')
+                           .with_asset_file('text.txt'))
+
+        self.mail_eml = create(Builder("mail")
+                               .with_message(MAIL_DATA)
+                               .within(self.dossier))
+
+    @browsing
+    def test_providers(self, browser):
+        self.fixture()
```
deiferni: if you create the fixture in `setUp` this is no longer needed

njohner:
Ok all done.

PR #3764, opened 2017-12-12
Move test_move_items to IntegrationLayer.

njohner:
I've added quite a few elements to the fixture as we needed different elements in different states. We could obviously do without, by changing the state of existing elements instead. This is pretty slow though (around 0.2s or so). I guess in the end this will depend on how many tests use different elements from the fixture.

### DIFF
--- a/opengever/testing/fixtures.py
+++ b/opengever/testing/fixtures.py
@@ -463,7 +463,7 @@ class OpengeverContentFixture(object):
                     bumblebee_asset('example.docx').bytes(),
                     u'vertragsentwurf.docx')))
 
-        task = self.register('task', create(
+        self.task = self.register('task', create(
             Builder('task').within(self.dossier)
             .titled(u'Vertragsentwurf \xdcberpr\xfcfen')
             .having(responsible_client=self.org_unit.id(),
@@ -475,7 +475,7 @@ class OpengeverContentFixture(object):
             .relate_to(self.document)))
 
         self.register('subtask', create(
-            Builder('task').within(task)
+            Builder('task').within(self.task)
             .titled(u'Rechtliche Grundlagen in Vertragsentwurf \xdcberpr\xfcfen')
             .having(responsible_client=self.org_unit.id(),
                     responsible=self.regular_user.getId(),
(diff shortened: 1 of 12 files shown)


opengever/testing/fixtures.py
```diff
@@ -559,12 +559,49 @@ def create_treaty_dossiers(self):
...
+            .titled(u'Vertr\xe4ge abschliessen')
+            .having(responsible_client=self.org_unit.id(),
+                    responsible=self.regular_user.getId(),
+                    issuer=self.dossier_responsible.getId(),
+                    task_type='correction',
+                    deadline=date(2015, 12, 31))
+            .in_state('task-state-resolved')
+            .relate_to(archive_document)))
+
+        inactive_dossier = self.register('inactive_dossier', create(
```
jone: This is currently only used in 1 test.
Do you expect that we will end up with more tests using an `inactive_dossier`?

If we have only 1 test the cost of adding it to the fixture is bigger as when we just deactivate `self.dossier`, if this is possible.

njohner: Yes currently I don't know if it's worth it or not. You can instead change the state of existing elements, but that is pretty slow. I'll try to find out if this would be used in other tests or not

njohner: @jone 
So an inactive dossier is used in the following tests:
* `opengever.basetests.test_archival_file_form.TestEditArchivalFormAccess`
* 4 times in `opengever.dossier.tests.test_activate.TestDossierActivation`
* `opengever.dossier.tests.test_base.TestDateCalculations`
* `opengever.dossier.tests.test_deactivate.TestDossierDeactivation`
* `opengever.dossier.tests.test_navigation.TestNavigation`
* `opengever.dossier.tests.test_resolve.TestResolving`
* 15 times in `opengever.officeconnector.tests.test_api.TestOfficeconnectorAPIDisabled`
* 37 times in `opengever.officeconnector.tests.test_templates`
* 9 times in `opengever/tabbedview/tests/test_dossier_listing`
* `opengever.disposition.tests.test_inactive_dossiers_are_always_not_archival_worthy`
* `opengever.disposition.tests.TestDestructionForNotArchivedDossiers`
* 3 times in `opengever.disposition.tests.TestDispositionEditForm`
* 14 times in `opengever.disposition.tests.TestDispositionOverview`
and `TestContentStatsIntegration` if we port it back to integration layer

Of course I don't know if we can move them all to the `IntegrationLayer`

jone: Ok, I think that sounds reasonable 😉

Rotonen: Since there is a performance gain proposed: split up a new issue for 'port existing integration tests to use the inactive dossier fixture' once this lands.

PR #3789, opened 2017-12-12
Fix dossier responsible widget, in the DossierAddFormView step.

phgross:
When accepting a multi adminunit task and selecting the store and editing in a newdossier step. The dossier responsible field does not work. The AJAX call to the search endpoint ist not working (404).

The reason is that the typename, which is used by the WizzardWrappedAddForm, is not set in the widget_traversal request, therefore I add an fallback to the default dossier type.

Backport-needed to `2017.6-stable`

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 ---------------------
 
 - Bump ftw.bumblebee to skip checksum calculation for unsupported mimetypes. [deiferni]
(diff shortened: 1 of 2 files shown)


phgross:
@deiferni I suppose that the field is broken since the responsible-widget rework (ftw.keywordwidget etc.). Because  it's not usually to change the dossier responsible.

I've tried to add a specific test, unsuccessfully ...

deiferni:
(has already been backported)

Issue #2500, opened 2017-01-30
Split document action buttons onto their own proper shared z3c.form widget


Rotonen:
Splitting off from:

https://github.com/4teamwork/opengever.core/pull/2496

Currently we have essentially duplicate functionality ripe for generalization in how we display the document manipulation options in file overview vs. bumblebee overlay vs. tooltip.

Rotonen:
Note to self also unify the (S)CSS while tackling this.

https://github.com/4teamwork/plonetheme.teamraum/pull/520

Rotonen:
This is effectively superseded by the new ui project.

PR #3799, opened 2017-12-13
Rework UX on checking in locked documents

### LINKED ISSUE #3478: Warnung bei Check-In eines Dokuments wenn es noch ein WebDav-Lock gibt
Wenn ein Benutzer ein Dokument mit Office Connector bearbeitet, so kann er das Dokument immer noch manuell einchecken. Mach er dies, so ist das Lock dann kaputt und muss vom Administrator aufgelöst werden.

Um dies zu verhindern, solle eine Warnung erscheinen wenn ein Benutzer ein Dokument einchecken möchte, welches noch durch einen WebDav-Lock geschützt ist. 

Bei Einchecken mit Kommentar erscheint zwar eine Warnung, jedoch zu wenig aussagekräftig.
 
Bei Einchecken OHNE Kommentar erschien KEINE Warnung, was dann zum angesprochenen Problem führt

mein brief finanzdirektion fd spv

Rotonen:
* As specified in the images in https://github.com/4teamwork/opengever.core/issues/3478
* https://github.com/4teamwork/ftw.keywordwidget/pull/31

Closes #3478


opengever/document/tests/test_overview.py
```diff
@@ -414,6 +414,24 @@ def test_archival_file_is_extended_with_mimetype_class(self, browser):
...
+        browser.open(self.document, view='tabbedview_view-overview')
+        browser.find('Checkout and edit').click()
+
+        lockable = IRefreshableLockable(self.document)
+        lockable.lock()
+
+        # Tabbedview gets in the way of the redirect so we'll have to revisit
+        browser.open(self.document, view='tabbedview_view-overview')
+
+        self.assertNotIn(
```
deiferni: the not here is a bit fragile, maybe better assert what actions must be visible?

opengever/document/tests/test_overview.py
```diff
@@ -2,100 +2,124 @@
...
+class TestDocumentOverviewVanilla(IntegrationTestCase):
+    @browsing
+    def test_overview_displays_related_documents_but_only_documents(self, browser):  # noqa
+        self.login(self.regular_user, browser)
+
+        document_chain_link = create(
+            Builder('document')
+            .within(self.dossier)
+            .having(
+                title=u'F\xf6\xf6',
```
deiferni: no reason to use multiple lines for one parameter, right 😄 ?

opengever/document/tests/test_overview.py
```diff
@@ -2,100 +2,124 @@
...
+                )
+            .relate_to(self.document)
+            )
+
+        document_chain_end = create(
+            Builder('document')
+            .within(self.dossier)
+            .having(
+                title=u'F\xe4\xe4',
```
deiferni: dito.

opengever/document/checkout/checkin.py
```diff
@@ -97,11 +96,37 @@ class CheckinContextCommentForm(form.Form):
...
+                    button = (
+                        'button_checkin_anyway',
+                        button_payload,
+                        )
+
         self.checkin_controller = CheckinController(self.request)
 
     @button.buttonAndHandler(_(u'button_checkin', default=u'Checkin'))
     def checkin_button_handler(self, action):
-        data, errors = self.extractData()
```
deiferni: uhoh, this looks a bit fishy. seems we skip error handling cause there are no errors **now**. id' prefer to always include the error handling code here.

i.e. add

```python
    if errors:
        return
```

also for other occurrences. what do you think?

opengever/document/checkout/controller.py
```diff
@@ -37,5 +38,9 @@ def is_cancel_allowed(self):
                                     ICheckinCheckoutManager)
         if manager:
             return manager.is_cancel_allowed()
-        else:
-            return False
+
+        return False
+
+    def is_locked(self):
```
deiferni: could we define `is_locked` once on the manager and the reuse?

opengever/document/tests/test_overview.py
```diff
@@ -150,267 +227,259 @@ def test_inactive_links_if_document_is_checked_out(self, browser):
...
-        browser.login().visit(self.document, view='tabbedview_view-overview')
+        self.login(self.regular_user, browser)
+
+        browser.open(self.document, view='tabbedview_view-overview')
 
         self.assertEquals(
             'unprotected',
-            browser.css(
-                '#form-widgets-IClassification-classification').first.text)
+            browser.css('#form-widgets-IClassification-classification')
```
phgross: Thats a good example of those situation where the 80 char limit makes the thing just less readable 😄 . 

In my opinion in situations like this, ignoring the 80 char limit makes sense. But's ok for now.

opengever/document/tests/test_overview.py
```diff
@@ -150,267 +227,259 @@ def test_inactive_links_if_document_is_checked_out(self, browser):
...
-
-        browser.login().visit(document, view='tabbedview_view-overview')
+        self.login(self.manager, browser)
+        self.set_workflow_state('dossier-state-resolved', self.dossier)
+        self.set_workflow_state('task-state-tested-and-closed', self.task)
 
-        self.assertEquals(
-            '{}/edit_archival_file'.format(document.absolute_url()),
-            browser.css('#archival_file_edit_link').first.get('href'))
+        browser.open(self.document, view='tabbedview_view-overview')
```
phgross: Should be self.taskdocument

sources.cfg
```diff
@@ -8,6 +8,8 @@ extensions += mr.developer
 development-packages =
   opengever.maintenance
   collective.js.timeago
+# https://github.com/4teamwork/ftw.keywordwidget/pull/31
```
phgross: Could you create a release of ftw.keywordwidget and updated the pinning in the versions.cfg instead.

opengever/document/checkout/checkin.py
```diff
@@ -95,6 +96,18 @@ class CheckinContextCommentForm(form.Form):
...
+            # Add a warning onto a locked document checkin view
+            msg = _(
+                u'label_warn_checkout_locked',
+                default=u' '.join((
+                    'This document is currently being worked on.',
+                    'When you check it in manually you will lose the changes.',
+                    'Please allow for the process to be finished first.'
+                    )),
+                )
+            IStatusMessage(self.request).addStatusMessage(msg, type=u'error')
```
phgross: I would go with `warning` instead of `error`

opengever/document/tests/test_checkout.py
```diff
@@ -363,6 +364,42 @@ def test_single_checkin_with_comment(self, browser):
...
+            'Checkinerino'
+            })
+        browser.css('#form-buttons-button_checkin').first.click()
+
+        manager = getMultiAdapter((self.document, self.portal.REQUEST),
+                                  ICheckinCheckoutManager)
+        self.assertEquals(None, manager.get_checked_out_by())
+
+        # check last history entry to verify the checkin
+        repository_tool = getToolByName(self.document, 'portal_repository')
```
phgross: Use opengever.document.versioner.Versioner.retrieve

opengever/document/browser/actionbuttons.py
```diff
@@ -25,12 +25,15 @@ def is_edit_metadata_link_visible(self):
 
         return True
 
+    def is_locked(self):
```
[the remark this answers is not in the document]
Rotonen: Then I'd need to implement that individually for all the things which can end up using the mixin.

The whole action button stack would need an overhaul with cleanly defined interfaces and a widget.

opengever/document/checkout/checkin.py
Rotonen: I for one welcome scope creeps when they go for quality.

opengever/document/tests/test_overview.py
Rotonen: That way it'd break every time anything unrelated to this test is changed.

We should do a single place where we test the full set of file action button permutations on the full set of data types if we want that.

opengever/document/tests/test_overview.py
Rotonen: `.having()` can take multiple parametres, thus the linediffability preference.

opengever/document/tests/test_overview.py
```diff
@@ -2,100 +2,124 @@
...
-
-    def setUp(self):
-        super(TestDocumentOverview, self).setUp()
-        self.repo, self.repo_folder = create(Builder('repository_tree'))
+class TestDocumentOverviewVanilla(IntegrationTestCase):
+    @browsing
+    def test_overview_displays_related_documents_but_only_documents(self, browser):  # noqa
+        self.login(self.regular_user, browser)
+
+        document_chain_link = create(
```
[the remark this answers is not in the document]
Rotonen: We do not have a chain of 3 related documents within a single dossier in the fixtures. But I can move this to live in the fixtures.

opengever/document/tests/test_overview.py
```diff
@@ -104,44 +128,97 @@ def test_overview_self_checked_out(self, browser):
...
+            'creator Ziegler Robert (robert.ziegler)',
+            document_metadata,
+            )
+
+        self.assertIn(
+            u'Checked out B\xe4rfuss K\xe4thi (kathi.barfuss)',
+            document_metadata,
+            )
+
+        file_actions = browser.css('.file-action-buttons a').text
```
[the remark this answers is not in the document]
Rotonen: I'm not fond of the idea of breaking very many tests all over the place just because we touched the file button actions macro. I'll rather test the minimal set of what the test at hand is actually about. I consider this to be more robust rather than more fragile.

opengever/document/tests/test_overview.py
Rotonen: I chose to break this up silly, can be broken up neater.

opengever/document/checkout/controller.py
Rotonen: It actually already was, amended.

opengever/document/checkout/checkin.py
Rotonen: Actually our logic relies on there being errors, but those not being handled by the form.

Split this off to: https://github.com/4teamwork/opengever.core/issues/3808

Rotonen:
Ready for round 2.

opengever/document/browser/actionbuttons.py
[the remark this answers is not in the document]
Rotonen: https://github.com/4teamwork/opengever.core/issues/2500

opengever/document/tests/test_overview.py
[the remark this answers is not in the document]
phgross: A tests who checks the actions when a file is self_checked out, should definitely fail when theres something added to the actions. Otherwise you limit the value of a test. A tests like this can help you to not forget thinking about a specific situation (a self checked out document in that case). I don't see any bad if tests about the document_actions are failing if you add/remove those actions.

opengever/document/tests/test_overview.py
phgross: Same here. I had several times the situation that I've had to fix a bug, which should be tested, but wasn't because of a simple rename in the action and a `assertNotIn` assertion.

A failing build when implementing a feature rather than a bug in production.

Rotonen:
@phgross tested locally and upon checking in a locked document without a comment the document gets unlocked and checked in. This seems to be the desirable outcome to me.

@phgross @deiferni reworked test logic to avoid `notIn`.

PR #3803, opened 2017-12-13
Have predictable intids in the fixture.

jone:
**Goal:**
We want to have predictable intids in the tests, so that we can assert the intids and also oguids (which are based on intids).

**Thoughts:**
We freeze the time and already a use a static UID generator. The fixture is fragmented into groups, frozen to a specific hour and with a certain UID prefix so that we avoid conflicts while still beeing able to keep the values as static as possible.

Since the ID is a ten-digit integer, we cannot prefix it with a string. We've therefore decided to base the intid generation on the current time, which we expect to be already frozen.

**Implementation:**
The generated ID consists of: `100, hour, minute, second, 00`, wherease the ending `00` is used for conflict resolution incrementation, although this should not happen in the fixture when we don't break it.

The IntIds utility is a persistent component in the local site manager; therefore we need to override it and override it back.

We cannot use a regular wrapper since we would not be able to patch the `._generateId()`, which is called by the `.register()` method. Therefore we inherit the class and share the state while overriding the `._generateId()` method.

### DIFF
--- a/opengever/testing/helpers.py
+++ b/opengever/testing/helpers.py
@@ -7,6 +7,9 @@ from opengever.document.versioner import Versioner
 from plone import api
 from Products.CMFCore.utils import getToolByName
 from Products.PloneLanguageTool.LanguageTool import LanguageBinding
+from zope.component import getUtility
+from zope.component.hooks import getSite
+from zope.intid.interfaces import IIntIds
 import pytz
 import transaction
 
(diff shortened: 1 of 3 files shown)


opengever/testing/helpers.py
```diff
@@ -6,13 +7,42 @@
...
+def time_based_intids():
+    """To ensure predictable IntIds in tests, this context manager patches
+    the IntIds utility so that IntIds are created based on the current time
+    """
+    original_intids = getUtility(IIntIds)
+
+    class TimeBasedIntIds(type(original_intids)):
+
+        def __init__(self):
+            self.__dict__ = original_intids.__dict__
```
tarnap: Why is this line needed?
Since `TimeBasedIntIds` is a subclass of the `original_intids`'s class, the method should be inherited.
Is the original method patched at runtime?

jone: > We cannot use a regular wrapper since we would not be able to patch the ._generateId(), which is called by the .register() method. Therefore we inherit the class and share the state while overriding the ._generateId() method.

The problem is that a regular wrapper does not work here: a regular wrapper overrides certain behavior while still redirecting certain calls to the original object.
This does not work here because the `register()` method internally calls the `_generateId()` method; so the wrapper would only work when we copy the `register()` method as well; and we don't want to do that.

So we subclass the original class and share the state (dict) but override the methods we need to patch.

tarnap: Thank you for the explanation :)

PR #3812, opened 2017-12-18
Delegate task form: Fix required input validation for the recipients field

phgross:
Set missing_value for the `responsible` field in the delegate task form, otherwise the `z3c.form.field.FieldWidgets.extract` those not raise an requiredinput validation error.

### DIFF
--- a/opengever/task/browser/delegate/metadata.py
+++ b/opengever/task/browser/delegate/metadata.py
@@ -12,6 +12,7 @@ from Products.statusmessages.interfaces import IStatusMessage
 from z3c.form.button import buttonAndHandler
 from z3c.form.field import Fields
 from z3c.form.form import Form
+from z3c.form.interfaces import IDataConverter
 from zope import schema
 from zope.interface import provider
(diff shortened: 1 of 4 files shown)


opengever/task/browser/delegate/metadata.py
```diff
@@ -93,15 +93,9 @@ def handle_cancel(self, action):
...
-        portal_state = getMultiAdapter((self.context, self.request),
-                                       name=u"plone_portal_state")
-        member = portal_state.member()
-        if not self.request.get('form.widgets.issuer', None):
-            self.request.set('form.widgets.issuer', [member.getId()])
-
-        super(UpdateMetadataForm, self).update()
+    def updateWidgets(self):
+        super(UpdateMetadataForm, self).updateWidgets()
+        self.widgets['issuer'].value = [api.user.get_current().getId()]
```
deiferni: you could use an `IDataConverter` here to guard against changes in the future, i.e. `widget.value = IDataConverter(widget).toWidgetValue(value)`.

PR #3813, opened 2017-12-18
Update lawgiver to newest version.

jone:
Update `ftw.lawgiver` to the newest version `1.13.0`.

**Restapi permissions:**
Since `ftw.lawgiver = 1.12.0` the restapi permissions are globally ignored by default. This means that lawgiver will not manage those permissions any workflow.
In GEVER we used to manage the permissions in the workflow, which is actually wrong, since it would lead to inconsistency in the API: a client could find ressources (by searching on root) which it would not be allowed to retrieve, since the local workflow would prohibit it.

We are updating our workflows to no longer manager restapi permissions.
Since only the meeting area is affected it is safe to update the security of those objects in an upgrade.

**Translations:**
With `ftw.lawgiver = 1.11.0` the proposed translations have changed: state-IDs are now translated as well.
We actually need to translate both, the workfow state ID and the workflow state title, since in Plone and in our code we use both to translate.
Therefore the translations were updated.


njohner:
Sure I'll look at it tonight or tomorrow morning

njohner:
While I'm at it should I do the others in that file?

PR #3815, opened 2017-12-18
Fix pre-filling committee group id in edit form.

deiferni:
This PR uses the widget to inject initial data instead of the request.

This is a better way of loading data especially since the data is set from a data converter which can also deal with vocabularies that have values that differ from their tokens.


deiferni:
@phgross please re-review, some more changes were necessary.

PR #3822, opened 2017-12-20
Add DocProperty ogg.document.version_number

### LINKED ISSUE #3791: Add DocProperty ogg.document.version_number.
`DefaultDocumentDocPropertyProvider` should be extended and also provide the current document version. The current version can be obtained from a `Document` with `get_current_version_id`. Name of the property: `ogg.document.version_number`.

Also update the docs: https://docs.onegovgever.ch/admin-manual/docproperties/list/#doc-properties-dokument.

njohner:
I decided to use `missing_as_zero=True` to have the property even when we are at version 0. This is not the default behavior for the `get_current_version_id` method.

resolves #3791


opengever/dossier/docprops.py
```diff
@@ -136,7 +136,7 @@ def _add_property(self, properties, name, value):
         If a namespace (NS) has been configured prefixes keys with that
         namespace.
         """
-        if not value:
+        if value == None:
```
deiferni: as `None` is a singleton you can (and should always) use `is None` i.e. identity instead of equality for such statements

njohner:
@deiferni Included your comment and updated the docs.

njohner:
@phgross 
All done and rebased on master.

deiferni:
@njohner please rebase and then merge. thanks!

Issue #3679, opened 2017-11-24
OGIP 17: Workflows for workspace family


jone:
### Principles and thoughts:
- The new workflows are built with `ftw.lawgiver`.
- We use different roles than the regular dossier workflows: the principle is that a workspace is closed by default (participants must be authorized first), whereas a dossier is open by default.

### Document workflow:
- We are using the standard document FTI, which has a dossier-optimized workflow.
- We can either use our own workflow (apply it with a placeful workflow policy) or extend the document workflow to support both in the same workflow.
- For now, we try to keep the same document workflow; this means
  - avoid placeful workflow policy for, as it complicates things
  - use non-conflicting role names in the workspace area
  - maybe rewrite the document workflow with lawgiver at some point

### New workflows:
- `opengever_workspace_root` -> `opengever.workspace.root` #3675
- `opengever_workspace` -> `opengever.workspace.workspace` #3676
- `opengever_workspace_folder` -> `opengever.workspace.folder` #3677

### Roles:
on workspaces root:
- `WorkspacesUser` (Teamräume Benutzer): makes the workspace root visible; allowing to activate the feature only for a certain group
- `WorkspacesCreator` (Teamräume Ersteller): is allowed to create new workspaces.
  - (base on `WorkspacesUser`)

on each workspace:
- `WorkspaceOwner` [1] (Federführung): `WorkspaceAdmin` + manage workspace (edit; close)
  - (base on `WorkspaceAdmin`)
- `WorkspaceAdmin` (Admin): `Contributor` + manage contributors
  - (base on `WorkspaceMember`)
- `WorkspaceMember` (Teammitglied): CRUD on workspace content
  - (base on `WorkspaceGuest`)
- `WorkspaceGuest` (Gast): read workspace content
- these roles base on each other, so that a user only has one of those 4 roles at a given time on a given context.

on folders:
- block inheriting roles so that a folder can be protected
- `WorkspaceOwner` and `WorkspaceAdmin` should probably always have access, I guess
- therefore configurable roles:
  - `WorkspaceMember`
  - `WorkspaceGuest`

### Questions:
- Which admin roles should have acces to the contents of the workspaces? (`Site Administrator`)

[1] We are not using the `Owner` role, so that we can grant permissions on workspace content to the workspace owner. This is not that easy possible with Plone's standard `Owner` role as it is treated specially.

### Todos

- [x] Extend the users in the fixture as necessary.
[...]

PR #3827, opened 2017-12-21
OGIP 17: Add workflows for workspaces and workspace roots

jone:
Implements parts of #3679

Changes:
- add workspace root workflow
- add workspace workflow
- add workspace folder workflow
- extend document workflow with workspace support:
  changes only affect workspace roles; we are not updating the security of non-workspace documents as it is too expensive.
- new roles for workspace module
- configuration of role appearance (sharing page)
- upgrades and tests

Roles / sharing page / permission mappings:
It is important that other workflows are not changed unnecesserily when new workflows are added. In order to achieve that, sharing page role permissions are ignored in workflows of other packages (e.g. committe workflow), so that those workflows do not change.
This is possible in combination with not having set any roles on the root (rolemap.xml); otherwise the roles would appear in the sharing page at least for manager user (we also want to avoid that so that managers cannot do stupid things).
This concept leads to a bit uf redundancy: the sharing page role permissions are repeatedly ignored for each workflow. I think this explicitness is necessery when we don't want to always update all workflows when adding roles. I think this is the best solution at the moment.

These upgrades wont affect existing gever objects.

bildschirmfoto 2017-12-21 um 10 52 23

bildschirmfoto 2017-12-21 um 10 53 13


sources.cfg
```diff
@@ -19,3 +21,5 @@ auto-checkout = ${buildout:development-packages}
 [branches]
 Products.LDAPUserFolder = ftw
 Products.LDAPMultiPlugins = ftw
+# https://github.com/4teamwork/ftw.lawgiver/pull/81
+ftw.lawgiver = jone-describe-roles-with-umlauts
```
jone: 🚧 https://github.com/4teamwork/ftw.lawgiver/pull/81

maethu: was approved ;-)

maethu:
@jone this is awesome!!

PR #3832, opened 2017-12-21
Tidy up XML files in default profile of opengever.core

raphael-s:
- Removes unnecessary comments
- Adds comments to structure the files
- Unify the structure of all the files

`types.xml` and `actions.xml` cant be sorted because it would change the order of things in the gui, and we dont want that.

### DIFF
--- a/opengever/core/profiles/default/actions.xml
+++ b/opengever/core/profiles/default/actions.xml
@@ -1,10 +1,8 @@
 <object xmlns:i18n="http://xml.zope.org/namespaces/i18n" name="portal_actions" meta_type="Plone Actions Tool">
 
-
-  <!-- ###################################################### /-->
+  <!-- DOCUMENT ACTIONS -->
   <object name="document_actions" meta_type="CMF Action Category">
 
-    <!--migrated from opengever/policy/base/profiles/default/actions.xml-->
     <object name="watch" meta_type="CMF Action">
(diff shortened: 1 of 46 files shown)


opengever/core/profiles/default/actions.xml
```diff
@@ -1,36 +1,36 @@
 <object xmlns:i18n="http://xml.zope.org/namespaces/i18n" name="portal_actions" meta_type="Plone Actions Tool">
 
-
-  <!-- ###################################################### /-->
+  <!-- ### DOCUMENT ACTIONS ### -->
   <object name="document_actions" meta_type="CMF Action Category">
 
-    <!--migrated from opengever/policy/base/profiles/default/actions.xml-->
+    <!-- WATCH -->
```
phgross: I would remove the comments for each action, it's just a extraction of the line below and does not bring much value.

raphael-s:
@phgross I updated `actions.xml` and added a changelog entry. The commit history is now clean as well.

raphael-s:
@deiferni Rebase done, please have a look at `rolemap.xml` again before merging. There were many conflicts while rebasing because of https://github.com/4teamwork/opengever.core/commit/eb5348647613d1897a0940b7b4474d60ccf6df98 
I had to completely redo the cleanup in this file.

PR #3833, opened 2017-12-21
Implement cross tab logout mechanism.

bierik:
gever-logout-worker

When a logout is triggered in one gever tab, all other tabs are reloaded
automatically.
This is done through a SharedWorker. All tabs are connected with this
SharedWorker. If one tab sends a logout message to the worker the worker
broadcasts a logout message to all open gever tabs.

According to https://caniuse.com/#feat=sharedworkers this is just
working in Chrome and Firefox.

⚠️ This implementation is not working for cross-site worker invocations. As far as I know the SharedWorkers obey the cross origin security. Maybe a simple CORS header will do the trick.

/cc @Rotonen Thanks for the tip 👍

### DIFF
--- a/opengever/base/browser/resources/logoutClient.js
+++ b/opengever/base/browser/resources/logoutClient.js
@@ -1,0 +1,30 @@
+(function(global, $) {
+  var logoutWorker;
+
+  // Setup the shared worker which holds all open gever tabs
+  function initWorker() {
+    var workerSrc = global.portal_url + "/++resource++opengever.base/logoutWorker.js";
+    logoutWorker = new SharedWorker(workerSrc);
+    // Reload the tab when the LogoutBus sends a broadcast message
+    logoutWorker.port.onmessage = function() { location.reload(); };
+    setupListeners();
+  }
+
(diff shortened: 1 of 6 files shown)


opengever/base/browser/resources/logoutClient.js
```diff
@@ -0,0 +1,30 @@
+(function(global, $) {
+  var logoutWorker;
+
+  // Setup the shared worker which holds all open gever tabs
+  function initWorker() {
+    var workerSrc = global.portal_url + "/++resource++opengever.base/logoutWorker.js";
+    logoutWorker = new SharedWorker(workerSrc);
```
deiferni: what happens on unsupported browsers? did you check that this does not have a negative side-effect.

bierik: I check that here: https://github.com/4teamwork/opengever.core/pull/3833/files#diff-94d23ef5a0b1defbe70af5973c0ce908R27

bierik:
@deiferni CL entry added

PR #3835, opened 2017-12-22
Checkout/Edit only for office connector supported file formats supported

### LINKED ISSUE #3665: Auschecken und Bearbeiten nur bei unterstützen Dokumentformaten
Zur Zeit wird der "Auschecken/Bearbeiten" Knopf bei allen Dokumentformaten dargestellt.
Das macht keinen Sinn, weil die Bearbeitung in Office Connector nur mit bestimmten Formaten/Programmen funktioniert.

njohner:
Only allow Checkout/Edit button for file formats supported by office connector.

If office connector is unavailable or if the file format cannot be handled by office connector:
* Make Checkout/Edit button discreet
* Add tooltip explaining why the function is unavailable (tooltip either says that office connector is unavailable or that the file type is not supported by office connector)
* Add Checkout button

**Remarks**
* I did not find the list of supported file formats for `OfficeConnector`, so the registry needs to be updated
* Would it make more sense to have the registry in the `OfficeConnector` code?

**For a content type that cannot be edited in office connector**
checkout

**For a content type that can be edited in office connector**
checkout_edit

resolves #3665

### DIFF
--- a/opengever/core/profiles/default/registry.xml
+++ b/opengever/core/profiles/default/registry.xml
@@ -82,6 +82,82 @@
 
   <!-- OFFICECONNECTOR -->
   <records interface="opengever.officeconnector.interfaces.IOfficeConnectorSettings" />
+  <record
+      interface="opengever.officeconnector.interfaces.IOfficeConnectorSettings"
+      field="officeconnector_editable_types">
+    <field type="plone.registry.field.List">
+      <value_type type="plone.registry.field.TextLine" />
+    </field>
+    <value>
+      <!-- TextEdit -->
+      <element>text/plain</element>
+      <element>application/rtf</element>
+      <element>text/csv</element>
+      <element>text/css</element>
+      <element>text/xml</element>
+      <element>text/tab-separated-values</element>
+      <element>text/richtext</element>
+      <element>text/html</element>
+
+      <!-- Acrobat Pro, Reader, Preview -->
+      <element>application/pdf</element>
+      <element>application/postscript</element>
+
+      <!-- Preview -->
+      <element>image/jpeg</element>
+      <element>image/jpg</element>
+      <element>image/tiff</element>
+      <element>image/gif</element>
+      <element>image/png</element>
+
+      <!-- MS Visio -->
+      <element>application/vnd.visio</element>
+
+      <!-- MS Project -->
+      <element>application/vnd.ms-project</element>
+
+      <!-- MS Project (generic) -->
+      <element>application/x-project</element>
+
+      <!-- Adobe InDesign -->
+      <element>application/x-indesign</element>
+
+      <!-- Adobe Photoshop -->
+      <element>image/vnd.adobe.photoshop</element>
+
+      <!-- Adobe Illustrator -->
+      <element>application/illustrator</element>
+
+      <!-- MS OneNote -->
+      <element>application/onenote</element>
+
+      <!-- MS Excel -->
+      <element>application/vnd.ms-excel</element>
+      <element>application/vnd.openxmlformats-officedocument.spreadsheetml.sheet</element>
+      <element>application/vnd.openxmlformats-officedocument.spreadsheetml.template</element>
+      <element>application/vnd.ms-excel.sheet.macroEnabled.12</element>
+      <element>application/vnd.ms-excel.sheet.binary.macroEnabled.12</element>
+      <element>application/vnd.ms-excel.template.macroEnabled.12</element>
+      <element>application/vnd.ms-excel.addin.macroEnabled.12</element>
+
+      <!-- MS Powerpoint -->
+      <element>application/vnd.ms-powerpoint</element>
+      <element>application/vnd.ms-powerpoint.addin.macroEnabled.12</element>
+      <element>application/vnd.ms-powerpoint.presentation.macroEnabled.12</element>
+      <element>application/vnd.ms-powerpoint.slideshow.macroEnabled.12</element>
+      <element>application/vnd.openxmlformats-officedocument.presentationml.presentation</element>
+      <element>application/vnd.openxmlformats-officedocument.presentationml.template</element>
+      <element>application/vnd.openxmlformats-officedocument.presentationml.slideshow</element>
+
+      <!-- MS Word -->
+      <element>application/msword</element>
+      <element>application/vnd.ms-word.document.macroEnabled.12</element>
+      <element>application/vnd.ms-word.template.macroEnabled.12</element>
+      <element>application/vnd.openxmlformats-officedocument.wordprocessingml.document</element>
+      <element>application/vnd.openxmlformats-officedocument.wordprocessingml.template</element>
+    </value>
+  </record>
+
 
   <!-- OGDS -->
   <records interface="opengever.ogds.base.interfaces.IAdminUnitConfiguration" />
--- a/opengever/core/profiles/default/registry.xml
+++ b/opengever/core/profiles/default/registry.xml
@@ -155,6 +155,16 @@
       <element>application/vnd.ms-word.template.macroEnabled.12</element>
       <element>application/vnd.openxmlformats-officedocument.wordprocessingml.document</element>
       <element>application/vnd.openxmlformats-officedocument.wordprocessingml.template</element>
+
+      <!-- Apple Numbers -->
+      <element>application/x-iwork-numbers-sffnumbers</element>
+
+      <!-- Apple Keynote -->
+      <element>application/x-iwork-keynote-sffkey</element>
+
+      <!-- Apple Pages -->
+      <element>application/x-iwork-pages-sffpages</element>
+
     </value>
   </record>
 
--- a/opengever/core/upgrades/20180123155812_checkout_edit_only_for_office_connector_editable/upgrade.py
+++ b/opengever/core/upgrades/20180123155812_checkout_edit_only_for_office_connector_editable/upgrade.py
@@ -1,0 +1,9 @@
+from ftw.upgrade import UpgradeStep
+
+
+class CheckoutEditOnlyForOfficeConnectorEditable(UpgradeStep):
+    """Checkout edit only for office connector editable.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
--- a/opengever/document/browser/templates/macros.pt
+++ b/opengever/document/browser/templates/macros.pt
@@ -25,32 +25,48 @@
         <tal:file tal:condition="context/has_file"
             tal:define="preview_supported view/is_preview_supported;
             checkout_and_edit_available view/is_checkout_and_edit_available;
+            office_connector_editable view/is_office_connector_editable;
             copy_download_available view/is_download_copy_available;
             attach_to_email_available view/is_attach_to_email_available;
             checked_out_by_current_user view/is_checked_out_by_current_user;
-            is_cancel_allowed view/is_checkout_cancel_available">
+            is_cancel_allowed view/is_checkout_cancel_available;
+            new_external_edit_enabled context/@@officeconnector_settings/is_checkout_enabled">
 
             <tal:checkout_and_edit_enabled tal:condition="checkout_and_edit_available">
+                <tal:oc_editable tal:condition="office_connector_editable">
+                    <tal:oc_checkout tal:condition="new_external_edit_enabled">
+                        <a tal:attributes="href string:javascript:officeConnectorCheckout('${context/absolute_url}/officeconnector_checkout_url');;;
+                                           data-document-url context/absolute_url"
+                            class="function-edit oc-checkout"
+                            href="#"
+                            i18n:translate="label_checkout_and_edit">
+                            Checkout and edit
+                        </a>
+                    </tal:oc_checkout>
 
-                <tal:oc_checkout tal:condition="context/@@officeconnector_settings/is_checkout_enabled">
-                    <a tal:attributes="href string:javascript:officeConnectorCheckout('${context/absolute_url}/officeconnector_checkout_url');;;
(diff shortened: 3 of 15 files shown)


njohner:
To insert the warning icon after the button description, modifications were made in `plonetheme.teamraum` https://github.com/4teamwork/plonetheme.teamraum/pull/606

Rotonen:
Please:

Rebase your plonetheme.teamraum pull request and get it merged.

Rebase this branch.

Then either check out the current master of `plonetheme.teamraum` with an appropriate comment to indicate why we use that or cut a release of `plonetheme.teamraum`.

njohner:
Ok I rebased the `plonetheme.teamraum` commit and added Kevin as reviewer.

opengever/core/profiles/default/registry.xml
```diff
@@ -150,4 +150,24 @@
   <record name="ftw.datepicker.interfaces.IDatetimeRegistry.various">
     <value>{"dayOfWeekStart": 1, "scrollMonth": false, "scrollTime": false, "scrollInput": false}</value>
   </record>
+
+  <record
+      interface="ftw.tabbedview.interfaces.ITabbedView"
```
deiferni: this is the wrong interface as it is from a foreign package we don't directly control here. i would add the setting as a new field to `opengever.officeconnector.interfaces.IOfficeConnectorSettings`.

opengever/core/upgrades/20171220144115_checkout_edit_only_for_office_connector_editable/registry.xml
```diff
@@ -0,0 +1,21 @@
+<registry>
+  <record
+      interface="ftw.tabbedview.interfaces.ITabbedView"
```
deiferni: same here.

opengever/document/browser/templates/macros.pt
```diff
@@ -25,35 +25,53 @@
...
+                        </span>
+                    </tal:oc_uneditable>
+                    <tal:oc_unavailable tal:condition="not: office_connector_available">
+                        <span class='function-edit-inactive discreet' title="Office connector unavailable"
+                        i18n:attributes="title oc_unavailable_message"
+                        i18n:translate="label_checkout_and_edit">
+                        Checkout and edit
+                        </span>
+                    </tal:oc_unavailable>
+                    <!--<span tal:attributes="class python: (office_connector_available and 'fa-exclamation-triangle-after ' or '') + 'function-edit-inactive discreet';title python: not office_connector_available and 'Office connector unavailable' or 'Office connector unsupported type'"
```
deiferni: please don't add commented out code. if it should not be run it should not be checked in 😉 .

opengever/document/locales/de/LC_MESSAGES/opengever.document.po
```diff
@@ -166,8 +166,9 @@ msgstr "Sie sind dabei, eine Kopie des Dokuments ${filename} herunterzuladen."
 msgid "error_file_and_preserved_as_paper"
 msgstr "Sie haben weder eine Datei ausgewählt noch ist das Dokument in Papierform aufbewahrt, bitte korrigieren."
 
-#. Default: "It's not possible to add E-mails here, please '                    'send it to ${mailaddress} or drag it to the dossier '                    ' (Dragn'n'Drop)."
+#. Default: "It's not possible to add E-mails here, please send it to ${mailaddress} or drag it to the dossier (Dragn'n'Drop)."
 #: ./opengever/document/document.py
+#, fuzzy
```
deiferni: this translation is marked as fuzzy and won't appear. you should not have any fuzzy translations. please check if the translation is ok and then remove the line with the `, fuzzy` in it.

opengever/document/locales/fr/LC_MESSAGES/opengever.document.po
```diff
@@ -168,8 +168,9 @@ msgstr "Vous étes en train de télécharger une copie du document ${filename}."
 msgid "error_file_and_preserved_as_paper"
 msgstr "Vous n'avez pas sélectionné un fichier ni conservé un document en version papier. Merci de corriger cela."
 
-#. Default: "It's not possible to add E-mails here, please '                    'send it to ${mailaddress} or drag it to the dossier '                    ' (Dragn'n'Drop)."
+#. Default: "It's not possible to add E-mails here, please send it to ${mailaddress} or drag it to the dossier (Dragn'n'Drop)."
 #: ./opengever/document/document.py
+#, fuzzy
```
deiferni: dito

opengever/core/profiles/default/registry.xml
```diff
@@ -150,4 +150,24 @@
...
+      field="officeconnector_editable_types">
+    <field type="plone.registry.field.List">
+      <value_type type="plone.registry.field.TextLine" />
+    </field>
+    <value>
+      <element>application/vnd.ms-excel</element>
```
deiferni: what is the motivation of putting the supported mime-types in the registry vs. a hardcoded constant in a python module?

njohner: Hum, that was just my interpretation of @phgross 's comment on the issue (https://github.com/4teamwork/opengever.core/issues/3665#issuecomment-348153350)
Maybe @phgross can comment on this?

deiferni: Ah in the meantime i figured out why that may be desired. It can help to avoid a late error message on the client from office-connector when certain applications are not installed there. So 👍  for this implementation.

njohner: If you say so 😄

opengever/document/locales/de/LC_MESSAGES/opengever.document.po
njohner: Ok, this came from the default string which was multiline but not protected.

njohner:
@deiferni 
No, this is not ready.
We started discussing with @Rotonen  and @buchi . Buchi thinks it does not make sense to have a registry of supported formats in Gever. We would also still have the issue that if a user does not have an application for a file format supported by office connector, then the "checkout and edit" would be there but the action would fail.
So in the long term we would need a solution where `officeconnector` and `opengever` communicate and @Rotonen has some ideas on how to do this (but not now...)
So I'm not sure what we want to do here. In case we still want this partial fix, I would need to:
* Have two registries of supported file formats, one for mac and one for windows.
* Write a test that extracts the supported applications from `officeconnector` and compares that with the registry in `opengever`

Rotonen:
I'm not fond of the idea of admin-configurable registries for this. I prefer hardcoding the common set and the (up to) two supersets into a class in this case. This would require the featureset releases of `opengever.core` and Office Connector to happen in tandem in the future and I'd actually prefer that to be the case for easier plannability.

We can already have a go at the two-way communication, but the major blocker there is the lack of ability to inform the user of what is going on. What will take the brunt effort is building the frontend and UX stuff through to the user as things happen. We already have the foundations of this in the polling spinner @bierik did - we just need to design the whole thing well and such niceties are not yet on the plate as we're still fighting with the core basics. Thus my earlier estimate of getting to this sometime in the August / September timeframe.

But as for the feature itself in the here and now: IMO doing this now is both harmless and better than what we're doing now. It does not catch all the corner cases, but setting the bar that high is a bit too much as this should be reusable as-is when already in place when tackling the other facets of this.

Stepping stone upgrades such as this are fine by me.

deiferni:
@Rotonen my first reaction was also to think of a hardcoded set. But the set of supported applications does not always match the set of installed applications. Having a registry will allow us to configure the supported apps per customer, especially larger customers might prefer this approach. Also this would allow us to sidestep the two separate registries for mac and windows as the customer environments are usually homogenous and don't mix mac and windows. This approach does not cover all the edge cases of course but i think it is still an improvement over the current state where the link is always displayed.
Once we have  two-way communication the registry can be dropped again, but as you said this might take a while and we're contractually bound to deliver a solution for this problem in `2018.1`.

For the default registry value i would suggest to:
- make it a set of all supported types of officceconnector (mac + windows)
- as suggested write a test that extracts the supported apps from the latest officeconnector and makes sure the set is equal. @Rotonen how would we do this best at the moment?

Rotonen:
@deiferni submodule the latest office connector release tag and import from there.

deiferni:
For the sake of completeness the supported applications can be found here:
https://github.com/4teamwork/OfficeConnector/blob/master/officeconnector/darwin/editors/__init__.py
https://github.com/4teamwork/OfficeConnector/blob/master/officeconnector/win32/editors/__init__.py

njohner:
@deiferni @Rotonen 
Ok so I completed the list of supported mimetypes as best I could. I now see the difficulty there. It's hard to know what file formats can be opened by the supported applications and of course several applications can read the same file formats... 
I also had to add mimetypes for the mac IWorks series of softwares and the corresponding upgrade step. For the upgrade step I had to really reset the mimetype. I had found a similar upgrade where objects were just reindexed (https://github.com/4teamwork/opengever.core/commit/420d5c203862a2455d4b23db21de64393a7f8416), but that did not work for me.

opengever/core/upgrades/20180123155812_checkout_edit_only_for_office_connector_editable/upgrade.py
```diff
@@ -0,0 +1,38 @@
...
+    def _update_document_metadata(self):
+        """Update the mimetype on documents.
+        """
+        catalog = api.portal.get_tool('portal_catalog')
+        brains = catalog.unrestrictedSearchResults(
+            portal_type='opengever.document.document')
+        print("brains",len(brains))
+        # A getIcon value of 'application.png' indicates that the mimetype
+        # coulnd't be resolved for this document - only consider these
+        brains = filter(lambda b: b.getIcon == 'application.png', brains)
```
deiferni: I'd think about dropping the upgrade that sets the mimetypes. We have added mimetypes in the past without updating existing objects: https://github.com/4teamwork/opengever.core/pull/3660/. 

As we have deployments with many, many documents i'd say not upgrading is preferable.

opengever/document/browser/templates/macros.pt
```diff
@@ -25,36 +25,50 @@
...
                         class="function-edit oc-checkout"
                         href="#"
                         i18n:translate="label_checkout_and_edit">
                         Checkout and edit
                     </a>
-                </tal:oc_checkout>
+                </tal:oc_editable>
 
-                <tal:zem_checkout tal:condition="not: context/@@officeconnector_settings/is_checkout_enabled">
+               <tal:oc_uneditable tal:condition="python: not (office_connector_available and office_connector_editable)">
```
lukasgraf: This whole part is wrong I think. This doesn't do what's being claimed in the Pull-Request description (show checkout action if file type is unsupported), but instead it does this:

- With direct checkout registry flag disabled:
  - Disables the "checkout and edit" action, claiming OC is unavailable
  - Renders the ZEM checkout and edit action with just the label "checkout", instead of "checkout and edit

- With direct checkout registry flag enabled, for unsupported types:
  - Disables the "checkout and edit" action
  - Renders the ZEM download action, but labels it "checkout"

This is all very confusing, I don't think this is what was intended.

Rotonen: As you now have a larger tree of options of what can happen you'll have to build that tree in TAL as well, it's not ZEM vs. URL anymore, but you'll have to handle the mimetype stuff for both.

opengever/core/upgrades/20180123155812_checkout_edit_only_for_office_connector_editable/upgrade.py
njohner: This would mean that external edit will not be available for those documents. But I guess that if they get checked out and back in once, the mimetype is reassigned.

opengever/document/browser/templates/macros.pt
njohner: Ok there was quite some misunderstanding on my side. Should be better now...

njohner:
Open questions/things:
* Remove mimetype assignment from upgrade step? This will disable external edit for these files
* Icons: I slacked @bierik

njohner:
@Rotonen @deiferni @lukasgraf 
I made the  requested changes. We now have:
* `Checkout and Edit` button when the file format is supported.
  * oc url if `direct_checkout_and_edit` is enabled
  * zem download if `direct_checkout_and_edit` is disabled
* Inactive `Checkout and Edit` button and a `Checkout` button when the file format is not supported.

I added a test to check for the presence of the inactive `Checkout and Edit` and the `Checkout` buttons when the file format is not supported.

deiferni:
ok
- i think we can merge before we get the icons, they can be added later. we'll get the default icon in the meantime i suppose?
- for the upgrade i'd remove the upgrade step nevertheless and create a maintenance script in https://github.com/4teamwork/opengever.maintenance/tree/master/opengever/maintenance/scripts to perform the changes. that way we can decouple it from the upgrade and only run it on demand i.e. for customers that actually have files of this type. please get back to me if you need more information here.

lukasgraf:
@njohner looks good! However, when testing locally, the newstyle office connector checkout and edit doesn't work for me (404 from OC client), whereas ZEM checkout and edit works.

Can you please re-verify locally that this works for you?

_Re: Icons: I also think we can ship this while new icons aren't done yet._

Rotonen:
@lukasgraf went through the delta here and cannot spot anything which would have changed the URL or the payload fetched based on the URL. None of that logic is touched. That could mean something else is broken.

njohner:
@lukasgraf were you logged in as zopemaster by any chance? That raises 404...

deiferni:
@njohner that has been fixed as of https://github.com/4teamwork/opengever.core/pull/3893, thus the rebase request 😉 .

njohner:
Nice!

PR #3836, opened 2017-12-27
Prepare HTML structure for dossier title alignment.

bierik:
screen shot 2017-12-27 at 11 25 54


bierik:
@phgross added changelog

Issue #3807, opened 2017-12-14
Send documents form: focus not on first field


phgross:
Focus is on the field `Externe Empfänger` but should be on the first field `Interne Empfänger`

<img width="877" alt="bildschirmfoto 2017-12-14 um 12 02 15" src="https://user-images.githubusercontent.com/485755/33989124-ab6c369e-e0c6-11e7-91b4-272ed16e10c0.png">

PR #3837, opened 2017-12-27
Fix autofocus on forms with keywordwidget.

bierik:
The auto focus was just limited to textareas and text input elements.

Closes https://github.com/4teamwork/opengever.core/issues/3807


bierik:
@phgross updated

PR #3838, opened 2017-12-27
Display document mimetype icon on livesearch results.

bierik:
screen shot 2017-12-27 at 15 07 17
screen shot 2017-12-27 at 15 07 30

Closes https://github.com/4teamwork/opengever.core/issues/3732

### DIFF
--- a/opengever/advancedsearch/skins/advancedsearch_resources/livesearch_reply.py
+++ b/opengever/advancedsearch/skins/advancedsearch_resources/livesearch_reply.py
@@ -49,6 +49,24 @@ def quote_bad_chars(s):
         s = s.replace(char, quotestring(char))
     return s
 
+
+# https://github.com/4teamwork/opengever.core/blob/master/opengever/base/browser/helper.py#L20
+def get_mimetype_icon_klass(item):
+    # It just makes sense to guess the mimetype of documents
+    if not item.portal_type == 'opengever.document.document':
+        return ''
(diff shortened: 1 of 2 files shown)


opengever/advancedsearch/skins/advancedsearch_resources/livesearch_reply.py
```diff
@@ -164,8 +181,14 @@ def write(s):
             display_title = full_title
 
         full_title = full_title.replace('"', '&quot;')
-        klass = 'contenttype-%s' % ploneUtils.normalizeString(result.portal_type)
-        write('''<a href="%s" title="%s" class="%s">%s</a>''' % (itemUrl, full_title, klass, display_title))
+
+        klass = []
```
maethu: I now it is picky: But why is this a list? With a if/else you never get more than one item in the list. now you may end up with one empty string and a content type klass. 

I would prefer something like:

```
css_klass = 'contenttype-%s' % ploneUtils.normalizeString(result.portal_type)
mime_type_klass = get_mimetype_icon_klass(result)
if mime_type_klass:
    css_klass = mime_type_klass

write('''<a href="%s" title="%s" class="%s">%s</a>''' % (itemUrl, full_title, css_klass, display_title))

bierik:
@maethu updated

bierik:
@maethu comment added

PR #3839, opened 2017-12-27
Save current tree portlet index in localstorage.

bierik:
Closes https://github.com/4teamwork/opengever.core/issues/3397


bierik:
@phabegger upated

PR #3847, opened 2018-01-05
Change msg2mime transform to use msgconvert executable from $PATH

lukasgraf:
Change `msg2mime` transform to use `msgconvert` executable from `$PATH` instead of shipping our own wrapper script.

This is the way the `Email::Outlook::Message` is intended to be used with the new package installed via puppet.

However, this approach should also be backwards compatible with the old way we used to ship the module (using a `local::lib` install):

If the module was installed properly using the old method,

- the `local::lib` shim should be present and
- the `msgconvert` executable should be in `$PATH` since the module was installed using CPAN.

If nevertheless the executable can't be found in `$PATH`, a message will be logged to Sentry.


lukasgraf:
Tested on `lab` 👍

lukasgraf:
@phgross good catch, thanks! Updated README.

PR #3848, opened 2018-01-05
Add action to revive bumblebee previews.

### LINKED ISSUE #3619: Add view and button to revive the bumblebee preview.
Sometimes creating a bumblebee preview image/thumbnail/overlay image fails and in rare circumstances this leaves bumblebee in an inconsistent state (GEVER-side). Also due to a bug some cached bumblebee document checksums may not be up to date, see https://github.com/4teamwork/opengever.core/pull/3617 for example. It also may happen that the bumblebee service is down temporarily and the documents can't be registered.

At the moment these documents won't show a preview until the document is modified and thus triggers a bumblebee update. 

We would like to add a button to "revive" the preview as a stop-gap measure so that an `Administrator` or `Manager` can attempt to reset bumblebee state and make sure a preview image is rendered again.

"reviving" should perform the following actions:
- re-calculate bumblebee checksum
- store in bumblebee

can most likely be implemented by using something from`IBumblebeeDocument`.

elioschmutz:
This PR adds the possibility to revive a bumblebee document.

screen2

See #3619 for more information.

closes #3619

### DIFF
--- a/opengever/bumblebee/browser/revive_preview.py
+++ b/opengever/bumblebee/browser/revive_preview.py
@@ -1,0 +1,63 @@
+from ftw.bumblebee.interfaces import IBumblebeeable
+from ftw.bumblebee.interfaces import IBumblebeeDocument
+from opengever.bumblebee import _
+from opengever.bumblebee import is_bumblebee_feature_enabled
+from plone import api
+from Products.Five import BrowserView
+
+
+class RevivePreview(BrowserView):
+    """Sometimes creating a bumblebee preview fails and in rare circumstances
+    this leaves bumblebee in an inconsistent state (GEVER-side).
+
+    It also may happen that the bumblebee service is down temporarily and the
+    documents can't be registered.
+
+    These documents won't show a preview until the document is modified and
+    thus triggers a bumblebee update.
+
+    This view lets you fix this inconsistent bumblebee document either through
+    calling the view directly in the browser with
+
+    @@revive_preview
+
+    or you use the action on the object to revive the current object.
+    """
+    def __call__(self):
+        """Handles the reviving process.
+        """
+        if self.available():
+
+            # Indicates that this context should not be converted if the
+            # checksum didn't change. Even the force attribute respects this
+            # marker-attribute.
+            #
+            # Because we really want to update the document, we just remove
+            # this property.
+            if hasattr(self.context, '_v_bumblebee_last_converted'):
+                delattr(self.context, '_v_bumblebee_last_converted')
+
+            # Forces a document-update. Means, regenerating the checksum
+            # and queue storing in bumblebee.
+            IBumblebeeDocument(self.context)._handle_update(force=True)
+
+            api.portal.show_message(message=_(u'preview_revived',
+                                              default=u'Preview revived and will be available soon'),
+                                    request=self.request, type='info')
+        else:
+            api.portal.show_message(message=_(u'preview_revive_not_available',
+                                              default=u'Reviving the preview is not available for this context.'),
+                                    request=self.request, type='error')
+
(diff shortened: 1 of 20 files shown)


opengever/bumblebee/browser/revive_preview.py
```diff
@@ -0,0 +1,64 @@
...
+            # Forces a document-update. Means, regenerating the checksum
+            # and queue storing in bumblebee.
+            IBumblebeeDocument(self.context)._handle_update(force=True)
+
+            api.portal.show_message(message=_(u'preview_revived',
+                                              default=u'Preview revived and will be available soon'),
+                                    request=self.request, type='info')
+        else:
+            api.portal.show_message(message=_(u'preview_revive_not_available',
+                                              default=u'Revive the preview is not available for this context.'),
```
deiferni: Reviving

opengever/bumblebee/browser/revive_preview.py
```diff
@@ -0,0 +1,64 @@
...
+
+    This view lets you fix this inconsistent bumblebee document either thorugh
+    calling the view directly in the browser with
+
+    @@revive_preview/revive
+
+    or you use the action on the object to revive the current object.
+    """
+
+    def revive(self):
```
deiferni: or just `__call__`? then calling this via browser would be easier as one has only to remember `revive_preview`.

opengever/bumblebee/browser/revive_preview.py
```diff
@@ -0,0 +1,64 @@
...
+        """
+        if self.available():
+
+            # Indicates that this context should not be converted if the
+            # checksum didn't change. Even the force attribute respects this
+            # marker-attribute.
+            #
+            # Because we really want to update the document, we just remove
+            # this property.
+            if hasattr(self.context, '_v_bumblebee_last_converted'):
```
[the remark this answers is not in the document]
elioschmutz: From the Zope-Docs: https://docs.zope.org/zope2/zdgbook/ZODBPersistentComponents.html

> The third rule is that all object attributes that begin with ‘_v_’ are “volatile” and are not saved to the database. This means that as long as the persistent object is in Zope memory cache, volatile attributes can be used. When the object is deactivated (removed from memory) volatile attributes are thrown away.

I think thats the reason, why the volatile-attribute is still on the object. The object comes from the memory.

opengever/bumblebee/browser/revive_preview.py
elioschmutz: I thought `<property name="available_expr">object/@@revive_preview/available</property>` is calling the object first before calling the method... but I tested it and it doesn't 🤔 

So I changed it to revive directly on calling the object.x

opengever/bumblebee/browser/revive_preview.py
```diff
@@ -0,0 +1,64 @@
...
+    """Sometimes creating a bumblebee preview fails and in rare circumstances
+    this leaves bumblebee in an inconsistent state (GEVER-side).
+
+    It also may happen that the bumblebee service is down temporarily and the
+    documents can't be registered.
+
+    These documents won't show a preview until the document is modified and
+    thus triggers a bumblebee update.
+
+    This view lets you fix this inconsistent bumblebee document either thorugh
```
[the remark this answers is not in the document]
elioschmutz: 👍

opengever/bumblebee/browser/revive_preview.py
elioschmutz: 👍

opengever/bumblebee/browser/revive_preview.py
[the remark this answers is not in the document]
deiferni: 🤔  mhmm. It can't hurt to keep it here.

PR #3851, opened 2018-01-09
Sort query only if column exists

### LINKED ISSUE #3468: Fehler bei Anzeige der Anträge in Dossiers
Unter dem Reiter "Anträge"  (In beliebigen Dossiers) lädt und lädt die Seite ohne Ergebnisse zurück zu geben. Problem lässt sich lösen in dem man via dem Zahnrädchen die Tabellenkonfiguration zurücksetzt.

Für weitere Infos siehe: https://sentry.4teamwork.ch/sentry/onegov-gever/issues/3581/

tarnap:
The tabbed view allows to configure the sorting of tables for each user.
If after a migration the column the user used to sort the data is not available anymore, the server responds with an error.
This prevents sorting by unavailable columns.

Resolves #3468

### DIFF
--- a/opengever/tabbedview/sqlsource.py
+++ b/opengever/tabbedview/sqlsource.py
@@ -24,6 +24,36 @@ def cast_to_string(field):
     return cast(field, String)
 
 
+def sort_column_exists(query, sort_on):
+    for _column in query.column_descriptions:
+
+        column_exists = (
+            not _column.get('aliased') and
+            str(sort_on) in _column.get('entity').__table__.columns or
+            False
+        )
+
+        entity_exists = (
+            not _column.get('aliased') and
+            str(sort_on).replace(
+                '{}.'.format(_column.get('name')),
+                ''
+            ) in _column.get('entity').__table__.columns or
+            False
+        )
+
+        alias_exists = (
+            _column.get('aliased') and
+            str(sort_on) in _column.get('name') or
+            False
+        )
+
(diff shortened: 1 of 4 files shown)


opengever/tabbedview/sqlsource.py
```diff
@@ -69,7 +69,24 @@ def extend_query_with_ordering(self, query):
 
             order_f = self.config.sort_reverse and desc or asc
 
-            query = query.order_by(order_f(sort_on))
+            # Apply sorting only if the column exists
+            do_sort = any([  # column
```
deiferni: this is a bit hard to read for me. suggestions to improve:
- the whole thing iterates all columns in each sql query three times. we're interested in the first `True`. Could you rewrite with one loop and `return` once you found what you're looking for?
- put this hole thing into its own method (`is_valid_sort_column` or something like that)

tarnap: @deiferni: There's a commit with the changes you requested for review. I'll squash it into the previous commit once you approve it.

tarnap:
@deiferni: Thank your for re-reviewing

tarnap:
@deiferni: I will need the different kinds of relations for the test. Will you help me build them up (of find existing ones inside the current application?)

deiferni:
@tarnap yes 😉

tarnap:
@deiferni: please re-re-re-re-re-review this pull request :)

PR #3868, opened 2018-01-16
Show documents and tasks on repository folders.

jone:
At the moment it is not possible to list all documents or all tasks for a specific repository folder. This is especially annoing when trying to figure out how many / which tasks are open in a specific area of the repository. As user I'm sometimes looking for a document but I'm not sure in which dossier it is stored, for example because I'm not responsible / used to this area of the repository.

## Solution:
Adding the tabs "documents" and "tasks" on the repository folders will resolve that problem and give the users more flexibility. It also helps users to navigate faster because there is no need to open the dossier first.

## Implementation details:
- Both tabs can be individually disabled with feature flags. The tabs are enabled by default and will be automatically enabled on update.
- The documents tab has a new dossier column; the task tab already had one.
- The actions were cleaned up so that only actions are provided which actually work from the repository folder context.

## Screenshots:
bildschirmfoto 2018-01-16 um 14 45 27
bildschirmfoto 2018-01-16 um 14 44 38

## Discussion:
The feature has been discussed in the feedback forum: https://feedback.onegovgever.ch/t/dokumente-und-aufgabenuebersicht-auf-stufe-ordnungsposition-aktivieren/523/9

- **Performance:** There was the objection that this might by very imperformant on top-level repository folders on very large installations. Therefore the tabs can be disabled by feature flag.

- **Visual similarity with dossier:** When the tabs were initially designed we did not include those tabs on repository folders in order to have a visual unambiguousness between repository folders and dossiers. This is problem is not anymore valid since we have changed the design completely in recent years and we have added water marks in the background.

### DIFF
--- a/opengever/repository/browser/tabs.py
+++ b/opengever/repository/browser/tabs.py
@@ -1,0 +1,28 @@
+from opengever.tabbedview import _ as tabbedview_mf
+from opengever.tabbedview.browser.tabs import Documents
+from opengever.tabbedview.helper import linked_containing_maindossier
+
+
+class RepositoryFolderDocuments(Documents):
+
+    disabled_actions = (
+        'zip_selected',  # disabled because ZIP does not work on repository folders
+        'send_as_email',  # target not available on repository folders
+        'trashed',  # target not available on repository folders
+        'create_task',  # cannot create task in repository folders
+    )
+
+    @property
+    def enabled_actions(self):
+        actions = super(RepositoryFolderDocuments, self).enabled_actions
+        return [action for action in actions if action not in self.disabled_actions]
+
+    @property
+    def columns(self):
+        columns = list(super(RepositoryFolderDocuments, self).columns)
+        after_column = filter(lambda col: col['column'] == 'containing_subdossier', columns)[0]
+        columns.insert(columns.index(after_column),
+                       {'column': 'containing_dossier',
+                        'column_title': tabbedview_mf('containing_dossier', 'Dossier'),
+                        'transform': linked_containing_maindossier})
+        return columns
--- a/opengever/repository/tests/test_repositoryfolder_tabs.py
+++ b/opengever/repository/tests/test_repositoryfolder_tabs.py
@@ -9,12 +9,76 @@ class TestRepositoryFolderTabs(IntegrationTestCase):
     def test_visible_tabs_for_regular_users(self, browser):
         self.login(self.regular_user, browser)
         browser.open(self.branch_repofolder)
-        self.assertEquals(['Dossiers', 'Info'],
+        self.assertEquals(['Dossiers', 'Documents', 'Info'],
                           tabbedview.tabs().text)
 
     @browsing
     def test_visible_tabs_for_administrator(self, browser):
         self.login(self.administrator, browser)
         browser.open(self.branch_repofolder)
-        self.assertEquals(['Dossiers', 'Info', 'Protected Objects'],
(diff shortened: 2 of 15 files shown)


opengever/repository/browser/tabs.py
```diff
@@ -0,0 +1,28 @@
+from opengever.tabbedview import _ as tabbedview_mf
+from opengever.tabbedview.browser.tabs import Documents
+from opengever.tabbedview.helper import linked_containing_maindossier
+
+
+class RepositoryFolderDocuments(Documents):
+
+    disabled_actions = (
+        'zip_selected',  # disabled because ZIP filename is invalid
```
lukasgraf: What exactly does this comment mean here?

opengever/repository/browser/tabs.py
```diff
@@ -0,0 +1,28 @@
...
+    )
+
+    @property
+    def enabled_actions(self):
+        actions = super(RepositoryFolderDocuments, self).enabled_actions
+        return [action for action in actions if action not in self.disabled_actions]
+
+    @property
+    def columns(self):
+        columns = list(super(RepositoryFolderDocuments, self).columns)
```
lukasgraf: Hmm. Luckily `Document.columns` is immutable (tuple), not a list. Otherwise the `insert()` below would mutate the `columns` on the `Document` class. Maybe create a copy? It only takes one Schlaumeier to change `Documents.columns` into a list for whatever reason, and this will break in horrible ways 😉

opengever/repository/tests/test_repositoryfolder_tabs.py
```diff
@@ -9,12 +9,66 @@ class TestRepositoryFolderTabs(IntegrationTestCase):
...
+        self.assertNotIn('Documents', tabbedview.tabs().text)
+
+
+class TestRepositoryFolderDocumentsTab(IntegrationTestCase):
+
+    @browsing
+    def test_visible_actions(self, browser):
+        # Not all actions of the dossier-tab work on the repositoryfolder-tab.
+        # We are testing as Manager so that we can make sure that actions are
+        # filtered without programatically, without respecting permissions.
```
lukasgraf: -`without programatically,` (I think?)

opengever/repository/browser/tabs.py
jone: I think I'm already making a copy by calling the `list()`-type 😜

opengever/repository/browser/tabs.py
jone: I tried to download a bunch of documents on the documents-tab of a repository folder. I got a ZIP file, but it's name was just `zip` and MacOS couldn't extract it. 🤔

jone: I have updated the comment.

opengever/repository/tests/test_repositoryfolder_tabs.py
jone: I have rewritten this part.

jone:
- Updated according to comments.
- I've also removed the tasks-tab, since it was actually wrong and not really used; it works well with the default tab.
- I have added a test for the tasks tab in order to reveal future problems with changes on the regular task tab, which have a negative effect to the repository folder tab.

opengever/repository/browser/tabs.py
lukasgraf: You're right - for some reason, I was under the assumption, that `(list(already_a_list) is already_a_list) == True` (that Python would directly return the reference). Which is obviously false. 😜

Issue #3681, opened 2017-11-24
OGIP 17: Configure workspace participants


jone:
Build a new view for showing and configuring the workspace participants.

Requires the workflows #3679

### Features
- Show current participants and their roles.
- Invite new participants.
- Change the roles of existing participants.
- Show open invitations.
- Journalization: all changes must strictly be journalized, so that we can tell for any given time in the past who had access to which documents.

### Details and principles
- Available in the action menu as "Participants" (Teilnehmer).
- Separated tables for "participants" and "invitations".
- The role descriptions from lawgiver should be visible somehow.
- Only allow one role per participant.
- The role is selected on invitation time; the local role should be added when the pariticipant accepts the inviation, not when the invitation is created.
- Only users with roles `WorkspacesUser` or `WorkspacesCreator` can be invited.
- Journalization: use the existing GEVER journal, when sufficient.

### Access to view
- No access for `WorkspaceGuest`
- Read-only access for `WorkspaceMember`
- Read-and-write for `WorkspaceAdmin` and  `WorkspaceOwner`
- Admin-Access TBD (e.g. `Site Administrator`)

PR #3871, opened 2018-01-16
OGIP 17: Implement workspace invitation storage

maethu:
Part of #3681

The origin implementation can be found here:
https://github.com/4teamwork/ftw.participation/blob/78a4e0b5a384a033ce780aec473f6ae726508a89/ftw/participation/storage.py

https://github.com/4teamwork/ftw.participation/blob/78a4e0b5a384a033ce780aec473f6ae726508a89/ftw/participation/invitation.py

I did the following changes and improvements:
- Removed the unnecessary SESSION based code. (We always have a user session in GEVER)
- Do not write on read: `def invitations` did a write on read before. for this purpose I introduced a new method `initialize_storage`. This will be called by `add_invitation` if necessary
- Remove the `else` block in `remove_invitation`. Seems to be not necessary, since the userid never changes.
- Tests are completely new, there were "only" browser tests in ftw.participations, the storage itself was barely tested. 

This PR implements only the storage. The actual feature (forms, etc.) will follow


opengever/workspace/participation/invitation.py
```diff
@@ -0,0 +1,63 @@
...
+        storage = IInvitationStorage(target)
+        # set the attributes
+        self.userid = userid
+        self.inviter = inviter
+        self.set_target(target)
+        self.role = role
+        # generate the id, which sets it to self.iid
+        storage.generate_iid_for(self)
+        # register the invitation
+        storage.add_invitation(self)
```
jone: I think it is not a good design that the init of the item adds itself into the storage.
I'd rather do that outside 😉

opengever/workspace/participation/invitation.py
```diff
@@ -0,0 +1,63 @@
...
+    def set_target(obj):
+        """Sets the target to the given `obj`. Internally we store UIDs.
+        """
+
+    def get_target(self):
+        """Returns the currently stored target or `None`
+        """
+
+
+class Invitation(Persistent):
```
jone: I don't like that we subclass `Persistent`: it can cause a lot of problems when we want to refactor it in the future.
I'd rather have a `PersistentMapping` stored and let the `Invitation` be constructed temporarily with a pointer to the mapping as storage.

opengever/workspace/participation/invitation.py
```diff
@@ -0,0 +1,63 @@
...
+    """The `Invitation` object is used for storing who has invited who for
+    which context.
+    """
+
+    implements(IInvitation)
```
jone: In GEVER we want to use `@implementer` instead of class-directives everywhere 😉

opengever/workspace/participation/invitation.py
```diff
@@ -0,0 +1,63 @@
...
+    def set_target(self, obj):
+        self._target = IUUID(obj)
+
+    def get_target(self):
+        if '_target' not in dir(self):
+            return None
+        site = getSite()
+        catalog = getToolByName(site, 'portal_catalog')
+
+        result = catalog.unrestrictedSearchResults(UID=self._target)
```
jone: Use `from plone.app.uuid.utils import uuidToObject` 😜

opengever/workspace/participation/storage.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+
+
+STORAGE_CACHE_KEY = '_invitations_list'
+
+
+class InvitationStorage(object):
+    """The invitation stores and manages the invitation. All modification in
+    the list of invitations (appending new invitations, deleting existing
+    invitations, etc.) should be done with through the invitation storage.
```
jone: ✂️ ~~with~~

opengever/workspace/participation/storage.py
```diff
@@ -0,0 +1,180 @@
...
+    The invitation storage is adapter adapting any context. The storage stores
```
jone: ✂️ ~~adapter~~

opengever/workspace/participation/storage.py
```diff
@@ -0,0 +1,180 @@
...
+    its data into annotations on the plone site root.
+    """
+
+    implements(IInvitationStorage)
+    adapts(Interface)
```
jone: `@adapter`; `@implementer`

opengever/workspace/participation/storage.py
```diff
@@ -0,0 +1,180 @@
...
+
+    ANNOTATIONS_DATA_KEY = 'opengever.workspace : invitations'
+
+    def __init__(self, context):
+        """Adapts any context within a plone site
+        """
+        self.context = context
+        portal_url = getToolByName(self.context, 'portal_url')
+        self.portal = portal_url.getPortalObject()
```
jone: Use `getSite`

opengever/workspace/participation/storage.py
```diff
@@ -0,0 +1,180 @@
...
+        """
+        return IAnnotations(self.portal)
+
+    @property
+    def invitations(self):
+        """A persistent dictionary of all invitations. The key is the userid
+        of the invited user, the values are `PersistentList`s of
+        `Invitations` of this user.
+        """
+        if STORAGE_CACHE_KEY not in dir(self):
```
jone: ✂️ I think that we do not need a cache; this seems to be premature optimization. I'd therefore remove the cache completely.

opengever/workspace/participation/storage.py
```diff
@@ -0,0 +1,180 @@
...
+                # didnt cache it yet, get it from the annotations (and cache it)
+                setattr(self, STORAGE_CACHE_KEY,
+                        self.portal_annotations[self.ANNOTATIONS_DATA_KEY])
+
+        return getattr(self, STORAGE_CACHE_KEY, {})
+
+    def initialize_storage(self):
+        if self.ANNOTATIONS_DATA_KEY not in self.portal_annotations:
+            # dictionary does not exist in the annotations, create it
+            self.portal_annotations[self.ANNOTATIONS_DATA_KEY] = PersistentDict()
```
jone: We should use an `OOBTree` instead of a persistent dict.

opengever/workspace/participation/storage.py
```diff
@@ -0,0 +1,180 @@
...
+        # create a string as unique as possible, which will be hashed later
+        # for that the info of the string are not decodable by the user
+        base = '-'.join((
+            str(datetime.now().isoformat()),
+            invitation.userid,
+            invitation.inviter))
+        counter = 1
+        iid = None
+        while not iid or iid in used_iids:
+            iid = md5(base + str(counter)).hexdigest().strip()
```
jone: I think we should use some kind of UUID here.
- `uuid4` should work, but is not predictable
- if we need predictable uuids, we could try using `IUUIDGenerator` in combination with static uids.

opengever/workspace/participation/storage.py
```diff
@@ -0,0 +1,180 @@
...
+                        self.portal_annotations[self.ANNOTATIONS_DATA_KEY])
+
+        return getattr(self, STORAGE_CACHE_KEY, {})
+
+    def initialize_storage(self):
+        if self.ANNOTATIONS_DATA_KEY not in self.portal_annotations:
+            # dictionary does not exist in the annotations, create it
+            self.portal_annotations[self.ANNOTATIONS_DATA_KEY] = PersistentDict()
+
+    def get_used_iids(self):
```
jone: When we use something like `uuid4`, we probably do not have to do that anymore.

opengever/workspace/participation/storage.py
```diff
@@ -0,0 +1,180 @@
...
+        # set the iid on the invitation
+        invitation.iid = iid
+        return iid
+
+    def get_invitation_by_iid(self, iid, default=None):
+        """Returns the invitation with the `iid` or `default` if there is none.
+        """
+        for users_invitiations in self.invitations.values():
+            for invitation in users_invitiations:
+                for invitation in users_invitiations:
```
jone: ✂️ duplicate line 🙈

jone:
I've refactored it a bit, so that we are no longer subclassing `Persistent`, because it's a pain to remove those objects later. We should try to not implement persistent objects.

I've also changed the storage to be a utility.

maethu:
@jone HAHA, The frontend in #3882 is still working, though ;-)

PR #3872, opened 2018-01-16
Use excerpt header template as master, if available.

deiferni:
If no excerpt header template is available, fall back to agenda item`s document as master.

This fixes an issue with headings not taken from the correct file in the merged word documents.

Needs backport to `2017.7`.

### DIFF
--- a/opengever/meeting/command.py
+++ b/opengever/meeting/command.py
@@ -45,8 +45,10 @@ MIME_DOCX = 'application/vnd.openxmlformats-officedocument.wordprocessingml.docu
 
 class MergeDocxExcerptCommand(CreateDocumentCommand):
     """Create or update a merged excerpt word file.
-    """
 
+    If an excerpt header is available it is considered the master document.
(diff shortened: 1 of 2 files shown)


opengever/meeting/command.py
```diff
@@ -62,15 +64,21 @@ def __init__(self, context, agenda_item, filename, title):
     def generate_file_data(self):
         header_template = self.agenda_item.get_excerpt_header_template()
         suffix_template = self.agenda_item.get_excerpt_suffix_template()
+        agenda_item_document = self.agenda_item.resolve_document()
 
         if header_template is None and suffix_template is None:
-            return self.agenda_item.resolve_document().file.data
+            return agenda_item_document.file.data
+
+        if header_template is not None:
```
jone: We've discussed that we might use a variable here, since this condition is repeated below. This could prevent us from failures when we change it in the future.

deiferni:
@jone updated as suggested

PR #3874, opened 2018-01-17
Ergänzung Bemerkung Titel-Länge bei Zip-Download

livschal:


deiferni:
Also please note that a free zip tool can handle longer path-names on older versions of windows.  I think we should add a link to the tool in our docs, too, and recommend to install it.

The zip program is available at http://www.7-zip.org/.

PR #3875, opened 2018-01-17
 Ensure documents can be attached to email from within forwardings within the inbox

### LINKED ISSUE #3849: Attach-to-outlook fails for documents in forwardings
Attach-to-outlook fails with `AttributeError: is_open` for documents in (some?) forwardings.

Possibly related to #2556

**Steps to reproduce**

As zopemaster:
- Enable the `opengever.officeconnector.interfaces.IOfficeConnectorSettings.attach_to_outlook_enabled` registry setting to allow attach to outlook

As user in the inbox group (`lukas.graf` works for local dev):
- Drag & Drop upload a file to the Inbox to create a document
- Check that document in the Inbox's `Documents` tab, and press the `Weiterleiten` button
- As responsible (Auftragnehmer), select your current user
- Submit, and open the Forwarding that just got created

- Try to attach-to-outlook the document in the forwarding using the button in the overlay. Either from the Fowarding overview page, section "Dokumente", or the separate "Dokumente" tab.

**Expected result**
- Opened as attachment in mail client

**Actual Result**
`AttributeError: is_open`

oc_attach_to_outlook_500

```
2018-01-08 15:13:37 ERROR Zope.SiteErrorLog 1515420817.050.0317682606825 http://localhost:8080/fd/oc_attach
Traceback (innermost last):
  Module plone.rest.service, line 23, in __call__
  Module opengever.officeconnector.service, line 147, in render
AttributeError: is_open
```

Rotonen:
* Split up the OC tests a bit better per theme
  * This process will continue as the matrix grows, but for now it's clean enough
* Only do BCC for business case dossiers

I decided against doing the full test matrix of OC vs. Forwardings and Tasks vs. documents at this point. I'll keep amending the matrix as something pops up - it took a year for this to pop up so we're not in a horrible state.

When we tackle reworking Task, we could build the whole matrix to act as the acceptance testing for the reworking efforts, though.

Closes #3849


opengever/testing/fixtures.py
```diff
@@ -585,6 +585,17 @@ def create_inbox(self):
             ['Contributor', 'Editor', 'Reader'],
             )
 
+        self.register('inbox_forwarding', create(
+            Builder('forwarding')
+            .within(self.inbox)
+            .titled(u'F\xf6rw\xe4rding')
+            .having(
```
deiferni: fowardings have a relationship to one or multiple documents. this is enforced when creating them via web-ui. the document then ends up being inside the forwarding. i wonder if we should also add one in the fixture as this seems to be required, or at least expected.

opengever/testing/fixtures.py
```diff
@@ -596,6 +596,13 @@ def create_inbox(self):
                 )
             ))
 
+        self.register('inbox_forwarding_document', create(
```
deiferni: never mind my previous comment 😉 .

opengever/officeconnector/service.py
```diff
@@ -148,7 +148,11 @@ def render(self):
             document = payload['document']
             parent_dossier = document.get_parent_dossier()
 
-            if parent_dossier:
+            if (
+                    parent_dossier
+                    and parent_dossier.portal_type
+                    == 'opengever.dossier.businesscasedossier'
```
deiferni: this check is incorrect. we have other dossier-types that provide the `is_open` method, please change this check to `IDossierMarker.providedBy(parent_dossier)`.

Also, what is going on here 🤔 , `get_parent_dossier` seems to return a non-dossier. That may be worth investigating a bit? Is an Inbox returned? Should we define `is_open` on Inbox and always return `True`?

opengever/officeconnector/tests/test_api_forwarding_checkout.py
```diff
@@ -0,0 +1,32 @@
+from ftw.testbrowser import browsing
+from opengever.officeconnector.testing import OCIntegrationTestCase
+
+
+class TestOfficeconnectorForwardingAPIWithCheckout(OCIntegrationTestCase):
+    features = (
+        'officeconnector-checkout',
+    )
+    @browsing
```
deiferni: \n ?

opengever/officeconnector/tests/test_api_forwarding_attach.py
```diff
@@ -0,0 +1,58 @@
...
+    @browsing
+    def test_attach_to_email_open_with_file(self, browser):
+        self.login(self.secretariat_user, browser)
+
+        oc_url = self.fetch_document_attach_oc_url(
+            browser,
+            self.inbox_forwarding_document,
+            )
+
+        self.assertTrue(oc_url)
```
deiferni: i think `assertIsNotNone` would indicate better what you're interested in here

opengever/officeconnector/service.py
Rotonen: It returns the inbox. This is what caught me off guard to begin with. Maybe I should just dive in there and move the marker there so we only get dossiers from there?

deiferni: hmmmm it almost certainly is relevant for building the reference number 🤔 , so i'd rather not. `IDossierMarker.providedBy` seems to be the way to go then i'd say.

Rotonen:
@deiferni ready for rereview.

lukasgraf:
The issue that got fixed here probably was what caused https://sentry.4teamwork.ch/sentry/onegov-gever/issues/4382/

PR #3882, opened 2018-01-19
Manage Participants on a Workspace

maethu:
This PR introduces a new manage participants view on workspaces.

- No longer show the sharing action.
- Show a participants action.
- Only User with sharing permission can manage participants. 

There's a JSON (not RESTFul) backend. with a minimal VueJS frontend.
The necessary resources are only loaded on the view itself.

**Invite a new participant:**
screen shot 2018-01-19 at 15 23 57
screen shot 2018-01-19 at 15 24 14

**Delete invitation**
screen shot 2018-01-19 at 15 25 10
Disable all other delete buttons while request is in progress.

**Modify role of invitation or participant:**
screen shot 2018-01-19 at 15 26 25
screen shot 2018-01-19 at 15 26 32
Disable all other modify dropdown while request is in progress.

Select2 widget is no everywhere in use, big impact in vuejs - solvable but too much effort for now. 

Further it also features a new my-invitations view, including a personal action. to access "my invitations".


jone:
The merge-base was merged into master; therefore this can be rebased onto master and the merge base changed.

opengever/workspace/locales/de/LC_MESSAGES/opengever.workspace.po
```diff
@@ -58,3 +131,8 @@ msgstr "Papierkorb"
 #: ./opengever/workspace/browser/tabbed.py
 msgid "label_workspaces"
 msgstr "Teamräume"
+
+#. Default: "Manage pariticipation of ${workspace}"
+#: ./opengever/workspace/participation/browser/templates/participants-view.pt
+msgid "title_manage_pariticpation"
+msgstr "Teilnehmer vom Teamraum \"${workspace}\" verwalten"
```
jone: Das dünkt mich relativ "schwer" / umständlich. Ich persönlich würde dies vereinfache,
z.B.: `Teilnehmer von "Teamraum für wichtiges"`

opengever/workspace/participation/browser/manage_participants.py
```diff
@@ -0,0 +1,181 @@
...
+        response = self.request.response
+        response.setHeader('Content-Type', 'application/json')
+        response.setHeader('X-Theme-Disabled', 'True')
+        response.enableHTTPCompression(REQUEST=self.request)
+        return json.dumps(data)
+
+    def get_participants(self):
+        return self.get_user_role_mapping()
+
+    def get_user_role_mapping(self):
```
jone: From `mapping` I had expected the return type to be a dictionary, but it is a list.
How about `get_user_role_infos`? or even `get_user_infos`?

opengever/workspace/participation/browser/manage_participants.py
```diff
@@ -0,0 +1,181 @@
...
+                        inviter=self.get_full_user_info(
+                            userid=invitation['inviter']),
+                        can_manage=self.can_manage_member(),
+                        type_='invitation',
+                        token=invitation['iid'])
+            entries.append(item)
+
+        return entries
+
+    def get_full_user_info(self, userid=None, member=None):
```
jone: Doesn't the OGDS service provide a method for that? We'd use the standard format then 🤔

opengever/workspace/participation/browser/manage_participants.py
```diff
@@ -0,0 +1,181 @@
...
+        CheckAuthenticator(self.request)
+        userid = self.request.get('userid', None)
+        role = self.request.get('role', None)
+
+        if not userid or not role or not self.can_manage_member():
+            raise BadRequest('No userid or role provided')
+
+        storage = getUtility(IInvitationStorage)
+        storage.add_invitation(self.context, userid,
+                               api.user.get_current().getId(), role)
```
jone: By constructing a request, I think I could grant myself an administrator role at this context.
We should at least verify that the role is one of `MANAGED_ROLES`.

I would also get the ISharingPageRole-Named-Utility and verify that the current user has the role-permission on the context. I'd do that because we then still follow Plone principles and respect its security.
The permission should be set correctly already because we use ftw.lawgiver.

opengever/workspace/participation/browser/manage_participants.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def delete(self):
+        """ A traversable method to delete a pending invitation"""
+        CheckAuthenticator(self.request)
+
+        token = self.request.get('token', None)
+        type_ = self.request.get('type', None)
+
+        if not token or not type_:
+            raise BadRequest('A token a a type is required')
```
jone: +nd

opengever/workspace/participation/browser/manage_participants.py
```diff
@@ -0,0 +1,181 @@
...
+        if not userid or not role or not self.can_manage_member():
+            raise BadRequest('No userid or role provided')
+
+        storage = getUtility(IInvitationStorage)
+        storage.add_invitation(self.context, userid,
+                               api.user.get_current().getId(), role)
+        return self.__call__()
+
+    def delete(self):
+        """ A traversable method to delete a pending invitation"""
```
jone: .. or local role 😉

opengever/workspace/participation/browser/manage_participants.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def modify(self):
+        """ A traversable method to modify a users local roles"""
+        CheckAuthenticator(self.request)
+
+        token = self.request.get('token', None)
+        role = self.request.get('role', None)
+        type_ = self.request.get('type', None)
+
+        if not token or not role or not type_ and role in MANAGED_ROLES:
```
jone: probably role _not_ in managed roles?
I'd move the "and .." into a seperate if-clause for clarity, and raise "Unauthorized" instead.

opengever/workspace/participation/browser/manage_participants.py
```diff
@@ -0,0 +1,181 @@
...
+            raise BadRequest('No userid or invalid role provided')
+
+        if type_ == 'user':
+            user_roles = api.user.get_roles(username=token, obj=self.context,
+                                            inherit=False)
+            if user_roles and 'WorkspaceOwner' not in user_roles:
+                self.context.manage_setLocalRoles(token, [role])
```
jone: Could we run into trouble because we are removing existing non-workspace roles here?
e.g. we are not taking care of the `Owner`-role, so should we keep it if the user has it?

opengever/workspace/participation/browser/my_invitations.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def __call__(self):
+
+        return self.template()
+
+    def is_feature_enabled(self):
+        registry = getUtility(IRegistry)
+        settings = registry.forInterface(IWorkspaceSettings, check=False)
+        if settings:
+            return settings.is_feature_enabled
```
jone: The convention would be to use [`is_workspace_feature_enabled`](https://github.com/4teamwork/opengever.core/blob/master/opengever/workspace/__init__.py#L7) here.

opengever/workspace/participation/browser/my_invitations.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def get_invitations(self):
+        userid = api.user.get_current().getId()
+
+        entries = list(self.storage().iter_invitions_for_recipient(userid))
+        entries.sort(key=lambda item: item['created'])
+
+        for entry in entries:
+            with elevated_privileges():
+                target = uuidToObject(entry['target_uuid'])
```
jone: I would probably even do:
```python
            with elevated_privileges():
                target_title = uuidToObject(entry['target_uuid']).Title()
```
It is small detail, but it causes the object retrieved with elevated security to not leave the context manager. We are using the context manager because we argument that we only get the title - so why not just do that?

opengever/workspace/participation/browser/participants_view.py
```diff
@@ -0,0 +1,55 @@
...
+                           'table.html'), 'r') as html_file:
+        table_template = html_file.read()
+
+    def __call__(self):
+        return self.template()
+
+    # @property
+    # def table_template(self):
+    #    """Enable for development."""
+    #     with open(os.path.join(os.path.dirname(__file__),
```
jone: ✂️ 😏

opengever/workspace/participation/browser/my_invitations.py
maethu: It may no exist? Then title gives a AttributEerror

jone: Indeed, an `if target` would be nice.
The point is that I'd try to not use the "target" object outside of the context manager.

opengever/workspace/participation/browser/resources/app.js
```diff
@@ -0,0 +1,107 @@
...
+      payload.append('role', event.currentTarget.value);
+      payload.append('type', entry.type_);
+      payload.append('_authenticator', this.authtoken);
+      axios.post(this.endpoint + '/modify', payload).then(function (response) {
+        if (response.status === 204) {
+          entry.roles = [newRole];
+          self.isModifying = false;
+          self.messanger.shout([{'messageTitle': 'Info', 'message': 'Rolle wurde aktualisiert', 'messageClass': 'info'}]);
+        }
+      });
```
jone: What happens when something goes wrong?

opengever/workspace/participation/browser/resources/app.js
```diff
@@ -0,0 +1,107 @@
...
+
+    roles: function() {
+      return [
+        {name: this.i18n.workspaceguest, value: 'WorkspaceGuest'},
+        {name: this.i18n.workspacemember, value: 'WorkspaceMember'},
+        {name: this.i18n.workspaceadmin, value: 'WorkspaceAdmin'},
+      ];
+    },
+
+    messanger: MessageFactory.getInstance,
```
jone: maybe `messenger`? 😏

opengever/workspace/participation/browser/resources/app.js
```diff
@@ -0,0 +1,107 @@
...
+      self.isModifying = true;
+      payload.append('token', entry.token);
+      payload.append('role', event.currentTarget.value);
+      payload.append('type', entry.type_);
+      payload.append('_authenticator', this.authtoken);
+      axios.post(this.endpoint + '/modify', payload).then(function (response) {
+        if (response.status === 204) {
+          entry.roles = [newRole];
+          self.isModifying = false;
+          self.messanger.shout([{'messageTitle': 'Info', 'message': 'Rolle wurde aktualisiert', 'messageClass': 'info'}]);
```
jone: I'd also translate `Rolle wurde aktualisiert`, since you already have solved i18n.

opengever/workspace/participation/browser/resources/app.js
```diff
@@ -0,0 +1,107 @@
...
+        }
+      });
+    },
+
+    invideUser: function () {
```
jone: invite

opengever/workspace/participation/browser/resources/app.js
```diff
@@ -0,0 +1,107 @@
...
+      payload.append('userid', this.$refs.userid.value);
+      payload.append('role', this.$refs.role.value);
+      payload.append('_authenticator', this.authtoken);
+
+      var self = this;
+      axios.post(this.endpoint + '/add', payload).then(function (response) {
+        self.entries = response.data;
+        jQuery('select[name="userid"]').val('').trigger('change');
+        jQuery('select[name="role"]').val('WorkspaceGuest').trigger('change');
+        self.messanger.shout([{'messageTitle': 'Info', 'message': 'Teilnehmer wurde eingeladen', 'messageClass': 'info'}]);
```
jone: i18n

opengever/workspace/participation/browser/resources/app.js
```diff
@@ -0,0 +1,107 @@
...
+      var payload = new FormData();
+      payload.append('token', entry.token);
+      payload.append('type', entry.type_);
+      payload.append('_authenticator', this.authtoken);
+
+      var self = this;
+      axios.post(this.endpoint + '/delete', payload).then(function (response) {
+        self.entries = response.data;
+        self.isDeleting = false;
+        self.messanger.shout([{'messageTitle': 'Info', 'message': 'Teilnehmer wurde gelöscht', 'messageClass': 'info'}]);
```
jone: i18n

opengever/workspace/participation/browser/templates/my-invitations.pt
```diff
@@ -0,0 +1,47 @@
...
+                    <th i18n:translate="">Workspace</th>
+                    <th i18n:translate="">Action</th>
+                </tr>
+            </thead>
+            <tbody>
+                <tr tal:repeat="invitation view/get_invitations">
+                    <td tal:content="invitation/inviter" />
+                    <td tal:content="invitation/target_title" />
+                    <td tal:define="authtoken string:${context/@@authenticator/token}">
+                        <a class="AcceptInvitation" tal:attributes="href string:${context/absolute_url}/my-invitations/accept?iid=${invitation/iid}&_authenticator=authtoken" i18n:translate="">
```
jone: I think it would be correct to escape the `&` with `&amp;` (github marks it red 😉 )

opengever/workspace/participation/browser/templates/participants-view.pt
```diff
@@ -0,0 +1,39 @@
...
+    <metal:js fill-slot="javascript_head_slot">
+      <script tal:attributes="src string:${here/portal_url}/++resource++opengever.workspace.participants.resources/vue.min.js"></script>
+
+      <script tal:attributes="src string:${here/portal_url}/++resource++opengever.workspace.participants.resources/axios.min.js"></script>
+
+    </metal:js>
+
+
+    <metal:title fill-slot="content-title">
+       <h1 class="documentFirstHeading" i18n:translate="title_manage_pariticpation">
```
jone: pariticpation 😉

opengever/workspace/participation/browser/templates/table.html
```diff
@@ -0,0 +1,40 @@
...
+
+    <div class="actions">
+      <select ref="userid" name="userid"
+              v-bind:data-ajaxoptions="ajaxOptions"
+              v-bind:data-select2config=select2Config>
+      </select>
+      <select ref="role" name="role">
+        <option v-for="role in roles" :value=role.value>{{role.name}}</option>
+      </select>
+      <input type="submit" :value="i18n.invite" @click.prevent="invideUser" />
```
jone: I suggest to move the "invite" actions above the table of current participants, because
- it would be faster to invite a new participant to e workspace which already has lots of participants (no need to scroll down)
- it has less / consistent height; I'd go from small to big from a UI point of view
- I'm personally more often inviting new participants than changing / removing existing ones on similary views

opengever/core/upgrades/20180122154722_add_my_invitation_action/actions.xml
```diff
@@ -0,0 +1,18 @@
+<object xmlns:i18n="http://xml.zope.org/namespaces/i18n" name="portal_actions" meta_type="Plone Actions Tool">
+
+  <object name="user" meta_type="CMF Action Category">
+    <object name="my-invitations" meta_type="CMF Action" i18n:domain="opengever.core">
+      <property name="title" i18n:translate="">My Invitations</property>
+      <property name="description" i18n:translate="" />
+      <property name="url_expr">string:${globals_view/navigationRootUrl}/@@my-invitations</property>
+      <property name="icon_expr" />
+      <property name="available_expr">here/portal_url/@@my-invitations/is_feature_enabled</property>
```
jone: The action should appear before the logout action. I always expect the logout action to be the last and I'm confused when this is not the case 😉 

bildschirmfoto 2018-01-26 um 13 38 11

opengever/core/locales/de/LC_MESSAGES/opengever.core.po
```diff
@@ -135,6 +135,11 @@ msgstr "Sitzungsdossier"
 msgid "Move Items"
 msgstr "Elemente verschieben"
 
+#: ./opengever/core/profiles/default/actions.xml
+#: ./opengever/core/upgrades/20180122154722_add_my_invitation_action/actions.xml
+msgid "My Invitations"
+msgstr "Meine Einladungen"
```
jone: `Meine Einladungen` and
`Meine Einstellungen`
look very similary, which did confuse me when I first opened the menu.
Can we find another label which is better visually diffable to the existing one?
bildschirmfoto 2018-01-26 um 13 41 28

opengever/workspace/participation/browser/templates/my-invitations.pt
```diff
@@ -0,0 +1,47 @@
...
+            <thead>
+                <tr>
+                    <th i18n:translate="">Invited by</th>
+                    <th i18n:translate="">Workspace</th>
+                    <th i18n:translate="">Action</th>
+                </tr>
+            </thead>
+            <tbody>
+                <tr tal:repeat="invitation view/get_invitations">
+                    <td tal:content="invitation/inviter" />
```
jone: This displays the userid instead of the full user-info (fullname, ..).
bildschirmfoto 2018-01-26 um 13 46 15

opengever/workspace/participation/browser/manage_participants.py
jone: When changing the role of another participant from "guest" to "member" we have a caching problem: it does not invalidate the p.a.caching cachekey and the editbar may not appear for the other user.

I think the modified-date of the workspace must change when we change the participants or the roles, so that we can invalidate certain caches.

I have also notified that we are not reindexing the object security at this point. With the current catalog this may not be needed, since we know that all selectable roles have the same "view"-permission in the workflow. I'd probably do it anways, since the workflow can change it the future and we won't remember this.

maethu: @jone  I guess not, the Owner does nothing at all. We have the WorkspaceOwner role for this purpose.

opengever/workspace/participation/browser/my_invitations.py
maethu: I'll do now. Should I change the endpoint name too? 
I did it manually, since this was more "upgrade save". But the upgrade will be run anyway before someone is hitting the view.

opengever/workspace/participation/browser/participants_view.py
maethu: @jone :-( but it is useful for development. If I remove another will fail hard at this point. What so you think?

opengever/core/locales/de/LC_MESSAGES/opengever.core.po
maethu: @jone I'll go for `Teamraum Einladungen`. `Teamraum` should be distinct enough.

opengever/workspace/participation/browser/manage_participants.py
maethu: `I have also notified that we are not reindexing the object security at this point. With the current catalog this may not be needed, since we know that all selectable roles have the same "view"-permission in the workflow. I'd probably do it anways, since the workflow can change it the future and we won't remember this.`

opengever/workspace/participation/browser/manage_participants.py
maethu: I changed that.

opengever/workspace/participation/browser/my_invitations.py
maethu: 👍 done

opengever/core/locales/de/LC_MESSAGES/opengever.core.po
maethu: 👍 done

opengever/core/upgrades/20180122154722_add_my_invitation_action/actions.xml
maethu: 👍

opengever/workspace/participation/browser/templates/my-invitations.pt
maethu: 👍 Done

opengever/workspace/participation/browser/templates/my-invitations.pt
maethu: oh damn good catch. 👍

maethu:
@jone thanks for feedback. I updated almost everything you mentioned, incl. the changelog 👍 

Look at that, it even works in IE11 😉 
screen shot 2018-01-30 at 08 21 03

maethu:
In my opinion it's now ready for the DEV system.

opengever/workspace/participation/browser/manage_participants.py
jone: Yes, but there might be things relying on the Owner-role (e.g. owner index, `.Owners`-method, etc).

opengever/workspace/participation/browser/participants_view.py
jone: I would not know that this is here, I'll just be annoyed that it does not apply the changes and reboot 😜 

Could we change it so that it just works as expected?
Could we just cache the template as long as the mtime of the file does not change?
Or maybe check the debug mode?

maethu: Ok: I just remove it for now :-)

PR #3883, opened 2018-01-19
 Prefix the reference numbers on private dossiers with a P

### LINKED ISSUE #3571: Aktenzeichen für Privaten Inhalt mit `P` versehen
Siehe OGIP 19 - Kapitel 4.3

- reference number for repositoryroots and PrivateRepositoryRoots
- behavior with a reference part field, only visible managers

https://github.com/4teamwork/opengever.core/blob/master/opengever/base/reference_formatter.py#L29-L44

Rotonen:
* Amended the private dossier reference number getter
  * Get the current formatter
  * If possible, use the current separator
  * Default to a space

Did not go for making the prefix configurable per repository root / repository folder. Splitting that onto its own issue to be tackled later down the line.

Closes #3571

### DIFF
--- a/opengever/private/dossier.py
+++ b/opengever/private/dossier.py
@@ -1,6 +1,10 @@
+from opengever.base.interfaces import IReferenceNumberFormatter
+from opengever.base.interfaces import IReferenceNumberSettings
 from opengever.dossier.base import DossierContainer
 from opengever.dossier.businesscase import IBusinessCaseDossier
 from opengever.private.interfaces import IPrivateContainer
+from plone import api
+from zope.component import queryAdapter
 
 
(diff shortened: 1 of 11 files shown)


opengever/private/dossier.py
```diff
@@ -12,3 +16,27 @@ class PrivateDossier(DossierContainer):
...
+            interface=IReferenceNumberSettings,
+            )
+
+        formatter = queryAdapter(
+            api.portal.get(),
+            IReferenceNumberFormatter,
+            name=active_formatter_name,
+            )
+
+        if 'repository_dossier_seperator' in formatter.__dict__:
```
lukasgraf: Can be shortened to `separator = getattr(formatter, 'repository_dossier_seperator', u' ')` (untested)

Rotonen:

PR #3893, opened 2018-01-24
 Fix JWT authentication for users from the Zope root acl_users for OC payloads

Rotonen:
Just mirroring `plone.restapi` here.

We had cloned the logic from there as we needed to customize it, so the fix did not propagate down to what we're actually doing just by upgrading `plone.restapi`.

### DIFF
--- a/opengever/officeconnector/tests/test_zopemaster.py
+++ b/opengever/officeconnector/tests/test_zopemaster.py
@@ -1,6 +1,7 @@
 from ftw.testbrowser import browsing
 from hashlib import sha256
 from opengever.officeconnector.testing import OCIntegrationTestCase
+from opengever.testing.assets import path_to_asset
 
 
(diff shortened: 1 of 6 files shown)


opengever/officeconnector/tests/test_zopemaster.py
```diff
@@ -0,0 +1,78 @@
...
+
+        self.checkout_document(browser, tokens, payload, self.document)
+
+        lock_token = self.lock_document(browser, tokens, self.document)
+
+        original_checksum = sha256(
+            self.download_document(browser, tokens, payload),
+            ).hexdigest()
+
+        with open('../../opengever/testing/assets/addendum.docx') as f:
```
deiferni: please use 

```python
from opengever.testing import assets

assets.load('…')
```

using a relative path is quite brittle.

Rotonen:
@deiferni ready for rereview.

PR #3901, opened 2018-01-25
Ensure bumblebee overlay view download links point to the correct version

### LINKED ISSUE #3863: Download copy on document version overlay downloads current file
Steps to Reproduce:

- Create Document with multiple versions
- Go to versions tab, open "Preview" for a non-recent version
- Click "Kopie herunterladen" in the Preview overlay

Actual outcome:
The most recent version of the file is downloaded.

Expected outcome:
The version for which the preview is rendered should be downloaded.

Rotonen:
Appends the version_id parametre to download links if it is present in the request and that is not the current version.

Closes #3863

### DIFF
--- a/opengever/document/browser/actionbuttons.py
+++ b/opengever/document/browser/actionbuttons.py
@@ -81,11 +81,26 @@ class ActionButtonRendererMixin(object):
         links to the orginal message download view.
         """
         dc_helper = DownloadConfirmationHelper(self.context)
-        return dc_helper.get_html_tag(
-            additional_classes=['function-download-copy'],
-            viewname=self.context.get_download_view_name(),
-            include_token=True,
-        )
+
+        kwargs = {
+            'additional_classes': ['function-download-copy'],
+            'viewname': self.context.get_download_view_name(),
+            'include_token': True,
+            }
+
+        requested_version_id = self.request.get('version_id')
+
+        if requested_version_id:
+            requested_version_id = int(requested_version_id)
+            current_version_id = self.context.get_current_version_id()
+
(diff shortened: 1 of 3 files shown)


opengever/document/browser/actionbuttons.py
```diff
@@ -81,11 +81,30 @@ def get_download_copy_tag(self):
...
+        kwargs = {
+            'additional_classes': ['function-download-copy'],
+            'viewname': self.context.get_download_view_name(),
+            'include_token': True,
+            }
+
+        requested_version_id = self.request.get('version_id')
+
+        if (
+                requested_version_id
```
lukasgraf: This will also evaluate to `False` for `requested_version_id == 0`. So it should be `requested_version_id is not None` instead.

opengever/document/browser/actionbuttons.py
```diff
@@ -81,11 +81,30 @@ def get_download_copy_tag(self):
...
+                and getattr(self.context, 'get_current_version_id', None)
```
deiferni: maybe:
```python
hasattr(self.context, 'get_current_version_id')
```
?

but why/where could this happen?

Rotonen: Mails.

deiferni: check `OGMail`.

Rotonen: Huh, hit something while developing, but tests do pass, no recollection at this point.

Rotonen:
@deiferni ready for rereview.

PR #3909, opened 2018-01-30
Add SKA admin unit for development.

jone:
In development, we have only one admin unit "FD" (=Plone site). In order to test things locally with two admin units, this adds a second admin unit "SKA" with one org unit, allowing for a easier setup.

When using it, make sure to not purge the SQL on setup.

### DIFF
--- a/opengever/examplecontent/configure.zcml
+++ b/opengever/examplecontent/configure.zcml
@@ -11,7 +11,7 @@
 
 
   <opengever:registerDeployment
-      title="Development with examplecontent"
+      title="Development with examplecontent (FD)"
       policy_profile="opengever.examplecontent:default"
       additional_profiles="opengever.setup:default_content
                            opengever.examplecontent:empty_templates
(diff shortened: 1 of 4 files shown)


opengever/examplecontent/profiles/default-ska/propertiestool.xml
```diff
@@ -0,0 +1,9 @@
+<object name="portal_properties" meta_type="Plone Properties Tool">
```
deiferni: to avoid this duplication i would suggest to:
- create two new profiles for fd and ska that only contain the org-units. use `opengever.examplecontent:default` as `policy_profile` for both and add one of the two unit-profiles to `additional_profiles`.

jone: I would rather just remove those files, e.g.:
- we want to use the default language - I can't see why we are "changing" for development only
- I'm don't understand why we add the Topic (twice!) to the default_page_types 😯 
- we could easily use the standard feature flags for the second development admin unit (I'd actually prefer that)

I'm going to do just that 😉

jone:
@deiferni I've removed the duplicated files 😏 😎

opengever/examplecontent/profiles/default-ska/propertiestool.xml
deiferni: fair enough, works for me 👍

PR #3910, opened 2018-01-30
Drop unused SQL sequences

### LINKED ISSUE #3477: Check for unused PostgreSQL sequences
It seems the Sequence `proposal_history_id_seq` is only ever added in an upgrade step, but not used/created any more on a new install.

This sequence, and maybe others, should probably be removed via an upgrade-step.

Rotonen:
Ran `SELECT c.relname FROM pg_class c WHERE c.relkind = 'S';` on an older client installation and a fresh installation and diffed:

```
--- fresh.txt	2018-01-30 18:41:38.000000000 +0100
+++ <redacted>.txt	2018-01-30 18:41:29.000000000 +0100
@@ -23,9 +23,11 @@
 participations_id_seq
 periods_id_seq
 phonenumber_id_seq
+proposal_history_id_seq
 proposal_id_seq
 resources_id_seq
 submitteddocument_id_seq
+subscriptions_resource_id_seq
 task_id_seq
 teams_id_seq
 urls_id_seq
```

Closes #3477.

### DIFF
--- a/opengever/core/upgrades/20180131153203_drop_unused_sequences/upgrade.py
+++ b/opengever/core/upgrades/20180131153203_drop_unused_sequences/upgrade.py
@@ -1,0 +1,28 @@
+from opengever.core.upgrade import SchemaMigration
+from sqlalchemy.schema import DropSequence
+from sqlalchemy.schema import Sequence
+
+
+class DropUnusedSequences(SchemaMigration):
+    """Drop unused sequences."""
+
+    def migrate(self):
+        # We'll only attempt to do this for postgresql
+        if self.is_postgres:
+            sequences_to_drop = (
+                'proposal_history_id_seq',
+                'subscriptions_resource_id_seq',
+                )
+
+            matching_sequences = tuple(
+                Sequence(sequence['relname'])
+                for sequence in self.execute(
+                    "select relname from pg_class "
+                    "where relkind = 'S' "
+                    "and relname in {}"
+                    .format(sequences_to_drop)
+                    ).fetchall()
+                )
+
+            for sequence in matching_sequences:
+                self.execute(DropSequence(sequence))


opengever/core/upgrades/20180130184240_drop_unused_sequences/upgrade.py
```diff
@@ -0,0 +1,15 @@
...
+    """Drop unused sequences."""
+
+    def migrate(self):
+        sequences_to_drop = (
+            'proposal_history_id_seq',
+            'subscriptions_resource_id_seq',
+            )
+
+        for sequence in sequences_to_drop:
+            DropSequence(sequence)
```
deiferni: have you tested this? i think this can't work since you just instantiate an SQL statement without actually executing it.

deiferni: also please only delete the sequence if it exists.

opengever/core/upgrades/20180130184240_drop_unused_sequences/upgrade.py
```diff
@@ -0,0 +1,15 @@
...
+class DropUnusedSequences(SchemaMigration):
+    """Drop unused sequences."""
+
+    def migrate(self):
+        sequences_to_drop = (
+            'proposal_history_id_seq',
+            'subscriptions_resource_id_seq',
+            )
+
+        for sequence in sequences_to_drop:
```
deiferni: also should we guard against systems without sequences? maybe there is a way to ask the dialect if it has sequences and `return` otherwise?

Rotonen:
@deiferni ready for rereview.

opengever/core/upgrades/20180131153203_drop_unused_sequences/upgrade.py
```diff
@@ -0,0 +1,28 @@
...
+from sqlalchemy.schema import DropSequence
+from sqlalchemy.schema import Sequence
+
+
+class DropUnusedSequences(SchemaMigration):
+    """Drop unused sequences."""
+
+    def migrate(self):
+        # We'll only attempt to do this for postgresql
+        if self.connection.dialect.name == 'postgresql':
```
deiferni: we have `self.is_postgres()`

Rotonen:
@deiferni ready for rereview.

PR #3915, opened 2018-01-31
Introduce system-wide tracking of schema migrations.

### LINKED ISSUE #3354: Schema-migration tracking table can't deal with orphaned upgrade

jone:
Fixes #3354

Schema migrations must only be applied once per system (=SQL-db).

In an environment with multiple admin units (=Plone sites) it is therefore necessary that only the first admin unit running upgrades runs schema migrations and the other admin units are not running the same schem migrations again.

This feature was implemented by storing the target version of the last executed schema migration per profile in the SQL database (opengever_upgrade_version table). The problem with this approach is that it does not support orphan schema migration because it does not track the installation for each schema migration but uses the latest version for comparison. (An orphan uprade step is one that is not yet installed but has a lower target version than the currently installed version).

The approach for fixing that is to track all schema migration separately, requiring a change in the primary key of the table. The tracking is moved to a GEVER-specific IUpgradeStepRecorder customization which also stores tracking information the SQL database for schema migrations, allowing us to actually mark migrations as executed on other admin units, rather than marking them as proposed and skipping the execution.

In order for making these changes easy, the knowledge of the operations class had to be changed so that it does not depend on upgrade steps.

### Updating multi-admin-unit systems
The hardest situation to update is an old system (with pending schema migrations from last year), having multiple admin units (Plone sites).

Updating the tracking table:
We need to change the primary key of the tracking table in order to allow for multiple entries per profile.
This cannot happen in a schema migration, because the code will start to insert rows into the table whenever the first schema migration is applied.
Therefore the tracking table is cannot be migrated in a regular schema migration; it is therefore migrated on the fly when accessing.

Tracking of already executed schema migrations:
Up to now, all schema migrations (which are installed before this change) are tracked also by ftw.upgrade in the Plone site annotations.
Since we used to also execute upgrade steps although they were executed by other admin units, each admin-unit has marked all schema migrations locally too.
Therefore there is no need to update the tracking for older schema migrations.

Schema migrations installed in the same release but before this change will just start beeing tracked in SQL and stop beeing proposed on other admin units.
Therefore we are safe there too.

The only import thing is that when we start installing any update including this change, all proposed upgrades on all admin units must have been executed before and all systems must start with the same state.

### Other changes

- The IdempotentOperations class is refactored so that it no longer has a pointer to the upgrade step; the metadata object is sufficient. This also is more in the sense of the tell-dont-ask principle.
- ftw.upgrade for the changes https://github.com/4teamwork/ftw.upgrade/pull/159 and https://github.com/4teamwork/ftw.upgrade/pull/158 .

### Tests

It is quite hard to make good tests for these mechanisms, which will lead us to bad tests (such as mock-tests) - the tests wont be able to really tell us whether "it works".
Therefore, from a economic point of view, writing tests would be way too expensive for the benefit.


opengever/core/upgrade.py
```diff
@@ -340,95 +371,89 @@ def is_postgres(self):
...
-        ).fetchall()
-        if result:
-            return
+        self.dialect_name = self.connection.dialect.name
+        self.metadata = MetaData(self.connection, reflect=True)
+        self.op = get_operations(self.connection)
+
+
+class GeverUpgradeStepRecorder(UpgradeStepRecorder):
+
```
deiferni: this thing deserves a docstring i'd say 😉 .

opengever/core/upgrade.py
```diff
@@ -340,95 +371,89 @@ def is_postgres(self):
...
+        return super(GeverUpgradeStepRecorder, self).is_installed(target_version)
+
+    def mark_as_installed(self, target_version):
+        super(GeverUpgradeStepRecorder, self).mark_as_installed(target_version)
+        self.mark_as_installed_in_sql(target_version)
 
-        self.execute(
-            versions_table.insert().values(profileid=self.profileid,
-                                           upgradeid=0))
+    def is_installed_marked_in_sql(self, target_version):
```
deiferni: 🔄  `is_marked_installed_in_sql`

jone:
@deiferni updated:
- added docsting to `GeverUpgradeStepRecorder`
- renamed method to `is_marked_installed_in_sql`
- rebased onto newest `master`

jone:
@deiferni My branch does not insert `profile-opengever.core:default`, but `opengever.core:default`, so I assume you did not check out my branch? 😏 

- it does create the tracking table on my machine on first access
- it inserts multiple records

bildschirmfoto 2018-02-01 um 18 56 55

deiferni:
@jone that may have been a leftover from the old tracking table. nevertheless i was on your branch 🤔 .

deiferni:
oh no. apparently i forgot to buildout and did not have the correct `ftw.upgrade` version 🙈 .

PR #3918, opened 2018-02-01
Add daily notification digest 

phgross:
- The digest is implemented as a separate notification dispatcher, which only sets the flag for all notifications which should be included in the next digest mail, according to the users notification settings. 
 - The setup entry point `send_digest`, will send a mail to all users with summary of all activities from the last 24 hours (grouped by reference). The entry-point should be called daily by a cron job.

bildschirmfoto 2018-02-01 um 01 08 04
bildschirmfoto 2018-02-01 um 01 08 21


opengever/activity/model/query.py
```diff
@@ -29,6 +32,10 @@ def by_subscription_roles(self, roles, activity):
 
         return self.filter_by(activity_id=activity.id).filter(Notification.userid.is_(None))
 
+    def from_last_24_hours(self):
+        since = datetime.now().replace(tzinfo=pytz.utc) - timedelta(hours=24)
```
deiferni: i would not architect it with a hardcoded timestamp. instead i'd build an endpoint that sends all unsent digest mails, and then track per digest-notification whether it has been included in the digest yet.

if we skip sending mails at some point, e.g. when we have a downtime due to a release or for other reasons, we should be able to just send the unsent digest mails regardless of the current time. this also allows for more flexibility for the timespan included in the digest. if we want to send the digest twice per day or only once per week for example, we can configure that with a cronjob per customer by using the same endpoint in gever.

opengever/activity/__init__.py
```diff
@@ -30,6 +34,12 @@ def notification_center():
         PloneNotificationMailer(), BadgeIconDispatcher(), DigestDispatcher()])
 
 
+def send_digest_zopectl_handler(app, args):
```
lukasgraf: A `zopectl` handler can be pretty risky from an operational standpoint. If a deployment's ZEO is stopped, and a cronjob tries to invoke that `zopectl` handler, that process will never timeout/terminate. Which means every night, a new process will be launched, until all filehandles or memory on the server are exhausted.

But, we already have that behavior with the `sync_ogds` zopectl handler. In my opinion, this needs to be solved as part of #2290 - so just pointing it out here.

phgross:
@deiferni @lukasgraf updated the PR according to your feedback and refactored the decision if a notification is part of a digest and if a digest mails should sent.
 - Introduced a flag `sent_in_digest` which tracks if a notification was part of a digest mail or not, this makes sure, that one activity is only contained by one digest mail.
 - Add a separate model `digest` which tracks, when the last digest mail has been sent per user. So the `DigestMailer` only send a digest mail if the digest-interval, currently 24h (23h because of a tolerance 1h), has been expired.

Please take another look, specially at the last 5 commits.

opengever/activity/digest.py
```diff
@@ -25,10 +25,9 @@ class DigestMailer(Mailer):
     template = ViewPageTemplateFile("templates/digest.pt")
 
     def get_notifications(self):
-        query = Notification.query.from_last_24_hours()
-        query = query.filter(Notification.is_digest.is_(True))
-
         notifications = {}
+
+        query = Notification.query.not_yet_sent_digest_notifications()
```
deiferni: not_yet_sent -> unsent

opengever/activity/tests/test_digest.py
```diff
@@ -23,29 +23,39 @@ def setUp(self):
...
-            DigestMailer().send_digests()
+        DigestMailer().send_digests()
 
         messages = [message_from_string(mail)
                     for mail in Mailing(self.portal).get_messages()]
         self.assertEquals(2, len(messages))
         self.assertEquals(['foo@example.com', 'robert.ziegler@gever.local'],
                           [message.get('To') for message in messages])
 
+    def test_sends_only_not_yet_sended_notifications(self):
```
deiferni: not_yet_sended -> unsent (forall, not gonna repeat 😉 )

phgross:
@deiferni updated ...

opengever/activity/templates/digest.pt
```diff
@@ -0,0 +1,294 @@
...
+                          <tr>
+                            <td style="padding: 40px; font-family: sans-serif; font-size: 15px; line-height: 140%; color: #555555;">
+                              <!-- Button : BEGIN -->
+                              <table role="presentation" cellspacing="0" cellpadding="0" border="0" align="center" style="margin: auto;">
+                                <tr>
+                                  <td style="border-radius: 3px; text-align: center;">
+                                    <a href="#"
+                                       class="button-a"
+                                       i18n:translate="label_notification_settings"
+                                       tal:attributes="href string:${portal_url}/@@personal-preferences"
```
lukasgraf: We need to use the public URL here to construct links. When called via a `zopectl` handler, the instance doesn't know about its canonical hostname - and we end up with the one [from `Testing.makerequest`, which is `foo`](https://github.com/zopefoundation/Zope/blob/2.13/src/Testing/makerequest.py#L53).

lukasgraf:
Otherwise looks good to me 👍  Didn't have time to do an in-depth review, but tested locally, and after the first :boom: during the upgrade it worked well:
- Upgrades of digest settings looks good
- After the upgrade on an existing site, the existing notifications (read as well as unread) don't cause a digest to be sent 👍 
- After enabling digests in the settings, digests get produced
- Already sent notifications don't get included in digests any more
- Even if the cronjob is triggered many times, the digest is only sent max. once per interval
- Mail formatting looks good (tested in Apple Mail and GMail).

opengever/activity/digest.py
```diff
@@ -0,0 +1,123 @@
...
+            'description': activity.translations[language].description}
+
+    def send_digests(self):
+        for userid, notifications in self.get_notifications().items():
+            # skip when digest interval is not expired yet
+            if not self.is_interval_expired(userid):
+                continue
+
+            msg = self.prepare_mail(
+                subject=_(u'subject_digest', default=u'Daily Digest'),
```
lukasgraf: I get mails with `subject_digest` as the actual title. I think this is because the subject never gets autotranslated because it doesn't go through the HTML template, but is instead used as a header directly.

```
subject = translate(_(u'subject_digest', default=u'Daily Digest'), domain='opengever.activity')
msg = self.prepare_mail(
subject=subject,
# ...
```
seems to work for me.

phgross:
@lukasgraf @deiferni updated ...

Issue #3716, opened 2017-12-01
Allow Admins (and Managers) to unlock documents locked by webdav.


deiferni:
This is a recurring support issue. At the moment solving this issue also includes a developer, which must perform the unlock. This is time-consuming and unnecessary.

We want to improve this situation and allow an `Administrator` (and `Manager`) to unlock a document. The `Administrator` is usually the person that decides if the unlock is justified (e.g. when a user left for holidays and forgot close officeconnector and checkin the document) and then opens the support issue.

A `WebDAV` locked document should be unlockable (for `Administrator` and `Manager`) just the same way as the document is unlockable when another user opens the `Edit medatata` view on a document:
![screen shot 2017-12-01 at 13 49 35](https://user-images.githubusercontent.com/736583/33483545-99c2d4b6-d69e-11e7-9ce6-896f8e5e0c2c.png)

For documentation purposes a short description of the current unlock process:
- create a locked document, checkout and edit the document, and keep officeconnector open.
- officeconnector will create a `WebDAV` lock. you should see the following info messages on the document overview:
![screen_shot_2017-12-01_at_12_32_36](https://user-images.githubusercontent.com/736583/33482255-6df4c930-d698-11e7-8f42-fa2a1900dcf5.png) The marked messages is the one displayed for the `WebDAV` lock.
- to force-unlock the `WebDAV` a `Manager` must open the **Zope** control panel, then navigate to `/Control_Panel/DavLocks/manage_main`
- in the form fill in the path relative to zope for the document's **parent**, e.g. `/fd/ordnungssystem/fuehrung/kommunikation/allgemeines/dossier-1/` and submit
- the lock can then be removed in the following form:
![screen shot 2017-12-01 at 13 38 35](https://user-images.githubusercontent.com/736583/33483197-f5c50a60-d69c-11e7-8f93-c305dda58459.png)

lukasgraf:
Also note that we already have a `@@lock_maintenance` view in `opengever.maintenance`.

This could probably be fleshed out.

njohner:
from @deiferni 
as discussed
- verify that both checkin views (with and without comment) remove the locks
- verify that this behavior is tested
- make sure that forcing checkin (checking in documens checked out by other users) is based on the `opengever.document: Force Checkin` permission
- actions in the file action button listing should check if they are still executable
- make sure that the (force) checkin action is also visible in the file action button listing:

[...]

PR #3919, opened 2018-02-01
Allow managers to clear webdav locks on documents.

njohner:
We had already overwritten the lock info view in `opengever/locking/info.py` to render different templates depending on the type of lock. The tested locks are all locks defined in `opengever`, so they have their own type and can be easily tested for. For the locks set by `officeconnector` there is no useful info in the `lock_info` so I had to overwrite that method to be able to identify them (by their `owner`).
I did not want to define a new lock type for these `officeconnector` locks as they really are of type `write`.

Closes https://github.com/4teamwork/opengever.core/issues/3716.


opengever/locking/browser/locking.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def force_unlock_unstealable(self, redirect=True):
+        """Steal the lock.
+
+        If redirect is True, redirect back to the context URL, i.e. reload
+        the page.
+        """
+        manager = getMultiAdapter((self.context, self.request),
+                                  ICheckinCheckoutManager)
+        manager.clear_locks()
```
Rotonen: This should already happen during a normal checkin and manager should have the permissions to just check the file in.

https://github.com/4teamwork/opengever.core/blob/e206bf0e7b27d0de3e2302ec0451261dfd37afba/opengever/document/checkout/manager.py#L146

What are we actually tackling here?

opengever/locking/info.py
```diff
@@ -84,3 +90,12 @@ def get_meeting_from_dossier_excerpt(self):
             return None
 
         return agenda_item.meeting
+
+    def lock_info(self):
```
Rotonen: We already have an API endpoint for exposing the document status.

https://github.com/4teamwork/opengever.core/blob/e206bf0e7b27d0de3e2302ec0451261dfd37afba/opengever/base/service.py#L24-L30

opengever/locking/browser/locking.py
njohner: This is for issue #3716.
But indeed, it seems that a manager can just check-in a locked document and this will unlock it. So I'm not sure what the point is here?

opengever/locking/info.py
njohner: Oh I hadn't found that one. But I think it wouldn't contain the information I seek. Also, if I understood @lukasgraf, I should not try to access that info from python, as I would be making a request while answering a request.

Rotonen: Two views on why I'm pointing that one out:

These deal with the same data so they should be broken out to draw their facts from a shared bit of code instead of duplicating the ~same queries.

If we need that info somewhere else it should also be amended to the API endpoint.

Rotonen:
Also do note the two tiers of review going on here at the same time: merge blockers and non-merge-blocking comments and queries.

njohner:
Yes, I know the tests are still missing for example. I just wasn't sure whether it was useful to correct this, as you pointed out that checkin also cleared the lock. @deiferni will also have a quick look.

njohner:
For sure the permissions would need to be set for admins and not managers to make sense.

opengever/locking/browser/locking.py
Rotonen: Figured this one out - the checkin action is not available on locked documents. Maybe just re-enable it for locked documents for managers?

deiferni:
as discussed
- verify that both checkin views (with and without comment) remove the locks
- verify that this behavior is tested
- make sure that forcing checkin (checking in documens checked out by other users) is based on the `opengever.document: Force Checkin` permission
- make sure that the (force) checkin action is also visible in the file action button listing:

screen shot 2018-02-27 at 13 30 05

phgross:
@njohner what's the state here? all done and ready for a review?

njohner:
@phgross @deiferni 
This is not ready no. The current state is that locked documents can be force-checked-in, which clears the locks (https://github.com/4teamwork/opengever.core/pull/4040).
I'm not sure whether the state with a webdav lock and no possibility to checkin a document still happens, so not sure whether this is still an issue (I could not reproduce such a state with new versions of OfficeConnector).
So should I finish this?

deiferni:
as discussed go for it ✅👍

njohner:
Ok this is ready for review. Checks should pass once we merge https://github.com/4teamwork/ftw.testbrowser/pull/86

I've mainly added the tests since last review. 
I have not ported the lock info to the api or merged with other codes giving information on a lock. The information we need here is very specific, to identify office connector locks and it mainly uses the `lock_info` from the `LockInfoViewlet` which serves as base class, only adding the information about the owner.

opengever/officeconnector/testing.py
```diff
@@ -296,22 +296,22 @@ def lock_document(self, browser, tokens, document):
...
                 )
 
             browser.webdav('LOCK', document, headers=headers, data=data)
             self.assertEquals(200, browser.status_code)
             self.assertTrue(lock_manager.is_locked())
 
-            lock_token = ET.fromstring(browser.contents).find(
-                './d:lockdiscovery/d:activelock/d:locktoken/',
-                {'d': 'DAV:'},
+            lock_token = browser.document.find(
```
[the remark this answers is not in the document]
Rotonen: Would be more accurate of a simulation (and a neat gever-side reminder) if you do the hack here too and parse `browser.contents`. Not bent on either way, though, but do leave a ´  # XXX - ...` comment in any case.

njohner:
Done

Issue #3359, opened 2017-09-08
Repo favorites JS looping or being recursively called


lukasgraf:
It looks like the JavaScript for fetching the repofolder favorites might be either looping, or issue a request recursively. I spotted the following requests to `repository-favorites/list` in close succession in the logs:

```
127.0.0.1 - john.doe [05/Sep/2017:14:54:04 +0200] "GET /VirtualHostBase/https/gever.example.org:443/ex/VirtualHostRoot/ordnungssystem/repository-favorites/list?cache_key=4fb59dff51aeb9b6539680a0a25e158c-john.doe HTTP/1.1" 200 544 "https://gever.example.org/ordnungssystem/some-repofolder" "Mozilla/5.0 (Windows NT 6.1; rv:47.0) Gecko/20100101 Firefox/47.0"
127.0.0.1 - john.doe [05/Sep/2017:14:54:04 +0200] "GET /VirtualHostBase/https/gever.example.org:443/ex/VirtualHostRoot/ordnungssystem/repository-favorites/list?cache_key=4fb59dff51aeb9b6539680a0a25e158c-john.doe HTTP/1.1" 200 544 "https://gever.example.org/ordnungssystem/some-repofolder" "Mozilla/5.0 (Windows NT 6.1; rv:47.0) Gecko/20100101 Firefox/47.0"
127.0.0.1 - john.doe [05/Sep/2017:14:54:04 +0200] "GET /VirtualHostBase/https/gever.example.org:443/ex/VirtualHostRoot/ordnungssystem/repository-favorites/list?cache_key=4fb59dff51aeb9b6539680a0a25e158c-john.doe HTTP/1.1" 200 544 "https://gever.example.org/ordnungssystem/some-repofolder" "Mozilla/5.0 (Windows NT 6.1; rv:47.0) Gecko/20100101 Firefox/47.0"
127.0.0.1 - john.doe [05/Sep/2017:14:54:04 +0200] "GET /VirtualHostBase/https/gever.example.org:443/ex/VirtualHostRoot/ordnungssystem/repository-favorites/list?cache_key=4fb59dff51aeb9b6539680a0a25e158c-john.doe HTTP/1.1" 200 544 "https://gever.example.org/ordnungssystem/some-repofolder" "Mozilla/5.0 (Windows NT 6.1; rv:47.0) Gecko/20100101 Firefox/47.0"
[...]

PR #3929, opened 2018-02-02
Prevent navigation tree from getting favorites multiple times.

bierik:
Closes https://github.com/4teamwork/opengever.core/issues/3359

Every time the navigation tree wants to mark an entry with a favorite
tag, it triggers an XHR for getting all favorites entries.
So use promises to keep the loading task of the favorites pending until the data is
retrieved by the ajax call.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2018.1.0 (unreleased)
 ---------------------
 
+- Prevent navigation tree from getting favorites multiple times. [Kevin Bieri]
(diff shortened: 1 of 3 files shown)


Rotonen:
Does this also guard for the variable access over the localstorage access or only for the network access over the localstorage access?

The caching is two-tiered.

deiferni:
@bierik 📥

bierik:
@deiferni I already discussed that with @Rotonen .
The problem that @Rotonen wanted to address is that the localstorage is probably accessed before the RAM cache in the variable. Accessing the localstorage also takes a reasonable amount of time and we should later have a look if this is actually the case. But for this PR the problem is solved and we do no longer access the network multiple times for the same data.

@lukasgraf updated

Issue #3930, opened 2018-02-02
Logout overlay is broken


lukasgraf:
Implementation of cross-tab logout in #3833 broke the **logout overlay** functionality.

**Steps to reproduce**

- Sign in with a Plone user (not `zopemaster`)
- Check out a document
- Choose the logout action from the personal tools menu

**Expected behavior**

A logout overlay should be displayed, warning the user about checked out documents and asking for confirmation before actually logging out:

![logout_overlay](https://user-images.githubusercontent.com/405124/35729343-985616aa-080e-11e8-9006-04ca2d6628de.png)

**Actual behavior**

Logout is triggered immediately. Also, in addition to skipping the logout overlay, user feedback is wonky (the menu entry link doesn't seem to visually register the click).

![logout2](https://user-images.githubusercontent.com/405124/35729499-2ce33cbc-080f-11e8-889c-4a30ac7295c8.gif)

Sometimes, it seems also a page refresh is missing entirely.

lukasgraf:
@bierik this one is **criticial**, please look into this ASAP!

lukasgraf:
Logout is still broken for browsers other than Chrome/Firefox. For example Safari or Mobile Chrome.

lukasgraf:
I didn't see the follow up PR #3932 - it's actually fixed now (tested in Chrome, FF, Safari, Chrome Mobile). Thanks @bierik

PR #3931, opened 2018-02-02
Fix logout overlay when used with cross tab logout.

bierik:
Closes https://github.com/4teamwork/opengever.core/issues/3930

### DIFF
--- a/opengever/base/browser/resources/logoutClient.js
+++ b/opengever/base/browser/resources/logoutClient.js
@@ -7,24 +7,28 @@
     logoutWorker = new SharedWorker(workerSrc);
     // Reload the tab when the LogoutBus sends a broadcast message
     logoutWorker.port.onmessage = function() { location.reload(); };
-    setupListeners();
   }
 
   // Prevent the default login behavior. Broadcast a logout message to all open gever tabs instead.
-  function trackLogout(e) {
-    e.preventDefault();
-    e.stopPropagation();
-    $.get(global.portal_url + '/logout').done(() => {
+  function logout(url) {
+    // Reject the promise immediately when the logout worker has not been initialized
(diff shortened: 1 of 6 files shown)


opengever/document/static/logout_overlay.js
```diff
@@ -1,46 +1,55 @@
...
+(function(global, $, logoutWorker) {
 
-    /* Workaround: If we just activate the prepOverlay on the link,
-    the event onBeforeLoad won't trigger. So we need to add a live-event on the
-    link at first. If we click on the logout link, the overlay is preparing.
-    To get the overlay, we have to click the button again programmatically.
-    */
-    $('[href*="logout_overlay"]').live('click', function(e){
-        e.preventDefault();
+  function logout(url) {
```
lukasgraf: Can you please add a comment here that describes what's being done?

(_Try logging out via service worker (Chrome, FF), to be found in `logoutClient.js`, if that fails lougout the old fashioned way_)

opengever/base/browser/resources/logoutClient.js
```diff
@@ -7,24 +7,24 @@
...
   }
 
   // Prevent the default login behavior. Broadcast a logout message to all open gever tabs instead.
-  function trackLogout(e) {
-    e.preventDefault();
-    e.stopPropagation();
-    $.get(global.portal_url + '/logout').done(() => {
+  function logout(url) {
+    if (!logoutWorker) {
+      return $.Deferred().promise().reject();
```
lukasgraf: Can you add some comments here when exactly / under what conditions `if (!logoutWorker)` will be true? Just IE, or are there other cases to be aware of?

bierik:
@lukasgraf updated

PR #3934, opened 2018-02-05
[Help needed] Sorting committee members by last name

### LINKED ISSUE #3905: Die Mitglieder eines Gremiums müssen alphabetisch nach Nachnamen sortiert dargestellt werden in der Übersichtsseite
https://github.com/4teamwork/opengever.core/blob/b6f3d5394daf295be5e7cf882158f0b352db267a/opengever/meeting/browser/committee.py#L53

Backref https://basecamp.com/2768704/projects/13732725/todos/339161197.

tarnap:
🚧 

The second and third commit do not solve the issue and can probably be removed.
The sorting is done in python (which is not the best solution).
Do you know how to improve this using SQL Alchemy?

Resolves #3905

### DIFF
--- a/opengever/meeting/browser/committee.py
+++ b/opengever/meeting/browser/committee.py
@@ -54,10 +54,7 @@ class CommitteeOverview(BrowserView, GeverTabMixin):
 
     def current_members(self):
         memberships = self.context.get_active_memberships().all()
-        members = [membership.member for membership in sorted(
-            memberships,
-            key=lambda membership: membership.member.lastname,
-        )]
+        members = [membership.member for membership in memberships]
         committee_container = self.context.get_committee_container()
         return [member.get_link(committee_container) for member in members]
 
--- a/opengever/meeting/browser/committee.py
+++ b/opengever/meeting/browser/committee.py
@@ -54,7 +54,10 @@ class CommitteeOverview(BrowserView, GeverTabMixin):
 
     def current_members(self):
         memberships = self.context.get_active_memberships().all()
-        members = [membership.member for membership in memberships]
+        members = [membership.member for membership in sorted(
+            memberships,
+            key=lambda membership: membership.member.lastname,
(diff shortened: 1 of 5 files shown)


opengever/meeting/browser/committee.py
```diff
@@ -54,7 +54,10 @@ def unscheduled_proposals(self):
 
     def current_members(self):
         memberships = self.context.get_active_memberships().all()
-        members = [membership.member for membership in memberships]
+        members = [membership.member for membership in sorted(
+            memberships,
+            key=lambda membership: membership.member.lastname,
```
deiferni: this also sorts, i guess this is not needed any more as the `relationship` is sorted already?

tarnap:
@deiferni 

We discussed this yesterday:
3a4288c does not sort
53637a2 does not sort either
fc7de6a actually sorts the memberships

Then you gave me this link: http://docs.sqlalchemy.org/en/latest/orm/loading_relationships.html#the-zen-of-joined-eager-loading

I'll use this to retry sorting in the database.

deiferni:
sorry i was notified ant thought this was ready now. when complete diff is small i usually don't have a look at the separate commit messages.

i have added a construction sign to the PR description 😉 .

tarnap:
What we're doing here is wrong.
What we try to get are the current members, and we do that, by querying the memberships and then getting each member.
We should rather query all members with a valid membership.

tarnap:
I'll do something like this:

https://stackoverflow.com/questions/8561470/sqlalchemy-filtering-by-relationship-attribute

http://docs.sqlalchemy.org/en/latest/orm/internals.html?highlight=has#sqlalchemy.orm.properties.RelationshipProperty.Comparator.has

tarnap:
Is a much better solution: https://github.com/4teamwork/opengever.core/pull/3940

PR #3938, opened 2018-02-05
Catch fuzzy translations

### LINKED ISSUE #2506: Amend the jenkins test script to fail on #fuzzy comments in compiled gettext files
Split off from https://github.com/4teamwork/opengever.core/pull/2496 .

Currently we do not monitor for #fuzzy comments in our compiled gettext locale files.

This should be a trivial shell glob + grep from the project root.

`if [ (! grep -q fuzzy opengever/*/locales/*/LC_MESSAGES/*.po) ]; then exit 1; fi`

The subshell is there to not trip up a potential `set -e`. Or we can just let the return value trip it up, but this design is cleaner for ever coming back to visit it. Explicitness and verbosity are IME values in contexts like bash and perl.

Rotonen:
Added an extra grep for 'fuzzy' to before and after compiling the locales.

Closes #2506

### DIFF
--- a/test-i18n-de.cfg
+++ b/test-i18n-de.cfg
@@ -22,6 +22,16 @@ input = inline:
 
     OUTPUT=""
 
+    # Catch fuzzy translations from committed locales
+    OUTPUT+=$(grep -nA5 fuzzy opengever/*/locales/*/LC_MESSAGES/*.po)
+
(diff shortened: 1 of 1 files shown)


Rotonen:
@deiferni here you go:
https://ci.4teamwork.ch/builds/144752/tasks/227486

PR #3951, opened 2018-02-07
Fix select2 value dropdowns, which has been hidden by the exposator.

phgross:
The z-index of the value dropdown has been alredy fixed in this PR https://github.com/4teamwork/plonetheme.teamraum/pull/605, but opengever.core
does not make sure that the ftw.keywordwidget css files are loaded
before the gever.css.
bildschirmfoto 2018-02-07 um 17 32 39

Change includes an upgradestep which fixes the order also for existing deployments.


deiferni:
(rebased on your behalf)

PR #3952, opened 2018-02-07
Fix mimetype registries.

deiferni:
This PR fixes the various mimetype registries in such a way that they yield results as epxpected by bumblebee, see also https://github.com/4teamwork/ftw.bumblebee/blob/bc7b6f9ad9efe7bffd91a50b7a984797b0018d32/ftw/bumblebee/mimetypes.py#L19-L64.

Bumblebee expects the extension to be set for a mime-type in the plone mimetypes registry:
- Add `bmp` extension to `image/bmp`
- Add `ini` extension to `text/plain`

Also we need to register `ini` with python:
- Treat `ini` as `text/plain`

### DIFF
--- a/opengever/policy/base/profiles/mimetype/mimetypes.xml
+++ b/opengever/policy/base/profiles/mimetype/mimetypes.xml
@@ -22,6 +22,7 @@
                  image/x-bmp
                  image/x-MS-bmp"
       icon_path="icon_dokument_bilddatei.gif"
+      extensions="bmp"
       />
 
   <mimetype
@@ -32,6 +33,7 @@
   <mimetype
       mimetypes="text/plain"
       icon_path="icon_dokument_text.gif"
+      extensions="txt diff conf def in text pot ksh list log asc ini"
       />
 
   <mimetype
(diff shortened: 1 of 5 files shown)


opengever/policy/base/profiles/mimetype/mimetypes.xml
```diff
@@ -32,6 +33,7 @@
   <mimetype
       mimetypes="text/plain"
       icon_path="icon_dokument_text.gif"
+      extensions="txt diff conf def in text pot ksh list log asc ini"
```
Rotonen: I'd rather leave gettext files for programs like poedit.

deiferni: i only added the `ini` here, the rest comes from somewhere else. i need to overwrite the whole attribute though as it cannot extend.

Rotonen: `.pot` is what I'm eyeing there. But I suppose that's a triviality which can be gotten to later, if ever.

PR #3963, opened 2018-02-09
Add titles to the builder groups

Rotonen:
🚧 pending on https://github.com/4teamwork/opengever.ogds.models/pull/52 🚧 

I'm a bit confused as to this one, seems we already have all the functionality, we're just not testing it nor using it in the case of the basecamp issue.

Added a test for the case presented here and tackling the actual data sourcing issue in the policy package.


lukasgraf:
@Rotonen I merged the dependency, could you please rebase and switch `opengever.ogds.models` back to `master` (no branch spec)?

Rotonen:
@lukasgraf dropped the branch pin and rebased.

PR #3969, opened 2018-02-11
Add REST API endpoints for trashing/untrashing documents

buchi:


opengever/api/trash.py
```diff
@@ -0,0 +1,39 @@
...
+import plone.protect.interfaces
+
+
+class Trash(Service):
+    """Trash a document"""
+
+    def reply(self):
+
+        # Disable CSRF protection
+        if 'IDisableCSRFProtection' in dir(plone.protect.interfaces):
```
lukasgraf: That check isn't necessary here - we have a [hard dependency on `plone.protect >= 3.0.15`](https://github.com/4teamwork/opengever.core/blob/master/setup.py#L139) in GEVER, that request layer will always be importable (same for `Untrash` below).

opengever/api/configure.zcml
```diff
@@ -35,6 +36,22 @@
       permission="opengever.document.Cancel"
       />
 
+  <plone:service
+      method="POST"
+      name="@trash"
+      for="opengever.document.document.IDocumentSchema"
```
lukasgraf: If mails (`ftw.mail.mail`) should also be trashable (which they are via the UI), this should be registered for `opengever.document.behaviors.IBaseDocument`.

lukasgraf:
Could you please also add a short mention of this endpoint to the [GEVER API docs](https://docs.onegovgever.ch/dev-manual/api/)? (`docs/public/dev-manual/api/` - maybe similar to the already documented `@scanin` endpoint).

opengever/api/trash.py
buchi: check removed

opengever/api/configure.zcml
buchi: changed to `opengever.document.behaviors.IBaseDocument`

buchi:
Moved permission and condition checks from browser view to trasher adapter.
Added some short docs.

PR #3970, opened 2018-02-11
Restapi items fieldlist

buchi:
It's now possible to specify additional fields that should be returned in summaries of GET requests.

The requested fields are given with the url parameter `items.fl`.
Example: `items.fl=filename,filesize,modified`


lukasgraf:
Could you please have a look if there's a good place in the [GEVER API docs](https://docs.onegovgever.ch/dev-manual/api/) where this could be documented? As far as I can tell, this mainly affects content `GET` and `@search`.

lukasgraf:
@buchi rebased, added docs, added support for `Creator`

PR #3976, opened 2018-02-12
Add missing translation domain.

deiferni:
Add missing ftw.tabbedview domain as we translate the label in that domain for all listings.

Fixed:
screen shot 2018-02-12 at 20 00 50

Before:
1feace2e-1018-11e8-87a1-a0369f35f0c0

Backref: https://basecamp.com/2768704/projects/12554551/todos/341841512

### DIFF
--- a/opengever/tabbedview/browser/selection_with_filters.pt
+++ b/opengever/tabbedview/browser/selection_with_filters.pt
@@ -1,7 +1,7 @@
 <div class="tabbedview_select" i18n:domain="ftw.tabbedview">
 
     <tal:condition tal:condition="view/show_selects">
-      <span i18n:translate="">Choose</span>
+      <span i18n:domain="opengever.tabbedview" i18n:translate="">Choose</span>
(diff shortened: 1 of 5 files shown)


opengever/tabbedview/browser/selection_with_filters.pt
```diff
@@ -1,7 +1,9 @@
 <div class="tabbedview_select" i18n:domain="ftw.tabbedview">
 
     <tal:condition tal:condition="view/show_selects">
-      <span i18n:translate="">Choose</span>
+      <tal:domain i18n:domain="opengever.tabbedview">
+        <span i18n:translate="">Choose</span>
```
lukasgraf: Can't you stick the `i18n:domain="opengever.tabbedview"` attribute on the `<span>` element? Or did that not work?

deiferni: Indeed it also works on the `span` 👍

PR #3977, opened 2018-02-12
Combine similar transitions to one in the activity settings page.

phgross:
There are some different task-transitions which ends in the same review_state, therefore ther are multiple options in the activity settings page for closing a task. Thats confusing for the user, therefore we combine them together and introduce a list of "aliased transitions".


lukasgraf:
@phgross upgrade step for migrating existing notification settings was missing. I hijacked your branch and added the missing upgrade step.

@deiferni as discussed.

PR #3984, opened 2018-02-14
Enable Mind Manager with Office Connector

buchi:
Mind Manager is supported in Office Connector 1.5.0 and later.


buchi:
From a technical point of view making a changelog entry is wrong. This PR amends a change that's already mentioned in the changelog for the next release. Also the current release already "supports" Mind Manager as there's no restriction of mimetypes.

We have a changelog entry in Office Connector for the marketing team.

PR #3990, opened 2018-02-15
Fix upgrade steps for oracle.

deiferni:
Creating the tracking table failed for oracle due to a different auto naming schema.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2018.1.0 (unreleased)
 ---------------------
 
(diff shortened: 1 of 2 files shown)


docs/HISTORY.txt
```diff
@@ -5,6 +5,7 @@ Changelog
 2018.1.0 (unreleased)
 ---------------------
 
+- Fix tracking table creation for oracle. [deiferni]
```
jone: Actually not creating is the problem, but migrating it 🤓

deiferni: True dat.

PR #3993, opened 2018-02-16
Anpassung Printscreen Stellvertertung-Dropdown

livschal:


deiferni:
This may need a rebase once we fix a test problem with flaky tests.

Issue #3936, opened 2018-02-05
Meeting TOC spinner gets stuck on IE11


Rotonen:
The icon spins, but when the download fires, it does not reset back to the download icon, instead the spinner icon remains frozen.

The icon does not spin on a subsequent click.

bierik:
Fixed through https://github.com/4teamwork/opengever.core/pull/3994

PR #3994, opened 2018-02-16
Fix toc download on IE11 and Edge

bierik:
https://github.com/4teamwork/opengever.core/issues/3936

IE11 and Edge do not support `blob` as repsonse type for XHR requests.
So use `arraybuffer` and build the blob manually.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2018.1.3 (unreleased)
 ---------------------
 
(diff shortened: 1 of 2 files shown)


deiferni:
@Rotonen as you have tested this yesterday, what is the state here?

- i also think this should go into `2018.1.x`
- could you provide a CL entry on behalf of @bierik and :shipit: ?

Rotonen:
Still pending testing on Windows 7 as Windows updates kicked in and forgot to get back to it.

I'll hijack the branch.

PR #4003, opened 2018-02-19
Solr schema fixes

buchi:


solr-conf/managed-schema
```diff
@@ -106,6 +106,7 @@
     <!-- Plone default fields -->
     <field name="allowedRolesAndUsers" type="string" indexed="true" stored="false" multiValued="true"/>
     <field name="created" type="pdate" indexed="true" stored="false" />
+    <field name="Creator" type="string" indexed="false" stored="false" />
```
lukasgraf: `indexed="false"` and `stored="false"`? That can't be right, can it? Or am I missing something?

buchi: We're using DocValues for string fields (see https://wiki.apache.org/solr/DocValues). Thus indexing is not always needed and storing is never needed.
However I changed to `indexed="true"` because that should be faster for filtering.

PR #4021, opened 2018-02-22
Trivialities of `bin/mtest`

Rotonen:
Observations piled up enough to take a small diversion:

1) We can save time by explicitly resetting buildout parts to the relevant ones for the main build
2) We should use `test-command-no-coverage` to fully adhere to the intentions of our buildouts
3) We should run the builds with the debug flag for better user feedback (and recorded timing data)
4) We should explicitly set the parallelity to match the build weight config

### DIFF
--- a/test-plone-4.3.x.cfg
+++ b/test-plone-4.3.x.cfg
@@ -16,4 +16,4 @@ eggs +=
     ${buildout:hotfix-eggs}
 
 [test-jenkins]
-test-command = bin/mtest -d -j5 $@
+test-command-no-coverage = bin/mtest -d -j5 $@
--- a/test-plone-4.3.x.cfg
+++ b/test-plone-4.3.x.cfg
@@ -16,4 +16,4 @@ eggs +=
     ${buildout:hotfix-eggs}
 
 [test-jenkins]
-test-command = bin/mtest -d $@
+test-command = bin/mtest -d -j5 $@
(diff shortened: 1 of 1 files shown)


jone:
As `development.cfg` extends from `test-plone-4.3.x.cfg`, I expect this change to remove all those tools for development too, which is probably not the intention.

Rotonen:
True. Back to the drawing board.

deiferni:
I'd say this should be mergeable without the part-resetting?

Rotonen:
The part-resetting is the major time saver here. But sure, will rebase to exclude that and get back to that on better time.

Rotonen:
Currently `bin/mtest` blindly assumes one has hyper threading and goes for "half of thread count minus one" when given no parallelity parametre via an environment variable or command line argument. The command line argument takes precedence over the environment variable as I'm considering that more specific. I've changed the buildout to be explicit over the command line parametre rather than relying on the Jenkins-side environment variable as we now have the build weight config in the repository and that'd then be off if the environment variable changes for some reason.

Would one like to be in the driver's seat for local test runs, one can supply an integer of choice for the `-j` parametre. See also the `-h` parametre.

jone:
Sounds fair 👍

Rotonen:
Locally I often just run `bin/mtest -d -j"$(getconf _NPROCESSORS_ONLN)"`, but the desired impact on workstation interactivity is up to each developer to decide on their own. Thus the rather conservative default parallelity.

PR #4030, opened 2018-02-27
Use ZCML instead of provideAdapter.

### LINKED ISSUE #3928: Test isolation failure for tasks
I had a test failure in a sequential test run of the whole suite in `test_reassign_task_in_progress_state_to_a_team_isnt_possible (opengever.task.tests.test_assign.TestAssignTask)`.

```
AssertionError: Lists differ: ['Team responsibles are only a... != []

First list contains 1 additional elements.
First extra element 0:
'Team responsibles are only allowed if the task or forwarding is open.'

- ['Team responsibles are only allowed if the task or forwarding is open.']
+ []
```

Superficially seems like some leak in the workflow state of a fixtured task.

jone:
`provideAdapter` should never be used on module level because:
1. it cannot be isolated in tests
2. it is not consistent: all other adapters are registered in ZCML.

Fixes #3928

### DIFF
--- a/opengever/dossier/archive.py
+++ b/opengever/dossier/archive.py
@@ -21,7 +21,7 @@ from ZODB.POSException import ConflictError
 from zope import schema
 from zope.annotation.interfaces import IAnnotations
 from zope.component import adapter
-from zope.component import getUtility, provideAdapter
+from zope.component import getUtility
 from zope.interface import implementer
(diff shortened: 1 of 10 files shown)


jone:
Tests are now fixed and did the mandatory pep8 cleanup.
The docstring is irrelevant for this change; it explains how the component can be used.

PR #4034, opened 2018-02-28
Do not check for delete permission in `@move` endpoint

buchi:
Customize `@move` endpoint and comment out delete permission checks.

### DIFF
--- a/opengever/api/move.py
+++ b/opengever/api/move.py
@@ -1,0 +1,76 @@
+# -*- coding: utf-8 -*-
+from Acquisition import aq_parent
+from Products.CMFCore.utils import getToolByName
+from plone.restapi.deserializer import json_body
+from plone.restapi.services.copymove.copymove import Move
+from zExceptions import BadRequest
+from zope.interface import alsoProvides
+from zope.security import checkPermission
+
+import plone
+
+
+class Move(Move):
+    """Moves existing content objects.
+    """
+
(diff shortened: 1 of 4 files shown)


opengever/api/move.py
```diff
@@ -0,0 +1,73 @@
...
+                         plone.protect.interfaces.IDisableCSRFProtection)
+
+        if not isinstance(source, list):
+            source = [source]
+
+        parents_ids = {}
+        for item in source:
+            obj = self.get_object(item)
+            if obj is not None:
+                # if self.is_moving:
```
lukasgraf: Can you add a comment here *why* this block needs to be commented out?

opengever/api/move.py
```diff
@@ -0,0 +1,73 @@
...
+import plone
+
+
+class Move(Move):
+    """Moves existing content objects.
+    """
+
+    # Copied from plone.restapi 1.1.0
+    # Disables DeleteObjects permission check
+    def reply(self):
```
deiferni: are we planning to fix this in `plone.restapi` also? @buchi we had a short discussion about this some days ago and iirc we couln't see a reason to duplicate the checks here. If this is planned a comment that we can remove this implementation eventually would be good.

opengever/api/move.py
buchi: I've added an additional comment

PR #4035, opened 2018-02-28
Add REST API endpoint for querying configuration settings

buchi:
Provides a new endpoint for querying configuration settings that may be needed in clients.

Currently it returns:
- maximum subdossier depth
- maximum repository depth
- activated and disabled features

Example Response:
```json
{
    "@id": "http://localhost:8080/fd/@config",
    "features": {
        "activity": true,
        "bumblebee": false,
        "contacts": false,
        "dossier_templates": true,
        "meetings": true,
        "officatwork": true,
        "solr": true,
        "workspace": false
    },
    "maximum_repository_depth": 3,
    "maximum_subdossier_depth": 1
}
```

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -10,6 +10,8 @@ Changelog
 - Fix deleting agenda items for non-word agenda items. [deiferni]
 - Improve checking for truthy dates in sablon template data. [deiferni]
 - Fix REST API @move endpoint: Do not require delete permission when moving objects. [buchi]
+- Add REST API endpoint for querying configuration settings. [buchi]
+
 
 2018.1.3 (2018-02-20)
 ---------------------
--- a/opengever/api/config.py
+++ b/opengever/api/config.py
@@ -1,0 +1,69 @@
+from opengever.activity.interfaces import IActivitySettings
+from opengever.base.interfaces import ISearchSettings
+from opengever.bumblebee.interfaces import IGeverBumblebeeSettings
+from opengever.contact.interfaces import IContactSettings
+from opengever.dossier.dossiertemplate.interfaces import IDossierTemplateSettings
+from opengever.dossier.interfaces import IDossierContainerTypes
+from opengever.dossier.interfaces import ITemplateFolderProperties
+from opengever.ech0147.interfaces import IECH0147Settings
+from opengever.meeting.interfaces import IMeetingSettings
+from opengever.officeatwork.interfaces import IOfficeatworkSettings
+from opengever.officeconnector.interfaces import IOfficeConnectorSettings
+from opengever.repository.interfaces import IRepositoryFolderRecords
+from opengever.workspace.interfaces import IWorkspaceSettings
+from plone import api
+from plone.restapi.services import Service
+
+
+class Config(Service):
+    """GEVER configuration"""
+
+    def reply(self):
+        config = {}
+
+        config['@id'] = api.portal.get().absolute_url() + '/@config'
+
+        config['max_repositoryfolder_levels'] = api.portal.get_registry_record(
+            'maximum_repository_depth', interface=IRepositoryFolderRecords)
+
+        config['max_dossier_levels'] = api.portal.get_registry_record(
+            'maximum_dossier_depth', interface=IDossierContainerTypes) + 1
+
+        config['features'] = {}
+
+        config['features']['activity'] = api.portal.get_registry_record(
+            'is_feature_enabled', interface=IActivitySettings)
+        config['features']['contacts'] = api.portal.get_registry_record(
+            'is_feature_enabled', interface=IContactSettings)
+        config['features']['doc_properties'] = api.portal.get_registry_record(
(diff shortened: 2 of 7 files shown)


jone:
There is also a list of features here: https://github.com/4teamwork/opengever.core/blob/master/opengever/testing/integration_test_case.py#L37
Wouldn't it make sense to extract both into one common list?
I also think that we have more features in the tests than here at the moment 😉

opengever/api/config.py
```diff
@@ -0,0 +1,47 @@
...
+            'is_feature_enabled', interface=IActivitySettings)
+        config['features']['bumblebee'] = api.portal.get_registry_record(
+            'is_feature_enabled', interface=IGeverBumblebeeSettings)
+        config['features']['contacts'] = api.portal.get_registry_record(
+            'is_feature_enabled', interface=IContactSettings)
+        config['features']['dossier_templates'] = api.portal.get_registry_record(
+            'is_feature_enabled', interface=IDossierTemplateSettings)
+        config['features']['meetings'] = api.portal.get_registry_record(
+            'is_feature_enabled', interface=IMeetingSettings)
+        config['features']['officatwork'] = api.portal.get_registry_record(
```
lukasgraf:
```
officatwork  -> officeatwork
                     ^
```

opengever/api/config.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def reply(self):
+        config = {}
+
+        config['@id'] = api.portal.get().absolute_url() + '/@config'
+
+        config['maximum_repository_depth'] = api.portal.get_registry_record(
+            'maximum_repository_depth', interface=IRepositoryFolderRecords)
+
+        config['maximum_subdossier_depth'] = api.portal.get_registry_record(
```
lukasgraf: As discussed, I would use this opportunity to use constistent semantics for these nesting depth settings (`maximum_subdossier_depth` refers to only *sub*dossiers, whereas `maximum_repository_depth` refers to nesting of repository folders inside themselves.

I would therefore simply call them
- `max_dossier_levels`
- `max_repositoryfolder_levels`

and adjust on of them by one accordingly compared to the internally stored value.

(I find `_level` easier to visualize, personally).

The meaning then is:
`1` - no nesting allowed, so no sub-dossiers or sub-repofolders
`2` - exactly one sub-object allowed
etc...

@deiferni thoughts?

docs/HISTORY.txt
```diff
@@ -8,6 +8,7 @@ Changelog
 - Remove all uses of enableHTTPCompression in GEVER. [lgraf]
 - Fix deleting agenda items for non-word agenda items. [deiferni]
 - Improve checking for truthy dates in sablon template data. [deiferni]
+- Add REST API endpoint for querying configuration settings. [buchi]
```
lukasgraf: Documentation is missing for https://docs.onegovgever.ch/dev-manual/api/

lukasgraf:
@jone I agree. I'm also less than happy that we're duplicating that logic here, and inconsistently so. But apparently this is urgent.

So I see this as a chance to commit to a "public" configuration schema (a small subset of what's currently configurable, but still). Once we publish this, and clients rely on it, we can't change it in non-backwards compatible ways.

Therefore we can consider this the canonical source of truth for naming settings in the future, eventually refactor names in our internal codebase, and use this to inform names in other contexts where we need to specify settings (OGGBundle import, policies, OGGBundle Validator, Excel sheets for Onboarding of new customers, tests)

opengever/api/config.py
deiferni: As discussed i agree, i also find levels easier to understand 👍 . Also since this is read-only we don't have to bother with preventing someone from setting `0` as the value yet.

opengever/api/config.py
```diff
@@ -0,0 +1,47 @@
...
+            'maximum_dossier_depth', interface=IDossierContainerTypes)
+
+        config['features'] = {}
+        config['features']['activity'] = api.portal.get_registry_record(
+            'is_feature_enabled', interface=IActivitySettings)
+        config['features']['bumblebee'] = api.portal.get_registry_record(
```
deiferni: As already suggested in a discussion by @lukasgraf i'd not call this `bumblebee` but rather `pdf_preview`, or maybe just `preview`. The name `bumblebee` is internal only and should not be exposed to customers/API clients.

opengever/api/config.py
buchi: Fixed.

opengever/api/config.py
buchi: changed to `max_repositoryfolder_levels` and `max_dossier_levels`.

opengever/api/config.py
buchi: Renamed to `preview`.

buchi:
I've added some more features found in https://github.com/4teamwork/opengever.core/blob/master/opengever/testing/integration_test_case.py#L37

Also added docs for `@config` endpoint.

PR #4036, opened 2018-03-01
Correct French translations

njohner:
In French translations:
* Correct a typo
* Change "appendices" in two more cases (generally replace by "pièces jointes", but in one case here simply by "documents additionnels" which made more sense in that context, as it is a confirmation of an action that was termed"Soumettre des documents additionnels")

resolves https://extranet.4teamwork.ch/support/edk/tracker-support-gever-edk/83

Other instances of this translation had been corrected in
https://github.com/4teamwork/opengever.core/pull/3811


njohner:
Although the translations were corrected and merged end of December from https://extranet.4teamwork.ch/support/edk/tracker-support-gever-edk/83, the client complained that it was not done. So I guess their opengever was not updated since?

PR #4037, opened 2018-03-01
Generalize mtest more

Rotonen:
I've made mtest more usable for local development. It now splits the tests across processes independent of preconfigured layer split lists and accepts --layer and -m arguments. It also now runs heaviest-by-classcount layer batches first.

For running on a single module, for example `bin/test -j3 -m opengever.document` is still faster, but when running on multiple modules, for example `bin/mtest -d -j3 -m opengever.core -m opengever.base -m opengever.repository -m opengever.dossier -m opengever.document -m opengever.mail` mtest ends up quicker. This has to do with mtest incurring layer setup costs when splitting within a layer and `bin/test` not splitting within a layer, but merely running layers in parallel.

### DIFF
--- a/bin/mtest
+++ b/bin/mtest
@@ -30,6 +30,13 @@ import locale
 import sys
 
 
+# From https://gist.github.com/moshekaplan/4678925
+def chunks(chunkable, n):
+    """Yield successive n-sized chunks from l."""
+    for i in xrange(0, len(chunkable), n):
+        yield tuple(chunkable[i:i + n])
+
+
 def split_into_batches(
         iterable,
         chunk_count=None,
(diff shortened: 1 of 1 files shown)


test-plone-4.3.x.cfg
```diff
@@ -10,10 +10,13 @@ package-name = opengever.core
 package-namespace = opengever
 test-egg = opengever.core[api, tests]
 
+parts =
+    test
+    test-jenkins
```
jone: I'm still not convinced by this change 😜 breaks `development.cfg`.

Rotonen: Rebase failure, this is not supposed to be here, sorry.

Rotonen:
@jone rebased, ready for rereview.

bin/mtest
```diff
@@ -30,6 +30,34 @@ import locale
...
+    length = len(iterable)
+
+    if length / float(minimum_chunk_count) < minimum_chunk_size:
+        return (iterable, )
+
+    if not chunk_count:
+        chunk_count = max(length / optimal_chunk_size, minimum_chunk_count)
+    chunksize = ceil(length / float(chunk_count))
+
+    batches = [[] for _ in range(chunk_count)]
```
deiferni: as discussed maybe rewrite with a batcher https://gist.github.com/moshekaplan/4678925 to improve readability
note: you can also slice tuples, so the conversion might not be necessary

Rotonen:
@deiferni amended.

PR #4038, opened 2018-03-02
Fix restapi reference number

buchi:
- Customize serializer for dossiers and documents to return the correct reference number.
- Make reference_number a read-only field to prevent setting it through the REST API.

Having the reference number as a field in the schema is highly questionable for me.
I'm keeping it in the edit form for now as removing it, is out of scope of this PR.

Having it in the schema seems to confuse people as various tests show that try to set/get the value of the reference number field.
Further the document schema doesn't contain the reference number field.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -12,6 +12,7 @@ Changelog
 - Fix REST API @move endpoint: Do not require delete permission when moving objects. [buchi]
 - Add REST API endpoint for querying configuration settings. [buchi]
 - Make sure notification-default for the forwarding-added kind exists. [phgross]
+- Fix reference number in REST API GET. [buchi]
 
 
 2018.1.3 (2018-02-20)
--- a/opengever/api/configure.zcml
+++ b/opengever/api/configure.zcml
@@ -12,6 +12,9 @@
   <adapter factory=".summary.GeverJSONSummarySerializer" />
   <adapter factory=".fields.ChoiceFieldDeserializer" />
 
+  <adapter factory=".dossier.SerializeDossierToJson" />
+  <adapter factory=".document.SerializeDocumentToJson" />
+
   <plone:service
       method="POST"
       name="@checkout"
--- a/opengever/api/document.py
+++ b/opengever/api/document.py
@@ -1,0 +1,20 @@
+from opengever.base.interfaces import IReferenceNumber
+from opengever.document.behaviors import IBaseDocument
+from plone.restapi.interfaces import ISerializeToJson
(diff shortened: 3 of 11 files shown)


lukasgraf:
As discussed, I very much like where this is going in the grand scheme of things.

However, I have some concerns regarding the `reference_number` field being used in `OGGBundle` imports to control the reference numbers of created content. I need to check though if and how that actually is being done, I'm not sure that ever worked as intended.

lukasgraf:
@buchi this indeed will break setting reference numbers on bundle import.
The field `reference_number` is [read from the bundle JSON here](https://github.com/4teamwork/opengever.core/blob/master/opengever/bundle/sections/constructor.py#L109-L125).

lukasgraf:
I just checked, the behavior for setting reference numbers upon bundle import is not affected by this change, it's just the tests (which were silly before) that need to be rewritten.

Instead of
```
        self.assertEqual(
            '7',
            IDossier(dossier_peter).reference_number)
```

we should test for

```
self.assertEqual('Client1 0.3 / 7', IReferenceNumber(dossier_peter).get_number())
```

(And not drop the `reference_number` data from the `opengever/bundle/tests/assets/basic.oggbundle/dossiers.json`)

buchi:
I've readded the reference number assertions in `test_oggbundle_pipeline.py`.
However there's a test failure if I readd the reference number in https://github.com/4teamwork/opengever.core/pull/4038/files#diff-2d56dc11275e67653eb1d207e02b001fL7
The assertion failing is in `assert_document_6_created`.
@lukasgraf Maybe you can look at that and provide some explanation?

lukasgraf:
@buchi interesting - with this change on top of your branch, oggbundle pipeline tests pass for me (in isolation):

```diff
diff --git a/opengever/bundle/tests/assets/basic.oggbundle/dossiers.json b/opengever/bundle/tests/assets/basic.oggbundle/dossiers.json
index 606c742..97d357c 100644
--- a/opengever/bundle/tests/assets/basic.oggbundle/dossiers.json
+++ b/opengever/bundle/tests/assets/basic.oggbundle/dossiers.json
@@ -5,6 +5,7 @@
     "comments": "Vreni Meier ist ein Tausendsassa",
     "keywords": [],
     "relatedDossier": [],
+    "reference_number": "1",
     "responsible": "lukas.graf",
     "review_state": "dossier-state-active",
     "start": "2010-11-11",
```

buchi:
@lukasgraf Yep. Tests work. I've added the line to the wrong dossier in my first attempt.

PR ready.

PR #4039, opened 2018-03-02
Enhance debug output of mtest

Rotonen:
An example of a normal run failure output:

https://ci.4teamwork.ch/builds/149436/tasks/235825


Rotonen:
Still amended the command lines onto the start and done debug messages as that seemed to be desired.

Test output:
https://ci.4teamwork.ch/builds/149464/tasks/235883

Anything else (and trivial) on the wishlist for now?

PR #4040, opened 2018-03-05
Allow force checkin for users with `Manage portal` permissions

### LINKED ISSUE #3716: Allow Admins (and Managers) to unlock documents locked by webdav.

njohner:
- verify that both checkin views (with and without comment) remove the locks:
  - yes, both actions call the `checkin` method of the `CheckinCheckoutManager`
- verify that this behavior is tested.
  - Behavior is tested for checkin without comment
  - Added a test to check that force checkin by another user 
- make sure that forcing checkin (checking in documents checked out by other users) is based on the `opengever.document: Force Checkin` permission
  - There are two permissions that allow force checkin:
    - `Force Checkin`
    - `Manage portal`
  - I also added a test that `Force Checkin` gives the permission. Note that with just this permission, the checkin will raise an `Unauthorized` probably because the user cannot save the file (the file gets checked in though and the lock cleared). So this is a bit of a strange situation, but it is also an artificial situation from the test setting.
- make sure that the (force) checkin action is also visible in the file action button listing:
  - Corrected the condition for showing the `checkin` action buttons
- actions in the file action button listing should check if they are still executable:
  - We now check whether the checkin is allowed when executing it.
  - We only allow checkin with comment for a force checkin (we actually check that the comment is not empty in that case)

**`checkin with comment` is now shown for administrators when force checkin is allowed**

screen shot 2018-03-05 at 10 33 54

**Simple `checkin` or checkin with an empty comment will not work for force checkin and display an error message**

screen shot 2018-03-05 at 11 06 20

resolves #3716

### DIFF
--- a/opengever/document/checkout/checkin.py
+++ b/opengever/document/checkout/checkin.py
@@ -230,6 +230,11 @@ class CheckinDocumentWithoutComment(BrowserView):
         self.checkin_controller = CheckinController(self.request)
 
     def __call__(self):
+        manager = getMultiAdapter((self.context, self.request),
+                                  ICheckinCheckoutManager)
+        if not manager.is_checkin_without_comment_allowed():
+            checkin_with_comment = CheckinDocuments(self.context, self.request)
+            return checkin_with_comment.__call__()
         self.checkin()
         return self.redirect()
 
@@ -245,6 +250,9 @@ class CheckinDocumentsWithoutComment(CheckinDocumentWithoutComment):
 
     This view is called from a tabbed_view.
     """
+    def __call__(self):
+        self.checkin()
+        return self.redirect()
 
     def checkin(self):
         try:
--- a/opengever/document/tests/test_checkout.py
+++ b/opengever/document/tests/test_checkout.py
@@ -36,6 +36,45 @@ import jwt
 import transaction
 
 
+class TestCheckinIntegration(IntegrationTestCase):
+
+    def test_manager_can_force_checkin(self):
+        self.login(self.regular_user)
+        manager = getMultiAdapter((self.document, self.portal.REQUEST),
+                                  ICheckinCheckoutManager)
+        manager.checkout()
+        self.assertEqual(self.regular_user.getName(),
+                         manager.get_checked_out_by())
+
+        # Dossier Manager can't force checkin
+        self.login(self.dossier_manager)
+        with self.assertRaises(Unauthorized):
+            manager.checkin(comment="Force checkin")
+
+        # Dossier Responsible can't force checkin
+        self.login(self.dossier_responsible)
+        with self.assertRaises(Unauthorized):
+            manager.checkin(comment="Force checkin")
+
+        # Administrator can force checkin
+        self.login(self.administrator)
+        manager.checkin(comment="Force checkin")
+        self.assertIsNone(manager.get_checked_out_by())
+
+    def test_force_checkin_clears_lock(self):
+        self.login(self.regular_user)
+        manager = getMultiAdapter((self.document, self.portal.REQUEST),
+                                  ICheckinCheckoutManager)
+        manager.checkout()
+        self.assertFalse(IRefreshableLockable(self.document).locked())
+        IRefreshableLockable(self.document).lock()
+        self.assertTrue(IRefreshableLockable(self.document).locked())
+
+        self.login(self.administrator)
+        manager.checkin(comment="Force checkin")
+        self.assertFalse(IRefreshableLockable(self.document).locked())
+
+
 class TestCheckin(FunctionalTestCase):
     """Tests for the checkin functionality."""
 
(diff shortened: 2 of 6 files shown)


njohner:
@Rotonen @deiferni
This replaces https://github.com/4teamwork/opengever.core/pull/3919

opengever/document/checkout/checkin.py
```diff
@@ -59,7 +59,12 @@ def perform_checkin(self, document, comment):
                 mapping=dict(title=document.Title().decode('utf-8')),
                 )
             IStatusMessage(self.request).addStatusMessage(msg, type=u'error')
-
+        if not comment and not manager.is_checkin_without_comment_allowed():
```
deiferni: i would not require a comment here. the motivation to disable the checkin without comment is not to require a comment but a quick solution to be able to display a warning message to the user. We wanted to indicate that he may cause trouble office-connector side for some other user. So in this case the form is used to display the warning message and thus also to provides slightly different button labels (Trotzdem einchecken).

opengever/document/tests/test_checkout.py
```diff
@@ -235,6 +250,18 @@ def test_file_upload_is_allowed_when_document_is_checked_out_and_not_locked(self
         doc = create(Builder('document').checked_out())
         self.assertTrue(self.get_manager(doc).is_file_upload_allowed())
 
+    def test_checkin_permissions_when_document_is_checked_out_by_other_user(self):
+        doc = create(Builder('document').checked_out_by('hugo.boss'))
+        self.assertFalse(self.get_manager(doc).is_checkin_allowed())
+        self.grant('Manager')
+        self.assertTrue(self.get_manager(doc).is_checkin_allowed())
+
+        self.portal.acl_users.userFolderAddUser('other_user', 'secret', ['WorkspaceMember'], [])
```
deiferni: I'd not use the workspace roles for tests for now as it is rather new and unfinished at the moment.

opengever/document/tests/test_checkout.py
```diff
@@ -82,6 +83,20 @@ def test_clear_locks(self):
         self.manager.checkin()
         self.assertFalse(IRefreshableLockable(self.document).locked())
 
+    def test_force_checkin(self):
```
deiferni: please also add a test that documents that an `Administrator` can always checkin (which should be the case according to `rolemap.xml`)

deiferni:
one more thing, could you double-check that this is documented in docs.onegovgever.ch? if not @livschal could update the docs accordingly.

njohner:
To avoid the situation where a user calling the `checkin_without_comment` directly does not see a warning message when the document he is trying to check-in is blocked or checked out by another user, we now redirect the `checkin_without_comment` action back to the `checkin_with_comment` form when necessary.
Note that neither of the two actions will display any warning when the multi-document checkin is used.

So the current situation is:

**In the document preview we indicate when a document is locked and now show the "Checkin with comment" button as well as the action in the menu**
screen shot 2018-03-16 at 08 56 03

**If a user directly calls the checkin without comment on a locked document, he is redirected to the checkin with comment form, displaying the appropriate warning**
screen shot 2018-03-16 at 08 56 32

**When multi document checkin is used, no warning is displayed**
screen shot 2018-03-16 at 09 02 53

Rotonen:
@njohner what happens when Office Connector tries to do a no-comment checkin in these circumstances?

njohner:
As for documentation, it does not talk at all about locked objects. Also it will be outdated as the button for `checkin_with_comment`will be available for managers, not only the menu action.

njohner:
@Rotonen 
If the situation you are referring to is:
* An administrator (Admin1) has a document open in office connector
* Another administrator (Admin2) force checks in that document
* Admin1 closes the document, so office connector tries to check it in (but the ).
So what seems to happen in that situation is:
1. If Admin1 did not make any changes to the document, he gets the following message:
screen shot 2018-03-16 at 14 59 08
2. If he did make changes, he gets the message that there is a conflict, and if he still wants to checkin, it fails:
screen shot 2018-03-16 at 15 03 59
screen shot 2018-03-16 at 15 05 48
3. And brings us to that state we couldn't reproduce where the document is locked and no one can check it in (it does not show but I'm logged in as zopemaster here):

screen shot 2018-03-16 at 15 14 21

Now that was with office connector 1.3. With 1.6, this does not happen anymore. In case of conflict it asks to first checkout the document again, so everything works fine in that case:

screen shot 2018-03-16 at 15 34 56

Rotonen:
Thank you for the validation with 1.6. That is reassuring.

@deiferni this should also go into the manual validation guide.

In general what happens when there is a redirect with a `200 OK` in the end? I suppose that's fine as then the file has gotten succesfully uploaded if OC gets to that point and simply silently fails to check in (as in indirectly behaves as it would have with a ZEM file).

PR #4056, opened 2018-03-08
Always manually sync proposals for dossier reference number prefixes

### LINKED ISSUE #3234: Wrong dossier-reference-number on proposal after moving dossier.
After moving a dossier, the `dossier_reference_number` cached in the proposal's SQL is not correct: the dossier part is missing.
This issue might appear when subscribers are in wrong order.
The problem was discovered in test, a production problem was not verified at this point.

https://github.com/4teamwork/opengever.core/blob/cb75d8d399dfe043e448960a206ba76fe59a95e0/opengever/meeting/tests/test_proposal.py#L276-L278

/cc @deiferni

Rotonen:
Seems there was already something similar in place for tasks.

Closes #3234

### DIFF
--- a/opengever/dossier/handlers.py
+++ b/opengever/dossier/handlers.py
@@ -8,6 +8,7 @@ from opengever.dossier.behaviors.dossier import IDossierMarker
 from opengever.dossier.resolve import get_resolver
 from opengever.globalindex.handlers.task import sync_task
 from opengever.globalindex.handlers.task import TaskSqlSyncer
+from opengever.meeting.handlers import ProposalSqlSyncer
 from plone import api
 from plone.app.workflow.interfaces import ILocalrolesModifiedEvent
 from zope.component import getAdapter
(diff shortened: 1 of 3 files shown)


opengever/dossier/handlers.py
```diff
@@ -76,6 +77,14 @@ def save_reference_number_prefix(obj, event):
     for task in tasks:
         TaskSqlSyncer(task.getObject(), None).sync()
 
+    # And also proposals
+    proposals = catalog({
+        'path': '/'.join(obj.getPhysicalPath()),
+        'object_provides': 'opengever.meeting.proposal.IProposal',
+        'depth': -1})
+    for proposal in proposals:
```
deiferni: I wonder why https://github.com/4teamwork/opengever.core/blob/a37a608d7247d561acc34b94b20c3111c8edc557/opengever/meeting/handlers.py#L74 does not apply here 🤔. is this an issue with the order in which events/operations are performed for nested objects?

Rotonen: https://github.com/4teamwork/opengever.core/issues/3234#issuecomment-359909667

But as to the why, the order of events is undefined. I'd not try to seek an answer there.

Issue #3941, opened 2018-02-06
OC UX for reauth requests


Rotonen:
This will draw from the work and approaches initiated in the officeatwork branches.

Rotonen:
Scaling this down from user-facing modal dialogs to just nixing a redirector warning. It is likelier we'll tackle this stuff with any upcoming UI overhauls.

PR #4061, opened 2018-03-12
Avoid checking out the document via the redirector on an OC reauth.

Rotonen:
Added a `reauth=1` querystrhing parametre to the checkout redirector to skip the checkout when used only for the reauth.

Closes https://github.com/4teamwork/opengever.core/issues/3941


deiferni:
(hijacked and rebased)

PR #4062, opened 2018-03-12
Reorder buildouts to minimize CI test run buildout time

Rotonen:
This time without destroying `development.cfg`. A basically free ~40s CI job runtime gain.


Rotonen:
Potential yes, but as to verifiability of what I'd break by doing so, no idea.

PR #4064, opened 2018-03-13
 Docs: Add documentation for ftw.tokenauth usage.

lukasgraf:
Docs: Add documentation for `ftw.tokenauth` usage.

A rendered version is available at https://docs.onegovgever.ch/tokenauth-docs/dev-manual/api/authentication/oauth2_token_auth/ for easier review.


docs/public/dev-manual/api/authentication/oauth2_token_auth.rst
```diff
@@ -0,0 +1,345 @@
...
+Nach dem Erzeugen wird der private Teil des Schlüssels genau einmal
+dargestellt, und muss gespeichert werden. Der öffentliche Schlüssel verbleibt
+auf dem Server, und der private Schlüssel wird auf dem Dateisystem so
+hinterlegt, dass er nur für die Service-Applikation zugänglich ist.
+
+Bei der Erstellung eines Schlüssels wird diesem eine eindeutige Client-ID
+vergeben. Diese Client-ID identifiziert die Service-Applikation, und ein
+Schlüssel sollte daher jeweils nur für eine Applikation verwendet werden.
+
+|tokenauth-download-key|
```
deiferni: Im screenshot hab ich einen typo erspäht: Diese**r** Schlüssel ermöglicht jedem …

docs/public/dev-manual/api/authentication/oauth2_token_auth.rst
```diff
@@ -0,0 +1,345 @@
...
+        response = session.get('http://localhost:8080/Plone/')
+        # ...
+
+Wenn das von der Client-Applikation gesendete Token abgelaufen ist, sendet
+der Server eine entsprechende Fehler-Response:
+
+.. code:: json
+
+    {
+      "error": "invalid_token",
```
deiferni: just a question to clarify, not relevant for this PR: is this error only used for expired tokens? theoretically the token could be invalid for other reasons. i think it is important that the client application can differ easily between an expired and an otherwise invalid token. Thus i would have expected an explicit `expired_token` error id for such cases. otherwise the client app may have to inspect the `error_description` in addition to the `error` to obtain the necessary information. it may be a bit theoretical but maybe worth a though?

docs/public/dev-manual/api/authentication/examples/oauth2-client-example.py
```diff
@@ -0,0 +1,84 @@
+import json
```
deiferni: thinking out loud: could we run this in a test 🤔

docs/public/dev-manual/api/authentication/oauth2_token_auth.rst
lukasgraf: Yes, this is a good point: There may in fact be other `invalid_token` responses, and the client absolutely needs to check for the `error_description` (even though this is intended to be a human readable field).

(This response gets produced by `ftw.tokenauth` [here](https://github.com/4teamwork/ftw.tokenauth/blob/master/ftw/tokenauth/pas/plugin.py#L217-L227)).

The [OAuth2 spec on token error responses](https://tools.ietf.org/html/rfc6750#section-3.1) leaves something to be desired when it comes to this aspect.
For one, it's very clear that it's the `error` type `invalid_token` that needs to be sent for expired tokens. So something like `expired_token`, which would be much easier for machines to check for would go against the spec.

On the other hand, it designates `error_description` as a human readable field.
I'll try and update the documentation to make this a bit more clear.

docs/public/dev-manual/api/authentication/examples/oauth2-client-example.py
lukasgraf: There is a [high-level end-to-end test in `ftw.tokenauth`](https://github.com/4teamwork/ftw.tokenauth/blob/master/ftw/tokenauth/tests/test_end_to_end.py) that pretty much does this (minus sessions (peristent HTTP connections)).

But this could indeed be a case where doctests in GEVER could make sense (the way they were originally intended, test that examples in your documentation are valid).

I might look into this at some point, but might require some trickery with executable test setup vs. readability of example code vs. embedding it in Sphinx docs.

docs/public/dev-manual/api/authentication/oauth2_token_auth.rst
lukasgraf: sad_panda

PR #4077, opened 2018-03-19
Update policy template and example content

njohner:
* Update the policy template to include the new features
* Add French translations for the `Ordnungssystem` (position 0) and German everywhere else, to avoid having `null` everywhere when looking at the example content in French.
This is for https://github.com/4teamwork/opengever.demo/issues/48


njohner:
@phgross 
Ok done

njohner:
Here is a screenshot of the `Ordnungssystem` in French, with the German added where no French translation is currently available.
screen shot 2018-03-16 at 10 21 06

PR #4080, opened 2018-03-19
Fix force checkin button

njohner:
Buttons get defined on the class level, so conditionally changing the button label is not a good idea. Instead we now have 2 buttons that are displayed conditionally.
I also added the French translation of the button label.

**Before, once the label was changed to "checking anyway", it could remain in that state even if the document is not locked anymore**
screen shot 2018-03-19 at 14 18 23

**Now  we get the correct labels**
screen shot 2018-03-19 at 13 56 20
screen shot 2018-03-19 at 13 57 16
**And the french translation**
screen shot 2018-03-19 at 14 02 09

resolves https://extranet.4teamwork.ch/support/edk/tracker-support-gever-edk/102

### DIFF
--- a/opengever/document/checkout/checkin.py
+++ b/opengever/document/checkout/checkin.py
@@ -102,8 +102,7 @@ class CheckinContextCommentForm(form.Form):
 
     def __init__(self, context, request):
         super(CheckinContextCommentForm, self).__init__(context, request)
-        self.locked = IRefreshableLockable(self.context).locked()
-        if self.locked:
+        if self.is_document_locked():
             # Add a warning onto a locked document checkin view
             msg = _(
                 u'label_warn_checkout_locked',
@@ -115,23 +114,12 @@ class CheckinContextCommentForm(form.Form):
                 )
             IStatusMessage(self.request).addStatusMessage(msg, type=u'warning')
 
-            # Swap the button label out on a locked document
-            for form_button in self.buttons.items():
-                if form_button[0] == 'button_checkin':
-                    button_payload = form_button[1]
-                    button_payload.title = _(
-                        u'button_checkin_anyway',
-                        default=u'Checkin anyway'
-                        )
-                    form_button = (
-                        'button_checkin_anyway',
-                        button_payload,
-                        )
-
         self.checkin_controller = CheckinController(self.request)
 
-    @button.buttonAndHandler(_(u'button_checkin', default=u'Checkin'))
-    def checkin_button_handler(self, action):
+    def is_document_locked(self):
+        return IRefreshableLockable(self.context).locked()
+
+    def checkin_document(self):
         # Errors are handled by the checkin
         data = self.extractData()[0]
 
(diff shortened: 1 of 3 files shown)


jone:
Could you add a screenshot? This could help giving some context 😉

njohner:
I had already made them, but forgot to add them to the PR

njohner:
Nooo it failed. I guess I'll need to fix the tests :-)

opengever/document/checkout/checkin.py
```diff
@@ -115,23 +114,27 @@ def __init__(self, context, request):
...
-                    form_button = (
-                        'button_checkin_anyway',
-                        button_payload,
-                        )
-
         self.checkin_controller = CheckinController(self.request)
 
-    @button.buttonAndHandler(_(u'button_checkin', default=u'Checkin'))
-    def checkin_button_handler(self, action):
+    def document_is_locked(self):
```
phgross: I would rename it to `is_document_locked`

opengever/document/checkout/checkin.py
```diff
@@ -115,23 +114,27 @@ def __init__(self, context, request):
...
+        return IRefreshableLockable(self.context).locked()
+
+    @button.buttonAndHandler(_(u'button_checkin', default=u'Checkin'),
+                             condition=lambda form: not form.document_is_locked())
+    def checkin_button_handler(self, action, condition=lambda form: not form.document_is_locked()):
```
phgross: Is it necessary to repeat the condition in the method itself? According to the [docs](https://docs.plone.org/develop/addons/schema-driven-forms/customising-form-behaviour/actions.html#conditional-actions) it's not.

njohner: Oh no, sorry. That is a "typo", or rather a copy paste fail.

opengever/document/checkout/checkin.py
```diff
@@ -115,23 +114,27 @@ def __init__(self, context, request):
...
+        # Errors are handled by the checkin
+        data = self.extractData()[0]
```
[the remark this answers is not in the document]
njohner: done

njohner:
* @phgross I hadn't done that translation because I was having some issues translating the last sentence. I added one now, but I feel it's not perfect. As I wrote in the gever channel, I think we need a real translation of `checkout` in French.
* I added a test to check that the proper buttons are displayed
* Corrected another translation (the warning said the document was "locked" instead of saying that it is checked-out)
* In the French warning "Avec check-out" means "with checkout", which is not was is meant here ("checked out"). I am unable to come up with a translation as "checkout" cannot be used as a verb in French. If we use something else, we would need to change it everywhere in gever. (I think "retirer" would be a good option if we want to go down that path)

**French view of force checkin**
screen shot 2018-03-19 at 15 34 09

opengever/document/tests/test_checkout.py
```diff
@@ -335,6 +336,31 @@ def get_manager(self, document):
...
+        browser.open(self.document)
+        browser.css('#checkin_with_comment').first.click()
+        with self.assertRaises(NoElementFound):
+            browser.css('#form-buttons-button_checkin_anyway').first
+
+        self.assertEqual(browser.css('#form-buttons-button_checkin').first.name, 'form.buttons.button_checkin')
+
+        # Lock document
+        IRefreshableLockable(self.document).lock()
+        transaction.commit()
```
deiferni: you should never have to commit in  an `IntegrationTestCase`

njohner:
@deiferni Ok everything done

PR #4082, opened 2018-03-20
Upgrade to cleanup leftovers from ICreatorAware

### LINKED ISSUE #4048: Reindex object_provides, remove references to ICreatorAware
`ICreatorAware` and `ICreator` have been removed  in https://github.com/4teamwork/opengever.core/pull/2949 and upgraded in https://github.com/4teamwork/opengever.core/pull/3165. However the catalog was not reindexed properly. Thus some objects still provide `ICreatorAware` according tot he catalog. 

We should add an upgrade step that reindexes the `object_provides` index for those objects. The misbehaving brains can be retrieved as follows:

```python
catalog(object_provides='opengever.base.behaviors.creator.ICreatorAware')
```

njohner:
Upgrade step to remove references to the `ICreatorAware` interface in the index.
This was tested locally by reverting the commits removing the interface, setting up a new gever with content and then reapplying the commits and corresponding upgrade steps. This upgrade step could then be applied and tested.

resolves #4048

### DIFF
--- a/opengever/core/upgrades/20180319170439_cleanup_i_creator_leftovers/upgrade.py
+++ b/opengever/core/upgrades/20180319170439_cleanup_i_creator_leftovers/upgrade.py
@@ -1,0 +1,19 @@
+from ftw.upgrade import UpgradeStep
+from plone import api
+
+
+class CleanupICreatorLeftovers(UpgradeStep):
+    """Cleanup i creator leftovers.
+    """
+
+    def __call__(self):
+        self.reindex_object_provides()
+
+    def reindex_object_provides(self):
+        catalog = api.portal.get().portal_catalog
+        query = {'object_provides': ['opengever.base.behaviors.creator.ICreatorAware']}
+        message = "Remove references to ICreatorAware interface:"
+        message += " reindexing object_provides"
+        for obj in self.objects(query, message):
+            catalog.reindexObject(
+                obj, idxs=['object_provides'], update_metadata=False)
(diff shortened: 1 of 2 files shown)


opengever/core/upgrades/20180319170439_cleanup_i_creator_leftovers/upgrade.py
```diff
@@ -0,0 +1,16 @@
...
+
+    def __call__(self):
+        self.reindex_object_provides()
+
+    def reindex_object_provides(self):
+        query = {'object_provides': ['opengever.base.behaviors.creator.ICreatorAware']}
+        message = "Remove references to ICreatorAware interface:"
+        message += " reindexing object_provides"
+        for obj in self.objects(query, message):
+            obj.reindexObject(idxs=["object_provides"])
```
deiferni: `object_provides` is index data only, there is no metadata. so you should make sure that metadata is not updated for performance reasons:
```python
    catalog.reindexObject(obj, idxs=['object_provides'], update_metadata=False)
```
Unfortunately the `update_metadata` argument is only available on `portal_catalog` but not on on each content object.

deiferni: yes it does, but unfortunately for the indexes only. metadata is always updated either completely or not at all.

njohner:
Sorry forgot the readme. Will do it now

njohner:
@deiferni I guess you meant an entry in the Changelog?
Everything is done.

PR #4089, opened 2018-03-22
Only allow to add contacts when `IContactSettings.is_feature_enabled` is disabled

### LINKED ISSUE #4086: Hinzufügen, Kontakt auch mit `IContactSettings.is_feature_enabled` sichtbar
The links will add an `opengever.contact.contact` content type object. this is clearly wrong as with `IContactSettings.is_feature_enabled` a different type of contact is used. also we don't offer a create/edit form for these types.

screen shot 2018-03-21 at 13 46 42

- the link should only be visible when the feature mentioned above is disabled
- bonus points: the add form should not be available either

njohner:
`IContactSettings.is_feature_enabled` is enabled, one should not be able to add contacts.
* implement `allowedContentTypes` to only allow `contact` when the feature is disabled
* Make the add form unavailable when the feature is enabled.

**With the feature enabled, contact is not in the add menu anymore**
screen shot 2018-03-22 at 10 09 13

**Unauthorized is raised when directly accessing the add form**
screen shot 2018-03-22 at 10 09 53

**When the feature is disabled, contacts can be added**
screen shot 2018-03-22 at 11 10 02

resolves #4086

### DIFF
--- a/opengever/contact/browser/contact_forms.py
+++ b/opengever/contact/browser/contact_forms.py
@@ -1,0 +1,23 @@
+from plone.dexterity.browser.add import DefaultAddForm
+from plone.dexterity.browser.add import DefaultAddView
+from plone.dexterity.interfaces import IDexterityFTI
+from Products.CMFCore.interfaces import IFolderish
+from zExceptions import Unauthorized
+from zope.component import adapter
+from zope.component import getUtility
+from zope.publisher.interfaces.browser import IDefaultBrowserLayer
+
+
+class ContactAddForm(DefaultAddForm):
+
+    def render(self):
+        fti = getUtility(IDexterityFTI, name=self.portal_type)
+        if fti not in self.context.allowedContentTypes():
+            raise Unauthorized
+
+        return super(ContactAddForm, self).render()
+
+
+@adapter(IFolderish, IDefaultBrowserLayer, IDexterityFTI)
+class ContactAddView(DefaultAddView):
+    form = ContactAddForm
--- a/opengever/contact/contactfolder.py
+++ b/opengever/contact/contactfolder.py
@@ -1,4 +1,5 @@
 from opengever.base.behaviors.translated_title import TranslatedTitleMixin
+from opengever.contact import is_contact_feature_enabled
 from opengever.contact.interfaces import IContactFolder
 from opengever.contact.models import Contact
 from opengever.ogds.base.wrapper import TeamWrapper
@@ -43,3 +44,13 @@ class ContactFolder(Container, TranslatedTitleMixin):
         if default is _marker:
             raise KeyError(id_)
         return default
+
+    def allowedContentTypes(self, *args, **kwargs):
+        types = super(ContactFolder, self).allowedContentTypes(*args, **kwargs)
+
+        def filter_type(fti):
+            if fti.id == "opengever.contact.contact":
+                return not is_contact_feature_enabled()
+            return True
+
+        return filter(filter_type, types)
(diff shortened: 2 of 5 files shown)


opengever/contact/contactfolder.py
```diff
@@ -43,3 +44,14 @@ def _getOb(self, id_, default=_marker):
         if default is _marker:
             raise KeyError(id_)
         return default
+
+    def allowedContentTypes(self, *args, **kwargs):
+        #types = super(ContactFolder, self).allowedContentTypes(*args, **kwargs)
+        types = Container.allowedContentTypes(self, *args, **kwargs)
```
njohner: Can someone explain why super doesn't work here? I get the error that self should be an instance of `ContactFolder`
types = super(ContactFolder, self).allowedContentTypes(*args, **kwargs)

opengever/contact/contactfolder.py
```diff
@@ -43,3 +44,13 @@ def _getOb(self, id_, default=_marker):
         if default is _marker:
             raise KeyError(id_)
         return default
+
+    def allowedContentTypes(self, *args, **kwargs):
+        types = Container.allowedContentTypes(self, *args, **kwargs)
```
njohner: Can someone explain why super doesn't work here? I get the error that self should be an instance of `ContactFolder`
types = super(ContactFolder, self).allowedContentTypes(*args, **kwargs)

deiferni: Did you use `plone.reload` to reload code? Reloading will re-interpret the file with python thus creating a new instance of the class. It cannot replace the class everywhere it is used though so such checks will fail. Restarting the associated python process solves the issue.

njohner: Oh thanks, I guess that was it. Seems to work now...

opengever/contact/tests/test_contact.py
```diff
@@ -31,6 +34,24 @@ def test_add_a_contact(self, browser):
...
+
+        factoriesmenu.add('Contact')
+        browser.visit(contactfolder, view="++add++opengever.contact.contact")
+
+        api.portal.set_registry_record(
+          'is_feature_enabled', True, interface=IContactSettings)
+        transaction.commit()
+
+        browser.open(contactfolder)
+        with self.assertRaises(ValueError):
```
deiferni: there are multiple reasons why a `ValueError` could be raised here as it is a very generic exception. this could potentially shadow what you want to test for here.
I'd prefer an explicit assertion that `Contact` is not in the factories menu, you should be able to do so with the `addable_types` function.

opengever/contact/browser/contact_forms.py
```diff
@@ -0,0 +1,23 @@
...
+from plone.dexterity.browser.add import DefaultAddView
+from plone.dexterity.interfaces import IDexterityFTI
+from Products.CMFCore.interfaces import IFolderish
+from zExceptions import Unauthorized
+from zope.component import adapter
+from zope.component import getUtility
+from zope.publisher.interfaces.browser import IDefaultBrowserLayer
+
+
+class AddForm(DefaultAddForm):
```
deiferni: I'd include the type name here: `ContactAddForm` and `ContactAddView`, that way the class can be easily found later. We already have too many badly named `AddForm` classes 😞.

njohner: Ha, I had started with `ContactAddForm` and then saw that for ContactFolder, `AddForm` was used :-D

njohner:
@deiferni 
Ok done

PR #4094, opened 2018-03-26
Disallow checkin-anyway for multi-checkin action

### LINKED ISSUE #4091: Check-in-anyway can be circumvented via table actions
The "check in anyway" safeguard can be accidentaly circumvented by choosing one of the `"Mit Kommentar einchecken"` or `"Ohne Kommentar einchecken"` actions on a table in the documents tab (for checkin in documents *en masse*).

**Steps to reproduce**

- Check out several documents using OfficeConnector
- Keep them open on the client side
- Select those documents in the `Documents` tab and choose the `"Ohne Kommentar einchecken"` action

mass_check_in

A possible solution should deal with the following apect:
- If the user is presented with a "check in anyway" option, it should be clear to him, which documents this applies to (only a subset of the selected documents may be in the "still open on client side" state). IMHO this is a piece of information the user needs to make an educated decision.

njohner:
* Do not check-in locked documents when the multi-checkin action is called. 
  * Display a message for each document that could not be checked-in
  * Addition of a `force` argument to methods of the checkin controller to make the use of force checkin more deliberate and clear.

Apart from the main bug fixed, another one was also fixed and tests improved:
* Correct a bug for checkin without comment. When doing a checkin without comment on a document that could not be checked-in, one would be redirected to the checkin with comment. This was intended for locked documents, but would also happen for documents that could not be checked in for other reasons (for example not checked out)
* Improved existing multi-checkin tests by better using the fixture (remove unnecessary document creation)

**Checkin from table action on two locked documents**
screen shot 2018-03-26 at 10 02 24

resolves #4091

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,9 @@ Changelog
 ---------------------
 
 - Update object security when objects are moved (account for changes in placeful workflow). [lgraf]
+- Disallow force-checkin for mutli-checkin action (table action) [njohner]
+- Make sure to warn users before they can force-checkin a document [njohner]
+- Add checkin button for force-checkin in the document actions [njohner]
 - Only allow to add contacts when IContactSettings.is_feature_enabled is disabled [njohner]
 - Cleanup leftovers from ICreatorAware interface in index [njohner]
 - Update policy template to set flags for new features [njohner]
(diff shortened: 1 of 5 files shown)


njohner:
@lukasgraf 
I realized the changelog was also missing for the other PR on force-checkin for single documents. I added changelogs for both here.

njohner:
sorry, I just rebased...

PR #4115, opened 2018-03-28
Fix displaying quickupload uploadbox for template folders.

### LINKED ISSUE #4114: D&D upload for template folder broken.
Can't upload stuff any more. Should work. Will fix.

deiferni:
This PR fixes an issue where the upload box would not be rendered for template folders. 

The template folder did no longer provide the `ITabbedviewUploadable` interface, which is required (also for containers that is uploaded into).

In https://github.com/4teamwork/opengever.core/pull/3081 the inheritance of template folders has changed. I suspect that this removed the upload functionality.

Fixes #4114.

### DIFF
--- a/opengever/dossier/templatefolder/interfaces.py
+++ b/opengever/dossier/templatefolder/interfaces.py
@@ -1,5 +1,5 @@
-from zope.interface import Interface
+from ftw.tabbedview.interfaces import ITabbedviewUploadable
 
 
-class ITemplateFolder(Interface):
+class ITemplateFolder(ITabbedviewUploadable):
     pass
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2018.2.0 (unreleased)
 ---------------------
 
+- Fix displaying quickupload uploadbox for template folders. [deiferni]
 - Homogenize eCH-0147 import/export action titles. [lgraf]
 - Update object security when objects are moved (account for changes in placeful workflow). [lgraf]
 - Disallow force-checkin for mutli-checkin action (table action) [njohner]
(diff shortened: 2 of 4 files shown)


opengever/dossier/templatefolder/interfaces.py
```diff
@@ -1,5 +1,5 @@
-from zope.interface import Interface
+from ftw.tabbedview.interfaces import ITabbedviewUploadable
 
 
-class ITemplateFolder(Interface):
+class ITemplateFolder(ITabbedviewUploadable):
```
njohner: you could maybe inherit from Interface first:
```
class ITemplateFolder(Interface, ITabbedviewUploadable)
```
Won't change anything right now but provides a safer MRO just in case... It also seems to be how it is done with other interfaces in opengever, e.g. https://github.com/4teamwork/opengever.core/blob/de0a0d6d0959f94055b2c30d2218b0f19582eb51/opengever/dossier/behaviors/dossier.py#L27

deiferni: `ITabbedviewUploadable` directly inherits from `Interface` already. `Interface` is the one common base that all `Interface`s will inherit from, i would not expect that to change 😄 .

So inheriting from `Interface` and some other interface that already inherits from `Interface` at the same time does not make much sense in my opinion. Unless there is some peculiarity in the registry i don't know about 🤔.
It may make sense to inherit from two more specific interfaces though, of course.

njohner: It changes the order of resolution of methods. Here it would first look up in `ITemplateFolder`, then in `ITabbedviewUploadable` and then in `Interface`. If you directly inherit from interface as first parent, then the order is `ITemplateFolder`, `Interface` and then `ITabbedviewUploadable`.
But as interfaces are mainly used as markers, so don't provide much methods, it won't change anything. It actually seems that in most cases in opengever they interfaces directly inherit from another interface or from `Interface` but rarely from both.
So anyway, leave it as is :-D

PR #4120, opened 2018-04-03
Enhance resolve_oguid view, to support other views.

### LINKED ISSUE #4101: OGIP 32: enhance resolve_oguid to support other views
In OGIP 32 we are not storing full URLs, but we want still be able to have document tooltips.
As document URL we'll use something like `/fd/resolve_oguid?fd:123`. This works for the default view of an object, but not for other views.

For the tooltip we want to use create an URL to the `/tooltip`-view of the document.
In order to do that we need to also support traversal-based IDs instead of GET-parameter based, since we cannot combine GET-parameter with additional path parts.

In the end we want to support these kinds of URLs (with their redirects):
- `/fd/resolve_oguid?fd:123` -> `/fd/os/1/dossier-5/document-7` (already supported)
- `/fd/resolve_oguid/fd:123` -> `/fd/os/1/dossier-5/document-7` (new)
- `/fd/resolve_oguid/fd:123/tooltip` -> `/fd/os/1/dossier-5/document-7/tooltip` (new)

phgross:
The view supports now traversal based oguids and other views.

GET paramater: `/fd/resolve_oguid?fd:123`
Traversal based paramater: `/fd/resolve_oguid/fd:123`
Specific view: `/fd/resolve_oguid/fd:123/tooltip`

Closes #4101

### DIFF
--- a/opengever/base/browser/resolveoguid.py
+++ b/opengever/base/browser/resolveoguid.py
@@ -6,15 +6,39 @@ from Products.CMFCore.utils import getToolByName
 from Products.Five import BrowserView
 from zExceptions import Unauthorized
 from zope.component import getUtility
+from zope.interface import implementer
 from zope.intid.interfaces import IIntIds
+from zope.publisher.interfaces import IPublishTraverse
 
 
+@implementer(IPublishTraverse)
 class ResolveOGUIDView(BrowserView):
+    """The view supports the following url formats:
+
+    GET paramater: /fd/resolve_oguid?oguid=fd:123
+    Traversal based paramater: /fd/resolve_oguid/fd:123
+    Specific view: /fd/resolve_oguid/fd:123/tooltip
+    """
+
+    oguid_str = None
+    view_name = None
+
+    def publishTraverse(self, request, name):
+        if self.oguid_str:
+            self.view_name = name
+        else:
+            self.oguid_str = name
+
+        return self
 
     @classmethod
-    def url_for(cls, oguid, admin_unit=None):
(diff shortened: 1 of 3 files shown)


opengever/base/browser/resolveoguid.py
```diff
@@ -6,32 +6,63 @@
...
+from zope.interface import implementer
 from zope.intid.interfaces import IIntIds
+from zope.publisher.interfaces import IPublishTraverse
 
 
+@implementer(IPublishTraverse)
 class ResolveOGUIDView(BrowserView):
+    """The view supports the following url formats:
+
+    GET paramater: /fd/resolve_oguid?fd:123
```
lukasgraf: This should be `?oguid=fd:123`, no?

opengever/base/tests/test_resolve_oguid.py
```diff
@@ -28,3 +28,37 @@ def test_redirect_if_correct_client(self, browser):
...
+
+        oguid = Oguid.for_object(self.task)
+        url = "{}/@@resolve_oguid?oguid={}".format(
+            self.portal.absolute_url(), oguid)
+
+        browser.open(url)
+        self.assertEqual(self.task.absolute_url(), browser.url)
+
+    @browsing
+    def test_with_traversable_based_ids(self, browser):
```
lukasgraf: `test_with_traversable_ids` or `test_with_traversal_based_ids` 😉

phgross:
@lukasgraf updated ...

PR #4128, opened 2018-04-04
Add favorite SQL model.

### LINKED ISSUE #4100: OGIP 32: favorites storage
The new storage for favorites is implemented as an SQL table in order to support inter-client favorites.

**Tasks:**
- [x] Create model
- [x] Add migration creating the SQL table
- [x] Implement method for looking up whether an object is marked as favorite for the current user (will be needed in #4103)

**Model:**
Package: `opengever.base`
Table: `favorites`
Columns:
- `id` (PK)
- `admin_unit_id` (index)
- `int_id` (index)
- `user_id` (index)
- `position` (index)
- `title`
- `is_title_personalized` (Default: `False`)
- `portal_type`
- `icon_class`
- `uid`
- `created` (time)
- `modified` (time)

phgross:
This PR adds a new table `favorites` to the SQL DB , as it is specified in #4100, including an schema migration upgradestep.

Differences to specification:
- `is_title_personalized` instead of `personalized_title`
- tooltip_view not stored in the DB
- `created` instead of `added`

Closes #4100


opengever/base/model/favorite.py
```diff
@@ -0,0 +1,60 @@
...
+
+    userid = Column(String(USER_ID_LENGTH), index=True)
+    position = Column(Integer)
+
+    title = Column(String(CONTENT_TITLE_LENGTH), nullable=False)
+    title_personalized = Column(Boolean, default=False, nullable=False)
+    portal_type = Column(String(PORTAL_TYPE_LENGTH))
+    icon_class = Column(String(ID_LENGTH))
+
+    tooltip_view = Column(String(ID_LENGTH))
```
deiferni: out of curiosity, what is this for 🤔?

opengever/base/model/favorite.py
```diff
@@ -0,0 +1,60 @@
...
+    uid = Column(String(UID_LENGTH))
+    created = Column(UTCDateTime(timezone=True), default=utcnow_tz_aware)
+    modified = Column(UTCDateTime(timezone=True),
+                      default=utcnow_tz_aware,
+                      onupdate=utcnow_tz_aware)
+
+
+from opengever.ogds.models.query import BaseQuery
```
deiferni: this import does not belong here 😉.

opengever/base/model/favorite.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class FavoriteQuery(BaseQuery):
+
+    def is_marked_as_favorite(self, obj, user):
+        query = self.by_object_and_user(obj, user)
+        return self.session.query(query.exists()).scalar()
```
deiferni: i vaguely remember there being a problem with exists and the advice to rather use a check that `.first()`  is not `None`. maybe someone may fill in the blanks?

opengever/base/model/favorite.py
```diff
@@ -0,0 +1,60 @@
...
+
+    favorite_id = Column('id', Integer, Sequence("favorites_id_seq"),
+                primary_key=True)
+
+    admin_unit_id = Column(String(UNIT_ID_LENGTH), index=True, nullable=False)
+    int_id = Column(Integer, index=True, nullable=False)
+    oguid = composite(Oguid, admin_unit_id, int_id)
+
+    userid = Column(String(USER_ID_LENGTH), index=True)
+    position = Column(Integer)
```
deiferni: this looks like something that we sort by? in that case it would deserve an `index`.

opengever/core/upgrades/20180403142317_add_favorites_table/upgrade.py
```diff
@@ -0,0 +1,43 @@
...
+    def migrate(self):
+        self.op.create_table(
+            'favorites',
+            Column('id', Integer, Sequence("favorites_id_seq"),
+                   primary_key=True),
+            Column('admin_unit_id', String(UNIT_ID_LENGTH),
+                   index=True, nullable=False),
+            Column('int_id', Integer, index=True, nullable=False),
+            Column('userid', String(USER_ID_LENGTH), index=True),
+            Column('position', Integer),
```
deiferni: (and also here, `index`)

opengever/core/upgrades/20180403142317_add_favorites_table/upgrade.py
```diff
@@ -0,0 +1,43 @@
...
+            Column('title', String(CONTENT_TITLE_LENGTH), nullable=False),
+            Column('title_personalized', Boolean, default=False, nullable=False),
```
deiferni: `is_title_personalized` or `has_personalized_title` would read better. here when i only see the attribute name i'd expect a string that contains the text of the personalized title

opengever/base/model/favorite.py
Rotonen: And could we use the `ENUM` type for something like that for operational safety anyway?

phgross: The tooltip_view column is not necessary anymore ...

phgross:
@4teamwork/gever updated ...

phgross:
@lukasgraf updated ...

PR #4132, opened 2018-04-06
Request confirmation before cancelling a document checkout

### LINKED ISSUE #4119: Bestätigung bei Checkout widerrufen
Der Button `Checkout widerrufen` bei einem ausgecheckten Dokument löscht die Working Copy und setzt die Datei auf die neuste Version zurück. Diese Funktionalität sollte vom Benutzer eine Bestätigung einfordern, weil sie destruktiv ist.

screen shot 2018-04-03 at 09 53 06

njohner:
Request a confirmation to cancel the checkout of a document.

When cancelling the checkout of a single document from the document actions, a popup window asks for confirmation. When using the table actions from the `TabbedView`, the confirmation is handled by a form (see snapshots below).

I also checked that it still works for `proposaltemplate`.

**pop up for cancelling checkout of a single document**
screen shot 2018-04-06 at 08 19 30

**form for cancelling checkout of multiple documents. The form lists the files that were selected for checkout cancel**
screen shot 2018-04-06 at 08 20 53

**Messages after cancelling checkout, here of three documents, with one of them that wasn't checked out, so the cancel failed for that one**
screen shot 2018-04-06 at 08 21 17

resolves #4119

### DIFF
--- a/opengever/core/upgrades/20180403175143_add_confirmation_to_cancel_checkout/jsregistry.xml
+++ b/opengever/core/upgrades/20180403175143_add_confirmation_to_cancel_checkout/jsregistry.xml
@@ -1,0 +1,16 @@
+<object name="portal_javascripts" meta_type="JavaScripts Registry">
+
+
+  <javascript
+      cacheable="True"
+      compression="safe"
+      cookable="True"
+      expression=""
+      enabled="True"
+      id="++resource++opengever.document/cancel_checkout_confirmation.js"
+      inline="False"
+      insert-after="++resource++opengever.document/download_confirmation.js"
+      />
+
+
+</object>
--- a/opengever/document/checkout/cancel.py
+++ b/opengever/document/checkout/cancel.py
@@ -1,10 +1,20 @@
 from ftw.mail.mail import IMail
 from opengever.document import _
 from opengever.document.document import IDocumentSchema
+from opengever.document.exceptions import NoItemsSelected
 from opengever.document.interfaces import ICheckinCheckoutManager
+from opengever.tabbedview.utils import get_containing_document_tab_url
 from plone import api
+from plone.z3cform import layout
 from Products.Five import BrowserView
+from Products.statusmessages.interfaces import IStatusMessage
+from z3c.form import button
+from z3c.form import field
+from z3c.form import form
+from z3c.form.interfaces import HIDDEN_MODE
+from zope import schema
 from zope.component import getMultiAdapter
+from zope.interface import Interface
 
 
 class CancelDocuments(BrowserView):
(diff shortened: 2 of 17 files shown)


Rotonen:
Just a comment, not a change request: one can get nicer looking modals by using the already bundled jquery-ui package. There is a PoC in the old and abandoned officeatwork branch:

https://github.com/4teamwork/opengever.core/commit/33f4a82d233c1c11ceae4f945c0ba61f19b83c01

opengever/core/upgrades/20180403175143_add_confirmation_to_cancel_checkout/jsregistry.xml
```diff
@@ -0,0 +1,26 @@
+<object name="portal_javascripts" meta_type="JavaScripts Registry">
+
+  <javascript
+      cacheable="True"
+      compression="safe"
+      cookable="True"
+      expression=""
+      enabled="True"
+      id="++resource++opengever.document/download_confirmation.js"
```
phgross: The `download_confirmation.js` is already registered ... ✂️ ?

opengever/document/locales/de/LC_MESSAGES/opengever.document.po
```diff
@@ -160,6 +168,12 @@ msgstr "Trotzdem einchecken"
 msgid "button_pdf"
 msgstr "PDF-Vorschau"
 
+#. Default: "Are you sure you want to cancel the checkout?"
+#: ./opengever/document/checkout/templates/cancelconfirmation.pt
+#: ./opengever/document/checkout/templates/cancelconfirmationform.pt
+msgid "cancel_checkout_confirmation"
+msgstr "Wollen Sie wirklich den Checkout widerrufen?"
```
phgross: `Wollen Sie den Checkout wirklich widerrufen? Bereits gemacht Änderungen am Dokument gehen dabei verloren.

opengever/document/checkout/cancel.py
```diff
@@ -85,3 +95,87 @@ def cancel(self, obj):
...
+    """Confirmation overlay view to cancel the checkout of a document.
+    This view is called directly on the document itself
+    (without any request parameters).
+    """
+
+    def get_checkout_cancel_url(self):
+        url = u'{}/@@cancel_document_checkouts'.format(
+            self.context.absolute_url())
+        paths = self.request.get('paths')
+        if not IDocumentSchema.providedBy(self.context) and paths:
```
phgross: I'm not sure why there is a need to handle `paths` here, the `CancelDocumentCheckoutConfirmation` is only used for a single document.

opengever/document/checkout/cancel.py
```diff
@@ -85,3 +95,87 @@ def cancel(self, obj):
...
+
+
+class IMultiCheckoutCancelSchema(Interface):
+    """Form schema to confirm multi checkout cancel."""
+
+    paths = schema.TextLine(title=u'Selected Items')
+
+
+class MultiCheckoutCancelForm(form.Form):
+    """Form to checkin one document, the form's context."""
```
phgross: Docstring is wrong.

opengever/document/checkout/cancel.py
```diff
@@ -85,3 +95,87 @@ def cancel(self, obj):
...
+    @button.buttonAndHandler(_(u'button_cancel', default=u'Cancel'))
+    def cancel(self, action):
+        return self.redirect()
+
+    def redirect(self):
+        return self.request.RESPONSE.redirect(
+            get_containing_document_tab_url(self.context))
+
+    def get_document_paths(self):
+        if self.widgets:
```
phgross: why are there two different ways to pass in the `paths`?

opengever/document/checkout/cancel.py
njohner: Wops. I first had a single view handling both cases, before making a form for the multi checkout. Apparently some leftovers.

opengever/document/checkout/cancel.py
njohner: The first time the form is called from the `TabbedView` the paths are sent in the request. 
Then, when the form is submitted, the paths were stored in the `paths` field of the form (see the `updateWidgets` method)

opengever/core/upgrades/20180403175143_add_confirmation_to_cancel_checkout/jsregistry.xml
njohner: Yes it is. I wasn't sure whether I needed to add it in the upgrade as I intercalated a new js file. I'll cut it out.

njohner:
For the translations, do we need the part about "loosing the changes already made to the documents"? Cancelling a checkout is not available for locked documents (there only force checkin is available), so there should be no data loss? Of course there could later on be conflicts...

Otherwise all done (I did the translations in a separate commit for now, if you want them like this I'll merge the two commits)

njohner:
@phgross All done

PR #4133, opened 2018-04-06
Favorites API endpoints

phgross:
Based on #4128 .
 
This PR adds the favorite API endpoints which are used for the new favorite-feature  (OGIP 32):

 - @favorites (GET, POST, PATCH and DELETE)
 - @repository-favorites (GET)

For further information see documentation https://github.com/4teamwork/opengever.core/blob/6636677a4dbe0b5b5ab09f737fae4ce97f2fe05b/docs/public/dev-manual/api/favorite.rst

**Permission and protection:**
 - the GET service requires`zope2.View` all other services requires the `setOwnProperties` permission.
 - All services are protected with an additional check `userid == api.user.get_current()`, which avoid that a user can manipulate favorites of an other user.

### DIFF
--- a/docs/public/dev-manual/api/favorite.rst
+++ b/docs/public/dev-manual/api/favorite.rst
@@ -1,0 +1,161 @@
+.. _favorites:
+
+Favoriten
+=========
+
+Der ``@favorites`` endpoint bietet alle Funktion für die Auflistung und Bearbeitung der globalen Favoriten welche pro Benutzer aber über einen kompletten GEVER Verbund zentral verwaltet werden. Der Endpoint steht nur auf Stufe PloneSite zur Verfügung und erwartet eine Einschränkung auf einen Benutzer via Benutzer-ID. Die URL setzt sich somit folgendermassen zusammen:
+
+``http://example.org/fd/@favorites/peter.mueller``
+
+
+Auflistung:
+-----------
+Mittels eines GET Request können Favoriten des Benutzers abgefragt werden. Dabei werden alle, also global über den ganzen Mandanten-Verbundes, zurückgegeben.
+
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+       GET /@favorites/peter.mueller HTTP/1.1
+       Accept: application/json
+
+
+**Beispiel-Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+      [
+          {
+              "@id": "http://localhost:8080/fd/@favorites/peter.mueller/3",
+              "favorite_id": 3,
+              "icon_class": "icon-dokument_word",
+              "oguid": "fd:68398212",
+              "title": "Richtlinien Gesetzesentwürfe",
+              "position": 1,
+              "target_url": "http://localhost:8080/fd/resolve_oguid/fd:68398212"
+          }
+          ,
+          {
+              "@id": "http://localhost:8080/fd/@favorites/peter.mueller/57",
+              "favorite_id": 57,
+              "icon_class": "contenttype-opengever-dossier-businesscasedossier",
+              "oguid": "fd:68336212",
+              "title": "Anfragen 2018",
+              "position": 2,
+              "target_url": "http://localhost:8080/fd/resolve_oguid/fd:68336212"
+          }
+      ]
+
+Favorit hinzufügen:
+-------------------
+Ein Favorit für ein beliebiges Objekt kann mittels POST Request hinzugefügt werden. Dabei wird die Oguid des Objektes im Paramter ``oguid`` erwartet.
+
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+       POST /@favorites/peter.mueller HTTP/1.1
+       Accept: application/json
+
+       {
+        "oguid": "fd:68398212"
+       }
+
+
+**Beispiel-Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 201 Created
+      Content-Type: application/json
+      Location: http://localhost:8080/fd/@favorites/peter.mueller/20
+
+      {
+          "@id": "http://localhost:8080/fd/@favorites/peter.mueller/20",
+          "favorite_id": 20,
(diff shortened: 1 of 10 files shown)


opengever/base/favorite.py
```diff
@@ -0,0 +1,80 @@
...
+            favorite.title = title
+            favorite.is_title_personalized = is_title_personalized
+
+        if position:
+            self.update_position(favorite, position, userid)
+
+        return favorite
+
+    def update_position(self, fav_to_updated, position, userid):
+        """Update the position all favorites affected by the reordering."""
```
lukasgraf: This algorithm will lead to gaps in the positions, and `max([f.pos for f in favorites]) != len(favorites) - 1`.

It will also result in an ever-increasing `max(pos)` because with every reorder, all later favorites in the list will be bumped +1, whether they need to or not. So `max(pos)` will keep growing and growing over time.

In practice, this isn't really a problem, but I think it would be much nicer if the favorite positions always were in a predictable state that makes it immediately obvious if something went wrong (there shouldn't be any gaps, and `max(pos)` should be ` == len(favorites) - 1`).

What about this algorithm:
```
- fetch all the users favorites
- build a Python list of the fav IDs in correct order according to their position
- Remove the fav ID to be moved from the list
- lst.insert(new_pos, fav_id)
- Loop through all the user's favorites:
  - If fav.pos is different from lst.index(fav.favorite_id), update its position
```

Speaking of debugging odd position states:
I think there's a [TOCTTOU race condition](https://en.wikipedia.org/wiki/Time_of_check_to_time_of_use) here: If multiple requests for updating favorite positions hit this endpoint at the same time, and they overlap, the result might be inconsistent. Depdending on how exactly the reordering UI is built on the frontend, this might or might not happen often. Because its just favorites and their positions, it's not critical though I think.

opengever/base/model/favorite.py
```diff
@@ -42,6 +45,35 @@ class Favorite(Base):
...
+            return 'tooltip'
+
+    def get_tooltip_url(self):
+        url = self.get_url()
+        if self.tooltip_view:
+            return u'{}/{}'.format(url, self.tooltip_view)
+
+        return None
+
+    def get_url(self):
```
lukasgraf: I would maybe call this `get_target_url` (and also `target_url` in the model) - it's the URL of the target object, not of the favorite.

docs/public/dev-manual/api/favorite.rst
```diff
@@ -0,0 +1,149 @@
+.. _favorites:
+
+Favoriten
+=========
+
+Der ``@favorites`` bietet alle Funktion für die Auflistung und Bearbeitung der globalen Favoriten welche pro Benutzer aber über einen kompletten GEVER Verbund zentral verwaltet werden. Der Endpoint steht nur auf Stufe PloneSite zur Verfügung und erwartet eine Einschränkung auf einen Benutzer via Benutzer-ID. Die URL setzt sich sommit folgendermassen zusammen:
```
lukasgraf: -> *Der ``@favorites`` endpoint bietet *

lukasgraf:
@phgross second part of review done. Here's the documentation for the plone.restapi content manipulation status codes:
http://plonerestapi.readthedocs.io/en/latest/content.html#id1

phgross:
@lukasgraf updated the PR according to your feedback. Please have another look.

opengever/api/favorites.py
```diff
@@ -0,0 +1,202 @@
...
+        if userid != api.user.get_current().getId():
+            raise Unauthorized(
+                "It's not allowed to access favorites of other users.")
+
+        portal_url = api.portal.get().absolute_url()
+
+        if favorite_id:
+            favorite = Favorite.query.get(favorite_id)
+            if favorite.userid != userid:
+                raise NotFound(u'Parameter mismatch: Favorite {} is '
```
lukasgraf: When raising `NotFound`, `plone.rest` apparently supplies its own error description:
```
Resource not found: http://nohost/plone/@favorites/nicole.kohler/1
```
Therefore I would simply do `raise NotFound`, since the message is ignored anyway.

However, what exactly happens if a favorite ID is specified that does not exist at all? Maybe both cases could be adressed by acutally doing a query with `filter_by(id= favorite_id, user_id=userid)` and then just always raising `NotFound` if `fav is None`?

opengever/api/tests/test_favorites.py
```diff
@@ -0,0 +1,467 @@
...
+                         headers={'Accept': 'application/json'})
+
+        self.assertEqual(
+            {"message": "Must supply user ID as URL and optional the "
+             "favorite id as path parameter.",
+             "type": "BadRequest"},
+            browser.json)
+
+    @browsing
+    def test_raises_notfound_when_accessing_favorites_of_other_user(self, browser):
```
lukasgraf: `test_raises_notfound` -> `test_raises_unauthorized`

opengever/api/tests/test_favorites.py
```diff
@@ -0,0 +1,467 @@
...
+            browser.open(url, method='POST', data="{}",
+                         headers={'Accept': 'application/json',
+                                  'Content-Type': 'application/json'})
+
+        self.assertEquals(
+            {u'message': u'Missing parameter oguid', u'type': u'BadRequest'},
+            browser.json)
+
+    @browsing
+    def test_raises_notfound_when_accessing_favorites_of_other_user(self, browser):
```
lukasgraf: `test_raises_notfound` -> `test_raises_unauthorized`

opengever/base/model/favorite.py
```diff
@@ -42,6 +45,38 @@ class Favorite(Base):
...
                       onupdate=utcnow_tz_aware)
 
+    def serialize(self, portal_url):
+        return {
+            '@id': self.api_url(portal_url),
+            'favorite_id': self.favorite_id,
+            'oguid': self.oguid.id,
+            'title': self.title,
+            'icon_class': self.icon_class,
+            'url': self.get_target_url(),
```
lukasgraf: I would call this `target_url`, just to avoid confusion.

phgross:
@lukasgraf updated again ...

PR #4145, opened 2018-04-11
Fix logout worker when using SSO

bierik:
Reload the page with the actual URL instead of a simple location reload.
This should trigger the reload with the correct URL eg. portal logout
URL.

Backport needed: `2018.2-stable`


bierik:
@phgross updated

PR #4152, opened 2018-04-12
 Break activity details out per paragraph

Rotonen:
* Breaks the activity details out per paragraphs on notify mails with multi line details

As far as I can tell these are not printed on digests.

Closes https://github.com/4teamwork/opengever.core/issues/3365

### DIFF
--- a/opengever/activity/tests/test_mail.py
+++ b/opengever/activity/tests/test_mail.py
@@ -20,6 +20,17 @@ class TestEmailNotification(IntegrationTestCase):
         # XXX - we cannot yet fixturize SQL objects
         create(Builder('watcher').having(actorid=self.regular_user.id))
         # XXX - we cannot yet fixturize SQL objects
+        create(
+            Builder('notification_setting')
+            .having(
+                kind='task-commented',
+                userid=self.regular_user.getId(),
+                mail_notification_roles=[TASK_RESPONSIBLE_ROLE],
+                badge_notification_roles=[],
+                digest_notification_roles=[],
+                ),
+            )
+        # XXX - we cannot yet fixturize SQL objects
         # The secretariat user is a part of the inbox group
         create(Builder('watcher').having(actorid=self.secretariat_user.id))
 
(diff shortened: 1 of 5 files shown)


opengever/activity/tests/test_mail.py
```diff
@@ -3,132 +3,110 @@
...
+                self.portal,
+                view='notification-settings/save',
+                data={
+                    'kind': 'task-commented',
+                    'mail': json.dumps([TASK_RESPONSIBLE_ROLE]),
+                    'badge': json.dumps([]),
+                    'digest': json.dumps([]),
+                    },
+                )
+        # XXX - did not quickly find a neater way to get the task
```
phgross: Return browser.context  in the `create_task_via_browser` method so you got the task plone object.

opengever/activity/tests/test_mail.py
```diff
@@ -3,132 +3,110 @@
...
-        browser.login().open(self.dossier, view='++add++opengever.task.task')
-        browser.fill({'Title': 'Test Task',
-                      'Task Type': 'comment'})
-
-        form = browser.find_form_by_field('Responsible')
-        form.find_widget('Responsible').fill(self.org_unit.id() + ':hugo.boss')
+    def test_notification_summary_is_split_into_paragraphs(self, browser):
+        self.login(self.dossier_responsible, browser)
+        self.create_task_via_browser(browser)
+        # XXX - not the neatest way to change the setting
```
phgross:
```
        create(Builder('notification_setting')
               .having(kind='task-commented',
                       userid=self.regular_user.getId(),
                       mail_notification_roles=[TASK_RESPONSIBLE_ROLE],
                       badge_notification_roles=[],
                       digest_notification_roles=[]))
```

opengever/activity/tests/test_mail.py
Rotonen: The browser lands on the dossier container of the task after the form, not on the task itself, and adding the extra click there slows down all of the tests in the class.

Rotonen:
@phgross ready for rereview.

PR #4153, opened 2018-04-13
Add favorite action viewlet

### LINKED ISSUE #4103: OGIP 32: mark as favorite
The UI should be enhanced so that all dexterity based favorites can be marked as favorite. A "star"-icon is added to the title. It should indicate whether the current item is marked as favorite (filled or not). `fa-star` should be used.
The REST-API endpoint `@toggle-favorite` from #4102 is used for toggling the state.
bildschirmfoto 2018-03-13 um 09 41 29
bildschirmfoto 2018-03-13 um 09 43 57
This should respect the feature flag #4104.

It is not possible to modify the title when marking as favorite; it is possible later though.

phgross:
This PR adds an additional viewlet for the `Above contenttitle`-manager with the new favorite action (star-icon): 
 - The requests to add or delete a favorite is done by an ajax call to the API (see #4133)
 - Multiple clicks are handled and ignored during a request.
 - The viewlet is only visible if the newly introduced feature-flag is enabled and the content supports oguid (dexterity objects)
 - Caching of the viewlet is handled by a separate eTag adapter.

Styling in a separate PR in plonetheme.teamraum https://github.com/4teamwork/plonetheme.teamraum/pull/627

favorite

Closes #4103

### DIFF
--- a/opengever/base/browser/resources/Favorites.js
+++ b/opengever/base/browser/resources/Favorites.js
@@ -1,0 +1,74 @@
+(function(global, $, Controller) {
+
+  function FavoriteController() {
+    Controller.call(this);
+
+    this.toggle_favorite = function(target) {
+      if ($('#mark-as-favorite').hasClass('loading')){
+        return;
+      }
+
+      if ($('#mark-as-favorite').hasClass('is-favorite')){
+        this.delete_favorite(target);
+      } else {
+        this.add_favorite(target);
+      }
+    };
+
+    this.add_favorite = function(target) {
+      url = $('#favorite-action').data('url');
+      target.addClass('loading');
+
+      return $.ajax({
+        url: url,
+        method: "POST",
+        headers: {
+          Accept: "application/json",
+          "Content-Type": "application/json"
+        },
+        data: JSON.stringify({
+          oguid: $('#favorite-action').data('oguid')})
+      }).done(function(data) {
+        target.data('favorite-id', data['favorite_id']);
+        target.addClass('is-favorite');
+        target.removeClass('loading');
+      }).fail(function(data){
+        target.removeClass('loading');
+      });
+
+    };
+
+    this.delete_favorite = function(target) {
+      var url = $('#favorite-action').data('url') + '/' + target.data('favorite-id');
+      target.addClass('loading');
+
+      $.ajax(url, {
+        url: url,
+        method: "DELETE",
+        headers: {
+          Accept: "application/json",
+          "Content-Type": "application/json"
+        }}).success(function() {
+          target.data('favorite-id', null);
+          target.removeClass('is-favorite');
+          target.removeClass('loading');
+        }).fail(function(data) {
+          target.removeClass('loading');
+        });
+    };
+
+    this.events = [
+      {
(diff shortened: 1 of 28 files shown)


phgross:
@lukasgraf updated ...

lukasgraf:
@phgross you may need to rerun the tests, pypi is having some issues right now (see https://status.python.org/)

opengever/base/browser/resources/Favorites.js
```diff
@@ -0,0 +1,74 @@
...
+        target.data('favorite-id', data['favorite_id']);
+        target.addClass('is-favorite');
+        target.removeClass('loading');
+      }).fail(function(data){
+        target.removeClass('loading');
+      });
+
+    };
+
+    this.delete_favorite = function     (target) {
```
lukasgraf: @phgross this is still present it seems?

deiferni:
so, uhm, why is the favourite button revolving once pressed? (i know the request takes time but ...) why not just highlight it? i find that a bit obtrusive for such a straightforward action 🤔.

phgross:
@lukasgraf updated according to you Feedback in #4164:
 - unqiueconstraint for the favorites table (oguid, userid)
 - Adjust @favorites POST endpoint, when adding an already existing favorite (return HTTP 409)
 - Separate handling for adding an already existing favorite or deleting an already deleted favorite in the favorite action.

Could you take another look to the last 3 commits?


@deiferni relating the revolving icon: 
I like to have an easy and simple way to show the user feedback when GEVER is adding or deleting the favorite and the request is slow fore some reasons. Usually the effect is not visible for the user, because the request is fast enough (I add an delay for the gif 😉). Better opinions to give the user feedback, that the request is "running"?

deiferni:
Honestly i would hide that from the user (i think he does not care if the request is slow) and instead just immediately toggle the selected state of the star. It may be personal preference, but i'd prefer an unobtrusive simple state change here.
And as you said, the request is expected to be fast. So it might be users see a revolving start for 100ms and then the filled star. That may appear as flickering, and i worry that this may be annoying (to me i guess it would be).

I'd still keep the functionality that prevents clicking until the request has finished though, just without the animation. If you're unhappy with my suggestions we could discuss introducing a "pushed" state, maybe with a small material-design style halo around the outline,  🤔 e.g.:
screen shot 2018-04-17 at 16 26 07


Btw here is how stackoverflow handles errors in their favorites system:
screen shot 2018-04-17 at 16 11 07

lukasgraf:
@deiferni the state change *is* pretty much immediate in practice. Except for when the request hangs for some reason, in which case there *should* be user feedback. Also, changing the the state in the GUI before success has been acknowledged is a bad interaction pattern.

Therefore I think the current solution is pretty much perfect as it is.

deiferni:
i strongly disagree and find the revolving animation quite horrible 😄. but as already mentioned this may be personal preference, and thus overruled.

lukasgraf:
@deiferni Overruled then ;) If a customer opens a ticket for flickering, I'll handle it and you get 🍻 - deal? 😜

PR #4162, opened 2018-04-16
 Prefix tasks in zip exports

### LINKED ISSUE #4112: Adjust ZIP representation for tasks
When creating a Zip export of a dossier which contains documents inside tasks, tasks are represented as folders named with the task title. Because this could be confusing for some users, the "task-folders" name should be prefixed with `Aufgabe: `.

Todos:
 - make path_prefix getter in ftw.zipexport easier customizable
 - implement prefix for task representation

See [Extranet](https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/239) for further information.

Rotonen:
* Bumped ftw.zipexport to get the new zip prefix functionality
* Added a property to tasks for the zip prefix

Closes #4112


opengever/task/task.py
```diff
@@ -267,6 +268,11 @@ class Task(Container):
     def __init__(self, *args, **kwargs):
         super(Task, self).__init__(*args, **kwargs)
 
+    @property
+    def zipexport_title(self):
+        translated_taskname = translate(_(u'Task', default=u'Task'), context=api.portal.get().REQUEST)
+        return u' - '.join((translated_taskname, self.Title().decode('utf-8'), ))
```
phgross: Please use `:` instead of `-`.

For example `Aufgaben: Anfrage Kennzahlen 2018`

Rotonen: Nocando, pathname sanitation converts `:` to `-`. Rather having it space-dash separated as that looks neatest of the alternatives available.

phgross: Oh yes, you're right 👍

Issue #4105, opened 2018-03-27
OGIP 32: switch repository favorites to new implementation


jone:
Since we implement a new generic favorites system, we want the existing repository favorites to also work together with the new system and use the same storage.

- [x] let the portlet use the new `@list-repository-favorites` REST API endpint for getting the user's favorites (#4102)
- [x] make the existing portlet JS use the new `@toggle-favorite`REST API endpoint for toggling favorites (#4102)
- [x] migrate [existing favorites](https://github.com/4teamwork/opengever.core/blob/026471b04ac1b4a1b68275283ad823e40ea36204/opengever/portlets/tree/favorites.py#L35) to the new SQL store (#4100)
- [x] reimplement caching of repository favorites
  - use last max(last modified) of the current user's favorites as cache key
  - put the cache key in the DOM and only load the favorites when the cache key changes
    - maybe use HTTP caching headers
  - when the cachekey is in the DOM, the [etag value](https://github.com/4teamwork/opengever.core/blob/026471b04ac1b4a1b68275283ad823e40ea36204/opengever/portlets/tree/favorites.py#L135) must be kept / adapted, otherwise it can be removed
- [x] remove the new longer used favorites  [storage](https://github.com/4teamwork/opengever.core/blob/026471b04ac1b4a1b68275283ad823e40ea36204/opengever/portlets/tree/favorites.py#L35) and [view](https://github.com/4teamwork/opengever.core/blob/026471b04ac1b4a1b68275283ad823e40ea36204/opengever/portlets/tree/favorites.py#L72) code
- [x] make sure that using the favorites-toggle in the title (#4103) also reloads the favorites portlet
- [x] review python and JS code of the repository favorites feature and remove everything which is not needed anymore

_harvest: `Feature: Favoriten (OGIP 32)`_

phgross:
@elioschmutz yes that make sense.

By the way the favorites feature is enabled everywhere.

PR #4200, opened 2018-04-20
Add favorites menu and details view

### LINKED ISSUE #4107: OGIP 32: favorites menu
When the feature-flag is enabled (#4104), a new button near the notifications button should appear, allowing access to the favorites.

- [ ] button (`fa-star`) near notifications button
- [ ] click on button opens a menu showing a list of favorites
  - with a scrollable area like notifications have
  - no size limitation (the most favorites we've seen in the field are 40)
- [ ] data is loaded with the `@list-favorites` endpoint (#4102)
- [ ] only when feature flag enabled (#4104)

### LINKED ISSUE #4108: OGIP 32: manage favorites
The user should be able to manage the personal favorites, therefore we build a managment view.

- implement with VueJS
- accessible via favorites menu from #4107
- list favorites from `@list-favorites` (#4102), including favorites of other admin units
- show tooltip for documents (`tooltip_url`) on hover
- show object type icon
- show admin unit title
- reorder favorites with drag'n'drop
- remove favorite (`@remove-favorite`, #4102)
- change favorite title (`@rename-favorite`, #4102)

phgross:
This PR adds the favorite menu and the favorite details view, both frontend components are implemented as separate Vue Apps.

**Favorites Menu**
 - Shows all favorites in the same manually defined order
 - Viewlet is only available when feature flag is enabled
 - Fetching the data via REST API 
 - Closes #4107 

favorite_menu

**Favorite View**
 - Provides functionality to rename, delete and update the order
 - Accessible via the overview link in the favorites menu
 - Closes #4108 

favorites view

Styling: https://github.com/4teamwork/plonetheme.teamraum/pull/628

During the implementation I've found some bugs in the endpoints, which I've fixed as well.


opengever/base/browser/resources/favorites_app.js
```diff
@@ -0,0 +1,168 @@
...
+
+    deleteAction: function (entry) {
+      var self = this;
+      var payload = {
+        _authenticator: this.authtoken,
+        title: this.newTitle
+      };
+      this.requester.delete(entry['@id'], payload).then(function (response) {
+        if (response.status === 204) {
+          self.fetchData();
```
elioschmutz: @phgross I came across your new implementation while realizing #4099.

You always refetch the data after you did changes on the content, i.e. removing an item. Instead of refeching the data, you can just update your `entries`-array with your changes. I.e. 

```js
self.entries = self.entries.filter(function(e) {
    return e['@id'] !== entry['@id']; 
});
```

And the view-component will update it.
You'll save a request and it's a common practice.

opengever/base/browser/resources/favorites_app.js
```diff
@@ -0,0 +1,168 @@
...
+    },
+
+    updatePosition: function (entry, newPosition, oldPosition) {
+      var self = this;
+      var payload = {
+        _authenticator: this.authtoken,
+        position: newPosition};
+      this.requester.patch(entry['@id'], payload).then(function (response) {
+        if (response.status === 204) {
+          self.fetchData();
```
elioschmutz: Same here...

opengever/base/viewlets/favorites_menu.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class FavoritesMenuViewlet(common.ViewletBase):
+
+    index = ViewPageTemplateFile('favorites_menu.pt')
+
+    @property
+    def vuejs_template(self):
+        return get_handlebars_template(
+            os.path.join(os.path.dirname(__file__), 'favorites_menu.html'))
```
deiferni: Same here i'd use `resource_filename`.

opengever/base/viewlets/favorites_menu.py
```diff
@@ -32,5 +33,4 @@ def translations(self):
 
 
     def available(self):
-        # todo check for feature flag
-        return True
+        return is_favorites_feature_enabled()
```
deiferni: the default for the feature is `False`, might be a good idea to enable it by default instead as it is genuinely useful?

opengever/base/favorite.py
```diff
@@ -17,6 +17,12 @@ def delete(self, userid, fav_id):
             # inexistent favorite-id or not ownded by given user
             raise NotFound
 
+        # update positions
+        favorites_to_reduce = Favorite.query.by_userid(userid).filter(
+            Favorite.position > favorite.position)
+        for fav_to_reduce in favorites_to_reduce:
+            fav_to_reduce.position = fav_to_reduce.position - 1
```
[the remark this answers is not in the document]
phgross: No gaps aren't allowed. 

I don't get your point with the `enumerate` or how this should work, maybe you can show be today.

opengever/base/viewlets/favorites_menu.py
phgross: Yes, the idea is to enable the favorites feature by default, but currently the repository favorites are not working with the new favorite model (see #4105), as soon the repository favorites are switched we can enable the favorites feature by default.

opengever/base/browser/resources/favorites_app.js
phgross: Hi @elioschmutz thanks for your input. Maethu did the implementation of the frontend part, but as far as i know, he were absolutely aware of this. The idea of the "refetching" request, is to avoid problems with "outdated" positions.

I would propose to not change the implementation in this PR and let @maethu take another look after his vacation, if necessary we can change this part with a later PR.

opengever/base/favorite.py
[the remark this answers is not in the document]
deiferni:
```python

    for new_pos, fav_to_reduce in numerate(favorites_to_reduce, start=favorite.position+1):
        fav_to_reduce.position = new_pos
```

opengever/api/tests/test_favorites.py
```diff
@@ -281,6 +281,31 @@ def test_removes_favorite_when_already_exists_for_user(self, browser):
...
+                      .for_object(self.leaf_repofolder))
+
+        url = '{}/@favorites/{}/{}'.format(
+            self.portal.absolute_url(), self.administrator.getId(),
+            fav1.favorite_id)
+        browser.open(url, method='DELETE', headers={'Accept': 'application/json'})
+
+        self.assertEqual(0, fav2.position)
+        self.assertEqual(1, fav3.position)
+
```
elioschmutz: This test already exists with the same title (see one test above https://github.com/4teamwork/opengever.core/pull/4200/files#diff-c0b8cd45221b7403f7e99c9c69785409R267)...

The first test would fail if it gets executed, but it doesn't.

phgross: 🙈 thanks for the hint, updated ...

phgross:
@deiferni @elioschmutz updated the PR according to your feedback

PR #4219, opened 2018-04-23
Implement a view for the manger to see the current config

### LINKED ISSUE #4093: Create view that displays current GEVER configuration
We would like to have a view that displays the current configuration for a particular GEVER admin unit.

The view should list the same settings that have been exposed in the `@config` API endpoint in #4035 - but be aware that this endpoint is already being used by customers, and is therefore considered part of a public API that needs to remain stable.

The view should look half decent, but not polished yet. In the end it should probably be accessible by the end user, but for now I would suggest to protect it with a custom permission that is managed to the manager role.

Rotonen:
* Move the config generation to an adapter for potential future configurable versionability
* Implement the manager view over the restapi endpoint in handlebars

Closes #4093

### DIFF
--- a/opengever/core/profiles/default/jsregistry.xml
+++ b/opengever/core/profiles/default/jsregistry.xml
@@ -529,6 +529,17 @@
 
   <javascript
       cacheable="True"
+      compression="safe"
+      cookable="True"
+      expression=""
+      enabled="True"
+      id="++resource++opengever.base/config.js"
+      inline="False"
+      insert-after="++resource++opengever.base/Controller.js"
+      />
+
+  <javascript
+      cacheable="True"
       compression="none"
       cookable="True"
       expression=""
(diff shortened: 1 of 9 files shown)


Rotonen:
And here is a screenshot:

screen shot 2018-04-23 at 15 53 09

opengever/core/profiles/default/jsregistry.xml
```diff
@@ -517,4 +517,15 @@
       insert-after="++resource++opengever.base/Controller.js"
       />
 
+  <javascript
+      cacheable="True"
+      compression="none"
```
phgross: why?

Rotonen: I was under the impression our handlebars integration was not compression safe. It seems I was mistaken as both notifications and settings are marked safe.

Will amend.

Rotonen:
@phgross also amended the new preview refresh flag. Ready for rereview.

PR #4220, opened 2018-04-23
Add bumblebee auto-refresh feature flag.

deiferni:
Add a feature-flag to enable/disable bumblebee auto-refresh. This allows us to avoid 404s in the browser console when the backend's `notifications` endpoint is not available.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 ---------------------
 
 - Implement bumblebee's auto refresh functionality. [siegy]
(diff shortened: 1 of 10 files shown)


deiferni:
the build seems to have passed https://ci.4teamwork.ch/builds/157947, but somehow this is not visible on the commits. might have something to do with the fact that this is a branch based on your branch 🤔.

deiferni:
i think it is less confusing to merge this one into your pr and to the final review there, otherwise that will happen twice.

PR #4224, opened 2018-04-24
Fix dates for transported objects.

### LINKED ISSUE #4156: Fix date for transported objects
Splitting off from https://github.com/4teamwork/opengever.core/pull/4155

There is a high likelihood every object with a date ever transported now has a datetime instead.

phgross:
Adds an upgradestep which fixes dates for transported objects (closes #4156). See #4155 for more information.

The upgradestep is skipped on a single admin_unit deployment, because the transporter is only used in multi admin_unit communication and potentially long running.

### DIFF
--- a/opengever/core/upgrades/20180424145246_fix_date_for_transported_objects/upgrade.py
+++ b/opengever/core/upgrades/20180424145246_fix_date_for_transported_objects/upgrade.py
@@ -1,0 +1,60 @@
+from datetime import datetime
+from opengever.core.upgrade import SQLUpgradeStep
+from opengever.document.behaviors import IBaseDocument
+from opengever.task.task import ITask
+
+
+TASK_DATE_FIELDS = [
+    ('deadline', 'deadline'),
+    ('date_of_completion', 'date_of_completion'),
+    ('expectedStartOfWork', None)]
+
+DOCUMENT_DATE_FIELDS = [
+    ('document_date', 'document_date'),
+    ('receipt_date', 'receipt_date'),
+    ('delivery_date', 'delivery_date')]
+
+
+class FixDateForTransportedObjects(SQLUpgradeStep):
+    """Fix date for transported objects.
+    """
+
+    def migrate(self):
+        if not self.has_multiple_admin_units():
+            # Transporter is only used on deployments with multiple adminunits,
+            # therefore we skip on single admin unit sites.
+            return
+
+        query = {'object_provides': ITask.__identifier__}
+        for obj in self.objects(query, 'Fix task dates'):
+            self.fix_task_dates(obj)
+
+        query = {'object_provides': IBaseDocument.__identifier__}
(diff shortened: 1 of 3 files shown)


opengever/core/upgrades/20180424145246_fix_date_for_transported_objects/upgrade.py
```diff
@@ -0,0 +1,39 @@
+from datetime import datetime
+from opengever.core.upgrade import SQLUpgradeStep
+from opengever.document.behaviors import IBaseDocument
+from opengever.task.task import ITask
+
+
+TASK_DATE_FIELDS = ['deadline', 'date_of_completion', 'expectedStartOfWork']
```
deiferni: are these fields in SQL and already truncated to `Date`? or would you need to sync the task?

opengever/core/upgrades/20180424145246_fix_date_for_transported_objects/upgrade.py
```diff
@@ -0,0 +1,39 @@
...
+DOCUMENT_DATE_FIELDS = ['document_date', 'receipt_date', 'delivery_date']
```
deiferni: i think there may be an index for some of these, so that would need updating, too.

opengever/core/upgrades/20180424145246_fix_date_for_transported_objects/upgrade.py
```diff
@@ -0,0 +1,39 @@
...
+            self.fix_task_dates(obj)
+
+        query = {'object_provides': IBaseDocument.__identifier__}
+        for obj in self.objects(query, 'Fix document dates'):
+            self.fix_document_dates(obj)
+
+    def fix_task_dates(self, task):
+        for date_field in TASK_DATE_FIELDS:
+            value = getattr(task, date_field)
+            if value and type(value) == datetime:
```
deiferni: shouldn't you be able to `isinstance(value, datetime)` 🤔? Will work this way, so no need to change. Just wondering.

Rotonen: `isinstance()` also considers inheritance. This is IMO more explicit to the intention here.

opengever/core/upgrades/20180424145246_fix_date_for_transported_objects/upgrade.py
phgross: Doublechecked, the dates in the SQL are correct, no sync necessary.

opengever/core/upgrades/20180424145246_fix_date_for_transported_objects/upgrade.py
phgross: @deiferni You're right, updated the PR with a reindexObject if necessary. please have another look.

PR #4247, opened 2018-04-30
Ignore common environment management

Rotonen:
Motivation: a cleaner git status for more people. More and more people seem to use pyenv and rbenv and a virtualenv for their gever buildout.

* Sorted the gitignore files
* Removed an empty gitignore file
* Added common environment management files into the main gitignore file


Rotonen:
`.gitkeep` would be the convention for that. Which way do I amend?

Rotonen:
@deiferni dropped the deletion commit. Ready for rereview.

Issue #3948, opened 2018-02-07
Ausbau non-Word SPV aus opengever.core


deiferni:
The following is a loose and chaotic, ever-growing list of stuff that may be removed that stumble upon while dropping functionality:
- [x] Feature Flag entfernen, search for `is_word_implementation_enabled`
- [x] Remove `activate_meeting_word_implementation`
- [x] Test feature `word_meetings` entfernen
- [x] `view='protocol'` entfernen
- [x] `generate_protocol` und `update_protocol` view entfernen
- [x] drop `manually_generated_excerpts` `generate_excerpt`
- [x] drop non-word meeting view 
- [x] ~~drop `Builder('generated_excerpt')`~~
- [x] `meeting_excerpts` Tabelle entfernen
- [x] `get_protocol_template`
- [x] double check `download_generated_protocol`
- [x] double check `download_protocol_json`
- [x] Drop unneeded params from `ExcerptProtocolData` (e.g.: `include_initial_position`)
- [x] Test flag `word-meeting` entfernen
- [x] Tests für Non-Word SPV entfernen
- [x] Non-Word Komponenten entfernen
- [x] `if`s im Code entfernen
- [x] `Meeting` drop ~~`GeneratedExcerpt` and~~ `excerpt_documents`
- [x] Drop `is_word_meeting_implementation_enabled` and `require_word_meeting_feature`
- [x] Drop usage of `protocol_template`
- [x] Drop usage of `excerpt_template`
- [x] `Proposal.update_excerpt`
- [x] `Proposal.generate_excerpt`
- [x] ~~`Proposal.register_excerpt`~~
- [x] `ExcerptOperations`
- [x] ~~`UpdateExcerptInDossierCommand`~~
- [x] ~~`CreateGeneratedDocumentCommand`~~
- [x] ~~`UpdateGeneratedDocumentCommand`~~
- [x] `ProtocolOperations`
- [x] `TrixFieldWidget` and all the trix stuff 😢 `TrixDataConverter``ITrixWidget` `TrixWidget`
- [x] `trix2sablon`
- [x] there should be no more "html:" prefixed dict keys used in sablon
- [x] remove `DownloadGeneratedProtocol` and `download_generated_protocol` debug view and also remove from docs
- [x] drop the fields:
  - considerations
  - copy_for_attention
  - decision
  - decision_draft
  - disclose_to
  - discussion
  - initial_position
  - legal_basis
  - proposed_action
  - publish_in
[...]

PR #4249, opened 2018-04-30
Respect feature flag when filling sample data.

deiferni:
The `html:*` fields only make sense when the word meeting feature is disabled, otherwise the fields have been replaced by the word-file and are no longer available.

_This is currently important as we're about to create many new templates. Merge ASAP for a process speedup. 💛!_


deiferni:
It is not beautiful 😉 but it will go away again with https://github.com/4teamwork/opengever.core/issues/3948

PR #4253, opened 2018-05-01
Add service to create OC JWT URLs

### LINKED ISSUE #4136: OneOffixx: Service for creating oc:<JWT> URLs
Similiar to `OfficeConnectorCheckoutURL`.
action=oneofffixx

https://github.com/4teamwork/opengever.core/blob/c723c47a8e668b9f9b35e9ba827a330e77260ba1/opengever/officeconnector/service.py#L71-L85

njohner:
Add the service to create a JWT for the `oneoffixx` action for office connector.

resolves #4136

### DIFF
--- a/opengever/document/document.py
+++ b/opengever/document/document.py
@@ -6,6 +6,7 @@ from Acquisition import aq_parent
 from collective import dexteritytextindexer
 from ftw.mail.interfaces import IEmailAddress
 from ftw.tabbedview.interfaces import ITabbedviewUploadable
+from opengever.oneoffixx import is_oneoffixx_feature_enabled
 from opengever.base.interfaces import IRedirector
 from opengever.document import _
 from opengever.document.base import BaseDocumentMixin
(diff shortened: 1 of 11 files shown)


njohner:
Ok so this is ready

opengever/document/document.py
```diff
@@ -362,16 +362,23 @@ def checkout_and_get_office_connector_url(self):
...
+                    request,
+                    self,
+                    dict(action=action),
+                ))
+            else:
+                redirector.redirect(
+                    '%s/external_edit' % self.absolute_url(),
+                    target='_self',
+                    timeout=1000)
+        elif action == "oneoffixx":
```
Rotonen: Missing the feature flag check here. Also this should be within the branch which checks for the direct checkin feature being available.

njohner: I consciously skipped the feature test here. My reasoning was that for the checkout, the action is there whether the officeconnector feature is enabled or not, so the check is necessary to decide what to do. In the oneoffixx case, the action is only available if the corresponding feature is enabled.
Of course we can add the test to be on the safe side for the future

Rotonen: But we also should not let it redirect to the OC URL logic if OC cannot then soon after checkout the document.

Rotonen: Actually, thinking about it, the feature check view for oneoffixx should probably return `False` even if it is enabled so long as the direct checkin feature is disabled.

We have no neat way to do nested features so I'm unsure if this would confuse anyone, but since it is only us mucking with the settings, we should be fine with that (and this should then get documented as well).

njohner:
I think this is ready

njohner:
Just added a comment on the logic

PR #4255, opened 2018-05-01
Add service to generate payload for officeconnector oneoffixx action

### LINKED ISSUE #4137: OneOffixx: oc_oneoffixx Service
Similiar to OfficeConnectorCheckoutPayload.

Lives on the site root and operates via the document UUID.
https://github.com/4teamwork/opengever.core/blob/c723c47a8e668b9f9b35e9ba827a330e77260ba1/opengever/officeconnector/configure.zcml#L61-L68

https://github.com/4teamwork/opengever.core/blob/c723c47a8e668b9f9b35e9ba827a330e77260ba1/opengever/officeconnector/service.py#L88-L128

Includes `oneoffixx-connect-xml-url` defined in #4135

njohner:
Add service to generate payload for the officeconnector oneoffixx action. The payload contains the necessary information to retrieve the connect XML (`connect-xml` key)

resolves #4137

### DIFF
--- a/opengever/officeconnector/helpers.py
+++ b/opengever/officeconnector/helpers.py
@@ -34,25 +34,25 @@ def parse_bcc(request):
 
 def parse_documents(request, context, action):
     documents = []
-
     if (
             request['REQUEST_METHOD'] == 'GET'
             or request['REQUEST_METHOD'] == 'POST'
-            and 'BODY' not in request
+            and ('BODY' not in request
+                 or isinstance(json.loads(request['BODY']), list))
         ):
         # Feature enabled for the wrong content type
         if not IBaseDocument.providedBy(context):
             raise NotFound
 
         # for checkout and attach actions, the document needs to have a file
-        # for the oneoffixx action, the document should be in the shadow state
-        if not (context.has_file()
-                or action == "oneoffixx" and context.is_shadow_document()):
+        # for the oneoffixx action or checkout action fetched from
+        # OfficeConnectorOneOffixxPayload, the document should be in the shadow state
+        if not (context.has_file() or context.is_shadow_document()):
             raise NotFound
 
         documents.append(context)
 
-    if request['REQUEST_METHOD'] == 'POST' and 'BODY' in request:
+    elif request['REQUEST_METHOD'] == 'POST' and 'BODY' in request:
         payload = json.loads(request['BODY'])
         paths = payload.get('documents', None)
 
--- a/opengever/officeconnector/service.py
+++ b/opengever/officeconnector/service.py
@@ -12,6 +12,7 @@ from plone.protect import createToken
 from plone.rest import Service
 from zExceptions import Forbidden
 from zExceptions import NotFound
+from zope.annotation.interfaces import IAnnotations
 from zope.event import notify
 from zope.i18n import translate
 import json
@@ -110,6 +111,10 @@ class OfficeConnectorPayload(Service):
         super(OfficeConnectorPayload, self).__init__(context, request)
         self.uuids = json.loads(request['BODY'])
 
+    @staticmethod
+    def document_is_valid(document):
+        return document and document.has_file()
+
     def get_base_payloads(self):
         # Require an authenticated user
         if not api.user.is_anonymous():
@@ -118,7 +123,7 @@ class OfficeConnectorPayload(Service):
             for uuid in self.uuids:
                 document = api.content.get(UID=uuid)
 
-                if document and document.has_file():
+                if self.document_is_valid(document):
                     documents.append(document)
 
         else:
@@ -131,12 +136,9 @@ class OfficeConnectorPayload(Service):
             for document in documents:
                 payloads.append(
                     {
-                        'content-type': document.get_file().contentType,
                         'csrf-token': createToken(),
                         'document-url': document.absolute_url(),
                         'document': document,
-                        'download': document.get_download_view_name(),
-                        'filename': document.get_filename(),
                         }
                     )
 
@@ -185,6 +187,9 @@ class OfficeConnectorAttachPayload(OfficeConnectorPayload):
                 dossier_notifications[parent_dossier_uuid].append(document)
 
             payload['title'] = document.title_or_id()
+            payload['content-type'] = document.get_file().contentType
+            payload['download'] = document.get_download_view_name()
+            payload['filename'] = document.get_filename()
             del payload['document']
             notify(FileAttachedToEmailEvent(document))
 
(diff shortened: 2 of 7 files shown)


opengever/officeconnector/service.py
```diff
@@ -228,3 +236,42 @@ def render(self):
...
 
         return json.dumps(payloads)
+
+class OfficeConnectorOneOffixxPayload(OfficeConnectorPayload):
+    """Issue JSON instruction payloads for OfficeConnector.
+
+    Consists of the minimal instruction set with which to perform a full
+    checkout checkin cycle for a file attached to a document.
+    """
+    @staticmethod
```
Rotonen: Rather module scope the method, make a mixin or accept the duplication.

opengever/officeconnector/service.py
```diff
@@ -129,12 +133,9 @@ def get_base_payloads(self):
             for document in documents:
                 payloads.append(
                     {
-                        'content-type': document.get_file().contentType,
```
Rotonen: Why do these get moved?

opengever/officeconnector/service.py
```diff
@@ -228,3 +236,42 @@ def render(self):
...
+        for payload in payloads:
+            # A permission check to verify the user is also able to upload
+            authorized = api.user.has_permission(
+                'Modify portal content',
+                obj=payload['document'],
+                )
+
+            if authorized:
+                del payload['document']
+                payload['checkin-with-comment'] = '@@checkin_document'
```
Rotonen: We do not need checkin or upload data in the oneoffixx payload. We'll need information on where the XML is, how to unshadow and then the checkout URL (JWT) of the document to skip a request and directly fetch the normal checkout payload.

opengever/officeconnector/service.py
```diff
@@ -108,6 +108,10 @@ def __init__(self, context, request):
         super(OfficeConnectorPayload, self).__init__(context, request)
         self.uuids = json.loads(request['BODY'])
 
+    @staticmethod
```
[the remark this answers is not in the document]
njohner: I did that to overwrite it in `OfficeConnectorOneOffixxPayload` which uses this one as a baseclass, but has a different condition for documents being valid. 
So module scope would not work here. Duplication would mean duplicate the whole baseclass

njohner:
Ok can you have another look?
The payload now contains:
* where to find the XML
* the checkout url JWT
* the filename
* the document url
I'm not passing anything for the unshadowing. I think this should not be necessary, as the checkin will unlock the document and the document unlock also removes the shadow state. (https://github.com/4teamwork/opengever.core/blob/master/opengever/document/document.py#L334)

opengever/officeconnector/service.py
```diff
@@ -230,3 +256,43 @@ def render(self):
         self.request.response.setHeader('Content-type', 'application/json')
 
         return json.dumps(payloads)
+
+
+class OfficeConnectorOneOffixxPayload(OfficeConnectorPayload):
+    """Issue JSON instruction payloads for OfficeConnector.
```
Rotonen: I think we need a fresh docstring here.

opengever/officeconnector/service.py
```diff
@@ -230,3 +256,43 @@ def render(self):
...
+    """
+    @staticmethod
+    def document_is_valid(document):
+        return document and document.is_shadow_document()
+
+    def render(self):
+        payloads = self.get_base_payloads()
+
+        # Upload API will be included as a registry setting in the future when
+        # the plone.api endpoint gets made - for now we've made a custom upload
```
Rotonen: This is not correct at the moment. We're pending TUS support on the client side, `plone.restapi` already has support baked in.

http://plonerestapi.readthedocs.io/en/latest/tusupload.html

Sidenote: `plone.api` is also the wrong product. I know we named it badly, but people objected to having called it `plone.ipa` in 2011, even though it's about the inverse of what an api is (maybe that had something to do with it having been birthed in the belgian beer sprint, see https://www.coactivate.org/projects/belgian-beer-sprint/project-home for some of the the history). It's essentially just a collection of shorthand utils for most commonly done things deemed too Zope-y and repetitive.

opengever/officeconnector/service.py
```diff
@@ -230,3 +256,43 @@ def render(self):
...
+            # A permission check to verify the user is also able to upload
+            authorized = api.user.has_permission(
+                'Modify portal content',
+                obj=payload['document'],
+                )
+
+            if authorized:
+                document = payload['document']
+                checkout_token = create_oc_url(self.request, document, {"action": "checkout"})
+                payload['filename'] = IAnnotations(document).get("filename")
```
Rotonen: We also need to do this conditionally in the normal checkout payload for documents if they're in the shadow state so the OC-side logic of avoiding to download when we already have a file kicks in.

opengever/officeconnector/service.py
njohner: That was a copy-paste fail from the Checkout Payload. That comment should not be here anyway as we do not include the upload information in the OneOffixx payload. But hen maybe that comment should be corrected in the Checkout payload...

opengever/oneoffixx/browser/form.py
```diff
@@ -153,17 +153,9 @@ def finish_document_creation(self, data):
...
+        """Add the oneoffixx external_editor URL to redirector queue.
         """
-        # Check out the new document
-        manager = self.context.restrictedTraverse('checkout_documents')
-
-        # When it will be implemented we should use the new oneoffix
-        # action instead of checkout
-        # we can extend setup_external_edit_redirect to take a keyword argument
-        # action
-        manager.checkout(new_doc)
```
njohner: The checkout will be done from OfficeConnector

opengever/officeconnector/helpers.py
```diff
@@ -34,25 +34,25 @@ def parse_bcc(request):
 
 def parse_documents(request, context, action):
     documents = []
-
     if (
             request['REQUEST_METHOD'] == 'GET'
             or request['REQUEST_METHOD'] == 'POST'
-            and 'BODY' not in request
+            and ('BODY' not in request
+                 or isinstance(json.loads(request['BODY']), list))
```
njohner: When fetching the checkout url from the OneOffixxPayload, the request is POST but has a BODY, but it's still the context that is the document... In that case the BODY is a list and not a dict in the `elif` below. If you have a cleaner solution here...

Rotonen: Not on this schedule. That'll do for now.

PR #4257, opened 2018-05-02
Show group title in the sharing view

### LINKED ISSUE #3864: Sharing and info view should display group title, if available.
The sharing view and tab currently display the group id for all groups that have been configured in `@@sharing` (Aktionen -> Freigabe).

Sharing-Tab:
screen shot 2018-01-16 at 13 19 05

Sharing-View:
screen shot 2018-01-16 at 13 27 36

- Instead of displaying the `group_id` only, the group title should be displayed, when available. This could be achieved by using the `label` method on `Group`, for example.
- Also preserve the `group_id` for debugging purposes, for example by using it as `title` of the html element or by adding it in brackets, similar to user.

See:
- label method
 https://github.com/4teamwork/opengever.ogds.models/blob/1d64efb3db926e0eb811e81fbf5c43da37707641/opengever/ogds/models/group.py#L70
- sharing view implementation
 https://github.com/4teamwork/opengever.core/blob/master/opengever/sharing/browser/sharing.py
- ogds sync that also syncs group title, if available https://github.com/4teamwork/opengever.core/pull/3638

https://basecamp.com/2768704/projects/13732725/todos/325803930

phgross:
With b1e2d59 a new configuration option `group_title_ldap_attribute` has been introduced, which sets the group title during the ogds syncrhonisation. We show now the title if exists in the Sharing view instead of only the id.

bildschirmfoto 2018-05-02 um 14 28 25

Because our LDAP stack does not support mapping additional attributes, we fetch the title information from ogds groups.

Closes #3864.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 ---------------------
 
 - Add view to get the Connect XML for OneOffixx. [njohner]
+- Show group title in the sharing view. [phgross]
 - No longer show unsortable columns as sortable, bump ftw.tabbedview to 4.1.1. [deiferni]
 - Add debug views for agenda_items and docxcompose. [deiferni]
 - Fix API get request on private dossiers. [phgross]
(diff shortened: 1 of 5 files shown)


phgross:
@deiferni yes it uses the same view: 

bildschirmfoto 2018-05-02 um 15 39 56

PR #4263, opened 2018-05-03
Enable favorites feature by default.

phgross:
Feature soll bei allen Kunden standardmässig aktiviert sein.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2018.3.0 (unreleased)
 ---------------------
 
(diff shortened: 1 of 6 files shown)


phgross:
@deiferni it's a brand-new feature-flag, which isn't configured in any customer policy. 

I'd like to still have the feature flag. So were still able to disable the feature with a simple registry configuration including an Upgradestep in the policy. The policy setting is always installed after the og.core stuff, because of the dependency.

phgross:
👍 

Just for information: we dropped the `repo-tree favourites flag`. One flag for the complete favorites- functionality is enough. 😉

PR #4269, opened 2018-05-07
Fix IE 11 JS-bug in tree.js

phgross:
IE 11 does not support the ecmascript 6 feature `default function parameters` (see http://kangax.github.io/compat-table/es6/) , therefore we have to implement the common workaround.
https://stackoverflow.com/questions/15178700/javascript-functions-and-default-parameters-not-working-in-ie-and-chrome


phgross:
@lukasgraf the tree.js, or at least the "bugy" part of it, was introduced with the favorites feature, imo no changelog entry necessary.

PR #4276, opened 2018-05-08
Make external link an href in the byline view for dossiers

### LINKED ISSUE #4126: externe Referenz klickbar machen
Es wäre 4tw-intern wie auch für unsere Kunden sehr hilfreich, wenn der Verweis auf die externe Referenz  klickbar wäre (sofern es eine URL ist). 

image

njohner:
We make the external reference clickable in the `byline` view for `businesscasedossier` and `meetingdossier`.
* Added a test that it is only an `href` when it is a url (we test that urlparse finds a `scheme`. `urlparse` recognizes a rather comprehensive list of schemes...)
* Moved `test_dossier_byline_external_reference_display` to the `TestDossierByline` class, so that it also gets tested for the `meetingdossier`.

**clickable for a "valid" href**

screen shot 2018-05-08 at 10 04 30

**non-clickable for an "invalid" href or any other referenz**

screen shot 2018-05-08 at 10 04 01
resolves #4126


docs/HISTORY.txt
```diff
@@ -5,6 +5,7 @@ Changelog
 2018.3.0 (unreleased)
 ---------------------
 
+- Make external reference clickable in dossiers. [njohner]
```
deiferni: but only if we detect an url

opengever/dossier/viewlets/byline.py
```diff
@@ -33,6 +34,14 @@ def mailto_link(self):
     def external_reference(self):
         return IDossier(self.context).external_reference
 
+    def external_reference_link(self):
+        if not self.external_reference():
```
deiferni: this should be improved a bit. I'd say the correct link should also appear for things like:
- `' http://www.gooogle.ch'`
- `'Hier: http://www.gooogle.ch.'`

fun fact: the scheme of 'Hier: http://www.gooogle.ch.' is `'hier'` 😄 so that would generate a link too.

I think it would be better to use a transform to `'text/x-web-intelligent'` which we already have, see https://github.com/4teamwork/opengever.core/blob/21ef08fcd76a7356073784380b26817595a5206b/opengever/dossier/base.py#L378-L386 for an example where this has been implemented for a dossiers comments.

njohner: Thanks, I was actually trying to see whether you already had something similar somewhere...

njohner:
Ok I think this is ready.
Tests passed when I reran them...
The failed tests were here: https://ci.4teamwork.ch/builds/164287/tasks/259948

PR #4282, opened 2018-05-08
Oneoffixx roundup

Rotonen:
The PR is currently broken and unfinished, but making it so @phgross has it somewhere visible (assigned to milestone).

What's still open is the user masquerading against their api.


lukasgraf:
Won't happen in `2018.4` anymore, mapping to `2018.5`

Rotonen:
I'm rebasing and cleaning up this branch to serve as a base to do local manual verification of the impersonating API to progress with the integration, and to later provide a springboard for actually implementing that upon those results.

https://github.com/4teamwork/opengever.core/issues/4865 will need to get tackled before this can be fully implemented.

Rotonen:
So, ready for review.

opengever/oneoffixx/api_client.py
```diff
@@ -0,0 +1,188 @@
...
+            raise OneoffixxConfigurationException('Oneoffixx configuration file missing!')
+        return creds_path
+
+    def get_credentials(self):
+        """Read the per deployment Oneoffixx backend credentials.
+
+        We've deemed hitting this one every time we need to fetch a new token
+        is OK.
+        """
+        if not self.credentials:
```
lukasgraf: This will actually cache credentials for the lifetime of the instance though, and never re-read a credentials file that's been changed on disk - is this intentional? I think a disk access on every re-auth is perfectly fine.

opengever/oneoffixx/api_client.py
```diff
@@ -0,0 +1,188 @@
...
+        """Get and blindly persist per user id the access token.
+
+        We will invalidate and refetch elsewhere if we have an invalid token.
+
+        The user id provides us a stable cache key to invalidate with.
+        """
+        url = '/'.join((self.get_oneoffixx_baseurl(), 'ids/connect/token', ))
+
+        credentials = self.get_credentials()
+        client_id = credentials.get('client_id')
```
lukasgraf: Maybe use `credentials[key]` for these instead of `.get()`? A `KeyError` during processing of the creds should be much more meaningful than whatever will happen down the line if `None` gets passed along.

opengever/oneoffixx/api_client.py
```diff
@@ -0,0 +1,188 @@
...
+
+        headers = {'Content-Type': 'application/x-www-form-urlencoded'}
+        try:
+            response = self.session.post(url, headers=headers, data=data)
+            response.raise_for_status()
+            access_token = response.json().get('access_token')
+            if access_token:
+                return access_token
+            else:
+                raise OneoffixxBackendException('Unable to fetch an OAUTH2 grant from Oneoffixx.', response)
```
lukasgraf: Might be helpful to log `response.text` for these (*all `OneoffixxBackendExceptions` basically*) with level `error` - otherwise we have not much more to go on than an HTTP status code.

Rotonen:
Still added the SID fetch support, per discussion with @buchi - can you double check?

Rotonen:
Still added whitelisting of content types and filename normalisation as we hit both in https://github.com/4teamwork/OfficeConnector/pull/299.

PR #4289, opened 2018-05-09
Add Subject index to solr.

phgross:
The Subject index from plone was missing in the solr schema, therefore the search on the Subject index, for example when clicking on a keyword on the dossier overview, returned a wrong results (all
objects).

Note: I wasn't sure if I should add an upgradestep or not, but for the smaller deployments it's definitely easier than manually triggering and for the big ones it's possible to patch it.


opengever/core/upgrades/20180509201121_reindex_subject_index_on_solr/upgrade.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def __call__(self):
+        solr_enabled = api.portal.get_registry_record(
+            name='use_solr', interface=ISearchSettings)
+        if not solr_enabled:
+            return
+
+        portal = api.portal.get()
+        solr_maintenance = queryMultiAdapter(
+            (portal, portal.REQUEST), name=u'solr-maintenance')
```
deiferni: does this work with other upgrade steps? the view calls `transaction.doom` in its `reindex` method, see https://github.com/4teamwork/ftw.solr/blob/1e7c7a0a3fd9ac27543b3f789585e473f5859b81/ftw/solr/browser/maintenance.py#L71. i'd say that causes problems.

lukasgraf:
@phgross PR on `ftw.solr` side that would allow for `solr_maintenance.reindex(idxs=['Subject'], doom=False)`: 4teamwork/ftw.solr#119

phgross:
@deiferni updated I use now the new keyword argument `doom` to avoid the transaction.doom() when using the `SolrMaintenanceView`

PR #4302, opened 2018-05-16
No longer require paragraph template.

deiferni:
This PR no longer requires the paragraph template for committees and their containers. It also removes the Exception that was raised in such a case.

It may not be required by customers, which is a valid state but uncommon. The system should not forbid it though as there is no reason to do so.


deiferni:
heh you're too fast 😉

Issue #4147, opened 2018-04-11
Tabbedview fails to load on document once on Firefox when created from a template


Rotonen:
1) Create a dossier
2) Create a document from a template
3) The tabbedview spinner spawns, spins and freezes

I suspect by symptom the tabbedview JS events are not handled properly in the new JS engine of Firefox, but I have no hard proof on that.

See https://extranet.4teamwork.ch/support/gever-ai/tracker/299 for a video demo.

bierik:
I could not get any errors in firefox after reproducing the issue. But there is an extjs specific warning in the console: `unreachable code after return statement`.
It would take me about 2 days for debugging the tabbedview code because I have to figure out what's going on behind the scenes of tabbedview when it tried to load more content.

@phgross Shall I spend these 2 days and how critical is this issue?

deiferni:
as just discussed, if i understood correctly this seems urgent as it has the potential to break firefox compatibility in the future. should attempt to tackle this in `2018.3` if possible 🤔.

bierik:
As discussed with @phgross I'm gonna spend half a day again on this issue.

bierik:
Triggering the `oc` URL-handler on tabbed-view load should fix the problem ->
https://github.com/4teamwork/opengever.core/blob/7250392ff29c8417d894cc070ebd1abb1d840b6e/opengever/base/redirector.py#L94-L104

Rotonen:
The same issue also freezes javascript execution on a document after clicking on a checkout (or attach) file action button. This is a minor usability issue for checkouts as they should poll the document status and reload-on-lock.

phgross:
@bierik bitte umsetzen oder mit @Rotonen besprechen was nun gemacht werden muss.

bierik:
Fixed through https://github.com/4teamwork/opengever.core/pull/4349

PR #4349, opened 2018-05-24
Fix stuck tabbedview spinner on firefox

bierik:
See https://github.com/4teamwork/opengever.core/issues/4147

The `reload` event is fired after the tabbedview has been loaded. So open the oc url after this event happened.


bierik:
@Rotonen Updated

PR #4351, opened 2018-05-24
Fix resolving tasks when original was delegated

### LINKED ISSUE #4230: TaskTransitionController.all_subtasks_finished does not handle successor pairs.
The current [`all_subtasks_finished`](https://github.com/4teamwork/opengever.core/blob/master/opengever/task/browser/transitioncontroller.py#L565) check in the transition controller, currently only checks for subtasks of the current task but it should also check for subtasks on the successor or predecessor if there are some. 

Otherwise it can lead to confusing failures when trying to resolve a task on the predecessor site, with subtasks on the successor site. 

https://extranet.4teamwork.ch/extern/opengever-kanton-zug/sprint-backlog/20377

tarnap:
When working with multi admin setups, tasks may span over several admin
units.

We assume to have two admin units: FD and RK

Anton creates a task in FD and gives it to Berta in RK
At the same time, Anton delegates the task to Carlos and David in FD
When Berta tries to resolve the task, the subtasks delegated to Carlos
and David should be checked for completeness as well.

Therefore the check on Berta's side should get all the subtasks from
Berta's and Anton's task.

Resolves #4230

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -10,6 +10,7 @@ Changelog
 - Normalize query strings of template filter. [Kevin Bieri]
 - Improve error handling on the @scan-in API endpoint. [Rotonen]
 - Fix missing translation for committee filter. [tarnap]
+- Fix resolving multi admin unit tasks when subtasks for original task exist. [tarnap]
 - Bumblebee bugfix: no longer defer preview for new documents. [deiferni]
 - Display dossier resolve properties in the config view. [tarnap]
 - Add a tasktemplatefolder to GEVER examplecontent. [phgross]
--- a/opengever/globalindex/model/query.py
+++ b/opengever/globalindex/model/query.py
@@ -3,6 +3,8 @@ from opengever.globalindex.model.task import Task
 from opengever.ogds.base.utils import get_current_admin_unit
 from opengever.ogds.models.query import BaseQuery
 from plone import api
+from sqlalchemy import and_
+from sqlalchemy import or_
 
 
 class TaskQuery(BaseQuery):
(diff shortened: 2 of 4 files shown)


opengever/task/tests/test_transitioncontroller_integration.py
```diff
@@ -0,0 +1,35 @@
...
+from ftw.builder import Builder
+from ftw.builder import create
+from opengever.task.browser.transitioncontroller import TaskChecker
+from opengever.task.interfaces import ISuccessorTaskController
+from opengever.testing import IntegrationTestCase
+
+
+class TaskTransitionController(IntegrationTestCase):
+
+    def test_task_cannot_be_resolved_if_predecessor_task_has_subtasks(self):
```
phgross: the name is wrong ... you not asserting that ...

opengever/testing/fixtures.py
```diff
@@ -809,6 +809,34 @@ def create_treaty_dossiers(self):
             .relate_to(self.document)
             ))
 
+        self.predecessor_task = self.register('predecessor_task', create(
```
phgross: i would not add two separate tasks to the fixtures for this special use case, the fixture should stay slim. Use the existing tasks instead and link them via the helper `register_successor` from IntegrationTestCase

opengever/task/tests/test_transitioncontroller_integration.py
phgross: as already discussed, I wouldn't add a specific test for this use case, but add an additional test_case to the all_subtasks_finished tests (https://github.com/4teamwork/opengever.core/blob/master/opengever/task/tests/test_task_checker.py#L80)

tarnap:
@phgross: please re-review 😄

tarnap:
@phgross: Made the changes in another pull request with a better description.

PR #4361, opened 2018-05-28
Fix resolving tasks if original task was delegated

### LINKED ISSUE #4230: TaskTransitionController.all_subtasks_finished does not handle successor pairs.

tarnap:
## Assuming

1. Anton created a successor task for Berta
2. Anton delegated the task to Carlos

```
FD                   RK

              1.
Task (Anton) +-----> Successor Task (Berta)
 +
 | 2.
 | 
 v
Delegated Task (Carlos)

```

When Berta tries to resolve the task, the status of the subtask delegated to Carlos should be checked as well.

Therefore the check on Berta's side should get all the subtasks from
Berta's and Anton's task.

Resolves #4230 
Closes #4351

### DIFF
--- a/opengever/task/tests/test_task_checker.py
+++ b/opengever/task/tests/test_task_checker.py
@@ -1,199 +1,132 @@
 from datetime import date
 from ftw.builder import Builder
 from ftw.builder import create
+from opengever.ogds.base.utils import get_current_org_unit
 from opengever.task.browser.transitioncontroller import Checker
 from opengever.task.browser.transitioncontroller import get_checker
 from opengever.task.browser.transitioncontroller import TaskChecker
 from opengever.task.interfaces import ISuccessorTaskController
-from opengever.testing import FunctionalTestCase
+from opengever.testing import IntegrationTestCase
 from plone.app.testing import TEST_USER_ID
-import transaction
+from unittest import skip
 
 
-class TestTaskControllerChecker(FunctionalTestCase):
-
-    def setUp(self):
-        super(TestTaskControllerChecker, self).setUp()
-
-        self.hugo = create(Builder('ogds_user')
-                           .having(userid='hugo.boss',
-                                   firstname='Hugo',
-                                   lastname='Boss',
-                                   email='hugo@boss.ch')
-                           .in_group(self.org_unit.users_group))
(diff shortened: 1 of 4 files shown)


tarnap:
@phgross: This PR is ready for review

opengever/task/tests/test_task_checker.py
```diff
@@ -1,199 +1,132 @@
...
-
-        task1 = create(Builder('task').having(responsible_client='client1'))
-
+        self.login(self.secretariat_user)
         self.assertTrue(
-            get_checker(task1).current_user.in_responsible_orgunits_inbox_group)
+            get_checker(self.task).current_user.in_responsible_orgunits_inbox_group)
 
-        checker = Checker(task1.get_sql_object(), self.request, self.hugo)
+        checker = Checker(self.task.get_sql_object(), self.request, self.secretariat_user)
```
phgross: I would add a short comment that one check is with `agency` and one without ...

opengever/task/tests/test_task_checker.py
```diff
@@ -1,199 +1,132 @@
...
-            .successor_from(predecessor)
-            .in_state('task-state-in-progress')
-            )
+        self.set_workflow_state('task-state-in-progress', self.subtask)
+        self.set_workflow_state('task-state-in-progress', self.archive_task)
+        self.register_successor(self.task, self.archive_task)
 
         sql_successor_task = ISuccessorTaskController(
-            predecessor).get_successors()[0]
+            self.task).get_successors()[0]
```
phgross: fetching the sql_successor_task should not be necessary its `self.archive_task`

tarnap:
@phgross: please re-re-re-review this PR 🙈

PR #4364, opened 2018-05-28
OGIP 26: Implement process handling for sequential task

### LINKED ISSUE #4175: Detailansicht von Aufgaben von sequentiellen Abläufen
Ansicht der Unteraufgabe anpassen: vordere und nächste Aufgabe des Sequentiellen Ablaufs ollen ersichtlich sein.

phgross:
- This PR links tasks generated by triggering a sequential tasktemplatefolder together, every task stores a reference to the predecessor in the sequential process. The reference is stored on the object itself, but also synced to the globalindex.
- Shows previous and next task on the task view:
bildschirmfoto 2018-05-28 um 13 56 00
- Automatically start next task when a task gets resolved
- Do not send notifications for planned tasks, but send them if they are opened by resolving the predecessor. To use the same type of activity we reword the TaskAddedActivity from "task added" to "task opened".

Closes part of #4182 
Closes #4175


phgross:
@lukasgraf I updated my PR with the following changes: 
 - Instead of saving the order of the subtask in a chain of linked subtasks, I store now a list of oguids in the annotations of a task. By the way, I had also tried to use the already existing "position in parent" list, but that can be confusing when the main task contains other objects than tasktemplate-tasks  therefore I preferred to use an explicit list.
- made the tasktemplate_predecessor relationship 1:1 instead of 1:n
- As already discussed, there is no clean way to by pass the security checks when doing a transition, therefore we still protect the transition with a request marker-interface. To makes sure the interface gets removed as soon the transition is done, I add an contextmanager `as_internal_workflow_transition` to run such transitions.

I've also checked if it would make sense to use plone contentrules for the "automatic start of the next task", but I decided against it:
 - I see currently no value which would to bring us plone contentrules, rather we need to make sure, all  the contentrule stuff is not visible for other users.
 - Similar functionality (disposition) is also done with event handlers
 - Event handlers is way more explicit, especially when we cleanup and merge together the event handlers.

lukasgraf:
Looks good to me 👍

PR #4319, opened 2018-05-17
OGIP 26: Reset responsible to issuer on task rejection

Rotonen:
* Ported some task transition action tests to integration tests where trivial
* Reset task responsible to issuer upon task rejection
  * Also remove the activity subscription of the previous responsible

Closes #4177

PR #4365, opened 2018-05-28
Fixups for missed code review points on #4319

Rotonen:
Continuing from: https://github.com/4teamwork/opengever.core/pull/4319

My bad, sorry.

### DIFF
--- a/opengever/task/handlers.py
+++ b/opengever/task/handlers.py
@@ -1,6 +1,7 @@
 from Acquisition import aq_inner, aq_parent
 from datetime import date
 from opengever.activity import notification_center
+from opengever.activity.roles import TASK_ISSUER_ROLE
 from opengever.activity.roles import TASK_RESPONSIBLE_ROLE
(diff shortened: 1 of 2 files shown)


opengever/task/handlers.py
```diff
@@ -76,6 +77,8 @@ def reassign_team_tasks(task, event):
 def set_responsible_to_issuer_on_reject(task, event):
     if event.action == 'task-transition-open-rejected':
         notification_center().remove_watcher_from_resource(task.oguid, task.responsible, TASK_RESPONSIBLE_ROLE)
+        notification_center().add_watcher_to_resource(task.oguid, task.issuer, TASK_ISSUER_ROLE)
```
phgross: That's not necessary the issuer is already registered ...

Rotonen: That's not guaranteed as the user can have unsubscribed. I figured this should be enforced upon rejection.

PR #4380, opened 2018-05-30
Display tooltip on favorite action

### LINKED ISSUE #4353: Tooltip of favorites icon does not appear
**Expected:**
Tooltip on-hover as specified in #4103
bildschirmfoto 2018-03-13 um 09 41 29

**Got:**
Tooltip does not appear.

tarnap:
Display the tooltip and javascript toggle behaviour.

tooltip

Resolves #4353

### DIFF
--- a/opengever/base/browser/resources/Favorites.js
+++ b/opengever/base/browser/resources/Favorites.js
@@ -13,8 +13,10 @@
       }
 
       if ($('#mark-as-favorite').hasClass('is-favorite')){
+        target.attr('title', target.data('markFavoriteTitle'));
         this.delete_favorite(target);
       } else {
+        target.attr('title', target.data('unmarkFavoriteTitle'));
         this.add_favorite(target);
       }
     };
--- a/opengever/base/viewlets/favorite_action.pt
+++ b/opengever/base/viewlets/favorite_action.pt
@@ -1,11 +1,17 @@
 <div id="favorite-action"
+     i18n:domain="opengever.base"
      tal:condition="view/available"
      tal:define="favorite view/get_favorite"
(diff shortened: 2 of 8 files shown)


opengever/base/viewlets/favorite_action.pt
```diff
@@ -1,11 +1,30 @@
...
+     i18n:domain="opengever.base"
      tal:condition="view/available"
      tal:define="favorite view/get_favorite"
      tal:attributes="data-oguid python: view.oguid().id;
                      data-url view/url">
 
-  <a href="#" id="mark-as-favorite"
-     tal:attributes="class python:'is-favorite' if favorite else '';
-                     data-favorite-id python: favorite and favorite.favorite_id" />
+  <a href="#"
```
phgross: If we have to switch the title attribute by javascript anyways, i would initialize it also via JS. So no second link in the template is necessary

opengever/base/browser/resources/Favorites.js
```diff
@@ -13,9 +13,13 @@
       }
 
       if ($('#mark-as-favorite').hasClass('is-favorite')){
+        var title = target.data('markFavoriteTitle')
```
phgross: `;` is missing, but i would prefer a oneliner:

```javascript
target.attr('title', target.data('markFavoriteTitle'));
```

opengever/base/viewlets/favorite_action.pt
tarnap: Took me a while, but managed to do it with one single tag.
Had to move the translations to the viewlet (recommended by @deiferni)

@phgross: Thank you for the review

PR #4385, opened 2018-05-30
Add today in the right format to the sablon data

### LINKED ISSUE #4216: Add `today` to sablon fields
The date of generation is required in the agenda item list.
Since date fields are not updated when the sablon process is run, the current date needs to be passed as additional field.

The date output format needs to be determined.

## Please read discussion:
https://basecamp.com/2768704/projects/13732725/todos/339163209#comment_614401493

tarnap:
In order to display the protocol generation data, the date needs to be passed inside the data.
This adds a formatted `today`.

Resolves #4216

### DIFF
--- a/docs/public/admin-manual/meeting/mergefields.rst
+++ b/docs/public/admin-manual/meeting/mergefields.rst
@@ -15,6 +15,13 @@ Worddokument zu erzeugen.
 
 Die folgenden Seriendruckfelder können standardmässig verwendet werden:
 
+Generelle Metadaten:
+~~~~~~~~~~~~~~~~~~~~
+
+- ``document.generated``
+
+  Generierungsdatum des Dokuments (String)
+
 Metadaten zur Sitzung:
 ~~~~~~~~~~~~~~~~~~~~~~
 
--- a/opengever/meeting/protocol.py
+++ b/opengever/meeting/protocol.py
@@ -1,6 +1,8 @@
+from datetime import datetime
 from opengever.meeting import _
 from opengever.meeting.model.membership import Membership
 from opengever.ogds.base.utils import get_current_admin_unit
+from plone import api
 from zope.globalrequest import getRequest
 from zope.i18n import translate
 import json
@@ -36,6 +38,7 @@ class ProtocolData(object):
         self.add_meeting()
         self.add_commitee()
         self.add_agenda_items()
+        self.add_general_metadata()
 
     def add_settings(self):
         if not self.meeting.protocol_start_page_number:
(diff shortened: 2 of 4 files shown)


opengever/meeting/protocol.py
```diff
@@ -145,6 +148,10 @@ def add_agenda_items(self):
                     include_copy_for_attention=self.include_copy_for_attention
                 ))
 
+    def add_general_metadata(self):
+        self.data['today'] = api.portal.get_localized_time(
```
deiferni: - should we namespace this `document.` 🤔?
- i'd prefer `modification_date` to `today`.

tarnap: That's much better naming!

tarnap:
@deiferni: I changed the naming of the field. I also did add the documentation to the admin docs.

docs/public/admin-manual/meeting/mergefields.rst
```diff
@@ -15,6 +15,13 @@ Worddokument zu erzeugen.
 
 Die folgenden Seriendruckfelder können standardmässig verwendet werden:
 
+Generelle Metadaten:
+~~~~~~~~~~~~~~~~~~~~
+
+- ``document.modified``
+
+  Erstellungsdatum des Dokuments (String)
```
deiferni: hmm `Erstellungsdatum` sounds like `created`, this should be `Modifikationsdatum` or something like that, no?

docs/public/admin-manual/meeting/mergefields.rst
```diff
@@ -15,6 +15,13 @@ Worddokument zu erzeugen.
 
 Die folgenden Seriendruckfelder können standardmässig verwendet werden:
 
+Generelle Metadaten:
+~~~~~~~~~~~~~~~~~~~~
+
+- ``document.modified``
```
deiferni: hah just had a spontaneous flash of inspiration: call this `generated`.

`modified` is a thing that is used by plone already but which will contain a different timestamp than what we provide with `generated` as  `datetime.today`. that way we don't have a naming conflict and we don't confuse developers/users should we decide to also expose the `modified` and `created` fields that are tracked by plone to sablon

tarnap:
@deiferni: please re-review

opengever/meeting/protocol.py
```diff
@@ -145,6 +148,10 @@ def add_agenda_items(self):
                     include_copy_for_attention=self.include_copy_for_attention
                 ))
 
+    def add_general_metadata(self):
+        self.data['document.generated'] = api.portal.get_localized_time(
```
deiferni: oh almost missed that one. these should be nested dicts for the namespace 😓.

tarnap:
@deiferni: please rereview

PR #4391, opened 2018-05-31
Implement recently touched menu

lukasgraf:
This implements a new menu that lists **checked out documents** as well as **recently touched** documents in a combined drop-down.

recently_touched

---

The upper part lists the user's **currently checked out documents** - all of them. If the list of checked out documents gets larger than 5, this group gets a scrollbar, and is individually scrollable.

The **number of currently checked out documents** is displayed as a **badge counter** in the same orange that's already being used for the meta-icon indicating that a document is checked out.

recently_touched_many_checked_out_transp

The bottom part lists the **recently touched documents**. Any document that is already listed in the upper part because it's checked out (by the current user), won't be listed again in the list of recently touched objects. This list is limited to 10 items by default (configurable via registry setting).

---

If no documents are currently checked out, a placeholder is displayed:

recently_touched_none_checked_out_transp


opengever/base/touched.py
```diff
@@ -0,0 +1,112 @@
...
+RECENTLY_TOUCHED_KEY = 'opengever.base.touched.recently_touched'
+RECENTLY_TOUCHED_INTERFACES_TO_TRACK = [
+    IBaseDocument,
+]
+
+logger = logging.getLogger('opengever.base.touched')
+
+
+class IObjectTouchedEvent(IObjectEvent):
+    """An object has been touched"""
```
lukasgraf: FYI: I considered splitting out the interface and the event to `opengever.base.interfaces` / `opengever.base.events` respectively. However, we don't yet have an `events.py`, and no event interfaces yet in `interfaces.py`. So since it made rebasing much easier, and because having everything in one file `touched.py` has its own advantages, I left it like that for now. Not sure though.

opengever/base/viewlets/recently_touched_menu.html
```diff
@@ -0,0 +1,54 @@
...
+                  <li v-for="entry in checkedOut">
+
+                    <div class="item-content">
+                      <a class="document-link" :class="entry.icon_class" :href="entry.target_url">{{entry.title}}</a>
+                      <span class="item-timestamp timeago relativetime discreet" :title="entry.last_touched">{{entry.last_touched}}</span>
+                    </div>
+
+                  </li>
+
+                  <li v-show="checkedOut.length == 0">
```
bierik: Always use three `=` for equality checks.

opengever/base/viewlets/recently_touched_menu.html
```diff
@@ -0,0 +1,54 @@
...
+                  <li v-for="entry in recentlyTouched">
+
+                    <div class="item-content">
+                      <a class="document-link" :class="entry.icon_class" :href="entry.target_url">{{entry.title}}</a>
+                      <span class="item-timestamp timeago relativetime discreet" :title="entry.last_touched">{{entry.last_touched}}</span>
+                    </div>
+
+                  </li>
+
+                  <li v-show="recentlyTouched.length == 0">
```
bierik: Always use three `=` for equality checks.

opengever/base/browser/resources/recently_touched_menu_app.js
```diff
@@ -0,0 +1,76 @@
...
+      userid: null,
+      i18n: {},
+      recentlyTouched: [],
+      checkedOut: [],
+      isOpen: false,
+    },
+
+    beforeMount: function () {
+      if (!this.$el.attributes['data-userid']){
+        return;
```
bierik: When is the userid not set on the template?
And does `return;` really stops the execution of the vue component?

opengever/base/browser/resources/recently_touched_menu_app.js
```diff
@@ -0,0 +1,76 @@
...
+
+      var requester = axios.create();
+      requester.defaults.headers.common['Accept'] = 'application/json';
+      requester.defaults.headers.common['Content-Type'] = 'application/json';
+      this.requester = requester;
+    },
+
+    updated: function () {
+      // Update relative times after DOM has been updated by Vue
+      $(".timeago").timeago();
```
bierik: This updates every element that matches `.timeago` selector which is not necessary, right?
You could probably select the elements by using `#recently-touched-vue-app .timeago`.

opengever/base/browser/resources/recently_touched_menu_app.js
```diff
@@ -0,0 +1,76 @@
...
+    },
+
+    computed: {
+      messenger: MessageFactory.getInstance
```
bierik: The messenger does not have to be a computed property here. You could generate the instance when the app is being mounted.

lukasgraf: @bierik copied from the [favorites app](https://github.com/4teamwork/opengever.core/blob/master/opengever/base/browser/resources/favorites_menu_app.js#L34-L36) - should that also be updated?

opengever/base/browser/resources/recently_touched_menu_app.js
lukasgraf: Good point, thanks! 👍

opengever/base/browser/resources/recently_touched_menu_app.js
lukasgraf: @bierik copied from the [favorites menu app](https://github.com/4teamwork/opengever.core/blob/master/opengever/base/browser/resources/favorites_menu_app.js#L17-L20) - I would assume it acts as a safeguard so it's not the entire JavaScript execution that fails in that case, just the Vue app? Might be paranoia code though.

opengever/base/viewlets/recently_touched_menu.html
lukasgraf: rage

lukasgraf:
@bierik updated with your feedback.

docs/HISTORY.txt
```diff
@@ -5,6 +5,8 @@ Changelog
 2018.4.0 (unreleased)
 ---------------------
 
+- Implement recently touched menu that lists checked out documents and recently touched objects. [lgraf]
+- Team add portal action permissions: Administrators to add. [Rotonen]
```
phgross: ✂️ I guess, something went wrong while rebasing ...

opengever/base/tests/test_object_touched_handling.py
```diff
@@ -0,0 +1,193 @@
...
+        recently_touched_log = self._get_log(self.regular_user)
+
+        limit = api.portal.get_registry_record(
+            'limit', IRecentlyTouchedSettings)
+
+        for i in range(limit * 3):
+            recently_touched_log.append(
+                {'last_touched': datetime.now(), 'uid': str(i)})
+
+        getMultiAdapter((self.subdocument, self.request),
```
phgross: you can use integrationtests `checkout_document`

opengever/base/interfaces.py
```diff
@@ -335,3 +335,12 @@ class IFavoritesSettings(Interface):
         title=u'Enable favorites feature',
         description=u'Whether favorite feature is enabled',
         default=True)
+
+
+class IRecentlyTouchedSettings(Interface):
+
+    limit = schema.Int(
```
phgross: should we provide this setting also in the `@settings` endpoint?

opengever/base/tests/test_object_touched_triggering.py
```diff
@@ -0,0 +1,43 @@
...
+        event = get_last_recorded_event()
+        self.assertTrue(IObjectTouchedEvent.providedBy(event))
+
+    def test_gets_fired_on_checkout(self):
+        self.login(self.regular_user)
+        manager = getMultiAdapter((self.document, self.request),
+                                  ICheckinCheckoutManager)
+
+        register_event_recorder(IObjectTouchedEvent)
+        manager.checkout()
```
phgross: self.checkout_document(self.document)

opengever/base/handlers.py
```diff
@@ -15,6 +15,8 @@
 
 
 def object_moved_or_added(context, event):
+    notify(ObjectTouchedEvent(context))
```
phgross: Hmm not sure about that 🤔 
Technically it's correct, but if a user move a dossier containing more than 10 documents to an other repositoryfolder, the complete list is replaced with those moved documents ...

opengever/base/interfaces.py
lukasgraf: @phgross should be covered in 87f3fb33

docs/HISTORY.txt
lukasgraf: Ah yes, this is an artifact of [updating the changelog on master after creating the hotfix release `2018.3.2`](https://github.com/4teamwork/opengever.core/pull/4393) - fixed 👍

opengever/base/handlers.py
lukasgraf: Good point! I think this might require some more discussion / thinking.

lukasgraf:
@phgross updated with your feedback and rebased

opengever/base/handlers.py
lukasgraf: @phgross as discussed: Changed so moves don't fire `ObjectTouchedEvent` any more.

lukasgraf:
@phgross updated - the most recently commit changes it so `ObjectMovedEvent` and `ObjectRemovedEvent` aren't considered "touched" anymore, only `ObjectAddedEvent` *(and `ObjectModifiedEvent` of course)*.

opengever/base/browser/resources/recently_touched_menu_app.js
```diff
@@ -0,0 +1,73 @@
...
+        return this.fetchData()
+          .then(function() {
+            this.isOpen = true;
+          }.bind(this));
+      },
+      close: function() {
+        this.isOpen = false;
+      },
+      fetchData: function () {
+        return this.requester.get(this.endpoint)
```
phgross: @lukasgraf i forgot to mentioned that on my last review: But this request, will be cached by IE 11 (see https://github.com/4teamwork/opengever.core/pull/4284) for more details.

lukasgraf: @phgross good point, thanks. That actualy proved to be an issue in IE11 - fixed, and tested again in IE 11 and Edge. 👍

PR #4405, opened 2018-06-05
Add Comments to proposals

### LINKED ISSUE #2181: Kommentarfunktion zu Anträgen
⚠️ Sollte zusammen mit https://github.com/4teamwork/opengever.core/issues/4186 erledigt werden.

Der Kunde wünscht die Möglichkeit einen Antrag vor der Traktandierung zu reviewen und zu kommentieren.

Lösungsvorschlag: Kommentarfunktion mit Separatem Formular zum Kommentieren anbieten. Kommentare im Verlauf einblenden.

screen shot 2016-09-22 at 14 21 33

Ref: https://basecamp.com/2768704/projects/13732725/todos/340577216

### LINKED ISSUE #4186: Kommentarfeld bei Statusänderungen von Anträgen
⚠️ Sollte zusammen mit https://github.com/4teamwork/opengever.core/issues/2181 erledigt werden.

Ändert man den Status eines Antrags (z.B. Einreichen) passiert dies meist direkt (Ausser beim Zurückweisen). Es besteht der Wunsch, dass bei Statusänderungen ein Kommentar angegeben werden kann (analog Aufgaben).

screen shot 2018-04-18 at 16 58 38

Ref https://basecamp.com/2768704/projects/13732725/todos/340577216.

njohner:
The stand was that you could add a comment when rejecting a submitted proposal but not otherwise. Now one can add a comment for any action on a `Proposal` or `SubmittedProposal` and there is a new action to simply comment a proposal, without changing its state.

I had to add new permissions to allow commenting a `Proposal` once it is submitted (in which case it can't normally be modified). This is done on the `dossier` workflow, as we only allow commenting a proposal if the containing dossier is open.

Comments made on a `SubmittedProposal` either on the dossier side or on the committee side will appear in the history on both sides, so can be used to share information on the proposal.

Summary of the changes:
* Actions on proposals now redirect to a form used to add a comment.
* The reject action form was deleted and this action is now treated as all the others
* New action to add comments on proposals. We use the newly defined `CommentProposalCommand`s to dispatch a request to the `CommentProposal` view to sync add the comment on the second proposal object (`Proposal` or `SubmittedProposal`)
* New permissions to allow commenting a submitted proposal.
* Use same icon for comments as for `Task`

**New comment action available**
screen shot 2018-06-05 at 08 57 45

**When executing an action, we get a form to add a comment**
screen shot 2018-06-05 at 08 58 44

**Once submitted the proposal can still be commented from the dossier**
screen shot 2018-06-05 at 08 59 17

**History on both sides shows the comments made from the dossier and from the committee**
**here seen from the committee**
screen shot 2018-06-05 at 09 00 52

**here seen from the dossier**
screen shot 2018-06-05 at 09 01 04

**When cancelled the proposal can still be commented. This also shows a fill history with comments on different actions, from the committee and from the dossier**
screen shot 2018-06-05 at 09 02 36

resolves #2181 
resolves #4186


opengever/meeting/tests/test_proposalhistory.py
```diff
@@ -101,28 +128,69 @@ def test_rejecting_proposals_creates_history_entries(self, browser):
...
+        browser.fill({'Comment': comment}).submit()
+        with self.login(self.meeting_user, browser):
+            self.assert_proposal_history_records(
+                u'Proposal commented by B\xe4rfuss K\xe4thi (kathi.barfuss)',
+                self.proposal,
+                browser,
+                expected_comments=comment,
+                with_submitted=True
+                )
+
```
deiferni: could you add add a test where i comment on a decided submitted proposal that is inside a closed dossier. this currently works but it may be something we mess up in the future if there is no test 😉.

opengever/core/profiles/default/workflows/opengever_proposal_workflow/definition.xml
```diff
@@ -12,6 +12,7 @@
   <permission>opengever.document: Checkin</permission>
   <permission>opengever.document: Checkout</permission>
   <permission>opengever.document: Force Checkin</permission>
+  <permission>opengever.meeting: Add Proposal Comment</permission>
```
deiferni: the permission should not be managed by the proposal workflow, but only by the dossier workflow. hint: you have configured it to acquire from the parent in all states, so it does not add anything here 😄. best leave it out entirely.

opengever/core/upgrades/20180604105148_add_proposal_comment_permission/upgrade.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class AddProposalCommentPermission(UpgradeStep):
+    """Add proposal comment permission.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
+        self.update_workflow_security(
+            ['opengever_dossier_workflow', 'opengever_proposal_workflow'],
```
deiferni: if you remove the permission, `opengever_proposal_workflow` should be removed here, too.

opengever/core/lawgiver.zcml
```diff
@@ -51,6 +51,7 @@
                    opengever.meeting: Add Member,
                    opengever.meeting: Add Proposal Template,
                    opengever.meeting: Add Proposal,
+                   opengever.meeting: Add Proposal Comment,
```
deiferni: might be easier to just ignore the permission for now. it is not used inside lawgiver-ed content at this moment. that way you don't have to update all the workflows, that do not use this permission.

opengever/meeting/browser/templates/macros.pt
```diff
@@ -35,6 +35,36 @@
...
+                </a>
+              </li>
+            </tal:block>
+            <li>
+              <a tal:condition="view/is_comment_allowed" tal:attributes="href view/comment_url"
+                 id="comment_proposal" class="proposalCommented"
+                 title="Comment"
+                 i18n:attributes="title">
+                  <span i18n:translate="" class="subMenuTitle actionText">
+                    Comment
```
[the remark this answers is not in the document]
njohner: done

opengever/meeting/proposal_transition_comment.py
```diff
@@ -0,0 +1,114 @@
...
+                             name='cancel', )
+    def handleCancel(self, action):
+        return self.request.RESPONSE.redirect('.')
+
+    @button.buttonAndHandler(_(u'save', default='Save'),
+                             name='save')
+    def handleSubmit(self, action):
+        data, errors = self.extractData()
+        if errors:
+            errorMessage = '<ul>'
```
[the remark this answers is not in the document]
njohner: I did not know about `self.formErrorsMessage`. I copied the error handling form the `RejectForm`

opengever/meeting/proposal_transition_comment.py
```diff
@@ -0,0 +1,114 @@
...
+            for error in errors:
+                if errorMessage.find(error.message):
+                    errorMessage += '<li>' + error.message + '</li>'
+            errorMessage += '</ul>'
+            self.status = errorMessage
+            return None
```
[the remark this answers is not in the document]
njohner: done

opengever/meeting/proposal_transition_comment.py
```diff
@@ -0,0 +1,114 @@
...
+
+        else:
```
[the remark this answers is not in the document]
njohner: done

opengever/meeting/proposalhistory.py
```diff
@@ -190,6 +189,17 @@ def message(self):
                  mapping={'user': self.get_actor_link()})
 
 
+@ProposalHistory.register
+class ProposalCommented(BaseHistoryRecord):
+
+    history_type = u'commented'
+
```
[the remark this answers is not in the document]
njohner: Yes right, done

opengever/meeting/tests/test_proposalhistory.py
njohner: done

njohner:
Ok everything is done here. Ready for re-review

opengever/meeting/model/proposal.py
```diff
@@ -34,7 +34,7 @@
 
 class Submit(Transition):
 
-    def execute(self, obj, model):
+    def execute(self, obj, model, text=None, **kwargs):
```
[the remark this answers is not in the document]
deiferni: Hm i'd say it is ok if the subclass adds more (optional) parameters that cannot be passed to the parent, this already happens with the keyword argument `text`. I don't see the benefit of the `**kwargs` here 🤔.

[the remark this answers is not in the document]
njohner: Ok fine with me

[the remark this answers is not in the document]
njohner: Ok not fine. As a memo, here is the reason behind it.
The chain of calls to execute a transition TTW is:
* `ProposalOverview` points to the `ProposalTransitionCommentAddFormView`
* Submitting the form calls `ProposalTransitionController.execute_transition`
* calls `Proposal.execute_transition`
* Which calls `Workflow.execute_transition`
* Which calls `Transition.execute`

The new `text` argument has to be handled by the whole chain, but `Workflow` and `Transition` is used by other objects than `Proposal`, so I added `**kwargs` to the workflow, which then in turn has to be handled by `Transition`.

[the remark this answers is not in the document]
deiferni: 😄 as discussed, that makes sense and was not obvious from the diff. in that case ready to 🚢!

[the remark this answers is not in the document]
njohner: I added some functional tests with comments to make sure this does not get broken later on.

opengever/core/profiles/default/workflows/opengever_proposal_workflow/definition.xml
```diff
@@ -53,6 +53,10 @@
         acquired="False"
         name="opengever.trash: Untrash content"
         />
+    <permission-map
```
deiferni: this is a left-over, the permission is no longer managed: ✂️

opengever/core/profiles/default/workflows/opengever_proposal_workflow/definition.xml
```diff
@@ -94,6 +98,10 @@
...
```
deiferni: ✂️

opengever/core/upgrades/20180604105148_add_proposal_comment_permission/upgrade.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class AddProposalCommentPermission(UpgradeStep):
+    """Add proposal comment permission.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
+        self.update_workflow_security(
+            ['opengever_dossier_workflow', 'opengever_proposal_workflow',
```
deiferni: only `opengever_dossier_workflow` should have changed now 🤔

opengever/meeting/proposal_transition_comment.py
```diff
@@ -0,0 +1,148 @@
...
+from zope import schema
+from zope.interface import Interface
+from zope.interface import provider
+from zope.i18n import translate
+from zope.schema.interfaces import IContextSourceBinder
+from zope.schema.vocabulary import SimpleVocabulary
+
+
+@provider(IContextSourceBinder)
+def getTransitionVocab(context):
```
deiferni: snake case please

versions.cfg
```diff
@@ -87,7 +87,7 @@ ftw.keywordwidget = 1.5.0
 ftw.lawgiver = 1.14.1
 ftw.mail = 2.5.2
 ftw.maintenanceserver = 1.1.2
-ftw.mobilenavigation = 1.4.7
+ftw.mobilenavigation = 1.4.8
```
deiferni: ???

opengever/meeting/proposalhistory.py
```diff
@@ -65,19 +66,28 @@ def append_record(self, history_type, timestamp=None, **kwargs):
         record.append_to(history)
 
         if record.needs_syncing:
-            # currently we only sync from the submitted side
-            # to the dossier
-            path = self.context.load_model().physical_path
+            if isinstance(self.context, SubmittedProposal):
```
deiferni: instance checks are always a last resort. if you can avoid it, do so!

here it would be better to add the same method to both objects proposal and submitted proposal. submitted proposal already has one `get_source_admin_unit_id`, which is only used here. id would propose to rename this to `get_sync_target_admin_unit_id` and add a method for the path, maybe `get_sync_target_path` 🤔.

opengever/meeting/tests/test_unit_workflow.py
```diff
@@ -9,7 +9,21 @@ class SomethingWithWorkflow(object):
 
     def __init__(self, initial_state):
         self.workflow_state = initial_state
+        self.history = []
 
+class Retract(Transition):
```
deiferni: the newlines here are f...ed up. did your linter stop working?

opengever/meeting/tests/test_unit_workflow.py
```diff
@@ -9,7 +9,21 @@ class SomethingWithWorkflow(object):
...
+
+    def __init__(self):
+        super(Retract, self).__init__('published', 'pending')
+
+    def execute(self, obj, model, reason=None, **kwargs):
+        if not self.can_execute(model):
```
deiferni: shouldn't this be a super call?

opengever/meeting/browser/proposaltransitions.py
```diff
@@ -2,95 +2,50 @@
...
-from z3c.form.field import Fields
-from z3c.form.form import button
-from z3c.form.form import Form
 from zExceptions import NotFound
-from zope import schema
-from zope.interface import Interface
 
 
-class ProposalTransitionController(BrowserView):
+class ProposalTransitionController(object):
```
deiferni: `ProposalTransitionController` does form stuff, i would merge that class into the form `ProposalTransitionCommentAddForm`. The responsibility of each class is not clear to me 🤔.

opengever/meeting/proposal.py
```diff
@@ -600,3 +609,7 @@ def reject(self):
         self.date_of_submission = None
         api.content.transition(obj=self,
                                transition='proposal-transition-reject')
+
+    def is_submitted(self):
+        model = self.load_model()
+        return bool(model.submitted_physical_path)
```
deiferni: we have a workflow state `submitted`. i would delegate this question to the model, i.e. add a `is_submitted` method also to model, check if workflow state is `submitted` `scheduled` or `decided`

njohner: I had thought about that, but had opted for this solution to avoid having the `is_submitted` method fail if a new state is added in the future... but anyway, as the method is defined on the model, it should be obvious enough not to get forgotten...

opengever/meeting/browser/proposaltransitions.py
njohner: That is true. I merged them.

opengever/meeting/tests/test_unit_workflow.py
njohner: No that should be fine, it inherits that method, but does not define it itself, so there is no reason for a super call here

opengever/meeting/tests/test_unit_workflow.py
njohner: yes it did

opengever/core/profiles/default/workflows/opengever_proposal_workflow/definition.xml
njohner: done

opengever/core/profiles/default/workflows/opengever_proposal_workflow/definition.xml
njohner: done

opengever/meeting/proposal_transition_comment.py
njohner: done

opengever/meeting/browser/proposaltransitions.py
```diff
@@ -13,14 +11,19 @@
 from zope.interface import Interface
 
 
-class ProposalTransitionController(BrowserView):
+class ProposalTransitionController(object):
```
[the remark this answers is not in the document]
njohner: done

opengever/meeting/proposal_transition_comment.py
```diff
@@ -0,0 +1,114 @@
...
+from zope.schema.interfaces import IContextSourceBinder
+from zope.schema.vocabulary import SimpleVocabulary
+import AccessControl
+import opengever.meeting
+from opengever.meeting.adapters import Comment
+from opengever.meeting.adapters import ICommentContainer
+
+
+@provider(IContextSourceBinder)
+def getTransitionVocab(context):
```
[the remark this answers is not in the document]
njohner: done

opengever/meeting/proposalhistory.py
njohner: done

opengever/core/upgrades/20180604105148_add_proposal_comment_permission/upgrade.py
njohner: yes done

opengever/meeting/proposal.py
deiferni: yes, it is the proper way to ask that question. otherwise you're relying on side-effects. this is rather theoretical in this case but we might change when/how the physical path is set.

opengever/meeting/tests/test_unit_workflow.py
deiferni: What i meant is that you copied the whole method body of `Transition.execute` into `Retract.execute`, but `Retract`  also inherits from `Transition`. Instead of copying you should do a super call and reuse the parents implementation. Or is there a subtle difference i did not see?

njohner: Oh sorry! I was thinking that the comment did not make much sense... :-D

PR #4408, opened 2018-06-05
Export repository roots as reimportable excel files

### LINKED ISSUE #4221: Export Möglichkeit für Ordnungssystem in xlsx Format
Es soll möglich sein, ein bestehendes Ordnungssystem (z.B. eines produktiven Mandanten) in Form einer xlsx Datei zu exportieren. Dabei soll die gleiche Vorlage, wie wir bereits für die Erstellung eines Ordnungssystems verwenden, benützt werden.

Im Export enthalten sein muss neben den Metadaten und der Positionsnummer auch die vergebenen Berechtigungen (local_roles und blocked_inheritance).

Rotonen:
* Can reimport the exported output
* Adds an action on the repository root
* Extends necessary missing functionality to the excel reporter

Closes #4221

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -12,6 +12,8 @@ Changelog
 - Document generation of policies in internal docs. [tarnap]
 - Add registry field to customize date format for sablon templates. [njohner]
 - Remove no longer used plonetheme.teamraum uninstall profile since it's integrated into opengever.core. [elioschmutz]
+- Add the ability to export repositories as excel files. [Rotonen]
+- Autosize columns on excel exports. [Rotonen]
 - Implement recently touched menu that lists checked out documents and recently touched objects. [lgraf]
 - Sort the entries of the reference browser widget. [tarnap]
 - Drop non-word meeting feature. [deiferni]
--- a/opengever/base/reporter.py
+++ b/opengever/base/reporter.py
@@ -1,7 +1,9 @@
+from math import floor
 from Missing import Value as MissingValue
 from opengever.ogds.base.actor import Actor
 from openpyxl import Workbook
 from openpyxl.styles import Font
+from openpyxl.utils import get_column_letter
 from plone.api.portal import get_localized_time
 from StringIO import StringIO
 from zope.i18n import translate
@@ -28,6 +30,21 @@ def readable_author(author):
 def readable_date(date):
     """Helper method to return a localized date"""
     return get_localized_time(date) if date else None
+
+
+def heuristic_for_11pt_calibri(charcount):
+    # Adapted from:
+    # https://msdn.microsoft.com/en-us/library/documentformat.openxml.spreadsheet.column.aspx
+    max_digit_width = 7.0
+    cell_padding = 5.0
+    return floor((charcount * max_digit_width + cell_padding) / max_digit_width * 256) / 256
+
+
+def approximate_column_width(cells):
+    return heuristic_for_11pt_calibri(max(
+        len(unicode(cell.value)) if cell.value else 0
+        for cell in cells
+        ))
 
 
 class StringTranslater(object):
@@ -81,6 +98,14 @@ class XLSReporter(object):
 
         self.insert_label_row(sheet)
         self.insert_value_rows(sheet)
+
+        column_widths = {
+            get_column_letter(i + 1): approximate_column_width(col)
+            for i, col in enumerate(sheet.columns)
+            }
+
+        for index, width in column_widths.iteritems():
+            sheet.column_dimensions[index].width = width
 
         return workbook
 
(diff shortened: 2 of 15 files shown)


opengever/base/reporter.py
```diff
@@ -95,9 +95,16 @@ def insert_value_rows(self, sheet):
...
-                if attr.get('callable'):
-                    value = value()
+
+                try:
+                    value = getattr(obj, attr.get('id'))
+                    if attr.get('callable'):
+                        value = value()
+                except AttributeError:
+                    if 'default' in attr:
+                        value = getattr(obj, attr.get('id'), attr.get('default'))
```
deiferni: why repeat the `getattr` here? that will fall back to default in any case.
i'd not use a try-except as this is normal and excepcted flow-control. better wrap falling back to `default` in a `hasattr`.

docs/HISTORY.txt
```diff
@@ -5,6 +5,7 @@ Changelog
 2018.4.0 (unreleased)
 ---------------------
 
+- Add the ability to export repository roots as excel files. [Rotonen]
```
deiferni: small detail, but you export the whole repo, right? so to avoid confustion i'd say:Add the ability to export repositories ~~roots~~ as excel files.

opengever/repository/browser/configure.zcml
```diff
@@ -107,4 +107,11 @@
       permission="zope2.View"
       />
 
+  <browser:page
+      for="opengever.repository.repositoryroot.IRepositoryRoot"
+      name="download_excel"
+      class=".excel_export.RepositoryRootExcelExport"
+      permission="cmf.ModifyPortalContent"
```
deiferni: the issue does not specify who should be able to export the excel. As this may be a quite expensive operation i'd prefer limited access. @phgross what is the motivation here, it is not specified in https://github.com/4teamwork/opengever.core/issues/4221 is this only for i.e. `Manager` or should the customer be able to export the file? In that case i'd prefer to limit this action to ~`Administrator` (not sure what permission we use).

opengever/repository/browser/excel_export.py
```diff
@@ -0,0 +1,140 @@
...
+from Products.statusmessages.interfaces import IStatusMessage
+
+
+def repository_number_to_outine_level(repository_number):
+    # All of the current repository number formatters are dot separated
+    return len(repository_number.split('.')) - 1
+
+
+def bool_label(value):
+    translater = StringTranslater(api.portal.get().REQUEST, 'opengever.repository').translate
```
deiferni: nice to see that you preserve local consistency 😂

opengever/repository/tests/test_excel_export.py
```diff
@@ -0,0 +1,72 @@
...
+from opengever.setup.sections.xlssource import XlsSource
+from opengever.testing import IntegrationTestCase
+from plone import api
+from tempfile import NamedTemporaryFile
+from tempfile import mkdtemp
+import os
+
+
+class TestRepositoryRootExcelExport(IntegrationTestCase):
+    def test_reimportability(self):
```
deiferni: this method name is a bit misleading as you test the exported excel, right?

opengever/base/reporter.py
Rotonen: So a a mix of objects can be put in and some have a callable while some have none and thus get the default.

opengever/repository/browser/configure.zcml
Rotonen: Exactly why I limited to editing rights on the repository. If something gets specified I'll rather add a new permission for it.

opengever/repository/tests/test_excel_export.py
Rotonen: Yep, originally had an import -> export -> import cycle here, but went for the quicker test.

opengever/repository/browser/configure.zcml
phgross: in my opinion, all users with the right to add/edit repositoryfolders on the current repository, should be able to generate an excel export. Those are in general only `Administrators`, therefore I would use the `opengever.repository.AddRepositoryFolder` Permission.

deiferni:
open question: should we be able to re-import the excel? what about multiple role assignments in such cases (afaik the import does not handle those yet):
screen shot 2018-06-06 at 17 42 22

Rotonen:
@deiferni ready for rereview.

@phgross ping re: reimportability upon multiple local roles.

Rotonen:
Still added a changelog entry for the excel column autosizing stuff as that might be a good one to pick out for any client facing change logs.

phgross:
@Rotonen @deiferni the import supports multiple local roles, therefore the export should also handle those ... https://github.com/4teamwork/opengever.core/blob/master/opengever/setup/sections/local_roles.py#L97-L101

Rotonen:
@phgross it already comma separates multiple roles and thus supports that as well. Seems the confusion was around not knowing we can import with multiple roles.

phgross:
@Rotonen ah ok. Then 🚢

PR #4397, opened 2018-06-04
Drop non-word meeting feature.

deiferni:
✅ before this PR is shipped to staging we need to run a meeting migration on our staging servers (`dev` and `test` with https://github.com/4teamwork/opengever.hatchery/pull/36)

With this PR there is only one way to do meetings, with word. We remove the non-word meeting features and different code-paths for non-word and word. Changes in this PR include but are not limited to:
- remove `is_word_implementation_enabled` check and registry setting, replace with `is_meeting_feature_enabled` if needed or make sure we only use the word meeting code path that was protected with a check for the flag
- port tests so that they work with the word based implementation, drop tests that tested no longer used features
- drop trix editor, widgets, javascript, …. we no longer use trix in any other places, so i decided to yank it out entirely
- running the upgrade will require `IMeetingSettings.is_word_implementation_enabled` to be `True` (only when `IMeetingSettings.is_feature_enabled` is `True`, of course), this is a necessary precondition to remove the non-word meeting stuff as otherwise data-loss will occur. if the flag is `False` a per-customer migration needs to be run first. The first upgrade acts as a guard and aborts if this condition is not met.
- upgrades will then delete no longer used fields from objects and columns from the relational database. that data is already stored in annotations of files generated in an upgrade, see https://github.com/4teamwork/opengever.duerrenaesch/pull/30 for an example of a per-customer migration.
- I have attempted split removal of functionality by subject area to maximise traceability, thus there are multiple upgrade-steps that migrate a separate content-type. This is necessary in any case as there is a mix of `UpgradeStep` and `SchemaMigration`.

This PR is a first batch that can be merged standalone and that gets the blocker out of the way. There will be follow-ups to rewrite ad merge some of the tests.

✅  i have manually tested the following things:
- [x] run the upgrades  content created on `master` (again)
- [x] manually test the whole meeting process in case there are gaps in our tests
- [x] test javascript as we removed a bunch of files
- [x] run coverage to make sure if have removed most of the darn thing
- [x] final testrun on a copy of `dev` which i had lying around from the migration test

Closes #3948.

_vorsichtig, gaaanz vorsichtig:_
careful_careful

PR #4417, opened 2018-06-06
Make date field format customisable for sablon templates

### LINKED ISSUE #4193: Formatierung Daten in Word-Files konfigurierbar pro Mandant.
Es besteht der Wunsch die Formatierung der von Sablon in die Word-Files eingefügten Daten anzupassen. Eine Anpassung in Sablon erscheint unwahrscheinlich https://github.com/senny/sablon/issues/103#issuecomment-381321062. Unser Vorschlag daher, pro Mandat einen konfigurierbaren Format-String zu hinterlegen mit dem die Formatierung der Daten **einmal** pro Mandant definiert werden kann. **Alle** in die Word-Files eingefügten Daten würden dann so formatiert, wie im Format-String definiert.

Wir sind uns bewusst, dass dies nicht alle Anforderungen abdeckt, es stellt aber eine Verbesserung zur aktuellen Situation dar in der man keine Möglichkeit hat die Datum-Formatierung anzupassen.

Umsetzung: 
- Registry-Eintrag mit gewünschtem Format für Datum
- Formatierung des Datums gemäss Eintrag für alle Sablon-Variablen

Ref: https://basecamp.com/2768704/projects/13732725/todos/340578518

njohner:
Add a registry field to customise the date format for sablon templates.

There are protocol fields provided by the `ProtocolData` class and its subclasses as well as by the `AlphabeticalToc` class. Instead of simply modifying these fields when the data is generated, I wrote a `JsonDataProcessor` class, which can be generally used to process json data before sending it to the Sablon template. This might be a bit of an overkill here, but it clearly separates getting and processing the data and leaves flexibility if we have to process other fields in the future.

resolves #4193


opengever/core/upgrades/20180605145426_add_sablon_date_formatting_string/registry.xml
```diff
@@ -0,0 +1,6 @@
+<registry>
+
+  <!-- MEETING -->
+  <records interface="opengever.meeting.interfaces.IMeetingSettings" />
```
deiferni: it may be dangerous to re-import the interface in upgrades. the interface may change/be renamed in the meantime, not sure if that also removes fields. the upgrade-step should always only include the delta at the time of its execution, and duplicate information, if necessary. that guards against changes that are made to the code-base afterwards.

for example, i have also removed a value in https://github.com/4teamwork/opengever.core/pull/4397. if your upgrade would run before mine it might already remove the setting, as what is in the python module counts (that includes **ALL** changes of the current release). my upgrade my then fail as it expects the setting to be present.

in your case have a look at https://github.com/4teamwork/opengever.core/blob/2696524bfd8c87dbe26ff40d80f0721915f7197b/opengever/core/upgrades/20170703121423_allow_zipexport_on_template_folder/registry.xml, that should cover what you need to do here.

opengever/meeting/utils.py
```diff
@@ -0,0 +1,33 @@
+from plone import api
+from opengever.meeting.interfaces import IMeetingSettings
+
+
+class JsonDataProcessor(object):
```
deiferni: this thing almost ask for being unit-tested excessively, please add some unit-tests. you might need to add a constructor that can be passed the desired format string.

njohner:
I'm done revising this one

njohner:
voila, rebased

PR #4427, opened 2018-06-08
Add support for properties to get_persisted_value_for_field() helper

lukasgraf:
The `get_persisted_value_for_field()` helper function previously wasn't able to fetch persisted values for fields that were implemented as properties. This change adds support for also correctly looking up those values.

If we're dealing with attribute storage and a field's value can't be found in the instance dict, we need to check for a property in the class dict as a last resort before raising an `AttributeError`.


lukasgraf:
rebased

PR #4433, opened 2018-06-11
Ensure default values get set by patching DX DefaultAddForm.create

lukasgraf:
Ensure default values get set by patching DX [`DefaultAddForm.create()`](https://github.com/plone/plone.dexterity/blob/2.2.8/plone/dexterity/browser/add.py#L47-L70):

Object construction in DX `DefaultAddForm.create()` previously wasn't covered by our default value patches. That means that there were still some cases where default or missing values weren't properly persisted on objects for content created via Z3C forms.

These cases are:
- Fields where the current user doesn't have write permission
  (these get [filtered out completely in `p.autoform.utils.processFields()`](https://github.com/plone/plone.autoform/blob/1.6.2/plone/autoform/utils.py#L189-L217))
- Fields that are omitted from the form
- Fields whose widgets are in `DISPLAY_MODE` even in AddForms.

In addition, I updated the `set_default_values()` helper function to canonicalize the names of the fields to their short form (without their `z3c.form` prefix) before checking whether a field already has a supplied *actual* value.

I also added tests that now check that the `test_default_values_for_types` tests now cover **every single field** for **every schema** of the types we test there.

This meta-test ensures that for every field, we expect either a **default value**, a **missing value** or an actual **value** to be persisted on the object after creation.

### DIFF
--- a/opengever/base/tests/test_default_values_for_types.py
+++ b/opengever/base/tests/test_default_values_for_types.py
@@ -25,15 +25,6 @@ FROZEN_TODAY = FROZEN_NOW.date()
 DEFAULT_TITLE = u'My title'
 DEFAULT_CLIENT = u'client1'
 
-OMITTED_FORM_FIELDS = [
-    'creators', 'predecessor', 'preview', 'archival_file_state',
-    'thumbnail', 'former_reference_number', 'reference_number',
-    'temporary_former_reference_number'
-]
-DISPLAY_MODE_FORM_FIELDS = [
-    'message_source', 'original_message',
-]
-
 REPOROOT_REQUIREDS = {
     'title_de': DEFAULT_TITLE,
 }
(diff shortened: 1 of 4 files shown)


opengever/base/tests/test_default_values_for_types.py
```diff
@@ -391,6 +415,8 @@ def test_z3c_add_form(self, browser):
 class TestDossierDefaults(TestDefaultsBase):
     """Test dossiers come with expected default values."""
 
+    portal_type = 'opengever.dossier.businesscasedossier'
```
deiferni: we have many other types, for example

- opengever.meeting.meetingdossier
- opengever.meeting.proposal
- opengever.meeting.submittedproposal
- opengever.private.*
- opengever.workspace.*

and so on. should we also test those?

lukasgraf: As discussed: Will add default value tests for `opengever.meeting.proposal` in a separate PR.

PR #4438, opened 2018-06-12
Improve meeting link render logic

### LINKED ISSUE #3779: Berechtigungen in Antragsansicht: Sitzungslink
<a href="https://github.com/jone"></img></a> **Issue by [jone](https://github.com/jone)**
_Monday Oct 09, 2017 at 07:43 GMT_
_Originally opened as https://github.com/4teamwork/gever/issues/133_

----

Ausgangslage:
- Sachbearbeiter betrachtet den eingereichten Antrag.
- Sachbearbeiter hat kein Zugriff auf Gremium / Sitzung / Sitzungsdossier.
- Antrag wurde bereits behandelt.

Problem:
- Sitzungslink: fehlende Prüfung der Berechtigung; führt zu "Unzureichende Berechtigungen" bei Klick.

Lösung:
- Nach View-Permission prüfen. Wenn keine Berechtigungen nur Titel der Sitzung darstellen statt Link.

Rotonen:
Should prevent users from getting links to meetings they cannot access.

Closes #3779

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2018.4.0 (unreleased)
 ---------------------
 
+- Only render links to meetings when the meeting itself can actually be accessed as well. [Rotonen]
 - Add support for properties to get_persisted_value_for_field() helper. [lgraf]
 - Update ftw.mobilenavigation to fix Unicode error. [tarnap]
 - Fix sorting of deeply nester repository folders. [tarnap]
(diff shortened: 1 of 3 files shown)


deiferni:
I'm sorry to say but this is the wrong approach. It won't work for multiple plone sites, unfortunately. We display the meeting link on the proposal overview in the source dossier of the proposal. That dossier may be on another plone site than the committee, in which case traversal will fail. The user may still have permissions on the other plone site though.

The correct approach would be to build something along the line of `TaskPrincipal` https://github.com/4teamwork/opengever.core/blob/master/opengever/globalindex/model/task.py#L427.

@Rotonen i'm closing the PR as we need to solve this in a completely different way 😞. Also the issues is not milestoned, if we're out of issues in 2018.4 it be best to start with stuff from 2018.5 https://github.com/4teamwork/opengever.core/milestone/48.

PR #4441, opened 2018-06-12
Fix default value persistence edge case for title and description fields.

lukasgraf:
This fixes an edge case for `title` or `description` fields with default values :

When a `title` or `description` field has a default(factory), there is one more pathological case where defaults weren't persisted correctly up until now:

The [constructor in `Products.CMFCore.PortalFolder.PortalFolderBase`](https://github.com/zopefoundation/Products.CMFCore/blob/ba14b57c0b3486a7bf5b60341ad97ead096b8f53/Products/CMFCore/PortalFolder.py#L71-L74) sets title and description specificially to the empty string if no **actual** field value was provided.

This then fools our own [`set_default_values()`](https://github.com/4teamwork/opengever.core/blob/master/opengever/base/default_values.py#L266-L310) which is called later: Because there already is a value persisted on the object for those fields, it doesn't attempt to determine and set default values. It has no way of knowing whether that empty string value was the value actually intended by the user, or the hardcoded value from the `PortalFolderBase` constructor.

We therefore treat these fields separately, and check if they have defaults **before** object construction. If so, and no **actual** value has been provided for the field in question, we inject the determined default into the `kw` dict, so that it then gets treated like it were a regular value.

---

This affected object construction for objects with a default for `title` or `descriptions`, constructed via `invokeFactory()` or `createContentInContainer()` - but not via `z3c.form` AddForms (because in add forms, the default values are pre-determined server side, and filled in as regular values into the client-side form).

---

One such case was the `title` field on the `Proposal` type: It defaults to the containing dossier's type. This PR therefore also contains a new default value test (which failed before this fix) for the Proposal type.

### DIFF
--- a/opengever/base/default_values.py
+++ b/opengever/base/default_values.py
@@ -93,6 +93,7 @@ from persistent.interfaces import IPersistent
 from plone.app.dexterity.behaviors import metadata
 from plone.behavior.annotation import AnnotationsFactoryImpl
 from plone.dexterity.utils import iterSchemata
+from plone.dexterity.utils import iterSchemataForType
 from zope.schema import getFieldsInOrder
 from zope.schema._bootstrapinterfaces import IContextAwareDefaultFactory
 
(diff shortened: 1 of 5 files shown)


lukasgraf:
rebased, fixed reason for previous test failures, this is now ready for review.

opengever/base/default_values.py
```diff
@@ -311,3 +312,43 @@ def set_default_values(content, container, values):
...
+            if not to_check:
+                return
+
+            for fn_to_check in to_check:
+                if name == fn_to_check:
+                    dv = determine_default_value(field, container)
+                    # If there is a default, and there is no *actual* value
+                    # in `kw` that should be set, inject the default into `kw`
+                    if dv is not NO_DEFAULT_MARKER:
+                        kw[fn_to_check] = dv
```
deiferni: i think you would also want to remove the name from `to_check` here for above early return.

lukasgraf: That was the intent indeed - nice catch! 😀

PR #4443, opened 2018-06-13
Implement async zip export for meetings including preview PDFs 

phgross:
- Based on netsight.async
- Preparation for long running zip generation
- Use preview PDFs from bumblebee if there exists
- No longer update protocol and documents 
- Readd Pin.js which has been removed by #4397 but is still used ...

**Fetching the previews PDF:**
 - We currently use the regular `pdf` endpoint to fetch the PDFs and add them to the ZIP. If currently no pdf exists we use the original format
 - Therefore we do not trigger an conversion of the document
 - As soon `OGIP 39 PDF speichern unter` is implemented, we can use the planed `demand` endpoint instead

**Why `netsight.async`:**
 - Currently in use in other projects (izug, my.teamraum.com etc.)
 - Its simple and works fine for this use case
 - If we tackle the "longrunning task problematic" and introduce a complete infrastructure for long running tasks we can rewrite the zip export implementation. But for this "small feature" it would be overkill.


opengever/meeting/browser/meetings/zipexport.py
```diff
@@ -53,8 +60,19 @@ def generate_zip(self):
...
 
+            if process_id:
+                self.set_progress(process_id, 90)
+
             # Return zip
             zip_file = generator.generate()
+
+            if process_id:
+                tmpdir = tempfile.gettempdir()
+                shutil.copy(zip_file.name, '{}/{}{}'.format(tmpdir,
```
deiferni: `os.path.join`

opengever/meeting/browser/meetings/zipexport.py
```diff
@@ -53,8 +60,19 @@ def generate_zip(self):
...
+                                                            process_id,
+                                                            MEETINGZIPPREFIX))
```
deiferni: this is a postfix

opengever/meeting/browser/meetings/zipexport.py
```diff
@@ -140,3 +158,47 @@ def get_meeting_json(self):
...
+
+    def download(self, process_id):
+        """Downloads the zipfile"""
+        response = self.request.response
+        tmpdir = tempfile.gettempdir()
+
+        zip_file_path = '{}/{}{}'.format(tmpdir,
+                                         process_id,
+                                         MEETINGZIPPREFIX)
+        with open(zip_file_path, 'r') as zip_file:
```
deiferni: there are multiple imho severe problems here:

- this is quite a bad security hole as the path-segment provided by the request is used directly to access files on the server.
- the file should be removed once downloaded/consumed? 
- how do we cleanup when the request fails for some reason? the file might stay around indefinitely.
- what if we have multiple servers without session stickiness, this approach wont' work as the file might be on another server

opengever/meeting/browser/meetings/zipexport.py
```diff
@@ -140,3 +158,47 @@ def get_meeting_json(self):
...
+    def __run__(self, process_id=None):
+        self.set_progress(process_id, 10)
+        self.generate_zip(process_id=process_id)
+
+    def download(self, process_id):
+        """Downloads the zipfile"""
+        response = self.request.response
+        tmpdir = tempfile.gettempdir()
+
+        zip_file_path = '{}/{}{}'.format(tmpdir,
```
deiferni: `os.path.join`

opengever/meeting/browser/meetings/zipexport.py
```diff
@@ -140,3 +158,47 @@ def get_meeting_json(self):
...
+                                         process_id,
+                                         MEETINGZIPPREFIX)
+        with open(zip_file_path, 'r') as zip_file:
+            filename = '{}.zip'.format(normalize_path(self.model.title))
+            response.setHeader(
```
deiferni: this duplicates `generate_zip` and should be extracted into its own method

opengever/core/profiles/default/jsregistry.xml
```diff
@@ -39,6 +39,17 @@
       insert-after="++resource++opengever.base/MessageFactory.js"
       />
 
+  <javascript
+      cacheable="True"
+      compression="safe"
+      cookable="True"
+      expression=""
+      enabled="True"
+      id="++resource++opengever.meeting/Pin.js"
```
deiferni: hmm i think the lines from `meeting.js` can be removed. did you verify they are still in use? and (out of curiosity), where?

iirc i removed that stuff 🤔, but might have been screwed by one of my rebases.

opengever/meeting/browser/meetings/zipexport.py
lukasgraf: This basically is a backdoor to fetch any file on the filesystem :boom: - the only thing that currently prevents this from being easily exploitable is the `MEETINGZIPPREFIX`suffix.

phgross:
@lukasgraf @deiferni Danke für euer Feedback und Input. Ich habe mich dazu entschlossen die Diskussion in unseren Internen Bereich zu verschieben und dem Architektur-Team zu übergeben (https://gever.4teamwork.ch/ordnungssystem/3/2/2/1/dossier-1763/dossier-2092/document-24575#overview), sie werden sicher auf euch zu kommen. Ich könnt das Dokument aber gerne bereits vorgängig ergänzen.

Ich schliesse deshalb diesen PR (bitte den Branch aber vorerst noch nicht löschen).

deiferni:
Wir müssen demfall separat kurz anschauen was mit dem `Pin.js` los ist. Man müsste dieses entweder wieder hinzufügen oder in `meeting.js` die nicht mehr verwendeten codezeilen löschen. Habe https://github.com/4teamwork/opengever.core/issues/4455 ergänzt.

PR #4446, opened 2018-06-13
Do not lock protocol until meeting is closed

### LINKED ISSUE #4190: Protokoll nicht mehr sperren, bis Sitzung abgeschlossen.
Im Moment ist das Protokoll erst editierbar, sobald die Sitzung abgeschlossen wurde. Diese Sperre ist historisch entstanden und macht mit der Word-SPV keinen Sinn mehr.

Das Locking des Vorprotokolls/Protokolls soll entfernt werden. Stand aktuell, folgender Hinweis erscheint:

screen shot 2018-04-19 at 13 37 39

 Stattdessen soll der Benutzer gewarnt werden, wenn er ein Protokoll mit manuellen Änderungen überschreibt. Dies ist möglich zu erkennen, wir speichern die generierte Version in `GeneratedDocument.generated_version`. Wenn die aktuelle Dokumentversion nun höher ist als die zuletzt generierte Version hat das Dokument manuelle Änderungen. Die Benutzer sollen in diesem Fall gewarnt werden, bevor das Protokoll überschrieben wird:

screen shot 2018-04-19 at 13 48 44

Das Protokoll soll neu bei Sitzungsabschluss nur noch erstellt werden, falls noch keines existiert. Bestehende Protokolle solle nicht mehr automatisch Überschrieben werden. Die Meldungen müssen angepasst werden.

Migration: Upgrade, das alle Locks auf schon bestehenden Protokollen entfernt.

Ref: https://basecamp.com/2768704/projects/13732725/todos/340578226

njohner:
Protocols can now be modified even if the meeting is not closed. Consequences of this are:
* We protect against overwriting the changes by asking for confirmation when regenerating a protocol that contains manual changes
* When closing the meeting we only regenerate the protocol if it does not contain any manual changes

The implementation is as follows:
* In the view, we add an overlay for confirmation when the protocol contains manual changes
* This protection only happens when the view is rendered, so in an old tab, the generation action could be called directly on a protocol with manual changes (or using the url). Hence I also added a protection in the action itself, to avoid overwriting the protocol without confirmation. Then that happens, we simply return to the meeting view, adding a message the protocol was not overwritten.

**Calling the action directly on a proposal that has been manually edited does not overwrite it and gives an info message**

screen shot 2018-06-13 at 10 39 39

**In the normal case, regenerating the protocol will need confirmation when it contains manual modifications**

screen shot 2018-06-13 at 11 00 13

resolves #4190


opengever/meeting/model/meeting.py
```diff
@@ -175,7 +175,7 @@ def protocol_manually_edited(self):
             return False
         self.protocol_document.generated_version
         document = self.protocol_document.resolve_document()
-        return self.protocol_document.generated_version != document.version_id
+        return self.protocol_document.generated_version != document.get_current_version_id(True)
```
deiferni: please always pass keyword arguments as keyword, `missing_as_zero` in this case

opengever/meeting/command.py
```diff
@@ -232,24 +238,33 @@ def __init__(self, context, meeting, document_operations):
...
+        if self.protocol_updated:
             api.portal.show_message(
                 self.document_operations.get_updated_message(self.meeting),
                 portal.REQUEST)
-            return
+        else:
+            api.portal.show_message(
+                self.document_operations.get_not_updated_message(self.meeting),
+                portal.REQUEST)
+        return
```
deiferni: unnecessary `return`

opengever/meeting/command.py
```diff
@@ -232,24 +238,33 @@ def __init__(self, context, meeting, document_operations):
...
-        if self.has_protocol:
-            portal = api.portal.get()
+        if not self.has_protocol:
+            return super(MergeDocxProtocolCommand, self).show_message()
+
+        portal = api.portal.get()
+        if self.protocol_updated:
             api.portal.show_message(
                 self.document_operations.get_updated_message(self.meeting),
                 portal.REQUEST)
```
deiferni: you can use `context.REQUEST`, there should be no need to fetch the portal 🤔

opengever/meeting/tests/test_meeting_word.py
```diff
@@ -147,47 +147,91 @@ def test_reopen_closed_meeting(self, browser):
...
              'Are you sure you want to close this meeting?'],
             browser.css('#confirm_close_meeting p').text)
         model.close()
+
         self.assertEquals(1, model.protocol_document.generated_version)
         self.assertEquals(u'closed', model.workflow_state)
 
     @browsing
-    def test_re_closing_meeting_regenerates_the_protocol(self, browser):
+    def test_closing_meeting_does_not_regenerates_edited_protocol(self, browser):
```
deiferni: regenerate~~s~~

opengever/meeting/browser/meetings/templates/meeting.pt
```diff
@@ -282,7 +282,7 @@
                   <div class="document-actions">
                     <a class="action generate" id="generate_protocol_with_confirm"
                        tal:condition="python:meeting.is_editable() and meeting.protocol_manually_edited()"
-                       tal:attributes="href view/url_merge_docx_protocol"
+                       tal:attributes="href python:view.url_merge_docx_protocol(True)"
```
deiferni: please always pass keyword arguments with a keyword

deiferni: actually, just push this to the view entirely, add the `if` to `url_merge_docx_protocol`, no need to have this login here in the template.

opengever/meeting/model/meeting.py
```diff
@@ -169,6 +169,14 @@ class Meeting(Base, SQLFormSupport):
         backref=backref('meeting', uselist=False),
         primaryjoin="GeneratedAgendaItemList.document_id==Meeting.agendaitem_list_document_id")
 
+    def protocol_manually_edited(self):
+        """checks whether the protocol has been manually edited or not"""
+        if not self.has_protocol_document():
+            return False
+        self.protocol_document.generated_version
+        document = self.protocol_document.resolve_document()
+        return self.protocol_document.generated_version != document.version_id
```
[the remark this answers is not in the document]
njohner: yes right, didn't see that method

opengever/meeting/browser/meetings/templates/meeting.pt
njohner: Then I would need to generate the whole tag in the view, because I need the `id=generate_protocol_with_confirm` to differentiate the actions in the javascript

njohner:
Ok I think this is done.

PR #4448, opened 2018-06-13
Include Dossier Doc-Properties for Documents inside Proposals and Tasks

### LINKED ISSUE #4410: Dossier Doc-Properties not included for Documents inside Proposals.
The `Dossier` doc-properties as specified in https://docs.onegovgever.ch/admin-manual/docproperties/list/#doc-properties-dossier should be included in each document that is inside a dossier. This does not seem to happen for `Document`s inside a `Proposal` though.

⚠️ `Document`s inside a `Task` could also be affected.

Analysis: In https://github.com/4teamwork/opengever.core/blob/b1c7fc347e25d651c88dc679b9c9f0efbd700f28/opengever/dossier/docprops.py#L330 we attempt to get the dossier with `aq_parent` instead of using  the documents `get_parent_dossier method `https://github.com/4teamwork/opengever.core/blob/b1c7fc347e25d651c88dc679b9c9f0efbd700f28/opengever/document/base.py#L38

Ref: https://basecamp.com/2768704/projects/13732725/todos/354274968

njohner:
properties from the `Dossier` were missing in the docproperties of documents inside `Task` and `Proposal` because we were getting the dossier with `aq_parent` instead of using `get_parent_dossier`.

I also added tests to make sure such documents contain the expected docproperties.

resolves #4410

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -591,6 +591,7 @@ Objects
           - self.mail
           - self.mail_msg
           - self.proposal
+            - self.proposaldocument
           - self.subdossier
             - self.subdocument
           - self.subdossier2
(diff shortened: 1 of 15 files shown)


deiferni:
Oh no, you forgot the CL and i didn't see.

PR #4457, opened 2018-06-18
Docs: added french translation

Linuexone:
Docs: added french translation


docs/public-fr/conf.py
```diff
@@ -124,9 +124,9 @@
     'logo': 'img/gever_logo.png',
     'analytics_id': 'UA-7414398-7',
     'fixed_sidebar': False,
-    'extra_nav_links': {'Produktseite': 'https://onegovgever.ch',
-                        'Feedback Forum': 'https://feedback.onegovgever.ch',
-                        'Online Demo': 'https://demo.onegovgever.ch'},
+    'extra_nav_links': {'Site du produit': 'https://onegovgever.ch',
```
deiferni: there should now be a (back-)link to the german docs, too. as one from german to french has been added in https://github.com/4teamwork/opengever.core/pull/4479

PR #4461, opened 2018-06-18
Bump integration test split count to 4

Rotonen:
We've arrived at a point we need to choose between congestion avoidance and porting tests to integration landing us any wallclock savings. I'm for the latter, thus the PR.

This is sparked by noticing the longest running tests can have been the integration tests on some branches pushed recently.


Rotonen:
Not relevant currently so closing. Do resurrect when appropriate.

PR #4464, opened 2018-06-18
Readd Pin.js.

deiferni:
Has been removed by mistake with #4397. Cherry picked from https://github.com/4teamwork/opengever.core/pull/4443.

_CL not required as this is a fix for changes of the currently developed release._

### DIFF
--- a/opengever/core/upgrades/20180613102451_readd_pin_js/jsregistry.xml
+++ b/opengever/core/upgrades/20180613102451_readd_pin_js/jsregistry.xml
@@ -1,0 +1,14 @@
+<object name="portal_javascripts" meta_type="JavaScripts Registry">
+
+  <javascript
+      cacheable="True"
+      compression="safe"
+      cookable="True"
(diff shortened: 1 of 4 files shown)


opengever/core/upgrades/20180613102451_readd_pin_js/jsregistry.xml
```diff
@@ -0,0 +1,508 @@
+<object name="portal_javascripts" meta_type="JavaScripts Registry">
```
lukasgraf: Is it really necessary to include the whole (?) contents of the JS registry in this upgrade step here?

phgross: No, should not be necessary. Sorry @deiferni

deiferni: thank you, fixed it.

PR #4474, opened 2018-06-19
Make sure excerpt document title is unicode.

### LINKED ISSUE #4452: Linked excerpt documents exist with title as bytestrings instead of unicode
We've seen cases of linked excerpt documents / copies of excerpts that have a `title` that is a bytestring instead of unicode.

There probably is an issue when generating those that needs to be fixed in code, and also existing documents need to be identified and their `title` attribute fixed.

See https://extranet.4teamwork.ch/support/gemeinde-mumpf/tracker/20 for details and an example.

deiferni:
The title entered in the create excerpt dialog was never decoded to unicode and stored as utf8. Thus displaying the title failed in various places.

This PR fixes:
- the form now correctly stores unicode
- also add a safeguard to command to convert title of newly created objects
- versions of affected documents
- messages in the journal that still contained the utf8-title, inspired by https://github.com/4teamwork/opengever.maintenance/pull/90

The upgrade-step only updates documents affected by the bug fixed here, i.e. all the different excerpts.

Fixes #4452.

### DIFF
--- a/opengever/core/upgrades/20180619143343_make_sure_excerpt_document_title_is_unicode/upgrade.py
+++ b/opengever/core/upgrades/20180619143343_make_sure_excerpt_document_title_is_unicode/upgrade.py
@@ -1,0 +1,154 @@
+from ftw.journal.config import JOURNAL_ENTRIES_ANNOTATIONS_KEY
+from ftw.journal.interfaces import IAnnotationsJournalizable
+from opengever.base.archeologist import Archeologist
+from opengever.core.upgrade import SQLUpgradeStep
+from opengever.ogds.base.utils import get_current_admin_unit
+from plone import api
+from Products.CMFPlone.utils import safe_unicode
+from sqlalchemy.sql.expression import column
+from sqlalchemy.sql.expression import table
+from zope.annotation.interfaces import IAnnotations
+from zope.app.intid.interfaces import IIntIds
+from zope.component import getUtility
+from zope.i18nmessageid import Message
+
+
+excerpts_table = table(
+    'excerpts',
+    column('excerpt_admin_unit_id'),
+    column('excerpt_int_id'),
+)
+
+
+generateddocuments_table = table(
+    'generateddocuments',
+    column('generated_document_type'),
+    column('admin_unit_id'),
+    column('int_id'),
+)
+
+
+class MakeSureExcerptDocumentTitleIsUnicode(SQLUpgradeStep):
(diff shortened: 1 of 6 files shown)


opengever/core/upgrades/20180619143343_make_sure_excerpt_document_title_is_unicode/upgrade.py
```diff
@@ -0,0 +1,154 @@
...
+        """Migrate excerpts for ad-hoc agenda items.
+
+        These excerpts are stored in the meeting dossier.
+        """
+        excerpts = self.execute(
+                excerpts_table.select()
+                .where(excerpts_table.c.excerpt_admin_unit_id == self.admin_unit_id)
+            ).fetchall()
+        for excerpt in excerpts:
+            excerpt_document = self.intids.queryObject(excerpt.excerpt_int_id)
```
phgross: Is it 100% sure, that those documents exists? I'd like to have an 

```
if not excerpt_document:
    continue
```

opengever/core/upgrades/20180619143343_make_sure_excerpt_document_title_is_unicode/upgrade.py
```diff
@@ -0,0 +1,154 @@
...
+    def migrate_excerpts_in_dossiers(self):
+        """Migrate excerpts that were returned to dossier of their proposal."""
+
+        excerpts = self.execute(
+                generateddocuments_table.select()
+                .where(generateddocuments_table.c.admin_unit_id == self.admin_unit_id)
+                .where(generateddocuments_table.c.generated_document_type == 'generated_excerpt')
+            ).fetchall()
+        for excerpt in excerpts:
+            excerpt_document = self.intids.queryObject(excerpt.int_id)
```
phgross: same here ...

opengever/core/upgrades/20180619143343_make_sure_excerpt_document_title_is_unicode/upgrade.py
deiferni: No it is not 100% guaranteed, that `if` is already in https://github.com/4teamwork/opengever.core/pull/4474/files#diff-f7d8680021d50f0988be9014214dd6e7R54 😄

phgross: Ah ok. So ready to 🚢

PR #4478, opened 2018-06-20
Fix updating mail title via API.

### LINKED ISSUE #4456: Changing the title of a mail through the REST API does not work
The title is set to the mail's subject when a mail is modified with PATCH.
It should be possible to set the title with PATCH to an arbitrary value.
PATCH should also not reset the title.

phgross:
Currently mail titles gets resetted to subject, when trying to update a mail via REST-API PATCH requests. It now only syncs the title from subject if a message object is part of the
parameters.

This fix makes it also possible to create a mail with a different title than the subject.

Closes #4456

Backport needed to: `2018.3-stable`

### DIFF
--- a/opengever/api/mail.py
+++ b/opengever/api/mail.py
@@ -41,7 +41,11 @@ class DeserializeMailFromJson(DeserializeFromJson):
         context = super(DeserializeMailFromJson, self).__call__(
             validate_all=validate_all, data=data)
 
-        context._update_title_from_message_subject()
-        initialize_metadata(context, None)
-        initalize_title(context, None)
+        if 'message' in data:
+            if not data.get('title'):
+                context._update_title_from_message_subject()
+                initalize_title(context, None)
+
+            initialize_metadata(context, None)
+
         return context
--- a/opengever/api/tests/test_mail.py
+++ b/opengever/api/tests/test_mail.py
@@ -3,6 +3,7 @@ from ftw.testbrowser import browsing
 from opengever.testing import IntegrationTestCase
 from opengever.testing.assets import load
 import base64
+import json
 
 
 class TestCreateMail(IntegrationTestCase):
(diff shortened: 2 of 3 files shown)


opengever/api/mail.py
```diff
@@ -41,7 +41,12 @@ def __call__(self, validate_all=False, data=None):
         context = super(DeserializeMailFromJson, self).__call__(
             validate_all=validate_all, data=data)
 
-        context._update_title_from_message_subject()
-        initialize_metadata(context, None)
-        initalize_title(context, None)
+        # Sync title message is given
```
deiferni: i don't understand this comment 🤔

opengever/api/mail.py
```diff
@@ -41,7 +41,12 @@ def __call__(self, validate_all=False, data=None):
...
+        if 'message' in data:
+            if not data.get('title'):
+                context._update_title_from_message_subject()
```
deiferni: `_update_title_from_message_subject` seems to be a copy of code already in `initalize_title`. do we really need to call both of them? or shouldn't `initalize_title` use `_update_title_from_message_subject`?

opengever/api/tests/test_mail.py
```diff
@@ -46,3 +47,54 @@ def test_create_mail_from_eml(self, browser):
...
+            data=json.dumps({"@type": "ftw.mail.mail",
+                             "message": {"data": msg,
+                                         "encoding": "base64",
+                                         "filename": "testmail.eml"},
+                             "title": "Separate title"}),
+            method='POST',
+            headers={'Accept': 'application/json',
+                     'Content-Type': 'application/json'})
+        self.assertEqual(browser.status_code, 201)
+        mail = self.dossier.objectValues()[-1]
```
deiferni: you could use `observe_children` here, see 
https://github.com/4teamwork/opengever.core/blob/8bcd987f7594bf52bd3e9e778596a18ba1fe9be1/opengever/meeting/tests/test_agendaitem_adhoc.py#L41-L45
for an example

opengever/api/mail.py
phgross: It's still needed: because of the current ftw.mail.mail and OGMail implementation. 😏I'd say this problem should be solved, when refactor the "mail" and make it really a document.

phgross:
@deiferni updated ...

opengever/api/mail.py
deiferni: yeah. we're not gonna do that 😝. but ok i agree, it is out of scope for this fix.

PR #4483, opened 2018-06-21
Fix bug when triggering a tasktemplatefolder with inbox as responsible.

### LINKED ISSUE #4458: OGIP 26: Auslösen von Standardablauf wirft Fehler
Beim Auslösen eines Standardablaufs, welcher Vorlagen enthält, die Eingangskörbe als Autragnehmer definiert kommt es beim Darstellen des "Auftragnehmer auswählen" Schrittes zu einem LookupError. Siehe https://sentry.4teamwork.ch/sentry/onegov-gever/issues/7373/

njohner:
The problem is in how the string representing the user is treated in the forms used to choose a task responsible (`SelectResponsiblesWizardStep`). When representing a user it has the form `orgunit:user_id` and for the inbox `inbox:orgunit`.

This is fixed here with conditionals, as it is in several other places in the code. In the future, this should all be delegated to the actor. 
resolves #4458

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 ---------------------
 
 - Add new notification roles for proposal activities. [elioschmutz]
+- Fix triggering TaskTemplateFolders with inboxes as task responsibles. [njohner]
 - Open successor task when predecessor is skipped or closed. [njohner]
 - Theme: Remove diazo rule that duplicated some JS scripts. [lgraf]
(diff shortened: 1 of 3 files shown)


njohner:
Created followup issue to delegate this to the `Actor`class:
https://github.com/4teamwork/opengever.core/issues/4484

njohner:
I amended an existing test to check that an inbox can be used as responsible of a task template.
I also checked that the test would fail before this bug-fix.

This should be ready to merge.

PR #4492, opened 2018-06-22
Match the plonetheme.teamraum inherited ruby stack to match the sablon one

Rotonen:
* Bumped the expected ruby version
* Bumped the expected bundler version
* Updated sass
* Updated rb-kqueue
* Verified it produces no output delta vs. the previous sass version

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-ruby '2.4.0'
+ruby '2.4.2'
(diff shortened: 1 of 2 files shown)


Rotonen:
Not sure if we actually still need the kqueue-select poller anymore now that there's unixy-platform independent inotify support, but as the Windows bit is definitely required, leaving it in for the sake of completeness. In the worst case it just does nothing now.

PR #4499, opened 2018-06-26
Sort user groups on user overview

### LINKED ISSUE #4223: Gruppen auf User-Übersicht sortiert darstellen
Die Gruppen eines Benutzers sollen sortiert dargestellt werden um die Übersichtlichkeit zu erhöhen.

screen shot 2018-04-24 at 12 12 18

Ref: https://extranet.4teamwork.ch/support/gever-ai/tracker/293

Rotonen:
Solved via:

https://github.com/4teamwork/opengever.ogds.models/pull/53

Closes #4223


opengever/ogds/base/tests/test_userdetails.py
```diff
@@ -69,17 +69,18 @@ def test_list_all_team_memberships(self, browser):
     @browsing
     def test_lists_group_memberships(self, browser):
         self.login(self.regular_user, browser)
-
+        # Make sure we actually sort the groups by reverse sorting the input
+        groups = ogds_service().find_user(self.regular_user.id).groups
+        groups = sorted(groups, key=lambda group: group.groupid, reverse=True)
```
deiferni: so, fundamental question: why not sort this in sql?

Rotonen: Could be I failed to see where this comes from, but I somehow thought there can also be LDAP in the mix. If not so, definitely.

deiferni: The layer this info comes from is the OGDS, thus it is always SQL. You'll probably have to implement this in the separate `opengever.ogds.models` package.

Rotonen:
@lukasgraf ready for review.

Issue #4312, opened 2018-05-17
Consider splitting out docs from opengever.core


lukasgraf:
We should consider splitting out the Sphinx documentation (`docs/intern` and `docs/public`) into its own git repository.

The frequent updating of screenshots leads to a problem where the `opengever.core` git repo is rapidly growing in size because of versioned binary blobs.

Challenges:
- Keeping schema dumps in sync
- Updating code and documentation at the same time
- Versioning of the documentation-repo

Possible alternative strategies:
- GIT LFS (disadvantage: LFS managed files *need* to be cloned via HTTPS)

lukasgraf:
Closed via #4507

PR #4507, opened 2018-06-27
Enable GIT LFS for images in docs.

lukasgraf:
This enables git-lfs for all image files under `docs/` to keep the repo from bloating.

This means that `*.jp[e]g` and `*.png` under `docs/` are now tracked by git-lfs, and are with this PR replaced with plain text pointers to the LFS resources.

Simply cloning the `opengever.core` repsitory **doesn't require either**
- git lfs
- or HTTPS

Only if someone needs these screenshots in order to work on the docs, they need
- to have `git-lfs` installed
- have enabled the repository to automatically fetch LFS managed resources

For these two steps I have provided a simple script and the necessary documentation.

This setup means, that by default, no LFS resources will be fetch (because that would count against our quota, and also slows down cloning quite a bit).

However, these defaults can be overriden by the per-repository local config (as does the `bin/enable-lfs` script for docs-people) or via a developer's `~/.gitconfig`.

*(No changelog entry since this is an internal change that doesn't affect released code)*


Rotonen:
Closes https://github.com/4teamwork/opengever.core/issues/4312 or still under consideration even with mitigation via LFS?

lukasgraf:
@Rotonen good catch, I would say this closes that issue for now.

PR #4509, opened 2018-06-27
Ensure creator gets set to the paster upon pasting content

### LINKED ISSUE #4386: Wrong ownership for copied documents
If user A creates a new document (from a template) and then user B copies the document, the creator of the copy is user A, which is wrong. The creator of the copy should be user B.

Rotonen:
We're potentially mucking with DC metadata assumptions here:

https://github.com/zopefoundation/Products.CMFCore/blob/e235f04aea5678f56693fdbaca152da0685e8fc7/Products/CMFCore/interfaces/_content.py#L235-L248

Closes #4386

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 ---------------------
 
 - Fix physical_path schema migration for oracle backends. [phgross]
+- Ensure creator sets get to the paster upon pasting content. [Rotonen]
 - Indicate change of responsible in response when task is rejected. [njohner]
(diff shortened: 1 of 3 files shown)


Rotonen:
The slicing should be safe:
https://github.com/plone/plone.dexterity/blob/master/plone/dexterity/content.py#L400

Rotonen:
As per discussion per @buchi we're, for the sake of simplicity, now considering pasting creating a new document.

Ready for review. Maybe @lukasgraf or @jone have thoughts on the semantic implications.

PR #4517, opened 2018-06-28
Display the checkin comment on bumblebee overlays

Rotonen:
* Uses the versioner to pull in the metadata history
* Uses the lazy history proxy to fetch the correct comment for the document version in question

Closes https://github.com/4teamwork/opengever.core/issues/4381

### DIFF
--- a/opengever/bumblebee/browser/overlay.py
+++ b/opengever/bumblebee/browser/overlay.py
@@ -1,7 +1,7 @@
 from Acquisition import aq_parent
 from ftw import bumblebee
-from ftw.bumblebee.mimetypes import is_mimetype_supported
 from ftw.bumblebee.interfaces import IBumblebeeDocument
+from ftw.bumblebee.mimetypes import is_mimetype_supported
 from opengever.base.browser.helper import get_css_class
 from opengever.base.interfaces import IReferenceNumber
 from opengever.base.interfaces import ISequenceNumber
@@ -12,6 +12,7 @@ from opengever.bumblebee.interfaces import IBumblebeeOverlay
 from opengever.bumblebee.interfaces import IGeverBumblebeeSettings
 from opengever.bumblebee.interfaces import IVersionedContextMarker
 from opengever.document.browser.actionbuttons import ActionButtonRendererMixin
+from opengever.document.browser.versions_tab import LazyHistoryMetadataProxy
 from opengever.document.browser.versions_tab import translate_link
 from opengever.document.checkout.viewlets import CheckedOutViewlet
 from opengever.document.document import IDocumentSchema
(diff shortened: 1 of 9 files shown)


opengever/bumblebee/browser/overlay.py
```diff
@@ -51,6 +52,17 @@ def __init__(self, context, request):
         self.context = context
         self.request = request
 
+    def get_checkin_comment(self):
+        version_id = self.version_id
+        if version_id is None:
+            version_id = getattr(self.context, 'version_id', None)
+        if version_id is not None:
+            history_data = Versioner(self.context).get_history_metadata()
+            history_proxy = LazyHistoryMetadataProxy(history_data, self.context.absolute_url(), self.context)
```
lukasgraf: There's no need to involve the `LazyHistoryMetadataProxy` here. That class is intended to lazily wrap an entire history, for listing metadata for each version in a way optimized for the versions tab (that's also [why the sort order is the way it is](https://github.com/4teamwork/opengever.core/blob/38fb1e91ee76832e9854d511239d688bb6818de6/opengever/document/browser/versions_tab.py#L210-L211)).

Instead we can simply `retrieve()` the version metadata for the specific `version_id` we want.

I would simply add a 

```python
    def get_version_metadata(self, version_id):
        return self.get_history_metadata().retrieve(version_id)['metadata']
```
to the `Versioner` class, and then do

```python
            version_metadata = Versioner(
                self.context).get_version_metadata(version_id)
            return version_metadata['sys_metadata']['comment']
```

here. No need to mess with off by one indexing pecularities and sort order.

opengever/bumblebee/browser/overlay.py
```diff
@@ -51,6 +52,17 @@ def __init__(self, context, request):
...
+            # XXX - these enumerate the history the opposite way from each other
+            return history_proxy[history_proxy._length - version_id - 1].comment
+        return ''
```
lukasgraf: This fallback should be `u''` (unicode), since that's what the comment is.

opengever/bumblebee/locales/de/LC_MESSAGES/opengever.bumblebee.po
```diff
@@ -14,6 +14,10 @@ msgstr ""
 "Preferred-Encodings: utf-8 latin1\n"
 "Domain: DOMAIN\n"
 
+#: ./opengever/bumblebee/browser/templates/bumblebeeoverlay.pt
+msgid "Checkin comment:"
+msgstr "Kommentar"
```
lukasgraf: I'm not sure if the German translation `Kommentar` makes it clear enough that this isn't a version-independent metadata field on the object, but a checkin comment instead. I'd maybe choose `Bearbeitungskommentar`. Not sure though.
bearbeitungskommentar

Rotonen: I went with the proposed one, but I also like this better. Amending.

opengever/bumblebee/browser/overlay.py
Rotonen: Thank you.

opengever/bumblebee/browser/overlay.py
Rotonen: Thank you.

Rotonen:

lukasgraf:
LGTM 👍

Issue #4515, opened 2018-06-28
Commentless noop-uploads from Office Connector break versions display


Rotonen:
1. Set up a fresh gever
2. Check out a document via OC http://localhost:8080/fd/ordnungssystem/ressourcen-und-support/personal/personalrekrutierung/dossier-4/document-2
3. Save file without changes
4. Upload with empty comment
5. Versions display is broken as the comment is now `None` instead of an empty string http://localhost:8080/fd/ordnungssystem/ressourcen-und-support/personal/personalrekrutierung/dossier-4/document-2#versions

PR #4519, opened 2018-06-29
Fix unicode error in versions tab comments.

lukasgraf:
Fixes unicode issue with `tooltip_helper` for versions tab comments that contain non-ASCII characters:

The tooltip_helper is currently used by two tabs, version and journal tabs. They don't both pass it the same type of value, one uses unicode and the other bytestrings. The `tooltip_helper` therefore needs to be able to accept both, work with unicode internally, and finally return a bytestring.

### DIFF
--- a/opengever/document/browser/versions_tab.py
+++ b/opengever/document/browser/versions_tab.py
@@ -10,11 +10,11 @@ from opengever.document import _
 from opengever.document.browser.download import DownloadConfirmationHelper
 from opengever.document.interfaces import ICheckinCheckoutManager
 from opengever.document.versioner import Versioner
-from opengever.journal.tab import tooltip_helper
 from opengever.ogds.base.actor import Actor
 from opengever.tabbedview import BaseListingTab
 from opengever.tabbedview import GeverTableSource
 from opengever.tabbedview.helper import linked_version_preview
+from opengever.tabbedview.helper import tooltip_helper
 from plone import api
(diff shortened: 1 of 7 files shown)


Rotonen:
Wanna also look at this while at it?
https://github.com/4teamwork/opengever.core/issues/4515

Rotonen:
Or could be this would actually close that:
https://github.com/4teamwork/opengever.core/pull/4519/files#diff-0cefbcf40fcdb22de623fdf58d344f0aR105

deiferni:
@Rotonen this is just a test helper, i don't think so 😉.

opengever/document/browser/versions_tab.py
```diff
@@ -10,11 +10,11 @@
 from opengever.document.browser.download import DownloadConfirmationHelper
 from opengever.document.interfaces import ICheckinCheckoutManager
 from opengever.document.versioner import Versioner
-from opengever.journal.tab import tooltip_helper
 from opengever.ogds.base.actor import Actor
 from opengever.tabbedview import BaseListingTab
 from opengever.tabbedview import GeverTableSource
 from opengever.tabbedview.helper import linked_version_preview
+from opengever.tabbedview.helper import tooltip_helper
```
tarnap: I was too lazy to move the helper :)

Thanks for doing so!

❤️

PR #3598, opened 2017-11-10
OGIP 19: Allow delete action for owner and administrator in private folder

tarnap:
Allows to remove documents and dossiers from the private folder in listings and detail views.

loschen in personlicher ablage

This is part of the OGIP 19.
Some tests are skipped, because the permissions for the records manager are implemented in an other PR.

Many workflows include the "Delete objects" permission. Please check that none of our workflows breaks due to the changes of this PR.

PR #4521, opened 2018-07-02
Restrict availability of Delete action button

### LINKED ISSUE #4334: Ordnungsposition hat unter "Aktionen" 2x löschen (2018.3 Dev)
Beim Human Testing Release 2018.3 wurde ein Fehler entdeckt.

Beschreibung:
Auf 3ter Stufe im Ordnungssystem, hat eine Ordnungsposition unter "Aktionen" 2x löschen vorhanden.

Plattform: DEV
Benutzer:  Gever Administrator
Beispielfall: https://dev.onegovgever.ch/fd/ordnungssystem/fuehrung/kommissionen-mit-selbstaendigen-verwaltungsbefugnissen/kommission-fuer-rechtsfragen/allgemeines#dossiers

Siehe Screenshot

zweimal loschen vorhanden
release2018 3op

njohner:
Since #3598, a delete action button was available everywhere for managers, and on third level of the repository it was available for admins even when the node was not empty, or twice when the node was empty. This was an unintended side effect of making the delete action available for the private folder, it became instead available everywhere where a user had delete rights on the parent. As we also have our own delete action for repos, this could lead to two available buttons.

The tests did not catch this error because we were testing on a branch repository folder, where the administrator does not have delete rights on the parent. Testing on a leaf made the error apparent.

I added a new view used to control where the delete button should be available and improved the tests for deleting repository nodes and tasktemplates and tasktemplatefolders.

resolves #4334

### DIFF
--- a/opengever/core/upgrades/20180629085145_restrict_delete_action_availability/actions.xml
+++ b/opengever/core/upgrades/20180629085145_restrict_delete_action_availability/actions.xml
@@ -1,0 +1,12 @@
+<object xmlns:i18n="http://xml.zope.org/namespaces/i18n" name="portal_actions" meta_type="Plone Actions Tool">
+
+  <!-- OBJECT BUTTONS -->
+  <object name="object_buttons" meta_type="CMF Action Category">
+
+    <object name="delete" meta_type="CMF Action" i18n:domain="plone">
+      <property name="available_expr">python:checkPermission("Delete objects", globals_view.getParentObject()) and here.restrictedTraverse('is_delete_available')()</property>
+    </object>
+
+  </object>
+
+</object>
--- a/opengever/repository/tests/test_deletion.py
+++ b/opengever/repository/tests/test_deletion.py
@@ -3,6 +3,7 @@ from Acquisition import aq_parent
 from ftw.builder import Builder
 from ftw.builder import create
 from ftw.testbrowser import browsing
+from ftw.testbrowser.pages import editbar
 from ftw.testbrowser.pages import statusmessages
 from opengever.repository.deleter import RepositoryDeleter
 from opengever.testing import IntegrationTestCase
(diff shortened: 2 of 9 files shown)


opengever/core/upgrades/20180629085145_restrict_delete_action_availability/actions.xml
```diff
@@ -0,0 +1,20 @@
+<object xmlns:i18n="http://xml.zope.org/namespaces/i18n" name="portal_actions" meta_type="Plone Actions Tool">
+
+  <!-- OBJECT BUTTONS -->
+  <object name="object_buttons" meta_type="CMF Action Category">
+
+    <object name="delete" meta_type="CMF Action" i18n:domain="plone">
+      <property name="title" i18n:translate="">Delete</property>
```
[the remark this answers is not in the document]
njohner: Right, I'm not always sure whether it will overwrite other attributes or not. But then I guess it only overwrites for list-type attributes? Where one can then use `purge=False`?

opengever/repository/tests/test_deletion.py
```diff
@@ -65,6 +65,39 @@ def test_delete_action_is_only_available_when_preconditions_satisfied(
...
+
+    @browsing
+    def test_delete_action_is_only_available_when_preconditions_satisfied_also_for_managers(
+            self, browser):
+        self.login(self.manager, browser)
+
+        browser.open(self.empty_repofolder)
+        self.assertIn(
+            'Delete',
+            browser.css('#plone-contentmenu-actions .actionMenuContent a').text,
```
deiferni: could you use `from ftw.testbrowser.pages import editbar` here? that wold prevent duplicating the css selectors for the menu.

njohner:
All done here...

opengever/repository/tests/test_deletion.py
```diff
@@ -12,6 +13,12 @@
 
 class TestRepositoryDeleter(IntegrationTestCase):
 
+    @staticmethod
+    def get_action_menu_content():
+        editbar.menu("Actions")
+        actions = editbar.menu("Actions").css("li")
```
deiferni: wouldn't `menu_options` do already what you implement here 🤔 ?

njohner: Yes you are right, sorry

njohner:
done

PR #4529, opened 2018-07-02
Untangle admin units and org units in FixtureBuilder for functional test cases

lukasgraf:
Both the admin unit as well as the org unit in the Fixture builder were created with the ID `client1`.

In addition, there was some code that attempted to link the org unit to the admin unit, but all it really did was *set the ID of the admin unit to the ID of the org unit*.

Because both they're IDs were 'client1' by default, that didn't cause any harm in cases where the admin unit ID wasn't customized.

This change untangles their IDs, setting them to `org-unit-1` and `admin-unit-1` respectively.

---

**Notes**:

- The viewlet failures on STDERR are not introduced by this change - they already exist in the STDERR of a [green build from a few days ago](https://ci.4teamwork.ch/builds/169194/tasks/269956). One culprit seems to be `test_toc.py` in `opengever.meeting`, where two periods (2016 and 2018) are present, but only one is expected to be active.

### DIFF
--- a/opengever/latex/tests/test_opentaskreport.py
+++ b/opengever/latex/tests/test_opentaskreport.py
@@ -20,6 +20,7 @@ from zope.component import adaptedBy
 from zope.component import getMultiAdapter
 from zope.interface.verify import verifyClass
 from zope.publisher.interfaces.browser import IDefaultBrowserLayer
+import unittest
 
 
 class TestOpenTaskReportPDFView(MockTestCase):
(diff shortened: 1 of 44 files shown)


opengever/latex/tests/test_opentaskreport.py
```diff
@@ -146,6 +147,9 @@ def test_task_report_is_only_available_for_current_inbox_users(self):
         self.assertFalse(
             self.portal.unrestrictedTraverse('pdf-open-task-report-allowed')())
 
+    @unittest.skip('The code that this test tests for is actually broken and '
+                   'needs to be fixed. The condition in opentaskreport.py:80 '
+                   'is wrong, it compares IDs of admin units vs org units.')
```
phgross: 🙈 

Could you fix the query extension in the opentaskreport? It should use the current admin_unit id to filter out predecessor/successor pairs not the org_unit.

lukasgraf: @phgross 👍 - will do so in a separate PR.

PR #4536, opened 2018-07-03
Format task template comments

### LINKED ISSUE #4526: Kommentar einer Aufgabenvorlage wird nicht gemäss Eingabe dargestellt.
bildschirmfoto 2018-07-02 um 14 29 25

bildschirmfoto 2018-07-02 um 14 29 31

Der Kommentar sollte mit dem `text/x-web-intelligent` transform dargestellt werden.

tarnap:
# Previously

bildschirmfoto 2018-07-02 um 14 29 31

When the text contains line breaks and similar, they were not displayed in the detail view.

bildschirmfoto 2018-07-02 um 14 29 25

# Changes with this PR

screen shot 2018-07-03 at 09 24 45

Resolves #4526

### DIFF
--- a/opengever/tasktemplates/tests/test_tasktemplate.py
+++ b/opengever/tasktemplates/tests/test_tasktemplate.py
@@ -47,6 +47,26 @@ class TestTaskTemplates(IntegrationTestCase):
                           browser.css('.documentFirstHeading').text)
 
     @browsing
+    def test_view_displays_formatted_comments(self, browser):
+        self.login(self.administrator, browser=browser)
+
+        self.tasktemplate.text = (u'- T\xfcsteintrag' '\n'
+                                  u'- m\xfet' '\n'
+                                  u'- F\xfdrmat')
+
+        browser.open(self.tasktemplate)
+
+        # Get the cell of the comments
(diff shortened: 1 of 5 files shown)


opengever/tasktemplates/tests/test_tasktemplate.py
```diff
@@ -46,6 +46,26 @@ def test_view(self, browser):
         self.assertEquals(['Arbeitsplatz einrichten.'],
                           browser.css('.documentFirstHeading').text)
 
+    @browsing
```
lukasgraf: Could you please add an additional test for XSS prevention here?

(Include `<script>alert("hello");</script>` as part of `self.tasktemplate.text`. It should be stripped from the rendered output after going through the transform, otherwise we have an XSS vulnerability).

tarnap: @lukasgraf: The tags are not stripped, but transformed. Is this enough?

lukasgraf: @tarnap yes, HTML-escaping them renders them ineffective, that works just as well. 👍 Thanks!

PR #4539, opened 2018-07-03
Make lock info visible even when it can be safely unlocked.

### LINKED ISSUE #4454: Locks are not visible for the user who created the lock
Locks created through the web interface or through the REST API are note visible for the user who created the lock. The locking information is only shown if the user can not safely unlock the item.

WebDAV locks are always treated as not safely unlockable because of a different lock type.

The locking information should be shown always if a document is locked.
There's an unlock button that should be shown only for administrators.

njohner:
Lock info is visible now even if the lock belongs to the current user.
Only managers will see the unlock button though.

**User see own lock**
screen shot 2018-07-03 at 12 24 17

**managers can unlock**
screen shot 2018-07-03 at 13 06 58

resolves #4454

### DIFF
--- a/opengever/locking/templates/info.pt
+++ b/opengever/locking/templates/info.pt
@@ -1,0 +1,45 @@
+<div id="plone-lock-status"
+     i18n:domain="plone"
+     tal:define="locked view/info/is_locked;
+                 stealable view/lock_is_stealable;
+                 lock_details view/lock_info;">
+  <tal:block condition="locked">
+    <dl class="portalMessage info">
+      <dt i18n:translate="label_locked">Locked</dt>
+      <dd>
+        <tal:author-page
+            tal:condition="lock_details/author_page"
+            i18n:translate="description_webdav_locked_by_author_on_time">
+          This item was locked by
+          <a i18n:name="author"
+             tal:content="lock_details/fullname"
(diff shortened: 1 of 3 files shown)


opengever/locking/templates/info.pt
```diff
@@ -0,0 +1,45 @@
+<div id="plone-lock-status"
```
deiferni: this looks like an exact copy of `plone/locking/browser/info.pt`, if i'm not mistaken? why did we need to copy?

njohner: It's almost an exact copy, but it's not. Here is the diff:
```
<      tal:define="locked view/info/is_locked;
---
>      tal:define="locked view/info/is_locked_for_current_user;
29c29
<         <form tal:condition="view/current_user_can_unlock_document"
---
>         <form tal:condition="stealable"
```
We now show the lock whenever the document is locked, not only when it `is_locked_for_current_user`, which is only True when the user cannot unlock safely. The second condition is adapted so that the `unlock` button is not shown when the user owns the lock and has no `manager` permissions.

deiferni: heh seem to have messed up my diff there 😄 👍

PR #4490, opened 2018-06-22
Ad-Hoc Agenda Item Templates

tarnap:
Allow users to chose the template from which to generate the ad-hoc agenda item.

## Committee edit view

The user may select a default template and select a list of allowed templates:
edit committee

The default template is pre-selected when adding an ad-hoc agenda item

The default template must be one of the allowed templates, or validation will complain:
edit committee validation

## Meeting view

Before adding a new ad-hoc agenda item, the template needs to be selected:
select template

Resolves #4187 
Closes #4476

PR #4542, opened 2018-07-04
Build missing french translations for opengever.meeting

elioschmutz:
Build missing french translations for opengever.meeting.

The translations should have been builded through the PR: #4490

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2018.4.0 (unreleased)
 ---------------------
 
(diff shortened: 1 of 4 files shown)


deiferni:
huh how could this happen? we have tests for outdated translations 🤔.

elioschmutz:
The german translations where build...

is the test only looking for the .pot file? or only for german translations?

deiferni:
ah yes it is only for the german translations as we might not be able to provide french one just in time. but the should always be built together with the french ones 🤔.

PR #4543, opened 2018-07-04
Extend activity icons

elioschmutz:
This PR extends the missing activity-icons for proposal-activities:

bildschirmfoto 2018-07-04 um 12 10 34 1


opengever/activity/__init__.py
```diff
@@ -125,3 +125,57 @@ def send_digest_zopectl_handler(app, args):
         'proposal-additional-documents-submitted',
         default=u'Additional documents submitted'),
 }
+
+# TODO: There are too many places where the activites are defined:
```
deiferni: yes i agree. but please open an issue on github instead so we can track it 😉. `tech-debt` i'd say.

opengever/activity/__init__.py
```diff
@@ -125,3 +125,57 @@ def send_digest_zopectl_handler(app, args):
...
+# i.e. ACTIVITIES = {
+#     'task-added': {
+#         'id': 'task-added',
+#         'css_class': 'created',
+#         'translation': _('task-added'),
+#         'category': 'task'
+#     }
+# }
+#
+ACTIVITIES_ICONS = {
```
deiferni: this will duplicate all the css classes currently used from response-descriptions. i don't like this. if you need to define some activity icons that are not linked to a response description implement a function that first checks the response description and then falls back to your mapping. your mapping should only contain additional classes though, not duplicate all of them.

elioschmutz:
@deiferni 👍

PR #4546, opened 2018-07-05
Disable remove action in private folder trash.

njohner:
`document-state-removed` does not exist for documents in private folders, so the action throws an error when executed.
We restrict the availability of the remove button so that it does not appear in private folders.

Do we actually need the transition at all in private dossiers? Shouldn't we also remove the transition from https://github.com/4teamwork/opengever.core/blob/master/opengever/core/profiles/default/workflows/opengever_private_folder_workflow/definition.xml ?


njohner:
Sorry unclear on my side, I meant just remove the transition `document-transition-remove`. I mean if the state `document-state-removed` is not defined, what use is there for the transition?

njohner:
Ok done. Can I merge it?

phgross:
Yes!

phgross:
https://sentry.4teamwork.ch/sentry/onegov-gever/issues/3184/

PR #4547, opened 2018-07-05
Rewrite API tests as integration tests

lukasgraf:
- Rewrite API content creation tests as integration tests
- Rewrite default value tests as integration tests
- Use documents with files in default value tests because this is our most important and common use case

### DIFF
--- a/opengever/api/tests/test_content_creation.py
+++ b/opengever/api/tests/test_content_creation.py
@@ -1,64 +1,58 @@
-from ftw.builder import Builder
-from ftw.builder import create
 from ftw.bumblebee.interfaces import IBumblebeeDocument
-from opengever.api.testing import RelativeSession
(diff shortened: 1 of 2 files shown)


opengever/api/tests/test_content_creation.py
```diff
@@ -1,64 +1,58 @@
...
-from plone.app.testing import TEST_USER_ID
-import transaction
+import json
 
 
-class TestContentCreation(FunctionalTestCase):
+class TestContentCreation(IntegrationTestCase):
 
-    layer = OPENGEVER_FUNCTIONAL_ZSERVER_TESTING
+    api_headers = {
```
phgross: I would like to have those on the IntegrationTestCase itself, there are used all over the place

lukasgraf: As discussed: We will factor this out at a later point. Follow up issue: #4591

PR #4553, opened 2018-07-06
Update filename normalisation

### LINKED ISSUE #4217: Anpassen des Dokument Dateinamens
Zurzeit noch ein Platzhalter, endgültige Spezifikation folgt noch.

Spezifikation:
- Der Dateinamen wird auf 100 Zeichen erweitert
- Dateinamen von bestehenden Dokumenten wird aktualisiert (Upgradestep)
- Normalisierung: 
  - Sonderzeichen (`\`, `/` etc.) sollen entfernt werden
  - Gross und Klein Schreibung soll erhalten bleiben
  - Leerzeichen sollen bestehen bleiben (evtl. müssen Sonder-Leerzeichen auch normalisiert werden)
  - Umlaute sollen wie gehabt normalisiert werden (`ä` wird zu `ae` etc.)

siehe [Feedback-Forum Diskussion](https://feedback.onegovgever.ch/t/dateinamen-und-dokumenttitel/917)

njohner:
In order to have filenames as close as possible to document titles, we change the filename normalisation method. We now notably allow:
* white spaces
* capital letters
* We use a German locale to transform umlauts, e.g. "ü -> ue"
* Maximal filename length is extended to 100 characters

resolves #4217

### DIFF
--- a/opengever/base/tests/test_filename_normalizer.py
+++ b/opengever/base/tests/test_filename_normalizer.py
@@ -1,0 +1,124 @@
+from unittest import TestCase
+from opengever.base.filename import GeverFileNameNormalizer
+import string
+import sys
+import unicodedata
+
+
+class TestFilenameNormalizer(TestCase):
+
+    normalizer = GeverFileNameNormalizer()
+    allowed_characters = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789 _.,=()+-'
+    max_length = 100
+
+    def test_character_cases_are_retained(self):
+        self.assertEqual(string.ascii_letters,
+                         self.normalizer.normalize_name(unicode(string.ascii_letters)))
+
+    def test_maximal_filename_length(self):
+        filename = u''.join('a' for i in range(2*self.max_length))
+        self.assertEqual(self.max_length, len(self.normalizer.normalize_name(filename)))
+
+    def test_filename_cropping_happens_after_character_mapping(self):
+        filename = u''.join(u'\xe4' for i in range(2*self.max_length))
+        self.assertEqual(self.max_length, len(self.normalizer.normalize_name(filename)))
+
+    def test_filename_cropping_happens_after_character_decoding(self):
+        filename = u''.join(u'\xe6' for i in range(2*self.max_length))
+        self.assertEqual(self.max_length, len(self.normalizer.normalize_name(filename)))
+
+    def test_filename_cropping_happens_after_consecutive_spaces_elimination(self):
+        filename = u''.join(u'  aa' for i in range(2*self.max_length))
+        self.assertEqual(self.max_length, len(self.normalizer.normalize_name(filename)))
+
+    def test_filename_cropped_length_does_not_include_file_extension(self):
+        filename = u''.join(u'\xe4' for i in range(2*self.max_length))
+        extension = '.txt'
+        expected_length = self.max_length + len(extension)
+        self.assertEqual(expected_length, len(self.normalizer.normalize(filename+extension)))
+
+    def test_normalizer_handles_non_unicode_strings(self):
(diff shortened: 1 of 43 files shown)


opengever/dossier/tests/test_resolve.py
```diff
@@ -144,7 +144,7 @@ def test_adds_journal_pdf_to_main_and_subdossier(self, browser):
         main_journal_pdf, = main_children['added']
         self.assertEquals(u'Journal of dossier An empty dossier, Apr 25, 2016 12:00 AM',
                           main_journal_pdf.title)
-        self.assertEquals(u'journal-of-dossier-an-empty-dossier-apr-25-2016-12.pdf',
+        self.assertEquals(u'Journal of dossier An empty dossier- Apr 25- 2016 12-00 AM.pdf',
```
lukasgraf: The dash followed by a space in `dossier- Apr` looks kinda weird. Maybe we should consider *stripping* unsupported special characters instead of replacing them with a dash. The spec in the feedbackforum unfortunately isn't very clear on this.

njohner:
Yes I saw that `Ä` becomes `AE`. If you think that it does not make sense, we can easily handle the German umlauts on our own, instead of letting the `FileNameNormalizer` take care of it.

For the upgrade step, I indeed need to add it, sorry! I think I did the PR to get easier feedback on th pycodestyle error I was getting...

opengever/base/interfaces.py
```diff
@@ -350,3 +351,19 @@ class IRecentlyTouchedSettings(Interface):
         description=u'How many entries to keep for a users recently touched '
                     u'items list.',
         default=10)
+
+
+class GeverFileNameNormalizer(FileNameNormalizer):
+
```
lukasgraf: I would like do see a docstring here that summarizes what *exactly* this normalizer does - i.e. what the end result of the transformation is, in the the way it is visible to the customer.

opengever/meeting/tests/test_agendaitem_list.py
```diff
@@ -50,6 +51,37 @@ def test_agendaitem_list_shows_statusmessage_when_no_template_is_configured(self
...
+            u'successfully.')
+
+        self.assertIsNotNone(meeting.agendaitem_list_document)
+        self.assertEqual(u'Agendaitem list-9. Sitzung der Rechnungspruefungskommission.docx',
+                         meeting.agendaitem_list_document.resolve_document().file.filename)
+        self.assertEqual(0, meeting.agendaitem_list_document.generated_version)
+
+        # update already generated agendaitem list
+        browser.css('.document-actions .action.generate').first.click()
+        #statusmessages.assert_message(
```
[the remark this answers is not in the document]
njohner: What we really expect here is the commented test, but because of a bug, we get the wrong message. I fixed this in a separate commit, that wasn't merged yet, hence the commented test lines here. https://github.com/4teamwork/opengever.core/pull/4550

njohner:
Ok this is ready for re-review.
I finally had to copy the `unidecode` function to have it do exactly what we wanted (notably output a space instead of the hex value of unknown characters...)

opengever/base/tests/test_filename_normalizer.py
```diff
@@ -0,0 +1,123 @@
...
+    def test_character_cases_are_retained(self):
+        self.assertEqual(string.ascii_letters,
+                         self.normalizer.normalize_name(unicode(string.ascii_letters)))
+
+    def test_maximal_filename_length(self):
+        max_length = 100
+        filename = u''.join('a' for i in range(2*max_length))
+        self.assertEqual(max_length, len(self.normalizer.normalize_name(filename)))
+
+        # Cropping should happen after character mapping
```
deiferni: is there a reason not to extract these blocks into their own method? that would improve test-results as all of these result in a separate assertion failure. if you combine them into one method it will abort after the first assertion that failed.

njohner: No reason no. They're split now.

Issue #3857, opened 2018-01-11
Port existing integration tests to use the inactive dossier fixture


njohner:
In #3764 we have added an inactive dossier containing an inactive document and task. We should make use of this in existing tests which use inactive elements.

A list of possible candidates is:
- [x]  `opengever.basetests.test_archival_file_form.TestEditArchivalFormAccess`
- [ ]  4 times in `opengever.dossier.tests.test_activate.TestDossierActivation`
- [ ]  `opengever.dossier.tests.test_base.TestDateCalculations`
- [ ]  `opengever.dossier.tests.test_deactivate.TestDossierDeactivation`
- [ ]  `opengever.dossier.tests.test_navigation.TestNavigation`
- [ ]  `opengever.dossier.tests.test_resolve.TestResolving`
- [x]  15 times in `opengever.officeconnector.tests.test_api.TestOfficeconnectorAPIDisabled`
- [x]  37 times in `opengever.officeconnector.tests.test_templates`
- [x]  9 times in `opengever/tabbedview/tests/test_dossier_listing`
- [x]  `opengever.disposition.tests.test_inactive_dossiers_are_always_not_archival_worthy`
- [ ]  `opengever.disposition.tests.TestDestructionForNotArchivedDossiers`
- [ ]  3 times in `opengever.disposition.tests.TestDispositionEditForm`
- [ ]  14 times in `opengever.disposition.tests.TestDispositionOverview`
- [ ]  `TestContentStatsIntegration` if we port it back to integration layer

njohner:
*  4 times in `opengever.dossier.tests.test_activate.TestDossierActivation`: We would also need an inactive subdossier to port these tests

njohner:
We have made use of the inactive dossier in all integration tests where it could be used and moved some tests to the integration layer.
The whole disposition module is in the functional layer and not that easily moved, so that would be a bit more work. Whether these have priority to be ported is unclear to me.

Rotonen:
The disposition tests are amongst the least recently touched and also amongst the slowest running functional tests - I see value in tackling those, but would wait for wider approval on that.

phgross:
I would say:
Let's open a separate Issue for converting the `dispositions` tests in to intergration tests, for 2019.1 release, Prio 2. and close this one.

@njohner could you do that?

njohner:
We opened a new issue for the conversion of the `dispositions` tests to the integration layer: #5133

PR #4563, opened 2018-07-09
Port archival file form tests to integration layer

njohner:
We make use of the new elements in the fixture to port the archival file form tests to the integration layer. This is part of issue #3857.

Note that:
1. There was a harmless bug in the `can_access_archival_file_form` function. While iterating over the `DOSSIER_STATES_OPEN` to check whether the user has `Modify portal content` rights on any of these states, we would return whether the user has these rights on the first of the states in the list.
2. The test `test_parent_of_content_needs_to_be_a_dossier` did not actually test what was advertised in the title, instead the second part
    ```
    with self.assertRaises(AssertionError):
        can_access_archival_file_form(user, self.document.aq_parent)
    ``` 
    actually tests that the form cannot be accessed for an object that is not `documentish`. This is already tested in a further test `test_not_accessible_on_not_basedocumentish_objects`. The new test fails instead with a `ValueError` which is raised when reaching the `PloneSiteRoot` in `find_parent_dossier`: https://github.com/4teamwork/opengever.core/blob/master/opengever/dossier/utils.py#L31

### DIFF
--- a/opengever/document/tests/test_archival_file_form.py
+++ b/opengever/document/tests/test_archival_file_form.py
@@ -1,88 +1,56 @@
-from ftw.builder import Builder
-from ftw.builder import create
+from Acquisition import aq_parent
 from ftw.testbrowser import browsing
 from opengever.document.behaviors.metadata import IDocumentMetadata
 from opengever.document.browser.archival_file_form import can_access_archival_file_form
-from opengever.testing import FunctionalTestCase
+from opengever.dossier.behaviors.dossier import IDossierMarker
+from opengever.testing import IntegrationTestCase
 from plone.namedfile.file import NamedBlobFile
 
 
-class TestEditArchivalFormAccess(FunctionalTestCase):
+class TestEditArchivalFormAccess(IntegrationTestCase):
 
-    def setUp(self):
-        super(TestEditArchivalFormAccess, self).setUp()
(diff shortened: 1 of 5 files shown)


opengever/document/tests/test_archival_file_form.py
```diff
@@ -1,88 +1,57 @@
...
-        self.document = create(Builder('document')
-                               .with_dummy_content()
-                               .within(dossier))
+        self.assertFalse(IDossierMarker.providedBy(aq_parent(self.taskdocument)))
+        self.assertTrue(can_access_archival_file_form(self.regular_user, self.taskdocument))
 
-    def test_parent_of_content_needs_to_be_a_dossier(self):
-        user = self.portal.portal_membership.getAuthenticatedMember()
+        self.login(self.administrator)
+        with self.assertRaises(ValueError):
```
phgross: The `can_access_archival_file_form` should also handle documents inside an inbox and returning False but should not raise.

njohner: Oh. I used inbox on purpose as the only way I found to not have a dossier as parent for a document 🤣.

njohner:
Ok this is ready. I also updated the `can_access_public_trial_edit_form` method to handle documents in inboxes. It now also returns `False` for documents in inboxes.

PR #4571, opened 2018-07-10
Move dossier_listing tests to integration layer.

njohner:
To avoid too much dependency on the fixtures, I modified the tests to check the listing not against fixed expected data, but against data generated on the fly from the fixture. While this makes the tests a bit more complex, I think the readability is almost better, notably for filters which are explicitly compared to filtering the generated data.

I added one test explicitly testing the function that actually generates the data from the fixture, to make sure the explicit content of the listing is also tested.

The speed gain here is very high (from 21.4 to 2.2 seconds).

This is part of issue #3857


opengever/tabbedview/tests/test_dossier_listing.py
```diff
@@ -1,165 +1,203 @@
...
-               .within(self.repository)
-               .titled(u'Dossier D')
-               .as_expired()
-               .in_state('dossier-state-archived'))
-
-        browser.login().open(
-            self.repository,
+        self.login(self.records_manager, browser=browser)
+
+        # update end dates to have expired retention periods
```
phgross: I would like to have an `expired_dossier` in the fixture. The dispositions will need them too.

njohner: Ok, fine with me. I rewrote some disposition tests yesterday and indeed always had to expire a dossier.

PR #4572, opened 2018-07-10
Add webcomponents-bundle which contains ie11 polyfills.

elioschmutz:
This PR includes missing IE11 polyfills from https://github.com/WebComponents/webcomponentsjs#webcomponentsjs-v1-spec-polyfills.

The polyfills where included in trix.js (https://github.com/basecamp/trix/tree/master/polyfills/vendor).

The Trix-Editor have been removed through the commit https://github.com/4teamwork/opengever.core/commit/c40bf4ed3cef8a0479cc02042e0ff359ed7d6fcf within the following PR: https://github.com/4teamwork/opengever.core/pull/4397

This polyfills are used by axios (Promise) and other GEVER-specific implementations (i.e. Event in https://github.com/4teamwork/opengever.core/blob/master/opengever/portlets/tree/resources/tree.js#L362)

closes #4570

PR #4579, opened 2018-07-11
Remove unnecessary webcomponents polyfill.

bierik:
In https://github.com/4teamwork/opengever.core/pull/4572 the webcomponents polyfill has been installed to have `CustomEvent` and `Promise` in IE>=9.
Beside these polyfills the webcomponents polyfill also provides a full featured webcomponents polyfill which is never used in GEVER.
The request to fetch those data are quite expensive and the code is never used.
So remove the webcomponents polyfill entirely and polyfill the `CustomEvent` and `Promise` by seperate polyfills provided by MDN.

The webcomponents polyfill also makes troubles in production mode when minified.

Tested on Chrome and IE11 in production mode.


opengever/core/profiles/default/jsregistry.xml
```diff
@@ -516,4 +516,26 @@
       insert-before="*"
       />
 
+  <javascript
+      cacheable="True"
+      compression="safe"
```
phgross: Those files are already minified, so the compression mode should be `none`, to avoid problem with two times minified js in production mode.

opengever/core/profiles/default/jsregistry.xml
```diff
@@ -516,4 +516,26 @@
...
+      expression=""
+      enabled="True"
+      id="++resource++opengever.base/es6_promise_polyfill.min.js"
+      inline="False"
+      insert-after="++resource++opengever.base/event_polyfill.js"
+      />
+
+  <javascript
+      cacheable="True"
+      compression="safe"
```
phgross: same here

PR #4581, opened 2018-07-12
Update favorite-icon after checking-out or checking-in a document.

### LINKED ISSUE #4567: Document checked out state im Favoriten Menü stimmt nicht
Bei der Dokument-Auflistung im Favoriten-Menü stimmt der checked-out-state (oranger Kreis) nicht. Er wird einmal beim Erstellen des Favorits bestimmt, und verbleibt dann für immer so.

(Ursprungs-Issue: 4teamwork/opengever.ftw#72)

### Schritte zur Reproduzierung
- Dokument in GEVER hochladen
- Favorit auf Dokument setzen
- Dokument auschecken

-> Das Dokument hat überall den richtigen checked-out-state "ausgecheckt" (oranger Kreis), ausser im Favoriten-Menü.

favoriten_checked_out_state

---

### Analyse

Das Problem hier ist, dass die [`icon_class` im SQL gespeichert wird](https://github.com/4teamwork/opengever.core/blob/3e9509988db2d35852ec17d0c6a96def01a9c80c/opengever/base/model/favorite.py#L46). Diese wird beim initialen erstellen eines Favoriten einmal bestimmt, und dann permanent [so gesetzt](https://github.com/4teamwork/opengever.core/blob/3e9509988db2d35852ec17d0c6a96def01a9c80c/opengever/base/favorite.py#L35).

Die CSS Klassen für Inhaltstypen in GEVER sind aber dynamisch - je nach State kommen da noch weitere Klassen dazu.

### Mögliche Lösungen
- Der `@favorite` Endpoint holt seine Informationen nicht bloss aus dem SQL, sondern macht auch noch ein Katalog-Query. Das würde aber nur für Mandanten-lokale Favoriten funktionieren, nicht Mandanten-übergreifende.
- Bei den [`ObjectCheckedOutEvent`](https://github.com/4teamwork/opengever.core/blob/3e9509988db2d35852ec17d0c6a96def01a9c80c/opengever/document/checkout/manager.py#L100) / [`ObjectCheckedInEvent`](https://github.com/4teamwork/opengever.core/blob/3e9509988db2d35852ec17d0c6a96def01a9c80c/opengever/document/checkout/manager.py#L176) werden jeweils alle betroffenen Favoriten im SQL aktualisiert
- Die Icons im Favoriten-Menü werden so zurechtgestutzt, dass sie keinen State enthalten. Das würde aber auch nicht wirklich richtig funktionieren: Ein Dokument in GEVER kann prinzipiell auch den Typ ändern (word -> excel), dann wär's immer noch falsch.

elioschmutz:
Updates the favorite-icon for checked-out documents:

Favorites for user who checked-out the document:
---------------------------------------------------
bildschirmfoto 2018-07-12 um 11 44 51

Favorites for user marked a checkedout document as a favorite:
---------------------------------------------------
bildschirmfoto 2018-07-12 um 11 44 38

@lukasgraf ❓ to get the different icon-classes (see method `_update_favorites_icon_class`, we have to lookup the icon for the current user and for all the other users (as already mentioned by @jone within the issue itself). I see three different approaches to reach the same result:

1. Manually manipulating the css-class (this is what I've implemented) 
We get the css-class for the current user and remove the unnecessary part of class for other users.

=> I've chosen this solution because it was fast to implement and i have nothing to refactor. But there are pro's to implement the other solutions, too...

pro: nothing to refactor, fast to implement
con: string-duplication and not object oriented

2. Using something like the `elevated_privileges` contextmanager to get the class with a not existing user. I.e. something like this:

```python
@contextmanager
def unrestricted_user(userid):
    old_manager = getSecurityManager()
    try:
        tmp_user = UnrestrictedUser(userid, '', ('manage', 'Manager'), '')
        tmp_user = tmp_user.__of__(api.portal.get().acl_users)
        newSecurityManager(getRequest(), tmp_user)
        yield
    finally:
        setSecurityManager(old_manager)

css_class_current_user = get_css_class(context)
with unrestricted_user('not_existing_user'):
  css_class_all_users = get_css_class(context)
  
```

pro: easy to understand and read
con: perhaps overkill and the problem is not really solved in the root (see 3rd case)

3. Refactoring the `get_css_class` (https://github.com/4teamwork/opengever.core/blob/master/opengever/base/browser/helper.py#L37) which is already marked to refactor it.
This is my favorite case. But to do this, we need more time for refactoring.

pro: we can use the required methods of the refactored class to get the icon-classes
con: we need more time for refactoring

@lukasgraf what do you think? Is the 1th case what you expected? Or do you prefer the 2nd or 3rd solution? or do you have another approach?

closes #4567

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -7,6 +7,7 @@ Changelog
 
 - Install CustomEvent and Promise polyfill. [Kevin Bieri]
 - Uninstall webcomponents polyfill. [Kevin Bieri]
+- Update favorite-icon after checking-out or checking-in a document. [elioschmutz]
 - Fix issue where empty version comments caused version tab to fail. [lgraf]
 - Add webcomponents-bundle which contains ie11 polyfills. The polyfills where included in trix.js. This file have been removed in commit c40bf4e. [elioschmutz]
 - Add skipped task icon. [Kevin Bieri]
--- a/opengever/base/browser/helper.py
+++ b/opengever/base/browser/helper.py
@@ -35,7 +35,7 @@ def checked_out_by(item):
 
 
 # XXX object orient me!
-def get_css_class(item, type_icon_only=False):
+def get_css_class(item, type_icon_only=False, for_user=None):
     """Returns the content-type icon css class for `item`.
 
     Arguments:
@@ -43,6 +43,9 @@ def get_css_class(item, type_icon_only=False):
     `type_icon_only` -- bool if it should only return the simple object
     type icon or the detailed content icon(mimetypes for document etc.).
     """
+    if for_user is None:
+        for_user = api.user.get_current().id
+
     css_class = None
 
     normalize = getUtility(IIDNormalizer).normalize
@@ -84,7 +87,7 @@ def get_css_class(item, type_icon_only=False):
         # is checked out by current user or someone else (or nobody at all)
         checked_out_userid = checked_out_by(item)
         if checked_out_userid:
-            if checked_out_userid == api.user.get_current().id:
+            if checked_out_userid == for_user:
                 css_class += " is-checked-out-by-current-user"
             else:
                 css_class += " is-checked-out"
--- a/opengever/base/tests/test_favorite.py
+++ b/opengever/base/tests/test_favorite.py
@@ -4,10 +4,12 @@ from ftw.testbrowser import browsing
 from opengever.base.model import create_session
 from opengever.base.model.favorite import Favorite
 from opengever.base.oguid import Oguid
+from opengever.document.interfaces import ICheckinCheckoutManager
 from opengever.testing import IntegrationTestCase
 from opengever.trash.trash import Trasher
 from plone import api
 from sqlalchemy.exc import IntegrityError
+from zope.component import getMultiAdapter
 
 
 class TestFavoriteModel(IntegrationTestCase):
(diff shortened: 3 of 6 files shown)


lukasgraf:
@elioschmutz I would go with the minimal required refactoring to `get_css_class()` to cover this use case:

- Add a new kwarg `for_user`:
  ```python
  def get_css_class(item, type_icon_only=False, for_user=None):
      if for_user is None:
          for_user = api.user.get_current().id
  ```
- and rewrite the check for the checked-out-state like
  ```python
  # Determine whether doc is checked out by user in question
  if checked_out_userid == for_user:
  ```
- then invoke like
  ```python
  get_css_class(item, for_user=owner_of_favorite)
  ```
  when updating the CSS class for a specific favorite.

Or am I missing something?

I wouldn't do a security manager switcheroo like described in 2. This might happen several times per request, and could get quite expensive as well as confusing.

elioschmutz:
@lukasgraf 👍 I updated the implementation with a minimal refactoring.

lukasgraf:
@elioschmutz can you please add somre more tests for the new `get_css_class()` behavior? I don't think your update really works. In the case where:
- The user has a favorite on a document he checked out himself
- Triggers an update to that document *as himself*
-> the class will incorrectly be determined as `is-checked-out` instead of `is-checked-out-by-current-user` for his own favorite.

Also I find the name `user_specific` confusing. The result of the `get_css_class` method is *always* user-specific, that's the whole point. The question is for *which* user - the one currently logged in (default), or a different one (`for_user='some_other_user'`).

elioschmutz:
@lukasgraf updated...

lukasgraf:
@elioschmutz as discussed, the bug I claimed didn't actually exist, I mistunderstood the code 😉 Looks good now, much easier to read IMHO 👍

elioschmutz:
@lukasgraf I agree! It's much easier to read 👍

PR #4588, opened 2018-07-12
Do no longer record `document submitted` activity while submitting a proposal with attachments.

### LINKED ISSUE #4583: Submitting proposal with attachments generates multiple notifications.
When submitting proposal with attachments currently the system generates multiple notifications. It is only necessary to generate one notification in this case.

- Philippe is in the commitee responsible group
- Dave creates a proposal with multiple attachments
- Dave submits the proposal

actual outcome:
- Philippe receives multiple notifications, one for the submitted proposal and one for each attachment

screen shot 2018-07-12 at 13 09 48

expected outcome:
- Philippe reveices one notification for the submitted proposal, separate notifications for attachments when the proposal is submitted initially are not necessary

elioschmutz:
This PR prevents from recording a `document submitted` activity on initially submitting a proposal with attachments.

closes #4583


opengever/meeting/browser/submitdocuments.py
```diff
@@ -244,8 +244,9 @@ def receive(self):
                 uuid=history_data['uuid']
             )
 
-        ProposalDocumentSubmittedActivity(
-            self.context, self.request, document.title).record()
+        if self.request.get('record_activity'):
```
deiferni: this will be a string? if so, it might be a good idea to parse a boolean here. otherwise `'False'` will be `True`.

elioschmutz: It was a boolean... but perhaps only on local requests... 

I updated the script to be failsave: I use a dict for activity-props (like the history-data).

PR #4607, opened 2018-07-16
Hide unlock button in the DocumentEditForm.

### LINKED ISSUE #4598: Lock info (and unlock button for managers) appear when editing document metadata
This is an unexpected consequence from https://github.com/4teamwork/opengever.core/pull/4539
It is quite confusing that the unlock button will appear for managers while editing the metadata of a document.

I think the lock info could be displayed while editing, that does make sense...

screen shot 2018-07-13 at 15 10 54

njohner:
We now show hide the unlock button when editing the document metadata.

I decided to leave the info that the document is locked when editing the metadata, as I think it is relevant for the user to know that while he is editing the metadata, the document is locked. Obviously we can also remove that info, if you think this is irrelevant/confusing for the user.

I also updated the changelog, as I had forgotten to add an entry in #4539 

resolves #4598 

screen shot 2018-07-16 at 18 22 08

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2018.4.0 (unreleased)
 ---------------------
 
+- Make lock info visible even when a document can be safely unlocked. [njohner]
 - Bump ftw.tokenauth to 1.1.0 which provides support for impersonation. [buchi]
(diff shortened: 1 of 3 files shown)


njohner:
Ok now the whole lock info is hidden in the edit form. This is ready for review/merge

PR #4615, opened 2018-07-17
Implement restapi @livesearch endpoint

maethu:
- Support for Plone Catalog
- Support for Solr (Untested)
- Add solr development cfg

### DIFF
--- a/development-solr.cfg
+++ b/development-solr.cfg
@@ -1,0 +1,9 @@
+[buildout]
+extends =
+    development.cfg
+    https://raw.githubusercontent.com/4teamwork/gever-buildouts/master/solr.cfg
+solr-port = 8983
+
+[solr]
+url = http://archive.apache.org/dist/lucene/solr/7.3.0/solr-7.3.0.tgz
+md5sum = dfb6893fc656e14919df48ff654d38f2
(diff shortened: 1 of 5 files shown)


buchi:
@phgross We need to customize the search endpoints as plone.restapi does not support Solr. Further `@livesearch` is optimized for speed and only queries and returns a minimal set of metadata.

Be aware that is PR is against the gever-ui branch. No need for an in-depth review. We'll make separate PRs for each feature in this branch.

development-solr.cfg
```diff
@@ -0,0 +1,9 @@
+[buildout]
+extends =
+    development.cfg
+    https://raw.githubusercontent.com/4teamwork/gever-buildouts/master/solr.cfg
+solr-port = 8983
+
+[solr]
+url = http://archive.apache.org/dist/lucene/solr/7.3.0/solr-7.3.0.tgz
```
buchi: Not needed. The current default is already version 7.3.1

opengever/api/livesearch.py
```diff
@@ -0,0 +1,51 @@
...
+        search_term = self.request.get('q', None)
+        limit = int(self.request.get('limit', 10))
+        path = self.request.get('path',
+                                '/'.join(api.portal.get().getPhysicalPath()))
+
+        if not search_term:
+            return []
+
+        if settings.use_solr:
+            view = self.context.restrictedTraverse('@@livesearch_reply')
```
buchi: Instead of traversal I would do a direct adapter lookup (getMultiAdapter). Should be much faster.

PR #4638, opened 2018-07-25
Add feature to generate a pdf listing the tasks when resolving a dossier.

### LINKED ISSUE #4184: Ablage Aufgaben-PDF bei Dossierabschluss
Beim Abschluss eines Hauptdossiers soll ein PDF-Extraktion aller im Dossier enthaltenen Aufgaben als separates Dokument im Dossier abgelegt werden. Ähnlich wie dies bereits für das Journal gemacht wird. 

 Pro Aufgaben sollen folgende Metadaten ersichtlich sein:
•	Titel
•	Beschreibung
•	Auftragstyp
•	Frist
•	Auftraggeber und Auftragnehmer
•	Erledigungsdatum

Zudem soll auch der Verlauf der Aufgabe ersichtlich sein, ähnlich wie das auch in der Detailansicht der Fall ist. Am sinnvollsten ist wohl eine Auflistung in Tabellenform. Dabei soll folgendes ersichtlich sein:
•	Zeitpunkt
•	Benutzer
•	Transition
•	Text

Die Erstellung des „Aufgaben-PDFs“ soll konfigurierbar sein, via zusätzlicher Option im IDossierResolveProperties und standardmässig deaktiviert sein.

Behandlung bei Wiedereröffnung:
Wird ein Dossier erneut abgeschlossen, so soll das Aufgaben-PDF Dokument in Form einer neuen Version aktualisiert werden. Damit das Aufgaben Dokument als solches wiedererkannt wird und aktualisiert werden kann, soll es mit einem Marker Interface markiert werden.

njohner:
We add a new feature to create a document of the tasks of a dossier when it gets resolved.
The document is a pdf containing a list of the tasks. For each task we show the metadata followed by a table presenting its history.
This feature is controlled by a configuration entry `tasks_pdf_enabled` in `IDossierResolveProperties` and is deactivated by default.

Finally we introduce an interface to mark the generated document to allow updating it instead of generating a new one when the dossier is reactivated and then closed again. We also adapt the journal pdf to follow that same behaviour.

Here an example of a generated pdf:
[Aufgabenliste des Dossier grege, 25.07.2018 10 05 (8).pdf](https://github.com/4teamwork/opengever.core/files/2228165/Aufgabenliste.des.Dossier.grege.25.07.2018.10.05.8.pdf)

resolves #4184

### DIFF
--- a/opengever/dossier/resolve.py
+++ b/opengever/dossier/resolve.py
@@ -3,6 +3,7 @@ from opengever.base.command import CreateDocumentCommand
 from opengever.base.security import elevated_privileges
 from opengever.document.archival_file import ArchivalFileConverter
 from opengever.document.behaviors import IBaseDocument
+from opengever.document.versioner import Versioner
 from opengever.dossier import _
 from opengever.dossier.base import DOSSIER_STATES_OPEN
 from opengever.dossier.behaviors.dossier import IDossier
@@ -10,6 +11,8 @@ from opengever.dossier.behaviors.dossier import IDossierMarker
 from opengever.dossier.behaviors.filing import IFilingNumberMarker
 from opengever.dossier.interfaces import IDossierResolveProperties
 from opengever.dossier.interfaces import IDossierResolver
+from opengever.dossier.interfaces import IDossierJournalPdfMarker
+from opengever.dossier.interfaces import IDossierTasksPdfMarker
 from plone import api
 from Products.CMFCore.utils import getToolByName
 from Products.Five.browser import BrowserView
@@ -17,6 +20,7 @@ from zope.component import adapter
 from zope.component import getAdapter
 from zope.component import getSiteManager
 from zope.i18n import translate
+from zope.interface import alsoProvides
 from zope.interface import implementer
 from zope.interface import implements
 from zope.schema.interfaces import IVocabularyFactory
@@ -247,12 +251,26 @@ class StrictDossierResolver(object):
         if dossier and dossier.end:
             kwargs['document_date'] = dossier.end
 
+        results = api.content.find(object_provides=IDossierJournalPdfMarker,
+                                   depth=1,
+                                   context=self.context)
+
         with elevated_privileges():
-            CreateDocumentCommand(
-                self.context, filename, view.get_data(),
-                title=translate(title, context=self.context.REQUEST),
-                content_type='application/pdf',
-                **kwargs).execute()
+            if len(results) > 0:
+                document = results[0].getObject()
+                document.title = translate(title, context=self.context.REQUEST)
+                comment = _(u'Updated with a newer generated version from dossier ${title}.',
+                            mapping=dict(title=self.context.title))
+                document.update_file(view.get_data(), create_version=True,
+                                     comment=comment)
+                return
+
+            document = CreateDocumentCommand(
+                        self.context, filename, view.get_data(),
+                        title=translate(title, context=self.context.REQUEST),
+                        content_type='application/pdf',
+                        **kwargs).execute()
+            alsoProvides(document, IDossierJournalPdfMarker)
 
     def create_tasks_listing_pdf(self):
         """Creates a pdf representation of the dossier tasks, and add it to
@@ -266,6 +284,7 @@ class StrictDossierResolver(object):
             return
 
         view = self.context.unrestrictedTraverse('pdf-dossier-tasks')
+
         today = api.portal.get_localized_time(
             datetime=datetime.today(), long_format=True)
         filename = u'Tasks {}.pdf'.format(today)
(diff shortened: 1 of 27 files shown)


opengever/latex/dossierjournal.py
```diff
@@ -40,7 +41,9 @@ def get_data(self):
 
         assembler = getMultiAdapter((self.context, self.request),
                                     IPDFAssembler)
-        return assembler.build_pdf()
+        data = assembler.build_pdf()
+        noLongerProvides(self.request, IDossierJournalLayer)
```
njohner: This is not super elegant, but when resolving a dossier, we now generate both the journal and the task list. If we don't clean up the added interface on the request, we then run into problems...

opengever/dossier/tests/test_resolve.py
```diff
@@ -197,7 +197,77 @@ def test_sets_journal_pdf_document_date_to_dossier_end_date(self, browser):
...
+        self.assertEquals(u'Task list of dossier An empty dossier, Apr 25, 2016 12:00 AM',
+                          main_tasks_pdf.title)
+        self.assertEquals(u'Task list of dossier An empty dossier, Apr 25, 2016 12 00 AM.pdf',
+                          main_tasks_pdf.file.filename)
+        self.assertEquals(u'application/pdf',
+                          main_tasks_pdf.file.contentType)
+        self.assertFalse(main_tasks_pdf.preserved_as_paper)
+
+        self.assertEquals(1, len(sub_children['added']))
+        sub_tasks_pdf, = sub_children['added']
```
phgross: It was not specified well, but i would only generate a task-pdf only for main dossiers, but make sure that tasks are recursively included, so the  main_task pdf should also contain tasks inside subdossiers.

opengever/latex/dossiertasks.py
```diff
@@ -0,0 +1,91 @@
...
+                                   'deadline': deadline,
+                                   'responsible': task.responsible,
+                                   'responsible_client': task.responsible_client,
+                                   'history': task_history.get_listing(response_container)})
+
+        return {'task_data_list': task_data_list,
+                'label': title}
+
+    def get_tasks(self):
+        task_brains = api.content.find(context=self.context, portal_type='opengever.task.task')
```
phgross: Use `object_provides=ITask` here to include task similar object like forwardings.

opengever/latex/listing.py
```diff
@@ -273,6 +274,32 @@ def get_columns(self):
...
+
+    def get_columns(self):
+        return [
+            Column('date',
+                   _('label_time', default=u'Time'),
+                   '10%',
+                   lambda item: helper.readable_date_time(item, item.date)),
+
+            Column('creator',
+                   _('label_creator', default='Response Creator'),
```
phgross: i would change the default here to `Changed by`

opengever/latex/locales/de/LC_MESSAGES/opengever.latex.po
```diff
@@ -129,14 +144,23 @@ msgstr "Art"
...
 #. Default: "Title"
 #: ./opengever/latex/dossierdetails.py
 #: ./opengever/latex/listing.py
 msgid "label_title"
 msgstr "Titel"
 
+#. Default: "Transition"
+#: ./opengever/latex/listing.py
+msgid "label_transition"
+msgstr "Transition"
```
phgross: Transition ist für den Benutzer wohl nicht verständlich, ich würde das ganze eher mit `Aktion` übersetzen.

opengever/dossier/resolve.py
```diff
@@ -245,12 +249,28 @@ def create_journal_pdf(self):
...
+                                   context=self.context)
+
         with elevated_privileges():
-            CreateDocumentCommand(
-                self.context, filename, view.get_data(),
-                title=translate(title, context=self.context.REQUEST),
-                content_type='application/pdf',
-                **kwargs).execute()
+            if len(results) > 0:
+                document = results[0].getObject()
```
phgross: I would like to have something like this and let update_document do the work:

```
document = results[0].getObject()
document.title = translate(title, context=self.context.REQUEST)
comment = _(u'Updated with a newer generated version from dossier ${title}.',
            mapping=dict(title=self.context.title))
document.update_file(view.get_data(), create_version=True, comment=comment)
```

opengever/latex/dossierjournal.py
phgross: you could refactor the `provide_request_layer` and implement a context manger, as we did for internal workflow transitions for example (see as_internal_workflow_transition).

njohner: Right good idea, that would be much cleaner.

opengever/dossier/resolve.py
njohner: Yes sounds good to me. I had actually asked in the gever channel whether we had a method to update the file properly and planned to maybe make another issue for that, to clean it also up at other places where it is used.

njohner:
Ok I think this is ready for re-review.
For the context manager to provide the request layers, I made a specific one for each view (`DossierTasksPDFView` and `DossierJournalPDFView`). I did not want to refactor `provide_request_layer` as this is provided by the `ftw.pdfgenerator` module and would introduce a non-backwards-compatible change there. We could also add a context manager while keeping the `provide_request_layer` function, but I think in most cases  `provide_request_layer` is fine.

phgross:
@njohner Wait with merge till we open the 2018.4 release lock

njohner:
Ok, I'll already rebase it anyways. There probably won't be much more conflicts coming in.

PR #4655, opened 2018-07-30
Refactor (some) activity functional test cases

tarnap:


Rotonen:
Hijacked and polished this one.

Ready for review.

opengever/activity/tests/test_resolve.py
```diff
@@ -1,125 +1,85 @@
...
-            browser.login().open(
-                self.portal, view='resolve_notification',
-                data={'notification_id': notification.notification_id})
+            self.login(self.regular_user, browser)
+            browser.open(self.portal, view='resolve_notification', data={'notification_id': notification.notification_id})
+
+    @browsing
+    def test_mark_notification_as_read(self, browser):
+        self.login(self.regular_user, browser)
+        # XXX - We have to hit the view twice here to mark it as read - a timing issue or a bug?
```
phgross: Re fetching the notification should solve your issue. Should definitely not be necessary to call the view twice.

Rotonen: As `notification_center().get_notification(self.notification.notification_id).is_read` also returns `False` when the view is hit only once, which is how I ended up trying this out. I'm very unsure as to what is going on there.

phgross: Seems weird/wrong i'll take a look into it tomorrow ...

Rotonen:
@phgross so are we now at the end of our hijack tennis here?

PR #4666, opened 2018-08-02
Add the document UUID to the OC document payloads

Rotonen:
This is an interim step towards tackling https://github.com/4teamwork/opengever.core/issues/4569 and also something which will allow for redownload prevention optimisation for Oneoffixx payloads while keeping backwards compatibility.


opengever/officeconnector/service.py
```diff
@@ -118,13 +118,13 @@ def document_is_valid(document):
     def get_base_payloads(self):
         # Require an authenticated user
         if not api.user.is_anonymous():
-            documents = []
+            documents = {}
```
lukasgraf: You might want to use a `collections.OrderedDict` here (or sort on iteration below) in order to get a stable payload order. Functionally the order probably doesn't matter, but that should make for much easier testing and debugging.

Rotonen:
Had to snatch a couple of upcoming changes from the oneoffixx branch, but moved the OC tests to use explicit assertions for the most part, as discussed.

Ready for rereview.

opengever/officeconnector/tests/test_api_dossier_oneoffixx.py
```diff
@@ -14,15 +17,102 @@ def test_create_with_oneoffixx(self, browser):
...
+            u'documents': [u'createtreatydossiers000000000002'],
+            u'exp': 4121033100,
+            u'sub': u'robert.ziegler',
+            u'url': u'http://nohost/plone/oc_checkout',
+            }
+        raw_token = payloads[0].get('checkout-url').split(':')[-1]
+        token = jwt.decode(raw_token, verify=False)
+        self.assertEqual(token, expected_token)
+
+        expected_oneoffixx_xml = [
```
lukasgraf: Just as a side note: For stuff like this, where you have whitespace-sensitive multiline-strings, a combination of a long `"""` multiline string and [`textwrap.dedent()`](https://docs.python.org/2/library/textwrap.html#textwrap.dedent) works nicely ([example](https://github.com/4teamwork/opengever.core/blob/f2edefe1726e7112f908004cbf9b237cc0a1b45f/opengever/base/tests/test_default_values_for_types.py#L849-L860)).

Rotonen: Thank you, a nice trick.

In this case a bit inconvenient to not anger pycodestyle with the first line of the XML.

PR #4680, opened 2018-08-06
Simplify agenda item titles for folder names for meeting zip exports

### LINKED ISSUE #4528: Rename folders in meeting export
The `export-meeting-zip` view currently creates a zip-file with long folder and filenames. This causes issues on windows due to some restrictions regarding path-length. We can mitigate these issues by no longer using the agenda-item title as foldername, but use sequentially numbered foldernames instead.

Use a translated title "Traktandum 01", "Traktandum 02" etc. instead of the whole agenda-item titles as foldernames for the agenda items.

The following things need to be adapted:

- the export where the file is written
- the export when it writes `meeting.json` metadata file

Ref https://basecamp.com/2768704/projects/13732725/todos/339165331.

Rotonen:
* Use the localized title string for the agenda item folders within the zip export
* Localize the relevant tests as well

Closes #4528

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -19,6 +19,7 @@ Changelog
 ---------------------
 
 - Make upgrade step 20180619143343 deferrable: Use of CMFEditions getHistory causes massive amounts of savepoints. [lgraf]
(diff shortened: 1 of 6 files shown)


Rotonen:
Ready for rereview.

PR #4681, opened 2018-08-06
Do not show lockinfo in mail edit form.

### LINKED ISSUE #4662: Mail edit form shows unnecessary locked message
When editing a mail (`ftw.mail.mail`), an unnecessary **lock message** is shown even if the user owning the lock is the one currently editing it:

mail_edit_locking

This has been fixed for the document edit form in #4607, but it seems this doesn't apply to mails yet.

njohner:
Not sure why but when editing an E-mail, the view attribute of the lockinfo viewlet is the form wrapper (`Products.Five.metaclass.MyFormWrapper`) and not the form itself. Hence it does not provide the `IEditForm` interface and we first need to get the form instance in that case.

I'd be happy with a cleaner solution though...

fixes #4662

### DIFF
--- a/opengever/locking/info.py
+++ b/opengever/locking/info.py
@@ -55,7 +55,8 @@ class GeverLockInfoViewlet(LockInfoViewlet):
         return "Manager" in api.user.get_roles()
 
     def is_edit_form(self):
-        return IEditForm.providedBy(self.view)
+        return (IEditForm.providedBy(self.view)
+                or IEditForm.providedBy(getattr(self.view, "form_instance", None)))
 
     def get_related_meeting_from_protocol(self):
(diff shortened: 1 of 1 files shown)


lukasgraf:
@njohner that form wrapper for the `MailEditForm` is unnecessary indeed - I just stumbled across this myself this afternoon, and already dropped it in my (WIP) branch: [b360dcba](https://github.com/4teamwork/opengever.core/commit/007a5c409eddee8689bb6dd7f4258825b360dcba).

There is also a highly confusing duplicate edit form for the mail, which [I dropped as well](https://github.com/4teamwork/opengever.core/commit/1bc8b9c403c49f31eecf1ad70c2cbd48def73ded).

Issue #4672, opened 2018-08-03
Base URL is incorrect for contentish types


lukasgraf:
For contentish types, the base URL in the `<base href="..." />` tag is incorrect - it contains the original, unmodified request URL, including tailingview names like `/tabbed_view`. It **should instead *just* contain the context URL**, like it does for folderish types.

The presence of a correct base URL that points to the context is required in order to build correct URLs on the frontend side.

This incorrect base URL is the underlying cause for issues like #4671 and #4665 where invalid URLs are constructed in JavasScript.

lukasgraf:
The base URL **got broken in Plone `4.3.12`**. People later discovered that that was a bad idea (see [community thread](https://community.plone.org/t/how-to-get-context-url-in-js-on-plone-4-3-12/4031) and plone/Products.CMFPlone#2051).

---

So the consensus was to introduce a **new data attribute on the body tag, `data-base-url`**, that should now contain the correct context URL. This was implemented in 

plone/Products.CMFPlone#2101
zopefoundation/Products.CMFDefault#1
plone/plonetheme.sunburst#17
plone/plonetheme.classic#6

and merged in Plone `4.3.16`.

---

However, that implementation contains a bug. It [implements a fallback](https://github.com/plone/Products.CMFPlone/pull/2101/files#diff-b830a98758551f60451fc335a0b01457R4) that, if the `data-base-url` attribute if its not present already, will set it to the value of the JavaScript global `base_url`.

That JavaScript global is determined using a different strategy: It [gets set to the `HTTP_REFERRER`](https://github.com/plone/Products.CMFPlone/pull/2101#pullrequestreview-52251166), which is just plain wrong. The HTTP_REFERRER would be disastrous if the `plone_javascript_variables.js` was included inline instead of fetched with a separate request. Because it isn't, we luckily don't get a base URL that points to random external sites, but it's still incorrect (not the context URL, but the unmodified request URL instead).

---

Because in our Diazo theme the `data-base-url` attribute on the <body /> tag [isn't copied over explicitely](https://github.com/4teamwork/opengever.core/blob/2c104ed2b0d9bfb4e7f388d8adb5890f2dfa00c3/plonetheme/teamraum/resources/rules.xml#L101), it gets omitted from the transformed markup. Therefore the fallback mentioned above kicks in, and we still wouldn't get a correct base URL even after updating to `4.3.16`.

lukasgraf:
#### Proposed Solution

[...]

Issue #4665, opened 2018-08-02
User lands on `/tabbed_view` of a dossier upon creation of a dossier


Rotonen:
Office Connector JS will not work on the `/tabbed_view` URL of a dossier.

Rotonen:
After discussing this with @lukasgraf, seems the root cause is not using the tabbedview-provided baseurl global correctly.

https://github.com/4teamwork/ftw.tabbedview/blob/f9ab4c706b98e04df558fbb41e1332daca94eeb9/ftw/tabbedview/browser/resources/tabbedview.js#L61-L66

PR #4686, opened 2018-08-07
Fix base URL tag for contentish objects.

### LINKED ISSUE #4671: Invoking DocumentTabbedView with explicit @@tabbed_view view suffix fails
When invoking the `DocumentTabbedView` view with it's `@@tabbed_view` view suffix explicitly appended to the URL, the subsequent TabbedView request issued from JavaScript will fail, because it constructs an incorrect URL:

`http://localhost:8080/fd/ordnungssystem/fuehrung/gemeinderecht/dossier-18/document-37/@@tabbed_view` ->

```python
2018-08-02 16:16:35 ERROR Zope.SiteErrorLog 1533219395.390.565403762446 http://localhost:8080/fd/ordnungssystem/fuehrung/gemeinderecht/dossier-18/document-37/tabbed_view/tabbed_view/listing
Traceback (innermost last):
  Module ftw.tabbedview.browser.tabbed, line 158, in listing
  Module ftw.tabbedview.browser.tabbed, line 50, in _resolve_tab
AttributeError: 'DocumentTabbedView' object has no attribute 'restrictedTraverse'
```

This is because on the JavaScript side, a target URL like
```
... document-37/tabbed_view/tabbed_view/listing
                ^^^^^^^^^^^^^^^^^^^^^^^
```

gets constructed.

Sentry Issue: https://sentry.4teamwork.ch/sentry/onegov-gever/issues/3586/

njohner:
We fix the base URL for contentish objects, which was incorrectly set to the request URL, hence including the view name.
This is for issue #4672, fixing the problem for the 2018.4 release, but the issue should not be closed as we will implement a better solution for 2018.5.

In the meantime, this resolves #4671

### DIFF
--- a/opengever/base/tests/test_layout.py
+++ b/opengever/base/tests/test_layout.py
@@ -1,5 +1,7 @@
 from ftw.testbrowser import browsing
 from opengever.testing import IntegrationTestCase
+from plone import api
+from zope.component import getMultiAdapter
 
 
 class TestGeverLayoutPolicy(IntegrationTestCase):
@@ -45,3 +47,14 @@ class TestGeverLayoutPolicy(IntegrationTestCase):
         self.activate_feature('meeting')
         browser.open(self.meeting)
         self.assertIn('model-meeting', browser.css('body').first.classes)
+
+    def test_render_base_returns_correct_url(self):
+        self.login(self.manager)
+        portal = api.portal.get()
+        contents = api.content.find(context=portal)
+        portal_types = set(el.portal_type for el in contents)
+        for portal_type in portal_types:
+            brains = api.content.find(portal_type=portal_type)
+            obj = brains[0].getObject()
+            layout = getMultiAdapter((obj, self.request), name=u'plone_layout')
+            self.assertEqual(layout.renderBase(), obj.absolute_url().rstrip("/") + "/")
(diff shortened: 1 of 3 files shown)


njohner:
issue #4665 will be fixed separately, as this is actually a somewhat different problem (the javascript does not retrieve the `href` from the `base` tag, but the window location).

opengever/base/tests/test_layout.py
```diff
@@ -45,3 +47,14 @@ def test_model_class_on_sql_wrapper(self, browser):
         self.activate_feature('meeting')
         browser.open(self.meeting)
         self.assertIn('model-meeting', browser.css('body').first.classes)
+
+    def test_render_base_returns_correct_url(self):
+        self.login(self.manager)
+        portal = api.portal.get()
+        contents = api.content.find(context=portal)
```
lukasgraf: Just as a side note:
If you ever need to know all the distinct values for a particular catalog index (like the `portal_type` index here), you can use `index.uniqueValues()`:

```python
index = catalog._catalog.indexes['portal_type']
distinct_portal_types = index.uniqueValues()
```

Doesn't make a difference here though, because in the end you'll have to query pretty much the entire catalog anyway 😜

njohner: Ah good to know.

PR #4690, opened 2018-08-07
Incorporate opengever.ogds.models into the opengever.core codebase

### LINKED ISSUE #4313: Move opengever.ogds.models into opengever.core
`opengever.ogds.models` is currently a separate package, but since the izug_bridge is no longer in use, it does not make sense to have a separate package.

Rotonen:
Simply moved it in as there is no profile and it's just code.

Had to give it a linter polish run so it passes the QA filter we have. I did not address everything flaring up for myself, but went a bit over the bare minimum.

Addressed the inheritance `TODO` found in `opengever.ogds.models.Team`.

Closes #4313


Rotonen:
Once this gets merged, this should also get merged https://github.com/4teamwork/opengever.ogds.models/pull/54.

Rotonen:
Retaining the commit history would either rewrite the dates on `opengever.ogds.models` objects or we'd have to force push on the `master` of `opengever.core` to shuffle the objects into the tree. AFAIK we're considering both routes not worth the hassle. Could be I'm missing some obvious thing on how to do tree merges in Git; this is partially uncharted territory for me.

As far as I could see there was only `opengever.ogds.models.Team` where there was a TODO left, but can take a second look as I need to rebase anyway. I'd rather do this all neatly in a single swoop to the agreed upon effort level.

@phgross is the history preserving a thing we want, and if so, by which compromise?

deiferni:
- the cleanup is not a todo iirc, just something i remembered when seeing this PR
- rewriting dates is not necessary imo, there is supposedly a subtreemerge script somewhere which should already be able to incorporate the history without a need for force-pushes. we have used it to merge all sub-packages into `opengevert.core`. we could not find it though in a quick search 😞should poke @buchi for more information.

Rotonen:
How to do the packageness stripping then, after the import or only importing from a rebase-edited branch?

Rotonen:
Anyone want to adopt this one?

deiferni:
i could look into this eventually if you don't get to do it.

docs/HISTORY.txt
```diff
@@ -114,6 +114,7 @@ Changelog
 - Make sure dossier manager field is set when protecting a dossier. [njohner]
 - Fix officeconnector issue when invoked from url with trailing view name. [njohner]
 - Sort excerpts by title. [Rotonen]
+- Vendor opengever.ogds.models into the opengever.core codebase. [Rotonen]
```
phgross: Move to 2018.5

Rotonen:
@phgross rebased the changelog.

PR #4691, opened 2018-08-07
Make sure dossier manager field is set when protecting a dossier.

### LINKED ISSUE #4673: dossier_manager field must be conditionally required
The "protect dossier" form fieldset currently allows to save the form with principals selected for the `reading` or `reading_and_writing` fields, but **without** a `dossier_manager` (for example by explicitly de-selecting the prefilled `dossier_manager`).

This shouldn't be allowed, as this would have unintended consequences (the dossier not actually getting protected). Therefore the `dossier_manager` field should be made required if any of those fields have a value.

*Discovered as part of #4653*

njohner:
We require that the dossier manager field is set when `reading` or `reading_writing` protection is set.
An exception is raised if this is not the case:

screen shot 2018-08-07 at 16 49 38

resolves #4673

### DIFF
--- a/opengever/dossier/locales/de/LC_MESSAGES/opengever.dossier.po
+++ b/opengever/dossier/locales/de/LC_MESSAGES/opengever.dossier.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
-"POT-Creation-Date: 2018-07-13 15:27+0000\n"
+"POT-Creation-Date: 2018-08-07 14:25+0000\n"
 "PO-Revision-Date: 2017-05-23 04:53+0000\n"
 "Last-Translator: Jacqueline Sposato <jacqueline.sposato@gmail.com>\n"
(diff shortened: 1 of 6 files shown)


opengever/dossier/locales/de/LC_MESSAGES/opengever.dossier.po
```diff
@@ -20,6 +20,10 @@ msgstr ""
 msgid "${copied_items} Elements were moved successfully"
 msgstr "${copied_items} Elemente wurden erfolgreich verschoben"
 
+#: ./opengever/dossier/behaviors/protect_dossier.py
+msgid "A dossier manager must be selected  when protecting a dossier"
+msgstr "Ein Dossier-Verwalter muss ausgewählt werden um das dossier schützen zu können."
```
lukasgraf: `dossier` -> `Dossier`

lukasgraf:
Oh, and changelog is missing.

njohner:
Ok all set for re-review.

Issue #4525, opened 2018-07-02
Bump plone.restapi to 3.x


lukasgraf:
Bump `plone.restapi` to a 3.x release.

This will address issues with the content creation process.

Expected work:

- Monkey patch `plone.restapi`'s new deserializer for proper default value handling.
   Until now, content created via `plone.restapi` was covered by our default value patches in `opengever.core` because [we patch `invokeFactory`](https://github.com/4teamwork/opengever.core/blob/3e9509988db2d35852ec17d0c6a96def01a9c80c/opengever/base/monkey/patches/default_values.py#L152-L190), which `plone.restapi` was using. Since that's not the case any more, `plone.restapi` probably needs to be patched separately. 
There's [tests that should fail](https://github.com/4teamwork/opengever.core/blob/3e9509988db2d35852ec17d0c6a96def01a9c80c/opengever/base/tests/test_default_values_for_types.py#L560-L583) for this. 

- Update any custom deserializers in GEVER
  - [Mail](https://github.com/4teamwork/opengever.core/blob/master/opengever/api/mail.py#L19-L49)

- Code invoking deserializers (opengever//ech0147/utils.py) might also need to be updated.

lukasgraf:
@buchi as discussed, I think it makes sense for you to take a stab at this, since you still have the changes to the deserializer interface freshly in memory. Happy to help with GEVER-specific oddities though, and the default value stuff.

lukasgraf:
Fixed via #4702

PR #4702, opened 2018-08-09
Bump plone.restapi to 3.4.1

phgross:
The new plone.restapi version contains some important fixes in the manner how contents gets created, thus we want to update to latest version (#4525).

Note: I add the necessary changes in different commits, although the tests pass only with all fixes, but that makes it easier to review and understand.

The following changes has been necessary:
 - Patching plone.restapi JSON deserializer to make sure default_values are set correctly, when adding content with plone.restapi
 - Update scanin endpoint, replace `rename` with the `add` helper method
 - Update our own mail json-deserializer 
 - Adjust ech0147, passing by the `create` flag when creating content via rest api.
 - Convert pyxb dates to real python datetime dates when adding content via eCH-0147 import
 - Adjust documentation about datetimes since restapi returns now always UTC datetimes. 

Further descriptions in the commit messages.


opengever/base/monkey/patches/default_values.py
```diff
@@ -385,3 +385,32 @@ def update_field(self, obj, field, item):
...
+        from opengever.base.default_values import set_default_values
+        from plone.restapi.deserializer import json_body
+
+        def __call__(self, validate_all=False, data=None, create=False):
+            if data is None:
+                data = json_body(self.request)
+
+            # set default values
+            container = aq_parent(aq_inner(self.context))
+            set_default_values(self.context, container, data)
```
lukasgraf: In this patch setting of default values happens *before* setting the actual values. In all the other default value patches we do it the other way round, first set the actual values, and then "fill in any holes" using `set_default_values`.

I would consider changing the order to do it the same way here. In terms of events it should be fine, since the `ObjectCreatedEvent` doesn't get fired inside the deserializer any more, but [afterwards instead](https://github.com/plone/plone.restapi/blob/38069323cb5737c347878ee934d26ff8293f0965/src/plone/restapi/services/content/add.py#L80-L81).

I also think setting default values should only happen when `create=True` - otherwise, we would also fill in any missing default values on `PATCH`, even if the user didn't specify those fields in their request. That wouldn't necessarily be bad, but it's not how the rest of the default value patches work - we only make sure things work out correctly on creation. After that, it's not the default value patches' responsibility any more.

docs/public/dev-manual/api/serialization.rst
```diff
@@ -19,8 +19,8 @@ Datum / Zeit
 --------------
 
 Da JSON keinen nativen Support für datetimes bietet, werden Python/Zope
-datetimes zu `ISO 8601 <https://de.wikipedia.org/wiki/ISO_8601>`_ Strings
-serialisiert.
+datetimes in UTC-Timezone `ISO 8601 <https://de.wikipedia.org/wiki/ISO_8601>`_
```
lukasgraf: I would also add a comment in the documentation changelog (`dev-manual/api/docs_changelog.txt`):
```
2018-08-09
----------

- REST API: datetime-Felder werden neu in UTC :doc:`serialisiert <serialization>` statt Lokalzeit.
```

docs/public/dev-manual/api/serialization.rst
```diff
@@ -19,8 +19,8 @@ Datum / Zeit
...
+Strings serialisiert.
```
lukasgraf: I would maybe phrase this like this:
```
Da JSON keinen nativen Support für datetimes bietet, werden Python/Zope
datetimes zu `ISO 8601 <https://de.wikipedia.org/wiki/ISO_8601>`_ Strings
serialisiert. Als Zeitzone wird bei der Serialisierung immer UTC verwendet,
nicht Lokalzeit.
```

Also, the table with examples just below this sections needs to be adjusted.

opengever/base/monkey/patches/default_values.py
phgross: As discussed, moving the `set_default_values` at the end of the deserialization is not possible, because the validation, which is done during the deserialization would raise for required fields without input but a default value.

phgross:
@lukasgraf updated ...

PR #4705, opened 2018-08-09
Bump Chameleon to 3.3 in order to fix edge case in exception handler causing recursion.

lukasgraf:
This fixes a recursion bug in Chameleon's `ExceptionFormatter`. In `3.2`, some exceptions that happened during template rendering got garbled in this way:

- The exception **message** got swallowed. Instead all you got was something like `<unprintable AttributeError object>` (no information as to what the attribute in question was)
- The template call stack recursed and was repeated over and over again

---

malthe/chameleon#260 fixes this. Instead of

`MissingParameterError: <unprintable MissingParameterError object>`

we now get

`MissingParameterError: Missing required parameter(s): obj`

(and the endless repetitions in the template call stack are gone).


lukasgraf:
@phgross IMHO we should seriously consider backporting this one. Otherwise we're going to have messed up exceptions in older releases for quite a while (ZG especially).

PR #4731, opened 2018-08-13
Exclude PreconditionFailed from Error log.

### LINKED ISSUE #4730: PreconditionFailed
https://sentry.4teamwork.ch/sentry/onegov-gever/issues/1517/

```
PreconditionFailed: 
(1 additional frame(s) were not displayed)
...
  File "ZPublisher/mapply.py", line 77, in mapply
    if debug is not None: return debug(object,args,context)
  File "home/zope/eggs/opengever.core-2017.6.8-py2.7.egg/opengever/document/document.py", line 335, in UNLOCK
    response = super(Document, self).UNLOCK(REQUEST, RESPONSE)
  File "webdav/Resource.py", line 650, in UNLOCK
    result = cmd.apply(self, token, url)
  File "webdav/davcmds.py", line 466, in apply
    raise PreconditionFailed

PreconditionFailed
```

njohner:
`PreconditionFailed` errors should be managed client side and can therefore be ignored on the GEVER side.

resolves #4730

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -18,6 +18,7 @@ Changelog
 - Only show create proposal button if meeting feature is enabled. [njohner]
 - Add action to switch to new GEVER UI. [njohner]
 - Add generic view to retrieve the value of a setting. [njohner]
+- Exclude PreconditionFailed from Error log. [njohner]
 - Lock oder of avaialbe roles in sharing endpoint. [phgross]
 - Fix archival_file and public_trial checks on documents overview. [phgross]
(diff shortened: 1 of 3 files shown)


njohner:
Sorry, that was pretty sloppy from me.
Should be ok now. I tested both a new install and the upgrade step.
I also modified the upgrade step to make sure we were adding two tuples, as it seems this can lead to type conflicts as suggested by @jone 's commit message https://github.com/4teamwork/opengever.core/commit/4c0a2dcea25dda98b43cb93c75c77a05f0f794b1

PR #4734, opened 2018-08-14
Sharing view improvements and IE 11 fixes

phgross:
- No longer use arrow functions and Object.values, there are not supported by IE 11.
- Drop sharing.js from the sharing template. The sharing.js is included globally (registered in the js registry), so no need to include them in the template.
- Show assignments_info button (?) only on rows with automatic roles


phgross:
Follow up Issue: https://github.com/4teamwork/opengever.core/issues/4737

PR #4739, opened 2018-08-14
Make sure widgets properly return persisted missing values

### LINKED ISSUE #4727: Widgets fall back to default_value if persisted value is equal to missing value
For fields that

- have a **default value** (`field.default` or `field.defaultFactory`)
- are **not required**

`z3c.form` widgets will perform a [misguided fallback](https://github.com/zopefoundation/z3c.form/blob/3.2.11/src/z3c/form/widget.py#L114-L118) (both in display and edit mode) to the field's default value.

This has the effect that for fields that get their value removed by the user (which is allowed, because they're not required), the persisted value will be set to `missing_value` (`None` for example), but the widget will show the default value.

This is mainly problematic in cases where the field's default is different from missing value - if they're the same, the condition still doesn't make a lot of sense, but ends up with the same value anyway (value == missing_value == default value).

#### Steps to reproduce

- Create a new document. The document will have its `document_date` set to today by default, and that value will be persisted
- Edit the document, and remove the `document_date`. Save.
- Check that the absence of a value has been persisted (`doc.__dict__['document_date']` should now be `None`).
- Yet in the document overview today's date will be rendered for`document_date`, and it will also be suggested in the edit form (overriding `None` if the form is saved)

lukasgraf:
Make sure widgets properly return persisted missing values for optional fields with default values:

In `z3c.form.widget.Widget.update` there is a [misguided check](https://github.com/zopefoundation/z3c.form/blob/3.2.11/src/z3c/form/widget.py#L114) that falls back to using a **default value** if a fields actual value is **equal to missing value**.

This leads to a bug where for fields that are

1) optional (not required)
2) have a default of some sort (field.default or field.defaultFactory)
3) have a legitimate missing value persisted
4) have a MV that is different from the default

an incorrect value is returned by the widget (the default value instead of the legitimate missing value).

This will result in display views that render an object and use the widget's display mode for that to render an incorrect field value (even though underneath, on the persistent object, the MV is properly persisted).

In addition, in an edit form the widget will incorrectly cause the default to be filled in instead of leaving the field at its legitimate MV. This will then result in the persisted MV being overriden by the default.

This change monkey patches `z3c.form.widget.Widget.update` in a way where this strategy is only applied for REQUIRED fields:

Only if the field is required AND the persisted value is equal to MV will it return the default - otherwise, the persisted value is used.

Fixes #4727

### DIFF
--- a/opengever/base/monkey/patches/default_values.py
+++ b/opengever/base/monkey/patches/default_values.py
@@ -257,6 +257,98 @@ class PatchZ3CFormChangedField(MonkeyPatch):
         self.patch_refs(util, 'changedField', changedField)
 
 
+class PatchZ3CFormWidgetUpdate(MonkeyPatch):
+    """
+    """
+
+    def __call__(self):
+        import zope.interface
+        import zope.component
+        import zope.location
+        import zope.schema.interfaces
+        from z3c.form import interfaces
+        from z3c.form.widget import PLACEHOLDER
+
+        def update(self):
+            """See z3c.form.interfaces.IWidget."""
+            # Step 1: Determine the value.
+            value = interfaces.NO_VALUE
+            lookForDefault = False
+            # Step 1.1: If possible, get a value from the request
+            if not self.ignoreRequest:
+                #at this turn we do not need errors to be set on widgets
+                #errors will be set when extract gets called from form.extractData
+                self.setErrors = False
+                widget_value = self.extract()
+                if widget_value is not interfaces.NO_VALUE:
+                    # Once we found the value in the request, it takes precendence
+                    # over everything and nothing else has to be done.
+                    self.value = widget_value
+                    value = PLACEHOLDER
+            # Step 1.2: If we have a widget with a field and we have no value yet,
+            #           we have some more possible locations to get the value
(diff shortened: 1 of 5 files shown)


opengever/base/monkey/patches/default_values.py
```diff
@@ -257,6 +257,95 @@ def changedField(field, value, context=None):
...
+                        interfaces.IDataManager).query()
+                # Step 1.2.2: If we still do not have a value, we can always use
+                #             the default value of the field, if set
+                # NOTE: It should check field.default is not missing_value, but
+                # that requires fixing zope.schema first
+                # We get a clone of the field with the context binded
+                field = self.field.bind(self.context)
+
+                # PATCHED:
+                # Only look for a default value if value is equal to missing_value
```
phgross: The condition is also true when the value is interfaces.NO_VALUE. Maybe a shor tnote in which case thats true would be helpful.

lukasgraf:
@phgross updated and rebased - this one is ready from my point of view. Did some extensive testing for the affected fields mentioned in https://github.com/4teamwork/opengever.core/issues/4727#issuecomment-412959708 - seems to work as intended.

PR #4740, opened 2018-08-15
Fix file conversion to PDF for mails

### LINKED ISSUE #4738: ArchivalFileConverter failing for OGMail
https://sentry.4teamwork.ch/sentry/onegov-gever/issues/3492/

```
AttributeError: file
(2 additional frame(s) were not displayed)
...
  File "home/zope/eggs/ftw.bumblebee-3.5.0-py2.7.egg/ftw/bumblebee/browser/callback.py", line 32, in __call__
    self.handle_success(mimetype, file_data)
  File "home/zope/eggs/opengever.core-2018.3.5-py2.7.egg/opengever/bumblebee/browser/callback.py", line 12, in handle_success
    ArchivalFileConverter(self.context).store_file(file_data, mimetype)
  File "home/zope/eggs/opengever.core-2018.3.5-py2.7.egg/opengever/document/archival_file.py", line 50, in store_file
    filename=self.get_file_name())
  File "home/zope/eggs/opengever.core-2018.3.5-py2.7.egg/opengever/document/archival_file.py", line 63, in get_file_name
    filename, ext = os.path.splitext(self.document.file.filename)

AttributeError: file
```

njohner:
Fix `ArchivalFileConverter` for `OGMail`. The converter was trying to access the filename through the `file` attribute, which only exists for documents, but not for mails.

I also fixed a test which did not convert the correct document.

resolves #4738 
Sentry Issue : https://sentry.4teamwork.ch/sentry/onegov-gever/issues/3492/


opengever/mail/tests/test_archival_file.py
```diff
@@ -0,0 +1,28 @@
...
+class TestArchivalFileForMails(TestArchivalFile):
+
+    def setUp(self):
+        FunctionalTestCase.setUp(self)
+        self.document = create(Builder('mail')
+                               .titled(u'\xdcberpr\xfcfung XY')
+                               .with_dummy_message())
+        self.grant('Manager')
+
+    def test_trigger_conversion_skip_files_in_manually_state(self):
```
phgross: 🤔 does this really test the filename getter for mails? or is just the test title wrong?

njohner: Not sure I understand the question. The title does not say that it tests the filename getter.
The test class inherits from `TestArchivalFile`, hence tests different aspects of the archival file, that were only tested for documents. It indirectly tests the filename getter, as it tests the the `ArchivalFileConverter` works for mails (several of these tests would fail without the correction to the `ArchivalFileConverter`).
I overwrite `test_trigger_conversion_skip_files_in_manually_state` as this one generates its own document for the test, hence needs to be overwritten to test this on a mail. 
Of course we could also just test the filename getter and make sure the `ArchivalFileConverter` works for mails if you think it's overkill to also test the form, states and so on for mails.

PR #4743, opened 2018-08-15
Fix inline-cell-icons since fontawesome update

bierik:
Closes https://github.com/4teamwork/opengever.core/issues/4715

screen shot 2018-08-15 at 16 35 32


phgross:
@bierik please rebase and 🚢

PR #4767, opened 2018-08-21
Add modification and creation date to documents listing

### LINKED ISSUE #4557: OGIP 27: Änderungs- und Erstelldatum Spalte in Dokumentauflistungen
- [ ] Das bereits existierende «Änderungsdatum» (`modified`) wird neu in der Auflistung der Dokumente angezeigt
- [ ] Das bereits existierende «Erstelldatum» (`created`) wird neu in der Auflistung der Dokumente angezeigt.
- [ ] Sofern technisch möglich wird es standardmässig ausgeblendet.

njohner:
We add the `created` and `modified` columns to the `Documents` listing (and by inheritance to all the other document listings).
We add the colums in the `hidden` mode, so that by default, these columns are not shown.

I also removed a duplicate definition of `columns` in the `MyDocuments` class, where it was first defined both as a class attribute and as a property. This seems to have been that way for quite some time and even both were modified several times, so is there any reason for that?

resolves #4557


njohner:
On a side note, if a user already has preferences stored for the listing, if he adds the two new columns, they will appear at the end of the table, after the empty column, which means after a little empty space.

njohner:
Yes right it can dropped

PR #4774, opened 2018-08-22
Add Activity settings for Dispositions.

### LINKED ISSUE #4498: OGIP 29: Benachrichtigungen des Aussonderungsmodul konfigurierbar machen
Die bereits bestehenden Benachrichtigungen für Events im Aussonderungsmodul sollten auch konfigurierbar sein:
 - [ ] Separater Reiter Aussonderung
 - [ ] Reiter nur darstellen falls entsprechende Rolle vorhanden (`Manager`, `Archivist` oder `Records Manager` (separate Berechtigung)

njohner:
We add a new tab to the activity settings for dispositions. The Tab is conditionally only showed when the user is one of `Manager`, `Archivist`, `Records Manager`.

I'm a bit confused with the idea of having a separate permission to manage whether a user sees the `Disposition` tab in the Activity settings or not. In which workflow would that be? Or should we make a new workflow for that? For now I'm just checking the permissions in the `NotificationSettingsForm` view.

Also I noticed that the notifications go to all `Archivist` and to the creator of the `Disposition`, but not to the `Records Manager`. I guess the creator is usually a `Records Manager`?

resolves #4498


njohner:
Ok all done.

PR #4777, opened 2018-08-23
Add creation and modification date to document overview.

### LINKED ISSUE #4558: OGIP 27: Änderungs- und Erstelldatum in Detailansicht darstellen
bildschirmfoto_2018-07-06_um_11_04_35

> \> Das Änderungsdatum wird neu in der Detailansicht des Dokuments angezeigt.
> \> Das Erstelldatum wird neu in der Detailansicht des Dokuments angezeigt.

njohner:
screen shot 2018-08-23 at 11 33 59

resolves #4558

### DIFF
--- a/opengever/document/browser/overview.py
+++ b/opengever/document/browser/overview.py
@@ -4,6 +4,7 @@ from ftw.bumblebee.interfaces import IBumblebeeDocument
 from opengever.base import _ as ogbmf
 from opengever.base.browser import edit_public_trial
 from opengever.base.browser.helper import get_css_class
+from opengever.base.behaviors.changed import IChanged
 from opengever.bumblebee import is_bumblebee_feature_enabled
 from opengever.document import _
 from opengever.document.behaviors.metadata import IDocumentMetadata
@@ -144,6 +145,10 @@ class Overview(DefaultView, GeverTabMixin, ActionButtonRendererMixin):
             FieldRow('IDocumentMetadata.document_date'),
             TemplateRow(self.file_template,
                         label=_('label_file', default='File')),
+            CustomRow(self.get_creation_date,
+                      label=_('label_created', default='Created')),
+            CustomRow(self.get_modification_date,
+                      label=_('label_modified', default='Modified')),
             FieldRow('IDocumentMetadata.document_type'),
             FieldRow('IDocumentMetadata.document_author'),
             CustomRow(self.render_creator_link,
@@ -329,3 +334,11 @@ class Overview(DefaultView, GeverTabMixin, ActionButtonRendererMixin):
             pass
 
         return 'contenttype-opengever-document-document'
+
+    def get_creation_date(self):
+        return self.context.toLocalizedTime(
+            self.context.created(), long_format=True)
+
+    def get_modification_date(self):
+        return self.context.toLocalizedTime(
+            IChanged(self.context).changed, long_format=True)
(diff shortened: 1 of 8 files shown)


lukasgraf:
When using Plone's `modified()`, this will also be affected by internal changes like reindexing the object, unless we specifically prevent this everywhere where it's being triggered (upgrade steps, events, ...).

@phgross Is this what we want / would this be acceptable to our customers?

njohner:
Oh that does not sound good... If reindexing sets the modification date, that might put some dossiers into forbidden states (dossier earliest possible end date later than dossier end date).

njohner:
Here is a summary of our discussion with @jone  and @phgross:
==============================================

* Using `modified` is not a good idea:
  * it gets changed by Plone, for example when reindexing an object
  * is also used in caching mechanisms, so controlling ourselves when it gets modified or not could lead to other issues

**Conclusion** 
we will have to introduce a new index, metadata and field for every object (let's call it `changed`). We need the index for tables, the metadata for solr, and the field for the API.

Implementation
---------------
**Considerations**
* While adding a new index is fast, adding a metadata is not.
* We want to add the metadata in a deferrable upgrade step
* Gever should obviously still work if the deferrable upgrade has not been run yet

**Proposed implementation**
* a `behavior` is used to get and set `changed`
  * getter has a fallback to `modified` if `changed` is not set
  * the behavior also provides a method to check whether the deferrable upgrade has been completed
* A first upgrade adds the `changed` index (copy of `modified`) and the empty metadata
* A second, deferrable, upgrade fills the metadata:
  * set field value via behavior
  * update metadata
  * reindex the `changed` index
  * set the flag in the behavior that upgrade has been completed
* table helper for the `changed` column, which uses the fallback
* in the table setup, if solr is used and the upgrade has not been run, do not display the `changed` column, as this would not work
* Event handlers
  * when `modified` gets changed and the upgrade step has not been run, copy `modified` to `changed`
  * set `changed` when desired (for documents maybe everytime there is a journal entry??)
* Add field to the API

lukasgraf:
Proposed implementation sounds good to me 👍

jone:
@njohner sounds good 👍
I have realized that it might not be very smart to put the "upgrades installed?" info on the behavior: the behavior can only be used by adapting dexterity content.
If we want to use this info in the table helper we would need to adapt an actual dexterity object. In the helper we usually only have the brain or the solr flair, and when used for `Persönliche Übersicht` the context is the Plone site and not a dexterity object.

Proposal:
Provide an importable function, implemented and availiable in the module of the behavior. The function does not need to require any arguments; it can use `getSite()` for getting the info.

njohner:
Ok I updated this PR to make use of the newly defined `changed` field (see #4962)

screen shot 2018-10-30 at 14 02 38

This should be ready to ship.

opengever/document/browser/overview.py
```diff
@@ -329,3 +334,9 @@ def get_archival_file_class(self):
             pass
 
         return 'contenttype-opengever-document-document'
+
+    def get_creation_date(self):
+        return self.context.created().strftime("%d.%m.%Y %H:%M")
```
phgross: Is it possible to use plone's toLocalizedTime here?

opengever/document/browser/overview.py
```diff
@@ -329,3 +334,9 @@ def get_archival_file_class(self):
...
+
+    def get_modification_date(self):
+        return IChanged(self.context).changed.strftime("%d.%m.%Y %H:%M")
```
phgross: same here

njohner:
Ok all done, this should be ready to ship.

PR #4778, opened 2018-08-23
Do not set document date during checkin or cancel.

### LINKED ISSUE #4560: OGIP 27: Dokumentdatum beim Checkin resp. Wiederrufen nicht mehr anpassen
> Das «Dokumentdatum» wird nicht mehr vom System gesetzt, die Aktualisierung des Dokument-Datums beim Checkin und Wiederrufen wird also entfernt.

njohner:
Document date should not be set by the system anymore, but only by the user. The modification date is already available in the `modified` attribute.

resolves #4560

### DIFF
--- a/opengever/document/tests/test_checkout.py
+++ b/opengever/document/tests/test_checkout.py
@@ -79,13 +79,15 @@ class TestCheckinIntegration(IntegrationTestCase):
 class TestCheckin(FunctionalTestCase):
     """Tests for the checkin functionality."""
 
+    document_date = date(2014, 1, 1)
+
     def setUp(self):
         super(TestCheckin, self).setUp()
         self.dossier = create(Builder('dossier'))
 
         self.document = create(Builder('document')
                                .with_dummy_content()
-                               .having(document_date=date(2014, 1, 1))
+                               .having(document_date=self.document_date)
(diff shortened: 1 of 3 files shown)


opengever/document/tests/test_checkout.py
```diff
@@ -122,11 +122,6 @@ def test_clear_locks(self):
         self.manager.checkin()
         self.assertFalse(IRefreshableLockable(self.document).locked())
 
-    def test_document_date_is_updated_to_current_date(self):
```
Rotonen: One should test the inverse here instead of just dropping the test.

njohner: Sure, I'll test the negative. I have to say I hesitated here, but decided a regression did not seem very likely.

Rotonen: Desired business logic should be tested. Stuff like workflows can throw a wrench into the gears in the far future.

njohner: 👍

njohner:
This is ready for (re)review

njohner:
@Rotonen Could you have another look at this one?

Rotonen:
@njohner rebase and merge.

PR #4782, opened 2018-08-24
Add proposal tabs to repository folders

### LINKED ISSUE #4411: Tab Anträge auf Ordnungspositionen
Mit https://github.com/4teamwork/opengever.core/pull/3868 wurden Dokumente und Aufgaben auf Ordnungspositionen dargestellt. Für Anträge soll nun auch ein solches Tab eingebaut werden da hier sehr ähnliche Anforderungen bestehen.

- Tab nur darstellen, wenn das `meeting` Feature aktiviert ist.
- Neues Flag, damit Tab deaktiviert werden kann (Analog Dokumente/Aufgaben) auch wenn `meeting` Feature aktiv ist. Per Default soll das Tab aber erscheinen (aber nur falls `meeting` Feature aktiv).
- Neue Spalte Dossier im Antrags-Tab, so dass direkt zum Dossier navigiert werden kann
- Reihenfolge beachten, gleich wie im Dossier, sprich nach Aufgaben-Tab
- In Gever-Config (`IGeverSettings`) View aufnehmen, z.B. als Option `repositoryfolder_proposals_tab`

Ref: https://basecamp.com/2768704/projects/13732725/todos/349420332#comment_626191116.

Rotonen:
* Add a registry flag
* Add the tab

Closes #4411

### DIFF
--- a/opengever/policytemplates/policy_template/opengever.+package.name+/opengever/+package.name+/profiles/default/registry.xml.bob
+++ b/opengever/policytemplates/policy_template/opengever.+package.name+/opengever/+package.name+/profiles/default/registry.xml.bob
@@ -79,6 +79,12 @@
   </records>
 
 {{% endif %}}
+{{% if setup.repositoryfolder_proposals_tab %}}
+  <records interface="opengever.repository.interfaces.IRepositoryFolderRecords">
+    <value key="show_proposals_tab">True</value>
+  </records>
+
+{{% endif %}}
 {{% if setup.workspace %}}
   <records interface="opengever.workspace.interfaces.IWorkspaceSettings">
     <value key="is_feature_enabled">True</value>
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,8 @@ Changelog
 ---------------------
 
 - Correct info messages in meetings. [njohner]
+- Add proposal tabs to repository folders. [Rotonen]
+- Fix global scoped "show more" links for SOLR livesearch [Rotonen]
 - Only set proposal to decided when excerpt has been generated and returned. [njohner]
 - Add actions pointing to the meeting and protocol debug views. [njohner]
 - Fix bug with document without file in a dossiertemplate. [njohner]
--- a/opengever/api/tests/test_config.py
+++ b/opengever/api/tests/test_config.py
@@ -55,6 +55,7 @@ class TestConfig(IntegrationTestCase):
                 u'preview_open_pdf_in_new_window': False,
                 u'purge_trash': False,
                 u'repositoryfolder_documents_tab': True,
+                u'repositoryfolder_proposals_tab': True,
                 u'repositoryfolder_tasks_tab': True,
                 u'resolver_name': u'strict',
                 u'sablon_date_format': u'%d.%m.%Y',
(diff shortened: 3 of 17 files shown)


opengever/policytemplates/policy_template/opengever.+package.name+/opengever/+package.name+/profiles/default/registry.xml.bob
```diff
@@ -78,6 +78,12 @@
     <value key="show_tasks_tab">True</value>
   </records>
 
+{{% endif %}}
+{{% if setup.repositoryfolder_proposals_tab %}}
```
deiferni: did you try creating a new policy? i think you also should add a question for this setting to the mr bob ini file.

opengever/policytemplates/policy_template/opengever.+package.name+/opengever/+package.name+/profiles/default/registry.xml.bob
```diff
@@ -78,6 +78,12 @@
...
+  <records interface="opengever.repository.interfaces.IRepositoryFolderRecords">
+    <value key="show_proposals_tab">True</value>
```
deiferni: these settings mirror the defaults, i think it would make more sense to invert above `if` and only include the setting if the feature flag is `False`.

deiferni:
I would re-use the proposals tab from the dossier, iirc there are no actions but you can re-use the columns it defines.

opengever/policytemplates/policy_template/opengever.+package.name+/opengever/+package.name+/profiles/default/registry.xml.bob
Rotonen: Good catch.

opengever/policytemplates/policy_template/opengever.+package.name+/opengever/+package.name+/profiles/default/registry.xml.bob
Rotonen: I'll mirror how the other similar settings are done. The default comes in from the ini file.

deiferni: The ini file has a different default than the interface schema definition for these fields. IMHO this is no good. That forces `[person that creates policy]` to be aware of these different defaults and to input accordingly. I would strongly prefer to use the same defaults in the mrbob ini file and in the schema definitions  so input can be omitted when creating a policy and "good" defaults are chosen.

I'm expecting the schema defaults to be chosen "wisely", so i would make sure that our policy templates uses the schema defaults.

Rotonen: How about I just drop this from here and we'll do a separate sweep of the policy templates?

deiferni: if you let it follow this PR immediately that'd be fine 😄. otherwise i prefer to have it fixed here since i consider it dangerous and not correct. also inverting the questions/statements is not complex and should not involve much effort.

Rotonen: I'm expecting this to have many cousins in the templates. But sure, can address just this one for now.

PR #4801, opened 2018-08-29
OGIP 45: Revoke roles when reassign task

phgross:
This is a missing, but essential part of the OGIP 45 work.

Because it can now happen, that the user which is reassigning a task, is no longer permitted to access them after the reassign, we add a special fallback and redirect to the personaloverview instead of the task and show a corresponding message.

Currently the reassign work is all implemented in the form itself. I don't like this fact, but that will be part of the #4786 work.


opengever/task/tests/test_assign.py
```diff
@@ -135,13 +140,51 @@ def test_modify_event_is_fired_but_only_once(self, browser):
...
+        self.login(self.regular_user, browser=browser)
+
+        responsible = 'fa:{}'.format(self.secretariat_user.getId())
+        self.assign_task(responsible, u'Thats a job for you.')
+
+        manager = RoleAssignmentManager(self.task)
+        self.assertEqual(
+            [{'cause': 1,
+              'roles': ['Editor'],
+              'reference': u'plone:1014093300',
```
deiferni: is this a hardcoded oguid? this may become cumbersome to update as the int-id changes when we adapt the fixture. is the reference important for the test? you could omit it and only test for a subset with `assertDictContainsSubset`.

opengever/task/tests/test_assign.py
```diff
@@ -195,3 +238,43 @@ def test_syncs_successor_when_reassigning_predecessor(self, browser):
...
+        form = browser.find_form_by_field('Responsible')
+        form.find_widget('Responsible').fill(
+            'fa:{}'.format(self.secretariat_user.getId()))
+        browser.find('Assign').click()
+
+        manager = RoleAssignmentManager(self.task)
+        self.assertEqual(
+            [{'cause': 1,
+              'roles': ['Editor'],
+              'reference': u'plone:1014093300',
```
deiferni: dito.

opengever/task/tests/test_assign.py
phgross: @deiferni Intids and so also oguids are fixed by using `time_based_intids` in the fixture. 

But it's easier to read when using `Oguid.for_object(self.task).id` so I updated the tests anyways.

deiferni: yes, they will change if the fixture changes though e.g. if more objects are created before the object you test against. if more and more tests hard-code the int-id the number of tests that could be affected when the fixture changes increases. thus i would prefer if we avoid hard-coding the int-id where it is not required (or does not benefit) by the test-case.

short version: i like your new suggestion that uses `Oguid` 😉 👍

PR #4805, opened 2018-08-29
Add activity for proposal unscheduled transition.

njohner:
There was no ativity defined for the transition 'scheduled-submitted'. Instead, the notification sent to the issuer when removing a proposal from a meeting schedule was the same as when the proposal got scheduled.

**personal-preferences view with the added setting**
screen shot 2018-08-29 at 16 00 40

**correct message is now displayed in the notifications**
screen shot 2018-08-29 at 16 01 54


opengever/activity/__init__.py
```diff
@@ -151,6 +153,7 @@ def send_digest_zopectl_handler(app, args):
 ACTIVITIES_ICONS = {
     'proposal-transition-reject': 'rejected',
     'proposal-transition-schedule': 'scheduled',
+    'proposal-transition-unschedule': 'unscheduled',
```
deiferni: this causes a naming conflict with proposals that are not yet scheduled, we call them unscheduled, too. i would not use unscheduled in another context for proposals. what do you think about deschedule 🤔?

Rotonen: How about 'pulled'?

deiferni: as just discussed `pulled` seems to be used in meetings and is unique enough. so we'll go for it.

njohner:
Ok this should be ready to ship. I tested the upgrade again after the change in naming and everything seems to still work.

PR #4813, opened 2018-08-30
REST API livesearch endpoint

buchi:
Supports Portal Catalog and Solr

### DIFF
--- a/opengever/api/livesearch.py
+++ b/opengever/api/livesearch.py
@@ -1,18 +1,13 @@
 from opengever.base.interfaces import ISearchSettings
 from opengever.base.solr import OGSolrContentListing
 from plone import api
-from plone.registry.interfaces import IRegistry
 from plone.restapi.services.search.get import SearchGet
 from zope.component import getMultiAdapter
-from zope.component import getUtility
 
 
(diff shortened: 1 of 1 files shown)


opengever/api/tests/test_livesearch.py
```diff
@@ -1,10 +1,5 @@
-from ftw.builder import Builder
```
phgross: please squash this commit to the first commit.

opengever/api/tests/test_livesearch.py
```diff
@@ -0,0 +1,52 @@
...
+from ftw.testbrowser import browsing
+from opengever.base.model.favorite import Favorite
+from opengever.base.oguid import Oguid
+from opengever.testing import IntegrationTestCase
+import json
+
+
+class TestLivesearchGet(IntegrationTestCase):
+
+    @browsing
```
phgross: The solr way is not tested at all ...

phgross: As discussed, solr-testing will be discussed and implemented  in the context of #4844, so we go without solr tests for now. ^

PR #4815, opened 2018-08-30
Bump lxml to 4.1.1

Rotonen:
Our lxml issues seem to have been entirely testbrowser-sided so far.

On the previous bump round we hit https://github.com/lxml/lxml/pull/199 in 3.6.1. Seems we're either not dealing with file fields the same way in the tests anymore or `ftw.testbrowser` has fixed something in the meanwhile.

As 4.0.0 and upwards promise speedups, I'm bumping this to as far as we can get to currently.

https://lxml.de/4.2/changes-4.2.4.html

4.2.0 introduces https://github.com/lxml/lxml/pull/255 prepopulated-per-default radio widgets, which break how some of our tests are built.

For example
https://github.com/4teamwork/opengever.core/blob/0f4281fbb75313786f008dfb81a3f8a700f7e245/opengever/ech0147/tests/test_export.py#L68-L74
is now properly prepopulated and does not return an error, but a valid zip file instead. This can be fixed with a testing style change, but that's a bit subtle of a change so not wishing to go there yet.

I've amended the testbrowser lxml issue https://github.com/4teamwork/ftw.testbrowser/issues/70 with this detail as well.


Rotonen:
@deiferni no build-time considerations.

lukasgraf:
I'm somewhat surprised that this one passes - I could have sworn I tried to update `lxml` in the past to a non-ancient version, and lots of things fell apart.

But since tests pass, I'm more than happy to bump it.

It seems that even for Plone 4.3 (the upcoming [4.3.18](http://dist.plone.org/release/4.3.18-pending/versions.cfg) specifically) `lxml` also got bumped, to `4.2.1`. More so [by accident](https://github.com/plone/buildout.coredev/commit/a639bda31d143df14d1df938eb3edb1b44fa8066) than intentionally, but I'll take it 😉 

But once we update Plone to 4.3.18, than means we'll need to tackle the testbrowser issues you pointed out.

So, 👍 from be for the bump to `4.1.1`.

PR #4827, opened 2018-09-04
Trash/remove Protocol Excerpts

### LINKED ISSUE #4486: Protokollauszüge "löschen"
Im Moment kann man Protokollauszüge nicht löschen/stornieren. Aus Benutzersicht soll dies folgendermassen funktionieren.

- Protokollauszüge können im Sitzungsdossier in den Papierkorb verschoben werden
- Diese Protokollauszüge, die im Papierkorb sind, sollten auf der Sitzungsansicht nicht mehr aufgelistet werden
- Werden die Protokollauszüge aus dem Papierkorb gelöscht wird die Relation zum Traktandum/Antrag entfernt

Umsetzung, Protokollauszüge nicht mehr darstellen:
- Objekte, die im Papierkorb sind, providen das `ITrashed` `Interface`
- Damit kann überprüft werden, ob die Dokumente dargestellt werden sollen oder nicht

Umsetzung Relation entfernen:
- Überprüfen ob `IObjectWillBeRemovedEvent` verwendet werden kann, ansonsten `IObjectRemovedEvent` brauchen.
- Bei Traktanden ohne Antrag wir die Relation in SQL gespeichert, siehe https://github.com/4teamwork/opengever.core/blob/44612e4a11a41a5340bd71c1a1506c97f2e970f6/opengever/meeting/model/excerpt.py. Der Zeile in der Tabelle muss in diesem Fall entfernt werden.
- Bei Traktanden mit Antrag wird die Relation als `RelationValue` gespeichert, siehe https://github.com/4teamwork/opengever.core/blob/master/opengever/meeting/proposal.py#L476, diese Relation muss beim Löschen entfernt werden.
- Sonderfall Protokollauszug schon zurückgeschickt: in diesem Fall soll das Löschen des Dokumentes aus dem Papierkorb
screen shot 2018-07-26 at 12 02 07
 verhindert werden, analog zu Dokumenten die Verweise von Aufgaben sind:

Ref: https://basecamp.com/2768704/projects/13732725/todos/350815610

njohner:
Trashing and/or removing protocol excerpts was not handled properly. Here we add the following features:
* Hide trashed excerpts from the meeting view
* when an excerpt is removed from trash, cleanup relations to proposal/agendaitem:
    * If the excerpt is from an ad-hoc agendaitem, remove its relation to the agendaitem from the sql database
    * If the excerpt is from an agendaitem with a proposal, remove the relation from the submitted proposal
* Forbid to trash an excerpt if it has been returned to the proposal.

screen shot 2018-09-04 at 14 48 57

resolves #4486

### DIFF
--- a/opengever/meeting/handlers.py
+++ b/opengever/meeting/handlers.py
@@ -1,3 +1,4 @@
+from Acquisition import aq_inner
 from opengever.base.browser.paste import ICopyPasteRequestLayer
 from opengever.base.model import create_session
 from opengever.base.oguid import Oguid
@@ -8,8 +9,11 @@ from opengever.meeting.command import UpdateExcerptInDossierCommand
 from opengever.meeting.model import GeneratedExcerpt
 from opengever.meeting.model import Proposal
 from opengever.meeting.model import SubmittedDocument
+from opengever.meeting.model.excerpt import Excerpt
+from opengever.meeting.proposal import ISubmittedProposal
 from opengever.meeting.sablontemplate import sablon_template_is_valid
 from plone import api
+from zc.relation.interfaces import ICatalog
 from zope.annotation.interfaces import IAnnotations
 from zope.component import getUtility
 from zope.component.interfaces import ComponentLookupError
@@ -36,6 +40,24 @@ def document_deleted(context, event):
     session = create_session()
     for doc in SubmittedDocument.query.by_document(context).all():
         session.delete(doc)
+
+
+def excerpt_delete(context, event):
+    # if it is an excerpt document from an agendaitem with a proposal,
+    # remove the excerpt from the submitted proposal.
+    for relation in getUtility(ICatalog).findRelations({
+            'to_id': getUtility(IIntIds).getId(aq_inner(context)),
+            'from_attribute': 'excerpts'}):
+        # We expect that there are 0 or 1 relation, because this document
+        # cannot be the excerpt of multiple proposals.
+        submitted_proposal = relation.from_object
+        assert(ISubmittedProposal.providedBy(submitted_proposal))
+        submitted_proposal.remove_excerpt(context)
+        return
+
+    # if it is an excerpt document from an agendaitem without a proposal
+    # remove it from the SQL database.
+    Excerpt.query.by_oguid(Oguid.for_object(context)).delete()
 
 
 def on_document_modified(context, event):
--- a/opengever/trash/remover.py
+++ b/opengever/trash/remover.py
@@ -4,7 +4,23 @@ from opengever.trash.trash import ITrashed
 from plone import api
 from zc.relation.interfaces import ICatalog
 from zope.component import getUtility
+from zope.component.interfaces import IObjectEvent
+from zope.component.interfaces import ObjectEvent
+from zope.event import notify
+from zope.interface import implements
 from zope.intid.interfaces import IIntIds
+
+
+class IObjectWillBeRemovedFromTrashEvent(IObjectEvent):
+    """Interface of an event which gets fired when removing a document from trash.
+    """
+
+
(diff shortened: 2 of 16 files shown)


opengever/meeting/handlers.py
```diff
@@ -38,6 +41,23 @@ def document_deleted(context, event):
...
 
+def excerpt_delete(context, event):
+    # if it is an excerpt document from an agendaitem with a proposal,
+    # remove the excerpt from the submitted proposal.
+    for relation in getUtility(ICatalog).findRelations({
+            'to_id': getUtility(IIntIds).getId(aq_inner(context)),
+            'from_attribute': 'excerpts'}):
+        # We expect that there are 0 or 1 relation, because this document
+        # cannot be the excerpt of multiple proposals.
+        submitted_proposal = relation.from_object
```
deiferni: this is quite dangerous as i think this may not do what you want. have you tested what happens if there are other relations to this document, e.g. from a task, or from another document? you should make sure that the object is indeed a submitted proposal.

since this is active for all documents you might otherwise bypass checks that are necessary and that should block the document from being removed. e.g. a non-excerpt document that is the target of a relation from another document should not be removable:

screen shot 2018-09-06 at 11 01 14

opengever/trash/remover.py
```diff
@@ -4,9 +4,25 @@
...
 from zc.relation.interfaces import ICatalog
 from zope.component import getUtility
+from zope.component.interfaces import IObjectEvent
+from zope.component.interfaces import ObjectEvent
+from zope.event import notify
+from zope.interface import implements
 from zope.intid.interfaces import IIntIds
 
 
+class IRemovedFromTrashEvent(IObjectEvent):
```
deiferni: this is fired before the object is removed, i would call this `IObjectWillBeRemovedFromTrashEvent`, plone has many of those `WillBe` events already and i think we should stick to the convention.

opengever/meeting/handlers.py
```diff
@@ -38,6 +41,23 @@ def document_deleted(context, event):
...
+        submitted_proposal.remove_excerpt(context)
+        return
+
+    # if it is an excerpt document from an agendaitem without a proposal
+    # remove it from the SQL database.
+    Excerpt.query.filter(Excerpt.excerpt_oguid == Oguid.for_object(context)).delete()
```
deiferni: you could add the method `by_oguid` to `ExcerptQuery` (`ExcerptQuery` may need to be created 😉), have a look at `ProposalQuery`, for example.

opengever/meeting/handlers.py
njohner: I think this is safe, as only submitted proposals have excerpts. This same assumption is also used in the document base. But if this changes in the future this would indeed break down, so I add an assertion.

opengever/meeting/handlers.py
njohner: Ok done. I needed to return the query, whereas for most others we have a `get_by_oguid` which returns the object itself. But with the different naming I don't think this is confusing.

deiferni: the naming convention i (or we) try to stick to is that if a `get` is in a query method name it will return an object (by calling `first`), otherwise it will build the query and return the query.

opengever/meeting/handlers.py
njohner: I'm actually not too sure how to test this for other relations, as I'm not sure in what situations relations are created

njohner:
Ok all done here

opengever/meeting/handlers.py
deiferni: ok, so upon closer inspection i think we should be guarded by the ```'from_attribute': 'excerpts'``` selector in the query. i think i overlooked that in my first comment as relations from other objects (e.g. other documents) will use another field.

PR #4829, opened 2018-09-04
Prevent pasting without `Copy or Move` on the target container

### LINKED ISSUE #4589: Probleme bei geschütztem Dossier mit Aufgabenzugriff
https://sentry.4teamwork.ch/sentry/onegov-gever/issues/8092/

Wird in einem geschützten Dossier, ein Benutzer mittels Aufgabe berechtigt, so wird ihm zwar die Möglichkeit geboten ein Dokument ins Dossier hinzuzufügen bzw. eine Dokument mittels  copy&paste einzufügen. Beim Versuch treten aber Berechtigungsproblem auf.

Rotonen:
* Amend the fixtures to have suitable data types
* Prevent pasting without `Copy or Move` on the target container as otherwise the freshly created object cannot be renamed

Closes #4589

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -607,6 +607,9 @@ Objects
             - self.meeting_subtask
         - self.protected_dossier
           - self.protected_document
+        - self.protected_dossier_with_task
+          - self.protected_document_with_task
+          - self.task_in_protected_dossier
     - self.empty_repofolder
   - self.templates
     - self.ad_hoc_agenda_item_template
--- a/README.rst
+++ b/README.rst
@@ -605,6 +605,8 @@ Objects
           - self.meeting_document
           - self.meeting_task
             - self.meeting_subtask
+        - self.protected_dossier
+          - self.protected_document
     - self.empty_repofolder
   - self.templates
     - self.ad_hoc_agenda_item_template
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2018.5.0 (unreleased)
 ---------------------
 
+- Do not present the paste UI to users without 'Copy or Move' on the target. [Rotonen]
 - Add trashing of protocol excerpts. [njohner]
 - Display title for NullObject in meeting template selection. [njohner]
 - Fix styling of favorite button. [njohner]
--- a/opengever/base/browser/pasting_allowed.py
+++ b/opengever/base/browser/pasting_allowed.py
@@ -3,6 +3,7 @@ from Acquisition import aq_parent
 from opengever.base.clipboard import Clipboard
 from opengever.dossier.behaviors.dossier import IDossierMarker
 from opengever.private.interfaces import IPrivateContainer
+from plone import api
 from Products.Five import BrowserView
 from ZODB.POSException import ConflictError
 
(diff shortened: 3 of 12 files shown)


README.rst
```diff
@@ -605,6 +605,9 @@ Objects
             - self.meeting_subtask
         - self.protected_dossier
           - self.protected_document
+        - self.protected_dossier_with_task
```
phgross: Why is an additional dossier with a task necessary? Whats the point to not just add the task to the protected_dossier

Rotonen: So we have an otherwise protected dossier to which `self.regular_user` has access only because of the task assignation.

Rotonen:
Juggling the rebases on this one are starting to eat unnecessary amounts of time in aggregate.

Additionally now the fixture test wants to have `self.mail` instead of `self.mail_eml` in there for me, and I cannot figure that one out.

```
diff --git a/README.rst b/README.rst
index 51d663caf..6c91611d4 100644
--- a/README.rst
+++ b/README.rst
@@ -577,7 +577,7 @@ Objects
           - self.draft_proposal
           - self.draft_word_proposal
           - self.empty_document
-          - self.mail
+          - self.mail_eml
           - self.mail_msg
           - self.proposal
             - self.proposaldocument
```

jone:
@Rotonen what is the question?

Rotonen:
@jone why does `bin/test -t TestDumpFixtureStrutureToReadme` fail on this branch?

jone:
@Rotonen this is strange indeed.
Actually it is correct that there must be a `mail_eml`, it is actually registered here: https://github.com/4teamwork/opengever.core/blob/f1dc9ba3340b2c1c87d0522caf26d652fa28ef42/opengever/testing/fixtures.py#L1145

But I've no idea why this is not yet on master and why the master can be green without that 🤔

Rotonen:
@jone the strangest thing there is it should want both in the readme, but it seems to want either or, based on a yet unknown pivot.

jone:
I think the problem is here: https://github.com/4teamwork/opengever.core/blob/91d11052545de20383618405f1a3d76853399839/opengever/testing/tests/test_dump_fixture_structure.py#L51
A path-based dictionary of all fixture objects is built.
The problem is that `self.mail` and `self.mail_eml` is the same object, so there are two entries with the same path. Since dictionaries are involved, it is a bit random which one is the winner.

So we should either register an object only once (maybe add an assertion) or fix the dump test so that it supports multiple entries with the same path.

Rotonen:
Ready for (re?)review.

Rotonen:
Rebased and ready for (re?)review.

docs/HISTORY.txt
```diff
@@ -32,6 +32,9 @@ Changelog
 - Prevent replacing files on proposal templates with non-.docx files via quickupload. [Rotonen]
 - Add an upgrade step to fix broken-by-broken-protocol-excerpt journal entries on dossiers. [Rotonen]
 - Do not list documents within dossiertemplates when creating a document from template. [njohner]
+- Fix task path queries for oracle backends. [phgross]
```

docs/HISTORY.txt
```diff
@@ -32,6 +32,9 @@ Changelog
...
+- Fix activity queries for oracle backends. [phgross]
```

PR #4833, opened 2018-09-04
Fix bug in document template list.

### LINKED ISSUE #4657: Creating a document from a template allows to select documents from a dossiertemplate
Assuming a dossier template with a document:
screen shot 2018-07-31 at 11 25 44

When creating a dossier from a template, the document in the dossier template is available as document template:
screen shot 2018-07-31 at 11 26 40

See as well https://extranet.4teamwork.ch/support/gemeinde-koeniz/tracker/76

njohner:
When creating a document from a template, `document` templates  inside a `dossiertemplate` should not be in the list of templates to choose from.

Documents can be inside the main `templatefolder`, but also in sub-`templatefolder`, `dossiertemplate`, subfolders in `dossiertemplate`, etc.
In the end we want as templates all `document` that are nor contained in a `templatefolder`. This would need to be recursively checked (as they can have dossiers as direct `parent_folder`). It is easier to get all documents that are directly contained in a `templatefolder`, which should be equivalent. This requires several catalog searches, and a custom sorting.

resolves #4657 
Also resolves https://extranet.4teamwork.ch/support/gemeinde-koeniz/tracker/76

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -14,6 +14,7 @@ Changelog
 - Fix bug with document without file in a dossiertemplate. [njohner]
 - Prevent replacing files on proposal templates with non-.docx files via quickupload. [Rotonen]
 - Add an upgrade step to fix broken-by-broken-protocol-excerpt journal entries on dossiers. [Rotonen]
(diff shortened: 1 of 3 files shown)


njohner:
Yes sorry. I've added a test.

PR #4834, opened 2018-09-04
Add an upgrade step to fix broken-by-broken-protocol-excerpt journal entries on dossiers

### LINKED ISSUE #4587: UnicodeDecodeError: Beim Darstellen des Journal Tabs in einem Dossier
https://sentry.4teamwork.ch/sentry/onegov-gever/issues/3975/

```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 92: ordinal not in range(128)
(9 additional frame(s) were not displayed)
...
  File "apps/eggs/zope.i18n-3.7.4-py2.7.egg/zope/i18n/__init__.py", line 126, in translate
    return util.translate(msgid, mapping, context, target_language, default)
  File "apps/eggs/zope.i18n-3.7.4-py2.7.egg/zope/i18n/translationdomain.py", line 78, in translate
    msgid, mapping, target_language, default, context)
  File "apps/eggs/zope.i18n-3.7.4-py2.7.egg/zope/i18n/translationdomain.py", line 140, in _recursive_translate
    text = interpolate(text, mapping)
  File "apps/eggs/zope.i18n-3.7.4-py2.7.egg/zope/i18n/__init__.py", line 164, in interpolate
    return _interp_regex.sub(replace, text)
  File "apps/eggs/zope.i18n-3.7.4-py2.7.egg/zope/i18n/__init__.py", line 159, in replace
    return unicode(mapping.get(param1 or param2, whole))

UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 92: ordinal not in range(128)
```

Rotonen:
Seems our saga with the protocol excerpt titles was not yet done.

@phgross @deiferni a candidate for a backport?

Closes #4587

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -11,6 +11,7 @@ Changelog
 - Add actions pointing to the meeting and protocol debug views. [njohner]
 - Fix bug with document without file in a dossiertemplate. [njohner]
 - Prevent replacing files on proposal templates with non-.docx files via quickupload. [Rotonen]
+- Add an upgrade step to fix broken-by-broken-protocol-excerpt journal entries on dossiers. [Rotonen]
 - Correct width of subdossier table in dossier details pdf. [njohner]
 - Do not set document date during check-in or cancel. [njohner]
 - Disallow grouping tasks by the checkbox column in list views. [Rotonen]
--- a/opengever/core/upgrades/20180907153824_fix_dossier_journal_entries_of_broken_titled_meeting_excerpts/upgrade.py
+++ b/opengever/core/upgrades/20180907153824_fix_dossier_journal_entries_of_broken_titled_meeting_excerpts/upgrade.py
@@ -1,0 +1,104 @@
+from ftw.journal.config import JOURNAL_ENTRIES_ANNOTATIONS_KEY
+from ftw.journal.interfaces import IAnnotationsJournalizable
+from opengever.core.upgrade import SQLUpgradeStep
+from opengever.ogds.base.utils import get_current_admin_unit
+from plone import api
+from sqlalchemy.sql.expression import column
+from sqlalchemy.sql.expression import table
+from zope.annotation.interfaces import IAnnotations
+from zope.app.intid.interfaces import IIntIds
+from zope.component import getUtility
+from zope.i18nmessageid import Message
+
+
+generateddocuments_table = table(
+    'generateddocuments',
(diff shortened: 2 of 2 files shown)


opengever/core/upgrades/20180904184647_fix_dossier_journal_entries_of_broken_titled_meeting_excerpts/upgrade.py
```diff
@@ -0,0 +1,92 @@
...
+        excerpts = self.execute(
+            generateddocuments_table.select()
+            .where(generateddocuments_table.c.admin_unit_id == admin_unit_id)
+            .where(generateddocuments_table.c.generated_document_type == 'generated_excerpt')
+            ).fetchall()
+        if not excerpts:
+            return
+        journal_fixer = JournalMessageFixer(keys_to_fix=['title'])
+        intids = getUtility(IIntIds)
+        dossiers = set(intids.queryObject(excerpt.int_id).get_parent_dossier() for excerpt in excerpts)
```
deiferni: there may be rare cases where the `int_id` points to `None`. Even though this is invalid state i would be careful here and handle `queryObject` returning `None`.

opengever/core/upgrades/20180904184647_fix_dossier_journal_entries_of_broken_titled_meeting_excerpts/upgrade.py
```diff
@@ -0,0 +1,92 @@
...
+    )
+
+
+class FixDossierJournalEntriesOfBrokenTitledMeetingExcerpts(SQLUpgradeStep):
+    """Fix dossier journal entries of broken-titled meeting excerpts."""
+
+    def migrate(self):
+        has_meeting_feature = api.portal.get_registry_record('opengever.meeting.interfaces.IMeetingSettings.is_feature_enabled')  # noqa
+        if not has_meeting_feature:
+            return
```
deiferni: IMO adding an extra newline in appropriate places would improve readability, for example in this method after each `return`.

Rotonen:
Ready for rereview.

Rotonen:
@phgross so now that we still do have an another bugfix release, do we backport this after all?

phgross:
@Rotonen I think this bug is not critical and there is no bugfix-release planned for our SPV customers, so I would say no backport yet.

PR #4835, opened 2018-09-05
Fix bug with document without file in a dossiertemplate.

### LINKED ISSUE #4828: Document with no file in dossier templates raises an error
When creating a dossier from a template, if the dossier template contains a document without a file, it throws `AttributeError: 'NoneType' object has no attribute 'filename'` hinting at a `document.file.filename` instead of `document.get_filename()`

njohner:
Creating a dossier from a dossiertemplate would fail when there was a document without a file in the dossiertemplate.

resolves #4828


njohner:
All done. I added a test for the `CreateDocumentFromTemplateCommand` and also one for `CreateDossierFromTemplateCommand`

PR #4857, opened 2018-09-07
Preserve proposal document title for proposal documents of submitted proposals

### LINKED ISSUE #4845: Proposal documents don't keep their title when submitted
There is an issue when submitting a proposal where the proposal document has a different title than the proposal. I would expect that the submitted proposal's document uses the same title as the proposal's document. The document title is reset to the submitted proposal title, however.

Steps to reproduce:
- Create a proposal
- Set a new title for the proposal's document which is different from the proposals title
- Submit the proposal
- Go to the submitted proposal and view its document

Expected outcome:
- The document of the submitted proposal should have the same title as the document of the proposal

Actual outcome:
- The document's title is set to the title of the proposal (see https://github.com/4teamwork/opengever.core/blob/1cff81b94320be2512a27841f1ea4c7d0f50f1c3/opengever/meeting/proposal.py#L312)

Rotonen:
Closes #4845

### DIFF
--- a/opengever/meeting/browser/createsubmittedproposal.py
+++ b/opengever/meeting/browser/createsubmittedproposal.py
@@ -41,6 +41,7 @@ class CreateSubmittedProposal(BrowserView):
             submitted_proposal.sync_model(proposal_model=proposal)
 
             submitted_proposal.create_proposal_document(
+                title=data['title'],
                 filename=data['file']['filename'],
                 content_type=data['file']['contentType'].encode('utf-8'),
                 data=base64.decodestring(data['file']['data']))
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -8,6 +8,7 @@ Changelog
 - Add TaskReminderActivity object. [elioschmutz]
 - Add ReminderSetting SQL Model. [elioschmutz]
 - Add task-reminder activity settings. [elioschmutz]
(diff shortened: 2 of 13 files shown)


opengever/meeting/browser/createsubmittedproposal.py
```diff
@@ -41,6 +41,7 @@ def __call__(self):
             submitted_proposal.sync_model(proposal_model=proposal)
 
             submitted_proposal.create_proposal_document(
+                title=proposal_oguid.resolve_object().get_proposal_document().title_or_id(),
```
deiferni: We're attempting to keep the workload necessary to make the meeting stuff work across admin units low. Thus you should not get the proposal document here as that may be on another plone site.
IIRC the document and file data is provided in the request. Could you pick the title out of there (title should be identical to the filename)?

deiferni: for the record as just discussed: the title could be added in https://github.com/4teamwork/opengever.core/blob/df27d21d1517041e2fdde8d6fe7809e6680daa93/opengever/meeting/command.py#L425-L428 and then retrieved from the request when creating the document.

Rotonen:
Ready for rereview.

Rotonen:
@deiferni nixed the reformats on anything I did not move or actually touch.

Issue #4854, opened 2018-09-07
Batching on SOLR search: second batch not working


phgross:
Steps do reproduce: 
 - erweiterte Suche
 - Dossiers, Status: `in Bearbeitung`, Suche nach responsible
 - Suchen 
 - Wechsel auf zweiter Batch

Bsp: https://dev.onegovgever.ch/rk/@@search?object_provides=opengever.dossier.behaviors.dossier.IDossierMarker&responsible=zopemaster&review_state%3Alist=dossier-state-active

lukasgraf:
@phgross could this be 4teamwork/opengever.core#4363 again? (Batching broken because heap size that is too small?)

deiferni:
Fixed by https://github.com/4teamwork/opengever.core/pull/4863

PR #4863, opened 2018-09-10
Fix batching links on the solr search.

phgross:
The request.form values has been mistakenly solr-escaped when generating solr filters, this led to broken batching links (resolves 4845).

Backport to `2018.4-stable`


phgross:
@lukasgraf updated

deiferni:
Fixes https://github.com/4teamwork/opengever.core/issues/4854

PR #4871, opened 2018-09-12
No longer use catalog.reindexObject.

### LINKED ISSUE #4861: Stop using catalog.reindexObject
`catalog.reindexObject` explicitly reindexes in Portal Catalog only, but it doesn't in Solr.
This leads to an inconsistent Solr index.

We should **always** use `obj.reindexObject()` instead of `catalog.reindexObject(obj)`

Wrong usage:
https://github.com/4teamwork/opengever.core/blob/master/opengever/document/checkout/manager.py#L90
https://github.com/4teamwork/opengever.core/blob/master/opengever/document/checkout/manager.py#L163
https://github.com/4teamwork/opengever.core/blob/master/opengever/document/checkout/manager.py#L217
https://github.com/4teamwork/opengever.core/blob/master/opengever/base/handlers.py#L44
https://github.com/4teamwork/opengever.core/blob/master/opengever/usermigration/creator.py#L84

Problem reported by customer:
https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/293

phgross:
Because using `catalog_tool.reindexObject` updates Portal Catalog only, but it doesn't in Solr.
This leads to an inconsistent Solr index and confusing search results for the user. 

Discussed to day with buchi, if ftw.solr should also handle `catalog.reindexObject` calls: 
He is of opinion  that  catalog's `reindexObject`, is an internal API which only updates the portal_catalog, it also skip other catalogs (UID ...) and therefore should not do anything with solr. This makes sense.

The PR contains an upgradestep which reindexes checked_out and allowedRolesAndUsers on solr. Because those could be affected by the reindexing problem. The upgradestep is marked as deferrable, because its not urgent.

I also overwrite the `plone.app.transmogrifiert` ReindexObject section, to include a manual SOLR reindex there.

Notes:
 - Upgradesteps before the SOLR integration (09.02.2018) are not fixed/adjusted.
 - ftw.upgrade does not use `catalog.reindexObject`

Resolves #4861 

Backport to `2018.4-stable`

### DIFF
--- a/opengever/base/handlers.py
+++ b/opengever/base/handlers.py
@@ -1,3 +1,5 @@
+from collective.indexing.interfaces import IIndexQueueProcessor
+from collective.indexing.queue import getQueue
 from ftw.upgrade.helpers import update_security_for
 from opengever.base.model import create_session
 from opengever.base.model.favorite import Favorite
@@ -9,6 +11,7 @@ from plone.app.workflow.interfaces import ILocalrolesModifiedEvent
 from Products.CMFCore.CMFCatalogAware import CatalogAware
 from Products.CMFPlone.interfaces import IPloneSiteRoot
 from sqlalchemy import and_
+from zope.component import getUtility
 from zope.container.interfaces import IContainerModifiedEvent
 from zope.event import notify
 from zope.lifecycleevent import IObjectRemovedEvent
@@ -43,6 +46,17 @@ def object_moved_or_added(context, event):
         catalog = api.portal.get_tool('portal_catalog')
         catalog.reindexObject(context, idxs=CatalogAware._cmf_security_indexes,
                               update_metadata=0)
+
+        # Using catalogs reindexObject does not trigger corresponding solr
+        # reindexing, so we do it manually.
+
+        # Register collective.indexing hook, to make sure solr changes
+        # are realy send to solr. See
+        # collective.indexing.queue.IndexQueue.hook.
+        getQueue().hook()
+
+        processor = getUtility(IIndexQueueProcessor, name='ftw.solr')
+        processor.index(context, CatalogAware._cmf_security_indexes)
 
 
 def remove_favorites(context, event):
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 ---------------------
 
 - Add missing mimetype for wmf files. [phgross]
+- Fix solr checked-in indexing problems. [phgross]
 - Fix tree portlet header styling for IE11. [Kevin Bieri]
 - Register bmp as editable by officeconnector. [phgross]
 - Fix batching links on the solr search. [phgross]
--- a/opengever/bundle/console.py
+++ b/opengever/bundle/console.py
@@ -1,4 +1,5 @@
 from collective.transmogrifier.transmogrifier import Transmogrifier
+from ftw.solr.interfaces import ISolrConnectionManager
 from opengever.base.interfaces import INoSeparateConnectionForSequenceNumbers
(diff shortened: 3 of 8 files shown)


lukasgraf:
@phgross `plone.app.transmogrifier` also uses `catalog.reindexObject` in its [reindexobject section](https://github.com/collective/plone.app.transmogrifier/blob/2d35bff0e72a1f177b070f58cd86287f7ccac951/src/plone/app/transmogrifier/reindexobject.py#L54-L58). This is used by [`ftw.inflator`'s content_creation.cfg pipeline](https://github.com/4teamwork/ftw.inflator/blob/34c1f1b5fa406b1538d705c59989ff7115580f01/ftw/inflator/creation/content_creation.cfg#L101-L102), and therefore by our [`gever_content.cfg` section](https://github.com/4teamwork/opengever.core/blob/ce8d75858ecfb796380b017753b024ccd72c7e41/opengever/setup/cfgs/gever_content.cfg#L3) used for example content and [bundle imports](https://github.com/4teamwork/opengever.core/blob/master/opengever/bundle/cfgs/oggbundle.cfg#L3).

Overriding the [registration for the `plone.app.transmogrifier.reindexobject`](https://github.com/collective/plone.app.transmogrifier/blob/2d35bff0e72a1f177b070f58cd86287f7ccac951/src/plone/app/transmogrifier/configure.zcml#L59-L62) in an `overrides.zcml` and substituting our own should do the trick.

opengever/base/handlers.py
```diff
@@ -44,6 +46,11 @@ def object_moved_or_added(context, event):
         catalog.reindexObject(context, idxs=CatalogAware._cmf_security_indexes,
                               update_metadata=0)
 
+        # Using catalogs reindexObject does not trigger corresponding solr
+        # reindexing, so we do it manually.
+        processor = getUtility(IIndexQueueProcessor, name='ftw.solr')
+        processor.index(context, CatalogAware._cmf_security_indexes)
```
lukasgraf: This alone isn't quite enough (at least not in all cases).

`processor.index()` only queues the operation (it will "add" it to the ftw.solr connection). The actual request to Solr is then made at the end of the transaction by [`connection.commit()`](https://github.com/4teamwork/ftw.solr/blob/f5aaa582cd9c4ed07477a138571455d8467813ec/ftw/solr/connection.py#L92). This however is invoked (via a few indirections) via `collective.indexing`'s [`Queue.commit()`](https://github.com/plone/collective.indexing/blob/33bd4bdbb7a78031a6654b326cb04283124b819d/src/collective/indexing/queue.py#L175-L178). This in turn is invoked by `collective.indexing`'s data manager [in `tpc_finish`](https://github.com/plone/collective.indexing/blob/33bd4bdbb7a78031a6654b326cb04283124b819d/src/collective/indexing/transactions.py#L58-L60). That DM get's registrered into the TransactionsManager (for every request) by [`Queue.hook`](https://github.com/plone/collective.indexing/blob/33bd4bdbb7a78031a6654b326cb04283124b819d/src/collective/indexing/queue.py#L86-L93). However. `Queue.hook` is only triggered via the `obj.reindexObject()` methods monkey patched by `c.indexing`, not via `portal_catalog.reindexObject`.

So this might seem to work because in the same request something else might call `obj.reindexObject()`, but to be sure we should explicitly call the hook function here.

```python
from collective.indexing.queue import getQueue

getQueue().hook()
```
should do the trick.

phgross:
@lukasgraf updated ...

phgross:
@lukasgraf updated with a "safeguard" when solr-reindexing is enabled but solr isn't running.

SG wiki updated ...

PR #4318, opened 2018-05-17
Fix too long physical_path value issue.

deiferni:
This PR fixes issue where the physical_path value could be too long. I'm fixing it for all the `physical_paths` we store in the relational database.

After a brief discussion with @Rotonen and @lukasgraf we decided against adding an index for now.

_The `existing_...` params would not be necessary for postgres but i find they are helping in documenting what we're doing._

Fixes #4235.

Needs backport to: `2018.1-stable` and `2018.2-stable`

PR #4884, opened 2018-09-19
Fix task and activity queries for oracle backends

phgross:
**Activities**
Do not use `is_(True)` for sqlalchemy filter statements, because this leads to an `ORA-00908: missing NULL keyword` exception on oracle backends. So i use filter(== true()) instead.
 - https://stackoverflow.com/questions/15143827/missing-null-keyword-error-when-trying-to-avoid-duplicates
 - https://stackoverflow.com/questions/18998010/flake8-complains-on-boolean-comparison-in-filter-clause for more information.

**Task**
With #4318 the physical_path column has been switched from a string to a text column. On oracle backends this leads to errors when filtering on physicalpath (`DatabaseError: (cx_Oracle.DatabaseError) ORA-00932: inconsistent datatypes: expected - got CLOB`).

Therefore I implemented a fallback for oracle backends, which first cast to string before filtering.
 - https://stackoverflow.com/questions/12980038/ora-00932-inconsistent-datatypes-expected-got-clob

Note: The path helper could be moved to the basequery, as soon ogds.models is merged into opengever.core. I added  a note in the issue there.

Backport needed for `2018.4`

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -14,6 +14,7 @@ Changelog
 - Prevent replacing files on proposal templates with non-.docx files via quickupload. [Rotonen]
 - Add an upgrade step to fix broken-by-broken-protocol-excerpt journal entries on dossiers. [Rotonen]
 - Do not list documents within dossiertemplates when creating a document from template. [njohner]
+- Fix activity queries for oracle backends. [phgross]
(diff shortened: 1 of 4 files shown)


lukasgraf:
Also Related: https://sentry.4teamwork.ch/sentry/onegov-gever/issues/8856/

PR #4889, opened 2018-09-19
Bump Ruby, nokogiri and sablon

Rotonen:
Mostly for easier local buildability in preparation of macOS 10.14.

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -8,8 +8,8 @@ extends =
 
 [ruby-versions]
 nokogiri = 1.8.4
-sablon = 0.0.21
 ruby = 2.4.4
+sablon = 0.3.0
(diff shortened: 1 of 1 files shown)


phgross:
@Rotonen Welchen Issue/Problem wird dadurch behoben? Was macht ein Update notwendig?

Rotonen:
As I've found workarounds for libxml build issues with the macOS 10.14 SDK for me locally in the meanwhile I'm dropping this as I do not need to drag custom versions with me locally anymore.

deiferni:
Don't get me wrong, i strongly agree we must update this eventually as some interesting features seem to have been added. Should we create an issue for 2019.X?

PR #4912, opened 2018-09-26
Only allow adding meeting templates, when meeting feature is enabled.

### LINKED ISSUE #4905: Aktion `Sitzungsvorlage` überprüft Meeting Feature flag nicht
Im Vorlagenbereich ist es möglich eine Sitzungsvorlage hinzuzufügen, obwohl das Meeting-Feature deaktiviert ist.

bildschirmfoto 2018-09-25 um 13 09 25

Korrektur für `2018.4-stable` nötig.

phgross:
This works for the action, as well for the API

Closes #4905 

Backport to: `2018.4-stable`

### DIFF
--- a/opengever/dossier/templatefolder/templatefolder.py
+++ b/opengever/dossier/templatefolder/templatefolder.py
@@ -43,10 +43,9 @@ class TemplateFolder(Container, TranslatedTitleMixin):
 
         def filter_type(fti):
             factory_type = fti.id
-            if factory_type in [u'opengever.meeting.sablontemplate']:
-                return is_meeting_feature_enabled()
-
-            if factory_type in [u'opengever.meeting.proposaltemplate']:
+            if factory_type in [u'opengever.meeting.sablontemplate',
+                                u'opengever.meeting.proposaltemplate',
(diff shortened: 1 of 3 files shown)


opengever/dossier/templatefolder/templatefolder.py
```diff
@@ -43,10 +43,9 @@ def allowedContentTypes(self, *args, **kwargs):
 
         def filter_type(fti):
             factory_type = fti.id
-            if factory_type in [u'opengever.meeting.sablontemplate']:
-                return is_meeting_feature_enabled()
-
-            if factory_type in [u'opengever.meeting.proposaltemplate']:
+            if factory_type in [u'opengever.meeting.sablontemplate',
```
deiferni: you could use an immutable thing here, e.g. a `tuple`. that would make sense if the iterable was intialized once, e.g. on class level as this would guard against unintended modification.

deiferni: never mind then, i wanted to say that this is an optional suggestion in the review, but @Rotonen beat me to it 😄.

Issue #4911, opened 2018-09-26
Double registering an object in the fixtures causes flaky testing tests


Rotonen:
Currently we register the EML mail twice in the fixtures to avoid refactoring every test using the old name. This causes the testing tests to pick up on either `self.mail` or `self.mail_eml` in the fixtures in a flaky way.

Split off from https://github.com/4teamwork/opengever.core/pull/4829

PR #4913, opened 2018-09-26
Prevent double registrations in the fixtures

Rotonen:
Ended up here from https://github.com/4teamwork/opengever.core/pull/4829

* Make everything use explicit `self.mail_eml`
* Add a test for preventing object double registrations

Closes https://github.com/4teamwork/opengever.core/issues/4911

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -577,7 +577,7 @@ Objects
           - self.draft_proposal
           - self.draft_word_proposal
           - self.empty_document
-          - self.mail
+          - self.mail_eml
           - self.mail_msg
           - self.proposal
             - self.proposaldocument
--- a/opengever/api/tests/test_mail.py
+++ b/opengever/api/tests/test_mail.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 from ftw.testbrowser import browsing
 from opengever.mail.mail import IOGMail
 from opengever.testing import IntegrationTestCase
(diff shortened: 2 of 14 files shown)


opengever/testing/tests/test_fixtures.py
```diff
@@ -91,3 +91,22 @@ def test_dossiers(self):
...
+            path
+            for _type, path in self.layer['fixture_lookup_table'].values()
+            if _type == 'object'
+        ]
+        duplicated_paths = [
+            path
+            for n, path in enumerate(registered_object_paths)
+            if path in registered_object_paths[:n]
+        ]
+        self.assertEqual([], duplicated_paths, 'Some objects are registered twice!')
```
deiferni: please also output the duplicated paths here.

opengever/testing/tests/test_fixtures.py
```diff
@@ -91,3 +91,22 @@ def test_dossiers(self):
...
         self.assertEquals('Abgeschlossene Vertr\xc3\xa4ge', self.expired_dossier.Title())
+
+    def test_no_double_registrations(self):
+        """Prevent objects from being registered more than once.
+
+        This is problematic for how we dump the object paths into the readme.
+        The implementation is a simple iteration over all the paths of
+        registered objects with a sliding slice list comprehension.
+        """
+        registered_object_paths = [
```
deiferni: couldn't we do this at `register_path` time? keep a set of all the paths in the fixture and immediately puke when one is registered twice. that would provide feedback faster, i.e. as soon as the fixture is modified and *any* test is run.

opengever/testing/tests/test_fixtures.py
Rotonen: It does, in the assertion delta, as those will be the contents of `duplicated_paths`.

opengever/testing/tests/test_fixtures.py
Rotonen: That nearly doubled the fixture setup time for me locally. I'm unsure as to where we'd like to draw the line on that.

deiferni: we might wanna look at how it was implemented then. i can't imagine that such a simple check should take one minute or somesuch of runtime 🤔.

Rotonen: Thinking about it, that should be a trivial `in list` comparison, I wonder what I did wrong during my local attempt. I'll reinvestigate.

deiferni: i'd use a `set` as that will most likely give you a faster `in` given the amount of objects we have in the fixture 😏

Rotonen:
@deiferni ready for rereview.

PR #4915, opened 2018-09-27
Ensure users with only 'Reader' on dossier see closed tasks

### LINKED ISSUE #4887: Reader is not able to see closed tasks inside a dossier.
Users which only have the Reader role on a dossier are not able to see closed tasks. In my Opinion thats wrong and is now a real problem since we remove local roles on task when closing them.

Example: https://dev.onegovgever.ch/fd/ordnungssystem/fuehrung/gemeinderecht/uebergeordnete-erlasse/dossier-448#overview

See [extranet issue](https://extranet.4teamwork.ch/support/gemeinde-koeniz/tracker/102)

Rotonen:
Closes #4887


Rotonen:
Got it now, but I guess we actually would like to acquire ´View` wider than just for the closed state?

Rotonen:
@phgross so, maybe I finally got what's going on - amended what the local roles setter does for tasks which are in the final state.

opengever/task/localroles.py
```diff
@@ -23,14 +23,17 @@ def __init__(self, task):
...
     def set_roles(self, event):
-        if self.task.is_in_final_state:
-            return
-
         self.event = event
-        self.set_roles_on_task()
-        self.globalindex_reindex_task()
-        self.set_roles_on_related_items()
-        self.set_roles_on_distinct_parent()
+
```
phgross: No, thats wrong, you add access to responsible users, which is not the case of the issue/pr and definitely not the wish (with OGIP 45 we recently introduced the exact opposite of this change - the revoke of local roles on task closing).

The only change which is needed is to ensure that users with only the 'Reader' role on a dossier are able to see closed tasks:
 => Add `Reader` role to the `view` permission for closed task state, in the task workflow.

Rotonen:
@phgross do I:

1) add to the forbiddenfruit mess of our unit tests to make `api.user.has_permission()` work in the unit tests
2) port the task transition controller unit tests to integration tests

?

Rotonen:
@phgross ready for rereview.

opengever/testing/builders/dx.py
```diff
@@ -228,6 +228,7 @@ def successor_from(self, predecessor):
 
     def change_workflow_state(self, obj):
         super(TaskBuilder, self).change_workflow_state(obj)
+        obj.get_sql_object().review_state = api.content.get_state(obj)
```
phgross: Thats not necessary, the sync_task(obj, None) a line below, those also update the review_state.

opengever/task/tests/test_localroles.py
```diff
@@ -198,6 +198,29 @@ def test_closing_a_task_revokes_responsible_roles_on_task(self, browser):
...
+        self.login(self.reader_user, browser)
+        browser.open(self.task, view='tabbedview_view-overview')
+        expected_actions = ['task-transition-delegate', 'task-transition-reassign']
+        self.assertEqual(expected_actions, browser.css('#action-menu a').text)
+
+        with self.login(self.dossier_responsible, browser=browser):
+            self.set_workflow_state('task-state-tested-and-closed', self.subtask)
+            self.set_workflow_state('task-state-resolved', self.task)
+
+        browser.open(self.task, view='tabbedview_view-overview')
```
phgross: Hmm you test with `dossier_responsible`, but I would say he got more than the `Reader` role.

opengever/testing/builders/dx.py
Rotonen: IIRC that only did it for the Plone object, thus I ended up doing this, but will reverify.

opengever/task/tests/test_localroles.py
Rotonen: ?

opengever/testing/builders/dx.py
Rotonen: Yep, seems I missed the `globalindex/model` logic and forgot to requery in early testing. Dropping this commit.

Rotonen:
@phgross ready for rereview.

opengever/task/tests/test_localroles.py
phgross: Too early review 🙈 ...

PR #4916, opened 2018-09-27
Allow users to retry creating a file from a Oneoffixx template

Rotonen:
Split off from https://github.com/4teamwork/opengever.core/pull/4282

In case anything goes wrong, this is enabling us to let users retry, and for development and trying things out, this is undispensable.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -12,6 +12,7 @@ Changelog
 - Remove leftover checkout and edit, and cancel actions for sablon and proposal templates. [njohner]
 - Do not include paragraphs in comittee table of contents. [njohner]
 - Add TaskReminder to handle reminders in annotations and sql. [elioschmutz]
+- Allow users to retry creating a file from a Oneoffixx template. [Rotonen]
 - Add TaskReminderActivity object. [elioschmutz]
 - Make paragraph templates deletable. [Rotonen]
 - Add ReminderSetting SQL Model. [elioschmutz]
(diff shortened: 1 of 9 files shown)


docs/HISTORY.txt
```diff
@@ -12,6 +12,8 @@ Changelog
 - Remove leftover checkout and edit, and cancel actions for sablon and proposal templates. [njohner]
 - Do not include paragraphs in comittee table of contents. [njohner]
 - Add TaskReminder to handle reminders in annotations and sql. [elioschmutz]
+- Allow users to retry creating a file from a Oneoffixx template. [Rotonen]
+- Only allow adding meeting templates, when meeting feature is enabled. [phgross]
```

opengever/document/document.py
```diff
@@ -299,6 +299,9 @@ def is_checkout_and_edit_available(self):
     def is_shadow_document(self):
         return api.content.get_state(self) == self.shadow_state
 
+    def is_oneoffixx_creatable(self):
+        return is_oneoffixx_feature_enabled and self.is_shadow_document()
```
phgross: is_oneoffixx_feature_enabled needs to be called ...

docs/HISTORY.txt
Rotonen: Autorebase. Sorry.

opengever/document/document.py
Rotonen: Good catch.

Rotonen:
@phgross ready for rereview.

PR #4921, opened 2018-09-28
 Fix bumblebee overlay link for not yet created initial version.

### LINKED ISSUE #4917: Bumblebee Version link raises not found for not existing initial version
If the initial version of a document is not yet created, the Preview link in the versions_tab is not working.

The ajax request raises an 404 because the version can not be retrieved.

phgross:
Since 2017.6 (PR #3467) the initial version of a document is only created when really used. Which leads to a not working bumblebee overlay link, when the initial version does not exist yet. Closes #4917

I add a special handling to the `InitialVersionDataProxy`, which uses the working copy to fetch the document.

Backport to `2018.4.-stable`

### DIFF
--- a/opengever/document/browser/versions_tab.py
+++ b/opengever/document/browser/versions_tab.py
@@ -154,6 +154,11 @@ class VersionDataProxy(object):
             label = translate_text(_(u'label_revert', default=u'Revert'))
             return u'<span class="discreet">{}</span>'.format(label)
 
+    @property
+    def _bumblebee_url(self):
+        return "{}/@@bumblebee-overlay-listing?version_id={}".format(
+            self.url, self.version)
+
 
 class LazyHistoryMetadataProxy(object):
     """A proxy for CMFEditions `ShadowHistory` objects as returned by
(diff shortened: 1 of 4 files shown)


opengever/document/browser/versions_tab.py
```diff
@@ -154,6 +154,14 @@ def revert_link(self):
             label = translate_text(_(u'label_revert', default=u'Revert'))
             return u'<span class="discreet">{}</span>'.format(label)
 
+    @property
+    def _bumblebee_url(self):
+        """Returns a formatted link to revert to this particular version if
```
deiferni: that docstring does not seem correct to me.

opengever/document/browser/versions_tab.py
```diff
@@ -299,6 +307,14 @@ def download_link(self):
             viewname='download', include_token=True)
         return link
 
+    @property
+    def _bumblebee_url(self):
+        """Returns a formatted link to revert to this particular version if
```
deiferni: that docstring does not seem correct to me.

opengever/document/browser/versions_tab.py
phgross: @deiferni mmm sorry 🙈 , I updated the PR

PR #4922, opened 2018-09-28
Bump Plone to 4.3.18

### LINKED ISSUE #4859: Update Plone to 4.3.18
Update Plone to `4.3.18`.

Also check
- whether there's remaining work from #4672 that needs or can be done.
- whether there's new CSRF false positives: See https://community.plone.org/t/plone-4-3-and-automatic-csrf-protection/7290/11

Rotonen:
* Bumped Plone and hotfix pinnings
* Unpinned everything Plone now pins to what we need

Closes #4859


deiferni:
what is pending for `ftw.tabbedview`, is there an issue or PR?

lukasgraf:
This can be shipped without the base URL work on the `ftw.tabbedview` side being complete yet IMHO.

Turns out, the `data-base-url` attribute gets manipulated by Plone JS during page load in more complex ways than I thought. So the fix to use `data-base-url` if it exists and otherwise just fall back to the `<base />` tag in the header is a bit hairier to implement than anticipated.

My latest change here makes sure we still don't drop `data-base-url` (introduced in Plone 4.3.16) in our Diazo theme, and includes tests for this as well as the `<base />` tag we're currently still using. Tests run on a separate integration test layer that combines the content fixture with the `plonetheme.teamraum:gever`.

*(Including the profile in all our tests via `OpengeverFixture` currently leads to [too many failing tests](https://ci.4teamwork.ch/builds/196302) to fix right now).*

PR #4923, opened 2018-10-01
Fix some french translations

phgross:
according to the feedback in https://extranet.4teamwork.ch/support/ph-vs/tracker/513.

Backport to 2018.4-stable

### DIFF
--- a/opengever/base/locales/fr/LC_MESSAGES/plone.po
+++ b/opengever/base/locales/fr/LC_MESSAGES/plone.po
@@ -125,6 +125,9 @@ msgstr "Usage"
 msgid "alphabetically"
 msgstr "Alphabétique"
 
+msgid "batch_x_items_matching_your_criteria"
+msgstr "${number} élément(s) trouvé(s)."
+
 msgid "button_download"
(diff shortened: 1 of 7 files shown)


opengever/base/locales/fr/LC_MESSAGES/plone.po
```diff
@@ -125,6 +125,9 @@ msgstr "Usage"
 msgid "alphabetically"
 msgstr "Alphabétique"
 
+msgid "batch_x_items_matching_your_criteria"
+msgstr "Il y a ${number} élément(s) qui correspondent à vos termes de recherche."
```
njohner: The translation is good, but it's much more precise than the German one. If you want something that is closer to the German one here, you could go with:
`"${number} élément(s) trouvé(s)."`
Also for including both singular and plural you would actually need one more bracket:
`"Il y a ${number} élément(s) qui correspond(ent) à vos termes de recherche."`

phgross: Thx for your proposal. I like the shorter one.

PR #4925, opened 2018-10-02
Do not include paragraphs in AlphabeticalToc.

### LINKED ISSUE #4907: Zwischentitel erscheinen in Inhaltsverzeichnis
In den Inhaltsverzeichnissen (alphabetisch und Ordnungsposition-basiert) erscheinen Zwischentitel. Dies sollte nicht der Fall sein.

Lösungsvorschlag:
`AgendaItem.is_paragraph` muss im query https://github.com/4teamwork/opengever.core/blob/master/opengever/meeting/toc/alphabetical.py#L55-L72 berücksichtig werden.

njohner:
We do not want paragraphs (intertitles) included in the table of contents (alphabetical or repository based).
I also updated the tests to make sure the intertitles are excluded from the tocs.

resolves #4907 

Sidenote: `description` of the agendaitems is not in the generated json, but as @deiferni pointed out, there is probably no one out there who wants the descriptions in the Toc. So I left it as is for now.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 ---------------------
 
 - Remove leftover checkout and edit, and cancel actions for sablon and proposal templates. [njohner]
+- Do not include paragraphs in comittee table of contents. [njohner]
 - Add TaskReminderActivity object. [elioschmutz]
 - Make paragraph templates deletable. [Rotonen]
 - Add ReminderSetting SQL Model. [elioschmutz]
--- a/opengever/meeting/toc/alphabetical.py
+++ b/opengever/meeting/toc/alphabetical.py
@@ -9,6 +9,7 @@ from opengever.meeting.utils import JsonDataProcessor
 from operator import itemgetter
 from sqlalchemy.orm import contains_eager
 from sqlalchemy.orm import joinedload
+from sqlalchemy.sql.expression import false
 import pytz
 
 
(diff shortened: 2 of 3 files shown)


opengever/meeting/toc/alphabetical.py
```diff
@@ -67,7 +67,8 @@ def build_query(self):
         # relevant day for the toc
         query = query.filter(Meeting.start >= datetime_from,
                              Meeting.start < datetime_to,
-                             Meeting.committee == self.period.committee
+                             Meeting.committee == self.period.committee,
+                             AgendaItem.is_paragraph == False  # noqa
```
deiferni: IIRC for oracle safety you'd want to import `from sqlalchemy.sql.expression import false` and write:

```python
AgendaItem.is_paragraph == false()
```

Also that will eliminate the need for `# noqa`. See https://github.com/4teamwork/opengever.core/pull/4884 for another PR deadling with that issue.

docs/HISTORY.txt
```diff
@@ -5,6 +5,7 @@ Changelog
 2018.5.0 (unreleased)
 ---------------------
 
+- Do not include paragraphs (intertitles) in comission table of contents. [njohner]
```
deiferni: hm i think intertitle is the wrong word here 🤔

deiferni: comission -> committee, i'd stick to the vocabulary we use already

opengever/meeting/toc/alphabetical.py
njohner: Ok thanks.

docs/HISTORY.txt
njohner: wops, that looks like a mixup with the German translation...

njohner:
Voilà, sorry...

PR #4930, opened 2018-10-03
Task-Reminder Restapi-Endpoints to add and delete reminders

elioschmutz:
Dieser PR fügt die Task-Reminder Endpoints zu hinzufügen und löschen von Task-Reminders hinzu: `@reminder`.

Issuer: #4205 
Based on: #4736

### DIFF
--- a/docs/public/dev-manual/api/reminder.rst
+++ b/docs/public/dev-manual/api/reminder.rst
@@ -1,0 +1,75 @@
+.. _reminders:
+
+Aufgaben-Erinnerungen
+=====================
+
+Der ``@reminder`` Endpoint behandelt Erinnerungen für Aufgabe. Jeder Benutzer kann für jede Aufgabe eine eigene Erinnerung setzten.
+
+
+Erinnerung setzten:
+-------------------
+Mit einem POST Request wird eine neue Erinnerung gesetzt. Als Body wird das Attribut ``option_type`` erwartet.
+Wenn das Attribut einen Wert enthält, welcher nicht zulässig ist, wird eine Liste mit gültigen Werten zurückgegeben.
+
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+       POST /task-1/@reminder HTTP/1.1
+       Accept: application/json
+
+       {
+        "option_type": "one_day_before"
+       }
+
+
+**Beispiel-Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 204 No content
+
+
+Erinnerung aktualisieren:
+-------------------------
+Eine bestehende Erinnerung kann durch einen PATCH Request aktualisiert werden.
+
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+       PATCH /task-1/@reminder HTTP/1.1
+       Accept: application/json
+
+       {
+        "option_type": "same_day"
+       }
+
+
+**Beispiel-Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 204 No content
+
+
+Erinnerung entfernen:
+---------------------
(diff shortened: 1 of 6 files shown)


opengever/api/reminder.py
```diff
@@ -0,0 +1,54 @@
...
+error_msgs = {
+    'missing_option_type': "The request body requires the 'option_type' attribute.",
+    'non_existing_option_type': "The provided 'option_type' does not exists. "
+                                "Use one of the following options: {}".format(
+                                    ', '.join(TASK_REMINDER_OPTIONS.keys())),
+}
+
+
+class TaskReminderPost(Service):
+    """API endpoint to set a task-reminder.
```
phgross: set a task-reminder **for the current user**

opengever/api/reminder.py
```diff
@@ -0,0 +1,54 @@
...
+
+error_msgs = {
+    'missing_option_type': "The request body requires the 'option_type' attribute.",
+    'non_existing_option_type': "The provided 'option_type' does not exists. "
+                                "Use one of the following options: {}".format(
+                                    ', '.join(TASK_REMINDER_OPTIONS.keys())),
+}
+
+
+class TaskReminderPost(Service):
```
phgross: Endpoint to change an reminer is missing (PATCH)

opengever/api/tests/test_reminder.py
```diff
@@ -0,0 +1,109 @@
...
+        self.assertIsNone(
+            self.task_reminder.get_reminder(self.task))
+
+        browser.open(
+            self.task.absolute_url() + '/@reminder',
+            method='DELETE',
+            headers={'Accept': 'application/json'},
+        )
+
+        self.assertEqual(204, browser.status_code)
```
phgross: 204 seems wrong to me, the resource (the reminder in this situation) could not be found --> so 404 not found should be the status_code.

opengever/api/reminder.py
```diff
@@ -0,0 +1,54 @@
...
+            raise BadRequest(error_msgs.get('non_existing_option_type'))
+
+        TaskReminder().set_reminder(self.context, reminder_option)
+
+        self.request.response.setStatus(204)
+        return super(TaskReminderPost, self).reply()
+
+
+class TaskReminderDelete(Service):
+    """API endpoint to remove a task-reminder.
```
phgross: for the current user.

opengever/api/reminder.py
elioschmutz: Currently, if you use the POST method, the reminder will be set or reset if it's already set.

This behavior seems to be good, otherwise, we first have to check if the user has already set a reminder or not and choose between POST or PATCH. And at all, we're not creating a new resource.

So I think, using PUT is more applicable and easier to use than a POST and a PATCH method for this usecase.

PUT will always create a new resource. Even if we don't create a new resource, it's more clear for the developer, that he will always set or override an existing reminder.

Is it OK for you, if I change the POST endpoint to PUT and skip implementing the PATCH mehtod?

elioschmutz: 👍 updated as discussed by phone. I added the PATCH endpoint.

docs/public/dev-manual/api/reminder.rst
```diff
@@ -0,0 +1,75 @@
+.. _reminders:
+
+Aufgaben-Erinnerungen
+=====================
+
+Der ``@reminder`` endpoint behandelt Erinnerungen für Aufgabenobjekte. Jeder Benutzer kann für jedes Aufgabenobjekt eine eigene Erinnerung setzten.
```
phgross: endpoint -> Endpoint
Aufgabenobjekte -> Aufgaben, Aufgabenobjekt -> Aufgabe

PR #4935, opened 2018-10-04
Custom Sablon Exceptions

### LINKED ISSUE #3792: Introduce custom sablon exception
Currently whenever `sablon` fails to process a document successsfully we fail with an `AssertionError`. `Sablon` success is checked with `assert sablon.is_processed_successfully(), sablon.stderr` statements.

This should be improved as follows:
- introduce a proper Exception
- add custom error view for the exception, see https://github.com/4teamwork/opengever.core/blob/962af6802f864facd31d582ceb1fa048f81f6127/opengever/quota/browser/configure.zcml#L15-L20 for an example
- some sablons are generated in JSON requests, also handle the error there
- do **not** ignore the exception yet

njohner:
We introduce a custom exception to handle failures when processing sablon templates.
Some of the sablons are requests issued from JavaScript, which are handled separately.
* We add a custom exception `SablonProcessingFailed` raised when processing fails. This exception adds an error message and reloads the page.
* For requests from JavaScript:
  * For agendaitems, we return the error in the JSON response
  * For TOC download, the response is not JSON, but an `arraybuffer`. We therefore cannot use the machinery of the `Controller` to handle the error. Instead we simply add an error message in Plone and then reload the page in JavaScript.
* Error handling of the TOC download was already present in the case of missing template, but it was broken. I fixed it as described above.

Here are snapshots of the different places the sablon errors are handled:
**Protocol generation**
protocol
**TOC download**
toc
**Agendaitem excerpts**
excerpt
**FillMeetingTemplateView**
fill_meeting_template

resolves #3792

### DIFF
--- a/opengever/base/browser/resources/generateToc.js
+++ b/opengever/base/browser/resources/generateToc.js
@@ -11,6 +11,9 @@
     request.responseType = "arraybuffer";
     request.onload = function() {
       loaded.resolve();
+      if (this.getResponseHeader('X-ogg-reload-page')) {
+        window.location.reload();
+      }
       var contentDisposition = this.getResponseHeader('content-disposition');
       var contentType = this.getResponseHeader('content-type');
       var filename = contentDisposition.match(/filename="(.+)"/)[1];
--- a/opengever/meeting/browser/meetings/agendaitem.py
+++ b/opengever/meeting/browser/meetings/agendaitem.py
@@ -7,6 +7,7 @@ from opengever.meeting import _
 from opengever.meeting.exceptions import CannotExecuteTransition
 from opengever.meeting.exceptions import MissingAdHocTemplate
 from opengever.meeting.exceptions import MissingMeetingDossierPermissions
+from opengever.meeting.exceptions import SablonProcessingFailed
 from opengever.meeting.exceptions import WrongAgendaItemState
 from opengever.meeting.proposal import ISubmittedProposal
 from opengever.meeting.protocol import ExcerptProtocolData
@@ -138,6 +139,11 @@ def return_jsonified_exceptions(func):
                   default=u"No ad-hoc agenda-item template has been "
                           u"configured."),
                 status=501).dump()
+
+        except SablonProcessingFailed:
+            return JSONResponse(getRequest()).error(
+                _('Error while processing Sablon template'),
+                status=500).dump()
 
     return wrapper
 
--- a/opengever/meeting/browser/sablontemplate.py
+++ b/opengever/meeting/browser/sablontemplate.py
@@ -1,5 +1,7 @@
 from opengever.meeting.command import MIME_DOCX
+from opengever.meeting.exceptions import SablonProcessingFailed
 from opengever.meeting.sablon import Sablon
+from Products.CMFPlone.utils import safe_unicode
 from Products.Five.browser import BrowserView
 import json
 
@@ -69,8 +71,10 @@ class FillMeetingTemplate(BrowserView):
 
     def __call__(self):
         sablon = Sablon(self.context)
-        sablon.process(json.dumps(SAMPLE_MEETING_DATA))
-        assert sablon.is_processed_successfully(), sablon.stderr
+        try:
+            sablon.process(json.dumps(SAMPLE_MEETING_DATA))
+        except SablonProcessingFailed as err:
+            return safe_unicode(err.message)
 
         response = self.request.response
         response.setHeader('X-Theme-Disabled', 'True')
--- a/opengever/meeting/command.py
+++ b/opengever/meeting/command.py
@@ -195,7 +195,6 @@ class CreateGeneratedDocumentCommand(CreateDocumentCommand):
         template = self.document_operations.get_sablon_template(self.meeting)
         sablon = Sablon(template)
         sablon.process(self.document_operations.get_meeting_data(self.meeting).as_json())
-        assert sablon.is_processed_successfully(), sablon.stderr
         return sablon.file_data
 
     def execute(self):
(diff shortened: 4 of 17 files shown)


opengever/meeting/browser/sablontemplate.py
```diff
@@ -70,7 +71,8 @@ class FillMeetingTemplate(BrowserView):
     def __call__(self):
         sablon = Sablon(self.context)
         sablon.process(json.dumps(SAMPLE_MEETING_DATA))
-        assert sablon.is_processed_successfully(), sablon.stderr
+        if not sablon.is_processed_successfully():
```
deiferni: `fill_meeting_template` is only available for managers, i still would like to see the error message/trace from sablon here in the browser for easier debugging

opengever/meeting/browser/views.py
```diff
@@ -38,3 +41,21 @@ def __call__(self):
...
+    def __call__(self):
+        # this is an error, we must not commit
+        transaction.abort()
+        message = _(u'Error while processing Sablon template')
+        IStatusMessage(self.request).addStatusMessage(
+            message, type='error')
+        raise Redirect(self.get_redirect_url())
+
+    def get_redirect_url(self):
+        if hasattr(self.context.container, "absolute_url"):
```
deiferni: not sure if with my other comments this mechanism will still be used/desired. it may be a better solution to omit this indirection and just handle the exception directly in the 1-2 views that still remain.

opengever/base/browser/resources/generateToc.js
```diff
@@ -11,6 +11,9 @@
     request.responseType = "arraybuffer";
     request.onload = function() {
       loaded.resolve();
+      if (this.getResponseHeader('reload')) {
```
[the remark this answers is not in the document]
njohner: As discussed, this cannot be handled by a redirect  as `XMLHttpRequest` follows the redirect automatically, so that the first response with the error and location header cannot be caught in javascript.
Instead of using a custom header, I changed this to use a `422 Unprocessable Entity`error instead.

opengever/meeting/browser/meetings/agendaitem.py
```diff
@@ -139,6 +140,11 @@ def wrapper(*args, **kwargs):
                           u"configured."),
                 status=501).dump()
 
+        except SablonProcessingFailed:
+            return JSONResponse(getRequest()).error(
+                _('Error while processing Sablon template'),
+                status=501).dump()
```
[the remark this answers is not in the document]
njohner: No good reason, apart that 501 is used when the template is missing, and I thought that was a very similar error. Changed it to 500.

opengever/base/browser/resources/generateToc.js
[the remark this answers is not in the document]
deiferni: could you use a code from the `5xx` range here? this is a server-side error. `4xx` are for client-side errors, so if you use those that may be misleading for the future reader.

opengever/meeting/command.py
```diff
@@ -308,34 +313,41 @@ def get_header_sablon(self):
...
         template = self._get_paragraph_template()
         if template is None:
             return
+        return self.process_sablon(
+            template, ProtocolData(self.meeting, [agenda_item]).as_json())
 
-        return Sablon(self._get_paragraph_template()).process(
-            ProtocolData(self.meeting, [agenda_item]).as_json())
+    def process_sablon(self, template, data):
+        sablon = Sablon(template).process(data)
```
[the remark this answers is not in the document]
njohner: I do need the `context` because I need to redirect, and I thought adding a context to the process seemed a bit wrong.

opengever/meeting/browser/sablontemplate.py
njohner: Ok I now display the sablon error message here.

opengever/base/browser/resources/generateToc.js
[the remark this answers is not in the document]
njohner: 500 it is

njohner:
Do not merge! I'll squash the last commits when review is finished

opengever/base/browser/resources/generateToc.js
```diff
@@ -11,6 +11,9 @@
     request.responseType = "arraybuffer";
     request.onload = function() {
       loaded.resolve();
+      if (this.status == 500) {
```
deiferni: ehmmm it just came to my mind this may be a very bad idea, what if there is another error? this will cause a reload loop 😂.

opengever/meeting/command.py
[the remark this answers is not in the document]
deiferni: that was my other comment 😉. i think in this case it would make more sense. i think if would be preferrable if the model always raises the exception (you do not need the context there), and then the view has (or does not have) a try-except statement where it then knows what to do with the exception in its context.

opengever/base/browser/resources/generateToc.js
njohner: As discussed, this cannot really lead to a loop, but still, let's get back to custom header for safety.

njohner:
Ok I rewrote history and I think this is ready

njohner:
Note that the meeting cannot be closed if the protocol cannot be regenerated.

opengever/meeting/model/meeting.py
```diff
@@ -220,7 +221,11 @@ def update_protocol_document(self, overwrite=False):
         operations = ProtocolOperations()
         command = MergeDocxProtocolCommand(
             self.get_dossier(), self, operations)
-        command.execute(overwrite=overwrite)
+        try:
+            command.execute(overwrite=overwrite)
+        except SablonProcessingFailed:
```
deiferni: this seems not necessary 😉. side-note: if you want to re-raise just type `raise`.

njohner: 🤣 I think that is some leftover

PR #4938, opened 2018-10-04
Add new docproperty ogg.dossier.external_reference.

### LINKED ISSUE #4929: Neues Docproperty `external_reference` einführen.
Es soll ein zusätzliches Docproperty `ogg.dossier.external_reference` hinzugefügt werden, als Wert soll das Feld `IDossier.external_reference`

Note: Bitte auch Dokumentation anpassen (https://docs.onegovgever.ch/admin-manual/docproperties/list/#doc-properties-dossier).

njohner:
Add a new docproperty `ogg.dossier.external_reference` and update the documentation accordingly.

resolves #4929


docs/public/admin-manual/docproperties/list.rst
```diff
@@ -33,6 +33,7 @@ Doc-Properties aktueller Benutzer:
 
 Doc-Properties Dossier:
 ~~~~~~~~~~~~~~~~~~~~~~~
+- ``ogg.dossier.external_reference`` - Externe Referenz, welches das Dokument enthält
```
phgross: I would remove the external_reference property to the bottom of the list, because its the least important.

njohner: Oh not even, seems I don't know my alphabet anymore...

njohner:
So this is ready.
For your question, the value will not be none, instead the property is just not there. `_add_property` does not add a docprop `if not value`. Then word simply assigns no value to the property if it cannot find it, and the field is replaced by nothing. So that works fine.

njohner:
Do we really want this in the oneoffixx xml? For now there are very few properties in the xml, and this one is actually a dossier docproperty?
I've made the required changes but I'm not sure this is really useful.

Rotonen:
I'm left with the impression we dump all of the docproperties into the documents in case a template will ever want to use those for whatever purpose.

njohner:
For now there is only the title, sequence_number and reference_number. But in that case we should maybe make a ticket to add all docprops in the oneoffixx xml?

Rotonen:
Or alternatively confirm if I'm wrong on this one. @buchi ?

PR #3092, opened 2017-07-05
Use OGMail original_message-file for bumblebee-conversion if available.

elioschmutz:
Dieser PR benutzt die Originale .msg-Datei bei Mails sofern diese vorhanden ist.

Er stellt u.a. einen Upgradestep zur Verfügung welcher vorhandene Mails mit .eml Dateien neu konvertiert.

screen2

closes #3023

PR #4943, opened 2018-10-08
Notify IAttachmentsDeletedEvent after updating the message.

### LINKED ISSUE #4939: Removing mail attachments does not updated the checksum
Steps to reproduce:
 - Upload a mail with attachments 
 - Extract attachments and remove them

The bumblebee checksum is not updated and no store is triggered tough the file has changed, this lead to checksum mismatch exceptions in the bumblebee service.

njohner:
There were two issues here:
* The `IAttachmentsDeletedEvent` was notified before updating the message. So the checksum was not updated
* `get_file` is used as `primary_key` in `OGMailBumblebeeDocument`. This will return the original_message if it is present, so that the checksum will never change, even when the attachements are deleted (this updates the `message`). 

After looking at the history it seems that we use the original message on purpose, although I'm not too sure why, except maybe if the conversion to eml doesn't work properly? (https://github.com/4teamwork/opengever.core/pull/2948 and https://github.com/4teamwork/opengever.core/pull/3092). So we either have to revert that change, so that we always use `message` as primary key, or we leave it as is and have two different behaviors for E-mails, depending on whether there is an original_message or not.

**Note**
I'm not completely sure how this actually led to checksum mismatch exceptions. The normal result of this bug was that the preview was simply not updated, but the checksum neither. To get a mismatch, we should then have called a `BaseBumblebeeDocument.calculate_checksum` without firing the bumblebee store. This happens for example if `BumblebeeConverter.reindex` is called, but I did not find any instance of such a reindexing without storing.

resolves #4939


opengever/core/upgrades/20181005183203_update_og_mail_bumblebee_document_checksums/upgrade.py
```diff
@@ -0,0 +1,23 @@
...
+from opengever.bumblebee import is_bumblebee_feature_enabled
+from opengever.mail.mail import IOGMailMarker
+from opengever.mail.mail import IOGMail
+from ftw.bumblebee.interfaces import IBumblebeeDocument
+
+
+class UpdateOGMailBumblebeeDocumentChecksums(UpgradeStep):
+    """Update og mail bumblebee document checksums.
+    """
+
```
deiferni: could this upgrade be `deferrable`?

njohner: I guess so. But shouldn't that actually be quite cheap to calculate? Of course we need to touch all E-mails, recalculate the checksum, but nothing else will happen for most. Only those where the checksum has changed will be reindexed and the preview rebuilt.

phgross: Yep, please make this Upgradestep deferrable.

njohner:
Ok all done. I added the CL and made the upgrade deferrable.

PR #4947, opened 2018-10-08
Move the favorite action just after the documentFirstHeading.

### LINKED ISSUE #4747: Favorite button incorrectly styled in "view metadata" view and for paragraph templates
The **favorite button ☆** in the "view metadata" views of most (all?) content types gets incorrectly wrapped onto the next line:

favorites_btn_metadata

Example: https://dev.onegovgever.ch/fd/ordnungssystem/fuehrung/gemeinderecht/uebergeordnete-erlasse/view

This is also true when viewing (and not editing) paragraphtemplates, see http://localhost:8080/fd/vorlagen/meetingtemplate-1/paragraphtemplate-2

---

Instead it should look like it does on the overview view:

favorites_btn_overview

Example: https://dev.onegovgever.ch/fd/ordnungssystem/fuehrung/gemeinderecht/uebergeordnete-erlasse#dossiers

njohner:
The favorite-action is rendered by the below-content-title viewlet, but we actually want it on the same line as the title. This viewlet is rendered as a block in the default template, below the title. We use diazo here to move it outside of that viewlet and just after the document title.

This seems to fix the problem in all the mentioned cases:

**Dossier "view metadata"**
screen shot 2018-10-08 at 14 42 29

**Paragraph template view**
screen shot 2018-10-08 at 14 45 21

And it did not break any of the other views, for example:
**Normal document with title spannign more than one line**
screen shot 2018-10-08 at 14 42 24

resolves #4747

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -7,6 +7,7 @@ Changelog
 
 - Add trashing of protocol excerpts. [njohner]
 - Display title for NullObject in meeting template selection. [njohner]
+- Fix styling of favorite button. [njohner]
 - Exclude searchroots from subdossier listings. [Rotonen, lgraf]
 - Add filters (active and all) to membership listing. [njohner]
 - Do not add a task-reminder activity if task is finished. [elioschmutz]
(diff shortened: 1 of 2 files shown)


njohner:
Ok this should be ready. I included your comments and added a changelog and made sure it still worked.

PR #4954, opened 2018-10-10
Generate meeting zips asynchronously, with PDFs from the new demand endpoint.

### LINKED ISSUE #4775: Use demand endpoint to include preview pdfs in meeting zipexport
We want to include PDFs in the meeting zipexport so that users can easily upload the pdfs to grimlock or distribute them on other channels (e.g. e-mail). This has the advantage that an "immutable" read-only file is distributed and not a mutable word-document.

- use https://github.com/4teamwork/bumblebee/pull/85 to demand pdfs from bumblebee
- build polling view that provides a link to the zip with pdfs eventually, or runs into a timeout
- build endpoint for bumblebee to put the files for said zip, provide endpoint on meeting or committee
- store data in committee
- include pdfs in current export, no longer provide original file
- [x] 🚧 provide id for bumblebee, don't expose it in request URL but make it part of payload, naming see https://www.elastic.co/guide/en/elasticsearch/reference/6.2/tasks.html#_identifying_running_tasks.
- [x] 🚧 fill me with better spec

Open for discussion:
- [x] 🚧 concerns have arisen about receiving the files one-by one from bumblebee. it may be a better idea to fetch them once all are available.

History:

1. https://github.com/4teamwork/opengever.core/pull/4443
2. https://gever.4teamwork.ch/ordnungssystem/3/2/2/1/dossier-1763/dossier-2092/document-24575
3. https://gever.4teamwork.ch/ordnungssystem/3/5/3/dossier-2515/document-24613
4. https://gever.4teamwork.ch/ordnungssystem/3/5/3/dossier-2515/document-24799

deiferni:
Generate Zips for meetings asynchronously and include PDFs from bumblebee, if available. The PDFs are generated using bumblebees new demand endpoint, see https://github.com/4teamwork/bumblebee/pull/85.

🚧 ⚠️ Requires:
- [x] https://github.com/4teamwork/bumblebee/pull/87
- [x] https://github.com/4teamwork/ftw.bumblebee/pull/122 once this is merged and released this PR needs to be updated with the new `ftw.bumblebee` in `version.cfg` and branch info in `sources.cfg` should be removed.

User visible changes are that upon requesting a zip export from a meeting the zip is no longer served directly, but instead the following page appears:

screen shot 2018-10-10 at 14 30 06

The page will then poll GEVER until a zip is available or a timeout (~1min) is exceeded.
In the background it will trigger an asynchronous demand request to Bumblebee for a PDF for each document in the meeting that needs to be exported. Bumblebee will POST back a PDF or a skip/fail message response eventually. If a document could not be converted to PDF its original file will be included in the zip and the following warning message will appear:

screen shot 2018-10-10 at 14 29 52

If there is an error with polling the following message will be displayed:

screen shot 2018-10-10 at 14 48 04

Furthermore if the timeout is exceeded the following message will be displayed:

screen shot 2018-10-10 at 14 50 04

Internally the zip is kept for ~2 days and can be fetched via the same URL without triggering a new zip creation for that period. Also the url could be shared between different users that have permission to view the committee.

With this PR we also refactor how zips and the grimlock/meetings-app JSONs are generated as that code has been in a bad state with much duplication and bad readability.

_My apology to the reviewers, this feature has been WIP for a long period but new things have only been added in small chunks, so i decided to leave some not easy to rebase early commits in. There could be value in preserving the history there, also i think it is not worth the extra effort of tearing everything apart again. Please keep this in mind while reviewing, especially code from the earlier commits will be heavily changed again in later commits._

Closes #4775.


opengever/meeting/browser/meetings/zipexport.py
```diff
@@ -1,30 +1,133 @@
...
+
+def require_public_id_parameter(request):
+    public_id = request.get('public_id')
+    if not public_id:
+        raise BadRequest('must supply valid public_id of zip-job.')
+
+    return public_id
+
+
+def require_exporter(request, meeting, str_public_id):
```
lukasgraf: `get_exporter`?

opengever/meeting/browser/meetings/zipexport.py
```diff
@@ -1,30 +1,133 @@
...
 from ZPublisher.Iterators import filestream_iterator
 import json
 import os
-import pytz
+import uuid
 
 
-class MeetingZipExport(BrowserView):
-    """Iterate over meeting contents and return results in a .zip archive."""
+def as_uuid(id_from_request):
```
lukasgraf: What's the reason for this? (i.e. why not just stick with strings?)

opengever/meeting/browser/meetings/zipexport.py
```diff
@@ -34,108 +137,106 @@ def visible_in_actions_menu(self):
...
+        return self.model.get_url(view='export-meeting-zip')
 
-        with ZipGenerator() as generator:
-            # Protocol
-            generator.add_file(*self.get_protocol())
 
-            # Agenda items
-            self.add_agenda_items_attachments(generator)
-            self.add_agenda_item_proposal_documents(generator)
+class MeetingZipExport(BrowserView):
```
lukasgraf: `MeetingZipExportView`? Would help distinguish it from `MeetingZipExporter`

opengever/meeting/zipexport.py
```diff
@@ -0,0 +1,385 @@
...
+        document_id = self._opaque_id_to_document_id(opaque_id)
+        document_job = self.zip_job['documents'][document_id]
+        document_job['status'] = 'skipped'
+
+    def is_finished_converting(self):
+        return self.get_status()['converting'] == 0
+
+    def get_status(self):
+        status = {
+            'skipped': 0,
```
lukasgraf: Stringly typed - use int constants?

opengever/meeting/zipexport.py
```diff
@@ -0,0 +1,385 @@
...
+            'finished': 0,
+            'converting': 0,
+            'zipped': 0,
```
lukasgraf: How could this ever change?

opengever/meeting/browser/resources/zip_export_app.js
```diff
@@ -0,0 +1,74 @@
...
+    template: '#meeting-zip-export',
+    el: '#meeting-zip-export-vue-app',
+    data: {
+      i18n: {},
+      isFinished: false,
+      isTimeout: false,
+      isError: false,
+      isDocumentSkipped: false,
+      downloadurl: null,
+      pollurl: null,
```
lukasgraf: Remove non-reactive properties.

opengever/meeting/zipexport.py
```diff
@@ -0,0 +1,385 @@
...
+        document_job['status'] = 'skipped'
+
+    def is_finished_converting(self):
+        return self.get_status()['converting'] == 0
+
+    def get_status(self):
+        status = {
+            'skipped': 0,
+            'finished': 0,
+            'converting': 0,
```
lukasgraf: I'm missing a `failed` here. `'skipped'` only accounts for unconvertable formats.

opengever/meeting/browser/resources/zip_export_app.js
```diff
@@ -0,0 +1,74 @@
...
+
+  var app = new Vue({
+    template: '#meeting-zip-export',
+    el: '#meeting-zip-export-vue-app',
+    data: {
+      i18n: {},
+      isFinished: false,
+      isTimeout: false,
+      isError: false,
+      isDocumentSkipped: false,
```
lukasgraf: `hasSkippedDocuments`?

opengever/meeting/zipexport.py
```diff
@@ -0,0 +1,385 @@
...
+        return tuple(self._documents)
+
+    def _collect(self, document):
+        if not IBumblebeeDocument(document).has_file_data():
+            return
+
+        self._documents.append(document)
+
+
+class MeetingZipExporter(object):
```
lukasgraf: `MeetingZipExporter` "adapts" full meetings as well as single documents, dependening on what it's getting asked, right? That took me a while to realize, and makes some of the code quite hard to read IMHO.

opengever/meeting/zipexport.py
```diff
@@ -0,0 +1,385 @@
...
+
+        # prepare annotations of committee
+        annotations = IAnnotations(self.committee)
+        if ZIP_JOBS_KEY not in annotations:
+            annotations[ZIP_JOBS_KEY] = OOBTree()
+        self.zip_jobs = annotations[ZIP_JOBS_KEY]
+
+        # prepare or figure out public/internal ids
+        if opaque_id:
+            self.internal_id = self._opaque_id_to_interal_id(opaque_id)
```
deiferni: as discussed, a rename to

```python
self.internal_id = self._get_internal_id_part(combined_doc_job_id)
```
will improve readability.

opengever/meeting/zipexport.py
```diff
@@ -0,0 +1,385 @@
...
+            self.zip_job = self.zip_jobs[self.internal_id]
+            self.public_id = self.zip_job['public_id']
+        elif public_id:
+            self.public_id = public_id
+            self.zip_job = self.zip_jobs[self.public_id]
+            self.internal_id = self.zip_job['internal_id']
+        else:
+            self.internal_id = uuid.uuid4()
```
deiferni: as discussed the requirement that public and internal id should be different has been dropped, thus `public_id` and `internal_id` can be unified, to e.g. `job_id`.

opengever/meeting/zipexport.py
```diff
@@ -0,0 +1,385 @@
...
+        zip_job['documents'] = OOBTree()
+        self.zip_jobs[self.internal_id] = zip_job
+        self.zip_jobs[self.public_id] = zip_job
+
+        self.zip_job = zip_job
+        return zip_job
+
+    def _append_document_job_metadata(self, zip_job, document, status):
+        document_info = OOBTree()
+        document_info['intid'] = getUtility(IIntIds).getId(document)
```
deiferni: we should use `IUUID` to resolve the document here, especially since it has been used in recent commits as part of `opaque_id`. stick to one id 😄

lukasgraf:
@deiferni as discussed yesterday, I'm hijacking this and will implement the changes we discussed.

lukasgraf:
Regarding the **ConflictErrors**:

These happen all the time when a single user creates single ZIP. They occur on `bumblebee_trigger_demand`, `receive_meeting_zip_pdf` and (occasionally) `bumblebee_download`. The reason for these ConflictErrors is that for all kinds of actions an `IBumblebeeJournal` entry gets written. These get persisted in the DB (on GET requests as well), in a Souper record. Because the log entries have a monotonically increasing timestamp as the key, those BTrees are extremely prone to conflicts.

Therefore I will introduce an `IBumblebeeJournal` adapter for GEVER that logs them to the regular instance log. At a later point, well probably want to separate those out to a dedicated `instance?-bumblebee.log`, but that will require some modifications to `ftw.recipe.deployment`.

lukasgraf:
@phgross this is ready for you to review my last commits on top. *(Will rebase afterwards)*

opengever/meeting/zipexport.py
```diff
@@ -369,13 +417,12 @@ def _queue_demand_job(self, document):
         else:
             return 'skipped'
 
-    def _get_doc_in_job_id(self, document):
-        return '{}:{}'.format(self.job_id, IUUID(document))
-
-    def _doc_in_job_id_to_job_id(self, doc_in_job_id):
+    @staticmethod
+    def _extract_job_id(doc_in_job_id):
```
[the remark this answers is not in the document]
lukasgraf: Changed public methods (and fixupped into commit that introduced them) 👍

opengever/meeting/zipexport.py
```diff
@@ -259,8 +258,11 @@ def demand_pdfs(self):
 
     def get_document(self, doc_in_job_id):
         document_id = self._doc_in_job_id_to_document_id(doc_in_job_id)
-        document_job = self.zip_job['documents'][document_id]
-        return getUtility(IIntIds).getObject(document_job['intid'])
+        assert document_id in self.zip_job['documents']
+        catalog = api.portal.get_tool('portal_catalog')
+        brain = catalog.unrestrictedSearchResults(UID=document_id)[0]
```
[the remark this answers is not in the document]
lukasgraf: This is needed because the `receive_meeting_zip_pdf` view is public, and will be called with an anonymous user context by bumblebee. Authentication is [handled by `ftw.bumblebee`](https://github.com/4teamwork/ftw.bumblebee/blob/master/ftw/bumblebee/browser/callback.py#L27-L34) using a Bumbleebee access token.

lukasgraf:
@phgross updated and rebased onto `master`.

deiferni:
just for the record: thank you for picking this up and for polishing it ✨ 💛!

PR #4955, opened 2018-10-10
Do not leak setting the sql dialect to Oracle in the Memory DB layer

Rotonen:
This gets us some 300 tests further along a sequential test run.


Rotonen:
There's still an issue where the session needs to get purged from the `z3c.saconfig` session cache, which fixes the savepoint support, as now the ZCML event gets handled and the connection settings are correct, but doing so then breaks dictstorage in the functional tests, but that'll get chased separately later.

Rotonen:
We did not have an issue for this one, I spotted this when looking at an another flaky test / leak issue.

PR #4957, opened 2018-10-12
Additional task field: is_private

elioschmutz:
Dieser PR implementiert das neue Feld für Aufgaben: `is_private`

bildschirmfoto 2018-10-10 um 13 54 12

Issuer: #4763

### DIFF
--- a/opengever/core/upgrades/20181012105107_persist_is_private_attribute_to_tasks/upgrade.py
+++ b/opengever/core/upgrades/20181012105107_persist_is_private_attribute_to_tasks/upgrade.py
@@ -1,0 +1,23 @@
+from ftw.upgrade import UpgradeStep
+from opengever.base.default_values import object_has_value_for_field
+from opengever.task.task import ITask
+
+
+class PersistIsPrivateAttributeToTasks(UpgradeStep):
+    """Persist is private attribute to tasks.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
+        for task in self.objects(
+                {'object_provides': 'opengever.task.task.ITask'},
+                "Set default value for is_private attribute for tasks"):
+            field = ITask['is_private']
+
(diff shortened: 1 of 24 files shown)


opengever/core/upgrades/20181012105107_persist_is_private_attribute_to_tasks/upgrade.py
```diff
@@ -0,0 +1,22 @@
...
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
+        for task in self.objects(
+                {'object_provides': 'opengever.task.task.ITask'},
+                "Set default value for is_private attribute for tasks"):
+            field = ITask['is_private']
+
+            if task.is_private is not field.default:
```
phgross: Why is that necessary? this field has been introduced within this commit, so should no be possible, to already has  differing value than the default.

elioschmutz: It's to prevent data-loose if you falsely run the upgradestep twice...

But it seems, that it's not wanted :-D I'll remove it.

phgross: Ohh good point, that totally makes sense! 

But please use `opengever.base.default_values.object_has_value_for_field` to check if there is a value or not.

elioschmutz: 👍 updated again...

Rotonen:
https://ci.4teamwork.ch/builds/195392/tasks/320246

Is `Failure in test test_keywords_are_linked_to_search_on_overview (opengever.dossier.tests.test_dossier.TestDossier)` a known flaky test, or did this somehow introduce a weird side effect?

That failure went away as I reran the tests.

PR #4982, opened 2018-10-19
Sharing form fixes

phgross:
- **Fix removal of all role assignments**: Currently the removal of all local roleassignments has been skipped by mistake in the sharing endpoint. We fix that, by making sure that all
RoleAssignments are cleared, when no assignments exists.

- Do not show search results multiple times.

Backport to `2018.4-stable`

### DIFF
--- a/opengever/sharing/browser/resources/sharing.js
+++ b/opengever/sharing/browser/resources/sharing.js
@@ -98,9 +98,14 @@ var sharingApp = {
 
       this.requester.get(this.endpoint, { params: params }).then(function (response) {
 
-        this.entries = this.entries.filter(function(i) { return !i.disabled; });
+        // Drop former search results, which has not been changed.
+        this.entries = this.entries.filter(function(i) { return i.disabled === false; });
+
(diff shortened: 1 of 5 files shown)


opengever/sharing/browser/resources/sharing.js
```diff
@@ -98,9 +98,14 @@ var sharingApp = {
...
 
-        this.entries = this.entries.filter(function(i) { return !i.disabled; });
+        // Drop former search results, which has not been changed.
+        this.entries = this.entries.filter(function(i) { return i.disabled === false; });
+
+        var current_ids = this.entries.map( function(i) { return i.id})
         this.entries = this.entries.concat(
-          response.data['entries'].filter(function(i) { return i.disabled != false; }));
+          response.data['entries'].filter(function(i) {
+            return i.disabled != false && current_ids.indexOf(i.id) == -1;
```
lukasgraf: Those should be `!==` and `===`, no?

phgross: You're right, updated ...

PR #4996, opened 2018-10-22
Sanitize eCH-0147 imports via a JSON cycle

### LINKED ISSUE #4376: Imported content from eCH-0147 messages has values of wrong type
It seems that textline fields are assigned a more specific XML type instead of a Python unicode string during eCH-0147 import.

In plone.restapi serialization this results in an error like this:
```
2018-05-29 15:51:20 ERROR Zope.SiteErrorLog 1527601880.970.295844106497 http://localhost:8081/fd/ordnungssystem/ressourcen-und-support/ict/allgemeines/dossier-2039/document-1098/GET_application_json_
Traceback (innermost last):
  Module plone.rest.service, line 23, in __call__
  Module plone.restapi.services, line 19, in render
  Module plone.restapi.services.content.get, line 19, in reply
  Module opengever.api.document, line 19, in __call__
  Module plone.restapi.serializer.dxcontent, line 84, in __call__
  Module plone.restapi.serializer.dxfields, line 76, in __call__
  Module plone.restapi.serializer.converters, line 42, in json_compatible
  Module zope.component.hooks, line 104, in adapter_hook
  Module plone.restapi.serializer.converters, line 103, in dict_converter
  Module plone.restapi.serializer.converters, line 42, in json_compatible
  Module zope.component.hooks, line 104, in adapter_hook
  Module plone.restapi.serializer.converters, line 56, in default_converter
TypeError: No converter for making u'application/pdf' (<class 'pyxb.binding.datatypes.token'>) JSON compatible.
```

We should always store the correct type defined by the field type.

Rotonen:
Made a custom JSON (de)serializer for eCH-0147  imports as we feed these to the `plone.restapi` JSON deserializer. This way everything is guaranteedly just a string unless it's whitelisted as a date field and we can easily amend any future additions to the field types.

There are also tests for importing data via eCH-0147 and a deferrable upgrade step for going through all of the fields on all of the schemata on all of the non-bundle-imported objects for PyXB data types and casting the strings, ints, booleans, dates and datetimes thereof into their appropriate field types.

Closes #4376

### DIFF
--- a/opengever/ech0147/utils.py
+++ b/opengever/ech0147/utils.py
@@ -3,25 +3,21 @@ from opengever.document.subscribers import set_digitally_available
 from opengever.ech0147.mappings import INV_CLASSIFICATION_MAPPING
 from opengever.ech0147.mappings import INV_PRIVACY_LAYER_MAPPING
 from opengever.ech0147.mappings import INV_PUBLIC_TRIAL_MAPPING
+from opengever.ech0147.serializer import ECH0147Serializer
 from plone.restapi.interfaces import IDeserializeFromJson
-from pyxb.binding.datatypes import date as pyxb_date
 from random import randint
 from zope.component import queryMultiAdapter
 from zope.container.interfaces import INameChooser
 from zope.event import notify
 from zope.lifecycleevent import ObjectModifiedEvent
+import json
 import os.path
 import transaction
 
 
-def convert_dates(data):
-    """convert pyxb dates to real python datetime dates, otherwise field
-    validation for date fields will raise.
-    """
-
-    for key, value in data.items():
-        if isinstance(value, pyxb_date):
-            data[key] = value.date()
+def sanitize_metadata(metadata):
+    """We do a JSON serialization cycle to get rid of any XML data types."""
+    return json.loads(json.dumps(metadata, cls=ECH0147Serializer), cls=ECH0147Serializer)
 
 
 def create_dossier(container, dossier, zipfile, responsible):
@@ -54,11 +50,8 @@ def create_dossier(container, dossier, zipfile, responsible):
         metadata['comments'] = u'\n'.join(
             [k.value() for k in dossier.comments.comment])
 
-    convert_dates(metadata)
-
-    deserializer = queryMultiAdapter((obj, obj.REQUEST),
-                                     IDeserializeFromJson)
-    deserializer(validate_all=True, data=metadata, create=True)
+    deserializer = queryMultiAdapter((obj, obj.REQUEST), IDeserializeFromJson)
+    deserializer(validate_all=True, data=sanitize_metadata(metadata), create=True)
 
     # Rename dossier
     chooser = INameChooser(container)
@@ -107,11 +100,8 @@ def create_document(container, document, zipfile):
     if document.keywords:
         metadata['keywords'] = [k.value() for k in document.keywords.keyword]
 
-    convert_dates(metadata)
-
-    deserializer = queryMultiAdapter((obj, obj.REQUEST),
-                                     IDeserializeFromJson)
-    deserializer(validate_all=True, data=metadata, create=True)
+    deserializer = queryMultiAdapter((obj, obj.REQUEST), IDeserializeFromJson)
+    deserializer(validate_all=True, data=sanitize_metadata(metadata), create=True)
 
     if document.files:
         file_ = document.files.file[0]
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -38,6 +38,7 @@ Changelog
 - Respect ISharingConfiguration black- and white_list_prefix for selectable groups in add-team form. [elioschmutz]
 - Restrict selectable orgunits on add-team form to orgunits of the current adminunit only. [elioschmutz]
 - Update ftw.tabbedview to 4.1.2. [Kevin Bieri]
+- Sanitize eCH-0147 imports. [Rotonen]
 - Fix dispatching notifications while recording a TaskReminderActivity. [elioschmutz]
 - Prefill task-reminder select field on task response form. [elioschmutz]
 - Use the same template for default notifiaction email like the daily-digest template. [elioschmutz]
--- a/opengever/core/upgrades/20181109105732_ensure_all_fields_of_e_ch_0147_imported_objects_are_properly_typed/upgrade.py
+++ b/opengever/core/upgrades/20181109105732_ensure_all_fields_of_e_ch_0147_imported_objects_are_properly_typed/upgrade.py
@@ -1,0 +1,87 @@
+from contextlib import contextmanager
+from datetime import date
+from datetime import datetime
+from dateutil.parser import parse as date_parser
+from ftw.upgrade import UpgradeStep
+from opengever.bundle.sections.constructor import BUNDLE_GUID_KEY
+from opengever.ech0147.interfaces import IECH0147Settings
+from plone import api
+from plone.dexterity.utils import iterSchemata
+from pyxb.binding.datatypes import boolean as pyxb_boolean
+from pyxb.binding.datatypes import date as pyxb_date
+from pyxb.binding.datatypes import dateTime as pyxb_datetime
+from pyxb.binding.datatypes import int as pyxb_int
+from pyxb.binding.datatypes import string as pyxb_string
+from zope.annotation import IAnnotations
+from zope.schema import getFieldsInOrder
+import logging
+import pyxb.binding.datatypes
+
+
+logger = logging.getLogger("opengever.core")
+
+
+@contextmanager
+def writable(field):
+    try:
+        original_readonly = field.readonly
+        field.readonly = False
+        yield field
+    finally:
+        field.readonly = original_readonly
+
+
+class EnsureAllFieldsOfECH0147ImportedObjectsAreProperlyTyped(UpgradeStep):
+    """Ensure all fields of eCH-0147 imported objects are properly typed.
+    """
+    deferrable = True
+
+    def __call__(self):
+        self.install_upgrade_profile()
+
+        # Skip if eCH-0147 imports are not enabled
+        if not api.portal.get_registry_record("ech0147_import_enabled", interface=IECH0147Settings):
+            return
+
+        # Only query non-imported-from-bundle documents and dossiers
+        affected_types = ("opengever.dossier.businesscasedossier", "opengever.document.document")
+        query = {"portal_type": affected_types, "bundle_guid": None}
+        primitive_pyxb_types = pyxb.binding.datatypes.__dict__.get("_PrimitiveDatatypes")
+        trivial_pyxb_types = (pyxb_int, pyxb_boolean, pyxb_string)
+        trivial_types = (int, basestring)
+
+        for obj in self.objects(query, "Ensure all fields of eCH-0147 imported objects are properly typed."):
+            # Skip imported-from-bundle objects not on the index
+            if IAnnotations(obj).get(BUNDLE_GUID_KEY):
+                continue
+            for schema in iterSchemata(obj):
+                for name, field in getFieldsInOrder(schema):
+                    value = getattr(field.interface(obj), name, None)
+                    value_type = type(value)
+                    # Only touch pyxb typed values
+                    if field._type and value is not None and isinstance(value_type, primitive_pyxb_types):
+                        object_path = "/".join(obj.getPhysicalPath())
+                        logger.info(
+                            "Found PyXB values in object %s field %s field type %s value type %s.",
+                            object_path,
+                            name,
+                            repr(field._type),
+                            repr(value_type),
+                        )
+                        if isinstance(value_type, trivial_pyxb_types) and field._type in trivial_types:
+                            with writable(field) as wfield:
+                                wfield.set(wfield.interface(obj), wfield._type(value))
+                        elif isinstance(value_type, pyxb_date) and field._type is date:
+                            with writable(field) as wfield:
+                                wfield.set(wfield.interface(obj), wfield._type.fromordinal(wfield.toordinal()))
+                        elif isinstance(value_type, pyxb_datetime) and field._type is datetime:
+                            with writable(field) as wfield:
+                                wfield.set(wfield.interface(obj), date_parser(value.ISO()))
+                        else:
+                            logger.warn(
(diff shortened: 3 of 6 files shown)


opengever/ech0147/serialiser.py
```diff
@@ -0,0 +1,28 @@
+from pyxb.binding.datatypes import date as pyxb_date
+import dateutil.parser
+import json
+
+
+ECH_0147_DATE_FIELDS = ('start', 'document_date')
```
phgross: What about all other dates from document and dossier?

opengever/ech0147/utils.py
```diff
@@ -3,27 +3,18 @@
 from opengever.ech0147.mappings import INV_CLASSIFICATION_MAPPING
 from opengever.ech0147.mappings import INV_PRIVACY_LAYER_MAPPING
 from opengever.ech0147.mappings import INV_PUBLIC_TRIAL_MAPPING
+from opengever.ech0147.serialiser import ECH0147Serializer
```
buchi: Also use US English spelling for the module name: `serializer` instead of `serialiser`.

opengever/ech0147/utils.py
```diff
@@ -54,10 +45,9 @@ def create_dossier(container, dossier, zipfile, responsible):
         metadata['comments'] = u'\n'.join(
             [k.value() for k in dossier.comments.comment])
 
-    convert_dates(metadata)
-
-    deserializer = queryMultiAdapter((obj, obj.REQUEST),
-                                     IDeserializeFromJson)
+    metadata = json.dumps(metadata, cls=ECH0147Serializer)
```
buchi: When reading the code, it's not obvious why this is done. I would at least add a comment or wrap it in a function with an appropriate name, e.g. sanitize_xml_types.

opengever/ech0147/serialiser.py
```diff
@@ -0,0 +1,28 @@
...
+        return obj
+
+    def decode(self, string):
+        """We only get flat enough dicts to know dates are on the top level."""
+        obj = json.loads(string)
+
+        for key, value in obj.items():
+            if isinstance(value, basestring):
+                if key in ECH_0147_DATE_FIELDS:
+                    obj[key] = dateutil.parser.parse(value).date()
```
buchi: Is this needed at all? plone.restapi Deserializer should be able to parse dates.

Rotonen: It does, but not from strings. Thus we explicitly convert known fields to dates.

opengever/ech0147/serialiser.py
Rotonen: Our tests use none.

https://www.ech.ch/dokument/26e12eb1-8982-4e48-9394-e706ab27f500

Yes, there are more, but seems we just test the minimum mandatory set so far.

Rotonen: Actually we discard everything else of the incoming metadata as far as I can see.

https://github.com/4teamwork/opengever.core/blob/9e1b74b6565bf3d41af827aa2a46c2376b9af135/opengever/ech0147/utils.py#L41

https://github.com/4teamwork/opengever.core/blob/9e1b74b6565bf3d41af827aa2a46c2376b9af135/opengever/ech0147/utils.py#L91

Rotonen:
I'm left a bit unsure as to how to do the upgrade step.

First of all these steps would be huge and we'd need a way to run it in segments.

Secondly I'm unsure as to how to abstractly enough handle versioning and non-string fields.

Rotonen:
@phgross @buchi ready for rereview.

opengever/core/upgrades/20181024152606_ensure_all_fields_of_all_objects_are_of_the_proper_type/upgrade.py
```diff
@@ -0,0 +1,43 @@
...
+        # Only query non-imported-from-bundle documents and dossiers
+        affected_types = ('opengever.dossier.businesscasedossier', 'opengever.document.document')
+        query = {'portal_type': affected_types, 'bundle_guid': None}
+
+        for obj in self.objects(query, "Fixing broken fields on non-imported-from-bundle documents and dossiers."):
+            # Skip imported-from-bundle objects not on the index
+            if IAnnotations(obj).get(BUNDLE_GUID_KEY):
+                continue
+            for schema in iterSchemata(obj):
+                for name, field in getFieldsInOrder(schema):
```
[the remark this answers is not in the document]
Rotonen: My intuition is pulling the object into memory being the expensive bit per outer loop cycle. If we can find a way to reliably skip things based on what is on the metadata / brain, that'd help.

lukasgraf:
@Rotonen what manual tests did you do for the upgrade step?

Rotonen:
@lukasgraf I ran it over the example content data set with and without having manually broken some objects in the way the imports do and also in other ways.

Rotonen:
To clarify on 'other ways', at least `str` vs. `unicode` was handled correctly by casting with `._type()`.

opengever/core/upgrades/20181024152606_ensure_all_fields_of_all_objects_are_of_the_proper_type/upgrade.py
```diff
@@ -0,0 +1,43 @@
...
+                    value = getattr(field.interface(obj), name, None)
+                    # Allow empty values on non-required fields
+                    # Only do for fields which declare a type
+                    if value is not None and field._type:
+                        if type(value) != field._type:
+                            # Strings and Booleans
+                            if isinstance(value, (basestring, bool)):
+                                field.set(obj, field._type(value))
```
lukasgraf: This is missing the adaption of the object to the schema interface.

The full pattern for setting a field is always `field.set(field.interface(obj), value)` (see [here](
https://github.com/4teamwork/opengever.core/blob/master/opengever/base/default_values.py#L314) for example).

Adapting the object to the interface is necessary because that could result in a different storage being returned than the object itself. Default attribute storage does indeed store fields directly on the object as attributes, and in that case it makes no difference. `obj is field.interface(obj)` is true in that case.

However, there are other storages that store field data off-object, and those would be returned by `field.interface(obj)` in that case. Failing to adapt the obj to the interface would in that case result in the field data getting written to the obj - where it will *not* be read from.

opengever/core/upgrades/20181024152606_ensure_all_fields_of_all_objects_are_of_the_proper_type/upgrade.py
```diff
@@ -0,0 +1,43 @@
...
+                            # Dates and Datetimes
+                            elif isinstance(value, date):
+                                field.set(obj, field._type.fromordinal(value.toordinal()))
```
lukasgraf: Same, `field.set(field.interface(obj), value)`

opengever/core/upgrades/20181024152606_ensure_all_fields_of_all_objects_are_of_the_proper_type/upgrade.py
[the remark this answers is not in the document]
lukasgraf: @phgross as discussed, I don't think there's any value in further optimizing this. It's easier and less risky to just make use of the `deferrable` upgrade step and install this one during a maintenance window for customers with large deployments.

Rotonen:
My validator currently (correctly) trips up on the mismatch of the field value and field type for `changed` on documents.

This is a good catch as now I also handle read only fields in the upgrade steps (in a somewhat naive way).

Also robustified the upgrade step in regards to dates and date times a bit.

Ripe for an interim checkup while we wait for https://github.com/4teamwork/opengever.core/issues/5038 to get solved.

opengever/core/upgrades/20181102133533_ensure_all_fields_of_all_objects_are_of_the_proper_type/upgrade.py
```diff
@@ -0,0 +1,56 @@
...
+from ftw.upgrade import UpgradeStep
+from opengever.bundle.sections.constructor import BUNDLE_GUID_KEY
+from opengever.ech0147.interfaces import IECH0147Settings
+from plone import api
+from plone.dexterity.utils import iterSchemata
+from zope.annotation import IAnnotations
+from zope.schema import getFieldsInOrder
+
+
+class EnsureAllFieldsOfAllObjectsAreOfTheProperType(UpgradeStep):
```
lukasgraf: The longer I think about it, I'm not a fan of this upgrade step being as broad / generic as it is.

I would suggest that it should only transform `pyxb` values, but leave everything else alone. This is the case we know about, and that we're trying to fix here. This is also the change in behavior that is accompanied by this upgrade step.

The logic that aborts early if
- ECH0147 is not enabled or
- The object obviously was imported via an OGGBundle import

is appropriate to rule out large amounts of objects that can't be from an ECH0147 import, as a *performance optimization*.

But it is not sufficient to guarantee that all encountered objects definitely have been imported via ECH0147. This means that we might be fixing some fields for non-ECH0147 content on *some* deployments, while they remain unfixed for other deployments. That would result in us having only partially fixed some incorrectly typed fields - so in essence, they're not really fixed at all, we just introduced another permutation of cases we would need to deal with.

So my suggestion is to limit the fixing to `pyxb` types - something like

```
if not isinstance(value, pyxb.SomeCommonBaseClassForAllPYXBTypes):
    continue
```

opengever/core/upgrades/20181102133533_ensure_all_fields_of_all_objects_are_of_the_proper_type/upgrade.py
```diff
@@ -0,0 +1,56 @@
...
+                continue
+            for schema in iterSchemata(obj):
+                for name, field in getFieldsInOrder(schema):
+                    original_readonly = field.readonly
+                    field.readonly = False
+                    value = getattr(field.interface(obj), name, None)
+                    # Allow empty values on non-required fields
+                    # Only do for fields which declare a type
+                    if value is not None and field._type:
+                        if type(value) != field._type:
```
lukasgraf: Singletons like types should be compared using `is` instead of `==`.

opengever/core/upgrades/20181102133533_ensure_all_fields_of_all_objects_are_of_the_proper_type/upgrade.py
```diff
@@ -0,0 +1,56 @@
...
+                            # Strings, floats, integers and Booleans - simple casting works here
+                            if isinstance(field._type, (basestring, float, int)):
+                                field.set(field.interface(obj), field._type(value))
+                            # Standard library dates
+                            elif field._type == date:
+                                field.set(field.interface(obj), field._type.fromordinal(value.toordinal()))
```
lukasgraf: Did you verify that this doesn't strip TZ awareness like you suspected?

opengever/core/upgrades/20181102133533_ensure_all_fields_of_all_objects_are_of_the_proper_type/upgrade.py
```diff
@@ -0,0 +1,56 @@
...
+                            # Standard library datetimes
+                            elif field._type == datetime:
+                                field.set(field.interface(obj), date_parser(value.ISO()))
```
lukasgraf: Shouldn't there also be a check regarding what type `value` is before we blindly invoke `.ISO()` or `.toordinal()` on it? We definitely know it's not of the `field._type`, but other than that, we don't have much to go on. I'd do an `isinstance` check and skip the field (and log a message) if it's not of one of the incorrect types we expect to see.

opengever/core/upgrades/20181102133533_ensure_all_fields_of_all_objects_are_of_the_proper_type/upgrade.py
```diff
@@ -0,0 +1,56 @@
...
+        # Only query non-imported-from-bundle documents and dossiers
+        affected_types = ('opengever.dossier.businesscasedossier', 'opengever.document.document')
+        query = {'portal_type': affected_types, 'bundle_guid': None}
+
+        for obj in self.objects(query, "Fixing broken fields on non-imported-from-bundle documents and dossiers."):
+            # Skip imported-from-bundle objects not on the index
+            if IAnnotations(obj).get(BUNDLE_GUID_KEY):
+                continue
+            for schema in iterSchemata(obj):
+                for name, field in getFieldsInOrder(schema):
```
lukasgraf: Generally I'd like to see some logging for field values that get changed by this upgrade step. Obj path, field name, repr(old_type), repr(new_type).

opengever/core/upgrades/20181102133533_ensure_all_fields_of_all_objects_are_of_the_proper_type/upgrade.py
```diff
@@ -0,0 +1,56 @@
...
+                            elif field._type == date:
+                                field.set(field.interface(obj), field._type.fromordinal(value.toordinal()))
+                            # Standard library datetimes
+                            elif field._type == datetime:
+                                field.set(field.interface(obj), date_parser(value.ISO()))
+                            # Zope DateTimes
+                            elif field._type == DateTime:
+                                field.set(field.interface(obj), field._type(value.ISO()))
+
+                    field.readonly = original_readonly
```
lukasgraf: Stuff like that should generally be done in a context manager. Something along these lines (untested):

```python
from contextlib import contextmanager

@contextmanager
def writable(field):
    try:
        original_readonly = field.readonly
        yield field
    finally:
        field.readonly = original_readonly

with writable(field) as writable_field:
    # ...
```

opengever/core/upgrades/20181102133533_ensure_all_fields_of_all_objects_are_of_the_proper_type/upgrade.py
```diff
@@ -0,0 +1,73 @@
...
+from pyxb.binding.datatypes import string as pyxb_string
+from zope.annotation import IAnnotations
+from zope.schema import getFieldsInOrder
+
+
+@contextmanager
+def writable(field):
+    try:
+        original_readonly = field.readonly
+        yield field
```
lukasgraf: As I wrote, untested - this is missing `field.readonly = False`

deiferni:
i have reverted this PR. See https://github.com/4teamwork/opengever.core/issues/4376.

PR #5004, opened 2018-10-25
Backport mtest improvements over from buildout.coredev experimentation

Rotonen:
Here's a batch of small to medium improvements and usability changes I'm backporting from my efforts of using an mtest variant in `buildout.coredev`.

### DIFF
--- a/bin/mtest
+++ b/bin/mtest
@@ -170,7 +170,8 @@ def split(batch):
         return (batch, )
 
     original_count = batch.get('original_count')
-    chunksize = max(minsize_mapping.get(layer), int(ceil(original_count / float(CONCURRENCY))))
+    chunk_count = min(ceil(original_count / float(minsize_mapping.get(layer))), CONCURRENCY)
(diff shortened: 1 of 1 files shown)


Rotonen:
Test discovery `STDERR` is not a payload we're interested in.

The universal newlines mode of `check_output`simplifies the logic of parsing the subprocess outputs.

I've dropped the hunt for unclean `STDERR` as that side of things has gotten way out of hand on both upstream and for us now that `zope.testrunner` (finally, correctly) bleeds deprecation warnings through.

We're not using the `MTEST_PROCESSORS` environment variable anymore and I find it cleaner when the command only bows down to command line flags.

The runtime gains are directly related to splitting the layers more evenly.

Passing around less data to the pool jobs alleviates some of the pickling bursting.

The tiered log flushing is more robust than expected so we can get away with `imap_unordered` without garbled output lines and thus get to also simplify the results handling logic.

Having the ability to pass in the `--shuffle` flag enables more shakedown categories of daily and weekly runs down the line.

PR #5007, opened 2018-10-26
Restrict TaskQueries according to the task principals.

### LINKED ISSUE #4765: OGIP 46: Rechte Überprüfung bei Aufgaben Auflistungen.
Die Auflistungen Alle Aufgaben, Alle erteilten Aufgaben in der persönlichen Übersicht sowie die Aufgaben Auflistung auf Stufe Dossier dürfen nur noch Aufgaben, welche ein Zugriff auf die Aufgaben erlauben, auflisten.
Für die Rechteüberprüfung wird der bereits bestehende Index task_principals verwendet.

phgross:
With the `TaskPrincipals` we already sync principals allowed to see a Task in to the globalindex. But currently the task queries has not been restricted and therefore a task was always visible, for all accessors of a Dossier and for inbox-users all tasks of a current admin_unit has been visible.

With "OGIP 46 - personal task", we now introduced a possibility to disable the `inbox-user` agency permission and have to restrict the task listings etc. I decided to restrict all TaskQueries, which also solves an [Issue](https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/365) wich allowed including tasks in to pdf_task_listing, where the user was not allowed to access.

During implementation I've also realized that the current implementation of the principals getter was wrong and did not respect the block_local_roles flag. 

Maybe backported to `2018.4-stable` 

Closes #4765

### DIFF
--- a/opengever/globalindex/model/task.py
+++ b/opengever/globalindex/model/task.py
@@ -465,6 +465,21 @@ class Task(Base):
 
 class TaskQuery(BaseQuery):
 
+    roles_allowed_to_see_tasks = ('Manager', 'Administrator', 'Reader')
+
+    def restrict(self):
+        member = api.user.get_current()
+
+        # Check for global roles. Global Administrators, Readers and
+        # Managers are always able to access a task, so no need to restrict.
+        global_roles = set(api.user.get_roles(user=member))
+        if not set(self.roles_allowed_to_see_tasks).isdisjoint(global_roles):
+            return self
+
+        principals = member.getGroupIds() + [member.getId()]
+        return self.join(Task._principals).filter(
+            TaskPrincipal.principal.in_(principals)).distinct()
+
     def _extend_with_physical_path(self, query, field, path):
         if is_oracle():
             return query.filter(func.to_char(field) == path)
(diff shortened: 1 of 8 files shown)


opengever/globalindex/model/task.py
```diff
@@ -465,6 +465,21 @@ def _task_state_wrapper(self, text):
...
 
+    roles_allowed_to_see_tasks = ('Manager', 'Administrator', 'Reader')
+
+    def restrict(self):
+        member = api.user.get_current()
+
+        # Check for global roles. Global Administrators, Readers and
+        # Managers are always able to access a task, so no need to restrict.
+        global_roles = set(api.user.get_roles(user=member))
+        if not set(self.roles_allowed_to_see_tasks).isdisjoint(global_roles):
```
Rotonen: Does this behave expectedly if the user does not have global roles for some reason?

opengever/task/task.py
```diff
@@ -478,15 +479,25 @@ def get_principals(self):
         # index the principal which have View permission. This is according to
         # the allowedRolesAndUsers index but it does not car of global roles.
         allowed_roles = rolesForPermissionOn(View, self)
+        return list(self._get_principals_for_roles(allowed_roles))
```
Rotonen: Would we have to refactor a lot to just always return the set here?

opengever/globalindex/model/task.py
phgross: Yes, the user should have the `Authenticated` role anyways but it would also work with a user without any roles.

opengever/task/task.py
phgross: Yes and it's a "public" method which is used in different places where a list is expected, i'd like to go with that implementation.

opengever/globalindex/model/task.py
Rotonen:
```
>>> {'a', 'b'}.isdisjoint({'a'})
False
>>> {'a', 'b'}.isdisjoint({})
True
```

Maybe rather use a negation of a binary and between the sets?

phgross: Not sure what you mean ... 

```
>>> roles = set(('Manager', 'Administrator', 'Reader'))
>>> roles.isdisjoint(set(('Member', 'Manager', 'Contributor')))
False
>>> roles.isdisjoint(set(('Member', 'Contributor')))
True
>>> roles.isdisjoint(set())
True
```

opengever/task/task.py
```diff
@@ -478,15 +479,25 @@ def get_principals(self):
...
 
-        for principal, roles in _mergedLocalRoles(self).items():
-            for role in roles:
-                if role in allowed_roles:
-                    principals.append(safe_unicode(principal))
-                    break
+            parent = aq_parent(aq_inner(parent))
 
-        return principals
+        return set(principals)
```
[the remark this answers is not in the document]
phgross: 👍 updated ...

opengever/globalindex/model/task.py
Rotonen: I guess I thought of the check itself the wrong way. Works as intended.

phgross:
updated ready to review ...

phgross:
@lukasgraf updated ...

lukasgraf:
@phgross I added the `backport needed` label just in case, even though we're not sure yet, just to make sure we don't forget to check.

PR #5011, opened 2018-10-30
Add some sharing form UI improvements

phgross:
- Disable buttons during save request in sharing form, and give the the user feedback during waiting time, with a spinning circle icon.
- Improve error handling in sharing form, show error message when saving request failed.

kapture 2018-10-30 at 8 52 00


opengever/sharing/locales/de/LC_MESSAGES/opengever.sharing.po
```diff
@@ -49,6 +49,11 @@ msgstr "Keine geschützte Objekte in diesem Bereich gefunden."
 msgid "label_roles_successfully_changed"
 msgstr "Lokale Rolle erfolgreich gespeichert"
 
+#. Default: "Local roles save failed."
+#: ./opengever/sharing/browser/sharing.py
+msgid "message_save_failed"
+msgstr "Berechtigungen konnten nicht gespeichert werden."
```
lukasgraf: I would use the same naming here as for success: `Lokale Rollen ...` instead of `Berechtigungen ...`

PR #5013, opened 2018-10-30
Don't html-escape description for JSON data.

### LINKED ISSUE #5012: Agenda item description is escaped for sablon data.
screen shot 2018-10-30 at 15 00 16

deiferni:
The description is passed on to sablon which no longer treats the input as html which needs to be converted to word. So html-escapes will be treated literally which is wrong.

Fixes #5012.

### DIFF
--- a/opengever/meeting/model/agendaitem.py
+++ b/opengever/meeting/model/agendaitem.py
@@ -142,9 +142,11 @@ class AgendaItem(Base):
             self.title = title
 
     def get_description(self):
-        description = (self.submitted_proposal.description if self.has_proposal
-                       else self.description)
-        return to_html_xweb_intelligent(description) or None
+        return (self.submitted_proposal.description if self.has_proposal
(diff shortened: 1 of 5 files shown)


opengever/meeting/model/agendaitem.py
```diff
@@ -142,9 +142,11 @@ def set_title(self, title):
             self.title = title
 
     def get_description(self):
-        description = (self.submitted_proposal.description if self.has_proposal
-                       else self.description)
-        return to_html_xweb_intelligent(description) or None
+        return (self.submitted_proposal.description if self.has_proposal
```
phgross: You changed `get_description` to a no-longer escaped value, which potentially could open some XSS holes, did you doublechecked all uses of `get_description`?

deiferni: yes, it should only be used once where it needs to be serialized.

Issue #5016, opened 2018-10-30
error while rendering live-search results


deiferni:
https://github.com/4teamwork/opengever.ftw/issues/78

PR #5019, opened 2018-10-31
Fix rendering issue in livesearch reply view.

deiferni:
Fix it for solr and for the skins folder livesearch. The title was not escaped at all which lead to markup not being rendered correctly in some cases.

Fixes https://github.com/4teamwork/opengever.core/issues/5016.


deiferni:
_(thank, you amended)_

Issue #5017, opened 2018-10-30
error while rendering search results


deiferni:
https://github.com/4teamwork/opengever.ftw/issues/77

PR #5021, opened 2018-10-31
Fix rendering issue in search view.

deiferni:
This PR fixes a rendering issue in search view. We fix it for SOLR and for the skins folder search results. The description was inserted as structure which would not render correctly due to missing escaping for catalog results.

SOLR search is configured by the view to return markup with highlighting using `<em>` tags. SOLR is now also configured to already escape its results before inserting those `<em>` tags so that the returned description can be safely inserted in the template using a `tal:structure`. 

For catalog results the view now escapes the description before it is being rendered in the template.

Fixes https://github.com/4teamwork/opengever.ftw/issues/77 https://github.com/4teamwork/opengever.core/issues/5017

### DIFF
--- a/opengever/base/browser/search.py
+++ b/opengever/base/browser/search.py
@@ -5,12 +5,12 @@ from ftw.solr.query import escape
 from ftw.solr.query import make_query
 from opengever.base.interfaces import ISearchSettings
 from opengever.base.solr import OGSolrContentListing
+from opengever.base.utils import escape_html
 from opengever.bumblebee import is_bumblebee_feature_enabled
 from plone import api
 from plone.app.search.browser import EVER
 from plone.app.search.browser import quote_chars
 from plone.app.search.browser import Search
-from plone.registry.interfaces import IRegistry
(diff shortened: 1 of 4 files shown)


opengever/base/browser/search.py
```diff
@@ -182,6 +184,25 @@ def breadcrumbs(self, item):
 
         return breadcrumbs
 
+    def description(self, item):
```
lukasgraf: I would consider naming this `escaped_description`, because that's what distinguishes it from `CroppedDescription`.

opengever/base/browser/search.py
```diff
@@ -182,6 +184,25 @@ def breadcrumbs(self, item):
...
+        """Return description and guarantee escaping of html special chars.
+
+        OGSolrContentListingObject.CroppedDescription will return markup with
+        highlihting using <em> tags and with html special chars already
```
lukasgraf: Nice try at spelling `highlighting` 😜

deiferni:
updated.

PR #5024, opened 2018-10-31
Add file related metadata and indices, without yet actually using them

### LINKED ISSUE #4966: Index filesize, filename and extension of documents
Use-cases:
- Display filesize and filename in document listings
- Query for large documents
- Use extension for mimetype icons 
- Query documents by type (extension)

Fields:
- filesize: indexed and stored
- filename: stored
- extension: indexed and stored

see also #4794
> Extension should be stored in the format .jpg, .tar.gz, etc.. (including the leading dot)

Maybe it's not necessary to store the filename because it can be generated from the title and the extension?

Metadata columns that can be dropped probably:
- `getObjSize`
- `getContentType`
- `getIcon`

Rotonen:
Did I get the addition of yet-unused metadata bit right here?

The translations on the schema items also probably need some love.

I'm also assuming we want to craft the deferrable upgrade step for rolling out this and the new `changed` timesteamp sometime later when the dust has settled on both.

Closes #4966

### DIFF
--- a/opengever/core/profiles/default/catalog.xml
+++ b/opengever/core/profiles/default/catalog.xml
@@ -30,6 +30,9 @@
   <column value="delivery_date" />
   <column value="checked_out" />
   <column value="public_trial" />
+  <column value="filesize" />
+  <column value="filename" />
+  <column value="file_extension" />
 
   <!-- DOSSIER -->
   <column value="containing_subdossier" />
--- a/opengever/core/upgrades/20181101150743_add_document_filesize__filename_and_file_extension_to_document_metadata/upgrade.py
+++ b/opengever/core/upgrades/20181101150743_add_document_filesize__filename_and_file_extension_to_document_metadata/upgrade.py
@@ -1,0 +1,21 @@
+from ftw.upgrade import UpgradeStep
+from opengever.document.behaviors import IBaseDocument
+
+
+class AddDocumentFilesize_filenameAndFileExtensionToDocumentMetadata(UpgradeStep):
+    """Add document filesize, filename and file extension to document metadata.
+    """
+
+    deferrable = True
+
+    def __call__(self):
+        self.install_upgrade_profile()
+        self.catalog_add_index('filesize', 'FieldIndex')
+        self.catalog_add_index('file_extension', 'FieldIndex')
+
+        query = {
+            'object_provides': IBaseDocument.__identifier__,
+            }
+
+        for basedocument in self.objects(query, 'Index filesize, filename and file extension of IBaseDocument objects.'):
+            basedocument.reindexObject(idxs=['filesize', 'file_extension'])
--- a/opengever/document/behaviors/metadata.py
+++ b/opengever/document/behaviors/metadata.py
@@ -7,7 +7,6 @@ from ftw.keywordwidget.widget import KeywordFieldWidget
 from ftw.mail.mail import IMail
 from opengever.document import _
 from opengever.document.interfaces import IDocumentSettings
-from opengever.document.behaviors.related_docs import IRelatedDocuments
 from plone import api
 from plone.autoform import directives as form
 from plone.autoform.interfaces import IFormFieldProvider
--- a/opengever/document/indexers.py
+++ b/opengever/document/indexers.py
@@ -21,6 +21,7 @@ from zope.component import queryMultiAdapter
 from zope.interface import implementer
 from zope.interface import Interface
 import logging
+import os.path
 
 
 logger = logging.getLogger('opengever.document')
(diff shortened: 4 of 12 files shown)


opengever/core/profiles/default/catalog.xml
```diff
@@ -30,6 +30,9 @@
   <column value="delivery_date" />
   <column value="checked_out" />
   <column value="public_trial" />
+  <column value="document_filesize" />
```
lukasgraf: I would drop the `document_` prefix. These metadata columns / indexes are by no means limited to documents (at least technically, that's how the catalog works. also not necessarily functionally, e.g. `ftw.mail`). We also don't really use namespacing for other index / metadata names.

opengever/core/upgrades/20181031135338_add_document_filename_to_document_metadata/upgrade.py
```diff
@@ -0,0 +1,9 @@
+from ftw.upgrade import UpgradeStep
+
+
+class AddDocumentFilenameToDocumentMetadata(UpgradeStep):
+    """Add document filename to document metadata.
+    """
```
lukasgraf: These upgrade steps should be
- made `deferrable`
- reindex the relevant indexes and metadata for `IBaseDocument` objects
- combined into one that covers all three indexes/metadata, so we don't have to iterate over objects three times

(upgrade step will be merged with the one for reindexing `changed` at a later point)

opengever/core/profiles/default/catalog.xml
```diff
@@ -30,6 +30,9 @@
...
+  <column value="document_filename" />
+  <column value="document_fileextension" />
```
lukasgraf: File size and extension are missing an index (only metadata is added in this PR). Add to `opengever/document/config.py` as a `FieldIndex`.

opengever/document/behaviors/metadata.py
```diff
@@ -96,6 +113,30 @@ class IDocumentMetadata(model.Schema):
         defaultFactory=document_date_default,
         )
 
+    form.omitted('document_filesize')
+    document_filesize = schema.Int(
```
lukasgraf: Uhm, that wasn't the plan 😉 We only need indexes and metadata for these. We already have all this data on the objects, no need to add them to the document schema.

opengever/document/indexers.py
```diff
@@ -115,6 +117,31 @@ def document_date(obj):
     return context.document_date
 
 
+@indexer(IDocumentSchema)
```
lukasgraf: Register these for `IBaseDocument`, and use `obj.get_file()` and `obj.get_filename()` respectively instead.

opengever/document/tests/test_indexers.py
```diff
@@ -53,6 +56,57 @@ def test_author_indexers(self):
         self.assertEquals(
             index_data_for(doc1).get('sortable_author'), u'H\xfcgo B\xf6ss')
 
+    def test_filesize_indexers(self):
+        file_ = NamedBlobFile('foobar barfoo', filename=u'test.txt')
```
lukasgraf: Use
```python
document = create(Builder("document")
                  .attach_file_containing(u"content", name=u"file.txt"))
```
instead.

opengever/core/upgrades/20181031143129_add_document_file_extension_to_document_metadata/catalog.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0"?>
```
lukasgraf: Changelog entry is missing.

Rotonen:

opengever/core/upgrades/20181101150743_add_document_filesize__filename_and_file_extension_to_document_metadata/upgrade.py
```diff
@@ -0,0 +1,22 @@
...
+class AddDocumentFilesize_filenameAndFileExtensionToDocumentMetadata(UpgradeStep):
+    """Add document filesize, filename and file extension to document metadata.
+    """
+
+    deferrable = True
+
+    def __call__(self):
+        self.install_upgrade_profile()
+        self.catalog_add_index('filesize', 'FieldIndex')
+        self.catalog_add_index('filename', 'FieldIndex')
```
lukasgraf: As described in [#4966](https://github.com/4teamwork/opengever.core/issues/4966),  we don't want the filename indexed, just `filesize` and `file_extension`.

Metadata:
- `filesize`
- `filename`
- `file_extension`

Indexes:
- `filesize`
- `file_extension`

opengever/document/indexers.py
```diff
@@ -195,3 +196,28 @@ def metadata(obj):
         metadata.append(doc_metadata.foreign_reference.encode('utf8'))
 
     return ' '.join(metadata)
+
+
+@indexer(IBaseDocument)
+def filesize(obj):
+    file = obj.get_file()
```
lukasgraf: `file` shadows the Python builtin `file`. Use `file_` instead please.

Rotonen:

PR #5041, opened 2018-11-02
Restrict owner permissions on shadow documents

### LINKED ISSUE #2168: Finalize document-state-shadow permissions
With the current POC implementation `Owner` has (too?) many Permissions on a document in shadow state. Investigate and remove unnecessary permissions for `Owner`.

Rotonen:
I left the `Delete` permission as it is conceivable users might want to clean up after themselves if the templating solution(s) fail at the time.

Closes #2168


opengever/core/upgrades/20181102131411_restrict_owner_permissions_of_owners_on_shadow_documents/upgrade.py
```diff
@@ -0,0 +1,9 @@
+from ftw.upgrade import UpgradeStep
+
+
+class RestrictOwnerPermissionsOfOwnersOnShadowDocuments(UpgradeStep):
+    """Restrict owner permissions of owners on shadow documents.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
```
phgross: You always need to update the `object_security` by using `self.update_workflow_security` when you change a workflow. 

In this case it's even better if you create a list with shadow documents by your own and use `self.update_security`. But please pass in `reindex_security=False` because your not changing the `View` permission ...

Rotonen:
@phgross ready for rereview.

PR #4806, opened 2018-08-29
Use document modification date to determine dossier end date.

njohner:
We adapt the `earliest_possible_end_date` method to use the document modification date instead of the document date in the determination of the dossier end date when resolving a dossier. We do this because the document date will not be used to reflect modifications to the documents anymore, instead it will be set by the user.

We also introduce a new marker interface `IAutomaticallyGeneratedDocument` to identify documents that were automatically generated, as opposed to user-generated. We add this as base class for `IDossierTasksPdfMarker` and `IDossierJournalPdfMarker`, identifying the two PDFs that potentially get generated when the dossier is resolved. 

Note that `earliest_possible_end_date` is also used when determining whether a dossier has a valid end date.

resolves #4555

PR #5042, opened 2018-11-02
Save PDF of a document as a separate document - OGIP 39

### LINKED ISSUE #4214: OGIP 39 - PDF speichern unter
Platzhalter Issue für [OGIP 39](https://gever.4teamwork.ch/ordnungssystem/3/2/2/5/dossier-2578/document-23463#overview)

njohner:
We add the possibility to save the PDF of a document as a new document in a different folder.

We use a form `SavePDFUnderForm` to select the destination container where the new document will be created.

screen shot 2018-11-06 at 14 34 27

Submitting the form redirects to the `SavePDFDocumentUnder` view. This view creates an empty document in the destination container, copying over the metadata from the source document. It then triggers the conversion of the source document to PDF, using the demand endpoint of `Bumblebee`.
This view also informs the user of the status of the conversion, and when the conversion is finished, allows the user to access the new document or get back to the source document.

**Waiting for the document to be converted**

waiting

**Message for a conversion error**

error

**Message after successfull conversion**

success

`Bumblebee` posts its response back to a callback view, `ReceiveDocumentPdf`, which will then set the file of the newly created document and take the document out of shadow state.

This process is either accessed through an action on the document or in the version tab if a specific version should be saved.

**Action in the document overview**

action_in_overview

**Action from the versions tab**

versions

resolves #4214


njohner:
⚠️ Once https://github.com/4teamwork/opengever.core/pull/4806/files is merged, we should make `IDocumentSavedAsPDFMarker` a subclass of `IAutomaticallyGeneratedDocument`. ⚠️

njohner:
Would it make sense to only show the `save pdf under` action if bumblebee is activated?

njohner:
Sorry my second comment was pending, seems I forgot to submit it. If you have input there, I can also adapt this

njohner:
Ok I integrated the last changes:
* only show action for documents that are convertable
* if queuing the demand to the bumblebee service fails, we raise an exception
* In the version tab, the save as PDB link is available if the bumblebee feature is turned on and the document is bumblebeeable. But we don't check for each version whether it is convertible. This is now done in the form `SavePDFUnderForm`. If the version is not convertable, we redirect to the source document and display an error message

**Action available for a convertable document**
screen shot 2018-11-08 at 13 40 06

**Action not available for a unconvertable document**
screen shot 2018-11-08 at 13 40 00

**Error displayed when trying to convert an unconvertable version**
screen shot 2018-11-08 at 16 21 51

PR #5044, opened 2018-11-02
Include physical path to item and container serialization

### LINKED ISSUE #4979: Feature: Path in response of REST API
When using the REST API in a frontend application (vue.js for example) there is no good way to store a unique identifier of a resource from GEVER.

In the RIS application, a business has a GEVER dossier. We persist the identifier of the dossier which is currently the URL. The problem is, that the URL also contains the cluster URL `https://ris.onegovgever.ch/` for example and in some circumstances the admin UID which could be something such as `/gs`. In general, it's a bad idea to persist URLs because they may change. So a better value to an identifier could be the path of the URL `/ordnungssystem/1/dossier-9`. This way the URL to the GEVER may change but the locator to the resource, the dossier, in that case, stays the same.
So the response of a GEVER resource (dossier, document, task, ...) should contain the path as well.

elioschmutz:
This PR provides the physical_path for serialized contents.

fixes #4979

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -11,6 +11,7 @@ Changelog
 - Make the changed field a python datetime. [njohner]
 - Fix Officeconnector permission checks for task documents tab actions. [Rotonen]
 - Ensure clamped down enough permissions on shadow documents. [Rotonen]
+- Include relative_path to dexterity item and folder serialization [elioschmutz]
 - Include bumblebee_checksum to dexterity item serialization [elioschmutz]
 - Include bumblebee-checksum into @listing endpoint [elioschmutz]
 - Drop verbose logging for ObjectTouched events. [lgraf]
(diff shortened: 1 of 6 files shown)


deiferni:
i took the liberty of changing the merge base, so this is reviewable. please switch that back to master once it is merged.

deiferni:
maybe @bierik or @mbaechtold could tell us how the relative path would be used in `RIS`?

deiferni:
we've discussed it and it should be implemented as specified in phippus comment at: https://github.com/4teamwork/opengever.core/issues/4979#issuecomment-434293819, i.e. with a path **relative** to the **plone site**.

elioschmutz:
@deiferni 👍 updated.

I implemented the get_relative_path method twice. Once for the serializer handling objects and once for the listing-objects. I could refactor the `extend_with_relative_path` and exclude the relative_path lookup to its own method. Then I could use it for the @listing, too. Do you prefer this way?

PR #5047, opened 2018-11-05
Add all docproperties to Oneoffixx XML payloads

### LINKED ISSUE #5025: Extend OneOffixx properties in the connect.xml
The oneoffixx connect.xml should be extended with all the following properties (see https://docs.onegovgever.ch/admin-manual/docproperties/list/#doc-properties-aktueller-benutzer):
 - ogg.user.*
 - ogg.dossier.*
 - ogg.document.*

_Note: ogg.meeting.* and ogg.recipient.* are not necessary_

Rotonen:
* Add an xml type and encoding declaration
* Encode in UTF-8
* Use all of the user, dossier and document doc properties

Closes #5025

### DIFF
--- a/opengever/oneoffixx/browser/connect_xml.py
+++ b/opengever/oneoffixx/browser/connect_xml.py
@@ -1,12 +1,13 @@
 from collections import OrderedDict
 from lxml import etree
-from opengever.base.interfaces import IReferenceNumber
+from opengever.dossier.interfaces import IDocProperties
 from opengever.officeconnector.helpers import create_oc_url
-from opengever.dossier.behaviors.dossier import IDossier
 from plone import api
+from Products.CMFPlone.utils import safe_unicode
 from Products.Five import BrowserView
 from zExceptions import NotFound
 from zope.annotation.interfaces import IAnnotations
+from zope.component import getMultiAdapter
 
 
 class OneoffixxConnectXml(BrowserView):
@@ -36,6 +37,8 @@ class OneoffixxConnectXml(BrowserView):
             )),
         ))
 
+        self.document_properties = getMultiAdapter((self.context, self.request), IDocProperties).get_properties()
+
         self.request.RESPONSE.setHeader("Content-type", "application/xml")
         return self.generate_xml()
 
@@ -52,7 +55,7 @@ class OneoffixxConnectXml(BrowserView):
 
         entries = etree.SubElement(batch, "Entries")
         entries.append(self.generate_one_offixx_connect_tag())
-        return etree.tostring(batch, pretty_print=True)
+        return etree.tostring(batch, pretty_print=True, xml_declaration=True, encoding="UTF-8")
 
     def generate_settings_tag(self):
         settings_tag = etree.Element("Settings")
@@ -119,22 +122,16 @@ class OneoffixxConnectXml(BrowserView):
         interface = etree.SubElement(arguments, "Interface")
         interface.set("Name", "OneGovGEVER")
 
-        node = etree.SubElement(interface, "Node")
-        node.set("Id", "ogg.document.title")
-        node.text = self.context.Title().decode("utf-8")
+        for key, value in self.document_properties.items():
+            node = etree.SubElement(interface, "Node")
+            node.set("Id", key)
+            if key in ('ogg.document.document_date',):
+                value = api.portal.get_localized_time(value)
+            if isinstance(value, basestring):
+                node.text = safe_unicode(value)
+            else:
+                node.text = unicode(value)
 
-        reference_number = IReferenceNumber(self.context)
-        node = etree.SubElement(interface, "Node")
-        node.set("Id", "ogg.document.reference_number")
-        node.text = reference_number.get_number()
-
-        node = etree.SubElement(interface, "Node")
-        node.set("Id", "ogg.document.sequence_number")
-        node.text = reference_number.get_local_number()
-
-        node = etree.SubElement(interface, "Node")
-        node.set("Id", "ogg.dossier.external_reference")
-        node.text = IDossier(self.context.getParentNode()).external_reference
         return function
 
     def generate_metadata_tag(self):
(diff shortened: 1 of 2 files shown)


opengever/oneoffixx/browser/connect_xml.py
```diff
@@ -1,8 +1,9 @@
 from collections import OrderedDict
 from lxml import etree
-from opengever.base.interfaces import IReferenceNumber
+from opengever.dossier.docprops import DefaultDocumentDocPropertyProvider as document_properties
```
deiferni: why would you rename the class, this is extra confusing.

deiferni: why would you rename the class to a method, this is extra confusing 🤔.

opengever/oneoffixx/browser/connect_xml.py
```diff
@@ -36,6 +37,10 @@ def __call__(self):
             )),
         ))
 
+        self.document_properties = user_properties(api.user.get_current()).get_properties()
```
deiferni: you are bypassing the adapter lookup here. you should multi-adapt (user/ dossier / document, request) to `IDocProperties` and then get the properties from the adapter.

that makes my first comment irrelevant as the imports will go away in any case 😉.

Rotonen:
@deiferni ready for rereview.

opengever/oneoffixx/browser/connect_xml.py
```diff
@@ -119,47 +121,26 @@ def generate_custom_interface_connector_tag(self):
...
-        reference_number = IReferenceNumber(self.context)
-        node = etree.SubElement(interface, "Node")
-        node.set("Id", "ogg.document.reference_number")
-        node.text = reference_number.get_number()
+        for key, value in self.document_properties.items():
+            node = etree.SubElement(interface, "Node")
+            node.set("Id", key)
+            if key in ('ogg.document.document_date',):
+                value = api.portal.get_localized_time(value)
+            node.text = unicode(value)
```
deiferni: should you escape here or will lxml take care of that 🤔?

deiferni: you should use `safe_unicode` here instead of casting to `unicode` to avoid `UnicodeDecodeError` in case of utf-8 encoded input values.

opengever/oneoffixx/browser/connect_xml.py
```diff
@@ -119,47 +121,26 @@ def generate_custom_interface_connector_tag(self):
...
-        node.set("key", "ogg.dossier.external_reference")
-        node.set("type", "string")
-        node.text = IDossier(self.context.getParentNode()).external_reference
+        for key, value in self.document_properties.items():
+            node = etree.SubElement(function, "Value")
+            node.set("key", key)
+            node.set("type", "string")
+            if key in ('ogg.document.document_date',):
+                value = api.portal.get_localized_time(value)
+            node.text = unicode(value)
```
deiferni: same here

opengever/oneoffixx/tests/assets/oneoffixx_connect_xml.txt
```diff
@@ -1,3 +1,4 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
deiferni: this is a `*.txt` file, shouldn't this be an `*.xml` file?

deiferni: just so i have asked this ... so oneoffixx will handle the encoding correctly?

Rotonen: @njohner why'd this originally become a .txt file - to skip the xml style validation?

opengever/oneoffixx/browser/connect_xml.py
Rotonen: If we escape, Office apps will not unescape?

opengever/oneoffixx/tests/assets/oneoffixx_connect_xml.txt
Rotonen: I cannot fully test that as there are no templates which actually use our docproperties, but at least it consumes the files errorless and produces valid output files.

opengever/oneoffixx/browser/connect_xml.py
deiferni: also what about input that will break your xml, i.e. what happens if you input `">"`?

opengever/oneoffixx/tests/assets/oneoffixx_connect_xml.txt
deiferni: that has to do it then 👍

opengever/oneoffixx/browser/connect_xml.py
Rotonen: How do I handle ints then?

njohner:
Is it worth adding the deprecated docproperties? They will be removed in the next release and were never in oneoffixx

opengever/oneoffixx/browser/connect_xml.py
Rotonen: Yes, verified manually and from the test output, it does:

```
-  '            <Node Id="ogg.document.sequence_number">33</Node>',
+  '            <Node Id="ogg.document.sequence_number">&lt;33</Node>',
```

We're setting `node.text` there which gets appropriately handled.

Rotonen:
Ready for rereview.

opengever/oneoffixx/browser/connect_xml.py
deiferni: 👍

deiferni: separately 😄? i can only tell you that this won't work if a string is in utf-8. and i think at the moment we cannot guarantee that all our strings are unicode.

PR #5048, opened 2018-11-05
Make "changed" a python datetime

### LINKED ISSUE #5038: Match changed values with the field type
Split off from https://github.com/4teamwork/opengever.core/pull/4996

Currently the field `._type` is `datetime.datetime` and the values stored are `DateTime.DateTime`.

njohner:
Once `ftw.testing` is released, we will obviously adapt the second commit and squash it.

In this PR we make `changed` a python `datetime` instead of a Zope `DateTime`. I also touched the upgrade step which fills the `changed` metadata column. If the upgrade step has been executed already or if the first upgrade step (adding the `changed` index) had been executed and the instance used without executing the second upgrade, some (or all) objects will have a Zope `DateTime` in the `changed` field, otherwise all `changed` fields will be empty. So in this upgrade we simply check whether the field is set, if so cast it to `datetime`, otherwise cast `modified` to `datetime` and use that to set the field. In this way we cover all possibilities.

I tested the upgrade for the two cases:

- Setup a new instance from current master, with the `changed` feature already in it, i.e. all objects have a `changed` date as zope `DateTime`. Then checkout this branch `nj_changed`. Upgrade properly sets all the `changed` dates to `datetime`.

- Setup an new instance from an older version of gever (just prior to adding `changed` field). Then switch to master and install only the first upgrade (adding the `changed` index). Touch some objects, then checkout `nj_modified`. So some objects have a zope `DateTime`, some have `changed` that is not set and points to `modified()`. We then run the second upgrade and check that `changed` is a python `datetime` for both types of objects.
 
resolves #5038


njohner:
I also tested it locally: setup an instance with from an older version of gever, before introduction of changed, then create several versions of a document and afterwards upgrade to this branch. `changed` gets set on all objects. When reverting to an older version, the document still has its `changed` field, as only the file gets overwritten

deiferni:
🚢👌

PR #5050, opened 2018-11-05
Bump ftw.contentstats to get disk usage logging.

lukasgraf:
Bump `ftw.contentstats` to `1.1.0` to get disk usage logging (4teamwork/ftw.contentstats#16).

I also added a bullet point to the Release 2018.5 Checklist (#4858) to verify the disk usage logging.

In order for the tests in `opengever.core` that veryify the `ftw.contentstats` integration to pass, I had to bring in the [`ContentStatsLayer`](https://github.com/4teamwork/ftw.contentstats/blob/71d615e8b2f038a4ad1030ce96e8be16b2d3d974/ftw/contentstats/testing.py#L51-L137) as a base for a new layer. This is because otherwise the [`get_log_dir_path()`](
https://github.com/4teamwork/ftw.contentstats/blob/master/ftw/contentstats/logger.py#L36-L57) that's being used in several places can't find a logfile logation (because usually in tests no eventlog handler is registered). Since this is already solved with setting up an eventlog configuration during tests in `ftw.contentstats`, I'd rather bring that layer into `opengever.core` than resort to any hacks.


lukasgraf:
@phgross this one is ready now.

PR #5057, opened 2018-11-07
Fix issue when returning an excerpt to an already decided proposal.

### LINKED ISSUE #5055: Return excerpt of closed meeting proposals
since https://github.com/4teamwork/opengever.core/pull/4837, we do not set a `proposal` to decided when an `agendaitem` is decided, but only when the protocol excerpt has been returned. 

For meetings where agendaitems have already been decided before we update to 2018.5, but the protocol has not been returned, we will try to close the `proposal` when returning the excerpt, which should raise an error.

njohner:
This will not happen in the long run, but agendaitems that were decided before we introduced that proposals get decided only when the excerpt is returned can be already decided even if the proposal has not been returned. This lead to an error when returning the excerpt and trying to decided an already decided proposal.
We fix this by only closing the proposal if it has not been closed already.

**returning the excerpt of a closed agendaitem could lead to an error**
screen shot 2018-11-07 at 10 28 06

**With the fix, we can return the excerpt without trying to close the proposal a second time**
screen shot 2018-11-07 at 10 30 01

resolves #5055

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2018.5.0 (unreleased)
 ---------------------
 
+- Fix issue when returning an excerpt to an already decided proposal. [njohner]
 - Display group label in teamdetails instead of group title. [njohner]
 - Add reference numbers to protected items admin view links. [Rotonen]
 - Sort repository excel exports by reference number. [Rotonen]
--- a/opengever/meeting/model/agendaitem.py
+++ b/opengever/meeting/model/agendaitem.py
@@ -353,7 +353,11 @@ class AgendaItem(Base):
         return False
 
     def return_excerpt(self, document):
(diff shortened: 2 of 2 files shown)


njohner:
Before merging this, let me point out the following: for meetings where certain `agendaitem`s have been decided before the update and others after the update to 2018.5, returning the excerpt before closing the meeting will be enforced only for the `agendaitem`s closed after the upgrade. For the others, the `proposal` will already be set to `decided`...

I've also tested the opposite situation, i.e. reopening an `agendaitem` that had been closed before the update but the excerpt not returned. That does not pose any problems as the `proposal` is not reopened anyway when an agendaitem is reopened

njohner:
Also not that I've made a follow up issue, to make the decided proposals uniformly have an excerpt.
https://github.com/4teamwork/opengever.core/issues/5076

PR #5058, opened 2018-11-07
Display group label in teamdetails instead of group title.

### LINKED ISSUE #5054: Team details: Group not displayed if it has no title
In the **team details view** the group name isn't displayed if the group doesn't have a title. In the team listing it works correctly however:

team_listing

---

team_details

---

The template should probably use `group/label` [instead of `group/title`](https://github.com/4teamwork/opengever.core/blob/bc1171801ae46ec70e6900a4af0ca6819ff7d592/opengever/ogds/base/browser/templates/teamdetails.pt#L28), just [like the `TeamListing` tab does](https://github.com/4teamwork/opengever.core/blob/bc1171801ae46ec70e6900a4af0ca6819ff7d592/opengever/contact/browser/tabs.py#L64). This also should be covered by a regression test.

njohner:
Using `label` we will display the `groupid` if the group does not have a `title`, instead of having an empty field.

resolves #5054

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2018.5.0 (unreleased)
 ---------------------
 
+- Display group label in teamdetails instead of group title. [njohner]
 - Add reference numbers to protected items admin view links. [Rotonen]
 - Sort repository excel exports by reference number. [Rotonen]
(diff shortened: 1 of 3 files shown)


njohner:
Ship it in 2018.5 or not?

njohner:
Oh sorry, ticket is set to 2018.5. So I'll ship it!

PR #5072, opened 2018-11-09
Exclude shadow documents from being relatable

### LINKED ISSUE #3385: Make sure shadow state documents are not selectable for relations
We offer relations to documents in various places:
- related items of a document
- related items of a task
- attachments of a proposal
- …?

Documents in shadow-state should not be selectable in those lists since they are only visible for `Owner` and may be deleted if creation is cancelled client-side.

Rotonen:
As we use a new enough `ZCatalog`, we can append a not query of the `review_state` to the path queries.

Put this PR to 2018.5 milestone as this is essentially an Oneoffixx thing.

Closes #3385

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -7,6 +7,7 @@ Changelog
 
 - Fix issue when returning an excerpt to an already decided proposal. [njohner]
 - Display group label in teamdetails instead of group title. [njohner]
+- Prevent shadow documents from being relatable. [Rotonen]
 - Add reference numbers to protected items admin view links. [Rotonen]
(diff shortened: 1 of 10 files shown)


Rotonen:
Indeed found a few more. Listing the found exceptions below.

Excerpts do not have a form so those do not have to be addressed.

The additional document(s) submit forms I could not make pick up on (shadow) documents at all in the tests, so I did not find a need to address those. In real use they're also autopopulated by wizards as far as I can tell.

Sending documents via mail is convoluted enough to not bother with and shadow documents will not have a file to begin with.

Manual journal entries should be able to refer to shadow documents

Ready for rereview.

PR #5079, opened 2018-11-13
Performance improvements on tasks local roles setter

phgross:
During the performance and long-request analysis, we detected some problems in the role assignment during task creation or some of the task transitions.
When creating a task, the responsible user and the inbox_group of the assigned_org_unit gets the `Contributor` role assigned on the distinct parent (the dossier or subdossier). This role assignment triggers a `reindexObjectSecurity` for the dossier, which is actually correct, but is also recursive and triggers therefore a reindex for every document and mail inside this dossiers. 

Because we know that the `Contributor` role has no impact on documents and mails, we are able to shorten this one and only reindex the Dossier and the containing subdossiers. The fact that the `Contributor` role has no impact on documents, mails, proposals and task, is always true, because otherwise the sharing of a specific document with the help of a task wouldn't work as expected, and this is a key use case of the task module.

In combination with the reindexObjectSecurity patch in ftw.solr this adjustment leads to an large performance improvement, when creating a task on a toplevel of a huge dossier:

local measures (on a fresh started instance):  
```
With adjustment: 1883 ms
Without adjustment: 21465 ms
```

This need to be backported to `2018.4-stable`.


opengever/task/tests/test_localroles.py
```diff
@@ -311,3 +313,94 @@ def test_can_delete_related_document_from_task(self):
         self.login(self.manager)
         api.content.delete(self.document)
         self.assertEqual([], self.task.relatedItems)
+
+
+class TestLocalRolesReindexing(IntegrationTestCase):
+
+    def test_reindexes_distinct_parent_and_related_documents(self):
+        self.login(self.dossier_responsible)
+        task = create(Builder('task')
```
deiferni: could we test/document that the user is not in the `allowedRolesAndUsers` index here, before creating the task.

opengever/task/tests/test_localroles.py
```diff
@@ -311,3 +313,94 @@ def test_can_delete_related_document_from_task(self):
...
+                      index_data_for(self.dossier)['allowedRolesAndUsers'])
+        self.assertIn(principal,
+                      index_data_for(self.document)['allowedRolesAndUsers'])
+        self.assertIn(principal,
+                      index_data_for(task)['allowedRolesAndUsers'])
+
+    def test_also_reindexes_containing_subdossier(self):
+        """The dossier are reindex manually
+        """
+        self.login(self.dossier_responsible)
```
deiferni: same here, please document the precondition.

phgross:
@deiferni updated ...

Issue #5070, opened 2018-11-09
OGIP 39: Bumblebee conversion not triggered correctly


phgross:
After saving a document as pdf, the bumblebee preview, does not work. Only if I trigger the conversion again via `Vorschau neu generieren` the preview shows up.

Example: https://dev.onegovgever.ch/fd/ordnungssystem/fuehrung/gemeinderecht/uebergeordnete-erlasse/dossier-448/document-2066#overview

deiferni:
could be related to https://github.com/4teamwork/opengever.core/issues/4999

njohner:
This has been resolved in https://github.com/4teamwork/opengever.core/pull/5083

PR #5083, opened 2018-11-13
Login as document owner in the ReceiveDocumentPDF callback.

njohner:
The ReceiveDocumentPDF callback view is called from bumblebee hence by an anonymous user. Because of that the preview is not generated by bumblebee as the anonymous user does not have view permissions on the document.


njohner:
This resolves #5070

PR #5085, opened 2018-11-13
Add Open XML Visio mimetypes.

### LINKED ISSUE #5080: Missing mimetype for visio documents
Similar we did for other mimetypes (see #4866) we need to extend the mimetype library for `vsdx` documents and add an upgradestep which fix existing ones.

deiferni:
Add Open XML Visio to extra mimetypes and the mimetype registry. Also add a deferrable upgrade to fix the `getContentType` and `getIcon` metadata so that icons are displayed correctly in GEVER.

The  upgrade step to update the mimetype registry is in registered in opengever.core and not in opengever.policy.base to prevent ordering issues as the mimetype registry *must* be upgraded before objects reindexed. Otherwise the `getIcon` metadata won't be filled correctly and thus the icons won't show up as expected.

A list of possible file extension has been found at microsoft (https://docs.microsoft.com/en-us/office/client-developer/visio/introduction-to-the-visio-file-formatvsdx) and the corresponding mimetype names has been taken from apache tika (http://svn.apache.org/viewvc/tika/trunk/tika-core/src/main/resources/org/apache/tika/mime/tika-mimetypes.xml?revision=1723581&view=markup#l2455).

After the fix:
screenshot 2018-11-13 at 17 06 22

And before:
screenshot 2018-11-13 at 17 03 29

Fixes #5080.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2018.5.0 (unreleased)
 ---------------------
 
+- Add Open XML Visio mimetypes. [deiferni]
 - Fix issue when returning an excerpt to an already decided proposal. [njohner]
 - Fix data in Task PDF when resolving a dossier. [njohner]
 - Make save as pdf button colored green. [deiferni]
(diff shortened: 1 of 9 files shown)


deiferni:
(as discussed, figure out if we can include the template and stencil or if we only include the drawings)

deiferni:
as discussed with @buchi and @Rotonen we only include:
- application/vnd.ms-visio.drawing.macroEnabled.12
- application/vnd.ms-visio.drawing

for office connector. templates and stencils should not be editable for now.

deiferni:
@Rotonen ready for re-review.

PR #5099, opened 2018-11-15
Improve task listing formatting.

njohner:
We improve the format of the task listing PDF:

* We add the possibility to set the text alignment in the `LaTexListing`. `Column`s now have a new `alignment` attribute.
* To avoid stretching of text and hyphenation, we set the first three columns (date, creator and transition) to left-aligned. The last column, `description`, is justified.
* For the creator we now show the full label (full name and userid)
* In the title of the document, we quote the dossier title.

**Old formatting**
screen shot 2018-11-15 at 08 26 05

**New formatting**
screen shot 2018-11-15 at 08 28 49

This should not change the formatting of other listings, as we use `justified` as default, which seems to be the default when no alignment is specified. The generated latex probably depends on the browser used though, as we go through HTML, and then from HTML to Latex.

### DIFF
--- a/opengever/latex/dossiertasks.py
+++ b/opengever/latex/dossiertasks.py
@@ -58,7 +58,7 @@ class DossierTasksLaTeXView(MakoLaTeXView):
         tasks = self.get_tasks()
         task_data_list = []
         title = translate(_('label_dossier_tasks',
-                            default=u'Task list for dossier ${title} (${reference_number})',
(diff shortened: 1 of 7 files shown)


njohner:
I'm not too sure whether using the `align` attribute is well supported:
https://www.w3schools.com/tags/att_td_align.asp

I should maybe use `text-align` instead, seems to be supported by all browser.

PR #5100, opened 2018-11-15
More refactorings of opengever.ogds.models

### LINKED ISSUE #4990: Finish integration of opengever.ogds.models.
In https://github.com/4teamwork/opengever.core/pull/4690 we have copied `opengever.ogds.models` into this package. Some cleanup still remains:

Split out everything that is not strictly `ogds` related
- `__init__.py` there is a `BASE` which should be merged into `opengever.base.model.Base`
- `query.py` move query to a base module
- `testing.py` there is layer setup in there, it should be together with the other layers if still needed
- `types.py` move to base module
- marry `PloneOGDSService` and `OGDSService`
- ~~get rid of `OGDSService` and `ogds_service` respectively. it is not a service nor will it be in the near future.~~

Rotonen:
* The test layer merge was already done separately
* Merging the bases made sense
* Don't fully see the point, but moved query and types
* Not touching dismantling the service as that's used too widely and not seeing a point in tearing that down / apart

Closes #4990


deiferni:
@Rotonen as discussed i picked this up and amended some more refactorings. X-review welcome.

fail.txt
```diff
@@ -0,0 +1,4 @@
+excerpts relation list
```
Rotonen: We should drop this accidental git add.

Rotonen:
Did not spot anything glaring.

deiferni:
for the record:
- the motivation to move query and types is that they define base types/query classes that are used throughout the application, not only in the `ogds` package. having such base classes in a central place seems a good thing to me as it reduces dependencies between packages.
- i agree pulling out `OGDSService` does not make sense at the moment. It does not cause issues IIRC.

deiferni:
i will merge this one today if nobody objects ⏳

PR #5101, opened 2018-11-15
Show participants in meeting tab-participants.

### LINKED ISSUE #4876: Participants of a meeting not updated when Memberships change
Currently when a meeting is created its participants are initialized https://github.com/4teamwork/opengever.core/blob/8746053de47e33eb297f4cb97d684ef2c05c5f02/opengever/meeting/model/meeting.py#L186-L192 and stored persistently in `Meeting.participants`. When memberships of those participants are changed at a later point the participants of a meeting are not updated correctly. 

Steps to reproduce:
1) Create a Meeting in a Committee that has Memberships that overlap the Meetings date
2) Modify Memberships of the Committee so that there are fewer Memberships at the meeting's date (e.g. remove memberships or shorten membership period)
3) Generate a protocol for the meeting, the protocol will still list the Members of the Memberships removed in step 2)

njohner:
Make sure all participants are displayed in the tab-participants of a meeting, even if their membership has expired, so that they can be removed from the participants list.

* We highlight participants that are in a meeting but not members (if their membership has been modified for example), with a red dot (green for present, grey for absent).
* If such a participant is removed from a meeting, it will be hidden immediately in the browser, this is important as such a participant cannot be added back, so clicking to add it back would fail.

**meeting with two participants**
screen shot 2018-11-15 at 10 20 28

**after changing the membership of Meier Hans, he is not a member at the time of the meeting anymore. It is shown with a red dot**
screen shot 2018-11-15 at 10 21 28

**after clicking the user to remove him from the meeting, it gets hidden**
screen shot 2018-11-15 at 10 21 36

resolves #4876


njohner:
We should probably amend the user documentation:
https://docs.onegovgever.ch/user-manual/spv/sitzung-durchfuehren/

opengever/meeting/browser/meetings/templates/meeting.pt
```diff
@@ -359,41 +359,42 @@
...
-                      <option value=""></option>
-                      <option value="presidency" i18n:translate="meeting_role_presidency">Presidency</option>
-                    </select>
-                  </div>
+                <span tal:repeat="participant participants">
+                  <li tal:define="cls python:meeting.is_editable() and 'editable ' + participant['member_cssclass'];
+                                  cls string:participant folded ${cls}"
+                      tal:attributes="data-member-id participant/member_id;
+                                      class cls">
+                    <div tal:attributes="class python:participant['presence_cssclass'] + ' ' + participant['member_cssclass']" />
```
deiferni: if those are used together it may be better to concatenate them in the view

njohner:
Ok this is ready. I included all of your comments. I also tried to use the new `absentees` property from the `Meeting` model, but that is really not useful here, as we anyway need the list of all `members` of the meeting to check whether a participant has an expired membership or not.

PR #5113, opened 2018-11-16
Fix the task reminder Vue app to also work without the reminder element in the markup

### LINKED ISSUE #5109: Task overview is broken for resolved tasks
Since #5098 the task overview is broken, because of an JS error. 

https://dev.onegovgever.ch/fd/ordnungssystem/gesellschaftliches/sport/projekte/dossier-37/task-16

Rotonen:
Reorganised around
* Only fires after we know we have a tabbedview
* Only fires after we know the reminder selector is there

Closes #5109


Rotonen:
@phgross as there need to be things visible to the vue app at its init time and one should only declare functions at the top level, I could not find a way to juggle that. I would also prefer that.

PR #5116, opened 2018-11-16
Translate z3c.formwidget.query (nothing).

### LINKED ISSUE #5081: translate z3c.formwidget.query (nothing)
The label for `(nothing)` is not translated in german.
screenshot 2018-11-13 at 14 17 43

We should provide a german and french manual translation for `z3c.formwidget.query`.

njohner:
This translation appears in proposal add forms and committee add forms (at least, maybe I missed other places where it appears).

**It is now correctly translated in the committee add form**
screen shot 2018-11-16 at 17 04 19

**and in the proposal add form**
screen shot 2018-11-16 at 17 04 00

Maybe the '(nothing)' checkbox is useful in the proposal add form, once a document has been selected, this allows to deselect it (see above)? But in that submitting the form would still raise an error though...

resolves #5081


njohner:
ready

PR #5121, opened 2018-11-19
Respect private_task feature flag also in forwarding forms.

### LINKED ISSUE #5120: Private task feature flag is not respected in the forwarding form.
The `private_task` field is shown up even when the feature flag is disabled.

phgross:
closes #5120

### DIFF
--- a/opengever/inbox/forwarding.py
+++ b/opengever/inbox/forwarding.py
@@ -7,6 +7,7 @@ from opengever.ogds.base.sources import AllUsersInboxesAndTeamsSourceBinder
 from opengever.ogds.base.utils import get_current_org_unit
 from opengever.ogds.base.utils import get_ou_selector
 from opengever.task import _ as task_mf
+from opengever.task import is_private_task_feature_enabled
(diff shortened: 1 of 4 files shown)


opengever/inbox/forwarding.py
```diff
@@ -166,6 +167,12 @@ def update(self):
 
         super(ForwardingAddForm, self).update()
 
+        # Disable is_private field according to the feature flag
+        if not is_private_task_feature_enabled():
+            common_group = [group for group in self.groups
```
Rotonen: Not too relevant here, but pattern-wise use a generator expression and next if you just want the first hit?

`next(group for group in self.groups if group.__name__ == u'common')`

Next can also take a default parametre in case of a miss.

```
>>> next(n for n in (1, 2, 3) if n == 2)
2
```
```
>>> next((n for n in (1, 2, 3) if n == 4), 5)
5
```

phgross: 👍 good to know, updated ...

PR #5124, opened 2018-11-20
Set \emergencystretch to 3em in latex templates.

### LINKED ISSUE #5123: Aufgaben PDF: Titel geht über rand hinaus
Siehe angehängtes Bsp. 

[Aufgabenliste des Dossiers 2008 - Renovation Regierungsgebaeude - 934850938, 13.11.2018 17 03.pdf](https://github.com/4teamwork/opengever.core/files/2598617/Aufgabenliste.des.Dossiers.2008.-.Renovation.Regierungsgebaeude.-.934850938.13.11.2018.17.03.pdf)

njohner:
This change allows latex to distribute 3em more of white space on a line when it cannot split it in a satisfying way, this is notably useful as titles often contain long numbers.

According to my tests the current setting should always work for numbers up to 10 digits. For longer numbers we can lines that extend over the margin or sheet. I feel like that is pretty reasonable, though we could obviously crank that up to some larger value, but then there is the risk of obtaining lines typeset with a lot of space.

Another possibility would be to redefine the titles to be `raggedright`, in the case of a section title as in the issue (#5123 ):
```
\titleformat*{\section}{\bf\Large\filright}
```
I think the generic solution should do though.

resolves #5123

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -7,6 +7,7 @@ Changelog
 
 - Fix bug in Windows8/10 and IE11/Edge: Select a task-reminder option will now properly stop spinning. [elioschmutz]
 - Show the task-reminder selector spinner directly and not only on long-requests to remove complexity. [elioschmutz]
+- Set \emergencystretch to 3em in latex templates. [njohner]
 - Update schemas in API documentation to include "changed". [njohner]
 - Ignore unauthorized traversal request in the response forms. [phgross]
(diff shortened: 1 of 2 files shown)


njohner:
Here a screen shot from a title that overshoots into the margin
screen shot 2018-11-20 at 10 37 42

With the `\emergencystretch` setting, the line is stretched and the number falls on the next line
screen shot 2018-11-20 at 10 37 58

PR #5128, opened 2018-11-21
Correctly handle containing_dossier, containing_subdossier and is_subdossier indices.

njohner:
The `containing_dossier` and `containing_subdossier` indices were not updated when an object was copied or moved. As an `ObjectModifiedEvent` is fired for every object that is moved, i.e. when moving a container an event is fired for every contained object as well, we can simply handle this in an event registered for `IDexterityContent`.

We also move the reindexing of `is_subdossier` to this handler, to have everything at the same place.

The `containing_subdossier` index was updated when an object was modified, event when it's title was not changed, although `containing_dossier` was only updated when the title gets modified. We now update the indexes only when necessary.

This should fix the sorting issues on subdossiers, which seemed to stem from inconsistent data between the metadata and the index:
https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/332

### DIFF
--- a/opengever/base/handlers.py
+++ b/opengever/base/handlers.py
@@ -8,6 +8,10 @@ from opengever.base.model.favorite import Favorite
 from opengever.base.oguid import Oguid
 from opengever.base.touched import ObjectTouchedEvent
 from opengever.base.touched import should_track_touches
+from opengever.dossier.behaviors.dossier import IDossierMarker
+from opengever.dossier.indexers import TYPES_WITH_CONTAINING_SUBDOSSIER_INDEX
+from opengever.globalindex.handlers.task import sync_task
+from opengever.task.task import ITask
 from plone import api
 from plone.app.workflow.interfaces import ILocalrolesModifiedEvent
 from Products.CMFCore.CMFCatalogAware import CatalogAware
@@ -59,6 +63,30 @@ def object_moved_or_added(context, event):
 
         processor = getUtility(IIndexQueueProcessor, name='ftw.solr')
         processor.index(context, CatalogAware._cmf_security_indexes)
+
+    # There are several indices that need updating when a dossier is moved.
+    # first make sure obj was actually moved and not created
+    if not event.oldParent or not event.newParent:
+        return
+
+    # When an object is moved, its containing_dossier needs reindexing.
+    to_reindex = ['containing_dossier']
+    # containing_subdossier is really only used for documents,
+    # while is_subdossier is only meaningful for dossiers.
+    if IDossierMarker.providedBy(context):
+        was_subdossier = IDossierMarker.providedBy(event.oldParent)
+        is_subdossier = IDossierMarker.providedBy(event.newParent)
+        if was_subdossier != is_subdossier:
+            to_reindex.append('is_subdossier')
+
+    if context.portal_type in TYPES_WITH_CONTAINING_SUBDOSSIER_INDEX:
+        to_reindex.append('containing_subdossier')
+
+    context.reindexObject(idxs=to_reindex)
+
+    # synchronize with model if necessary
+    if ITask.providedBy(context):
+        sync_task(context, event)
 
 
 def remove_favorites(context, event):
--- a/opengever/dossier/handlers.py
+++ b/opengever/dossier/handlers.py
@@ -131,12 +131,6 @@ def reindex_containing_dossier(dossier, event):
                     sync_task(brain.getObject(), event)
 
 
-def reindex_is_subdossier(dossier, event):
-    """Reindex the is_subdossier index after the dossier have been moved.
-    """
-    dossier.reindexObject(idxs=['is_subdossier'])
-
-
 def reindex_blocked_local_roles(dossier, event):
     """Reindex blocked_local_roles upon the acquisition blockedness changing."""
     dossier.reindexObject(idxs=['blocked_local_roles'])
--- a/opengever/dossier/handlers.py
+++ b/opengever/dossier/handlers.py
@@ -4,11 +4,12 @@ from opengever.base.interfaces import IReferenceNumber
 from opengever.base.interfaces import IReferenceNumberPrefix
 from opengever.bundle.sections.constructor import IDontIssueDossierReferenceNumber
 from opengever.dossier.behaviors.dossier import IDossier
-from opengever.dossier.behaviors.dossier import IDossierMarker
+from opengever.dossier.indexers import TYPES_WITH_CONTAINING_SUBDOSSIER_INDEX
 from opengever.dossier.resolve import get_resolver
 from opengever.globalindex.handlers.task import sync_task
 from opengever.globalindex.handlers.task import TaskSqlSyncer
 from opengever.meeting.handlers import ProposalSqlSyncer
+from opengever.task.task import ITask
 from plone import api
 from plone.app.workflow.interfaces import ILocalrolesModifiedEvent
 from zope.component import getAdapter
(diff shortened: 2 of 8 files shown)


opengever/base/handlers.py
```diff
@@ -60,6 +62,32 @@ def object_moved_or_added(context, event):
...
+    to_reindex = ['containing_dossier']
+    # containing_subdossier is really only used for documents,
+    # while is_subdossier is only meaningful for dossiers.
+    if IDossierMarker.providedBy(context):
+        was_subdossier = IDossierMarker.providedBy(event.oldParent)
+        is_subdossier = IDossierMarker.providedBy(event.newParent)
+        if was_subdossier != is_subdossier:
+            to_reindex.append('is_subdossier')
+
+    if context.portal_type in ['opengever.document.document',
```
deiferni: i do not like how the list of types hat needs reindexing is spread around the whole application. Could we check against interfaces instead? Or could we keep the list of portal types in one place and use it here and in the indexer (and wherever else it may be relevant).

opengever/base/handlers.py
```diff
@@ -60,6 +62,32 @@ def object_moved_or_added(context, event):
...
+                               'opengever.task.task',
+                               'ftw.mail.mail']:
+        to_reindex.append('containing_subdossier')
+
+    context.reindexObject(idxs=to_reindex)
+
+    # synchronize with model if necessary
+    if context.portal_type in ['opengever.task.task', 'opengever.inbox.forwarding']:
```
deiferni: same here, is this functionality needed elsewhere? could we put the knowledge in a central location where it is taken from? such explicit lists carry a high risk of not being in sync throughout the application.

opengever/core/upgrades/20181120132150_reindex_containing_dossier_or_subdossier_indices/upgrade.py
```diff
@@ -0,0 +1,24 @@
+from ftw.upgrade import UpgradeStep
+from opengever.dossier.interfaces import IDossierMarker
+
+
+class ReindexContainingDossierOrSubdossierIndices(UpgradeStep):
+    """Reindex containing_dossier, containing_subdossier and
+    is_ubdossier indices.
```
deiferni: typo 😄  is_**s**ub...

opengever/core/upgrades/20181120132150_reindex_containing_dossier_or_subdossier_indices/upgrade.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+
+    deferrable = True
+
+    def __call__(self):
+        message = "Ensure containing_dossier, containing_subdossier and "
+        message += "is_ubdossier are properly indexed on all objects"
```
deiferni: 😄

opengever/core/upgrades/20181120132150_reindex_containing_dossier_or_subdossier_indices/upgrade.py
```diff
@@ -0,0 +1,24 @@
...
+        for obj in self.objects({}, message):
```
deiferni: shouldn't you limit the query to everything that can be in a dossier?

opengever/core/upgrades/20181120132150_reindex_containing_dossier_or_subdossier_indices/upgrade.py
```diff
@@ -0,0 +1,24 @@
...
+            if IDossierMarker.providedBy(obj):
+                obj.reindexObject(idxs=["is_subdossier", "containing_dossier"])
+            elif obj.portal_type in ('opengever.document.document',
+                                     'opengever.task.task',
+                                     'ftw.mail.mail'):
+                obj.reindexObject(idxs=["is_subdossier", "containing_dossier",
+                                        "containing_subdossier"])
+            else:
+                obj.reindexObject(idxs=["containing_dossier"])
```
deiferni: will these objects be in a dossier 🤔?

opengever/dossier/handlers.py
```diff
@@ -88,53 +88,53 @@ def save_reference_number_prefix(obj, event):
...
+        )
+    if 'IOpenGeverBase.title' not in attrs:
+        return
 
-def reindex_is_subdossier(dossier, event):
-    """Reindex the is_subdossier index after the dossier have been moved.
-    """
-    dossier.reindexObject(idxs=['is_subdossier'])
+    parent = aq_parent(aq_inner(dossier))
+    is_subdossier = IDossierMarker.providedBy(parent)
```
deiferni: there are methods on a dossier that can to this `get_parent_dossier` and `is_subdossier`, better ask the object than duplicate the functionality

opengever/core/upgrades/20181120132150_reindex_containing_dossier_or_subdossier_indices/upgrade.py
njohner: yes we could, but that is not that easy to do. Would be a long list of portal_types. We could use the same one in the handlers. We would need to manage the list manually. Obtaining that programatically is a bit of a mess.

opengever/core/upgrades/20181120132150_reindex_containing_dossier_or_subdossier_indices/upgrade.py
njohner: If they are not, the indexer sends back `None`...
But OK, let's make that more specific...

opengever/dossier/handlers.py
njohner: Yes right, I simply moved the old code... it is indeed cleaner to use the methods defined in `DossierContainer`. This is ok as all portal_types that provide the `IDossierMarker` also inherit from `DossierContainer` (the marker interface is actually for the `IDossier` behavior).

njohner:
Ok this is ready for re-review. I've incorporated all your comments. Main changes are:
* I've created two lists of the portal_types for which the `containing_dossier` and `containing_subdossier` indices should be updated. These are then used in the different handlers, indexers and upgrade step
* I removed two unnecessary indexers `isSubdossierIndexer` and `is_subdossier_dossiertemplate`. All they did was get `is_subdossier` of the object, which is the name of the index, so that this is the default behavior.
* Check whether objects provide `ITask` instead of their `portal_type` to decide whether a call to `sync_task` is necessary.
I've left the new changes in separate commits, for readability, but I'll rewrite history once the review process is done.

opengever/core/upgrades/20181120132150_reindex_containing_dossier_or_subdossier_indices/upgrade.py
```diff
@@ -1,23 +1,24 @@
...
 
     deferrable = True
 
     def __call__(self):
         message = "Ensure containing_dossier, containing_subdossier and "
-        message += "is_ubdossier are properly indexed on all objects"
-        for obj in self.objects({}, message):
+        message += "is_subdossier are properly indexed on all objects"
+
+        for obj in self.objects({'portal_type': TYPES_WITH_CONTAINING_DOSSIER_INDEX}, message):
```
deiferni: sorry i unintentionally created a trap. the upgrade should snapshot the list. we may add new items to the list later, but these will have their own upgrade and only after that upgrade is run indexing them is guaranteed to work. in this case it is rather theoretical, but i would still stick to the practice to avoid the error in places where it actually might cause problems.

njohner: right, that's safer.

njohner:
Ok I made the suggested correction in the upgrade step. I also rebased and touched the upgrade step. This should be ready to 🚢

PR #5131, opened 2018-11-22
Add buttons to download toc json for managers

### LINKED ISSUE #4908: add debug links for toc json files
The JSON files for alphabetical and repository based TOC can be downloaded via an `as_json` method, see https://github.com/4teamwork/opengever.core/blob/master/opengever/meeting/browser/toc.py#L68. In https://github.com/4teamwork/opengever.core/pull/4846 we have added debug views as links/actions into the UI. We should do the same for the TOC links, i propose to add links in the highlighted area (for Manager only) 
screen shot 2018-09-25 at 14 44 20

njohner:
We add buttons in the period view to download the toc json. The view was available for all users before, we make it only available to managers and also only show the buttons for managers.

**The new buttons for managers in the periods view**
screen shot 2018-11-22 at 10 36 49

**The toc JSON buttons are not shown for non managers**
screen shot 2018-11-22 at 10 40 59

There were problems when displaying the periods in a narrow window, as buttons would wrap. This was already present before these changes, but less evident as the minimal width needed to display the list correctly was smaller. We modify the css as to forbid button wrapping:

**Button wrapping as it was before**
screen shot 2018-11-22 at 10 07 24

**Now buttons will not wrap**
screen shot 2018-11-22 at 10 07 56

resolves #4908


opengever/meeting/tests/test_periods.py
```diff
@@ -131,3 +133,90 @@ def test_period_title_with_start_and_end_date(self):
             date_to=date(2010, 12, 31), committee=self.committee_model))
         self.assertEqual('Foo (Jan 01, 2010 - Dec 31, 2010)',
                          period.get_title())
+
+    @browsing
+    def test_toc_json_button_only_shown_for_managers(self, browser):
```
deiferni: this would be a good occasion to rewrite this test as an integration test 🙄.

njohner: Boy scout got a bit lost in the test isolation issues, but the tests are now on the integration layer.

njohner:
⚠️ Do not merge ⚠️ 
It seems button display got broken during rebasing.

njohner:
@deiferni Can you have another look? I just ported the tests in `opengever/meeting/tests/test_periods.py` to the integration layer, as you suggested.

PR #5144, opened 2018-11-26
Make ReindexTaskPrincipalsInGlobalindex upgradestep more robust.

phgross:
Skip tasks where no representing sql task is stored in the globalindex. This situation should never happen and is something which should be visible in the validation scripts (https://github.com/4teamwork/opengever.maintenance/pull/149), but should not let the
upgrade fail.

So we skip and log a warning instead if a corresponding sql task is missing.


lukasgraf:
Sorry, just a bit too late - still, LGTM 👍 😉

PR #5148, opened 2018-11-27
 Show one line of each of title and description in livesearch result

njohner:
* We improve the livesearch result display. We now show one line of the title and one line of the description. Both are truncated if necessary and an ellipsis is added if they are truncated.
* Instead of truncating them in the python code, we now do this in css.
* There was also an issue with solr, where the description would not be displayed. This has now been fixed (we forgot to get the description field in the request to solr).

Here a snapshot on IE.

screen shot 2018-11-27 at 16 50 54

### DIFF
--- a/opengever/advancedsearch/skins/advancedsearch_resources/livesearch_reply.py
+++ b/opengever/advancedsearch/skins/advancedsearch_resources/livesearch_reply.py
@@ -180,19 +180,20 @@ else:
         if mime_type_klass:
             css_klass = mime_type_klass
 
-        write('''<a href="%s" title="%s" class="dropdown-list-item LSRow"><span class="%s"/><div>%s</div>'''
-              % (itemUrl, title, css_klass, title))
-        description = html_quote(safe_unicode(result.Description))
+        write('''<a href="%s" title="%s" class="dropdown-list-item LSRow">''' % (itemUrl, title))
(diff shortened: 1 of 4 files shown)


opengever/advancedsearch/skins/advancedsearch_resources/livesearch_reply.py
```diff
@@ -175,37 +173,27 @@ def write(s):
...
-        write('''</a>''')
-        full_title, display_title, display_description = None, None, None
+        write('''<a href="%s" title="%s" class="dropdown-list-item LSRow">''' % (itemUrl, title))
+        write('''<span class="%s"/><span class="dropdown-list-item-content">''' % (css_klass))
+        write('''<div class="LSTitle">%s</div>''' % (title))
+
+        description = html_quote(safe_unicode(result.Description))
+        write('''<div class="LSDescr">%s</div>''' % (description))
+        write('''</span></a>''')
+        title, description = None, None
```
njohner: If anyone knows why we reset these to `None`... to me this seems completely useless, but I didn't want to break anything, so left it there...

phgross: Hmm no idea, we'll improve the livesearch in the next release, so no need to investigate this in detail.

PR #5152, opened 2018-11-29
Ignore local virtualenv folders from git

jone:
I started using a local virtual env for installing buildout ([my script](https://github.com/jone/dotfiles/blob/master/bin/buildout), for the curious).

This left me with the untracked folders `include/` and `lib/`. This PR adds those folders to the git ignore file.

Since a lot of the entries in the git ignore file were not very explicit, I cleaned up the git ignores file in a separate commit:
- Ignored files and folders in the root of the project should be prefixed with a `/` in the `.gitignore` file in order to not match recursively.
- Entries which should match recursively should not begin with a slash.
- Directories should end with a `/` in order to indicate that this is a directory.


jone:
@Rotonen I'm using `pyenv-virtualenv` too.

PR #5156, opened 2018-11-29
Testserver for GEVER

jone:
### Goal
The goal is that GEVER provides a testing server, so that the new GEVER user interface as well as third party applications can run tests against a real, stable and isolated GEVER testserver.

Third party applications which are built on top of GEVER (e.g. using GEVER as DMS) and heavily rely on GEVER need to be able to have a GEVER installation running in their tests. In order to work properly, the GEVER server must provide decent isolation between tests.

### Idea
- The third party application somehow installs GEVER (this is out of scope here; might be builtout or Docker or whatever).
- The third party application starts the server in a subprocess (`bin/testserver`).
- The testserver installs a GEVER and sets up a fixture; probably the existing one. This is important so that the third party tests must do as few as possible.
- The testserver provides a second server which accepts "setup" and "teardown" calls, applying the testing isolation.
- The third party testrunner remote-controls the testserver state by using the "setup" and "teardown" commands.

### Implementation
- The `bin/testserver` script is generated, using `plone.app.robotframework`.
- The testserver starts up a GEVER testing layer and installs the standard testing fixtures.
- The testserver also starts an XMLRPC-Sever, accepting `zodb_setup` and `zodb_teardown` calls for each test in order to isolate the database.
- The `bin/testserver-selftest` boots up the testserver and tests that it works and isolates properly.
- A test-buildout-config adds a new job to Jenkins for the testserver selftest.
- The SQL is currently NOT isolated properly because it is a hard problem to solve as we have seen when changing the testing infrastructure. The plan is to ignore that for now and solve it later as most REST-API-calls do not use SQL data anyways.
- Fixture / database caches are not supported yet as it does not work properly. The pain is not big though when working on third party applications as the testserver can be left running for a long time.
- My expectation is that there are things which are not properly isolated or not properly set up. We will need to fix and update that on the way as we use it practice.
- A new service user is added to the fixture with a publicly known service key, so that third party applications can start testing right away.
- The testserver is documented in the readme.
- Ports are configurable (see readme).
- An example how it can be used with VCR-tapes can be found in the RIS project: https://github.com/4teamwork/ris/pull/505

Demo usage:
```sh
# Start testserver in foreground
$ bin/testserver

# After waiting a long time for the fixture to be loaded, remote-control the server in a second terminal tab:
$ bin/testserverctl zodb_setup

# Visit http://localhost:55001/plone and change some things.
# Then trigger the isolation:
$ bin/testserverctl zodb_teardown
$ bin/testserverctl zodb_setup

# The changes are now reverted. Do other changes.
# ....
```


opengever/core/testserver_selftest.py
```diff
@@ -0,0 +1,113 @@
...
+            self.selftest()
+        finally:
+            print ansi_green('> STOPPING TESTSERVER')
+            self.stop_testserver()
+
+    def selftest(self):
+        with browser:
+            self.testserverctl('zodb_setup')
+            with browser.expect_unauthorized():
+                browser.open('http://localhost:55001/plone/')
```
Rotonen: This assumption will break with a newer `plone.testing` as I've changed the default to pick a random free port.

Rotonen:
Can this be made so multiples of these can run in parallel to each other? Should work by moving the `Data.fs` and sqlite dump to a private temp directory changing the relevant paths appropriately.

Additionally it should then return the port it picks and this port can then be used as a parametre as to which instance the commands apply to.

I'm predicting multiples of these will have to run in parallel to each other on the same CI server and this would give us the ability to parallelise within a run as well.

jone:
Yes, it makes sense to run it in parallel, I just didn't yet finish 😉 
Regarding caching: at the moment the caching layer is not included, therefore no files are written and only the port is the problem regarding running in parallel.

I didn't yet include caching because it is buggy combined with functional testing / ZSERVER. I'm getting "connection closed" and cannot recover. This means though that the setup is slow.

I would like to have a service which provides testservers and solves parallel builds. That means a service which has a pool of already running testservers. As third party application I can then contact the service, initiate a session and receive an URL with custom port. I run my test against it and then tell the service to terminate the session. This could drastically reduce the setup time when testing against a specific release of GEVER or a "more or less latest" build. It does not help a lot when testing against the latest commit (we probably need to start a new testserver) but it could work the same way.

lukasgraf:
Just noticed something when running the selftest locally:

The repository can't be added because german translations seem to not work (completely).

I've noticed this as well when browsing the testserver instance in a real browser. In some places German translations work, but in others they don't:

german_translations



The first failure in the selftest is because of the missing translation of `Ordnungssystem`, but replacing that with `RepositoryRoot` then just leads to different failures where missing German translations are the cause. So the problem seems to be with the translations, not the selftest.

I don't think this is a blocker for merge though, I say we go ahead and merge it, and fix things along the way.

Still, for reference, here's the traceback:
```
lukasgraf@helios ~/Plone/buildouts/opengever/og-testserver $ bin/testserver-selftest
> STARTING TESTSERVER
> bin/testserver
Plone:  http://localhost:60601/plone
XMLRPC: http://localhost:60602
11:19:04 [ wait ] Starting Zope 2 server
ftw.testing PATCHING: Products.CMFQuickInstallerTool.events.handleProfileImportedEvent with noop
ftw.testing PATCHING: Products.CMFQuickInstallerTool.events.handleBeforeProfileImportEvent with noop
Installing fixture. Have patience.
Finished installing fixture.
collective.indexing: Processing queue...
11:20:30 [ ready ] Started Zope 2 server
> TESTSERVER IS READY
> PERFORMING SELFTEST
> bin/testserverctl zodb_setup
INFO:plone.protect:auto rotating keyring _system
INFO:plone.protect:auto rotating keyring _forms
INFO:plone.protect:auto rotating keyring _anon
INFO:plone.protect:auto rotating keyring _system
INFO:plone.protect:auto rotating keyring _forms
INFO:plone.protect:auto rotating keyring _anon

ftw.testbrowser dump: /var/folders/d5/8djs4qxd78lb98g8r7hyp7wr0000gn/T/ftw.testbrowser-PXBEJs.html > STOPPING TESTSERVER

11:20:35 [ wait ] Stopping Zope 2 server
  File "./bin/robot-server", line 417, in <module>
    sys.exit(plone.app.robotframework.server.server())
  File "/Users/lukasgraf/Plone/eggs/plone.app.robotframework-1.2.0-py2.7.egg/plone/app/robotframework/server.py", line 185, in server
    start(args.layer)
  File "/Users/lukasgraf/Plone/eggs/plone.app.robotframework-1.2.0-py2.7.egg/plone/app/robotframework/server.py", line 60, in start
    zsl.stop_zope_server()
  File "/Users/lukasgraf/Plone/eggs/plone.app.robotframework-1.2.0-py2.7.egg/plone/app/robotframework/server.py", line 266, in stop_zope_server
    tear_down()
  File "/Users/lukasgraf/Plone/eggs/plone.app.robotframework-1.2.0-py2.7.egg/plone/app/robotframework/server.py", line 345, in tear_down
    l.tearDown()
  File "/Users/lukasgraf/Plone/eggs/plone.app.testing-4.2.6-py2.7.egg/plone/app/testing/layers.py", line 113, in tearDown
    with z2.zopeApp() as app:
  File "/Users/lukasgraf/Plone/python/parts/opt/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "/Users/lukasgraf/Plone/eggs/plone.testing-4.1.2-py2.7.egg/plone/testing/z2.py", line 278, in zopeApp
    app = addRequestContainer(Zope2.app(connection), environ=environ)
  File "/Users/lukasgraf/Plone/eggs/Zope2-2.13.28-py2.7.egg/Zope2/__init__.py", line 52, in app
    return bobo_application(*args, **kw)
  File "/Users/lukasgraf/Plone/eggs/Zope2-2.13.28-py2.7.egg/App/ZApplication.py", line 75, in __call__
    return connection.root()[aname]
  File "/Users/lukasgraf/Plone/eggs/ZODB3-3.10.7-py2.7-macosx-10.4-x86_64.egg/ZODB/Connection.py", line 366, in root
    return RootConvenience(self.get(z64))
  File "/Users/lukasgraf/Plone/eggs/ZODB3-3.10.7-py2.7-macosx-10.4-x86_64.egg/ZODB/Connection.py", line 248, in get
    p, serial = self._storage.load(oid, '')
AttributeError: 'NoneType' object has no attribute 'load'
  File "bin/testserver-selftest", line 408, in <module>
    sys.exit(opengever.core.testserver_selftest.selftest())
  File "/Users/lukasgraf/Plone/buildouts/opengever/og-testserver/opengever/core/testserver_selftest.py", line 126, in selftest
    TestserverSelftest()()
  File "/Users/lukasgraf/Plone/buildouts/opengever/og-testserver/opengever/core/testserver_selftest.py", line 34, in __call__
    self.selftest()
  File "/Users/lukasgraf/Plone/buildouts/opengever/og-testserver/opengever/core/testserver_selftest.py", line 46, in selftest
    factoriesmenu.add('Ordnungssystem')
  File "/Users/lukasgraf/Plone/eggs/ftw.testbrowser-1.30.0-py2.7.egg/ftw/testbrowser/pages/factoriesmenu.py", line 42, in add
    ', '.join(addable_types(browser=browser))))
ValueError: The type "Ordnungssystem" is not addable. Addable types: Document, Task, E-Mail, Business Case Dossier, Disposition, Bild, Committee, ContactFolder, Datei, Dossier template, Dummy, Forwarding, Inbox, Inbox Container, Kollektion, Link, Meeting Dossier, Nachricht, Ordner, Paragraph Template, Private Dossier, Private Root, Private folder, Proposal, RepositoryFolder, RepositoryRoot, Sablon Template, Seite, Submitted Proposal, TaskTemplate, Template Folder, Termin, WorkspaceRoot
```

bin/testserverctl
```diff
@@ -0,0 +1,19 @@
...
+
+import os
+import sys
+import xmlrpclib
+
+CTL_PORT = os.environ.get('TESTSERVER_CTL_PORT', '55002')
+
+proxy = xmlrpclib.ServerProxy('http://localhost:{}'.format(CTL_PORT))
+
+if len(sys.argv) < 2:
```
Rotonen: Please use argparse.

opengever/core/testserver_selftest.py
```diff
@@ -0,0 +1,126 @@
...
+        args = ['bin/testserver']
+        print ansi_blue('>', *args)
+        self.testserver_process = subprocess.Popen(args)
+        self.testserver_thread = Thread(target=self.testserver_process.communicate)
+        self.testserver_thread.start()
+
+    def wait_for_testserver(self):
+        """Block until the testserver is ready.
+        """
+        timeout_seconds = 60 * 60
```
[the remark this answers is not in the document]
jone: Indeed. Changed it to 5 minutes since we want our tests to be faster than 10 minutes so it needs to be started in 5 minutes in order to be well in time. Should be fine 👍

bin/testserverctl
jone: Indeed 👍

jone:
Updated and addressed the comments.
I didn't look into translations though, but I changed the selftest anyway to use the rest api.

Changes:
- frozen time with ftw.tokenauth compatibility (not frozen there)
- static uids
- static intids
- changed selftest to use the REST api

Freezing all the dynamic values make testing third party applications a lot more comfortable, especially when using VCR tapes.

bin/testserverctl
```diff
@@ -1,19 +1,14 @@
...
 
+import argparse
 import os
-import sys
 import xmlrpclib
 
 CTL_PORT = os.environ.get('TESTSERVER_CTL_PORT', '55002')
 
-proxy = xmlrpclib.ServerProxy('http://localhost:{}'.format(CTL_PORT))
+parser = argparse.ArgumentParser()
```
Rotonen: To keep the previous functionality of telling what went wrong, you can pass a `description` and/or `epilog` kwargs to `ArgumentParser()`. Merely a FYI of a useful thing.

Example:
https://github.com/plone/buildout.coredev/blob/6dbce32de42b070d7e87b42904e745796357a590/bin/mtest3#L481-L485

jone: Yes, I was too lazy 🙈 
But I think it is good enough as the purpose is more to serve as an example.
In other (python) projects I'd prefer implementing that myself in the project.
Example: https://github.com/4teamwork/ris/pull/505/files#diff-b358d6cbb3b5023fe1115080abeb775fR130

opengever/core/testserver.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class TestServerFunctionalTesting(FunctionalTesting):
+
+    @contextmanager
+    def isolation(self):
+        start = datetime(2018, 11, 22, 14, 29, 33, tzinfo=pytz.UTC)
+        with freeze(start, ignore_modules=['ftw.tokenauth.oauth2.jwt_grants']):
+            with staticuid('testserver-session'):
+                with time_based_intids():
```
jone: Probably a bad idea since the time does not move 🤔

jone:
I have updated the PR: the tests are now passing.
The tests were failing because of how the testserver process was terminated in the selftest: since `bin/testserver` was a wrapper shell script, I terminated it by killing the process group. This caused a process of jenkins to be killed, which caused the test failure.

So I had to refactor `bin/testserver` to no longer use a shell wrapper script but directly be the python process by changing the buildout.

I have also added a clock freezer, a static uid generator and an incrementing intid generator in order to receive the exact same payload when e.g. reading a dossier through the REST API every time the test is run.

So the PR is ready for another review 😉 Thanks!

lukasgraf:
@jone looks good to me! 👍 Tested again locally, works like a charm 🎉
I say 🚢 it!

PR #5160, opened 2018-12-03
Fix proposal workflows, don't acquire View.

### LINKED ISSUE #5091: Proposal workflows permits Contributor to view
The current proposal workflow acquires the View permission which therefore allows `Contributor` to View proposals. This breaks the "rule" that a user which gets permitted on a dossier with a task only sees the related documents. 

Which means, Users which are only permitted on a dossier by a task, does also see all proposals, which is wrong.

deiferni:
Tasks set local roles on dossiers for users that would otherwise have no permission to see the dossier. They do so by enabling `Contributor` to `View`. Proposals currently acquired the `View` permission from their parent. Thus they were also able to `View` Proposals that they should not have been able to see.

This PR fixes the proposal workflows. Fixing the workflow had two consequences:
- if a task relates a proposal document, the proposal should also be visible to the task responsible. This is taken care of by setting the same local roles on the proposal as for the proposals document.
- the proposal listing from sql currently assumed that whoever is able to `View` the dossier is also allowed to see all proposals. As this assumption no longer holds the proposal tab has to implement security checks now. This is a stopgap measure as switching to brain based listings is not possible yet, this will be described in `OGIP 50`.

Fixes #5091.


deiferni:
@4teamwork/gever a review would be much appreciated 💛!

PR #5163, opened 2018-12-04
Fix file indexes upgradestep for deployments using solr.

phgross:
The change from PR #5024 only added catalog indexes, but did not add corresponding indexes to the solr schema. So the upgradestep was useless for deployments with an enabled solr search or even worse it almost removed all solr data because of a curios behave of solr or ftw.solr (I'll investigate this behavior in a separate issue).

So I added the new indexes to the schema and fix the current upgradestep.

### DIFF
--- a/opengever/core/upgrades/20181101150743_add_document_filesize__filename_and_file_extension_to_document_metadata/upgrade.py
+++ b/opengever/core/upgrades/20181101150743_add_document_filesize__filename_and_file_extension_to_document_metadata/upgrade.py
@@ -10,12 +10,21 @@ class AddDocumentFilesize_filenameAndFileExtensionToDocumentMetadata(UpgradeStep
 
     def __call__(self):
         self.install_upgrade_profile()
-        self.catalog_add_index('filesize', 'FieldIndex')
-        self.catalog_add_index('file_extension', 'FieldIndex')
+
+        if not self.catalog_has_index('filesize'):
+            self.catalog_add_index('filesize', 'FieldIndex')
+
+        if not self.catalog_has_index('file_extension'):
+            self.catalog_add_index('file_extension', 'FieldIndex')
 
         query = {
             'object_provides': IBaseDocument.__identifier__,
             }
 
         for basedocument in self.objects(query, 'Index filesize, filename and file extension of IBaseDocument objects.'):
-            basedocument.reindexObject(idxs=['filesize', 'file_extension'])
+
+            # There is no index called filename, but we need to add it to the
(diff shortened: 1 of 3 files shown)


solr-conf/managed-schema
```diff
@@ -137,6 +137,9 @@
     <field name="dossier_review_state" type="string" indexed="true" stored="false" />
     <field name="email" type="string" indexed="true" stored="false" />
     <field name="end" type="pdate" indexed="true" stored="false" />
+    <field name="file_extension" type="string" indexed="true" stored="false" />
+    <field name="filename" type="string" indexed="true" stored="false" />
+    <field name="filesize" type="string" indexed="true" stored="false" />
```
lukasgraf: That should probably by of type `pint` instead of `string`. Double check with buchi please.

opengever/core/upgrades/20181101150743_add_document_filesize__filename_and_file_extension_to_document_metadata/upgrade.py
```diff
@@ -10,12 +10,21 @@ class AddDocumentFilesize_filenameAndFileExtensionToDocumentMetadata(UpgradeStep
...
             }
 
         for basedocument in self.objects(query, 'Index filesize, filename and file extension of IBaseDocument objects.'):
-            basedocument.reindexObject(idxs=['filesize', 'file_extension'])
+
+            # There is no index called filename, but we need to add it to the
+            # idxs list, to make sure that solr's filename index is indexed
+            # as well.
+            basedocument.reindexObject(
+                idxs=['filesize', 'file_extension', 'filename'])
```
deiferni: couldn't you do something as in https://github.com/4teamwork/opengever.core/blob/6c1e077285f68b8e8e0cf30f9e95b514ab3f7a00/opengever/core/upgrades/20180515074210_reindex_subject_index_on_solr/upgrade.py for the solr-only index 🤔? if this is a solr-issue only, maybe only fix the solr side? or is the original upgrade-step a new/recent one?

phgross: I don't see the goal by implementing a separate one. The existing upgradestep is broken, and need to be fixed. As you see I did not touch this one, because existing solr deployments will already be fixed by a manual reindex.

solr-conf/managed-schema
phgross: @lukasgraf updated ...

PR #5164, opened 2018-12-04
Add support for simple language codes in request language negotiation

### LINKED ISSUE #5162: Language request negotiation doesn't work for combined lang codes
The way our `language_tool` is currently set up, request language negotiation based on the brower's `Accept-Language` header should be performed, if no explicit language choice has been made by the user (==cookie).

---

lang_tool

---

However, this only works if the user has a language preference for a **combined** language code in their browser (like `de_ch` (German (Switzerland))), not for simple codes (like `de` (German)).

### Steps to reproduce

- Set up a fresh GEVER, and change the default language to `French (Switzerland)`
- Make sure your browser's first language perference is just plain `German` (and `German (Switzerland)` isn't part of the browser preference list)
- Open GEVER (in inkognito mode, or clear your cookies)

### Expected Results

Request negotiation should have picked German as the negotiated language

### Actual results

The browser preferences are ignored, and the negotiation instead falls back to the site default language (French) 

Internal Ticket for reference: https://extranet.4teamwork.ch/support/ph-vs/tracker/513

### Analysis

This is because the [LanguageTool doesn't normalize language codes to the simple form](
https://github.com/plone/Products.PloneLanguageTool/blob/master/Products/PloneLanguageTool/LanguageTool.py#L485-L497) if `use_combined_language_codes` is enabled, therefore works if the browser sends combined codes, but fails to look up something like `de` in the list of supported languages (because we currently *only* specify combined language codes in those).

So for request language negotiation to work properly with `use_combined_language_codes` enabled, we would need to explicitly add both `de` and `de_ch`, as well as `fr` and `fr_ch` to supported languages.

The other alternative would be to disable `use_combined_language_codes` - this would mean that all language codes everywhere would be canonicalized to just `de` and `fr`. I currently prefer this approach.

lukasgraf:
This extends request language negotiation in a way where we also support browsers sending **simple language codes** in their `Accept-Language` headers (even though we have `use_combined_language_codes` turned on).

We do this by using a hardcoded mapping from simple to combined language codes, which we know to be appropriate because of our customer base.

Fixes #5162

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2019.1.0 (unreleased)
 ---------------------
 
+- Add support for simple language codes in request language negotiation [lgraf]
 - Fix typo in favorite error message. [njohner]
 - Only display .docx files as possible proposal documents. [Rotonen]
 - Fix task revoking permissions on close/reassign. [phgross]
--- a/opengever/base/configure.zcml
+++ b/opengever/base/configure.zcml
@@ -68,6 +68,11 @@
       name="clipboard"
       />
 
+  <adapter
+      factory=".language.NegotiateLanguage"
+      for="* opengever.base.interfaces.IOpengeverBaseLayer"
+      />
+
   <configure package="collective.quickupload.browser">
     <browser:page
         class="opengever.base.quickupload.OGQuickUploadView"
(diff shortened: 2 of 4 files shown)


lukasgraf:
rebased...

PR #5166, opened 2018-12-05
Add missing changed index to solr and fix tabbedview helper fallback.

phgross:
-  Fix changed fallback in documents listings for solr documents. This fixes an Bug on document listings when filtering and solr is enabled (see https://sentry.4teamwork.ch/sentry/onegov-gever/issues/9683/).
- Added missing `changed` field to the solr schema 
-  Added tests which checks if solr conf contains all catlaog indexes, and metadata or are marked as catalog only. This should help to not forget the solr-conf when adding a new index/metadata

There is no upgradestep needed, we'll need to fix/reindex the existing solr installations manually.


njohner:
👍 Looks good to me

PR #5174, opened 2018-12-10
Fix task revoking permissions on close/reassign

phgross:
Avoid unauthorized raises, when current user loses the access rights on the dossier during task closing. The Unauthorized has been raised, because we revoked the permission and fetch the subdossiers afterwards, so the getObject on the brain raised the `Unauthorized`. I've fixed this by changing the order, first fetch all subdossiers and than revoke the permission on the main dossier.

Besides that I also implemented a fallback to the add response form, when the current users loses access rights. If that's the case we now redirect to the personal overview and show a corresponding statusmessage.

Backport needed: `2018.5`

### DIFF
--- a/opengever/task/localroles.py
+++ b/opengever/task/localroles.py
@@ -174,12 +174,6 @@ class LocalRolesSetter(object):
 
     def revoke_on_distinct_parent(self):
         distinct_parent = self.get_distinct_parent()
-        manager = RoleAssignmentManager(self.get_distinct_parent())
-        manager.clear(
-            ASSIGNMENT_VIA_TASK,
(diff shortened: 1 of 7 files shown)


opengever/task/localroles.py
```diff
@@ -191,10 +185,16 @@ def revoke_on_distinct_parent(self):
                              IProposal.__identifier__],
             path='/'.join(distinct_parent.getPhysicalPath()))]
 
+        manager = RoleAssignmentManager(self.get_distinct_parent())
```
lukasgraf: `distinct_parent` is already fetched on line 176, and could be reused here.

phgross: @lukasgraf your right, updated.

PR #5177, opened 2018-12-10
Implement is_shadow_document() for mails

Rotonen:
* Used `unix2dos` to convert the test asset for easier comparability
* Implemented `is_shadow_document()` for mails
* Added tests for attaching mails via Office Connector


opengever/mail/mail.py
```diff
@@ -352,6 +352,9 @@ def get_download_view_name(self):
             return '@@download/original_message'
         return 'download'
 
+    def is_shadow_document(self):
```
deiferni: if this is now part of `BaseDocumentMixin` either move it up as default implementation or provide an abstract definition in `BaseDocumentMixin` that raises `NotImplementedError`.

Rotonen:
Ready for rereview.

PR #5190, opened 2018-12-12
Merge plonetheme.teamraum profiles into opengever.core

Rotonen:
Settled for just merging the profiles into `opengever.core` for now and making the default publisher encoding UTF-8 as we now go through diazo transforms in the tests and that'd not work out with the Zope 2 default encoding of latin9.


Rotonen:
As https://github.com/4teamwork/opengever.core/pull/5212 is merged, merging this one should be now in any case safe as well.

Rotonen:
@lukasgraf and now with changelog entries.

PR #5201, opened 2018-12-13
Closing subdossiers with unrestricted searching and elevated privileges

### LINKED ISSUE #5153: Subdossier active in deactivated dossier
When deactivating a dossier, we loop only over the dossier that the current user sees, so we can end up with activ subdossiers in an inactive dossier. The main dossier can then not be reactivated, because the activate transition does not exist for the already activ subdossiers.

https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/446

Steps to reproduce:

create a dossier with a subdossier, remove all rights on the subdossier for a given user, then with that user deactivate the main dossier.

Fix proposition: Forbid deactivating a dossier when there are subdossiers that are still active and that the current user cannot deactivate.

Rotonen:
* Implemented an unrestricted search option for `get_subdossiers()`
* Closing unrestrictedly searched subdossiers with elevated privileges

Closes #5153

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -28,6 +28,7 @@ Changelog
 - Translate z3c.formwidget.query (nothing). [njohner]
 - Add buttons for managers to get the toc json for meeting periods. [njohner]
 - Add absent members to the meeting protocolData. [njohner]
+- Also close subdossiers the current user cannot view. [Rotonen]
 - Add description field to paragraph add form. [njohner]
 - Fix bug in disposition ech0160 folder model. [njohner]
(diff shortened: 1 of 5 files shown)


lukasgraf:
This is a misguided approach: This deliberately circumvents the permissions needed for resolving (sub)dossiers, which is not what should happen.

Instead, it should be addressed exactly as proposed in the original issue:
> Fix proposition: Forbid deactivating a dossier when there are subdossiers that are still active and that the current user cannot deactivate.

(so, prevent the user from resolving the main dossier alltogether, and display a meaningful `error` level status message).

PR #5207, opened 2018-12-17
Store agenda item numbers as integers

### LINKED ISSUE #5046: Provide unformatted agenda item number
Currently the agenda item number is stored in the database as string which is terminated by a dot. This is also provided like that to sablon when creating word documents. It is desired that this behavior is changed to only providing the number so that word-documents can then include or not include the dot as they want.

- migrate database, use int for number
- provide number without dot to sablon
- migrate existing templates, or alternatively provide backwards compatibility by offering two attributes in `JSON`, one with and one without the dot

see 
https://github.com/4teamwork/opengever.core/blob/11b310b01362ff3faa122f6677513543ae757a27/opengever/meeting/model/meeting.py#L471-L482

Ref: 
https://feedback.onegovgever.ch/t/traktandennummer-punktieren/1174
https://basecamp.com/2768704/projects/13732725/todos/365659132

Rotonen:
* Swap the column type for agenda item numbers
* Preserve the old API
* Introduce a new unformatted attribute

Closes #5046


opengever/core/upgrades/20181214121838_migrate_agenda_item_numbers_from_strings_to_int/upgrade.py
```diff
@@ -0,0 +1,46 @@
...
+    def create_transitory_table_query():
+        return table(
+            'agendaitems',
+            column('id'),
+            column('item_number'),
+            column('item_number_raw'),
+        )
+
+    def migrate_data(self, transitory_table):
+        agendaitems = self.connection.execute(transitory_table.select()).fetchall()
```
deiferni: use `self.execute`

opengever/core/upgrades/20181214121838_migrate_agenda_item_numbers_from_strings_to_int/upgrade.py
```diff
@@ -0,0 +1,46 @@
...
+        self.migrate_data(transitory_table)
+        self.drop_old_column()
+        self.rename_new_column()
+
+    def create_new_column(self):
+        item_number_raw = Column('item_number_raw', Integer)
+        self.op.add_column('agendaitems', item_number_raw)
+
+    @staticmethod
+    def create_transitory_table_query():
```
deiferni: this name confused me, this is not a query, just the table definition.

the convention in other upgrade is to define these on a per module level. would make sense to follow the convention.

opengever/core/upgrades/20181214121838_migrate_agenda_item_numbers_from_strings_to_int/upgrade.py
```diff
@@ -0,0 +1,46 @@
...
+            'agendaitems',
+            column('id'),
+            column('item_number'),
+            column('item_number_raw'),
+        )
+
+    def migrate_data(self, transitory_table):
+        agendaitems = self.connection.execute(transitory_table.select()).fetchall()
+        for agendaitem in agendaitems:
+            stripped_item_number = agendaitem.item_number.strip('.')
```
deiferni: what happens if it is `None`, theoretically that could be the case as the column is nullable.

Rotonen:
Ready for rereview.

PR #5209, opened 2018-12-18
Remove deprecated docprops.

### LINKED ISSUE #2218: Entfernen von deprecated Doc-Properties
Die folgenden Doc-Properties sind mit der Implementation #2186 deprecated und können bei Gelegenheit (und mit einem Sicherheitsabstand, der gross genug ist) entfernt werden:
- `Document.ReferenceNumber`
- `Document.SequenceNumber`
- `Dossier.ReferenceNumber`
- `Dossier.Title`
- `User.ID`
- `User.FullName`

Zuvor muss jedoch sichergestellt werden, dass alle Kunden die neuen Doc-Properties verwenden. Sobald die Properties ausgebaut sind sollte auch die Doku (intern) aktualisiert werden.

njohner:
Here we remove the deprecated docproperties. I also adapted the asset files in the tests.

resolves #2218


njohner:
Ok sure. I thought that if it was set for 2019.1, it meant that we had already made sure the clients weren't using those anymore 😄 .

njohner:
Ok this is ready

njohner:
After discussing this with @lukasgraf , it came out that we will probably have to support the deprecated doc properties for ever and more. Otherwise this could lead to hard to detect problems where some docproperties get updated in a file and others not anymore.
He proposes we:
1. run a list of deprecated docproperties that we continue to update if present.
2. log to sentry everytime we encounter a document using a deprecated docproperty.

lukasgraf:
@njohner so I would say we proceed like this:

- We keep your changes that update the `*.docx` files in the examplecontent and the accompaning test changes
- But don't remove of the deprecation notice in the docs (it actually still serves a purpose, and we're never gonna fully drop support for those legacy properties)

njohner:
Ok all done here.

PR #5211, opened 2018-12-18
Amend the keywordwidget js template to also list team members

### LINKED ISSUE #5129: Klick auf Team bei der Dossierdarstellung sollte Team-Members auflisten
Ähnlich wie dies bei der Auswahl von Gruppen in der Dossier schützen Ansicht der Fall ist, wäre es sinnvoll wenn bei einem Klick auf ein Team die Members dargestellt würden.

Bsp. "Dossier schützen":
bildschirmfoto_2018-11-21_um_17_56_49
bildschirmfoto_2018-11-21_um_17_56_57

Soll:
bildschirmfoto_2018-11-21_um_17_52_34

Rotonen:
2018-12-18-team-keywordwidget

* Added teams to example content
* Added an empty group and a team for it into the fixtures
* Added a view for the keyword widget to use to fetch team members
* Extended the keyword widget js template to also work with teams
* Used the form on the add task form responsible field

Did I miss some other forms where this new template should now be used for the widget?

Closes #5129

### DIFF
--- a/opengever/base/browser/list_teammembers.py
+++ b/opengever/base/browser/list_teammembers.py
@@ -1,0 +1,45 @@
+from opengever.ogds.base.actor import Actor
+from opengever.ogds.models.team import Team
+from Products.Five import BrowserView
+from Products.Five.browser.pagetemplatefile import ViewPageTemplateFile
+from sqlalchemy.orm.exc import NoResultFound
+from zExceptions import BadRequest
+
+
+class ListTeamMembers(BrowserView):
+    """Fetch team group and group members by team id."""
+
+    template = ViewPageTemplateFile("list_groupmembers.pt")
+
+    def __init__(self, context, request):
+        super(ListTeamMembers, self).__init__(context, request)
+
+        self.active = None
+        self.members = None
+        self.team_id = None
+        self.group_id = None
+
+    def __call__(self):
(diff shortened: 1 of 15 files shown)


opengever/base/browser/list_teammembers.py
```diff
@@ -0,0 +1,40 @@
...
+        self.group_id = None
+
+    def __call__(self):
+        self.team_id = self.context.REQUEST.get('team', None)
+        self.members = []
+
+        if not self.team_id:
+            return 'no team id'
+
+        team = Team.query.filter_by(team_id=self.team_id).first()
```
lukasgraf: This should be `.one()` instead of `.first()`. Multiple results should never happen, and if they do, we want to know about it.

opengever/base/tests/test_list_teammembers.py
```diff
@@ -0,0 +1,26 @@
+from ftw.testbrowser import browsing
+from opengever.testing import IntegrationTestCase
+
+
+class TestTeamMembersView(IntegrationTestCase):
+
+    @browsing
+    def test_list_groupmembers_view(self, browser):
```
lukasgraf: `test_list_groupmembers` -> `test_list_teammembers` (3x)

PR #5231, opened 2018-12-21
Fix a new integration test which got merged after a fixture amend merge

Rotonen:
Did not realise there was a new integration test in `opengever.base` relying on the dossier count.

My bad.


Rotonen:
I'll just self merge as it is a triviality and this is blocking me from investigating sane layer splits.

PR #5232, opened 2018-12-23
Fix flaky relation tests

phgross:
the order in the relation catalog can't be controlled therefore we check for existence in all relations, instead of relying on on the order.

The test has been converted to integration tests recently and has been noticed as flaky since.

### DIFF
--- a/opengever/base/tests/test_relations.py
+++ b/opengever/base/tests/test_relations.py
@@ -23,8 +23,10 @@ class TestRelations(IntegrationTestCase):
             {'to_id': intids.getId(self.dossier)})]
 
         self.assertEqual(2, len(relations))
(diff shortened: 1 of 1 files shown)


opengever/base/tests/test_relations.py
```diff
@@ -23,8 +23,10 @@ def test_relation_gets_added_in_zc_catalog_when_dossier_is_added(self, browser):
             {'to_id': intids.getId(self.dossier)})]
 
         self.assertEqual(2, len(relations))
-        self.assertEqual(intids.getId(self.dossier), relations[-1].to_id)
-        self.assertEqual(intids.getId(browser.context), relations[-1].from_id)
+        self.assertIn(intids.getId(self.dossier),
```
Rotonen: Rather assert on the complete relation list as the same item can be in there more than once.

x4

PR #5234, opened 2018-12-27
Fix loading of document preview with tooltip

bierik:
As documented in https://extranet.4teamwork.ch/support/stadt-nidau/tracker-gever/95 it may be possible that the documents overview page reloads when loading a tooltip of a document and the document detail view simultaneously.
So when user clicks on the document link in the overview while the browser awaits the response from the tooltip, the tooltip request is aborted. The javascript code checks the response of the tooltip request, which has been aborted, for a `X-Tooltip-Response` header. But because the request has been canceled the header is not present. As a result the javascript code triggers a reload of the documents overview page because it is treated as an error.
Using a flag setting to true when the user clicks on the documents links solved the problem. The flag is checked before reloading the page.
Also adjust the error message when the `X-Tooltip-Response` header is missing.

### DIFF
--- a/opengever/base/browser/resources/tooltip.js
+++ b/opengever/base/browser/resources/tooltip.js
@@ -3,6 +3,7 @@
   "use strict";
 
   var tooltipHideTimer = null;
+  var viewingDocument = false;
 
   var settings = {
     overwrite: false, // Do not reload the tooltip when it's already created
@@ -40,7 +41,7 @@
 
   function isTooltipResponse(data, status, jqXHR) {
     if(jqXHR.getResponseHeader('X-Tooltip-Response') !== "True") {
-      return $.Deferred().reject(jqXHR, "error");
+      return $.Deferred().reject(jqXHR, "X-Tooltip-Response missing on tooltip response.");
     }
     return $.Deferred().resolve(data, status, jqXHR);
   }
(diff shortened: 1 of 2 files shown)


opengever/base/browser/resources/tooltip.js
```diff
@@ -40,7 +41,7 @@
 
   function isTooltipResponse(data, status, jqXHR) {
     if(jqXHR.getResponseHeader('X-Tooltip-Response') !== "True") {
-      return $.Deferred().reject(jqXHR, "error");
+      return $.Deferred().reject(jqXHR, "X-Tooltip-Response missin on tooltip response.");
```
lukasgraf: `missin` -> `missing`

opengever/base/browser/resources/tooltip.js
```diff
@@ -3,6 +3,7 @@
   "use strict";
 
   var tooltipHideTimer = null;
+  var viewingDocumment = false;
```
lukasgraf: `viewingDocumment` -> `viewingDocument`

opengever/base/browser/resources/tooltip.js
```diff
@@ -3,6 +3,7 @@
   "use strict";
```
lukasgraf: Changelog entry is missing.

opengever/base/browser/resources/tooltip.js
```diff
@@ -54,7 +55,9 @@
         $(".showroom-reference").on("click", function() { api.hide(); });
       })
       .fail(function() {
-        location.reload();
+        if (!viewingDocumment) {
```
lukasgraf: Could you please add a comment as to why we do this? Something like `// Avoid cancelling a pending navigation request initiated by the user`

bierik:
@lukasgraf updated

PR #5238, opened 2018-12-29
Squash all mtest runs into one

Rotonen:
After the efforts from @phgross and @njohner we're now in a position where adding anything into the integration tests directly adds to wallclock runtimes.

On the flipside we now have tests running fast enough to have them run in about the same wallclock time as a monolithic 6-wide run, landing us in a position where all further test run improvements land as direct wallclock savings.


test-docs-intern.cfg
```diff
@@ -7,6 +7,8 @@ parts +=
 
 jenkins_python = $PYTHON27
 
+git-clone-depth = 1
```
lukasgraf: Any reason why this can't be upstreamed to[`test-base.cfg`](https://github.com/4teamwork/ftw-buildouts/blob/master/test-base.cfg) in `ftw-buildouts`?

PR #5239, opened 2019-01-03
Add test for handling of special characters in dossier journal.

njohner:
⚠️ do not merge, first release `ftw.pdfgenerator` ⚠️ 

A vertical tab character in a dossier title was preventing closing of a dossier. We now handle this correctly in `ftw.pdfgenerator` and add a regression test here.

This problem emerged in 
https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/493


lukasgraf:
@njohner I just released `ftw.pdfgenerator = 1.6.3` with your changes (didn't do anything else on top of it after all).

PR #5241, opened 2019-01-04
Fix resolved subdossier invalid end dates when resolving main dossier.

### LINKED ISSUE #5204: correct dossier end date after migration to changed
The situation is the following:
* We can easily have a document with a modification date that is posterior to the document date.
* A dossier that was resolved prior to adding the `changed` date could therefore have an `end` date prior to the modification date of the contained documents. 
* After switching to `changed`, which is set to modified, we have a dossier with an invalid end date. Its end date is prior to the contained document `changed` date. 

So if this dossier was a subdossier of a dossier that was not closed yet, the main dossier cannot be closed anymore. We need to correct `end` date of such subdossiers. The `changed` date of the documents is fine in such a case, and it is the subdossier's `end` date that is not correct (according to the new rules for a dossier end date). So it makes more sense to correct the subdossier's `end` date.

njohner:
Invalid end dates for resolved dossiers can happen for several reasons, recently because of the switch from document_date to changed for the calculation of the earliest_possible_end_date. These are always due to software bugs and not user input and should hence be fixed automatically.

We need to be careful because open subdossiers with invalid end dates are due to user input and should still lead to an error message and let the user fix it himself. The approach implemented here works as follows:

* when checking the validity of subdossier end dates (in `ResolveConditions.check_end_dates`), we consider as invalid only dossiers that are not resolved and have an invalid end date. These will prevent dossier resolution and the user will receive an error message
* When resolving the dossier in `Resolver._recursive_resolve`, we set the end date of subdossiers:
  * to the end date of the main dossier if the subdossier has no end date or is not `resolved` and has an invalid end date (This should never happen in my opinion, but the condition was already there, so I left it for safety for now)
  * to the `earliest_possible_end_date` of the subdossier if it is resolved and has an invalid end date.

With this, invalid end dates of resolved dossier will no longer prevent dossier resolution and get corrected.

Note that the end date of inactive dossiers can also get set automatically by the software, to the day of inactivation. While before such dossiers could have an invalid end date (if a document had a `document_date` posterior to the inactivation date), this should not be possible anymore (all `changed` dates will be prior to the inactivation date, as nothing can be modified once the dossier is inactive). Of course they can have an invalid end date if the user set it himself, but then it is his responsibility to reset the end date correctly.

resolves #5204

### DIFF
--- a/opengever/dossier/resolve.py
+++ b/opengever/dossier/resolve.py
@@ -366,8 +366,18 @@ class Resolver(object):
 
         if recursive:
             # check the subdossiers end date
-            # set the parent endate when no or a invalid end date is set
-            if not IDossier(dossier).end or not dossier.has_valid_enddate():
+            # If a subdossier is already resolved, but seems to have an invalid
+            # end date, it's because we changed the resolution logic and rules
+            # over time, and that subdossier's end date has retroactively become
+            # invalid. In this case, we correct the end date according to current
+            # rules and proceed with resolving it.
+            if dossier.is_resolved() and not dossier.has_valid_enddate():
+                IDossier(dossier).end = dossier.earliest_possible_end_date()
+            # if no end date is set or the end_date is invalid, set to the parent
+            # end date. Invalid end_date should normally be prevented by
+            # ResolveConditions._recursive_date_validation, but this correction
+            # would happen for example when resolving a dossier in debug mode.
+            elif not IDossier(dossier).end or not dossier.has_valid_enddate():
                 IDossier(dossier).end = end_date
         else:
             # main dossier set the given end_date
@@ -447,7 +457,12 @@ class ResolveConditions(object):
 
         if not main:
             # check end_date
-            if not dossier.has_valid_enddate():
+            # If a dossier is already resolved, but seems to have an invalid
+            # end date, it's because we changed the resolution logic and rules
+            # over time, and that subdossier's end date has retroactively become
+            # invalid. In this case, should allow the resolving the main dossier
+            # anyway, and correct the invalid end date during dossier resolution.
+            if not dossier.is_resolved() and not dossier.has_valid_enddate():
                 self._invalid_dates.append(dossier.title)
 
         # recursively check subdossiers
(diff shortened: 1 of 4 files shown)


njohner:
Another note is that running the `fix_journal_pdf_document_date.py` is still useful, though not necessary. Journal PDFs in resolved dossiers will no longer prevent resolving a dossier, but they would lead to the dossier end dates being reset to the resolution date (as the journal pdf will have `changed` set to that date). Running the script marks these as journal PDFs so that they are not considered in the `earliest_possible_end_date` calculation anymore. 
I would avoid running the script with the `-f` option though, as this does not make much sense anymore in the light of the present bugfix (this would set the `IDossierJournalPDFMarker` interface to documents that are not a journal PDF).

opengever/dossier/resolve.py
```diff
@@ -366,9 +366,12 @@ def _recursive_resolve(self, dossier, end_date, recursive=False):
 
         if recursive:
             # check the subdossiers end date
-            # set the parent endate when no or a invalid end date is set
-            if not IDossier(dossier).end or not dossier.has_valid_enddate():
+            # if no end date is set, set to the parent end date
+            # if invalid and the subdossier is resolved, set to earliest_possible_end_date
+            if not IDossier(dossier).end or not dossier.has_valid_enddate(ignore_resolved=True):
```
lukasgraf: This works, but I find it a bit hard to read. I would try to make the special case special, and leave the existing logic untouched otherwise. Something like this:

- Drop the `ignore_resolved` keyword argument for `has_valid_enddate()` - even when taking resolved state into consideration, a resolved subdossier enddate that's too young is still technically invalid. It's just that we want to allow it in that special case, and correct it. Encompassing two different definitions of valid makes `has_valid_enddate()` hard to reason about IMHO.

- Specifically check for the special case we want to handle first, and document it as such:

```python
        if recursive:
            # If a subdossier is already resolved, but seems to have an invalid
            # end date, it's because we changed the resolution logic and rules
            # over time, and that subdossier's end date has retroactively become
            # invalid. In this case, we correct the end date according to current
            # rules and proceed with resolving it.
            if dossier.is_resolved() and not dossier.has_valid_enddate():
                IDossier(dossier).end = dossier.earliest_possible_end_date()
```

Issue #5104, opened 2018-11-15
Get to a state where a sequential test run passes again


Rotonen:
Currently a simple test run with `bin/test` will not complete succesfully.

Rotonen:
Currently `bin/test --layer '!opengever.core.testing.opengever:core:memory_db' --layer '!opengever.core.testing.opengever.core:functional:contentstats'` passes.

Rotonen:
And the `opengever.core.testing.opengever:core:memory_db` somehow leaks via `z3c.saconfig` so the `opengever.ogds.base.events` do not fire as 'there already is a database'.

Can probably be remedied by a teardown fix at layer teardown.

PR #5242, opened 2019-01-04
Plug an sqlalchemy session registry leak in the standalone memory db layer

Rotonen:
As `z3c.saconfig` holds a module level dictionary as a cache of scoped sessions we must pop our known-named session off this cache as otherwise our `create_session()` will not do what we expect it to do on other layers.

Additionally sqlalchemy itself also holds a cache of sessions and engines, so we also need to remove the session from them, which also drops the engine and connection so our ZCML declared events for `z3c.saconfig.interfaces.IEngineCreatedEvent` actually fire as intended upon the first `create_session()` of the next layer.

Closes https://github.com/4teamwork/opengever.core/issues/5104

### DIFF
--- a/opengever/core/sqlite_testing.py
+++ b/opengever/core/sqlite_testing.py
@@ -6,6 +6,7 @@ from plone.testing import Layer
 from sqlalchemy.pool import StaticPool
 from z3c.saconfig import EngineFactory
 from z3c.saconfig import GloballyScopedSession
+from z3c.saconfig import scopedsession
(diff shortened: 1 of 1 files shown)


lukasgraf:
If you can provide some explanation for that feel free to self merge at a later point, code LGTM otherwise.

lukasgraf:
Nice, 💚 the docstring. It's actually surprising that this hasn't caused more issues.

PR #5255, opened 2019-01-09
Add new meeting toc options sorted by reference number and only ad hoc agendaitems

### LINKED ISSUE #5176: Inhaltsverzeichnis Sortierung anhand Aktenzeichen der OS-Positionen
Zusätzlich zu den bestehenden Inhaltsverzeichnissen `AlphabeticalToc` und `RepositoryBasedTOC` sollen zwei zusätzliche Inhaltsverzeichnisse gebaut werden:

- Gruppierung nach Positionsnummer der Ursprungs-Ordnungsposition
- Gruppierung nach Aktenzeichen des Ursprungs-Dossiers

Bei beiden Inhaltsverzeichnissen werden Ad-Hoc Anträge separat ohne Aktenzeichen aufgelistet.

Bei Anträgen ist das Aktenzeichen des Dossiers als Column `Proposal.dossier_reference_number` verfügbar. Daraus kann die Nummer der Ordnungsposition abgeleitet werden.

Sortierung mit "Human-Sorting", also "AB 1.2.3 / 3" vor "AB 1.11.3 / 7", siehe https://nedbatchelder.com/blog/200712/human_sorting.html.

njohner:
We add two new toc types for meeting periods:
* only agendaitems with proposals, ordered by dossier reference number
* only ad hoc agendaitems, ordered by titles

I'm not quite sure whether this is what was wanted here? The proposed solution is only partially satisfactory as it relies on a second entry in the TOC json `toc_adhoc`, which can be used in the sablon template to have a different listing type (without displaying the reference number) for adhoc agendaitem elements. I modified the template accordingly, see below for example TOCs:

**Only agendaitems with proposals, ordered by reference number**
screen shot 2019-01-09 at 13 36 22

**Only ad hoc agendaitems, ordered by title**
screen shot 2019-01-09 at 13 36 34

Another solution could be to have two different templates. Or maybe only one new TOC type, with two listings, one for agendaitems with proposal and one for ad hoc?

resolves #5176

### DIFF
--- a/opengever/meeting/browser/toc.py
+++ b/opengever/meeting/browser/toc.py
@@ -3,6 +3,8 @@ from opengever.meeting.command import MIME_DOCX
 from opengever.meeting.exceptions import SablonProcessingFailed
 from opengever.meeting.sablon import Sablon
 from opengever.meeting.toc.alphabetical import AlphabeticalToc
+from opengever.meeting.toc.dossier_refnum import DossierReferenceNumberBasedTOC
+from opengever.meeting.toc.repository_refnum import RepositoryReferenceNumberBasedTOC
 from opengever.meeting.toc.repository import RepositoryBasedTOC
 from plone import api
 from plone.i18n.normalizer.interfaces import IIDNormalizer
@@ -98,6 +100,46 @@ class DownloadRepositoryTOC(DownloadAlphabeticalTOC):
         return u"{}.docx".format(
             translate(_(u'filename_repository_toc',
                         default=u'Repository Toc ${period} ${committee}',
+                        mapping={
+                          'period': period_title,
+                          'committee': committee_title,
+                        }),
+                      context=getRequest()))
+
+
+class DownloadDossierReferenceNumberTOC(DownloadAlphabeticalTOC):
+
+    def get_data(self):
+        return DossierReferenceNumberBasedTOC(self.model).get_json()
+
+    def get_filename(self):
+        normalizer = getUtility(IIDNormalizer)
+        period_title = normalizer.normalize(self.model.title)
+        committee_title = normalizer.normalize(self.model.committee.title)
+
+        return u"{}.docx".format(
+            translate(_(u'filename_dossier_reference_number_toc',
+                        default=u'Dossier Reference Number Toc ${period} ${committee}',
+                        mapping={
+                          'period': period_title,
+                          'committee': committee_title,
+                        }),
+                      context=getRequest()))
+
+
+class DownloadRepositoryReferenceNumberTOC(DownloadAlphabeticalTOC):
+
+    def get_data(self):
+        return RepositoryReferenceNumberBasedTOC(self.model).get_json()
+
+    def get_filename(self):
+        normalizer = getUtility(IIDNormalizer)
+        period_title = normalizer.normalize(self.model.title)
+        committee_title = normalizer.normalize(self.model.committee.title)
+
+        return u"{}.docx".format(
+            translate(_(u'filename_repository_reference_number_toc',
+                        default=u'Repository Reference Number Toc ${period} ${committee}',
                         mapping={
                           'period': period_title,
                           'committee': committee_title,
--- a/opengever/meeting/tests/test_toc.py
+++ b/opengever/meeting/tests/test_toc.py
@@ -10,12 +10,57 @@ from opengever.core.testing import OPENGEVER_FUNCTIONAL_MEETING_LAYER
 from opengever.meeting.command import MIME_DOCX
 from opengever.meeting.toc.alphabetical import AlphabeticalToc
 from opengever.meeting.toc.repository import RepositoryBasedTOC
+from opengever.meeting.toc.utils import first_title_char
+from opengever.meeting.toc.utils import normalise_string
 from opengever.testing import FunctionalTestCase
 from z3c.relationfield.relation import RelationValue
 from zope.component import getUtility
 from zope.intid.interfaces import IIntIds
 import pytz
 import transaction
+import unittest
+
+
+class TestFirstTitleChar(unittest.TestCase):
+
+    def test_returns_the_first_character_of_title_key(self):
+        item = {"title": u"test title", "other": u"another one"}
+        self.assertEqual(first_title_char(item), u"t")
+
+    def test_supports_plain_string(self):
+        item = {"title": "test title"}
+        self.assertEqual(first_title_char(item), u"t")
+
+    def test_returns_lower_case(self):
+        item = {"title": u'A test'}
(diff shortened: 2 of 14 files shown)


opengever/meeting/toc/reference_number.py
```diff
@@ -0,0 +1,48 @@
+from itertools import groupby
+from opengever.meeting.model import AgendaItem
+from opengever.meeting.toc.alphabetical import AlphabeticalToc
+import re
+
+
+def alphanum_key(string):
```
deiferni: i'd like to have unit tests for just these functions. no plone at all should be necessary, just a plain `UnitTest`.

opengever/meeting/sablontemplate.py
```diff
@@ -12,7 +12,7 @@
 
 VALIDATION_DATA = {
     'meeting': SAMPLE_MEETING_DATA,
-    'toc': {'toc': []},
+    'toc': {'toc': [], 'toc_adhoc': []},
```
deiferni: as you already mention this is not satisfying as we would need to modify the templates. can't you just prepend the items to `toc`?

opengever/meeting/toc/ad_hoc.py
```diff
@@ -0,0 +1,47 @@
+from itertools import groupby
+from opengever.meeting.model import AgendaItem
+from opengever.meeting.toc.alphabetical import AlphabeticalToc
+
+
+def first_title_char(item):
```
deiferni: as with the other toc a `UnitTest` would be extra nice 😄.

opengever/meeting/browser/toc.py
```diff
@@ -124,3 +125,23 @@ def get_filename(self):
                           'committee': committee_title,
                         }),
                       context=getRequest()))
+
+
+class DownloadAdHocAgendaitemTOC(DownloadAlphabeticalTOC):
```
deiferni: i'm sorry the spec in the issue was garbled. the ad-hoc part should apply to both new tocs. please see https://github.com/4teamwork/opengever.core/issues/5176, i've updated it. there should be the following two tocs, 

- Gruppierung nach Positionsnummer der Ursprungs-Ordnungsposition
- Gruppierung nach Aktenzeichen des Ursprungs-Dossiers

in both of them you should filter out the ad-hocs and display them first, without a reference-number, as none exists.

njohner:
Ok I think I've gone a bit overboard here 😓. The code for `AlphabeticalToc`, which was used as baseclass for the `RepositoryBasedTOC` was a bit messy, with some unnecessary sorting operations (I think they were leftovers from before it was realised that the items needed to be sorted before the `groupby` statement) and the organization of the code made it not too easy to use as baseclass for other TOC classes. I therefore:
* refactored the `AlphabeticalToc` class a bit, moved `item_sort_key`and `group_by_key` to methods on the class, so that they can be easily overwritten. This allows to really just overwrite the sorting and grouping keys instead of the whole sorting and grouping methods.
* Moved some of the helper functions into a new `utlis.py` file
* adapted `RepositoryBasedTOC` to the new `AlphabeticalToc` baseclass 
* Added string normalisation before sorting to avoid things like having two different groups for titles starting with `A` or with `Ä`, sorting `Ä` after `Z`, sorting `A` after `z` and so on.
* Added the two new TOC classes (also based on `AlphabeticalToc`)
* Improved testing by adding `UnitTest`s for the helper functions 
* Added a group title for `Ad hoc proposals` in the `RepositoryBasedTOC` and the two new TOCs

**For example before in the `AlphabeticalToc` we had two different groups for `A` and `Ä`, now we have only one, and sorted as expected:**
alphabetical_toc

**In the `RepositoryBasedTOC` we would sort first all Upper case, then lower case, and then special characters:**
repo_toc_old
**Now we instead disregard case in sorting and normalise the special characters**
repo_toc_new

The new TOCs, there is one grouped by dossier reference number and one by repository number:
**grouped by repository number, sorting inside the group by title**
screen shot 2019-01-21 at 09 42 35
**grouped by dossier reference number**
screen shot 2019-01-21 at 09 42 45

opengever/meeting/tests/test_toc.py
```diff
@@ -10,12 +10,57 @@
...
+
+    def test_supports_plain_string(self):
+        string = "test"
+        self.assertEqual(normalise_string(string), u"test")
+
+    def test_returns_lower_case(self):
+        string = u'A tEst'
+        self.assertEqual(normalise_string(string), u'a test')
+
+    def test_removes_umlauts(self):
```
deiferni: not sure if we should do this. how does a swiss phonebook sort, or the libary 🤔? what about the "french" special chars.

opengever/meeting/toc/utils.py
```diff
@@ -15,3 +16,30 @@ def normalise_string(string):
...
+    """
+    if string is None:
+        return []
+    return [int(block) if block.isdigit() else normalise_string(block)
+            for block in re.split('([0-9]+)', string)]
+
+
+def repo_refnum(item):
+    if item['dossier_reference_number'] is not None:
+        return item['dossier_reference_number'].split("/")[0].strip()
```
deiferni: instead of hardcoding the `"/"` you should ask the current `IReferenceNumberFormatter` for its `repository_dossier_seperator`. theoretically it could be different between formatters.

njohner: Oh did not realize that...

deiferni: Good point, yes if it can be reused we should attempt to do so. IIRC it is tied closely to the plone content objects which we do not have access to at that moment. We can discuss it, if you want!

opengever/meeting/tests/test_toc.py
deiferni: as discusssed sort chars with diacritics like `ä` after their "base" char like `a`.

njohner:
Ok all done, though not 100% happy yet.
* I'm now using the sorter from `IReferenceNumberFormatter` to sort toc entries
  * This fixed sorting in the toc when using the `GroupedByThreeReferenceFormatter`, as there, simply casting to integers for the repository reference number does not give correct sorting.
  * Note that the sorter for `GroupedByThreeReferenceFormatter` ignores the `client_id`. This is not ideal in my opinion, as meetings can span several `client_id`s?
  * For the `RepositoryBasedTOC`, it seems that the repository titles are stored translated on the brains, so that if two agendaitems are generated from different languages, they will end up in different groups

I think the code for the `IReferenceNumberFormatter` would benefit from some refactoring. There seems to be a lot of duplicated code.

I think we could make an issue for a future release to refactor the `IReferenceNumberFormatter`s.

Apart from that, I think this is ready to merge.

PR #5256, opened 2019-01-09
Provide solr for local development.

### LINKED ISSUE #5168: Enable solr for local development by default
We're going to enable solr on all our productive deployments in the future, so it makes sense to develop and test on a solr enabled GEVER.

njohner:
We include solr in `development.cfg`, making it available for local development, but we do not activate it by default. To activate run `bin/instance run src/opengever.maintenance/opengever/maintenance/scripts/activate_solr.py`

To facilitate the starting and stopping of solr and making sure the correct instance of solr is running for a given instance of gever, we provide a script which checks if the port used by solr is already in use and only starts solr and the gever instance if it isn't. When exiting gever, solr will also be stopped.

After installing a fresh GEVER, everything works fine without using solr. But strangely during deployment of the example content, errors of the type
```
2019-01-09 14:03:51 ERROR ftw.solr.connection Update command failed. An exception occured: [Errno 61] Connection refused
2019-01-09 14:03:51 ERROR ftw.solr.connection Solr response error. An exception occured: [Errno 61] Connection refused
``` 
are thrown, although solr should not be active at that point. Not sure what to do about this (although as said, it does not prevent the deployment and does not seem to lead to any problems afterwards).

resolves #5168

### DIFF
--- a/development.cfg
+++ b/development.cfg
@@ -6,6 +6,8 @@ extends =
     https://raw.github.com/4teamwork/ftw-buildouts/master/bumblebee.cfg
     sphinx.cfg
     https://raw.githubusercontent.com/4teamwork/ftw-buildouts/master/format-xml.cfg
+    https://raw.githubusercontent.com/4teamwork/gever-buildouts/master/solr.cfg
+    start_all.cfg
 
 always-checkout = false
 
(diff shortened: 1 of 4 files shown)


njohner:
Ok this should be ready.
We now provide a shell script `bin/start_all`, which is generated during buildout to get the `solr-port` from buildout. This allows to have different instances of GEVER running at the same time on different ports, each with a solr also running on different ports.

If we use this script tot start gever, then deploying the example content will not throw any solr errors anymore, as solr will be running.
Of course if we want to use solr for the listings and such, we still need to run 
`bin/instance run src/opengever.maintenance/opengever/maintenance/scripts/activate_solr.py`

PR #5262, opened 2019-01-11
ResolveOGUIDView: Preserve query string.

lukasgraf:
This change makes sure that any additional **query string parameters** passed to the `resolve_oguid` view **are preserved**, for internal redirects as well as redirects to remote admin units.

This is to ease frontend development of favorites.

### DIFF
--- a/opengever/base/browser/resolveoguid.py
+++ b/opengever/base/browser/resolveoguid.py
@@ -4,6 +4,8 @@ from opengever.ogds.base.utils import get_current_admin_unit
 from opengever.ogds.base.utils import ogds_service
 from Products.CMFCore.utils import getToolByName
 from Products.Five import BrowserView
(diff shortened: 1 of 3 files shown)


lukasgraf:
@buchi rebased (/cc @all4manu)

opengever/base/browser/resolveoguid.py
```diff
@@ -1,9 +1,12 @@
 from AccessControl.SecurityManagement import SpecialUsers
+from collections import OrderedDict
```
buchi: Import is no longer needed

lukasgraf: @buchi D'oh. Fixed and rebased.

lukasgraf:
Self-merge since only trivialities remaining...

PR #5263, opened 2019-01-15
Refactor solr livesearch reply

### LINKED ISSUE #5018: Solr livesearch_reply needs some 💛
Livesearch reply for solr based deployments needs some 💛. Currently it is a copy of `livesearch_reply.py` from the skins folder. Which means that there is no real template but rendering is implemented by appending strings to a list and then concatenating them in the end as the output string. Thus escaping and charset conversion has to be implemented by the programmer which is error prone (and has already caused issues, see https://github.com/4teamwork/opengever.core/issues/5016). Also the template is not readable and thus hard to modify.

- To be able to modify the output quicker and also to understand faster what the view actually does and how results are rendered it would make much sense to use a real template here. 
- Also it is not tested, some tests documenting and ensuring basic functionality would be a good idea.

njohner:
Livesearch reply with activated solr was overwritten by a browser view, but copying the rendering from the `livesearch_reply.py` from the skins folder, so that no template was used for the view. Instead html was generated from the python code by concatenating strings, making the code hard to read and modify. 
* We refactor the `LiveSearchReplyView` to use a template
* We add tests for the `LiveSearchReplyView` (which was untested until now)
* We correct a small xml bug in the standard `livesearch_reply` (no solr)
* We add a test for when the standard `livesearch_reply` returns no results.

To allow more flexible testing we modify the `IntegrationTestCase.mock_solr` method to also accept a response dictionary as argument instead  of an asset filename. This allows to more easily modify the response in the tests.

resolves #5018

### DIFF
--- a/opengever/base/browser/livesearch.py
+++ b/opengever/base/browser/livesearch.py
@@ -1,4 +1,4 @@
-# Copied over from skins/livesearch_reply.py
+# Adapted from skins/livesearch_reply.py
 from ftw.solr.interfaces import ISolrSearch
 from ftw.solr.query import escape
 from ftw.solr.query import make_query
@@ -9,7 +9,7 @@ from Products.CMFPlone.browser.navtree import getNavigationRoot
 from Products.CMFPlone.utils import normalizeString
 from Products.CMFPlone.utils import safe_unicode
 from Products.Five.browser import BrowserView
-from Products.PythonScripts.standard import html_quote
+from Products.Five.browser.pagetemplatefile import ViewPageTemplateFile
 from Products.PythonScripts.standard import url_quote_plus
 from zope.component import getUtility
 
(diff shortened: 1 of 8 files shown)


njohner:
⚠️ something seems to be leaking in the tests. Can't reproduce it locally.

opengever/base/browser/livesearch.py
```diff
@@ -82,65 +79,41 @@ def write(s):
...
+        self.show_more = {}
+
+        self.legend = ts.translate(legend_livesearch, context=self.request)
+        self.nothing_found = ts.translate(label_no_results_found, context=self.request)
+        self.advanced_search_url = portal_url + '/advanced_search?SearchableText=%s' % url_quote_plus(self.search_term)
+        self.advanced_search_label = ts.translate(label_advanced_search, context=self.request)
+
+        for result in results:
+
+            itemUrl = result.getURL()
```
deiferni: please use python naming conventions

opengever/base/browser/livesearch.py
```diff
@@ -82,65 +79,41 @@ def write(s):
...
+            if result.portal_type in useViewAction:
+                itemUrl += '/view'
+
+            itemUrl = itemUrl + u'?SearchableText=%s' % url_quote_plus(self.search_term)
+
```
deiferni: i'll leave it up to you but i'd prefer to have this extra white space between those lines removed. imho it does not help making the code more readable.

PR #5269, opened 2019-01-16
Also hide re-risen Unauthorized tracebacks for non-manager users.

### LINKED ISSUE #5191: Avoid tracebacks in response
The following request results in a response that contains a traceback in the body:
`/portal_factory/File/1/edit/@@plone_lock_operations/safe_unlock?_=1`

Rotonen:
Return an empty user-facing traceback for non-manager users.
This is done as `Unauthorized` is explicitly rerisen for PAS plugins and this ends up bypassing the `opengever.base` exception template and falling back to normal exception rendering mechanisms.

Additionally we always display the tracebacks in debug mode or with verbose security enabled.

Adding tests for this one would be quite a chore as one would have to create a new testing layer per combination of ZServer settings and the traversal driver on the testbrowser always follows redirects and the issue we have is on the body of the redirect response. Thus I'm requesting manual validation.

We're hitting this logic (relevant for PAS challenge plugins) in https://github.com/zopefoundation/Zope/commit/7b65910f826130e456c32fd33a0ccec26687e324 and monkey patching the exception formatting logic was the least impactful route I found of addressing this.

An alternative would be to try to muck around with the exception hook itself, but that'd be potentially a lot more dangerous.

https://github.com/zopefoundation/Zope/blob/d916c812bdaf518053b0c3cb2cb3545ff73bc288/src/ZPublisher/Publish.py#L336

Also the exception rendering code is peppered a bit all around the place, depending on exact circumstance, but all of them I've managed to find use the `zExceptions.ExceptionFormatter`.

Original implementation:
https://github.com/zopefoundation/zExceptions/blob/85343157fa49f04b5304f8394d27400804b292df/src/zExceptions/ExceptionFormatter.py#L178-L196

Closes #5191

### DIFF
--- a/opengever/base/monkey/patches/exception_formatter.py
+++ b/opengever/base/monkey/patches/exception_formatter.py
@@ -1,0 +1,41 @@
+from opengever.base.monkey.patching import MonkeyPatch
+
+
+class PatchExceptionFormatter(MonkeyPatch):
+    """Return an empty user-facing traceback for non-manager users.
+
+    This is done as ``Unauthorized`` is explicitly rerisen for PAS plugins and
+    this ends up bypassing the ``opengever.base`` exception template and
+    falling back to normal exception rendering mechanisms.
+    """
+
+    def __call__(self):
+        from plone import api
+
+        def formatException(self, etype, value, tb, limit=None):
+            # The block within this if is the original implementation
+            if api.user.has_permission("Manage Portal", obj=api.portal.get()):
+                # The next line provides a way to detect recursion.
+                __exception_formatter__ = 1
+                result = [self.getPrefix() + "\n"]
+                if limit is None:
+                    limit = self.getLimit()
+                n = 0
+                while tb is not None and (limit is None or n < limit):
+                    if tb.tb_frame.f_locals.get("__exception_formatter__"):
+                        # Stop recursion.
+                        result.append("(Recursive formatException() stopped)\n")
+                        break
+                    line = self.formatLine(tb)
+                    result.append(line + "\n")
+                    tb = tb.tb_next
+                    n = n + 1
+                exc_line = self.formatExceptionOnly(etype, value)
+                result.append(self.formatLastLine(exc_line))
+                return result
+
+            # We fall through in our patch with a notification of a suppression
+            return ["Exception suppressed."]
+
+        from zExceptions.ExceptionFormatter import TextExceptionFormatter
+        self.patch_refs(TextExceptionFormatter, "formatException", formatException)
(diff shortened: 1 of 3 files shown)


opengever/base/monkey/patches/exception_formatter.py
```diff
@@ -0,0 +1,48 @@
...
+
+    Additionally we always display the tracebacks in debug mode or with verbose
+    security enabled.
+    """
+
+    def __call__(self):
+        from App.config import getConfiguration
+        from plone import api
+
+        def formatException(self, etype, value, tb, limit=None):
```
lukasgraf: Please provide a link in the PR description to the original implementation that is being patched here (in the actual version currently in use by GEVER) so that I can compare them side-by-side.

opengever/base/monkey/patches/exception_formatter.py
```diff
@@ -0,0 +1,48 @@
...
+            config = getConfiguration()
+            debug_mode = getattr(config, "debug_mode", False)
+            verbose_security = getattr(config, "verbose_security", False)
+            configured_verbosity = debug_mode or verbose_security
```
lukasgraf: Please indicate (using comments) directly in the code here which parts are the unchanged original implementation, and which is the delta. (For example like [this](https://github.com/4teamwork/opengever.core/blob/master/opengever/base/monkey/patches/cmf_catalog_aware.py#L108-L110) or [this](https://github.com/4teamwork/opengever.core/blob/master/opengever/base/monkey/patches/verify_object_paste.py#L73-L80), or any other way that makes it immediately obvious what the change is).

opengever/base/monkey/patches/exception_formatter.py
```diff
@@ -0,0 +1,48 @@
...
+    security enabled.
+    """
+
+    def __call__(self):
+        from App.config import getConfiguration
+        from plone import api
+
+        def formatException(self, etype, value, tb, limit=None):
+            config = getConfiguration()
+            debug_mode = getattr(config, "debug_mode", False)
```
lukasgraf: > Additionally we always display the tracebacks in debug mode or with verbose security enabled.

Hmm. Why? Just checking for the `Manage Portal` permission is enough, IMHO.

lukasgraf:
@Rotonen can you update this one please? Then we can ship it I think.

opengever/base/monkey/patches/exception_formatter.py
Rotonen: Here is the original (also amending the PR body):
https://github.com/zopefoundation/zExceptions/blob/85343157fa49f04b5304f8394d27400804b292df/src/zExceptions/ExceptionFormatter.py#L178-L196

opengever/base/monkey/patches/exception_formatter.py
Rotonen: Figured that'd be a thing for local development and foreground instances, but sure, dropping the extra logic.

Adding a suppression marker string into the stack trace instead so we see if we ever encounter this in a way we'd like to revisit the idea.

Rotonen:

PR #5280, opened 2019-01-21
Avoid naming conflicts in meeting zipexport

### LINKED ISSUE #5192: Zip-Export naming conflict.
If two or more attachments of a proposal have the same title/filename, or if the proposal document has the same title/filename as an attachment this causes a naming conflict within zip-export and grimlock import. Grimlocks zip import then always imports the first file.

Possible solutions:
- put attachments into an additional folder (Anhänge) + prefix them with a number (preferred). I think we could just change the paths in the metadata `.json` inside the zip and it should work fine even if there are additional folders, but i'm not quite certain. @phgross would this work in grimlock or do we need to keep the existing folder structure? 
- don't add a folder but prefix all files with an id and a type prefix (Antrag or Anhang)

⛑ **workaround**: rename attachment documents and/or proposal documents so that there are no more conflicts

njohner:
When a proposal has several attachments with the same title/filename, or if the proposal document has the same title/filename as an attachment, this leads to a naming conflict in the zip file and in the grimlock zip-import. To avoid that, we add a new folder `attachments` for each proposal, and within that folder we prefix the attachments with a number.

Also note that in the first commit I removed an unnecessary check in the `MeetingTraverser`. Indeed in `_get_agenda_item_document` we do not need to check whether the agendaitem has a document, as `AgendaItem.resolve_document` already returns `None` when if the agenda_item does not have a document.

resolves #5192

### DIFF
--- a/opengever/meeting/traverser.py
+++ b/opengever/meeting/traverser.py
@@ -36,14 +36,19 @@ class MeetingTraverser(object):
             self.traverse_agenda_item_document(
                 agenda_item_doc, agenda_item)
 
+        attachment_number = 0
         for attachment in self._get_agenda_item_attachments(agenda_item):
             if attachment:
+                attachment_number += 1
                 self.traverse_agenda_item_attachment(
-                    attachment, agenda_item)
+                    attachment, agenda_item, attachment_number)
 
+        excerpt_number = 0
         for excerpt in self._get_agenda_item_excerpts(agenda_item):
             if excerpt:
-                self.traverse_agenda_item_excerpt(excerpt, agenda_item)
+                excerpt_number += 1
+                self.traverse_agenda_item_excerpt(
+                    excerpt, agenda_item, excerpt_number)
 
     def _get_protocol_document(self):
         if not self.meeting.has_protocol_document():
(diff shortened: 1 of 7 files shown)


opengever/meeting/zipexport.py
```diff
@@ -67,6 +67,18 @@ def get_agenda_item_filename(self, document, agenda_item_number):
...
         )
 
+    def get_agenda_item_attachment_filename(self, document, agenda_item_number, attachment_number):
+        return normalize_path(u'{}/{}/{}_{}'.format(
+            translate(
+                _(u'title_agenda_item', default=u'Agenda item ${agenda_item_number}',
+                  mapping={u'number': agenda_item_number}),
+                context=getRequest(),
+                ),
+            'attachments',
```
deiferni: i would prefer if we translate the folder name, i.e. "Anhang" in german. Users may extract the zip manually.

opengever/meeting/traverser.py
```diff
@@ -36,14 +36,19 @@ def traverse_agenda_item(self, agenda_item):
             self.traverse_agenda_item_document(
                 agenda_item_doc, agenda_item)
 
+        attachment_number = 0
         for attachment in self._get_agenda_item_attachments(agenda_item):
```
deiferni: you could use `enumerate` if you want 😉

njohner: Not really, because `attachment` can be `None` in some cases, and I did not want this to count in the numbering of the attachments...

deiferni: ah 👍

njohner:
Ok all done.
I also corrected an error in the french translation for agendaitem titles, which were using the wrong variable name, and hence it would not get replaced.

PR #5289, opened 2019-01-24
Fix logo upload in the theme control-panel.

phgross:
The upload was broken and ended up in an AttributeError and a cryptic error message gets
displayed in browser.


Rotonen:
Seems browser caching got me, it was green, now it is not.

deiferni:
@phgross this still 💥😞

phgross:
Tests are green now ...

PR #5293, opened 2019-01-25
Bump Ruby version to 2.4.5

lvonlanthen:
While installing gever on my machine today, I stumbled upon the wrong Ruby version in the documentation. Also, I had to activate the correct Ruby version after installing it.

The ruby version in the documentation now matches the one specified in the `Gemfile`.

On a side note: I also had to use `rbenv exec gem install bundler` instead of only `gem install bundler`, but maybe that is only because of my setup and does not need to be documented?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-ruby '2.4.2'
+ruby '2.4.5'
 gem 'sass', '~> 3.4'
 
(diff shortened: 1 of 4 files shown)


Rotonen:
@lvonlanthen I've started a process to put everything on 2.4.5 - I'll ping you back when that lands.

Rotonen:
And we are there. Everything can go onto 2.4.5 now.

Mind the Gemfile and the lockfile thereof for sass-watcher.

PR #5297, opened 2019-01-25
Scrub HTTP response headers

### LINKED ISSUE #5185: Clean Up Response Headers
Currently some responses return HTTP headers that are not optimal in terms of security.

**Cache-Control**
Some responses do not contain a Cache-Control header (e.g. file_download_confirmation, breadcrumb_by_uid, tooltip). Set the header if it's missing.
For sensitive content:
`Cache-Control: no-cache, no-store, must-revalidate`
For assets:
`Cache-Control: public, max-age=31536000`

**Remove headers that disclose unnecessary information**
E.g.
- `Bobo-Exception-File`
- `Bobo-Exception-Type`

**Remove server version**
instead of `Server: Zope/(2.13.28, python 2.7.14, linux2) ZServer/1.1`
return `Server: Zope`

**Content-Security-Policy**
Define and implement an appropriate CSP.
E.g. only allow content from the site's own origin: `Content-Security-Policy: default-src 'self'`
If we have inline JS, can we avoid it? If not, can we allow it?

Implementation:
Implement a transform chain (plone.transformchain) that adds/removes/cleans "security"-related headers.
The transform chain should be called as late as possible (after plone.protect, order=9000, maybe even after plone.caching, order=12000).

Rotonen:
* Bumped `ftw.testbrowser` to allow for not following redirects to test 302 response headers and bodies
* Added an even listener for `ZPublisher.interfaces.IPubEnd` to scrub the server version string
* Monkey patched setting Bobo Call Interface headers out
* Also added a test for the recently added traceback scrub from https://github.com/4teamwork/opengever.core/pull/5269

Original implementation of setting Bobo Call Interface headers onto responses:
https://github.com/zopefoundation/Zope/blob/d916c812bdaf518053b0c3cb2cb3545ff73bc288/src/ZPublisher/HTTPResponse.py#L758-L787

Closes #5185


opengever/base/monkey/patches/scrub_bobo_exceptions.py
```diff
@@ -0,0 +1,13 @@
+from opengever.base.monkey.patching import MonkeyPatch
+
+
+class ScrubBoboExceptions(MonkeyPatch):
+    """We do not use the Bobo exception headers for anything."""
+
+    def __call__(self):
+        def _setBCIHeaders(self, t, tb):
```
deiferni: i'm quite unhappy about the continuous use of monkey patches 😞 . we should use them as a last resort only. You might have already considered this but just to be sure, is there an event-handler you could register instead of patching and then scrub the headers again? Maybe `IPubFailure` would work?

Rotonen:
I've dropped the added test for the now-reverted rerisen error body scrubbing.

Ready for rereview.

PR #5306, opened 2019-01-29
Also hide re-risen Unauthorized tracebacks for non-manager users

Rotonen:
Redo of https://github.com/4teamwork/opengever.core/pull/5269

Return an known-redacted user-facing traceback for non-manager users.
This is done as Unauthorized is explicitly rerisen for PAS plugins and this ends up bypassing the opengever.base exception template and falling back to normal exception rendering mechanisms.

Additionally we always display the tracebacks in tests, debug mode or with verbose security enabled.

Adding tests for this one do not seem possible.

We're hitting this logic (relevant for PAS challenge plugins) in zopefoundation/Zope@7b65910 and monkey patching the exception formatting logic was the least impactful route I found of addressing this.

An alternative would be to try to muck around with the exception hook itself, but that'd be potentially a lot more dangerous.

https://github.com/zopefoundation/Zope/blob/d916c812bdaf518053b0c3cb2cb3545ff73bc288/src/ZPublisher/Publish.py#L336

Also the exception rendering code is peppered a bit all around the place, depending on exact circumstance, but all of them I've managed to find use the zExceptions.ExceptionFormatter.

Original implementation:
https://github.com/zopefoundation/zExceptions/blob/85343157fa49f04b5304f8394d27400804b292df/src/zExceptions/ExceptionFormatter.py#L178-L196

Delta declarations:

* Patching the HTML formatter instead of the Text formatter
* Reintroduction of conditionals for when to suppress the traceback
* Addition of checking for if we are using demo storage

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -36,6 +36,7 @@ Changelog
 - Include portal_type in @favorites endpoint response. [lgraf]
 - Supply date format locale settings for fr-ch. [lgraf]
 - Add meeting error view displayed when user has permission issues. [njohner]
+- Also hide re-risen Unauthorized tracebacks for non-manager users. [Rotonen]
 - Kill the theme functional test layer. [Rotonen]
 - Kill the theme integration test layer. [Rotonen]
 - Merge the plonetheme.teamraum gever profile into opengever.core. [Rotonen]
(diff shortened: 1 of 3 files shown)


lukasgraf:
@Rotonen as discussed: Please update the PR body with a succinct delta of the changes as compared to #5269, and why this one should fix the issues we've had.

deiferni:
what is going on here? i think this requires a description update? is this ready otherwise?

Rotonen:
I've already amended the description, now rebased.

PR #5307, opened 2019-01-30
Fix bug in Excerpt overview with no permissions on meeting.

njohner:
In the excerpt overview in the meeting dossier, when a user had no permission on the meeting, the get_proposal method would fail trying to access the proposal instead of the submitted proposal.

This resolves https://extranet.4teamwork.ch/support/gemeinde-koeniz/tracker/136


njohner:
all done

PR #5309, opened 2019-01-30
It can be None apparently.

deiferni:
…


deiferni:
i saw that coming: https://github.com/4teamwork/opengever.core/pull/5207#discussion_r248352324

PR #5315, opened 2019-02-01
Provide poor man's cache invalidation mechanism  for vue apps.

### LINKED ISSUE #5180: Vue Apps are cached for 24h without invalidation.
At the moment we include `Vue` apps javascript directly without using the javascript registry, see https://github.com/4teamwork/opengever.core/blob/7c1a241a6053c7e654a5b9e2f8370ec3d7af9698/opengever/meeting/browser/meetings/templates/demand_zip.pt#L27 for an example. These files will be cached for 24h which causes issues after updates since the cache can't be invalidated they way the files are served.

Possible solutions:
-  preferred: find a way to switch to plones js registry to register the javascript file and maybe instanciate the Vue app in the template somehow? input from @bierik this could be solved as follows: (https://github.com/4teamwork/bernmobil.instructions/blob/3d31bf0a38947fc89cf269486bb1584674a9d6b3/frontend/src/main.js)
-  append a query-string parameter when including the file, increase manually when the file changes

Non-Exhaustive list of other affected apps (or their filenames):
- [x] recently_touched_menu_app.js
- [x] favorites_app.js
- [x] favorites_menu_app.js
- [x] task_reminder_selector.js
- [x] save_pdf_under_app.js
- [x] zip_export_app.js
- [x] app.js (participation)

deiferni:
Provide a query string parameter wherever we include vue apps directly in a template. This will invalidate all those apps, theoretically that would not be necessary now but only when the file changes next time. We do it now anyway so it serves as an example for future uses, also new files we include will then already have the version query parameter when the line is copied from an existing template. Also hopefully hints that the version must be increased manually whenever the javascript file changes. 

We should come up with something better soon, but until then this approach must do.

I have manually verified that the apps affected by this change still work 😓.

Also provide tests that make sure that the query string parameter is updated when the files change and that we do not introduce new includes without cache invalidation. Sample test output.

Query string parameter not updated correctly:
screenshot 2019-02-01 at 20 23 47

Missing parameter:
screenshot 2019-02-01 at 21 09 57

closes #5180

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2019.1.0 (unreleased)
 ---------------------
 
+- Add simple cache invalidation mechanism for javascript included in templates. [deiferni]
 - Fix bug in paragraph agendaitem item_number display. [njohner]
 - Group fields in Committee edit forms. [njohner]
 - Scrub Bobo Call Interface data out of the HTTP response headers. [Rotonen]
--- a/opengever/base/browser/templates/favorites.pt
+++ b/opengever/base/browser/templates/favorites.pt
@@ -24,7 +24,7 @@
                              data-userid view/get_userid;
                              data-authtoken context/@@authenticator/token"></div>
 
-        <script tal:attributes="src string:${here/portal_url}/++resource++opengever.base/favorites_app.js?_v=1"></script>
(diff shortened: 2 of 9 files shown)


Rotonen:
We could use a hash of the file as the query string and add assertions of that into the tests.

deiferni:
yeah i've been thinking about something along that line, too. kinda hard to do it in a generic way, though without things falling through. how to reliably find all the js includes in templates for example? you quickly miss some that use different markup.

deiferni:
Actually it is possible with minimal effort. I have amended the PR.

PR #5317, opened 2019-02-04
Add custom sortable_title indexer to avoid cropping of content titles.

### LINKED ISSUE #4398: Ordnungspositionen werden nicht korrekt sortiert.
[Extranet Issue](https://extranet.4teamwork.ch/support/gemeinde-koeniz/tracker/70)

Der SortableTitle Indexer (`Products.CMFPlone.CatalogTool.sortable_title`) limitiert den Inhalt des sortable_titles auf 40 Zeichen, durch den vierstelligen 0 prefix welcher bei allen Nummern hinzugefügt wird, führt dies zu einer Limite von nur 5 Nummern. Dies ist bei Ordnungspositionen zu knapp:

**Auszug Analyse Debugging:**
```
self.context.Title
'1.1.1.5.1.1. Dienstzweig Abfallbewirtschaftung und Deponie (AD)'
from Products.CMFPlone.CatalogTool import sortable_title
(Pdb) sortable_title(self.context)()
'0001.0001.0001.0005.0001.00...ponie (ad)'
```

### Lösung:
- [x] Einen separaten `sortable_title` indexer für Ordnungspositionen erstellen, welcher bei der Berechnung der `MAX_SORTABLE_TITLE` Limite, die Aktenzeichennummer nicht beachtet.
 - [ ] Upgradestep der den sortable_title indexer für alle Ordnungspositionen neu berechnet

njohner:
Add custom `sortable_title` indexer to avoid cropping of content titles. We set the maximal length for the `sortable_title` larger than the maximal allowed content title length, to allow for number padding without cropping long titles.

resolves #4398

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -9,6 +9,7 @@ Changelog
 - Skip creation of the tasks pdf on resolve for dossiers without tasks. [phgross]
 - Give View permission to Editors on mails. [njohner]
 - Clear role assignments on contained object after forwarding creation. [njohner]
+- Add custom sortable_title indexer to avoid cropping of content titles. [njohner]
 - Testserver: add support for custom fixtures. [jone]
 - Reindex and store additionally supported bumblebee documents. [elioschmutz]
 - Fix scrubbing the server version out of the HTTP response headers. [Rotonen]
--- a/opengever/base/indexes.py
+++ b/opengever/base/indexes.py
@@ -3,8 +3,14 @@ from opengever.base.behaviors.changed import IChangedMarker
 from opengever.base.behaviors.translated_title import ITranslatedTitle
 from opengever.base.behaviors.translated_title import ITranslatedTitleSupport
 from opengever.base.interfaces import IReferenceNumber
+from opengever.base.model import SORTABLE_TITLE_LENGTH
 from plone.dexterity.interfaces import IDexterityContent
 from plone.indexer import indexer
+from plone.i18n.normalizer.base import mapUnicode
+from Products.CMFPlone.CatalogTool import num_sort_regex
+from Products.CMFPlone.CatalogTool import zero_fill
+from Products.CMFPlone.utils import safe_callable
+from Products.CMFPlone.utils import safe_unicode
 from zope.annotation import IAnnotations
 from opengever.bundle.sections.constructor import BUNDLE_GUID_KEY
 
@@ -45,3 +51,32 @@ def changed_indexer(obj):
         # The indexer transforms this to UTC and then represents it as a integer
         return IChanged(obj).changed
     return None
+
+
+@indexer(IDexterityContent)
+def sortable_title(obj):
+    """Custom sortable_title indexer for all content types.
+
+    This is a copy from Products.CMFPlone.CatalogTool.sortable_title
+    except that we overwrite MAX_SORTABLE_TITLE with SORTABLE_TITLE_LENGTH.
+    We set set it to a high enough value to basically avoid ever cropping
+    the title, even with number padding. This is to avoid sorting issues
+    in content listings.
+    """
+    title = getattr(obj, 'Title', None)
+    if title is not None:
+        if safe_callable(title):
+            title = title()
+
+        if isinstance(title, basestring):
+            # Ignore case, normalize accents, strip spaces
+            sortabletitle = mapUnicode(safe_unicode(title)).lower().strip()
+            # Replace numbers with zero filled numbers
+            sortabletitle = num_sort_regex.sub(zero_fill, sortabletitle)
+            # Truncate to prevent bloat, take bits from start and end
+            if len(sortabletitle) > SORTABLE_TITLE_LENGTH:
+                start = sortabletitle[:(SORTABLE_TITLE_LENGTH - 13)]
+                end = sortabletitle[-10:]
+                sortabletitle = start + u'...' + end
+            return sortabletitle.encode('utf-8')
+    return ''
--- a/opengever/base/model/__init__.py
+++ b/opengever/base/model/__init__.py
@@ -13,6 +13,9 @@ DEFAULT_LOCALE = 'de'
 SUPPORTED_LOCALES = ['de', 'fr', 'en']
 
 CONTENT_TITLE_LENGTH = 255
+# Sortable title length is a bit larger than content title due to number
+# padding (we make room for 5 numbers separated by dots and then some margin).
+SORTABLE_TITLE_LENGTH = CONTENT_TITLE_LENGTH + 45
 CSS_CLASS_LENGTH = 100
 EMAIL_LENGTH = 255
 FIRSTNAME_LENGTH = 255
--- a/opengever/core/upgrades/20190204085532_increase_sortable_title_max_length/upgrade.py
+++ b/opengever/core/upgrades/20190204085532_increase_sortable_title_max_length/upgrade.py
@@ -1,0 +1,23 @@
+from collective.indexing.interfaces import IIndexQueueProcessor
+from collective.indexing.queue import getQueue
+from ftw.upgrade import UpgradeStep
+from plone.dexterity.interfaces import IDexterityContent
+from plone import api
+from zope.component import getUtility
+
+
+class InccreaseSortableTitleMaxLength(UpgradeStep):
+    """Increase maximal length of sortable title.
+    """
+
+    deferrable = True
+
+    def __call__(self):
+        getQueue().hook()
+        processor = getUtility(IIndexQueueProcessor, name='ftw.solr')
+        catalog = api.portal.get_tool('portal_catalog')
+        for obj in self.objects({'object_provides': IDexterityContent.__identifier__},
+                                "Reindex sortable title."):
+            catalog.reindexObject(obj, idxs=['sortable_title'],
+                                  update_metadata=False)
+            processor.index(obj, ['sortable_title'])
--- a/opengever/base/configure.zcml
+++ b/opengever/base/configure.zcml
@@ -351,6 +351,11 @@
       name="changed"
       />
 
+  <adapter
+      factory=".indexes.sortable_title"
+      name="sortable_title"
+      />
+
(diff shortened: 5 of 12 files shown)


opengever/core/upgrades/20190204085532_reindex_repository_sortable_title/upgrade.py
```diff
@@ -0,0 +1,17 @@
...
+
+class ReindexRepositorySortableTitle(UpgradeStep):
+    """Reindex repository sortable title.
+    To avoid sorting issues with deeply nested repository folders, we
+    had previously added a custom indexer, but the upgrade step to reindex
+    the sortable title had been forgotten.
+    """
+
+    def __call__(self):
+        repfolder_brains = api.content.find(object_provides=IRepositoryFolder)
```
lukasgraf: We almost always want to use `catalog.unrestrictedSearchResults()` during upgrade steps, to make sure we don't miss any objects. So generally I would advise against using `plone.api.content.find()` at all during upgrade steps - it contains more magic than I'm comfortable with in situations where we need to precisely control what we query.

opengever/core/upgrades/20190204085532_reindex_repository_sortable_title/upgrade.py
```diff
@@ -0,0 +1,17 @@
...
+        for repfolder_brain in repfolder_brains:
+            repfolder = repfolder_brain.getObject()
+            repfolder.reindexObject(idxs=['sortable_title'])
```
lukasgraf: This will also update metadata, which is expensive and not needed in this case (only the `sortable_title` index needs to be indexed). So use the method on the catalog tool, and the keyword argument `update_metadata=false`:

```python
catalog.reindexObject(obj, idxs=['sortable_title'], update_metadata=False)
```

lukasgraf: As pointed out by @njohner: Using `catalog.reindexObject` will circumvent Solr. And our solr schema does indeed have a `sortable_title`. So we need to specifically trigger Solr indexing, [e.g. like this](https://github.com/4teamwork/opengever.core/blob/master/opengever/base/handlers.py#L59-L65). (`processor.index(context, [`sortable_title`])` should do the trick).

deiferni: I think for past upgrades we have decided to use the interface provided by the objects only (despite the fact that medatada is updated as well), so it would be `object.reindexObject(idxs=['whatever'])` here. I'd probably stick to that decision 🤔.

njohner: I guess I'm fine with either. `object.reindexObject(idxs=['whatever'])` is safe in the sense that we are sure not to forget solr. `catalog.reindexObject` would be faster...

So apart from that, this is ready.

lukasgraf:
@njohner @deiferni [as discussed](https://github.com/4teamwork/opengever.core/issues/4398#issuecomment-461375986), we want to go a step further here and increase the `sortable_title` length alltogether.

njohner:
Ok this should be ready now

opengever/base/indexes.py
```diff
@@ -45,3 +51,31 @@ def changed_indexer(obj):
         # The indexer transforms this to UTC and then represents it as a integer
         return IChanged(obj).changed
     return None
+
+
+@indexer(IDexterityContent)
+def sortable_title(obj):
+    """Custom sortable_title indexer for all content types.
+
+    This is a copy from from Products.CMFPlone.CatalogTool.sortanle_title
```
deiferni: typo `sortanle_title`?

opengever/base/indexes.py
```diff
@@ -45,3 +51,31 @@ def changed_indexer(obj):
...
+        if isinstance(title, basestring):
+            # Ignore case, normalize accents, strip spaces
+            sortabletitle = mapUnicode(safe_unicode(title)).lower().strip()
+            # Replace numbers with zero filled numbers
+            sortabletitle = num_sort_regex.sub(zero_fill, sortabletitle)
+            # Truncate to prevent bloat, take bits from start and end
+            if len(sortabletitle) > MAX_SORTABLE_TITLE:
+                start = sortabletitle[:(MAX_SORTABLE_TITLE - 13)]
+                end = sortabletitle[-10:]
+                sortabletitle = start + '...' + end
```
deiferni: you are concatenating unicode and byte strings here. `sortabletitle` is an unicode string, whereas `'...'` is a bytestring. it will work in this example as the bytestring can be decoded but is bad practice in general. please use an unicode literal, i.e. `u'...'`.

opengever/base/model/__init__.py
```diff
@@ -13,6 +13,9 @@
 SUPPORTED_LOCALES = ['de', 'fr', 'en']
 
 CONTENT_TITLE_LENGTH = 255
+# Sortable title length is a bit larger than content title due to number
+# padding (we make room for 5 numbers separated by dots, and a space).
+MAX_SORTABLE_TITLE = CONTENT_TITLE_LENGTH + 26
```
deiferni: this is 281 _(ok, cool, a prime number!)_, but why not just 300 or something? that would be a bit easier to communicate to customers should they as for some reason and it does not seem so arbitrary. the few occasional extra bytes should not hurt 🤔. what do you think?

opengever/base/indexes.py
```diff
@@ -45,3 +51,31 @@ def changed_indexer(obj):
...
+        if safe_callable(title):
+            title = title()
+
+        if isinstance(title, basestring):
+            # Ignore case, normalize accents, strip spaces
+            sortabletitle = mapUnicode(safe_unicode(title)).lower().strip()
+            # Replace numbers with zero filled numbers
+            sortabletitle = num_sort_regex.sub(zero_fill, sortabletitle)
+            # Truncate to prevent bloat, take bits from start and end
+            if len(sortabletitle) > MAX_SORTABLE_TITLE:
```
deiferni: `MAX_SORTABLE_TITLE`**`_LENGTH`**?

opengever/base/indexes.py
njohner: a bit blindly copied from plone I guess...

opengever/base/model/__init__.py
njohner: Sure I don't mind :-) But I'm not sure we would communicate 300 to the customers anyway, in the sense that it's a title with 255 characters, among which 5 numbers. They can never set more than 255 characters for a title, then depending on how many numbers are in there, the actual part of their title that will be used as sortable_title will depend. For repositories this is a bit more complicated, because they set a title and then we add the reference number in the sortable title, so depending on the nesting depth, it's not the same length available for them...

opengever/base/indexes.py
njohner: I thought of changing this to make it more consistent with definitions `opengever.base.model`, but then the `sortable_title` would diverge more from the method from which it was copied. So not sure what is better.

njohner: Made it `SORTABLE_TITLE_LENGTH`, consistent with other variable names in `opengever.base.model`

njohner:
Ok this should be ready

opengever/base/indexes.py
```diff
@@ -45,3 +51,32 @@ def changed_indexer(obj):
         # The indexer transforms this to UTC and then represents it as a integer
         return IChanged(obj).changed
     return None
+
+
+@indexer(IDexterityContent)
+def sortable_title(obj):
+    """Custom sortable_title indexer for all content types.
+
+    This is a copy from from Products.CMFPlone.CatalogTool.sortable_title
```
lukasgraf: `from from` -> `from`

docs/HISTORY.txt
```diff
@@ -18,6 +18,7 @@ Changelog
 - Implement WebActions storage and REST API. [lgraf]
 - Fix changing task to or from private via edit form. [deiferni]
 - Fix creating private tasks by mistake. [deiferni]
+- Reindex sortable_title to fix sorting of repository folders. [njohner]
```
lukasgraf: Changelog entry should be updated to reflect the new change in functionality. (`Add custom sortable_title indexer to avoid cropping of content titles.`)

opengever/core/upgrades/20190204085532_increase_sortable_title_max_length/upgrade.py
```diff
@@ -0,0 +1,14 @@
+from ftw.upgrade import UpgradeStep
+from plone.dexterity.interfaces import IDexterityContent
+
+
+class InccreaseSortableTitleMaxLength(UpgradeStep):
+    """Inccrease maximal length of sortable title.
```
lukasgraf: `Inccrease` -> `Increase`

opengever/core/upgrades/20190204085532_increase_sortable_title_max_length/upgrade.py
```diff
@@ -0,0 +1,14 @@
...
+    """
+
+    deferrable = True
+
+    def __call__(self):
+        for obj in self.objects({'object_provides': IDexterityContent.__identifier__},
+                                "Reindex sortable title."):
+            obj.reindexObject(idxs=['sortable_title'])
```
lukasgraf: Mhh, this has the same issue again that this will update all metadata *for all objects*. See [previous comment](https://github.com/4teamwork/opengever.core/pull/5317#discussion_r253405803)

njohner: Yes I forgot to update the upgrade, sorry.

njohner: on my ~300 objects that I have locally, the upgrade took 12s, instead of 27 when reindexing also reindexing the metadata

njohner:
I hope I did not forget anything this time! Should be ready

lukasgraf:
Looks good to me now 👍 Feel free to self-merge once tests are done :)

PR #4648, opened 2018-07-27
Add placeful workflow policy for inboxes

phgross:
Currently inbox inbox group users are wrongly able to edit the inbox. Because those users should still be able to edit documents inside the inbox, we introduce a separate workflow for inbox_documents and adjust the inbox workflow:
 - Add a separate placeful workflow policy `opengever_inbox_workflow` enabled on inboxes and inbox-containers and below.
 - Separate inbox_document workflow which no longer acquires the permissions `Modify portal_content`, `Request review`, `opengever.document: Cancel`, `opengever.document: Checkin` and `opengever.document: Checkout`
 - Revoke `Modify_portal_content` for `Editors` and `Administrators on inboxes`
 - Upgradestep which enables the placeful_workflow policy for all inboxes or inbox-container and below.

Currently inbox users are not been able to Checkout/Checkin documents, but only edit metadata. This restriction does not make much sense in a functional view and there is also no technical reason  against. So the new inbox_document workflow also allow to checkout/checkin documents. 

Closes #1408

Besides the permissions adjustments, I found a bug in the `archival_file` and `public_trial` checks, and fixes those as well.

PR #5325, opened 2019-02-05
Add placeful workflow for mails in inbox area

### LINKED ISSUE #5198: Beiarbeiten der Metadaten bei E-Mail-Dateien im Eingangskorb nicht möglich
- Das Bearbeiten der Metadaten bei E-Mail-Dateien welche per D&D od. via Mail-In in den Eingangskorb von GEVER hochgeladen werden ist aktuell nicht mehr möglich -> Feld "Metadaten Bearbeiten" bleibt ausgegraut.

- Andere Dateien wie Word, Excel ect. hingegen sind davon nicht betroffen, sondern auschliesslich E-Mail-Dateien.

Getestet u. reproduziert unter -> https://dev.onegovgever.ch/fd/eingangskorb#documents-proxy
Ticket -> https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/484

njohner:
Mails cannot be edited in the inbox, because of the inbox workflow introduced in #4648, preventing inbox group users are to edit the inbox. A placeful workflow was added for documents to allow editing of documents in the inbox, but not for mails. We correct this here.

For the upgrade step, we cannot simply use the `PlacefulWorkflowPolicyActivator`. The problem is that the  `opengever_inbox_policy` is already active on the inbox and that we simply add a workflow for mails, so that as soon as we call `self.install_upgrade_profile()`, the objects in the inbox will be on the new workflow, and `WorkflowChainUpdater` called from `PlacefulWorkflowPolicyActivator` will not apply the state mapping. Hence we have to do this by hand.

For easier review I have a first commit simply copying the `opengever_mail_workflow` as a new placeful workflow, and then modifying it.

resolves #5198


opengever/core/upgrades/20190204133732_add_mail_placeful_worfklow_for_inbox_area/upgrade.py
```diff
@@ -0,0 +1,39 @@
...
+                     'opengever_inbox_mail_workflow'): {
+                         'mail-state-active': 'mail-state-active',
+                         'mail-state-removed': 'mail-state-removed'}
+                    }
+
+        with WorkflowChainUpdater(objects, review_state_mapping):
+            self.install_upgrade_profile()
+            for inbox in inboxes:
+                activator = PlacefulWorkflowPolicyActivator(inbox)
+                activator._activate_placeful_policy(
```
jone: I think you don't need a `PlacefulWorkflowPolicyActivator` at all, because you are not activating a new placeful policy, right? The `PlacefulWorkflowPolicyActivator` actually just enables a new placeful workflow policy and then uses the `WorkflowChainUpdater`. So this should be doable with only using `WorkflowChainUpdater`.

I would do it like this:
```python
from operator import methodcaller

inbox_paths = map(methodcaller('getPath'),
                  self.catalog_unrestricted_search(…, full_objects=False))

mails_in_inboxes = self.catalog_unrestricted_search(
    {'path': inbox_paths, portal_type='ftw.mail.mail'},
    full_objects=True)

with WorkflowChainUpdater(objects, review_state_mapping):
    self.install_upgrade_profile()
```

njohner: Thanks for the clarification

Issue #5331, opened 2019-02-05
Test server self test runs get stuck if XMLRPC port allocation fails


Rotonen:
https://ci.4teamwork.ch/builds/220341/tasks/362106

How do we allocate that port?

deiferni:
@jone can we have a look at this tomorrow. a quick temporary fix would also be welcome, maybe an elegantly placed `@skip` somewhere 😉.

jone:
Sorry for the trouble. I'll look into it.
My suspicion is that Jenkins' port allocator plugin does not play well with the Jenkins slave: the ports are probably allocated on master but used on the slave, so collisions cannot be avoided.

- Using port `0` does not really work because the selftest script starts the server in a subprocess but needs to know the port for connecting to.
- Using port `0` and printing the port would probably involve patching `plone.app.robotframework` or rewriting parts of it, I'd like to avoid that.
- Using a path based socket does not work with testbrowser, I guess.
- Looking into Jenkins is probably the best approach.

Rotonen:
@jone yes, this is the reason why I made logic in `plone.testing` for letting the operating system allocate a free port for the socket.

Monkey patch `plone.app.robotframework` to also parse the xmlrpc port off the test layer and store the port there on the layer setup.

PR #5336, opened 2019-02-05
Delete the testserver selftest test config

Rotonen:
After discussing with @deiferni, I'm making this PR for dropping the testserver selftest config for now as it's getting stuck in a nasty way at a bad time.

@jone you can simply revert this commit on a branch for further testing and then PR that back in. Or does that break history browsing and you'd prefer a commented out Jenkins test command in the config and a substituted `exit 0` instead?


deiferni:
I've just discussed it with jone, it is fine if we proceed and delete for now. The rest can be tackled with https://github.com/4teamwork/opengever.core/issues/5331.

PR #5344, opened 2019-02-07
Fix permission issues in forwardings

### LINKED ISSUE #5338: Cannot view mail in reassigned forwarding
When a mail is forwarded and then the forwarding is reassigned to a user with not permissions on the inbox, he will not be able to see the mail.

This is because Task reassignment grants the `Editor` role, which does not give `View` permissions on mails (but does on documents)

see https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/548

### LINKED ISSUE #5339: Role Assignments on objects inside a forwarding are not properly reset when reassigning
When a forwarding is created, role assignments are added on the `relatedItems` through the `set_roles_after_adding` event handler, just before the related items are moved into the forwarding and the `relatedItems` field is cleared by `move_documents_into_forwarding`. When the forwarding then gets reassigned, these local roles do not get updated as the `relatedItems` field is now empty.

Note that the event order is probably not fixed as they are both registered on `IObjectAddedEvent`

njohner:
There were two problems with forwardings:
- Mails inside a forwarding (or a task for that matter) were not necessarily accessible to the task_responsible, as the `Editor` role did not grant `View` permissions on a task
- There was an issue with the RoleAssignments during creation of forwardings, basically we use the `relatedItems` field during creation for the documents and mails that will be moved into the forwarding. Then we move them into the forwarding and clear the field, but roles get assigned to the related items of a task, and this happens on the same Event as moving the items, so that roles can get assigned and then will never get cleared anymore (by reassignment for example)

resolves #5339 
resolves #5338 
This also resolves https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/548


opengever/inbox/tests/test_workflow.py
```diff
@@ -84,3 +88,95 @@ def test_switch_to_regular_workflow_when_assign_to_dossier_via_forwarding(self,
...
+        self.assertEqual(["Editor"], assignment["roles"])
+        self.assertEqual(self.secretariat_user.getId(), assignment["principal"])
+
+    @browsing
+    def test_forwarding_grants_view_permissions(self, browser):
+        self.login(self.secretariat_user, browser)
+
+        mail = create(Builder('mail').within(self.inbox_forwarding))
+
+        self.assertIn("user:{}".format(self.regular_user),
```
njohner: Note that using `plone.api.user.has_permission` or `plone.api.user.get_permissions` does not work here, permission returned by the plone api are the same before and after reassigning the forwarding, even though `obj.__ac_local_roles` is up to date.
I've opened an issue for that: https://github.com/4teamwork/opengever.core/issues/5345

njohner:
Sorry @deiferni , I had to rebase and add a commit, now that there is a new placefull workflow for E-mails in inboxes. I kept the new commit separate for now. It's just adding View permission for Editors in the placefull workflow and the corresponding upgrade.

lukasgraf:
@njohner can you please rebase and push again?

PR #5349, opened 2019-02-07
Update docproperties when updating an agendaitem list or a protocol.

### LINKED ISSUE #5292: SPV: DocProperties werden nicht aktualisiert bei Generierung Protokoll u. Traktandenliste
Wird in einer Sitzung ein Protokoll oder eine Traktandenliste generiert und heruntergeladen, verfügt die Word-Datei anschliessend nicht über die entsprechenden DocProperties, welche sonst automatisch bei einer Bearbeitung via OC (Auschecken/Bearbeiten) mitgegeben werden.

Reproduziert auf Testsystem -> https://dev.onegovgever.ch/fd/sitzungen/committee-13/meeting-68
Extranet-Ticket -> https://extranet.4teamwork.ch/support/gemeinde-koeniz/tracker/130

bildschirmfoto 2019-01-25 um 09 13 37

phgross:
Because the UpdateGeneratedDocumentCommand manually direclty creates a new version of a document, without using the checkout/checkin process, the docproperty update is skipped. So we need to manually trigger the docproperty update.

_To test this change I added a docproerty to the sablon template in the fixture_ 

Closes #5292


opengever/meeting/command.py
```diff
@@ -358,6 +358,10 @@ def execute(self):
         document = Oguid.resolve_object(self.generated_document.oguid)
         document.update_file(self.generate_file_data())
 
+        # Avoid circular imports:
+        from opengever.dossier.docprops import DocPropertyWriter
+        DocPropertyWriter(document).update()
```
deiferni: you should do this gracefully so that the operation still succeeds even if docprops could not be written. `opengever.document.handlers._update_docproperties` could take care of that.

phgross: @deiferni make sense, updated ...

opengever/meeting/command.py
```diff
@@ -358,6 +358,10 @@ def execute(self):
         document = Oguid.resolve_object(self.generated_document.oguid)
         document.update_file(self.generate_file_data())
 
+        # Avoid circular imports:
+        from opengever.document.handlers import _update_docproperties
```
deiferni: and the circular import problem still persists 😢?

phgross: yes 😏 ...

PR #5355, opened 2019-02-08
Reenable testserver selftest on jenkins by fixing port choosing.

### LINKED ISSUE #5331: Test server self test runs get stuck if XMLRPC port allocation fails

jone:
Closes #5331
 
The port choosing method is switched to using a 0-port so that the OS chooses the port in the testserver.

This is done in order to avoid problems where the jenkins master chooses a port, but the test is run at the slave, on which the port is already used.

The robot server is extended so that it prints out the URLs including the chosen ports for the ZServer and the XMLRPC-server. The selftest script now needs to watch the stdout of the testserver process and wait until the URLs appear

### DIFF
--- a/base-testserver.cfg
+++ b/base-testserver.cfg
@@ -20,8 +20,6 @@ initialization =
     os.environ.setdefault('FTW_STRUCTLOG_MUTE_SETUP_ERRORS', 'true')
     import sys
     sys.argv.insert(1, 'opengever.core.testserver.OPENGEVER_TESTSERVER')
-    print 'Plone:  http://localhost:{}/plone'.format(os.environ.get('ZSERVER_PORT'))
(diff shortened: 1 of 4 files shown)


sources.cfg
```diff
@@ -8,6 +8,7 @@ extensions += mr.developer
 development-packages =
   opengever.maintenance
   collective.js.timeago
+  plone.app.robotframework
```
jone: 🚧 https://github.com/plone/plone.app.robotframework/pull/105
Wait for release; update `versions.cfg`.

jone:
Ready for review.

The build https://ci.4teamwork.ch/builds/221662/tasks/364408 proves that this works on our jenkins slave.

PR #5357, opened 2019-02-08
Testserver: add support for custom fixtures.

jone:
The testserver is useful for testing other projects which use / work
with a GEVER.
Those projects have specific needs: they may need certain content or a specific structure and may not need some parts of GEVER at all.

Thats why we want to be able to inject a custom fixture when starting up the testserver. We need to be able to subclass the default fixture though, so that we can profit from the features in the default fixture.

The custom fixture can be inserted by setting the ``FIXTURE`` environment variable to the path to the fixture python file. The directory of the file needs to be a package, so there needs to be a ``__init__.py``. Since the fixture is stored and maintained in the third party project, it is not part of the python path. The testserver loads the package of the fixture file under the name ``customfixture``, so that importing from gever as well as from this directory works well.

The fixture is changed so that the content creation is moved into a separate method for easier subclassing and changing. We do not want to repeat the init code.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1032,6 +1032,39 @@ The ports used by the testserver can easily be changed through environment varia
 - ``TESTSERVER_CTL_PORT`` - the port of the XMLRPC control server (default: ``55002`).
 
 
+Custom fixtures
+~~~~~~~~~~~~~~~
+
+A custom fixture can be loaded in the testserver.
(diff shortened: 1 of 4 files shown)


lukasgraf:
CI Governor seems to have choked on your second push somehow...

jone:
I've pushed an amended commit; let's see 😉

PR #5367, opened 2019-02-11
Fix mail for task added activity with multiline comment.

### LINKED ISSUE #5290: Darstellungsproblem Benachrichtigungsmail
Siehe https://github.com/4teamwork/opengever.ftw/issues/81.

njohner:
Notification mails for task added activity were not displayed correctly if the Description was multiline.
The problem stems from https://github.com/4teamwork/opengever.core/pull/4152/commits/a2b502635b4576d4b9a29d6b99bc49a6ca3e6b19, where we break up the description into html paragraph blocks if it is multiline. Problem is that for task-added activity, the Description is already an html table, leading to overlapping tags in the final html. gmail then tries to fix the html, but doesn't manage it correctly (whereas mail for example succeeded in fixing the markup).

Proposed solution is to treat the linebreaks beforehand for the case of task-added activity.

**task-added activity before this fix**

screen shot 2019-02-11 at 11 32 52

**task-added activity before this fix**

screen shot 2019-02-11 at 11 32 36

resolves #5290 and https://github.com/4teamwork/opengever.ftw/issues/81

### DIFF
--- a/opengever/task/activities.py
+++ b/opengever/task/activities.py
@@ -56,9 +56,12 @@ class TaskAddedActivity(BaseActivity):
     def render_description_markup(self, data, language):
         msg = u'<table><tbody>'
         for label, value in data:
-            msg = u'{}<tr><td class="label">{}</td><td>{}</td></tr>'.format(
-                msg, self.translate(label, language), value)
-
+            msg = u'{}<tr><td class="label">{}</td><td>'.format(
+                msg, self.translate(label, language))
+            # Break lines correctly in the table rows
+            if value:
+                msg += u"<br />".join(value.splitlines())
+            msg += u"</td></tr>"
         return u'{}</tbody></table>'.format(msg)
 
     def collect_description_data(self, language):
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 ---------------------
 
 - Make sure favorite button is in front of the watermark header. [njohner]
+- Fix mail for task added activity with multiline comment. [njohner]
 - Skip creation of the tasks pdf on resolve for dossiers without tasks. [phgross]
 - Give View permission to Editors on mails. [njohner]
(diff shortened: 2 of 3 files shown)


opengever/task/activities.py
```diff
@@ -56,9 +56,16 @@ def description(self):
...
-
+            msg = u'{}<tr><td class="label">{}</td><td>'.format(
+                msg, self.translate(label, language))
+            # Break lines correctly in the table rows
+            if value:
+                lines = value.splitlines()
+                for i, line in enumerate(lines):
+                    msg += u"{}".format(line)
+                    if i < len(lines) - 1:
+                        msg += u"<br>"
```
lukasgraf: This should be `<br />` instead of `<br>` (we're sending the mail as  XHTML, which [requires all elements to be closed](https://en.wikipedia.org/wiki/XHTML#cite_ref-16)).

opengever/task/activities.py
```diff
@@ -56,9 +56,16 @@ def description(self):
...
         for label, value in data:
-            msg = u'{}<tr><td class="label">{}</td><td>{}</td></tr>'.format(
-                msg, self.translate(label, language), value)
-
+            msg = u'{}<tr><td class="label">{}</td><td>'.format(
+                msg, self.translate(label, language))
+            # Break lines correctly in the table rows
+            if value:
+                lines = value.splitlines()
+                for i, line in enumerate(lines):
```
lukasgraf: Use `msg += u'<br />'.join(lines)` instead? 😉 [`str.join()`](https://docs.python.org/2/library/stdtypes.html#str.join) is suited perfectly for [fencepost problems](https://en.wikipedia.org/wiki/Off-by-one_error#Fencepost_error) like this.

njohner: Oh yes, of course. Stupid...

lukasgraf:
Looks good to me, but tests somehow don't agree 😉

njohner:
Ok should go through now. decoding the `raw_mail` in the tests helped.

PR #5368, opened 2019-02-11
Make sure favorite button is in front of the watermark header.

### LINKED ISSUE #5277: Favorit (Stern) bei Dossiers mit langen Titeln nicht anwählbar
Wenn man die Schriftgrösse (zum Teil auch Fenstergrösse) auf 1 Zeile im Titel vergrössert/verkleinert, so kann der Stern nicht an/abgewählt werden. (s. Anhang im Ticket: *Stern nicht an und abwaehlbar - .mp4)

- Wenn man den Titel wieder auf 2 Zeilen anpasst, so kann man diesen wieder an/abwählen.

Extranet Ticket: https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/474

njohner:
The problem is when the favorite button overlaps the header watermark. Because the watermark ends up being in fron of the button it's not clickable anymore.

screen shot 2019-02-11 at 16 25 53

resolves #5277


plonetheme/teamraum/resources/css/gever/partials/_favorites.scss
```diff
@@ -10,6 +10,8 @@ $favorites-icon-size: 1em;
   display: inline-block;
   vertical-align: 2px;
   margin-left: 4px;
+  z-index: 99;
+  position: inherit;
```
deiferni: why is this necessary could you quickly describe in the PR or drop the line? i would have expected a one-liner 🤔.

njohner: Hum from what I read, and tested, `z-index` only works when position is also set to either `absolute`, `fixed`, or `relative`

njohner: https://www.w3schools.com/cssref/pr_pos_z-index.asp

njohner: `inherit` here will lead to `position: relative` inherited from `#tabbedview-header`
Could also directly set it to `relative`

PR #5372, opened 2019-02-12
Make bumblebee access tokens predictable in testing.

jone:
In testing, we want to have predictable access tokens in order to test against or when using `testserver` with a VCR tape.
In order to make the access token predictable we need a predictable user salt and a configured bumblebee salt.

Bumblebee's `bid` is also random, but we decided to not solve that in GEVER yet because we probably would have to monkey patch it.
There are other solutions, such as:
- append `?bid=static-bid-xy` to all requests; the same bid will be returned
- remove the bid from all urls before serializing VCR tapes

The bid is not relevant for the bumblebee or GEVER functionality. The sole purpose is to be able to correlate requests over multiple services.

/cc @lvonlanthen

### DIFF
--- a/opengever/core/testserver.py
+++ b/opengever/core/testserver.py
@@ -31,6 +31,12 @@ class TestserverLayer(OpengeverFixture):
         portal.portal_languages.use_combined_language_codes = True
         portal.portal_languages.addSupportedLanguage('de-ch')
 
+        os.environ['BUMBLEBEE_DEACTIVATE'] = 'True'
+        os.environ['BUMBLEBEE_APP_ID'] = 'local'
+        os.environ['BUMBLEBEE_SECRET'] = 'secret'
+        os.environ['BUMBLEBEE_INTERNAL_PLONE_URL'] = 'http://nohost/plone'
+        os.environ['BUMBLEBEE_PUBLIC_URL'] = 'http://bumblebee'
+
         setRequest(portal.REQUEST)
         print 'Installing fixture. Have patience.'
         self.get_fixture_class()()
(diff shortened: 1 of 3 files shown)


jone:
It does not work and I don't understand why.

deiferni:
😢 input required?

deiferni:
Hmm yes iirc naming of the tokens is not so good and a bit confusing. It could be it is one of the time base tokens like this one:

https://github.com/4teamwork/ftw.bumblebee/blob/9ef7f36efb95a2ce82ae5ae5ca2cfb5ec339153a/ftw/bumblebee/hashing.py#L12-L13

I don't see a `freeze` anywhere in the selftest. so that may help to keep it constant.

jone:
The testserver is actually freezed ( https://github.com/4teamwork/opengever.core/blob/master/opengever/core/testserver.py#L78 ) and the access token does not expire and should not be time based. It's probably something different 🤔

Rotonen:
Does the annotation store work?

https://github.com/4teamwork/ftw.bumblebee/blob/c5bad600728e5f0c75786df86e39a8ffd505ffca/ftw/bumblebee/usersalt.py#L58

If that assertion fails, you end up in a branch where you get a new UUIDv4.

jone:
Yes the annotation store should work, otherwise the bumblebee preview wouldn't work in production, if I understand correctly what you mean.
See https://github.com/4teamwork/opengever.core/pull/5372/files#diff-80aa9c702cc0bd4985609181b4d9873bR1750

Rotonen:
I mean to ask if the testing can break the annotation store.

deiferni:
hm we might wanna sit down and have a look at this in a debug session. i also think it "should work"™.

PR #5375, opened 2019-02-13
Add date string localization for sablon data.

### LINKED ISSUE #4989: Sablon date strings not translated in user language.
In https://github.com/4teamwork/opengever.core/pull/4417 we have introduced custom format strings for dates in sablons. Unfortunately we use `strftime` and thus the servers locale. For that reason translated days/months will only and incorrectly return the english names.

While we hook into this we could think about also allowing non-zero padded numbers, as that **will** come up, too.

Proposed solution:  somehow find a way to reuse `i18nl10n.ulocalized_time` or some other translation mechanism provided by plone.

njohner:
In #4417 we have introduced custom format strings for dates in sablons. The problem is that these dates are not localized to the user's locale, but to the servers, i.e. english.

To allow for localizations of these dates with a custom format string, we introduce our own date localization method, based upon `i18nl10n.ulocalized_time`. Below an example of a generated agendaitem list with GEVER in french and in german:

screen shot 2019-02-13 at 09 32 07

screen shot 2019-02-13 at 09 32 17
Also note that the resulting dates are zero padded.

resolves #4989

### DIFF
--- a/opengever/meeting/tests/test_agendaitem_list.py
+++ b/opengever/meeting/tests/test_agendaitem_list.py
@@ -8,7 +8,9 @@ from ftw.testing import freeze
 from ooxml_docprops import read_properties
 from opengever.dossier.docprops import TemporaryDocFile
 from opengever.meeting.command import MIME_DOCX
+from opengever.meeting.interfaces import IMeetingSettings
 from opengever.testing import IntegrationTestCase
+from plone import api
 import pytz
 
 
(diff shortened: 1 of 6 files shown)


opengever/meeting/tests/test_agendaitem_list.py
```diff
@@ -191,3 +193,73 @@ def test_json_data(self, browser):
         self.assertDictContainsSubset(expected_metadata, browser.json)
         self.assertDictContainsSubset(expected_participants, browser.json['participants'])
         self.assertEqual(expected_agenda_items, browser.json['agenda_items'])
+
+    @browsing
+    def test_json_data_is_locale_dependent(self, browser):
```
deiferni: should we split this into 4 separate test-cases? it looks like they could be separated as you have multiple do-assert blocks that don't seem to depend on each other.

njohner: Yes sure.

njohner: I also split up `TestMeetingTocDataLocalisation`

njohner:
This should be ready

Issue #5346, opened 2019-02-07
Lock dossier subtree during resolve transition


lukasgraf:
### Background

The `dossier-transition-resolve` transition for a large dossier (with many subdossiers and documents) can be quite expensive. During this time the user currently isn't prevented from triggering the transition a second time. If they do this, this will invariably lead to conflict errors, and only slow down the resolution process and cause unnecessary system load.

We should therefore come up with a locking mechanism that prevents triggering of `dossier-transition-resolve` if one is already running.

### Proposed Solution

Implement a simple persistent locking mechanism for this case. We could
- At the very start of the code for resolving a dossier, store a persistent lock object on the main dossier, and immediately commit.
- Resolution would then proceed as usual, in a new transaction
- On a transaction rollback (because of an unhandled exception, or a `ConflictError`, the persistent lock needs to be removed (and that removal needs to be committed)
- On resolution sucess, the lock also needs to be removed. We should be able to commit this in the same transaction that does the resolution.
- If a lock exists on any of the dossier-type-parents, the resolve transition code should redirect to the context and display a status message that a resolution is already in progress
- The lock should expire after 24h (not consider subtree locked for locks older than that, remove any older locks whenever we get the chance).

lukasgraf:
Closed via #5387

PR #5387, opened 2019-02-14
Lock dossier subtree during resolve transition

lukasgraf:
Implements a **locking mechanism to prevent concurrent dossier resolution**.

This mechanism is intended to prevent users from simultaneously triggering the `resolve` transition for a dossier (which would result in degraded performance because retries of conflicting transactions).

We do this by issuing a **persistent lock** that gets committed in its own transaction as the very first thing in the resolution process. Further attempts at resolving a dossier are then rejected as long as such a lock exists (and hasn't expired). Once the dossier is resolved successfully, the lock is removed and the removal will be committed again. The lock will **expire after 24h**, and which point it will just be treated as if it wouldn't exist.

In case of an exception, we catch it, abort the transaction, remove the lock, commit the removal, and re-reaise the exception to be handled by the usual error handling in `ZPublisher`.

When checking whether a subdossier is locked, possible locks on any of its parents are also considered. So basically an **entire dossier subtree is locked as a whole**, by attaching a lock to the main dossier. If a dossier is locked, this is displayed by appending a message to the workflow status in the byline - this is however only displayed for the current context, not determined recursively (for performance reasons).

In addition, this change includes a **guard condition to prevent dossiers from being resolved twice**.

Logging is currently very much on the verbose side, and could eventually be toned down once this proves to be a reliable strategy in the field.

**Note:** I added a check for a dirty transaction when committing the lock. When the transaction is found to be dirty (which it usually shouldn't be), this is logged to Sentry, but we still proceed with committing. 

#### View of a dossier during resolve process
view_during_resolve

---

#### View of a dossier when a concurrent resolution is attempted
attempt_at_concurrent_resolution

---

#### Example logs for a concurrent attempt at resolving the same dossier

```
2019-02-14 18:24:41 INFO opengever.dossier [<Connection at 10a79c050>] Acquiring resolve lock for <DossierContainer at dossier-54>...
2019-02-14 18:24:41 INFO opengever.dossier [<Connection at 10a79c050>] Resolve lock acquired.
2019-02-14 18:24:52 INFO opengever.dossier [<Connection at 11ada90d0>] <DossierContainer at dossier-54> is resolve locked
2019-02-14 18:24:55 INFO opengever.dossier [<Connection at 11ada90d0>] <DossierContainer at dossier-54> is resolve locked via lock on <DossierContainer at /fd/ordnungssystem/bildung/dossier-54>
2019-02-14 18:24:55 INFO opengever.dossier [<Connection at 11ada90d0>] Concurrent attempt at resolving <DossierContainer at dossier-54> rejected
2019-02-14 18:24:55 INFO opengever.dossier [<Connection at 11ada90d0>] <DossierContainer at dossier-54> is resolve locked
2019-02-14 18:25:11 INFO opengever.dossier [<Connection at 10a79c050>] Successfully resolved <DossierContainer at dossier-54>
2019-02-14 18:25:11 INFO opengever.dossier [<Connection at 10a79c050>] Releasing resolve lock...
2019-02-14 18:25:11 INFO opengever.dossier [<Connection at 10a79c050>] Resolve lock released for <DossierContainer at dossier-54>
```
---

#### Example logs for an exception during resolution
```
2019-02-14 18:21:31 INFO opengever.dossier [<Connection at 10a79c050>] Acquiring resolve lock for <DossierContainer at dossier-54>...
2019-02-14 18:21:31 INFO opengever.dossier [<Connection at 10a79c050>] Resolve lock acquired.
2019-02-14 18:22:01 INFO opengever.dossier [<Connection at 10a79c050>] Releasing resolve lock...
2019-02-14 18:22:01 INFO opengever.dossier [<Connection at 10a79c050>] Resolve lock released for <DossierContainer at dossier-54>
2019-02-14 18:22:01 ERROR Zope.SiteErrorLog 1550164921.490.941219581584 http://localhost:8080/fd/ordnungssystem/bildung/dossier-54/transition-resolve
Traceback (innermost last):
  Module opengever.dossier.resolve, line 81, in __call__
  Module opengever.dossier.resolve, line 132, in execute_recursive_resolve
Exception: Boom
2019-02-14 18:22:01 ERROR Zope.SiteErrorLog 1550164921.50.650119714009 http://localhost:8080/fd/ordnungssystem/bildung/dossier-54/transition-resolve
Traceback (innermost last):
  Module opengever.dossier.resolve, line 81, in __call__
  Module opengever.dossier.resolve, line 132, in execute_recursive_resolve
Exception: Boom
```

*(**Note**: The exception being reported twice has nothing to do with the changes in this PR. I checked, the actual code is only run once, it's just a case of double-reporting that already happens in `maser`. The reason for this is that first ZPublisher [invokes`SiteErrorLog.raising()`](https://github.com/zopefoundation/Zope/blob/2.13.28/src/Zope2/App/startup.py#L214), later [looks up the exception handling view and calls it](https://github.com/zopefoundation/Zope/blob/2.13.28/src/Zope2/App/startup.py#L232), which finds our own error handling view, which then [calls `SiteErrorLog.raising()` again](https://github.com/4teamwork/opengever.core/blob/master/opengever/base/browser/errors.py#L48-L50). )*

### DIFF
--- a/opengever/dossier/resolve.py
+++ b/opengever/dossier/resolve.py
@@ -13,6 +13,7 @@ from opengever.dossier.behaviors.dossier import IDossierMarker
 from opengever.dossier.behaviors.filing import IFilingNumberMarker
 from opengever.dossier.interfaces import IDossierResolveProperties
 from opengever.dossier.interfaces import IDossierResolver
+from opengever.dossier.resolve_lock import ResolveLock
 from opengever.task.task import ITask
 from plone import api
 from Products.CMFCore.utils import getToolByName
@@ -26,6 +27,7 @@ from zope.interface import implementer
 from zope.interface import implements
 from zope.schema.interfaces import IVocabularyFactory
 from zope.schema.vocabulary import SimpleVocabulary
+import transaction
 
 
 NOT_SUPPLIED_OBJECTS = _(
@@ -60,13 +62,45 @@ class ValidResolverNamesVocabularyFactory(object):
 class DossierResolveView(BrowserView):
 
     def __call__(self):
+        # Ensure already resolved dossier can't be resolved again
         if self.is_already_resolved():
             return self.show_already_resolved_msg()
 
-        self.execute_recursive_resolve()
+        resolve_lock = ResolveLock(self.context)
+
+        # Check whether a resolve is already in progress
+        if resolve_lock.is_locked():
+            resolve_lock.log('Concurrent attempt at resolving %s rejected' % self.context)
+            return self.show_being_resolved_msg()
+
+        # No resolve in progress, proceed with resolving this dossier
+        # by acquring a lock, resolving, and then releasing the lock
+        try:
+            resolve_lock.acquire(commit=True)
+            result = self.execute_recursive_resolve()
+            resolve_lock.log('Successfully resolved %s' % self.context)
+            return result
+
+        except Exception:
+            transaction.abort()
+            raise
+
+        finally:
+            # We end up here in two cases:
+            #
+            # 1) Resolve was successful (no exception)
+            # 2) An exception happened during resolve, and the txn has been
+            #    aborted just above (but the exception has not been
+            #    propagated yet).
+            #
+            # In either case, we remove the lock and commit the removal.
+            #
+            # In the case of a previous abort, the removal will be committed
+            # as as separate transaction, in the success case the removal will
+            # be part of the successful dossier resolution transaction.
+            resolve_lock.release(commit=True)
 
     def execute_recursive_resolve(self):
-
         resolver = get_resolver(self.context)
 
         # check preconditions
@@ -83,6 +117,7 @@ class DossierResolveView(BrowserView):
             return self.redirect('transition-archive')
 
         resolver.resolve()
+
         if self.context.is_subdossier():
             return self.show_subdossier_resolved_msg()
 
@@ -102,6 +137,12 @@ class DossierResolveView(BrowserView):
     def show_already_resolved_msg(self):
         api.portal.show_message(
             message=_('Dossier has already been resolved.'),
+            request=self.request, type='info')
+        return self.redirect(self.context_url)
+
+    def show_being_resolved_msg(self):
+        api.portal.show_message(
+            message=_('Dossier is already being resolved'),
             request=self.request, type='info')
         return self.redirect(self.context_url)
 
--- a/opengever/dossier/resolve.py
+++ b/opengever/dossier/resolve.py
@@ -67,39 +67,54 @@ class DossierResolveView(BrowserView):
         # check preconditions
         errors = resolver.is_resolve_possible()
         if errors:
-            for msg in errors:
-                api.portal.show_message(
-                    message=msg, request=self.request, type='error')
-
-            return self.request.RESPONSE.redirect(
-                self.context.absolute_url())
+            self.show_errors(errors)
+            return self.redirect(self.context_url)
 
         # validate enddates
         invalid_dates = resolver.are_enddates_valid()
         if invalid_dates:
-            for title in invalid_dates:
-                msg = _("The dossier ${dossier} has a invalid end_date",
-                        mapping=dict(dossier=title,))
-                api.portal.show_message(
-                    message=msg, request=self.request, type='error')
-
-            return self.request.RESPONSE.redirect(
-                self.context.absolute_url())
+            self.show_invalid_end_dates(titles=invalid_dates)
+            return self.redirect(self.context_url)
 
         if resolver.is_archive_form_needed():
-            self.request.RESPONSE.redirect('transition-archive')
+            self.redirect('transition-archive')
         else:
             resolver.resolve()
             if self.context.is_subdossier():
-                api.portal.show_message(
-                    message=_('The subdossier has been succesfully resolved.'),
-                    request=self.request, type='info')
+                self.show_subdossier_resolved_msg()
             else:
-                api.portal.show_message(
-                    message=_('The dossier has been succesfully resolved.'),
-                    request=self.request, type='info')
+                self.show_dossier_resolved_msg()
 
-            self.request.RESPONSE.redirect(self.context.absolute_url())
+            self.redirect(self.context_url)
+
+    @property
+    def context_url(self):
+        return self.context.absolute_url()
+
+    def redirect(self, url):
+        return self.request.RESPONSE.redirect(url)
+
+    def show_errors(self, errors):
+        for msg in errors:
+            api.portal.show_message(
+                message=msg, request=self.request, type='error')
+
+    def show_invalid_end_dates(self, titles):
+        for title in titles:
+            msg = _("The dossier ${dossier} has a invalid end_date",
+                    mapping=dict(dossier=title,))
+            api.portal.show_message(
+                message=msg, request=self.request, type='error')
+
+    def show_subdossier_resolved_msg(self):
+        api.portal.show_message(
+            message=_('The subdossier has been succesfully resolved.'),
+            request=self.request, type='info')
+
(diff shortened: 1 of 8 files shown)


opengever/dossier/resolve.py
```diff
@@ -59,44 +62,115 @@ def __call__(self, context):
...
+            # In the case of a previous abort, the removal will be committed
+            # as as separate transaction, in the success case the removal will
+            # be part of the successful dossier resolution transaction.
+            resolve_lock.release(commit=True)
+
+    def msg_being_resolved(self):
+        api.portal.show_message(
+            message=_('Dossier is already being resolved'),
+            request=self.request, type='info')
+        return self.request.RESPONSE.redirect(self.context.absolute_url())
```
deiferni: `self.redirect`?

opengever/dossier/resolve.py
```diff
@@ -59,44 +62,115 @@ def __call__(self, context):
 class DossierResolveView(BrowserView):
 
     def __call__(self):
+        # Ensure already resolved dossier can't be resolved again
+        if self.is_already_resolved():
+            self.show_already_resolved_msg()
```
deiferni: below you redirect in the method that shows the message, here you don't. it is a detail but it helps the reader if both things work the same way.

opengever/dossier/resolve.py
```diff
@@ -59,44 +62,115 @@ def __call__(self, context):
...
+            #
+            # In the case of a previous abort, the removal will be committed
+            # as as separate transaction, in the success case the removal will
+            # be part of the successful dossier resolution transaction.
+            resolve_lock.release(commit=True)
+
+    def msg_being_resolved(self):
+        api.portal.show_message(
+            message=_('Dossier is already being resolved'),
+            request=self.request, type='info')
```
deiferni: we could also show this as a warning-message 🤔.

opengever/dossier/resolve.py
lukasgraf: That was the plan all along 😄 Thanks.

opengever/dossier/resolve.py
lukasgraf: Yep - I actually refactored all that stuff with the plan to homogenize the handling of redirects and showing messages. Somehow some stuff got left by the wayside.

opengever/dossier/resolve.py
lukasgraf: I also considered this, and was on the fence for a while. In the end I went with `info` because there really isn't anything to warn the user about - we've all handled it for them. All that's left to do is for us to tell them nicely, and for them to reflect upon their actions 😉

opengever/dossier/resolve_lock.py
```diff
@@ -0,0 +1,132 @@
...
+        self.catalog._increment_counter()
+
+    def is_expired(self, lockinfo):
+        """Determine whether a lock is expired.
+        """
+        ts = lockinfo['timestamp']
+        age = datetime.now() - ts
+        expired = age > RESOLVE_LOCK_LIFETIME
+        if expired:
+            self.log("Resolve lock is expired (age: %s): %s" % (age, lockinfo))
```
njohner: Here expired locks could be cleared

opengever/dossier/resolve.py
```diff
@@ -59,44 +62,115 @@ def __call__(self, context):
...
+            # 2) An exception happened during resolve, and the txn has been
+            #    aborted just above (but the exception has not been
+            #    propagated yet).
+            #
+            # In either case, we remove the lock and commit the removal.
+            #
+            # In the case of a previous abort, the removal will be committed
+            # as as separate transaction, in the success case the removal will
+            # be part of the successful dossier resolution transaction.
+            resolve_lock.release(commit=True)
```
njohner: Ok that's good. Means that if we have a conflict during the transaction, we roll back to a state where the lock is still there, dossier not resolved and try again.
If the transaction fails 3 times, we will end up in a state with the dossier locked and not resolved. But that should not really happen anymore.

njohner:
Missing resolves #5346

opengever/dossier/resolve_lock.py
lukasgraf: @njohner I considered doing this. However, not here, because here we're only *testing* whether a lock is expired. Meaning, if we were to clean up locks here, we would also do this during reads, when just viewing a dossier triggers this check via the byline info. And writes on read are very bad for several reasons.

So I considered clearing existing locks during `acquire()` - but turns out, that really wouldn't add anything to what's being done right now. An existing lock simply gets overwritten with a fresh one during `acquire()`.

lukasgraf:
@njohner thanks for the review and the French translations! 💚

opengever/dossier/resolve_lock.py
njohner: Right, I still forget the to be careful with write on read...

opengever/dossier/resolve.py
lukasgraf: I believe the lock should even be removed in the case of conflicts and 3 retries - because then the `ConflictError` will actually be propagated, our `except` should catch it, and `finally` should still remove the lock. I think. I'll see if I can actually test this case, though it's quite tricky to reliably produce conflicts for testing this.

opengever/dossier/resolve_lock.py
```diff
@@ -0,0 +1,166 @@
...
+        if not recursive:
+            lockinfo = self.get_lockinfo(self.context)
+            if lockinfo is not None and not self.is_expired(lockinfo):
+                self.log("%s is resolve locked" % self.context)
+                return True
+            return False
+
+        item = self.context
+
+        while IDossierMarker.providedBy(item):
```
jone: I prefer using `aq_chain` over a while loop since it is harder to produce an infinite loop when not using `while` 😉 
I'd do it like that:
```python
for item in filter(IDossierMarker.providedBy, aq_chain(item)):
    ...
```

I'm aware that the `while`  loop is not infinit now, but someone might "refactor" it and I practice the for / aq_chain approach as good habit 😜 

You could also make the function more DRY by checking `recursive` in the for-loop, I guess:

```python
for item in filter(IDossierMarker.providedBy, aq_chain(item)):
    ...
    if not recursive:
        return False
return False
```
Although this would have a different result when `context` is not providing `IDossierMarker`; not sure if this is even possible.

opengever/dossier/resolve.py
lukasgraf: @njohner as discussed, you were right, this actually is an issue:

Because a `ConflictError` happens during `commit()`, and in my first approach I committed the removal of the lock together with all the work from resolving the dossiers, the transaction that conflicted was the same one that was supposed to remove the lock 😉 Which lead to a left over lock, which the retried request then stumbled upon (already locked).

So I changed it that we actually `commit()` at the end of the `try` clause (after successfully resolving). This means that a possible `ConflictError` would then manifest there and then, during that `commit()`, which can then by caughty by our `finally` like any other exception: The lock will be removed, removal committed, and we're ready for the retry of the request. Phew. 😉

opengever/dossier/resolve_lock.py
lukasgraf: @jone as discussed: For now, I'm not going with the `filter(..., aq_chain)` approach yet. I would prefer this one too, but then I'd like to actually implement this in a well-tested helper function that is then used in all the places in `opengever.core` where we already do this. Well-tested would also include tests for the odd cases of multiple AQ wrappers, not just containment, that make the `aq_inner` necessary.

lukasgraf:
Updated:
- Improved the handling of ConflictErrors
- Made `ResolveLock.is_locked()` more DRY
- Worked in the French translations
- Made the use of `self.redirect` and early returns more consistent in the resolve view
- Fixed typo in commit message

Also tested again for MeetingDossiers.

PR #5390, opened 2019-02-15
Make sure all workflow IDs are unique.

### LINKED ISSUE #5326: Duplicate workflow_id in workflows
Some workflows have the same `workflow_id`. According to @jone , this is not important, as for traversal it is the id stored on the parent that counts, and this is determined by the folder name. Still this is a bit confusing and could/should be fixed.

Here two culprits:

https://github.com/4teamwork/opengever.core/blob/master/opengever/core/profiles/default/workflows/opengever_private_mail_workflow/definition.xml#L2
https://github.com/4teamwork/opengever.core/blob/master/opengever/core/profiles/default/workflows/opengever_private_document_workflow/definition.xml#L2

We could also add a test that `workflow_id`s are unique.

njohner:
We correct workflow IDs that are not unique and do not match the directory name.

I also add some tests to make sure new workflows will have correct IDs matching their directory name and registration name.

resolves #5326


opengever/core/tests/test_workflow_ids.py
```diff
@@ -0,0 +1,40 @@
...
+from unittest import TestCase
+from zope.dottedname.resolve import resolve
+import os
+
+
+class TestWorkflowIds(TestCase):
+
+    @property
+    def workflow_dir_path(self):
+        return os.path.join(os.path.dirname(resolve(self.__module__).__file__),
```
deiferni: you might be able to use `resource_filename` from `pkg_resources`.

opengever/core/tests/test_workflow_ids.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def test_all_workflows_are_registered(self):
+        workflow_dirnames = self.get_workflow_directories()
+        workflow_xml = etree.parse(os.path.join(self.workflow_dir_path,
+                                                "../workflows.xml"))
+        workflow_names = []
+        for workflow in workflow_xml.getroot().findall("object"):
+            workflow_names.append(workflow.attrib.get("name"))
+        self.assertItemsEqual(workflow_dirnames, workflow_names,
+                              "All workflows should be registered in worklows.xml"
```
deiferni: just for the record, that may not have to be true necessarily. theoretically we could override an existing workflow without having to re-register it in `workflows.xml`.

but for practical reasons i think the assertion is good as it covers our current use-case, we currently always provide our own workflows which all have to be registered. so no need to change anything now 👍 😉.

njohner:
> just for the record, that may not have to be true necessarily. theoretically we could override an existing workflow without having to re-register it in `workflows.xml`

Yes the `name` in `workflows.xml` has to be the same as the directory in which it is located. The `name`, stored on the parent, is used to access a given workflow (through traversal). So the `workflow_id` is not really relevant for anything.

opengever/core/tests/test_workflow_ids.py
njohner: I added a comment in the test.

njohner:
Ok I included the suggested change and addressed the other comments.
This should be ready.

PR #5379, opened 2019-02-13
Set request language based on site wide preferred language for task reminders

njohner:
To make sure that (hopefully) everything gets translated correctly down the line, we set the language on the request from the cronjobs generating the reminders.

resolves #5378

PR #5415, opened 2019-02-25
Set default language for dossier overdue activity.

njohner:
More translations problems for mails generated via cron jobs. We corrected this for task reminders in https://github.com/4teamwork/opengever.core/pull/5379, and now the same fix is applied for dossier overdue activity. We are simply making sure the language is set on the request before generating the E-mails.

This resolves https://basecamp.com/2768704/projects/12554551/todos/380774579

### DIFF
--- a/opengever/dossier/cronjobs.py
+++ b/opengever/dossier/cronjobs.py
@@ -1,6 +1,7 @@
 from opengever.core.debughelpers import all_plone_sites
 from opengever.core.debughelpers import setup_plone
 from opengever.dossier.activities import DossierOverdueActivityGenerator
+from plone import api
 import logging
 import transaction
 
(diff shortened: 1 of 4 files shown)


opengever/dossier/cronjobs.py
```diff
@@ -16,7 +17,15 @@ def generate_overdue_notifications_zopectl_handler(app, args):
     logger.setLevel(logging.INFO)
 
     for plone_site in all_plone_sites(app):
-        setup_plone(plone_site)
+        plone = setup_plone(plone_site)
+
+        # Set up the language based on site wide preferred language. We do this
+        # so all the i18n and l10n machinery down the line uses the right language.
+        lang_tool = api.portal.get_tool('portal_languages')
+        lang = lang_tool.getPreferredLanguage()
```
lukasgraf: @njohner mhh, `getPreferredLanguage()` doesn't (necessarily) get the site wide language - it gets the currently *preferred* language, according to all negotiation steps, including HTTP Request Negotiation.

So, this shouldn't even make a difference really. 😕

njohner: From what I see it returns the current language if it is set, or the default language if there is no current language.
This seems to have fixed the issue for the task reminders though. At least I guess it did, Erich closed the to-do

lukasgraf: @njohner you're right in that the

> Set up the language based on site wide preferred language.
> We do this so all the i18n and l10n machinery down the line uses the right language.

bit was written by me, actually 🤕 I was wrong I guess 😛

njohner:
Created a new issue to have a closer look at the language negotiation:
https://github.com/4teamwork/opengever.core/issues/5416

and I added a comment in the code to more easily find the places to cleanup once we resolve that issue.

PR #5420, opened 2019-02-26
Hide system-actor in the my-notification listing.

phgross:
The "SYSTEM_ACTOR" used for internal generated activities/notifications, is confusing for the user, therefore the "My Notification" listing does no longer display the SYSTEM_ACTOR id.

See https://basecamp.com/2768704/projects/12554551/todos/380672644 for more information.


opengever/activity/browser/listing.py
```diff
@@ -20,7 +21,12 @@ def resolve_notification_link(item, value):
 
 
 def readable_actor(item, value):
-    return Actor.lookup(item.activity.actor_id).get_label()
+    actor_id = item.activity.actor_id
```
deiferni: The object oriented way would be to implement a `SystemActor` that does not have a label. Otherwise we'll have to implement this special case everywhere.

phgross:
@deiferni updated according to your feedback.

PR #5437, opened 2019-02-28
Add sharing white and black list prefix to config endpoint/view.

### LINKED ISSUE #5423: Include sharing information in the config endpoint/view

The settings `white_list_prefix` and `black_list_prefix` from `opengever.sharing.interfaces.ISharingConfiguration` should be included in the config adapter, so that values are visible in the config view and endpoint.

njohner:
We add the `black_list_prefix` and `white_list_prefix` from `ISharingConfiguration` to the config endpoint.

I wonder whether we should group them under `sharing_config` or something like that? Just the names as they are now are not very explicit.

screen shot 2019-02-28 at 15 13 25

resolves #5423

### DIFF
--- a/docs/public/dev-manual/api/config.rst
+++ b/docs/public/dev-manual/api/config.rst
@@ -54,7 +54,11 @@ GEVER-Mandanten abgefragt werden.
           "max_repositoryfolder_levels": 3,
           "recently_touched_limit": 10,
           "root_url": "http://localhost:8080/fd",
-          "version": "2018.4.0.dev0"
+          "sharing_configuration": {
+              "black_list_prefix": "^$",
+              "white_list_prefix": "^.+"
+          }
+          "version": "2018.4.0.dev0",
       }
 
 
@@ -120,3 +124,11 @@ features
 
     workspace
         Arbeitsräume
+
+sharing_configuration
+
+    white_list_prefix
+        regex Muster für Gruppen die in der Freigabe angezeigt werden sollen
+
+    black_list_prefix
+        regex Muster für Gruppen die in der Freigabe nicht angezeigt werden sollen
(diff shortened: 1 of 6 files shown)


docs/public/dev-manual/api/config.rst
```diff
@@ -67,6 +69,12 @@ max_repositoryfolder_levels
 max_dossier_levels
     Maximale Verschachtelungstiefe von Dossiers
 
+white_list_prefix
+    regex muster für Gruppen die in der Freigabe angezeigt werden sollen
```
deiferni: Regex Muster

docs/public/dev-manual/api/config.rst
```diff
@@ -67,6 +69,12 @@ max_repositoryfolder_levels
...
+
+black_list_prefix
+    regex muster für Gruppen die in der Freigabe nicht angezeigt werden sollen
```
deiferni: same

opengever/sharing/interfaces.py
```diff
@@ -29,13 +29,13 @@ class ISharingConfiguration(Interface):
...
-        description=u'The prrefix pattern, for groups wich should displayed'
-        'in the sharing view, equal the black list prefix would also match.',
+        description=u'The prefix pattern, for groups which should be displayed'
+        'in the sharing view, even if the black list prefix also matches.',
         default=u"^.+")
 
     black_list_prefix = schema.TextLine(
         title=u'black list prefix',
-        description=u"The prrefix pattern, for groups wich shouldn't be "
+        description=u"The prefix pattern, for groups which shouldn't be "
```
deiferni: pattern, -> pattern

deiferni: pattern, -> pattern

njohner:
Ok I've included your comments and nested the settings in `sharing_configuration`
This should be ready to 🚢

PR #5452, opened 2019-03-05
Do not list resolved tasks as pending in the 'My Tasks' tab

Rotonen:
Added a new filter for the 'My Tasks' tab.

Closes https://github.com/4teamwork/opengever.core/issues/5202

### DIFF
--- a/opengever/globalindex/model/task.py
+++ b/opengever/globalindex/model/task.py
@@ -577,8 +577,11 @@ class TaskQuery(BaseQuery):
         """Queries all subtask of the given task sql object."""
         return self.subtasks_by_tasks([task])
 
+    def in_state(self, states):
+        return self.filter(Task.review_state.in_(states))
+
     def in_pending_state(self):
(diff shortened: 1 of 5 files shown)


opengever/globalindex/model/task.py
```diff
@@ -61,7 +61,6 @@ class Task(Base):
     PENDING_STATES = [
         'task-state-open',
         'task-state-in-progress',
-        'task-state-resolved',
```
phgross: 🤔 hmm, this change has impact on other listing too, but the idea behind the issue #5202 was to hide resolved task only on the `My Task` listing not on every one. 

Because in general its correct that the `resolved` task are still pendent, because the transition to the `closed` state is missing. But for task assigned to the current user, the resolved state is the final state for him, because the task has to be closed by the issuer.

Rotonen: Added an extra filter just for that use case just for that tab instead.

Issue #5453, opened 2019-03-05
Double-Check and re-enable template mimetypes.


deiferni:
In https://github.com/4teamwork/opengever.core/pull/5382 support for templates was removed. But it seems they are supported by OC, i just tested a `*.dotx` with @buchi. We should

- Verify that the other office templates actually work with OC
- Re-enable those that work, discuss those that don't

This has surfaced in https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/611.

Rotonen:
Templates do work on Windows. They currently do not work on macOS.

Rotonen:
Additionally editing restricted templates also works.

PR #5454, opened 2019-03-05
Readd Office template, macro, stencil files into Office Connector editable MIME types

Rotonen:
We're readding Office Connector editability for Office template formats, as they work on Windows.

Closes https://github.com/4teamwork/opengever.core/issues/5453

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 2019.2.0 (unreleased)
 ---------------------
 
+- Readd Office template files into Office Connector editable MIME types. [Rotonen]
 - Readd Office macro files into Office Connector editable MIME types. [Rotonen]
 - Complete French translations of repository in examplecontent. [andresoberhaensli]
 - Adapt footer to new 4teamwork website. [njohner]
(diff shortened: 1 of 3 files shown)


Rotonen:
We cannot guarantee any macro type to work on any platform.
We cannot guarantee any addin type to work on any platform.

The empty skeleton files of all the variants work on Windows.

We can blacklist template MIME types in macOS client policies for now, if desired.

Added Excel addin dependent and binary macro enabled formats.
Added Powerpoint addin dependent, slideshow and template macro enabled formats.
Added all Visio drawing, stencil and template macroenabled formats.

deiferni:
(as discussed, double-check mimetypes registry)

Rotonen:
@deiferni we indeed had some stuff not present in the document MIME type registry - added a test to ensure everything on the default editables list is present in that registry.

Ready for rereview.

Rotonen:
As per discussion I'm dropping the added test and most of the changes to our extra types.

https://github.com/python/cpython/blob/2.7/Lib/mimetypes.py#L40-L50

https://github.com/python/cpython/blob/2.7/Lib/mimetypes.py#L405-L536

We want to minimise mucking around with the Python mimetype registry as per how it works under the hood.

I'm leaving in the a change to fix the entry for the powerpoint macro enabled template in the extra registry entries we provide.

PR #5456, opened 2019-03-06
Add support for filters in listing endpoint

buchi:
Filters can be passed with the `filters` record query parameter.

Example:
`/@listing?name=dossiers&filters.review_state:record:list=dossier-state-active&filters.review_state:record:list=dossier-state-archived&filters.responsible:record=hugo.boss`


buchi:
The listing endpoint is intentionally not documented yet as it's not public.
We are developing it for the new gever-ui and want to be able to break compatibility if required.

deiferni:
the listing endpoint is also used in other internal projects such as ris. i'd rather mention in the docs that it is not considered stable yet.

buchi:
This PR ist not about adding the listing endpoint and thus also not about adding documentation for it.
It just adds an additional option.

I'll add the documentation in a separate PR.

buchi:
documentation is in #5494

Issue #5434, opened 2019-02-27
Do a documentation pass on the REST API endpoints


Rotonen:
_Harvest: `Projekt - Umsetzung Neues User Interface GEVER 2019/2020` - `Entwicklung`_

As wished through https://github.com/4teamwork/opengever.core/pull/5433, we currently have major gaps in our REST API documentation we should address.

Known gaps:

* Listings
* ~~Documents~~
* ~~Dossiers~~

And probably many more.

We should thus do a cleanup sweep of our REST API documentation as we now have, including, but not limited to, active internal clients depending on it.

phgross:
@Rotonen what do you mean with?
 `Documents`
 `Dossiers`

Rotonen:
We have no explicit examples of how to interact with those data types over the REST API.

deiferni:
Ok we should document only the `@listing` endpoint in a first step as this is a major gap.

Rotonen:
Maybe also ask any internal consumers for what they already use and where they would have found documentation useful?

PR #5461, opened 2019-03-06
Extend listing with workspaces

elioschmutz:
- Erweitert die vorhandenen @listing-queries mit `workspaces`.

Issuer: https://github.com/4teamwork/gever-ui/issues/15

### DIFF
--- a/opengever/api/listing.py
+++ b/opengever/api/listing.py
@@ -131,6 +131,9 @@ SOLR_FILTERS = {
     ],
     u'documents': [
         u'object_provides:opengever.document.behaviors.IBaseDocument',
+    ],
+    u'workspaces': [
+        u'object_provides:opengever.workspace.interfaces.IWorkspace',
     ]
 }
 
@@ -140,7 +143,11 @@ CATALOG_QUERIES = {
     },
     'documents': {
         'object_provides': 'opengever.document.behaviors.IBaseDocument',
+    },
+    'workspaces': {
+        'object_provides': 'opengever.workspace.interfaces.IWorkspace',
     }
+
 }
 
 
--- a/opengever/api/tests/test_listing.py
+++ b/opengever/api/tests/test_listing.py
@@ -217,6 +217,23 @@ class TestListingEndpoint(IntegrationTestCase):
         self.assertEqual(1, len(start_dates))
         self.assertEqual('2016-01-01', start_dates[0])
 
+    @browsing
+    def test_workspaces_listing(self, browser):
+        self.login(self.workspace_member, browser=browser)
+        query_string = '&'.join((
+            'name=workspaces',
+            'columns=title',
+            'columns=responsible_fullname',
+        ))
+        view = '?'.join(('@listing', query_string))
+        browser.open(self.workspace_root, view=view, headers={'Accept': 'application/json'})
+
+        self.assertDictEqual(
+            {u'responsible_fullname': u'Fr\xf6hlich G\xfcnther',
+             u'@id': u'http://nohost/plone/workspaces/workspace-1',
+             u'title': u'A Workspace'},
+            browser.json['items'][-1])
+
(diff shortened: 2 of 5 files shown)


opengever/api/listing.py
```diff
@@ -95,6 +95,7 @@ def relative_path(brain):
     'relative_path': (None, relative_path, DEFAULT_SORT_INDEX),
     'responsible': ('responsible', 'responsible', 'responsible'),
     'responsible_fullname': ('responsible', 'responsible_fullname', 'responsible'),
+    'responsible_fullname_with_principal': ('responsible', 'responsible_fullname_with_principal', 'responsible'),  # noqa
```
buchi: Why do you need that additional column? `responsible` and `responsible_fullname` already provide the information. Concatenating those strings should be rather done in the frontend.

elioschmutz: 🤔 that was my first thought. But my consideration was, that we have to use this kind of user-representation in a lot of different listings. and instead of using a special method or injection in the frontend, it would be great if we can just get it directly from the backend.

Do you really think it's better to handle it in the frontend?

buchi: Not sure if we are going to use this representation in the new frontend at all. It wastes a lot of space and the username is not that important. I would recommend to only show the fullname and maybe the username on hover if it's really required. But that's another story.

I would still handle it in the frontend. Gives you more flexibility and it's just about representation.
But if you're still convinced that it's the better solution, go on.

elioschmutz: 👍 i just started a discussion with @maethu in the [related frontend-issue](https://github.com/4teamwork/gever-ui/pull/79) to change the spec to only display the fullname.

If so, this implementation will be useless. Otherwise, I'll move this implementation to the frontend as you have proposed.

elioschmutz: decision: at the moment, we only display the fullname.

elioschmutz:
@buchi 👍 update the PR:

removed the `responsible_fullname_with_principal` extension because we only display the fullname in the frontend.

elioschmutz:
@phgross 👍 added tests and changelog.

The  #5434 already includes the note, that the @listing endpint needs to be documented. The workspaces-listing is a part of the @listing endpoint like `dossiers` or `documents`.

elioschmutz:
@phgross ...

opengever/api/tests/test_listing.py
```diff
@@ -165,3 +165,20 @@ def test_review_state_and_review_state_label(self, browser):
...
+        self.login(self.workspace_member, browser=browser)
+        query_string = '&'.join((
+            'name=workspaces',
+            'columns=title',
+            'columns=responsible_fullname',
+        ))
+        view = '?'.join(('@listing', query_string))
+        browser.open(self.workspace_root, view=view, headers={'Accept': 'application/json'})
+
+        self.assertItemsEqual(
```
phgross: I would like to have a tests which fetchs the same fields as they are planned for the new UI (is it really only the title? 🤔 )

elioschmutz: Jup: https://github.com/4teamwork/gever-ui/pull/79

elioschmutz:
@phgross rebased...

opengever/api/tests/test_listing.py
```diff
@@ -217,6 +217,23 @@ def test_filter_by_start_date(self, browser):
...
+            {u'responsible_fullname': u'',
```
phgross: That seems to be wrong? Should be the self.workspace_owner fullname ...

elioschmutz: 👍 updated

elioschmutz:
@phgross 👍 updated

elioschmutz:
@buchi could you  final-review and merge please?

PR #5476, opened 2019-03-07
Make sure task and journal PDFs object_provides index is up to date

### LINKED ISSUE #5473: Dossier journal PDF generated twice
Currently when `resolving` a dossier we create PDFs for journal and tasks. Those seem to be generated twice when i:

- create a dossier
- close it
- reopen it
- close it again

expected outcome:
the already existing PDF should be updated with a new version.

screenshot 2019-03-07 at 09 40 38

as correctly suspected by @njohner the object is not reindexed correctly in the catalog and thus the marker interface is not in `object_provides`.

Possibly related with https://github.com/4teamwork/opengever.core/issues/5458.

njohner:
after resolving a dossier.

Wow, that looks like a 🐷 worthy one for me! I actually did the reindexing manually in the tests! Really not sure why I thought it was normal to have to reindex in the tests... Probably just too new to plone back then
resolves #5473

### DIFF
--- a/opengever/core/upgrades/20190307132714_fix_task_and_journal_pdf_object_provides/upgrade.py
+++ b/opengever/core/upgrades/20190307132714_fix_task_and_journal_pdf_object_provides/upgrade.py
@@ -11,6 +11,8 @@ class FixTaskAndJournalPDFObjectProvides(UpgradeStep):
     """Fix task and journal pdf object provides.
     """
 
+    deferrable = True
+
     def __call__(self):
         catalog = self.portal.portal_catalog
         object_provides = catalog._catalog.getIndex("object_provides")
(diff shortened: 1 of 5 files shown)


njohner:
I could quickly write an upgrade step. We will clearly need one, and I don't think it would be too much of a problem. We search for documents with the appropriate title and then just check that the interface is on the object and not in the index.

opengever/core/upgrades/20190307132714_fix_task_and_journal_pdf_object_provides/upgrade.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+journal_interface_name = 'opengever.document.interfaces.IDossierJournalPDFMarker'
+task_interface_name = 'opengever.document.interfaces.IDossierTasksPDFMarker'
+
+
+class FixTaskAndJournalPDFObjectProvides(UpgradeStep):
+    """Fix task and journal pdf object provides.
+    """
+
```
deiferni: make this deferable

PR #5490, opened 2019-03-11
Fix performance issue with search root exclusion in tabbed view listings

### LINKED ISSUE #5487: TabbedView listings for containers with many immediate children are slow
TabbedView listings for containers with many immediate children are slow. This is caused by the mechanism for excluding the search root in 4teamwork/ftw.table#53.

See 
- https://extranet.4teamwork.ch/support/gever-ai/tracker/340
- https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/623

lukasgraf:
The original implementation for excluding the search root in `ftw.table`, introduced in 4teamwork/ftw.table#53, creates one path query for each immediate child. Turns out that performs much worse than expected for containers with lots of immediate children.

Because in GEVER we have a more recent version of `Products.ZCatalog`, and also [monkeypatch the UIDIndex's query options to add support for the 'not' operator](https://github.com/4teamwork/opengever.core/blob/master/opengever/base/monkey/patches/uuidindex.py), we're able to use a more efficient strategy here, and simply exclude the current context using `{'not': uuid}` filter.

Fixes #5487

### DIFF
--- a/opengever/tabbedview/catalog_source.py
+++ b/opengever/tabbedview/catalog_source.py
@@ -7,8 +7,12 @@ from opengever.base.solr import OGSolrDocument
 from opengever.tabbedview.filtered_source import FilteredTableSourceMixin
 from opengever.tabbedview.interfaces import IGeverCatalogTableSourceConfig
 from plone.registry.interfaces import IRegistry
+from plone.uuid.interfaces import IUUID
+from Products.CMFCore.interfaces import IContentish
 from zope.component import adapter
 from zope.component import getUtility
+from zope.component import queryAdapter
+from zope.globalrequest import getRequest
 from zope.interface import implementer
 from zope.interface import Interface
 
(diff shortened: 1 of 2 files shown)


opengever/tabbedview/catalog_source.py
```diff
@@ -30,6 +34,37 @@ def search_results(self, query):
...
+        That implementation has serious performance issues for containers
+        with many immediate children.
+
+        In GEVER, we have a newer Products.ZCatalog version and monkey patch
+        support for the 'not' operator onto the UIDIndex's query options, so
+        we can use a more efficient approach here.
+        """
+        if not getattr(self.config, 'exclude_searchroot', True):
+            return query
+
```
njohner: Is there no use to also exclude cases where `query.get('path').get('depth') in (0, 1)`?

opengever/tabbedview/catalog_source.py
```diff
@@ -30,6 +34,37 @@ def search_results(self, query):
...
+        """Attempt to find the context from the request.
+
+        This is needed for excluding the search root below, and should
+        be considered 'good enough' for here, but not used anywhere else.
+        """
+        request = getRequest()
+        for item in request.get('PARENTS', []):
+            if IContentish.providedBy(item):
+                return item
+
```
deiferni: i would add an explicit `return None` here as that may happen and is expected control flow. helps the reader to quickly see that even though `None` is returned implicitly.

opengever/tabbedview/catalog_source.py
lukasgraf: Good point! It does indeed make sense to also adress these cases here.
=> Included the respective code from the original implementation.

lukasgraf:
Updated.

njohner:
Careful with the HISTORY that got messed up

lukasgraf:
@njohner thanks! Fixed 👍

PR #5493, opened 2019-03-12
Don't list autogenerated documents in recently touched menu

### LINKED ISSUE #5471: Don't list auto-generated documents as recently touched.
We tag automatically generated documents with the `IAutoGeneratedDocument` interface (currently task and journal PDF). These documents should not show up in the recently touched menu.

Beware, with the current implementation we only add the interfaces when the document has been created and added to the dossiers. Unfortunately the events that are used to figure out if a document has been touched have been fired by the time we add the `IAutoGeneratedDocument` interface in https://github.com/4teamwork/opengever.core/blob/71f484ef5a3f268219bf3e2223862f634acce901/opengever/dossier/resolve.py#L405-L410

As we already monkey patch `createContentInContainer` in 
https://github.com/4teamwork/opengever.core/blob/71f484ef5a3f268219bf3e2223862f634acce901/opengever/base/monkey/patches/default_values.py#L137-L145
it may be an option to add an optional parameter to the patched `createContentInContainer`. That parameter could accept a list of interfaces that slapped onto the object with `alsoProvides` after it is created by `createContentWithDefaults` but before calling `addContentToContainer`.

https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/635

njohner:
To avoid having the recently touched menu flooded with journal and task pdfs when a large dossier is resolved, we want to avoid showing them in the recently touched menu.

Because the touched is handled on the `ObjectAddedEvent`, we need to add the `IAutoGeneratedDocument` interface during the creation of the object. Fortunately we already overwrite `createContentInContainer`...

resolves #5471


opengever/base/monkey/patches/default_values.py
```diff
@@ -98,8 +99,11 @@ class PatchDXCreateContentInContainer(MonkeyPatch):
     default values for fields that haven't had a value passed in to the
     constructor.
 
-    Additionaly, have the request provide IDuringContentCreation while content
+    Additionaly:
+    - Have the request provide IDuringContentCreation while content
     creation is in progress.
+    - Acept an *interfaces* argument to add interfaces onto the created object
```
phgross: Typo: Acept

PR #5495, opened 2019-03-13
 Respect the local roles and the inheritance thereof of dossier templates

### LINKED ISSUE #5470: Apply dossiers template's local roles to content created from that template.

Currently when creating dossiers from a template the template's local roles are discarded.
This should be changed in such a way that the dossier template's local role and block inheritance setting are also applied to the dossiers created from the template.

Template dossiers that can't be seen by the current user shall not be created.

Something similar has already been done when copying dossiers and could maybe be reused, see: https://github.com/4teamwork/opengever.core/pull/5300.

https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/628

Rotonen:
* Added documents to the dossier templates in the fixtures
  * This allowed for the reduction of the use of the builder in the tests
* Copying over, if set, the local roles inheritance block from the template to the target
  * For dossiers
  * For documents
* Copying over, if set, the annotation storage for local roles assignation from the template to the target
  * For dossiers
  * For documents

Had a couple of false starts in trying to reuse the existing logic from copy/pasting, but that turned out too complex and I thus fell back to a simple solution for the scope.

Closes #5470


Rotonen:
I've created this PR from the link in https://ci.4teamwork.ch/builds/228291 - now I'm confused as to why the run status has not propagated to GitHub.

Rotonen:
And after a rebase, everything has managed to talk to each other.

Rotonen:
And this one again got weirdly stuck after a rebase. This time the build is not yet starting on the CI. Hopefully just a matter of webhook delivery congestion.

opengever/dossier/command.py
```diff
@@ -16,9 +19,27 @@ def __init__(self, context, template_doc, title, recipient_data=tuple()):
             title=title)
         self.recipient_data = recipient_data
 
+        # Grab blocking of role inheritance
+        self.block_role_inheritance = getattr(
+            template_doc, '__ac_local_roles_block__', None)
+
+        # Grab a deep copy of the local roles assignations of the template
+        self.storage_key = RoleAssignmentManager(template_doc).storage.key
+        self.role_assignments = IAnnotations(template_doc).get(self.storage_key)
```
njohner: I feel like this create command knows too much about the role assignment storage. I'd try to use the `RoleAssignmentManager` interface more. For example here:
```self.role_assignments = RoleAssignmentManager(template_doc).storage._storage```
Or even better would be to add a method to the `RoleAssignmentManager` to get all assignments and use it here.

File: opengever/dossier/command.py [hunk 2]
```diff
@@ -16,9 +19,27 @@ def __init__(self, context, template_doc, title, recipient_data=tuple()):
...
         obj = super(CreateDocumentFromTemplateCommand, self).execute()
         DocPropertyWriter(obj, recipient_data=self.recipient_data).initialize()
+
+        # Set blocking of role inheritance based on the template object
+        if self.block_role_inheritance is not None:
+            obj.__ac_local_roles_block__ = self.block_role_inheritance
+
+        # Copy the local roles assignations over from the template
+        if self.role_assignments:
+            IAnnotations(obj)[self.storage_key] = self.role_assignments
```
njohner: Same here, we're relying on the knowledge of the underlying storage in a distant part of the code. I think it would be nicer to handle this on the level of the `RoleAssignmentManager`.

opengever/dossier/command.py
```diff
@@ -32,6 +53,17 @@ def __init__(self, context, template):
...
         del kw["IOpenGeverBase"]
         self.additional_fields = kw
+
+        # Grab blocking of role inheritance
+        self.block_role_inheritance = getattr(
+            template, '__ac_local_roles_block__', None)
+
+        # Grab a deep copy of the local roles assignations of the template
+        self.storage_key = RoleAssignmentManager(template).storage.key
+        self.role_assignments = IAnnotations(template).get(self.storage_key)
```
njohner: Same as above

opengever/dossier/tests/test_dossier_template.py
```diff
@@ -428,6 +430,102 @@ def test_add_recursive_documents_and_subdossiers(self, browser):
             [obj.title for obj in subdossier.listFolderContents()],
             "The content of the subdossiertemplate is not correct"
         )
+        expected_role_assignments = []
+        role_assignments = RoleAssignmentManager(subdossier).storage._storage()
+        self.assertEqual(expected_role_assignments, role_assignments)
+
+    @browsing
+    def test_add_recursive_documents_and_subdossiers_local_roles_block_true(self, browser):
```
njohner: My understanding is that you're testing that `subdossiertemplate` is not copied because the user does not have permission to see it. The test title does not really reflect the tested behavior.

opengever/dossier/tests/test_dossier_template.py
```diff
@@ -428,6 +430,102 @@ def test_add_recursive_documents_and_subdossiers(self, browser):
...
+        token = browser.css(
+            'input[name="form.widgets.template"]').first.attrib.get('value')
+        browser.fill({'form.widgets.template': token}).submit()
+        browser.click_on('Save')
+        dossier = browser.context
+
+        self.assertEqual('Bauvorhaben klein', dossier.title)
+        self.assertEqual(
+            [u'Werkst\xe4tte'],
+            [obj.title for obj in dossier.listFolderContents()],
```
njohner: does `listFolderContents` ignore permissions? Or could the subdossier not be in the list because the user does not have the permissions to see it? You could maybe use the `with self.observe_children(self.leaf_repofolder)` construct instead and really look at what has been added

opengever/dossier/command.py
Rotonen: If I access `._storage`, it creates an annotation, if there was none.

https://github.com/4teamwork/opengever.core/blob/master/opengever/base/role_assignments.py#L179-L184

opengever/dossier/tests/test_dossier_template.py
Rotonen: As `observe_children` simply sees the dossier as added, that leaves the same dilemma.

I can log in as manager for that assertion.

opengever/dossier/tests/test_dossier_template.py
Rotonen: Thank you. Improving the naming.

opengever/dossier/command.py
njohner: True, but the annotation is anyway created by any `get` on the `RoleAssignmentManager`, for example by the sharing API endpoint... If that is a problem, then I'd say we need to fix it in the `RoleAssignmentManager`.

opengever/dossier/tests/test_dossier_template.py
njohner: 👍

opengever/dossier/command.py
Rotonen: It being created on normal access is not a problem.

When trying to copy an exact copy of the current state of the source object, it is a problem.

njohner: I get your point. But then why not add a `has_assignments` method on the `RoleAssignmentManager`, which takes care of checking whether assignments are present, without creating an entry in the annotation, and/or a `get_all_assignments`, which returns all assignments from the storage if any. For me the `RoleAssignmentManager` should be the interface between the storage and code using the role assignments. 

I don't feel too strongly about this though.

Rotonen: I'm fine with adding a `has_assignments` to the `RoleAssignmentManager`.

Will do. Thank you.

Rotonen:
I'm now doing everything through the manager on the dossier command, but I'm not fully happy how opaque the case of preserving there being an empty mapping became, as it gets created by a side effect of passing in an empty iterable of assignations. Looks cleaner, though.

docs/HISTORY.txt
```diff
@@ -36,6 +36,9 @@ Changelog
 - Exclusively handle templates on Committee and not on CommitteeContainer anymore. [njohner]
 - Extend @listing endpoint with `tasks`-listing. [elioschmutz]
 - Add option not to revoke permissions associated with a task when closing it. [njohner]
+- Fix performance issue with search root exclusion in tabbed view listings. [lgraf, buchi]
```
njohner: Rebase got you

Rotonen: Thank you for spotting that one, will rebase and merge.

PR #5500, opened 2019-03-14
Respect tabbedview settings when generating an dossier or task excel export.

### LINKED ISSUE #5469: Excel export should respect tabbedview-settings.

The excel export that is available for listing via `Weitere Aktionen` -> `Auswahl exportieren` should also only include the columns that are currently selected in the users tabbed-view setting. Also the order of the columns as customized by the current user should be respected.

Care should be taken to make sure this works in all places where such an export is available.

https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/627

phgross:
The personal tabbeview extjs listing settings (visible columns, column order) are now respected in Excel exports (closes #5469).

Implementation details and additional work:
 - To fetch the correct tabbedview setting i add the `view_name` as an additional tabbedview form value.
 - Fetching the currently visible columns and their order is implemented conservative, so that the default settings are used, when no personal settings exists or the settings couldn't be loaded.
 - Convert all existing tests to integration tests.
 - Implement base class for excel export views  and smaller refactorings.


opengever/dossier/browser/report.py
```diff
@@ -16,8 +20,9 @@ class DossierReporter(BrowserView):
     """
 
     def get_dossier_attributes(self):
-        return [
+        attributes = [
```
lukasgraf: These are rather columns than attributes, no? (Both in the Excel sheet as well as in the TabbedView table).
So I would probably rather go with `excel_columns` here...

opengever/dossier/browser/report.py
```diff
@@ -37,6 +42,47 @@ def get_dossier_attributes(self):
                         default=u'Reference Number')},
         ]
 
+        return self.filter_and_order_by_tabbedview_settings(attributes)
+
+    def filter_and_order_by_tabbedview_settings(self, attributes):
```
lukasgraf: `attributes` -> `excel_columns`

opengever/dossier/browser/report.py
```diff
@@ -37,6 +42,47 @@ def get_dossier_attributes(self):
...
+        attributes_dict = {
```
lukasgraf: `excel_columns_by_name`?

opengever/dossier/tests/test_reporter.py
```diff
@@ -1,116 +1,109 @@
...
 
-class TestDossierReporter(FunctionalTestCase):
+    @browsing
+    def test_respects_column_tabbedview_settings_if_exists(self, browser):
+        self.login(self.regular_user, browser=browser)
 
-    def setUp(self):
-        super(TestDossierReporter, self).setUp()
+        data = {'view_name': 'mydossiers',
+                'gridstate': '{"columns":[{"id":"path_checkbox","width":30},{"id":"sortable_title","width":300,"sortable":true},{"id":"review_state","width":110,"sortable":true},{"id":"reference","width":110,"sortable":true},{"id":"end","width":110,"sortable":true},{"id":"responsible","width":110,"sortable":true},{"id":"start","width":110,"hidden":true,"sortable":true},{"id":"Subject","width":110,"sortable":true},{"id":"dummy","width":1,"hidden":true}],"sort":{"field":"modified","direction":"ASC"}}'}
```
lukasgraf: I would also prefer the same style (using `json.dumps` etc..) here like the one you used for the task report tests.

phgross:
@lukasgraf updated ...

PR #5501, opened 2019-03-14
Allow reassigning tasks to other org- and adminunits.

### LINKED ISSUE #5374: Erweiterung Aufgabe "Neu Zuweisen"

Gemäss https://feedback.onegovgever.ch/t/erweiterung-aufgabe-neu-zuweisen/892 soll folgende Anpassung vorgenommen werden:

Wird aktuell eine bereits erstellte Aufgabe neu zugewiesen, kann diese nur auf potentielle Auftragnehmer innerhalb des selben Mandanten zugewiesen werden. Dies soll auf eine mandantenübergreifende Zuweisung ausgeweitet werden.

Technische Spezifikation, Todos:
 - [x] `responsible` Feld im `IAssignSchema` wird auf alle Mandanten erweitert (`only_current_orgunit`, `only_current_inbox` flags)
 - [x] Zusätzlicher Validator für das Feld `responsible`: Bei bereits akzeptierten Aufgaben ist die Neu Zuweisung an einen anderen Mandanten (AdminUnit) nicht mehr möglich. Der Benutzer würde in diesem Fall mit einer entsprechenden Fehlermeldung informiert: "Bereits akzeptierte mandantenübergreifende Aufgaben können nicht an einen anderen Mandanten zugewiesen werden."

https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/634

phgross:
Extend the field validator, which makes sure that admin-unit changes are only possible when the task is in the open state. Registering two `SimpleFieldValidators` for the same field, does not worked, therefore I combined the two checks in one validator.

Because we do not use the exact same interface in the form than the transition extender, which is used by API requests, we need to add separate validators for each field of the transition extender interface. 

I also realized that the BaseTransitionExtender did not correctly validates the values (SimpleFieldValidators has not been called in the former implementation), I've fixed that as well.

Closes #5374

### DIFF
--- a/opengever/task/transition.py
+++ b/opengever/task/transition.py
@@ -8,6 +8,8 @@ from opengever.ogds.base.sources import AllUsersInboxesAndTeamsSourceBinder
 from opengever.task import _
 from opengever.task.activities import TaskAddedActivity
 from opengever.task.activities import TaskReassignActivity
+from opengever.task.browser.assign import validate_no_admin_unit_change
+from opengever.task.browser.assign import validate_no_teams
 from opengever.task.browser.delegate.metadata import IUpdateMetadata
 from opengever.task.browser.delegate.recipients import ISelectRecipientsSchema
 from opengever.task.browser.delegate.utils import create_subtasks
@@ -19,6 +21,7 @@ from opengever.task.task import ITask
 from opengever.task.util import add_simple_response
 from plone import api
 from plone.supermodel.model import Schema
+from z3c.form import validator
 from z3c.relationfield.relation import RelationValue
 from z3c.relationfield.schema import RelationChoice
 from z3c.relationfield.schema import RelationList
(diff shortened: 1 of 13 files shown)


opengever/task/transition.py
```diff
@@ -88,6 +92,46 @@ class INewResponsibleSchema(Schema):
         required=True)
 
 
+class NoTeamsInProgressStateValidator(validator.SimpleFieldValidator):
```
lukasgraf: As discussed, it would be nice if you could avoid completely duplicating the validation logic here. It should be possible to split out most of the core validation logic to a plain Python function, and reuse it from the assign-form validators and transition extender validators (putting handling for `responsible_client, responsible = value.split(':')` outside that function).

phgross: @lukasgraf updated, moved validation in to separte methods to avoid duplication (changes amended to the last commit)

PR #1967, opened 2016-06-24
Add previews on versions tab

elioschmutz:
This PR add the bumblebee-previews for the document-versions:

bildschirmfoto 2016-06-24 um 16 50 53

closes #1670

PR #5505, opened 2019-03-19
Display a portal warning on the overview of trashed documents

### LINKED ISSUE #5481: Display hint for trashed documents on document overview and in overlay

- Display warning on document overview if the document is trashed. This should be a plone status message.
- Display a warning in the document overlay, this could be implemented similar to what we already do when viewing older document versions

https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/637

Rotonen:
* Adding a portal warning onto the tabbedview of a document or a mail
  * Only if the request is a 'simple get'
* Adding a warning to bumblebee overlays for documents and mails
* Added a versioned attribute to the versioned overlay as the adapter loses the context proper

Document overview:
Screenshot 2019-03-25 at 15 46 29

Document overview overlay:
Screenshot 2019-03-25 at 15 46 40

Document version listing overlay of current version:
Screenshot 2019-03-25 at 15 47 05

Document version listing overlay of old version:
Screenshot 2019-03-25 at 15 47 12

Closes #5481

### DIFF
--- a/opengever/document/browser/tabbed.py
+++ b/opengever/document/browser/tabbed.py
@@ -1,5 +1,8 @@
 from ftw.tabbedview.browser.tabbed import TabbedView
+from opengever.document import _
 from opengever.document.interfaces import ICheckinCheckoutManager
+from opengever.trash.trash import ITrashed
+from plone import api
 from zope.component import getMultiAdapter
 
 
@@ -7,6 +10,22 @@ class DocumentTabbedView(TabbedView):
     """Tabbedview for documentish types, implements a separate check for the
     uploadbox, to allow drag and drop replacing of a file.
     """
+
+    def __init__(self, context, request):
+        """Slap a warning onto the overview of a trashed document.
+
+        We're asserting on the request not having a form as the tabs themselves,
+        which get requested by AJAX, rely on a form in the request data. If
+        we'd also slap the portal warning onto those requests, the next 'full'
+        page view would display them, as the tabs do not consume a portal
+        warning.
+        """
+        super(DocumentTabbedView, self).__init__(context, request)
+        if ITrashed.providedBy(self.context):
+            if not self.request.form:
+                msg = _(
+                    u'warning_trashed', default=u'This document is trashed.')
+                api.portal.show_message(msg, self.request, type='warning')
 
     def show_uploadbox(self):
         """Only checks if a file_upload is available, means document is
--- a/opengever/document/locales/de/LC_MESSAGES/opengever.document.po
+++ b/opengever/document/locales/de/LC_MESSAGES/opengever.document.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
-"POT-Creation-Date: 2018-11-08 13:36+0000\n"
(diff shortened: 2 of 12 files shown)


opengever/document/browser/tabbed.py
```diff
@@ -8,6 +11,24 @@ class DocumentTabbedView(TabbedView):
     uploadbox, to allow drag and drop replacing of a file.
     """
 
+    def __init__(self, context, request):
```
deiferni: why not __call__, __init__ may be a bit early in the lifecycle 🤔.

Rotonen: I recoiled from doing that as the `__call__` of the parent is overriding the `__call__` of the Five `BrowserView` in a weird way and I've never had a side effect free ride when interfacing with a `ViewPageTemplateFile` bouncing around in the publisher.

https://github.com/4teamwork/ftw.tabbedview/blob/649f6f3e05e1241374cd05f1ee2bdeadd12affad/ftw/tabbedview/browser/tabbed.py#L37

I'll give it a shot.

Rotonen: And the publisher loses the context if I try that. Sticking with `__init__` here.

Rotonen:
* Added missing overlay functionality
* Also now tagging mails

Rotonen:
@deiferni thank you. I completely missed the overlay adapter mangles the overlay context on versioned documents.

That seems to be a side effect of https://github.com/4teamwork/opengever.core/pull/1967, but I'm not certain if it's intentional. In any case one can work around this as the original implementation has worked around the retrival of the version ID as well.

* Fixed version listing overlays
* Made the overlay warning position consistent

Also added screenshots to the main PR body.

opengever/document/locales/de/LC_MESSAGES/opengever.document.po
```diff
@@ -694,6 +694,11 @@ msgstr "Dieses Dokument kann nicht zu einem PDF konvertiert werden"
 msgid "version_id"
 msgstr "Version"
 
+#. Default: "This document is trashed."
+#: ./opengever/document/browser/tabbed.py
+msgid "warning_trashed"
+msgstr "Das Dokument ist im Papierkorb."
```
deiferni: sorry for mentioning this only now but i think the translation could be improved slightly:

Das Dokument ist im Papierkorb -> Dieses Dokument befindet sich im Papierkorb.

Same for the mails.

opengever/mail/locales/de/LC_MESSAGES/opengever.mail.po
```diff
@@ -244,3 +244,8 @@ msgstr "E-Mails"
 #: ./opengever/mail/browser/send_document.py
 msgid "receiver"
 msgstr "Empfänger"
+
+#. Default: "This mail is trashed."
+#: ./opengever/mail/browser/tabbed.py
+msgid "warning_trashed"
+msgstr "Das E-Mail ist im Papierkorb."
```
deiferni: @Rotonen unfortunately this translation was not updated correctly as requested in https://github.com/4teamwork/opengever.core/pull/5505#discussion_r268990657

PR #5506, opened 2019-03-19
Add option not to revoke permissions associated with a task when closing it

### LINKED ISSUE #5399: OGIP 53 - Berechtigungsentzug bei Aufgaben aussetzen
_Harvest: `OneGov GEVER` - `OGIP 53 - Berechtigungsentzug bei Aufgaben aussetzen`_

Umsetzung gemäss OGIP https://gever.4teamwork.ch/ordnungssystem/3/2/2/5/dossier-3072/document-26852

njohner:
When tasks transition into their final state, assignments through the task get revoked. Here we add an option to avoid revoking these permissions. For that we:
* Add a new field `revoke_permissions`, defaulting to `True`
* Only revoke the permissions if `revoke_permissions == False`
* Add an action to revoke the assignment permissions. The action is only available on tasks in their final state and for `manager`s and the task issuer
* I added a message to warn the user that permissions will not be revoked on the final transition when  `revoke_permissions == False`
* Add a feature flag to control whether the optional revoking is available or not. When the feature is disabled:
    * The corresponding field is hidden in the form
    * The action is not shown on closed tasks
    * The view cannot be called directly
* I added a message shown when the action to revoke the permissions is executed successfully, otherwise the user has no feedback that anything happened at all. 
* I did NOT write an upgrade step to set the field on all existing tasks. For now those will have the default value for the field, which is the expected behaviour. The field will get set by Lukas'script to set default values for all fields on all objects...

**When creating a task the field is available (when the feature flag is set to True)**
Screen Shot 2019-03-27 at 09 21 12

**Closing a task with `revoke_permissions` set to `False` will state in an info message that permissions will not be revoked**
Screen Shot 2019-03-27 at 09 23 20

**On a closed task, there is an action to revoke the permissions. Note that this action is also available on tasks where `revoke_permissions` set to `True`, so where permissions normally have been revoked when closing the task.**
Screen Shot 2019-03-27 at 09 24 11

**After revoking the permissions on task through the action, an info message is displayed to inform the user the action was performed.**
Screen Shot 2019-03-27 at 10 14 12

resolves #5399


opengever/task/tests/test_localroles.py
```diff
@@ -338,8 +410,44 @@ def test_closing_a_task_revokes_responsible_roles_on_distinct_parent(self, brows
...
+        # close subtask
+        browser.open(self.meeting_subtask, view='tabbedview_view-overview')
+        browser.click_on('task-transition-resolved-tested-and-closed')
+        browser.click_on('Save')
+
+        browser.open(self.meeting_task, view='tabbedview_view-overview')
+        browser.click_on('task-transition-resolved-tested-and-closed')
+        browser.click_on('Save')
+
+        self.assertEqual([], storage._storage())
```
phgross: So it revokes the roles on distinct parent 🤔, thats not what the test promise.

opengever/task/locales/de/LC_MESSAGES/opengever.task.po
```diff
@@ -679,6 +689,11 @@ msgstr "Status"
 msgid "msg_deadline_change_successfull"
 msgstr "Frist wurde erfolgreich angepasst."
 
+#. Default: "Temporary permissions will not be revoked by this transtion as revoke permissions is unchecked for this task"
+#: ./opengever/task/response.py
+msgid "msg_not_revoking_permissions"
+msgstr "Auf dieser Aufgabe ist eingestellt, dass die temporäre Berechtigungen für den Auftragnehmer und dessen Stellvertretung nicht entzogen werden."
```
phgross: Die temporäre Berechtigungen für den Auftragnehmer und dessen Stellvertretung bleiben bestehen, der Entzug der Berechtigungen ist für diese Aufgabe deaktiviert.

opengever/task/tests/test_localroles.py
njohner: 😓 Had set the field to False on the wrong task! Sorry

njohner:
Ok so I included all your comments. The major changes are:
* Add a feature flag to control whether the optional revoking is available or not. When the feature is disabled:
    * The corresponding field is hidden in the form
    * The action is not shown on closed tasks
    * The view cannot be called directly
* I added a message shown when the action to revoke the permissions is executed successfully, otherwise the user has no feedback that anything happened at all.

njohner:
I noticed two edge cases:
* `UserA` creates a task in `DossierA`, then a manager revokes the users permission on that dossier. `UserA` can still access the task (because of local role assignments). Let's say the task was created with `revoke_permissions=False`. When the task is closed, permissions remain. `UserA` goes to the task and clicks the `revoke_permissions` action, locking himself out of the task. As he is redirected to the task after the action he gets an "Insufficient privileges" page.
* Now a manger gives `Reader` permissions to `UserA` on `DossierA`. He can access the task, and he could trigger the `revoke_permissions` action again, as he is the issuer, but he does not see the action, because the Edit bar is not shown to the user (probably because he has no edit rights)

njohner:
Following our discussion I updated the PR with the following features:
* Add fallback for user losing access rights when revoking roles on task (he gets redirected to site root and a special message is displayed)
* Added an invariant to avoid users setting `revoke_permissions` through the API when the feature is disabled.
    * The invariant is written a bit differently than the one for `is_private`. I was having issues when `is_private` is set to `False` when hidden, it does not appear at all in the `data` in the invariant (the `getattr` fallback raises an error)
    * This also allowed to remove the automatic setting of `revoke_permissions` in the add form when creating a task. A correct value already being ensured by the invariant
* I also noticed that I had forgotten to hide the `revoke_permissions` field in the edit form when the feature is disabled. I also fixed this.

opengever/task/locales/de/LC_MESSAGES/opengever.task.po
```diff
@@ -221,6 +221,11 @@ msgstr "Teams sind nur bei offenen Aufgaben oder Weiterleitungen als Auftragnehm
 msgid "error_private_task_feature_is_disabled"
 msgstr "Die Funktionalität private Aufgaben ist deaktiviert."
 
+#. Default: "The revoke permissions feature is disabled"
+#: ./opengever/task/task.py
+msgid "error_revoke_permissions_feature_is_disabled"
+msgstr "Die Funktionalität Berechtigungen nach Abschluss optional entziehen ist deaktiviert."
```
phgross: Die Funktionalität Berechtigungen nach Abschluss optional zu entziehen ist deaktiviert.

PR #5517, opened 2019-03-27
Add solrsearch REST API endpoint.

### LINKED ISSUE #5406: Solr REST Search Endpoint
_Harvest: `Projekt - Umsetzung GEVER-RIS` - `GEVER API`_

RIS möchte nach Dokumenten in GEVER suchen können, z.B. nach Dokumentart oder Volltext. Wir haben dort Solr im Einsatz für GEVER.

Da aber der `@search`-Endpoint nur mit `portal_catalog` funktioniert benötigen wir hier eine Lösung, wie wir den Endpoint mit Solr kompatibel machen.

Der Vorschlag ist, ein neuer `@solr-search`-Endpoint zu bauen aus folgenden Gründen:
- `@search` funktioniert in einem Solr-Setup solange nicht `SearchableText` verwendet wird; das heisst wenn man mit und ohne Solr kompatibel sein will und keine Volltextsuche absetzt ist der bestehende `@search` Endpoint immer noch sinnvoll.
- Beim neuen Endpoint können wir mehr Solr-Query-Syntax verwenden und von der Katalog-Syntax abweichen und so mehr Möglichkeiten anbieten.

https://plonerestapi.readthedocs.io/en/latest/searching.html

@deiferni mir war nicht bewusst dass dies noch ein Problem ist. Können wir das schätzen und herausfinden, bis wann das in einen Release passt?
/cc @buchi

phgross:
The endpoint passes the parameter to ftw.solr's SolrSearchUtiliy, so we don't have to care about the security index and preparing the different parameters.

If solr is not enabled the endpoints raises a BadRequest exception.

Closes #5406

### DIFF
--- a/opengever/api/tests/test_solrsearch.py
+++ b/opengever/api/tests/test_solrsearch.py
@@ -1,0 +1,111 @@
+from ftw.testbrowser import browsing
+from opengever.testing import IntegrationTestCase
+
+
+class TestSolrSearchGet(IntegrationTestCase):
+
+    # TODO: Replace mock tests using a real solr as soon this
+    # is possible (see #4844).
+
+    api_headers = {'Accept': 'application/json'}
+    features = ['solr']
+
+    solr_search_response = {
+        "responseHeader": {
+            "status": 0,
+            "QTime": 3,
+            "limit": 10,
+            "params": {
+                "json": "{\n  \"query\": \":\"\n}"
+            }
+        },
+        "response": {
+            "numFound": 3,
+            "start": 0,
+            "docs": [
+                {
+                    "UID": "85bed8c49f6d4f8b841693c6a7c6cff1",
+                    "Title": "My Folder 1",
+                },
+                {
+                    "UID": "85a29144758f494c88df19182f749ed6",
+                    "Title": "My Folder 2",
+                },
+                {
+                    "UID": "9398dad21bcd49f8a197cd50d10ea778",
+                    "Title": "My Document",
+                }
+            ]
+        }
+    }
+
+    @browsing
+    def test_raises_bad_request_if_solr_is_not_enabled(self, browser):
+        self.deactivate_feature('solr')
+
+        self.login(self.regular_user, browser=browser)
+
+        with browser.expect_http_error(400):
+            url = u'{}/@solrsearch'.format(self.portal.absolute_url())
+            browser.open(url, method='GET', headers=self.api_headers)
+
+        self.assertEqual(
+            {u'message': u'Solr is not enabled on this GEVER installation.',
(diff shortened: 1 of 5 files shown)


opengever/api/tests/test_solrsearch.py
```diff
@@ -0,0 +1,122 @@
...
+        self.solr.search.assert_called_once_with(fl='UID,Title')
+
+    @browsing
+    def test_returns_json_serialized_solr_docs(self, browser):
+        self.login(self.regular_user, browser=browser)
+
+        self.solr = self.mock_solr(response_json=self.solr_search_response)
+
+        url = u'{}/@solrsearch?fl=UID,SearchableText,Title'.format(
+            self.portal.absolute_url())
```
njohner: You probably don't want `SearchableText` in the field list here

opengever/api/tests/test_solrsearch.py
```diff
@@ -0,0 +1,122 @@
...
+        browser.open(url, method='GET', headers=self.api_headers)
+
+        self.assertEqual(
+            [{u'path': u'/my-folder-1',
+              u'UID': u'85bed8c49f6d4f8b841693c6a7c6cff1',
+              u'Description': u'',
+              u'Title': u'My Folder 1'},
```
njohner: Why are not only the selected fields being returned (i.e. `UID` and `Title`)?

opengever/api/tests/test_solrsearch.py
phgross: I do, because I want to test that the unsupported field `SearchableText` is removed from the list. 

I'll add a comment to the test, that `SearchableText` is an unsupported field.

phgross:
@njohner Updated the PR according to your feedback, thx for the review. 

There is nothing broken with the `fl` parameter, just the mocked results had been wrong. I've fixed them now. Because we currently can't test against a real solr, I relay on the fact, that the SearchUtility from ftw.solr works as expected. So I only check in the tests, that the field parameter is passed in correctly.

opengever/api/tests/test_solrsearch.py
njohner: Oh I just realized, this is a mock test 😓 That's it was just returning whatever you make it return...

opengever/api/tests/test_solrsearch.py
njohner: Isn't the preivous test `test_not_supported_fields_are_skipped` testing for that? Of course you can also make sure that the skipping works as expected in the response...

phgross: 🙈ahh ... Never mind I already updated the fl parameters in the other test ...

PR #5520, opened 2019-03-28
Prevent creating private task or switching a task to private when feature is disabled.

### LINKED ISSUE #5441: Prevent tasks from being created as private or switched to private when feature is not enabled
_Harvest: `OneGov GEVER` - `Feature: Persönliche Aufgaben - OGIP 46`_

Follow up from #5332 make sure the creation or switch to a private change is not possible when the feature is disabled, also via REST API.

phgross:
I added an invariant to the ITask schema, thats imho the only way to have a schema validation, which is also called on api POST/PATCH requests.

Because invariants get an `z3c.form.validator.Data` object passed in, which provides his own getattr with a fallback, fetching the value from the context when the field was not passed in via form/api , I had to check if the is_private value was part of the input-data with the help of the attribute `_Data_data___`. I don't see any other way to check if the field is passed in or if it's already set on the object.

Besides that Joni did some testing and cleanup work, thx ...

Closes #5441

### DIFF
--- a/opengever/dossier/tests/test_overview.py
+++ b/opengever/dossier/tests/test_overview.py
@@ -35,7 +35,8 @@ class TestOverview(IntegrationTestCase):
 
     @property
     def tested_subtask(self):
-        return self.subtask
+        # XXX - not expected, but a quick way around the visibility limit
+        return self.private_task
 
     @property
     def participants(self):
@@ -44,6 +45,17 @@ class TestOverview(IntegrationTestCase):
 
     @property
     def task_titles(self):
+        return [
+            u'Diskr\xe4te Dinge',
+            u'Vertragsentw\xfcrfe 2018',
+            u'Personaleintritt',
+            u'Mitarbeiter Dossier generieren',
+            u'Vertragsentwurf \xdcberpr\xfcfen',
+        ]
+
+    @property
+    def task_titles_minus_pending(self):
+        # As we have more than 5 tasks, they shift in like this
         return [
             u'Vertragsentw\xfcrfe 2018',
             u'Personaleintritt',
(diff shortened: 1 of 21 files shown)


Rotonen:
The test failure is related to the fixture amendment - the fixture and the relevant tests on this branch have been changed before that fixture dependent test was added to master.

opengever/dossier/tests/test_overview.py
```diff
@@ -45,10 +46,21 @@ def participants(self):
...
             u'Mitarbeiter Dossier generieren',
             u'Vertragsentwurf \xdcberpr\xfcfen',
+        ]
+
+    @property
+    def task_titles_minus_pending(self):
+        # As we have more than 5 tasks, they shift in like this
+        return [
+            u'Vertragsentw\xfcrfe 2018',
+            'Personaleintritt',
```
deiferni: why do we mix unicode and non-unicode strings here 🤔?

opengever/task/task.py
```diff
@@ -251,6 +253,17 @@ class ITask(model.Schema):
         title=_(u'label_predecessor', default=u'Predecessor'),
         required=False)
 
+    @invariant
+    def private_task_is_not_changed_when_disabled(data):
+        if not is_private_task_feature_enabled() and data.is_private:
+
+            # Because the z3c.form.validator.Data object has implemented a
+            # getattr fallback, which fetchs the value from the context, we
```
deiferni: typo `fetchs` `fetches`

opengever/dossier/tests/test_overview.py
Rotonen: I think copy pasting from pretty printing is the likely culprit: it prefers to use non-unicode strings for things which are fully representable in ASCII.

PR #5522, opened 2019-03-28
Fix filtering by multiple values in Solr

buchi:
Put filter queries in braces to properly support OR operations.


buchi:
Support for filtering in the listing endpoint (with this bug) has not been released yet. Thus no changelog entry.

Shouldn't be an issue with the solrsearch endpoint as it expects a filter query and doesn't build one itself.

PR #5531, opened 2019-04-03
Fix creation of scaneingang dossier in the scan-in endpoint.

### LINKED ISSUE #5529: Creation of `Scaneingang` dossier with scan-in endpoint not correct.
https://extranet.4teamwork.ch/support/gemeinde-koeniz/tracker/156

The private dossier creation in the `scan-in` endpoint, does not correctly fire the `ObjectCreatedEvent` which leads to objects without an UUID.

phgross:
the scan-in endpoint used the `create` helper from plone.restapi to create the private dossier `Scaneingang`, but this helper does not fires the CreatedEvent which leads to object without an UID. In combination with the recently introduced not-query on UIDs those dossiers are no longer listed in the dossier listing. 

I changed now the creation to `createContentInContainer` (the way we use for other in code content creation) and provide an upgradestep which fires (again) the ObjectCreatedEvent and reindex the object.

Closes #5529 

Needs backport to `2019.1-stable`

### DIFF
--- a/opengever/core/upgrades/20190403172610_fix_scaneingang_privatedossiers/upgrade.py
+++ b/opengever/core/upgrades/20190403172610_fix_scaneingang_privatedossiers/upgrade.py
@@ -1,0 +1,22 @@
+from ftw.upgrade import UpgradeStep
+from plone import api
+from zope.event import notify
+from zope.lifecycleevent import ObjectCreatedEvent
+
+
+class FixScaneingangPrivatedossiers(UpgradeStep):
+    """Fix scaneingang privatedossiers.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
+        catalog = api.portal.get_tool('portal_catalog')
+
+        dossiers = catalog(sortable_title='scaneingang',  # Exact match
+                           portal_type='opengever.private.dossier')
+
+        for brain in dossiers:
+            if not brain.UID:
+                dossier = brain.getObject()
+                notify(ObjectCreatedEvent(dossier))
+                dossier.reindexObject()
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -8,6 +8,7 @@ Changelog
 - Add archival file management view on dossier level. [njohner]
 - Show archival file state on documents overview for managers. [njohner]
 - Fix tests failing due to timezone leading to date shift. [njohner]
+- Fix creation of scaneingang dossier in the scan-in endpoint. [phgross]
 - Prevent tasks from being created as private or switched to private when feature is not enabled. [Rotonen, phgross]
 - Bump ftw.mail to 2.6.0 to get error logging on inbound mail failures. [lgraf]
 - Handle complex URLs as titles on journal PDF exports. [Rotonen]
--- a/opengever/api/scanin.py
+++ b/opengever/api/scanin.py
@@ -1,9 +1,8 @@
 from opengever.base.command import CreateDocumentCommand
 from opengever.base.security import elevated_privileges
 from plone import api
+from plone.dexterity.utils import createContentInContainer
 from plone.restapi.services import Service
-from plone.restapi.services.content.utils import add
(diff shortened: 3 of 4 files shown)


opengever/core/upgrades/20190403172610_fix_scaneingang_privatedossiers/upgrade.py
```diff
@@ -0,0 +1,22 @@
...
+class FixScaneingangPrivatedossiers(UpgradeStep):
+    """Fix scaneingang privatedossiers.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
+        catalog = api.portal.get_tool('portal_catalog')
+
+        dossiers = catalog(sortable_title='scaneingang',  # Exact match
+                           portal_type='opengever.private.dossier')
```
deiferni: could we limit depth to make sure we only search in the toplevel of each users private folder.

opengever/core/upgrades/20190403172610_fix_scaneingang_privatedossiers/upgrade.py
```diff
@@ -0,0 +1,22 @@
...
+
+        for brain in dossiers:
+            if not brain.UID:
+                dossier = brain.getObject()
```
deiferni: i have some paranoia here:
- theoretically there could be no entry in the `UID` index, but the object could have an `IUUID`. So firing the event should not be necessary in such cases.
- Should we also re-fire an `ObjectAddedEvent` or how can we know that the object is in a good state now? Can you safely re-fire an `ObjectCreatedEvent` for an already added object 🤔?
- We have discussed it yesterday i think, it might be worth to completely re-create a new object and move the content to make sure we are in a good state.

opengever/core/upgrades/20190403172610_fix_scaneingang_privatedossiers/upgrade.py
phgross: I don't see any need for that, we do nothing if the dossier provides an UID

opengever/core/upgrades/20190403172610_fix_scaneingang_privatedossiers/upgrade.py
phgross: > theoretically there could be no entry in the UID index, but the object could have an IUUID. So firing the event should not be necessary in such cases.

Thats handled by the IUUID handler which makes sure, that an object with an additional UID does not get a new uuid.

> Should we also re-fire an ObjectAddedEvent or how can we know that the object is in a good state now? Can you safely re-fire an ObjectCreatedEvent for an already added object 🤔?

No, the added event has been fired by the `add` method. Yes, locally tested, I don't see any issue there and the object worked well.

> We have discussed it yesterday i think, it might be worth to completely re-create a new object and move the content to make sure we are in a good state.
I'm not a fan of making an upgradestep prophylactic much more complex, than it has to be. There are 30 Objects affected right now. IMHO it's not worth it ...

deiferni: yes it has been fired once, that is clear. my question was could there have been anything initialized by a created event handler that would have been expected by an added event handler but was not there. so the added event handler did not do something it should have done. thus there is still something else fucked up even after this fix.

i'm assuming the answer is yes, so the real question is are we willing to take the risk 😉.

phgross: Yes i'm willing to take the risk, when considering the risk a migration-upgradestep would bring. But let me know I can also live with that, and implement such an migration/recreation of the content ...

phgross: @deiferni should I or not? we need to fix this problem ...

deiferni: yes, i can live with that 👍.

PR #5535, opened 2019-04-05
Fix root path exclusion performance

buchi:
Turns out that NOT queries perform very bad for indexes with a lot of unique values.
Thus we patch the ExtendedPathIndex now for root path exclusion.

- Patch ExtendedPathIndex to allow exclusion of root path
- Exclude root path within path query in tabbed view's catalog source and in the REST API's listing endpoint
- Remove no longer needed UUIDIndex patch for NOT support

### DIFF
--- a/opengever/api/listing.py
+++ b/opengever/api/listing.py
@@ -15,7 +15,6 @@ from plone.restapi.batching import HypermediaBatch
 from plone.restapi.serializer.converters import json_compatible
 from plone.restapi.services import Service
 from plone.rfc822.interfaces import IPrimaryFieldInfo
-from plone.uuid.interfaces import IUUID
 from Products.CMFCore.utils import getToolByName
 from Products.ZCatalog.Lazy import LazyMap
 from Products.ZCTextIndex.ParseTree import ParseError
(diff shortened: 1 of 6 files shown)


opengever/api/listing.py
```diff
@@ -221,15 +220,16 @@ def catalog_results(self, name, term, start, rows, sort_on, sort_order,
 
         query = CATALOG_QUERIES[name].copy()
         query.update({
-            'path': '/'.join(self.context.getPhysicalPath()),
+            'path': {
+                'query': '/'.join(self.context.getPhysicalPath()),
+                'depth': -1,
```
lukasgraf: depth: -1 [is the default](https://github.com/plone/Products.ExtendedPathIndex/blob/a3d73d2768da413ca2ac0febfba9d1050fd4e341/Products/ExtendedPathIndex/ExtendedPathIndex.py#L356) - is there any particular reason to specify it here, or is this a leftover from testing?

opengever/base/monkey/patches/extendedpathindex.py
```diff
@@ -0,0 +1,44 @@
...
+                set_func = union
+            else:
+                set_func = intersection
+
+            result = None
+            for k in record.keys:
+                rows = self.search(k, level, depth, navtree, navtree_start,
+                                   resultset=resultset)
+                if exclude_root:
+                    root = self._index_items[k]
```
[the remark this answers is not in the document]
buchi: Fixed.

opengever/api/listing.py
buchi: No, just "explicit is better than implicit".

PR #5542, opened 2019-04-09
Add document_type index.

### LINKED ISSUE #5444: Add document_type to Solr schema
_Harvest: `Projekt - Umsetzung GEVER-RIS` - `aktueller Sprint`_ 

In RIS, I'd like to mark certain documents as [`RRB`](https://www.informatik.sg.ch/tools/glossar.handle__%7Bslash%7Dcontent%7Bslash%7Dapplikationen%7Bslash%7Dglossar%7Bslash%7Dinformatik%7Bslash%7Drrb.html).

The RIS application would set the `document_type` when it uploads the document through the REST API.

Later, we want to be able to find all RRB documents for one year. In order to do that, we would query the catalog for the `document_type` through the new [`@solr-search`](https://github.com/4teamwork/opengever.core/issues/5406) API endpoint.

### LINKED ISSUE #5548: Bug with solr indexing when index value is None

When indexing a single index, when the value of that index is `None`, solr will overwrite the whole entry by an empty one.

njohner:
This needs https://github.com/4teamwork/ftw.solr/pull/128, otherwise the upgrade will overwrite all documents with no `document_type` in solr and replace them with empty items.

I added an index for the `document_type` and also added it in the solr schema. You can now query for a specific `document_type` through the API:
```@solrsearch?query=document_type:offer```

Note that we are indexing the tokens, so you might want/need to fetch the tokens for a given value (in a given language). The vocabulary can be retrieved from:
```@vocabularies/opengever.document.document_types```

The titles for the tokens you get back seem to depend on the language set in the request to the `@vocabularies` endpoint.

The vocabulary itself, which will need to be overwritten if you need specific `document_type`s that are not in the default vocabulary is found here:
https://github.com/4teamwork/opengever.core/blob/master/opengever/document/vdexvocabs/document_types.vdex

resolves #5444 
resolves #5548

### CODE REVIEW

File: opengever/document/config.py [hunk 1]
```diff
@@ -5,6 +5,7 @@
          'lexicon_id': 'plone_lexicon'}),
     ('checked_out', 'FieldIndex'),
     ('document_date', 'DateIndex'),
+    ('document_type', 'FieldIndex'),
```
lukasgraf: Why a `FieldIndex` and not a `KeywordIndex`?

njohner: As discussed, `KeywordIndex` does not seem more efficient for a small set of possible values and is meant for a sequence of items.

PR #5553, opened 2019-04-11
 Fix solr error during copy/paste of word document.

### LINKED ISSUE #5547: Problem with solr indexing during copy/paste operation

Steps to reproduce:
* create a dossier containing a word document with docproperties
* copy and paste the dossier

The following error will be raised:
```
2019-04-05 16:42:49 ERROR Products.ZCatalog A different document with value '304111e4a4704fd6b346e7a3beb207a0' already exists in the index.'
2019-04-05 16:42:50 ERROR txn.123145386012672 Error in after commit hook exec in <function hook at 0x120858320> 
  File "/Users/njohner/.buildout/cache/eggs/transaction-1.1.1-py2.7.egg/transaction/_transaction.py", line 416, in _callAfterCommitHooks
    hook(status, *args, **kws)
  File "/Users/njohner/.buildout/cache/eggs/ftw.solr-2.3.1-py2.7.egg/ftw/solr/connection.py", line 115, in hook
    file_ = blob.committed()
  File "/Users/njohner/.buildout/cache/eggs/ZODB3-3.10.7-py2.7-macosx-10.12-x86_64.egg/ZODB/blob.py", line 202, in committed
    raise BlobError('Uncommitted changes')
BlobError: Uncommitted changes
```
@deiferni could also reproduce it on a fresh GEVER deployment and we found that this error was also present on productive systems. The end result is that the document is not indexed in solr.

see also https://github.com/4teamwork/opengever.core/pull/5533.

njohner:
Docproperties get updated several times during a copy/paste operation of a word document, overwriting the blob file each time. Each time the blob file is changed, a reindex operation is queued in solr, which is processed after the transaction ends. When that happens, solr tries to process some blobs that do not exist anymore and fails. We now update the docproperties only once during a copy/paste operation.

I don't think I can write a test for this, as long as we do not have a real solr running for testing purposes. For now I've written a test making sure that my changes did not break the update of the docproperties when copy/pasting a dossier containing a document. I added a comment in the test to also test sorl indexing of the object once we can.

**Here some notes on the analysis of the problem:**
So basically the problem comes from the renaming. In the copy/paste, we first create the new objects
and then rename them. During renaming we recursively rename all objects, but renaming of folderish objects also propagate the `ObjectMovedEvent` down to their children. For each `ObjectMovedEvent`, we update the docproperties, which means overwriting the blob file.

So when copying a dossier containing a document we updated the docproperties:
- When the document is added
- When renaming the dossier
- When renaming the document

The update of the docproperties when the document is added does not cause any problems (I think because the docproperties did not change, so the blob does not get overwritten, that's why copy/pasting a word document actually works). The one causing the issue is when renaming the dossier, as there we overwrite the blob of the document, then overwrite it again when renaming the document, but solr wants to index both blobs...

resolves #5547

### DIFF
--- a/opengever/document/tests/test_handlers.py
+++ b/opengever/document/tests/test_handlers.py
@@ -12,6 +12,7 @@ from plone import api
 from plone.app.testing import TEST_USER_ID
 from zipfile import BadZipfile
 from zope.component import getMultiAdapter
+from ftw.testbrowser import browsing
 
 
 class TestHandlers(FunctionalTestCase):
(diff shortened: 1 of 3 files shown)


opengever/document/tests/test_handlers.py
```diff
@@ -135,9 +136,13 @@ def test_document_checkin_updates_doc_properties(self):
...
 
-    def test_copying_documents_updates_doc_properties(self):
-        api.content.copy(source=self.doc_with_gever_properties,
-                         target=self.target_dossier)
+    @browsing
+    def test_copying_documents_updates_doc_properties(self, browser):
+        browser.login()
+        browser.open(self.doc_with_gever_properties, view='copy_item')
+        browser.open(self.target_dossier, view='tabbed_view')
+        browser.css('#contentActionMenus a#paste').first.click()
```
njohner: Copy/paste operation through the browser is different than through the api, we should stay close to what actually happens in normal operation here

deiferni:
ok, so, but but shouldn't we also tackle this in `ftw.solr`? we have now avoided the symptoms in this case, but the problem could surface again.

njohner:
Hum, you might be right, but how do we know in `ftw.solr` which request should be ignored or not? This looks like risky business, where we might end up not indexing certain files that we actually should... The other question this raises is catching such errors better. I mean, I stumbled on this one randomly, but basically solr was throwing an error, we were not aware of (at least not that I know) and some documents were not indexed...

PR #5122, opened 2018-11-19
🚧 WIP - Use a @restapi decorator from ftw.testbrowser for the api tests

Rotonen:
Creating this PR as an evaluation point for the pending `ftw.testbrowser` side PR at this point.

PR #5556, opened 2019-04-12
Fix broken rest endpoint to add workspace

elioschmutz:
⚠️ dieser PR wird durch #5563 überflüssig. Bitte noch warten mit Mergen... ⚠️ 

Dieser PR behebt ein Problem, bei dem es nicht möglich war, einen Arbeitsraum über die REST-API zu erstellen.

Siehe Kommentar im Code für weitere Informationen.

Issuer: https://github.com/4teamwork/gever-ui/issues/16

~ℹ️ Dieser PR fügt u.a. ein neues Test-Objekt hinzu: `RestAPIHandler`. Mit dem `RestAPIHandler` können REST-API einfacher getestet werden. Der Handler ist bereits in einem anderen Projekt erfolgreich im Einsatz (https://github.com/4teamwork/bernmobil.instructions/blob/master/bernmobil/instructions/tests/__init__.py#L17)~

~Diese Erweiterung hat keinen Einfluss auf bestehende Tests. Ein Refactoring wäre jedoch angebracht.~

### DIFF
--- a/opengever/workspace/utils.py
+++ b/opengever/workspace/utils.py
@@ -23,6 +23,10 @@ def roles_of_permission(context, permission):
 
 def is_within_workspace(context):
     """ Checks, if the content is a workspace or is within the workspace.
+
+    This function is required for the ogds-vocabularies. If the current context
+    is within a workspace, a special workspace-query hook will be used for
+    extending the query with workspace specific functionality.
     """
     # Avoid circular imports
     from opengever.workspace.interfaces import IWorkspace
--- a/opengever/workspace/utils.py
+++ b/opengever/workspace/utils.py
@@ -2,6 +2,8 @@ from AccessControl.interfaces import IRoleManager
 from AccessControl.Permission import Permission
 from Acquisition import aq_base
 from Acquisition import aq_chain
+from Acquisition import aq_parent
+from Acquisition import aq_inner
 from operator import itemgetter
 from plone import api
 
@@ -24,6 +26,25 @@ def is_within_workspace(context):
     """
     # Avoid circular imports
     from opengever.workspace.interfaces import IWorkspace
+
+    if context.getId() not in aq_parent(aq_inner(context)):
+        # Creating a new object through the REST-API will create a temporary
+        # object first. Then it validates all the field values and adds the
+        # new context to the container.
+        #
+        # While validating the schema, the context will be the temporary
+        # workspace instead the container object which is the assumed context.
+        #
+        # This behavior differs to content-creation without the REST-API. So we
+        # have to change the context here if we have a temporary context.
+        #
+        # Unfortunately, there is no other way to detect a temporary object
+        # than checking, if the current object is already added to the parent
+        # object.
+        #
+        # See: https://github.com/plone/plone.restapi/blob/3.7.2/src/plone/restapi/services/content/add.py#L83
+        context = aq_parent(aq_inner(context))
+
     return bool(filter(IWorkspace.providedBy, aq_chain(context)))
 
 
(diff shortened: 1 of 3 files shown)


opengever/testing/restapi.py
```diff
@@ -0,0 +1,69 @@
+from ftw.testbrowser import browser as default_browser
+from plone import api
+from plone.app.testing import TEST_USER_PASSWORD as DEFAULT_PASSWORD
+import json
+
+
+class RestAPIHandler(object):
```
phgross: I don't see really a need to have an additional way to login/authenticate in tests than the one we introduced with the IntegrationTestsCase and the testing fixtures. Or did I miss some benefits of the RestAPIHandler object?

Mi guess, having api_headers on the IntegrationTestCase would solve all your problems:

```
class IntegrationTestCase(TestCase):
    layer = OPENGEVER_INTEGRATION_TESTING
    features = ()

    api_headers = {
        'Accept': 'application/json',
        'Content-Type': 'application/json'}
```

Example from the solrsearch tests
```
    @browsing
    def test_raises_bad_request_if_solr_is_not_enabled(self, browser):
        self.deactivate_feature('solr')

        with browser.expect_http_error(400):
            url = u'{}/@solrsearch'.format(self.portal.absolute_url())
            browser.open(url, method='GET', headers=self.api_headers)

        self.assertEqual(
            {u'message': u'Solr is not enabled on this GEVER installation.',
             u'type': u'BadRequest'}, browser.json)

```

By the way Joni did some work on providing some api_requests method on ftw.testbrowser, which would make the api call even easier.

opengever/workspace/utils.py
```diff
@@ -24,6 +26,25 @@ def is_within_workspace(context):
     """
     # Avoid circular imports
     from opengever.workspace.interfaces import IWorkspace
+
+    if context.getId() not in aq_parent(aq_inner(context)):
+        # Creating a new object through the REST-API will create a temporary
```
phgross: The problem description is useful and understandable, but what I miss is a docstring what `is_withing_workspace` used for, the current docstring only describe what it does but not what this method is used for. that would help to understand this fallback.

opengever/testing/restapi.py
Rotonen: Look at the office connector tests.

elioschmutz: @Rotonen I can't find such tests... could you provide a link to an explicit example?

@phgross as you want... for me, the current implementation is too noisy and requires too much configuration. Using an own object for handling restapi requests let us use specific restapi-functions and it is much more readable... The new Login/Auth is like the real front-end is using the GEVER.

Please, decide... leave or discard?

Rotonen: https://github.com/4teamwork/opengever.core/blob/master/opengever/officeconnector/testing.py#L110-L140

https://github.com/4teamwork/opengever.core/blob/master/opengever/officeconnector/tests/test_api_dossier_attach.py#L23-L59

elioschmutz: Thanks... but this is not what I try to solve with the `RestAPIHandler`.

Rotonen: There are whole categories of things you cannot test with the traversal driver with the normal auth - some things are different when doing a forced JWT auth - especially so when testing across org units.

The so far stalled approach on the ftw.testbrowser side:
https://github.com/4teamwork/opengever.core/pull/5122
https://github.com/4teamwork/ftw.testbrowser/pull/89

The gever side is a demo of the desired usability. The ftw.testbrowser side implementation is an ugly hack @jone did not like too much yet.

elioschmutz: Ok... I see... there is too much going on. I'll remove my implementation and use the current approach until there is a useful solution ready.

elioschmutz:
@phgross I also extended the docstring of the `is_withing_workspace` function . Could you take another look?

elioschmutz:
Wird durch PR #5578 überflüssig.

PR #5564, opened 2019-04-15
Allow non-member contributors to use the REST API

Rotonen:
* Amended the fixture to be a two org unit setup
* Non-`Member` `Contributor`s can now use the REST API

Closes https://github.com/4teamwork/opengever.core/issues/5541

### DIFF
--- a/opengever/base/tests/test_role_assignments.py
+++ b/opengever/base/tests/test_role_assignments.py
@@ -105,6 +105,10 @@ class TestRoleAssignmentManager(IntegrationTestCase):
             [{'principal': self.regular_user.id,
               'roles': ['Editor'],
               'cause': ASSIGNMENT_VIA_TASK,
+              'reference': Oguid.for_object(task).id},
+             {'principal': u'fa_inbox_users',
+              'roles': ['Editor'],
+              'cause': ASSIGNMENT_VIA_TASK_AGENCY,
               'reference': Oguid.for_object(task).id}],
             RoleAssignmentManager(task).storage._storage())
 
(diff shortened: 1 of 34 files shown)


Rotonen:
On some of the gnarlier testing deltas (local roles / assignations), to get this done in a somewhat sane amount of time, I've used black on the test file in question so I can simply copy/paste output from the console and run black on the file instead of spending time manually fiddling the relevant list-dict structures neatly into the test file.

opengever/base/tests/test_role_assignments.py
```diff
@@ -7,255 +7,432 @@
...
 class TestRoleAssignmentManager(IntegrationTestCase):
-
     def test_add_register_assignment_in_the_storage(self):
         self.login(self.regular_user)
 
         manager = RoleAssignmentManager(self.empty_dossier)
-        manager.add_or_update(self.secretariat_user.id,
-                    ['Editor', 'Contributor', 'Reader'],
-                    ASSIGNMENT_VIA_TASK, self.task)
+        manager.add_or_update(
```
phgross: Why all this format changes? There is no need to do that and makes it harder to read the history! 

The review is unnecessary complex, please update first ...

Rotonen:
Now with maximally preserved local formatting conventions and other optimization for cherrypickability.

opengever/task/tests/test_localroles.py
```diff
@@ -4,20 +4,44 @@
...
 from z3c.relationfield.relation import RelationValue
 from zope.component import getUtility
 from zope.event import notify
 from zope.intid.interfaces import IIntIds
 from zope.lifecycleevent import ObjectModifiedEvent
 
 
-class TestLocalRolesSetter(IntegrationTestCase):
+class TestLocalRolesSetterFunctional(FunctionalTestCase):
+    # XXX - I found no other way to reset back to a lone org unit strategy
```
phgross: inactivate an org_unit and manually switch to the LoneOrgUnitStrategy wasn't working?

Rotonen: I did not manage to get it to actually inactivate and it thus kept forcing itself onto the multi org strategy - I'll still quickly try once again before merging.

Rotonen:
I've now attended to the testing curiosity and managed to pull it off.

PR #5566, opened 2019-04-16
Add nightly job runner

njohner:
We add a `NightlyJobRunner` class to execute nightly jobs (such as resolving dossiers, generating archive PDFs ...). The construction is as follows:
* `NightlyJobRunner`: It looks for all `INightlyJobProvider`s and executes the jobs provided by them. It will abort if outside of a given time window (defined in the registry `INightlyJobsSettings`) or if the system load is too high.
* `BaseJobProvider`: 
    * Base implementation of a job provider, which is meant to be subclassed (subclasses will have to be registered as implementing `INightlyJobProvider`).
    * Takes care of iteration and bookkeeping (how many jobs were executed, how many are left, etc.). 
    * subclasses have to implement the  
        * `execute_job` method, which will do the actual work 
        * `initialize` method which needs to provide a generator (or any iterable) of the jobs as well as the number of jobs to be executed. 
    * The class is designed to be flexible and it should be fairly easy to subclass it for a provider managing its jobs with a queue (for example remove job from queue in the `post_job_run` method)
* `NightlyJob` is the baseclass for the nightly jobs themselves. This basically does nothing except indicate that subclasses have to define the `provider_name` class attribute. This is used by the runner to know which provider to use to execute a particular job.
* `run_nightly_jobs` is the `zopectl.command` meant to be run by a cronjob.

An example implementation of a nightly job and a job provider can be found in the tests (`DocumentTitleModifierJobProvider` and `ModifyTitleJob`)

For system load monitoring we add `psutil` as dependency. For now monitoring is kept very simple. We could make the load limits registry settings?

I tested the whole implementation by registering a job provider and running jobs with the `run_nightly_jobs` command. The only thing that is not tested is the logging to sentry, as I don't know how to test it.

Also note that I had to overwrite an adapter for the registry to allow `Time` fields in the registry. I did not make a `registry` module for that, but maybe we should?

resoves #5554


opengever/nightlyjobs/runner.py
```diff
@@ -0,0 +1,119 @@
...
+        log_msg_to_sentry(message, request=getRequest())
+
+    @property
+    def njobs_remaining(self):
+        if not self.job_providers:
+            return 0
+        return sum(len(provider) for provider in self.job_providers.values())
+
+    @property
+    def njobs_total(self):
```
lukasgraf: I would lean towards tracking this state on the runner instead requiring the providers to do so.

So, instead of separately asking the providers to list their current job count and keep track of their total as well as executed job count, I would simply require them to provide a `__len__` (current job count).

The runner can then make sure to query the job providers once in the beginning for their `__len__` keep track of it itself (same for number of execute jobs - the runner doesn't need the provider's help to accomplish that).

opengever/nightlyjobs/tests/test_nightly_job_runner.py
```diff
@@ -0,0 +1,199 @@
...
+from opengever.testing import IntegrationTestCase
+from plone import api
+from Products.CMFPlone.interfaces import IPloneSiteRoot
+from zope.component import adapter
+from zope.interface import alsoProvides
+from zope.interface import implementer
+from zope.publisher.interfaces.browser import IBrowserRequest
+
+
+@implementer(INightlyJobProvider)
```
lukasgraf: Just a side note: Inheriting from a class that implements a particular interface means that the subclass will also be marked as implementing it (but could implement additional ones that the parent class doesn't).

opengever/nightlyjobs/tests/test_nightly_job_runner.py
```diff
@@ -0,0 +1,199 @@
...
+        self.assertTrue(runner.check_in_time_window())
+
+        # 6:00
+        self.clock.forward(seconds=3 * 3600)
+        self.assertFalse(runner.check_in_time_window())
+
+    def test_runner_aborts_when_not_in_time_window(self):
+        self.login(self.manager)
+        runner = NightlyJobRunner()
+        runner.check_in_time_window = lambda: False
```
lukasgraf: Nice - I like how readable these tests are and how they are structured! 💚

opengever/nightlyjobs/runner.py
```diff
@@ -81,15 +96,21 @@ def interrupt_if_necessary(self):
...
     def check_in_time_window(self):
         current_time = datetime.now().time()
         current_timedelta = self._positive_timedelta(self.window_start, current_time)
         return current_timedelta < self.window_length
 
+    def _is_cpu_overladed(self):
+        return psutil.cpu_percent() > self.CPU_LIMIT
+
+    def _is_memory_full(self):
+        return psutil.virtual_memory().available < self.MEMORY_LIMIT
```
lukasgraf: As discussed:
This would ideally abort if `available` is lower than either one of
- an absolute low watermark (e.g. 2 GB)
- a relative percentage of total memory (e.g. 5%)

opengever/nightlyjobs/runner.py
```diff
@@ -0,0 +1,140 @@
...
+            raise TimeWindowExceeded(self.window_start, self.window_end)
+        if self.check_system_overloaded():
+            raise SystemLoadCritical(self.CPU_LIMIT, self.MEMORY_LIMIT)
+
+    def check_in_time_window(self):
+        current_time = datetime.now().time()
+        current_timedelta = self._positive_timedelta(self.window_start, current_time)
+        return current_timedelta < self.window_length
+
+    def _is_cpu_overladed(self):
```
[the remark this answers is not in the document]
njohner: `psutil.getloadavg()` is not available in the current version (it's tagged as available in 5.6.2, but the latest version is 5.6.1, so it will probably available soon). It would probably be available if we pulled master, but I guess we don't really want to do that. That's why I'd gone for CPU percent, which is, as you explained, not a great measure...

opengever/nightlyjobs/runner.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+class SystemLoadCritical(Exception):
+
+    message = "System overloaded.\n"\
+              "CPU load: {}; limit: {}"\
+              "Available memory: {}MB; limit: {}MB"
+
+    def __init__(self, cpu_limit, memory_limit):
+        self.cpu_percent = psutil.cpu_percent()
```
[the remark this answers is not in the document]
njohner: agreed

opengever/nightlyjobs/job_providers.py
```diff
@@ -0,0 +1,66 @@
...
+    'IBrowserRequest' implementing 'INightlyJobProvider'.
+    'initialize' and 'execute_job' have to be defined in the subclasses
+
+    """
+
+    def __init__(self, context, request):
+        self.context = context
+        self.request = request
+        self.initialize()
+        self.njobs_executed = 0
```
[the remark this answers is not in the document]
njohner: Yes sorry. I've more clearly split the runner from the provider now. The `NightlyJobProvider` now just needs to define a `__len__` and a `__iter__` method

njohner:
As a note `psutil.Process().cpu_percent()` is the sum of all CPUs whereas `psutil.cpu_percent()` is the average over all CPUs. See the docs and https://stackoverflow.com/questions/49514617/how-psutil-cpu-percent-function-works

Rotonen:
@njohner of all the logical CPUs, so hyperthreading and homogenous multiprocessing can also throw an extra wrench into the works there. So do laggy frequency scaling governors.

njohner:
Ok this should be ready for re-review.
I did not rewrite the whole history, instead, tried to make the refactor as clean as possible in separate commits, although chances are that the commits are not completely self-sufficient, as I refactored and then tried to split into commits that made sense.

PR #5574, opened 2019-04-17
Future proof mtest

Rotonen:
Two main prongs:

* Somewhat guard against runtime penalties from fixture growth
  * Split off anything which causes the fixtures to be regenerated off the integration testing layer
  * Split off anything which can disproportionately hit a single integration testing batch

* Automatically adjust to changes in CI infra in regards to having more or less cores

### DIFF
--- a/bin/mtest
+++ b/bin/mtest
@@ -297,15 +297,6 @@ def run_tests(test_run_params):
         )
 
     if returncode:
-        logger.error(*done_args)
-    else:
-        logger.info(*done_args)
-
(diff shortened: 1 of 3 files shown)


Rotonen:
Was going to make this a draft PR, but cannot find the UI for that currently. Did that change recently?

njohner:
Once it's a PR, I don't think you can come back to draft...

Rotonen:
This will not actually lead into useful reuse of the fixture until we can go about 20-wide, and then running the other layers takes shorter than generating the fixture, making this not useful.

PR #5576, opened 2019-04-18
Testing: make group ids consistent in OrgUnitBuilder.

jone:
The OrgUnitBuilder creates Plone Groups and OGDS Groups for each org unit. But the configured group ids in the org unit did not match the automatically generated group ids.

This change makes the group ids consistent by not setting a default argument for the sql object and setting the default when building the groups. This allows us to use the default ids, to override the ids by using a "having" but also to create the groups by hand.

### DIFF
--- a/opengever/testing/builders/sql.py
+++ b/opengever/testing/builders/sql.py
@@ -162,8 +162,6 @@ class OrgUnitBuilder(SqlObjectBuilder):
     def __init__(self, session):
         super(OrgUnitBuilder, self).__init__(session)
         self.arguments[self.id_argument_name] = u'rr'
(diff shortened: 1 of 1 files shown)


jone:
~150 tests have failed with the first implementation because they are using the org unit builder without creating groups. Since I've removed the default group IDs, the org units couldn't be written to the SQL database since the not-null constraint was violated.

I've updated the change so that we set group ids on the sql org unit object even when no groups are created, so that those tests pass without being rewritten.

Rotonen:
Shouldn't we rather make a sane fixture?

jone:
This fixes the fixture to be more sane 😉 the workaround is only necessary when testing without fixtures and using the builder directly..

PR #5582, opened 2019-04-23
Update the french documentation index to include chapters 4 through 7

andresoberhaensli:


njohner:
Chapter 4 was added in https://github.com/4teamwork/opengever.core/pull/5595

PR #5583, opened 2019-04-23
Extend @listing endpoint with `workspace_folders`-listing.

elioschmutz:
Issuer: https://github.com/4teamwork/gever-ui/issues/19


elioschmutz:
👍 updated

PR #5585, opened 2019-04-24
Add has_sametype_children metadata column.

### LINKED ISSUE #5527: Has children metadata
_Harvest: `Projekt - Umsetzung Neues User Interface GEVER 2019/2020` - `Entwicklung`_

For the new frontend we need to know if an object has children of the same type. This information should be available from the catalog in search results, particularly the `@listing` endpoint. Also a GET of an object should return this info.

Example questions we need to answer:
- Does a dossier have subdossiers?
- Is a repository folder a leaf node?
- Does a task have subtasks?

I propose to add a new boolean metadata column/Solr stored field `has_children` or `is_leafnode`.

Currently there's no need to know if an object has children of another type, e.g. if a dossier has documents. I'm not sure if there will be a need for that in the future.

`has_children` may be a little bit misleading as a dossier that contains only documents would have a value of `has_children=False`.

Suggestions welcome.

njohner:
* added a new column `has_sametype_children` to the metadata.
* add the `has_sametype_children` to solr.
* add it as a column in the `listing` endpoint

A question that surfaced is whether this is what we really wanted. For example:
* should `has_sametype_children` be `True` for a dossier containing an inactive or trashed dossier?
* should `has_sametype_children` be `True` for a dossier containing a dossier that the current user cannot see?

resolves #5527

### DIFF
--- a/opengever/base/tests/test_indexers.py
+++ b/opengever/base/tests/test_indexers.py
@@ -1,0 +1,192 @@
+from ftw.testbrowser import browsing
+from ftw.testbrowser.pages import factoriesmenu
+from opengever.testing import IntegrationTestCase
+
+
+class TestHasSameTypeChildren(IntegrationTestCase):
+
+    def move_items(self, items, source=None, target=None):
+        paths = u";;".join(["/".join(i.getPhysicalPath()) for i in items])
+        self.request['paths'] = paths
+        self.request['form.widgets.request_paths'] = paths
+        self.request['form.widgets.destination_folder'] = "/".join(
+            target.getPhysicalPath())
+
+        view = source.restrictedTraverse('move_items')
+        form = view.form(source, self.request)
+        form.updateWidgets()
+        form.widgets['destination_folder'].value = target
+        form.widgets['request_paths'].value = paths
+
+        form.handle_submit(form, object)
+
+    def test_has_sametype_children_for_businesscasedossier(self):
+        self.login(self.regular_user)
+
+        self.assert_metadata_value(True, 'has_sametype_children',
+                                   self.dossier)
+        self.assert_metadata_value(True, 'has_sametype_children',
+                                   self.subdossier)
+        self.assert_metadata_value(False, 'has_sametype_children',
+                                   self.subsubdossier)
+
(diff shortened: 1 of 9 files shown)


phgross:
@buchi what do you think about Niklaus' questions? What we need for the gever-ui?

buchi:
Grabbing the info from the catalog metadata for the GET seems reasonable to me as I expect using the indexer to be slower.

Ideally `has_sametype_children` would be `False` if the current user cannot see the children. This would however require a KeywordIndex similar to `allowedRolesAndUsers`.

In the frontend we want to display an expand button based on the `has_sametype_children` info in things like a folder tree browser. Clicking on the expand button will fetch the children from the backend. So, basically we can handle the case when we get no children. But it would be nicer if we wouldn't display the expand button at all in this case.

phgross:
I was just talking to Buchi: a limitation according to the visability for a specific user, is technical  complicated and heavy to calculate. Considering that a view restriction at subdossier/subtask level is extremely rare, we think your current implementation makes sense.

The has_sametype_children information on the GET of a specific object is not necessary, @njohner you can drop this part.

opengever/base/tests/test_indexers.py
```diff
@@ -0,0 +1,192 @@
...
+                                   self.private_root)
+        # private_folder does not have the same type as private_dossier
+        self.assert_metadata_value(False, 'has_sametype_children',
+                                   self.private_folder)
+        self.assert_metadata_value(False, 'has_sametype_children',
+                                   self.private_dossier)
+
+    @browsing
+    def test_is_updated_when_adding_subdossier(self, browser):
+        self.login(self.regular_user, browser)
```
phgross: How does that work, which handlers calls a reindex of the container?

njohner: When submitting the form we at some point call `addContentToContainer(container, object)`, which fires an `ContainerModifiedEvent` which is caught by the `plone.dexterity.content.reindexOnModify` handler. This one does as its name suggests.

phgross: all right, was my assumption.

PR #5587, opened 2019-04-24
Pass context and orgunit as GET parameters to webactions.

### LINKED ISSUE #5586: OGIP 42 - Webaction parameters
We need to add
* context url
* userid
* orgunit

as GET parameters to the webaction urls

njohner:
We add the parameters directly in the `WebActionsSafeDataGetter`, which is used by all `Renderers` and `ItemPreparers`, so that we directly have the parameters passed for all the display locations.

resolves #5586

### DIFF
--- a/opengever/webactions/renderer.py
+++ b/opengever/webactions/renderer.py
@@ -1,11 +1,13 @@
 from opengever.base.utils import escape_html
+from opengever.ogds.base.utils import get_current_org_unit
+from opengever.webactions.interfaces import IWebActionsProvider
 from opengever.webactions.interfaces import IWebActionsRenderer
+from urllib import urlencode
 from zope.component import adapter
+from zope.component import queryMultiAdapter
 from zope.interface import implementer
 from zope.interface import Interface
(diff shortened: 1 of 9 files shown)


opengever/webactions/renderer.py
```diff
@@ -31,9 +33,19 @@ def _pre_formatting(self, webactions_dict):
                     for display, webactions in webactions_dict.items())
 
     def _prepare_webaction_data(self, action):
-        return {key: value if key in self._attributes_not_to_escape else escape_html(value)
+        data = {key: value if key in self._attributes_not_to_escape else escape_html(value)
                 for key, value in action.items()}
 
+        params = "&".join("{}={}".format(name, value) for name, value in
```
phgross: you should use urllib.urlencode() here (see https://stackoverflow.com/questions/1233539/python-dictionary-to-url-parameters)

njohner:
Ok this should be ready. Using `urlencode` also encodes the get parameters and notably the forward slashes of the `context.absolute_url()`. This is why I now also use `urlencode` in the tests to encode the expected parameters...

PR #5593, opened 2019-04-25
Update content controls when updating doc properties

buchi:
If a Word document contains plain text content controls with a name (alias) that matches the name of a doc property, the content of those content controls also gets updated.


opengever/dossier/docprops.py
```diff
@@ -103,11 +108,29 @@ def write_properties(self, only_existing, properties):
...
+            doc = Document(tmpfile.path)
+
             if changed:
                 # Update cached properties
-                doc = Document(tmpfile.path)
                 CustomProperties(doc).update_all()
-                Composer(doc).save(tmpfile.path)
+
+            # Update content controls
+            stds = StructuredDocumentTags(doc)
```
njohner: `sdts` to be consistent with the module name?

buchi: Sure. Somehow typing this correctly seems hard to me. 🙃

buchi:
Related Issue: https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/587

PR #5595, opened 2019-04-25
Translate chapter 4 of documentation to French.

njohner:
I took over @andresoberhaensli 's branch and made a new PR, as it seems that the snapshots were not being stored through our git lfs.

### DIFF
--- a/docs/public-fr/chapitre4-notifications.rst
+++ b/docs/public-fr/chapitre4-notifications.rst
@@ -1,0 +1,46 @@
+Paramètres de notification
+==========================
+
+Dans les paramètres de notification, il est possible d’ajuster manuellement de quels évènements OneGov GEVER notifie son utilisateur respectif. Pour faire cela, cliquez sur le nom d’utilisateur en haut à droite et sélectionnez « Préférences » dans le menu déroulant.
+
+|img-notification1|
+ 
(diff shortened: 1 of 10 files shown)


docs/public-fr/chapitre4-notifications.rst
```diff
@@ -0,0 +1,46 @@
...
+Notifications pour les transmissions, rappels et tâches. 
+--------------------------------------------------------
+
+Comme pour les notifications concernant les tâches, il est possible d’ajuster les notifications concernant les transmissions et rappels dans les onglet « Transmissions » et « Rappels » respectivement.
+
+|img-notification5|
+ 
+Notifications concernant les dossiers
+-------------------------------------
+
```
njohner: I find that part a bit strange, but it's the same in German. Two steps to create an overdue dossier notification. Isn't it supposed to be the other way around, like it's not the user who wants to generate a notification, rather the user gets the notification to tell him the dossier is overdue. So this does not make much sense to me...

andresoberhaensli: - It's a bit too easy to forget how things look on lower resolutions when using those lovely 4k displays... I agree that it would be less invasive when scaled down 33%:

img-notification1

- It's a bit overkill to have the user check that the notification is enabled as a distinct step, but this way of explaining it is a simple way to make them realize the link between those 2 properties. For the time being, I would leave it as is...

njohner:
As discussed with @lukasgraf we will deal with improving screenshot uniformity at some point, but this is not a priority for now.
As for my other comment, we want to stick close to the German version of the docs, we can deal with changing that part in both documentations at a later point if necessary.

PR #5598, opened 2019-04-26
Nightly job runner: Immediately check for time window and system load

lukasgraf:
Nightly job runner: Immediately check for time window and system load upon invocation, before even attempting to look up providers and collect jobs. This will result in an unhandled exception every time the runner is invoked outside the time window (from a misconfigured cron job), and allows us to notice these misconfigurations.

(Otherwise a situation with 0 pending jobs would lead to "success", because `NightlyJobRunner.interrupt_if_necessary()` is never called).

For testing purposes I had to introduce an `early_check=True` kwarg to be able to disable this behavior when we want to test for when these conditions happen later, during job execution.

_(No change log entry, since introduced in the same release)_

### DIFF
--- a/opengever/nightlyjobs/runner.py
+++ b/opengever/nightlyjobs/runner.py
@@ -70,7 +70,12 @@ class NightlyJobRunner(object):
             for job in job_provider:
                 yield job
 
-    def execute_pending_jobs(self):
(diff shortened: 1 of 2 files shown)


lukasgraf:
Agreed, my description isn't quite accurate. I'll ship this for now and rephrase it on the next pass.

PR #5603, opened 2019-04-28
Add feature flag for running nightly jobs

lukasgraf:
This adds a **feature flag** (registry, `INightlyJobsSettings.is_feature_enabled`) that controls whether the  nightly job feature is enabled.

If the feature is disabled, and there's a cron job anyway that invokes `bin/instance run_nightly_jobs`, it will not run any jobs but **fail gracefully**. This will allow us to create cron job configurations ahead of time, but they won't have any effect until the execution of jobs is turned on via the registry flag.

This feature flag will also be used in the future to determine whether certain actions should be performed immediately during the day, or if the work should be skipped and scheduled for a nightly job in some way.

---

In addition, I had to clean up the testing a bit, to make sure that *real* nightly job providers that will eventually be registered don't get found and interfere with the test conditions for the nightly job runner tests.

I also dropped the now unnecessary `NightlyJob` class: The runner now does the bookkeeping of what provider which job belongs to, and the `job`s returned by a provider's `__iter__` now are just plain dicts with the job arguments that define a job uniquely.


opengever/nightlyjobs/testing.py
```diff
@@ -0,0 +1,24 @@
...
+from zope.component import getAdapters
+from zope.globalrequest import getRequest
+
+
+class ITestingNightlyJobProvider(INightlyJobProvider):
+    """Interface to mark a job provider used for testing only.
+    """
+
+
+class TestingNightlyJobRunner(NightlyJobRunner):
```
njohner: 👍 Good point!

opengever/nightlyjobs/interfaces.py
```diff
@@ -25,11 +30,21 @@ class INightlyJobProvider(Interface):
...
 
     def __iter__():
-        """ needs to return an iterable
+        """Returns an iterator of jobs.
+
+        The individial jobs must be dictionaries with the arguments that
+        uniquely define a job.
+
+        The specific keys in that dictionary are in the domain of the job
+        provider, and it needs be able to make sense of them when those
```
njohner: "needs to be"

lukasgraf:
> you probably meant "disabled" and not "enabled" no?

Exactly that, thanks 😉

opengever/nightlyjobs/interfaces.py
lukasgraf: Right 👍 . Will fix that one later, otherwise I'll end up in rebase hell.

PR #5606, opened 2019-04-29
Improve the styling of the tabbedview keyword filter.

phgross:
The recently introduced (release 2019.1) tabbedview keyword filter, was not optically well embedded in the other tabbedview filters and selection options. 

The new styling is only a minimal improvement, to make it  less prominent and eye-catching.

Kapture 2019-04-29 at 10 31 22

tested on: ie11, edge and chrome


njohner:
LGTM.

PR #5607, opened 2019-04-29
Implement nightly job to perform jobs after dossier resolution.

lukasgraf:
Implement **nightly job to perform jobs after dossier resolution** as part of #5477 (OGIP 55).

The basic mode of operation is like this:

- The `AfterResolveJobs` class tracks (on a flag in the dossier's annotations) whether the `AfterResolveJobs` already have been executed for a particular dossier or not.

- This flag's value is also indexed into a `BooleanIndex`.

- When the nightly job feature flag is enabled, the `AfterResolveJobs` are  **skipped during normal operation**, and the **flag is set** to indicate that they are still pending for that dossier.

- A nightly job is registered that collects all dossiers that have
  - `review_state='dossier-state-resolved'`
  - `after_resolve_jobs_pending=True`

  and then executes the `AfterResolveJobs` for them (which will also set the flag to `False`).

In order to get the archival PDF conversion via Bumblebee to work, I had to have the nightly job runner register its own taskqueue:

Because `collective.taskqueue` requires a ZDaemon (webserver) that usually gets started during Zope2.Startup, this can't work in the context of `bin/instance [run | zopectl_cmd]` invocations, where no ZDaemons are started.

We therefore set up our own task queue (just the queue part, not a server that performs the tasks) that does nothing more than collect the jobs that get queued.

Afterwards we process all these jobs using `plone.subrequest` in the same process.

---

The cron job needs to be registered like this:
```
/path/to/01-deployment/bin/instance run_nightly_jobs
```

With release `2019.2` and upwards, it can be registered even though the nightly jobs feature flag is disabled in the registry. It will fail gracefully and just not run any jobs.

### DIFF
--- a/opengever/document/archival_file.py
+++ b/opengever/document/archival_file.py
@@ -1,3 +1,4 @@
+from ftw.bumblebee.config import bumblebee_config
 from ftw.bumblebee.config import PROCESSING_QUEUE
 from ftw.bumblebee.interfaces import IBumblebeeServiceV3
 from opengever.document.behaviors.metadata import IDocumentMetadata
@@ -71,8 +72,16 @@ class ArchivalFileConverter(object):
         IDocumentMetadata(self.document).archival_file_state = None
 
     def get_callback_url(self):
-        return "{}/archival_file_conversion_callback".format(
-            self.document.absolute_url())
+        """Get the URL for the callback to upload the archival PDF to.
+
+        This URL must be accessible by bumblebee, so we must not use the
+        virtual hosting based URL here, because that wouldn't work when
+        using SSH tunnels or nightly jobs run via bin/instance.
+        """
+        base_url = bumblebee_config.internal_plone_url
+        relative_path = '/'.join(self.document.getPhysicalPath()[2:])
+        doc_url = '/'.join([base_url, relative_path])
+        return "{}/archival_file_conversion_callback".format(doc_url)
 
     def store_file(self, data, mimetype='application/pdf'):
         if isinstance(mimetype, unicode):
--- a/opengever/dossier/nightly_after_resolve_job.py
+++ b/opengever/dossier/nightly_after_resolve_job.py
@@ -1,0 +1,43 @@
+from logging import getLogger
+from opengever.dossier.behaviors.dossier import IDossierMarker
+from opengever.dossier.resolve import AfterResolveJobs
+from opengever.nightlyjobs.interfaces import INightlyJobProvider
+from plone import api
+from Products.CMFPlone.interfaces import IPloneSiteRoot
+from zope.component import adapter
+from zope.interface import implementer
+from zope.publisher.interfaces.browser import IBrowserRequest
+
+
+logger = getLogger('opengever.dossier.nightly_after_resolve_job')
+
+
+@implementer(INightlyJobProvider)
+@adapter(IPloneSiteRoot, IBrowserRequest)
+class ExecuteNightlyAfterResolveJobs(object):
+
+    def __init__(self, context, request):
+        self.catalog = api.portal.get_tool('portal_catalog')
+        self.portal = api.portal.get()
+
+        # Get all dossiers that are resolved, but still have
+        # AfterResolveJobs pending
+        self.query = dict(
+            object_provides=IDossierMarker.__identifier__,
+            review_state='dossier-state-resolved',
+            after_resolve_jobs_pending=True,
+        )
+
+    def __iter__(self):
+        pending = self.catalog.unrestrictedSearchResults(self.query)
+        return iter([{'path': brain.getPath()} for brain in pending])
+
+    def __len__(self):
+        resultset = self.catalog.unrestrictedSearchResults(self.query)
+        return resultset.actual_result_count
+
+    def run_job(self, job, interrupt_if_necessary):
+        path = job['path']
+        dossier = self.portal.unrestrictedTraverse(path)
+        logger.info("Running AfterResolve jobs for %r" % dossier)
+        AfterResolveJobs(dossier).execute(nightly_run=True)
--- a/opengever/dossier/resolve.py
+++ b/opengever/dossier/resolve.py
@@ -16,6 +16,7 @@ from opengever.dossier.behaviors.filing import IFilingNumberMarker
 from opengever.dossier.interfaces import IDossierResolveProperties
 from opengever.dossier.interfaces import IDossierResolver
 from opengever.dossier.resolve_lock import ResolveLock
+from opengever.nightlyjobs.runner import nightly_jobs_feature_enabled
 from opengever.task.task import ITask
 from plone import api
 from Products.CMFCore.utils import getToolByName
@@ -363,7 +364,7 @@ class AfterResolveJobs(object):
             object_provides=ITask.__identifier__)
         return len(tasks) > 0
 
-    def execute(self):
+    def execute(self, nightly_run=False):
         """After resolving a dossier, some cleanup jobs have to be executed:
 
         - Remove all shadowed documents.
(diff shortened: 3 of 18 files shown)


opengever/dossier/tests/test_resolve.py
```diff
@@ -493,10 +563,503 @@ def test_shadowed_documents_are_deleted_recursively_when_resolving_a_dossier(sel
...
+
+        # Now run the nightly jobs
+        with self.observe_children(subdossier) as sub_children:
+            with freeze(datetime(2016, 4, 25)):
+                self.execute_nightly_jobs(expected=1)
+
+        self.assertEquals(1, len(sub_children['added']))
+        sub_journal_pdf, = sub_children['added']
+        self.assertEqual(date(2016, 3, 15), sub_journal_pdf.document_date,
+                         "End date should be set to dossier end date")
```
njohner: `"Journal PDF document date should be set to dossier end date"`

opengever/dossier/tests/test_resolve.py
```diff
@@ -493,10 +563,503 @@ def test_shadowed_documents_are_deleted_recursively_when_resolving_a_dossier(sel
...
+        self.assertEqual(0, len(children['added']))
+
+        self.assert_after_resolve_jobs_pending(
+            False, [self.empty_dossier])
+
+    @browsing
+    def test_sets_tasks_pdf_document_date_to_dossier_end_date(self, browser):
+        """When the document date is not set to the dossiers end date the
+        subdossier will be left in an inconsistent state. this will make
+        resolving the main dossier impossible.
```
njohner: I guess this description is not true anymore. We can resolve the main dossier when the subdossier is in an "inconsistent" state

opengever/dossier/tests/test_resolve.py
```diff
@@ -493,10 +563,503 @@ def test_shadowed_documents_are_deleted_recursively_when_resolving_a_dossier(sel
...
+        self.assert_after_resolve_jobs_pending(
+            True, [self.empty_dossier])
+
+        # Now run the nightly jobs
+        with self.observe_children(self.empty_dossier) as children:
+            self.execute_nightly_jobs(expected=1)
+
+        self.assertEquals(1, len(children['added']))
+        tasks_pdf, = children['added']
+        tasks_pdf.reindexObject()
```
njohner: Why do we need a `reindexObject` here? That makes me fear another instance of our indexes not being updated when they should no?
I tested and the tests pass without the `reindexObject` here, so I'd remove it (there is also one in the test from the basecalss)

opengever/document/archival_file.py
```diff
@@ -71,8 +72,16 @@ def remove_state(self):
...
-        return "{}/archival_file_conversion_callback".format(
-            self.document.absolute_url())
+        """Get the URL for the callback to upload the archival PDF to.
+
+        This URL must be accessible by bumblebee, so we must not use the
+        virtual hosting based URL here, because that wouldn't work when
+        using SSH tunnels or nightly jobs run via bin/instance.
+        """
+        base_url = bumblebee_config.internal_plone_url
+        relative_path = '/'.join(self.document.getPhysicalPath()[2:])
```
njohner: Seems a bit fragile, but I would not know of a better solution here

opengever/dossier/nightly_after_resolve_job.py
```diff
@@ -0,0 +1,43 @@
...
+        # AfterResolveJobs pending
+        self.query = dict(
+            object_provides=IDossierMarker.__identifier__,
+            review_state='dossier-state-resolved',
+            after_resolve_jobs_pending=True,
+        )
+
+    def __iter__(self):
+        pending = self.catalog.unrestrictedSearchResults(self.query)
+        return iter([{'path': brain.getPath()} for brain in pending])
```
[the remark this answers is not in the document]
lukasgraf: I'll reverse the question: Is there any benefit to using a generator? 😉 

I'm absolutely open to optimize this in the future to use lazy iterators in as many places as possible. But for now I'd like to keep things simple and predictable. Materializing the resultset just behaves in ways that are more predictable and easier to debug / reason about.

So by knowing that I didn't use a generator here, I completely eliminate one possible source of subtle runtime bugs that I don't have to worry about yet when we undoubtedly need to debug this stuff.

opengever/dossier/resolve.py
```diff
@@ -363,32 +364,46 @@ def contains_tasks(self):
...
+        a resolved dossier, or are still pending (because they have been
+        deferred to a nightly job).
+        """
         ann = IAnnotations(self.context)
         return ann.get(AFTER_RESOLVE_JOBS_PENDING_KEY, False)
 
-    def set_after_resolve_jobs_pending(self, value):
-        ann = IAnnotations(self.context)
+    @after_resolve_jobs_pending.setter
+    def after_resolve_jobs_pending(self, value):
```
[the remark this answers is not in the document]
lukasgraf: Yeah, I made it a setter later, and mistakenly squashed that commit onto the wrong one - the non-setter variant was never meant to see the light of day, but oh well 😉

opengever/dossier/resolve.py
```diff
@@ -363,32 +364,46 @@ def contains_tasks(self):
...
+            return
 
         self.trash_shadowed_docs()
         self.purge_trash()
         self.create_journal_pdf()
         if not self.context.is_subdossier() and self.contains_tasks():
             self.create_tasks_listing_pdf()
         self.trigger_pdf_conversion()
-        self.set_after_resolve_jobs_pending(False)
+        self.after_resolve_jobs_pending = False
```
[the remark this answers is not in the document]
lukasgraf: True. But on the other hand, with all those tasks the containing dossier is definitely already getting indexed, so it really doesn't make a difference in terms of performance. Also, by explicitely setting `False` we can actually distinguish dossiers that have been resolved using this new mechanism (`False`) vs. ones from before (no annotations).

Also, not persisting the `False` value means we indefinitely tie ourselves to the fallback in the getter (which is unlikely to ever change, but I'd still prefer things to be rather explicit).

opengever/dossier/tests/test_resolve.py
lukasgraf: This stems from the original test. I also disliked a lot of these existing tests, but I left them the way they were for the sake of diffability. Adding new ones or otherwise working of these sets of coupled tests becomes very difficult when they start to diverge.

And cleaning them up is way out of scope for this PR.

opengever/dossier/tests/test_resolve.py
lukasgraf: Actually, I think we have a regression regarding this 😉 I noticed some cases where resolving a dossier (with subdossiers), reactivating and attempting to resolve again resulted in such an issue again. Haven't had time to fully investigate it though.

But yeah, the message doesn't really reflect the current state very well any more, I'll see if I can reword it or just drop it alltogether.

opengever/dossier/tests/test_resolve.py
lukasgraf: Good point, thanks, I'll look into this.

opengever/document/archival_file.py
lukasgraf: Yeah, I don't like it too much either, it's a bit ugly. But fragile it shouldn't be.

opengever/dossier/tests/test_resolve.py
lukasgraf: I checked, there is indeed a regression here. I created an issue: #5611
Will leave these messages intact for now, because they still describe *why* we need these tests.

lukasgraf:
@njohner thanks for the review! I addressed all your feedback (updated or commented).

PR #5609, opened 2019-04-30
French translation of documentation

njohner:
French translation of most of the user-manual done by @andresoberhaensli . For now I've just made the necessary corrections so that the documentation gets built correctly. I still need to look at the content.


njohner:
Snapshots for dossier creation were done with too much privileges and show the 5th tab "Protéger".

General notes on documentation:
* Why are the dossier actions described in the chapter "modify dossier"?
* In the configuration options (in the part on archiving), option for journal pdf is described but not for task pdf.

andresoberhaensli:
Updated screencaps: [img-creernouveaudossier-nodossierprotégé.zip](https://github.com/4teamwork/opengever.core/files/3133911/img-creernouveaudossier-nodossierprotege.zip)


> * Why are the dossier actions described in the chapter "modify dossier"?
Makes sense to me, as, for example, closing a dossier is a modification of its status as well. 
> * In the configuration options (in the part on archiving), option for journal pdf is described but not for task pdf.
As far as i understand it, it is automatic and enabled server side. User has no control over it.

docs/public-fr/chapitre13-deroulements-standards.rst
```diff
@@ -44,7 +44,7 @@ Sélectionnez ensuite le déroulement standard à utiliser, puis cliquez sur *Co
 
 |img-deroulstandard-4|
 
-Les modèles de tâches marquées comme présélectionnés sont déjà inclus. Les tâches requises peuvent être ajoutées/ôtées par l’intermédiaire d’un clic dans la case à cocher.
+Les modèles de tâches marquées comme présélectionnées sont déjà inclus. Les tâches requises peuvent être ajoutées/ôtées par l’intermédiaire d’un clic dans la case à cocher.
```
andresoberhaensli: Accord sur _modèles_, n.m.

njohner: but in that case `marqués` and not `marquées`. They have to go with the same word

andresoberhaensli: I agree.

PR #5619, opened 2019-05-02
Correctly handle dossier deactivation through RESTAPI

njohner:
We add the RESTAPI endpoint to properly handle the deactivate dossier transition. As for dossier resolution, the transition is handled by custom transition handler. We do not need to use a `TransitionExtender` here, as there are no jobs that need to be executed after transition to the inactive state.

I refactored the tests of the dossier deactivate transition to allow for easier code reuse and have the same tests for the transition via browser or via RESTAPI.

I also added a test that trying to resolve a dossier non-recursively through the RESTAPI is not allowed. It seemed that this was not tested, or at least I did not find a test for it. 

### DIFF
--- a/opengever/dossier/tests/test_deactivate.py
+++ b/opengever/dossier/tests/test_deactivate.py
@@ -13,6 +13,7 @@ from opengever.dossier.behaviors.dossier import IDossier
 from opengever.dossier.behaviors.protect_dossier import IProtectDossier
 from opengever.testing import IntegrationTestCase
 from plone import api
+import json
 
 
 class TestDossierDeactivation(IntegrationTestCase):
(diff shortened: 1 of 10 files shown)


opengever/dossier/tests/test_deactivate.py
```diff
@@ -139,3 +140,44 @@ def test_subdossiers_the_user_cannot_view_can_also_block_deactivation(self, brow
         expected_msgs = [u"The Dossier 2016 contains a subdossier "
                          u"which can't be deactivated by the user."]
         self.assert_errors(self.subdossier, browser, expected_msgs)
+
+
+class TestDossierDeactivationRESTAPI(TestDossierDeactivation):
+
+    api_headers = {
```
phgross: api_headers is now part of the IntegrationTestCase

PR #5620, opened 2019-05-02
 Nightly jobs: Log all output to a dedicated self-rotating logfile.

lukasgraf:
Nightly jobs: Log all output to a dedicated self-rotating logfile in `var/log/nightly-jobs.log`.

In order to accomplish that, we set up the logger appropriately in the cronjob wrapper (in other context, a logfile handler would already be present). That properly configured logger then gets passed to the providers by having them adapt a third discriminator, `logging.Logger` (i.e., an instance of the Python stdlib `Logger` class).

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2019.2.0 (unreleased)
 ---------------------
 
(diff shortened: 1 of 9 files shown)


lukasgraf:
_(Also quickly looked over this with @njohner to keep him in the loop)_

PR #5622, opened 2019-05-03
 Trash: Update and reindex modification date when trashing documents.

lukasgraf:
In order to be able to query for recently trashed documents from RIS, we need to make sure that
- **Our automatic `trashed` filter can be overridden**.
  So our patch must only inject the filter for non-trashed docs if a query against the `trashed` index isn't already present (by looking at both `REQUEST` and `kw` in the patched `searchResults()`)
- the **modification date is updated and indexed** correctly

With these in place, a query via the REST API like this works:
```http
GET .../dossier-43/@search?portal_type=opengever.document.document&trashed:boolean=True
```

(Because `trashed` in GEVER is a `FieldIndex` and not a `BooleanIndex`, `plone.restapi` can't automatically cast the query arguments. Therefore a [ZPublisher type hint](https://zope.readthedocs.io/en/latest/zdgbook/ObjectPublishing.html#argument-conversion) like `trashed:boolean` needs to be used).

/cc @jone

### DIFF
--- a/opengever/trash/trash.py
+++ b/opengever/trash/trash.py
@@ -101,7 +101,8 @@ class Trasher(object):
         notify(UntrashedEvent(self.context))
 
     def reindex(self):
-        self.context.reindexObject(idxs=['trashed', 'object_provides'])
+        self.context.setModificationDate()
+        self.context.reindexObject(idxs=['trashed', 'object_provides', 'modified'])
 
     def verify_is_not_already_trashed(self):
         if ITrashed.providedBy(self.context):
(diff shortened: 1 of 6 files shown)


opengever/trash/trash.py
```diff
@@ -101,7 +101,8 @@ def untrash(self):
         notify(UntrashedEvent(self.context))
 
     def reindex(self):
-        self.context.reindexObject(idxs=['trashed', 'object_provides'])
+        self.context.setModificationDate()
+        self.context.reindexObject(idxs=['trashed', 'object_provides', 'modified'])
```
phgross: the opposite of trashing - reactivate/untrash - uses also the `reindex` method and modified is therefore reindexed?

lukasgraf: Yes - the `untrash` method also calls `self.reindex()`, so the modification date also gets updated in that case.

PR #5624, opened 2019-05-03
When adding document_type index we do not need to update metadata.

### LINKED ISSUE #5612: document_type upgradestep by mistake calculates all metadata
The current upgradestep for adding and indexing the document_type recalculates all metadata, this is unnecessary https://github.com/4teamwork/opengever.core/blob/master/opengever/core/upgrades/20190409141826_add_document_type_index/upgrade.py

njohner:
`document_type` is only an index and not in the metadata, but is also indexed in solr. We therefore do not need to update the metadata, but need to explicitly index it in solr.

resolves #5612

### DIFF
--- a/opengever/core/upgrades/20190409141826_add_document_type_index/upgrade.py
+++ b/opengever/core/upgrades/20190409141826_add_document_type_index/upgrade.py
@@ -1,5 +1,9 @@
+from collective.indexing.interfaces import IIndexQueueProcessor
+from collective.indexing.queue import getQueue
 from ftw.upgrade import UpgradeStep
 from opengever.document.behaviors import IBaseDocument
+from plone import api
+from zope.component import getUtility
 
 
 class AddDocumentTypeIndex(UpgradeStep):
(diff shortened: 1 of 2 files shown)


opengever/core/upgrades/20190409141826_add_document_type_index/upgrade.py
```diff
@@ -9,11 +13,18 @@ class AddDocumentTypeIndex(UpgradeStep):
...
         if not self.catalog_has_index('document_type'):
             self.catalog_add_index('document_type', 'FieldIndex')
 
         query = {
             'object_provides': IBaseDocument.__identifier__,
             }
 
-        self.catalog_reindex_objects(query, idxs=['document_type'])
+        for obj in self.objects(query, "Index document type."):
+            catalog.reindexObject(obj, idxs=['document_type'],
```
Rotonen: We're starting to use this so often now, should we generalize the catalog insert from `opengever.base.security` and reuse that here too?

https://github.com/4teamwork/opengever.core/blob/bf41ae7266eecfbf2bd55f3ac41afe29380d3711/opengever/base/security.py#L28-L53

That'd also help uniformly feature flag the solr stuff.

PR #5628, opened 2019-05-05
 Add test for archival PDF generation using nightly jobs.

lukasgraf:
Add test for archival PDF generation using nightly jobs.

These were missing from the initial implementation in #5607. I refactored the existing tests in such a way that there isn't _too_ much repetition between the two sets of tests.


opengever/dossier/tests/test_resolve.py
```diff
@@ -1128,6 +1138,62 @@ class TestAutomaticPDFAConversionRESTAPI(ResolveTestHelperRESTAPI, TestAutomatic
...
+        self.login(self.secretariat_user, browser)
+
+        api.portal.set_registry_record(
+            'archival_file_conversion_enabled', True,
+            interface=IDossierResolveProperties)
+
+        doc = self.create_additional_doc()
+
+        # Resolving doesn't trigger the AfterResolveJobs yet...
+        self.resolve(self.resolvable_dossier, browser)
```
njohner: Maybe assert that the queue is empty here?

lukasgraf: Good point, updated 👍

PR #5629, opened 2019-05-06
Add file_extension indexer for mails.

### LINKED ISSUE #5623: Filextension for Mails

Currently the file_extension index only indexes the extension for document but not emails. We would like to have the extension of the original_file (`.msg` or `.eml`) indexed and visible in the listings.

phgross:
The indexer uses the file_extension of the original_message, if exists.

Provides a deferrable upgradestep which reindexes the file_extension for all mails.

Closes #5623.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2019.2.0 (unreleased)
 ---------------------
 
+- Add file_extension indexer for mails. [phgross]
 - Fix upgrade step adding document_type index to not update the metadata. [njohner]
 - Log nightly job output to a dedicated, self-rotating logfile. [lgraf]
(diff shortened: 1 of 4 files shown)


opengever/mail/indexer.py
```diff
@@ -9,3 +10,12 @@ def checked_out(obj):
     empty string."""
 
     return ''
+
+
+@indexer(IMail)
```
lukasgraf: Hmm. Any particular reason we don't just register the [existing indexer](https://github.com/4teamwork/opengever.core/blob/master/opengever/document/indexers.py#L222-L228) for `IBaseDocument` instead of `IDocumentSchema`?

phgross: Good question 😅. I dropped the mail specific indexer and changed the existing document indexer instead. please take another look.

PR #5631, opened 2019-05-06
Correctly handle dossier activation through RESTAPI

njohner:
In the same spirits as the other PRs for #5432:
* We add the RESTAPI endpoint to properly handle the activate dossier transition.
* I refactored the tests of the dossier activate transition to allow for easier code reuse and have the same tests for the transition via browser or via RESTAPI.

### DIFF
--- a/opengever/dossier/tests/test_activate.py
+++ b/opengever/dossier/tests/test_activate.py
@@ -69,24 +69,16 @@ class TestDossierActivation(IntegrationTestCase):
 
     @browsing
     def test_end_date_is_reindexed(self, browser):
-        enddate = date(2013, 2, 21)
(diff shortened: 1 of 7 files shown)


opengever/dossier/tests/test_activate.py
```diff
@@ -104,3 +106,56 @@ def test_end_date_is_reindexed(self, browser):
 
         self.assert_index_value('', 'end', self.inactive_dossier)
         self.assert_metadata_value(None, 'end', self.inactive_dossier)
+
+
+class TestDossierActivationRESTAPI(TestDossierActivation):
+
+    api_headers = {
```
phgross: you can use now self.api_headers from integration_test_case (maybe a rebase is necessary).

PR #5633, opened 2019-05-07
Correctly handle dossier reactivation through RESTAPI

njohner:
We add the RESTAPI endpoint to properly handle the reactivate dossier transition. As for dossier resolution, the transition is handled by custom transition handler. We do not need to use a `TransitionExtender` here, as there are no jobs that need to be executed after the transition to the active state.

I refactored the tests of the dossier reactivate transition to allow for easier code reuse and have the same tests for the transition via browser or via RESTAPI. I also ported them to the `IntegrationTestCase`

Note: reactivating a dossier that is inactive does not throw an error, but that was already that way before... I made an issue for this https://github.com/4teamwork/opengever.core/issues/5638

### DIFF
--- a/opengever/dossier/reactivate.py
+++ b/opengever/dossier/reactivate.py
@@ -1,9 +1,11 @@
 from opengever.dossier import _
 from opengever.dossier.base import DOSSIER_STATES_OPEN
 from opengever.dossier.behaviors.dossier import IDossier
+from opengever.dossier.resolve import PreconditionsViolated
 from plone import api
 from Products.CMFCore.utils import getToolByName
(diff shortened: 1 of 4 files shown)


opengever/dossier/reactivate.py
```diff
@@ -23,8 +23,6 @@ def get_precondition_violations(self):
         return errors
 
     def reactivate(self):
-        if self.get_precondition_violations():
```
njohner: We are already checking the preconditions before calling the `reactivate` method

phgross:
@njohner changelog entry is missing otherwise, ready to 🚢

Issue #5632, opened 2019-05-07
Editing a dossier creates duplicate of dossiers.


phgross:
Steps to reproduce:
 - Create a main dossier
 - Edit the metadate from the dossier for example the description 
 - The dossier is now listed twice, with to different reference numbers

Example on dev:  https://dev.onegovgever.ch/fd/ordnungssystem/ressourcen-und-support/ict/informatik/software-fachapplikationen/dossier-913#overview

Also reproducable local with current master.

phgross:
Has been solved via #5633

PR #5636, opened 2019-05-07
Fix has_children indexer which lead to duplicate brains.

### LINKED ISSUE #5635: Closing a dossier with a cancelled task is not resolvable
A dossier which contains a cancelled task is not longer resolvable.

Steps to reproduce:
 - Open a Dossier
 - Create a Task
 - Cancel the task
 - Try to resolve the dossier
 -> the message `not all tasks closed` appears

njohner:
The has_children indexer was making a catalog query, which leads to processing the indexing queue in the middle of operations like creating a folderish object. This led to duplicate brains in the catalog. At least that is my best guess as to what the problem was...

I could not reproduce the problem in tests (tried Integration and Functional tests), but I can confirm that I could reproduce the two issues locally and that they do not happen anymore with these changes.

resolves #5635 #5632

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2019.2.0 (unreleased)
 ---------------------
 
+- Fix has_children indexer which lead to duplicate brains. [njohner, phgross]
 - Trash: Update and reindex modification date when trashing documents. [lgraf]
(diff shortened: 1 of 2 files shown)


njohner:
Huh, might have a flaky test. Or a test that passes locally but not on CI...

PR #5641, opened 2019-05-07
Oneoffixx configurability changes

Rotonen:
Extranet issue:
https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/714

As per potential upcoming changes in their API, we've been alerted to make more things configurable at runtime.

### DIFF
--- a/opengever/oneoffixx/interfaces.py
+++ b/opengever/oneoffixx/interfaces.py
@@ -37,3 +37,11 @@ class IOneoffixxSettings(Interface):
         description=u'Quite many of the backend responses are very slow, so we cache them per user. Defaults to 30 days.',
         default=30 * 24 * 60 * 60,
     )
+
+    scope = schema.TextLine(
+        title=u'A scope parameter for API versioning.',
+        description=u'This is their OAuth2 grant scope they use for '
+                    u'versioning the API.',
+        default=u'oo_V1WebApi',
+        required=False,
+    )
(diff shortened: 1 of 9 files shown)


opengever/oneoffixx/interfaces.py
```diff
@@ -37,3 +37,11 @@ class IOneoffixxSettings(Interface):
         description=u'Quite many of the backend responses are very slow, so we cache them per user. Defaults to 30 days.',
         default=30 * 24 * 60 * 60,
     )
+
+    scope = schema.TextLine(
+        title=u'A scope parametre for API versioning.',
```
phgross: typo

opengever/oneoffixx/tests/test_oneoffixx.py
```diff
@@ -617,3 +618,105 @@ def test_oneoffixx_favorites_lists_only_favorites(self, browser):
             expected_options,
             [e.get('title') for e in browser.css('input[type=radio]')],
         )
+
+
+class TestOneoffixxClientGrantScopeDefault(IntegrationTestCase):
```
phgross: I do not understand why your splitting this test in to two separate test classes? The setup is nearly identical. 
I think it would be easier to read/understand when those two tests would be in the same testcase ...

Rotonen: The caching happening within the API client singleton prevents me from doing this - the second one would never end up making the request for me to assert on.

opengever/oneoffixx/interfaces.py
Rotonen: Amended.

PR #5643, opened 2019-05-08
Testing handover documentation

Rotonen:
* Unified and validated all of the legacy scripts I've had around for profiling the tests
  * Based them all on the current `bin/mtest`
* Got rid of all the external dependencies for the scripts
* Documented the scripts
* Documented the nightlies
* Added some tips and tricks for local testing


docs/intern/development/testing.rst
```diff
@@ -181,3 +181,58 @@ Here is an example run of it being run for one module: ::
...
+
+The result files will be prefixed with the current date and the git commit hash
+of ``HEAD`` and can be found in ``parts/test/*prof``.
+
+A good starting point for digging into the results is setting the root of your
+view onto the test function, immediately under which are all the tests which
+have gotten run and then zooming in test by test and resetting the zoom to get
+back to the root.
+
+Viewing profiling results
```
lukasgraf: Could also drop a link to [KCacheGrind](https://kcachegrind.github.io/html/Home.html) here. It's Linux only, but easy enough to fire up a VM for that. And alternatives are good here, because as you pointed out, both of these two have their own pros and cons.

Rotonen: Amended.

PR #5652, opened 2019-05-13
Add CMFEditions modifier that prevents journals from being versioned.

### LINKED ISSUE #5400: Versionierung des Journals verhindern

Genaue Spezifikation folgt noch, dient als Platzhalter für Releaseplanung.

Bei der aktuellen Versionierung eines Dokuments wird auch das Journal eines Dokuments mitversioniert, dies ist sowohl aus technischer wie auch fachlicher Sicht unnötig. Die Versionierung durch cmf.editions soll so angepasst werden, dass das "versionierte Objekt" über keine Journaleinträge mehr verfügt.

lukasgraf:
This adds a modifier that will **prevent the `ftw.journal` entries from getting versioned** and piling up in the storage over time.

The modifier
- removes the ftw.journal `PersistentList` before cloning and pickling the object for versioning
- and will remove the `'ftw.journal.journal_entries_annotations_key'` entirely before saving the version

When retrieving a version, the journal is never being restored, so it's only ever available on the working copy.

Please see the docstring of the `get_journal_callbacks()` function for an in-depth explanation of how this mechanism works in detail (because the CMFEditions interface specs are severely lacking - [their words, not mine](https://github.com/plone/Products.CMFEditions/blob/4538606eb5739e5f36ca52fd55605658fd35d7e4/Products/CMFEditions/interfaces/IModifier.py#L71)). That docstring also explains why using the [`getCallbacks` function from `plone.app.versioningbehavior`](https://github.com/plone/plone.app.versioningbehavior/blob/c57163c2007e48301fc4179da6f2a142fd4fa904/plone/app/versioningbehavior/modifiers.py#L35-L55) would be a mistake.

Closes #5400 _(This will prevent versioning future journal entries. The existing ones will still have to be removed from the DB using a one-off maintenance script)._


opengever/document/modifiers.py
```diff
@@ -0,0 +1,164 @@
...
+    pickler. It will also namespace the persistent IDs generated by the
+    modifiers by prefixing them with the modifier name, and use that namespace
+    to look up the appropriate modifier's persistent_load during unpickling.
+
+    (See P.CMFEditions.ModifierRegistryTool.getOnCloneModifiers).
+
+    WARNING: The getCallbacks() function from plone.app.versioningbehavior
+    might seem like a good fit here, and will give the impression to work
+    as intended, but only on the surface. In pracice, it would lead to the
+    journals still being pickled to the DB in their entirety, because it uses
```
buchi: I'm not sure about that. `getOnCloneModifiers()` is only used to clone an object (cloning is done by pickling and unpicking an object). IMO the persistent ids should never land in the db. But maybe I'm getting something wrong.

lukasgraf: As discussed: I did indeed misunderstand how those pieces work together exactly - will update the comment accordingly.

opengever/document/modifiers.py
```diff
@@ -0,0 +1,164 @@
...
+
+        return {}, [], []
+
+    def afterRetrieveModifier(self, obj, repo_clone, preserve=()):
+        """Don't do anything, since we explicitly don't want to restore
+
+        journal data after retrieval of a version. The journal is only
+        supposed to be available on the working copy of the object.
+        """
+        return [], [], {}
```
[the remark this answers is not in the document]
lukasgraf: I extended the commend accordingly and added a test to verify that the journal doesn't get dropped after a revert (using our own mechanism).

PR #5653, opened 2019-05-13
Fix resolving of multi-adminunit tasks.

### LINKED ISSUE #5651: Erledigung von mandantenübergreifenden Aufgaben führt zu Fehler

Bei der Erledigung von mandantenübergreifenden Aufgaben tritt ein Fehler auf (Sentry -> https://sentry.4teamwork.ch/sentry/onegov-gever/issues/25036/) oder die Dokumente werden nicht richtig zur Ursprungsaufgaben kopiert.

phgross:
Avoid unnecessary workflow changes syncs, when resolve a multi adminunit task. Because this leads do erorrs or incomplete task completions (not delivered documents etc.)

Unfortunately adding a test which checks for this issues was not possible, because we can not completly test the multi-adminunit communication.

Closes #5651

### DIFF
--- a/opengever/task/util.py
+++ b/opengever/task/util.py
@@ -148,12 +148,13 @@ def add_simple_response(task, text='', field_changes=None, added_object=None,
     return response
 
 
-def change_task_workflow_state(task, transition, **kwargs):
+def change_task_workflow_state(task, transition, disable_sync=False, **kwargs):
     """Changes the workflow state of the task.
     """
 
     wftool = api.portal.get_tool('portal_workflow')
-    wftool.doActionFor(task, transition, transition_params=kwargs)
+    wftool.doActionFor(
+        task, transition, disable_sync=disable_sync, transition_params=kwargs)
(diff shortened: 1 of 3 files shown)


opengever/task/util.py
```diff
@@ -148,12 +148,13 @@ def add_simple_response(task, text='', field_changes=None, added_object=None,
...
 
 
-def change_task_workflow_state(task, transition, **kwargs):
+def change_task_workflow_state(task, transition, disable_sync=False, **kwargs):
     """Changes the workflow state of the task.
     """
 
     wftool = api.portal.get_tool('portal_workflow')
-    wftool.doActionFor(task, transition, transition_params=kwargs)
+    return wftool.doActionFor(
```
njohner: Why do we add a return here? Do we need it? seems like `doActionFor` does not return anything anyway.

phgross: You're right, not necessary.

PR #5654, opened 2019-05-13
Fix the setting of the content-type for Oneoffixx templates

Rotonen:
Seems I actually built all the machinery we needed, but then forgot to to actually plug it in in the end.

Original PR: https://github.com/4teamwork/opengever.core/pull/4282

🐷 will get fed.

### DIFF
--- a/opengever/oneoffixx/tests/test_oneoffixx.py
+++ b/opengever/oneoffixx/tests/test_oneoffixx.py
@@ -4,6 +4,7 @@ from opengever.oneoffixx.api_client import OneoffixxAPIClient
 from opengever.oneoffixx.exceptions import OneoffixxBackendException
 from opengever.oneoffixx.exceptions import OneoffixxConfigurationException
 from opengever.oneoffixx.interfaces import IOneoffixxSettings
+from opengever.oneoffixx.utils import whitelisted_template_types
 from opengever.testing import IntegrationTestCase
 from plone import api
 from zope.annotation.interfaces import IAnnotations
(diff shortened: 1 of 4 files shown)


Rotonen:
And we ended up testing this functionality against the shadow document in the fixture, which staticly sets it in the document annotations.

Now testing the annotations of an actually created shadow document.

opengever/oneoffixx/tests/test_oneoffixx.py
```diff
@@ -128,9 +129,29 @@ def test_document_creation_from_oneoffixx_template_creates_shadow_doc(self, brow
...
+            'application/vnd.openxmlformats-officedocument.wordprocessingml'
+            '.document',
+            annotations['content-type'],
+        )
+
+        whitelisted_content_types = [
+            template['content-type']
+            for template in whitelisted_template_types.values()
+        ]
+        self.assertIn(annotations['content-type'], whitelisted_content_types)
```
njohner: This seems a bit disconnected from the rest of the test, and the added value seems slim.

njohner:
Oh forgot, but can you reference the original PR in the description?

opengever/oneoffixx/tests/test_oneoffixx.py
Rotonen: The annotations having been set properly are an integral part of having created a shadow document from an Oneoffixx template. Where else should the assertions go?

Rotonen:
@njohner amended the PR body.

opengever/oneoffixx/tests/test_oneoffixx.py
njohner: Sorry maybe that wasn't clear. Testing the annotations above is fine with me. I was just wondering about that last assertion here, where you test that the `annotations['content-type']` is in the `whitelisted_content_types`. You have already tested what `annotations['content-type']` was, just above, so here you are affectively testing that `'application/vnd.openxmlformats-officedocument.wordprocessingml.document'` is in `whitelisted_content_types`. I was wondering whether that made much sense or not.

Rotonen: As the content type is pulled from that whitelist based on the meta template ID, I felt it prudent to add that assertion as well.

njohner: I approved the PR because I did not have a strong feeling about this, so feel free to leave it in.

PR #5656, opened 2019-05-14
 Fix bug with resolving a reopened dossier.

### LINKED ISSUE #5611: Regression: Resolving, reopening and resolving dossier again doesn't work in all cases
_Harvest: `[3.2.2.1] OneGov GEVER` - `Softwarepflege (Refactoring/Bug)`_

There seems to be a regression where resolving a dossier, reopening and resolving it again doesn't work in all cases.

**Steps to reproduce**

- Create a dossier
- Create a subdossier inside that dossier
- Add a document to the subdossier
- Resolve the main dossier
- Reopen the main dossier
- Attempt to resolve the main dossier again

**Expected result**
- Main dossier should be resolved recursively

**Actual Result**
- `Fehler: Es sind nicht alle Dokumente und Aufgaben in Subdossiers abgelegt.`

Screen Shot 2019-04-30 at 13 23 48

njohner:
When the `tasks-pdf` or `journal-pdf`feature is enabled and a dossier with subdossiers gets resolved, it will end up with documents in the main dossiers. These documents will prevent resolving the dossier again after having reopened it, as the `StrictDossierResolver` expects all documents to be in subdossiers.

resolves #5611


docs/HISTORY.txt
```diff
@@ -4,6 +4,7 @@ Changelog
 2019.2.0 (unreleased)
 ---------------------
 
+- Fix bug with resolving a reopened dossier. [njohner]
```
phgross: Take care when rebasing because of the version bump ...

PR #5660, opened 2019-05-14
Also update empty doc properties

buchi:
If a document has a doc property with a specific value and that value changes in GEVER to None it does not get updated in the Word document. This PR fixes that an sets the value to an empty string.


opengever/dossier/tests/__init__.py
```diff
@@ -82,7 +82,11 @@
     'ogg.document.reference_number': 'Client1 1.1 / 1 / 35',
     'ogg.document.document_date': datetime(2016, 8, 31, 0, 0),
     'ogg.document.sequence_number': '35',
-    'ogg.document.version_number': 0
+    'ogg.document.version_number': 0,
+    'ogg.document.delivery_date': '',
```
deiferni: just to be sure we don't break anything … how is an empty string handled? is it valid? i can't create such a doc-property via the word ui so it would be good to know that word can also update these properties.

opengever/contact/docprops.py
```diff
@@ -13,8 +13,8 @@ def __init__(self, prefix):
 
     def _add_property(self, properties, kind, name, value):
         """Add single property to collection of properties."""
-        if not value:
-            return
+        if value is None:
```
deiferni: We handle value conversion already in https://github.com/4teamwork/ooxml_docprops i would prefer if conversion of `None` values would happen in the same place.
Also after looking at this for a bit i think there may be an issue when attempting to overwrite existing `bool`, `int` or `datetime` values with a now empty value. This conversion would simply convert to an empty string but i am fairly certain that will trip up when we attempt to retrieve the written values in https://github.com/4teamwork/ooxml_docprops/blob/master/ooxml_docprops/datatypes.py#L45-L60 or even before when writing. I don't think that the type of the doc-property will be converted when writing.

I would therefore choose a slightly different approach

Drop the 
```python
if not value:
    return
```
part as you suggested in this pull-request but don't do value conversion, put `None` into the doc-property key-value mapping.

Then when updating doc-properites:
- If the value is `None` remove the doc-property from the document's properties part. If i saw correctly word does not support `None` values (also i can't add empty stringed values over the UI). So i think the equivalent of `None` would be to not have the doc-property present altogether in the document.
- Also think about updating cached values, we should behave the same as word when updating cached values of doc-properties that are "no longer available". The spec may help on how we should behave in this case.

deiferni:
What is our way forward here? IIRC we've discussed modifying `ooxml_docprops` or even incorporating it into `docxcompose` to fix the issues mentioned above.

buchi:
I've changed the implementation. Doc properties are now updated with docxcompose instead of ooxml_docprops.

Because doc properties always must have a value, they are now deleted if we don't have a value in GEVER. The cached value of the field referencing the doc property is updated with an empty string. The field is kept in the document intentionally to allow readding a doc property.

setup.py
```diff
@@ -130,7 +130,6 @@
           'jsonschema',
           'Markdown',
           'mr.bob',
-          'ooxml_docprops',
```
deiferni: you should also get rid of the two `ooxml-docprops` and `ooxml_docprops` pinnings in `versions.cfg`.

sources.cfg
```diff
@@ -8,6 +8,7 @@ extensions += mr.developer
 development-packages =
   opengever.maintenance
   collective.js.timeago
+  docxcompose
```
deiferni: Just a reminder so we don't accidentally forget: this needs to go as soon as https://github.com/4teamwork/docxcompose/pull/40 is merged and released 😉.

sources.cfg
```diff
@@ -18,3 +19,4 @@ auto-checkout = ${buildout:development-packages}
 [branches]
 Products.LDAPUserFolder = ftw
 Products.LDAPMultiPlugins = ftw
+docxcompose = edit-custom-props
```
deiferni: same here, of course.

setup.py
buchi: removed

sources.cfg
buchi: removed and updated pinning of docxcompose

sources.cfg
buchi: removed and updated pinning of docxcompose

PR #5661, opened 2019-05-14
Docs: Update nightly cron jobs example

lukasgraf:
The nightly cron jobs need to run with the zope user's environment variables. Because these are not normally sourced in a crontab, we prefix the nightly job command with `bash -l -c` in order to run them in a login shell, which will load the environment variables.

This is of particular importantance for the nightly jobs for at least two reasons:

- When triggering archival PDF generation via bumblebee, the `http_proxy` and `no_proxy` environment variables must be set correctly, otherwise requests may fail (with 502 notresolvable / BadStatusLine errors).

- We're attempting to log errors from the nightly jobs to  Sentry. For that to work, the `RAVEN_DSN` env variable needs to be set properly.

### DIFF
--- a/docs/intern/operations/allgemein.rst
+++ b/docs/intern/operations/allgemein.rst
@@ -110,7 +110,14 @@ Die Uhrzeit für den Cron-Job muss so gewählt werden, dass sie dem in den
 .. code:: bash
 
     # Nightly Jobs
-    0 01 * * * /home/zope/server/01-gever.example.org/bin/instance run_nightly_jobs >/dev/null 2>&1
+    0 01 * * * /bin/bash -l -c "/home/zope/server/01-gever.example.org/bin/instance0 run_nightly_jobs >/dev/null 2>&1"
+
+.. warning::
+    Der ``/bin/bash -l -c "<command>"`` wrapper ist nötig damit
+    der Cron-Job mit einer Login-Shell ausgeführt wird (damit die
+    Umgebungsvariablen für den ``zope`` User geladen werden). Dies ist
+    insbesondere relevant damit Umgebungsvariablen wie ``http_proxy``,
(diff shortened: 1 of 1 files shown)


docs/intern/operations/allgemein.rst
```diff
@@ -110,7 +110,14 @@ Die Uhrzeit für den Cron-Job muss so gewählt werden, dass sie dem in den
 .. code:: bash
 
     # Nightly Jobs
-    0 01 * * * /home/zope/server/01-gever.example.org/bin/instance run_nightly_jobs >/dev/null 2>&1
+    0 01 * * * /bin/bash -l -c "/home/zope/server/01-gever.example.org/bin/instance0 run_nightly_jobs >/dev/null 2>&1"
```
Rotonen: Just `bash` would work here as well.

docs/intern/operations/allgemein.rst
```diff
@@ -110,7 +110,14 @@ Die Uhrzeit für den Cron-Job muss so gewählt werden, dass sie dem in den
...
+
+.. warning::
+    Der ``/bin/bash -l -c "<command>"`` wrapper ist nötig damit
```
Rotonen: Just `bash` would work here as well.

docs/intern/operations/allgemein.rst
Rotonen: And since we now run bash, you can shorten the redirection down to `&>/dev/null`.

docs/intern/operations/allgemein.rst
Rotonen: And since we now run bash, you can shorten the redirection down to `&>/dev/null`.

docs/intern/operations/allgemein.rst
lukasgraf: @Rotonen I'd rather not change this. This is how it's already set up in production, this PR is really just to bring the documentation in line.

PR #5662, opened 2019-05-14
Update favorite-icon after cancel a document checkout.

phgross:
Since #4581 the document icon of documents marked as favorite gets updated on checkout and checkin. But on checkout-cancel the icon gets not updated, which leads in a wrong inconsistent document icon. This change solves this issue.

### DIFF
--- a/opengever/base/tests/test_favorite.py
+++ b/opengever/base/tests/test_favorite.py
@@ -6,12 +6,10 @@ from opengever.base.model import CONTENT_TITLE_LENGTH
 from opengever.base.model import create_session
 from opengever.base.model.favorite import Favorite
 from opengever.base.oguid import Oguid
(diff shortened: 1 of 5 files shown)


opengever/base/tests/test_favorite.py
```diff
@@ -185,25 +183,59 @@ def test_icon_class_of_favorites_get_updated_on_checkin_checkout_document(self):
...
+
+        fav_admin = create(Builder('favorite')
+                           .for_object(self.document)
+                           .for_user(self.administrator))
+
+        fav_user = create(Builder('favorite')
+                          .for_object(self.document)
+                          .for_user(self.regular_user))
+
         # Checkin document again
```
njohner: Should be `# Cancel document checkout`

PR #5666, opened 2019-05-15
Split up Oneoffixx configurability

Rotonen:
We discovered we cannot currently handle a path prefix in regards to where the Oneoffixx web service(s) are mounted.

I've split the configuration up a bit with sane defaults. Now the only thing which, normally needs to get configured, is the backend hostname. The second most common thing to configure is the path prefix.


Rotonen:
This is tested locally against the test instance we have access to.

PR #5673, opened 2019-05-20
Fix upgrade step migrating templates from committee_container to committee.

njohner:
Add a safeguard in case the attributes are not present on the committee_container.


njohner:
I couldn't really reproduce the error Philippe had when updating test. I tried locally, but the upgrade went through, although not all templates are set on the committee_container. I think I had discussed that with @lukasgraf at the time and concluded that the attribute would be on the CommitteeContainer, even if it's not in the schema anymore.
I don't have more time today, I could have a closer look tomorrow if you think it necessary.

PR #5677, opened 2019-05-22
Update documentation for update of SaaS deployments.

njohner:
Mainly added some info on the two step process of the update with first the non-deferrable and then the deferrable upgrade steps.


docs/intern/operations/saas.rst
```diff
@@ -97,22 +110,31 @@ Vorbereitung Release-Einspielung
         [solr]
         vm-opts = -Xms512m -Xmx2048m -Xss256
 
+- Der erwartete downtime am Betrieb ankündigen.
```
phgross: Die erwartete Downtime dem Betrieb ankündigen.

PR #5686, opened 2019-05-27
Fix revoke_permission field and validation in the fowarding forms.

### LINKED ISSUE #5685: Forwarding creation not working with disabled revoke_permissions feature
Harvest: `[3.2.2.1] OneGov GEVER` - `OGIP 53 - Berechtigungsentzug bei Aufgaben aussetzen`

Bildschirmfoto 2019-05-27 um 09 16 08

phgross:
The recently added revoke_permission fields, wasn't handled in the add and edit forwarding forms, therefore the creation of a forwarding, when the feature flag was disabled, did not worked.

Closes #5685

### DIFF
--- a/opengever/inbox/forwarding.py
+++ b/opengever/inbox/forwarding.py
@@ -8,8 +8,7 @@ from opengever.ogds.base.sources import AllUsersInboxesAndTeamsSourceBinder
 from opengever.ogds.base.utils import get_current_org_unit
 from opengever.ogds.base.utils import get_ou_selector
 from opengever.task import _ as task_mf
-from opengever.task import is_private_task_feature_enabled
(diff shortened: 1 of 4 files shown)


opengever/inbox/forwarding.py
```diff
@@ -197,6 +202,16 @@ def updateFieldsFromSchemata(self):
         super(ForwardingEditForm, self).updateFieldsFromSchemata()
         _drop_empty_additional_fieldset(self.groups)
 
+        if not is_private_task_feature_enabled():
```
deiferni: this seems to be highly duplicated. i would prefer to have it in a separate helper method or function.

phgross: 👍 Moved hidding feature flagged fields in to a helper method. (for both task and forwarding forms).

PR #5687, opened 2019-05-27
Hide byline for the teams

phgross:
There is no sensible data which should be displayed in the team 
details view. So we register an hidden byline for teams.

Bildschirmfoto 2019-05-27 um 14 17 44

See https://basecamp.com/2768704/projects/14549453/todos/388925855

### DIFF
--- a/opengever/contact/browser/byline.py
+++ b/opengever/contact/browser/byline.py
@@ -28,3 +28,12 @@ class ContactByline(BylineBase):
         if self.context.model.is_active:
             return PMF('Yes')
         return PMF('No')
+
+
+class TeamByline(BylineBase):
+    """Hidden byline for TeamWrapper objects.
+    There is no sensible byline data for team objects - so we hide them.
+    """
+
+    def show(self):
+        return False
--- a/opengever/ogds/base/tests/test_team_details.py
+++ b/opengever/ogds/base/tests/test_team_details.py
@@ -63,3 +63,16 @@ class TestTeamDetails(IntegrationTestCase):
             ['http://nohost/plone/@@user-details/kathi.barfuss',
              'http://nohost/plone/@@user-details/robert.ziegler'],
             [link.get('href') for link in links])
+
+    @browsing
+    def test_byline_is_not_shown(self, browser):
+        self.login(self.administrator, browser=browser)
+
+        byline_css_selector = '.documentByLine'
+
+        # test-precondition: byline is shown for the contactfolder
+        browser.open(self.contactfolder)
+        self.assertIsNotNone(browser.css(byline_css_selector))
+
+        browser.open(self.contactfolder, view='team-1/view')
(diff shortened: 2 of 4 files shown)


opengever/ogds/base/tests/test_team_details.py
```diff
@@ -63,3 +63,10 @@ def test_list_and_link_team_members(self, browser):
             ['http://nohost/plone/@@user-details/kathi.barfuss',
              'http://nohost/plone/@@user-details/robert.ziegler'],
             [link.get('href') for link in links])
+
+    @browsing
+    def test_byline_is_not_shown(self, browser):
+        self.login(self.administrator, browser=browser)
+        browser.open(self.contactfolder, view='team-1/view')
+
+        self.assertEquals([], browser.css('.documentByLine'))
```
deiferni: this looks a bit brittle. this test does not guard agains changing the css class of the byline. if we do change the class of the byline and due to a regression the byline starts showing again this test will still be green. i see two options to improve:
- as a test-precondition document that the byline shows in this test
- extract the byline css class into a constant and make sure it is used for the byline tests.

opengever/contact/browser/byline.py
```diff
@@ -28,3 +28,13 @@ def active_label(self):
         if self.context.model.is_active:
             return PMF('Yes')
         return PMF('No')
+
+
+class TeamByline(BylineBase):
+    """Hidden byline for the TeamWrapper objects.
+
+    There ise no sensible byline data for team objects - so we hide them.
```
deiferni: typo ise

opengever/contact/browser/byline.py
```diff
@@ -28,3 +28,13 @@ def active_label(self):
         if self.context.model.is_active:
             return PMF('Yes')
         return PMF('No')
+
+
+class TeamByline(BylineBase):
+    """Hidden byline for the TeamWrapper objects.
```
deiferni: ~~the~~

opengever/ogds/base/tests/test_team_details.py
phgross: > as a test-precondition document that the byline shows in this test
Don't get it 😅, how could I check that the byline is shown, thats exactly what should no longer be the case? Or do you mean the byline of an other objects? 

> extract the byline css class into a constant and make sure it is used for the byline tests.
The `documentByLine` is the default viewlet class from plone.app.viewlet, I see no option or sense to extract this in to a separate constant.

deiferni: clarifying the two points:

yep i meant for another object. i admit it may be a bit theoretical. but if you show
1) the byline is there for another object
2) the byline is not there for a team
in the same test you guard against the css class being changed, also you show that the changes don't accidentally destroy the byline for all objects.

extracting into a constant: i would of course only do this for our local tests, not for the code that renders the byline.

PR #5691, opened 2019-05-28
 Update creation date when copying an object.

njohner:
Creation date was not updated when copy/pasting an object. This is not the expected behaviour, as copy/paste is often used to create a new document with an other one as template.

The creation date is updated on the `IObjectCopiedEvent` so that it does not get updated when a document is moved.

I wrote generic tests for the update of document indexes and metadata during copy/paste and move operations. There seems to be some indexes and metadata that are not updated properly for example:
- `Creator` index during copy/paste
- `reference`, `SearchableText` and `is_subdossier`when moving a document.
- `start` index seems to behave a bit weirdly to me (it gets updated, probably because the default value of the field is `now`)
- `listCreators` metadata gets added a second creator during a move operation, not sure whether that is desired...
- Some indices/metadata like `is_subdosiser` are set for the document by acquisition of the field from the dossier...

I think it would be useful to write similar tests for dossiers and other content types. Would be useful that someone experienced carefully looks at the tests I wrote, where I tried to make explicit what indices and metadata get modified or not. In my opinion the other indices look ok.

resolves https://github.com/4teamwork/opengever.core/issues/5479


opengever/base/handlers.py
```diff
@@ -19,6 +19,12 @@
 from zope.lifecycleevent import IObjectRemovedEvent
 from zope.lifecycleevent import ObjectAddedEvent
 from zope.sqlalchemy.datamanager import mark_changed
+from DateTime import DateTime
+
+
+def object_copied(context, event):
+    context.creation_date = DateTime()
+    context.reindexObject(idxs=('created', 'CreationDate'))
```
deiferni: i think `CreationDate` isn't an index we support/have, only metadata.

njohner: You're right, it's only metadata.

njohner:
@deiferni just re-requested review, so you don't forget 😄

opengever/document/tests/test_copy_document.py
```diff
@@ -0,0 +1,231 @@
...
+                "copy/paste of a document, or to 'modified_metadata' otherwise")
+
+        self.assertDictEqual(expected_metadata, copy_metadata)
+
+        # Make sure the metadata was up to date
+        copy.reindexObject()
+        reindexed_copy_metadata = self.get_catalog_metadata(copy)
+
+        # Some metadata get updated because of indexing like "modified"
+        # Other data should be up to date but is not...
```
deiferni: so if i understand correctly these metadata values are not updated on paste? i think that is something we/you should have a closer look at.

opengever/document/tests/test_copy_document.py
```diff
@@ -0,0 +1,231 @@
...
+
+        # Make sure the indexdata was up to date
+        copy.reindexObject()
+        reindexed_copy_indexdata = self.get_catalog_indexdata(copy)
+
+        # Some indexes get updated because of indexing like "modified"
+        # Some index data is not up to date, but does not have to be
+        # such as "is_subdossier".
+        # Other data should be up to date but is not...
+        not_up_to_date = ['Date', 'modified', 'start', 'Creator']
```
deiferni: same here that should not happen imo.

opengever/dossier/tests/test_move_items.py
```diff
@@ -54,6 +76,207 @@ def test_move_items_to_valid_target(self):
...
+            'modified': paste_time_index,
+            # 'start': paste_time_index,
+
+            # containing dossier and subdossier
+            'containing_dossier': self.empty_dossier.Title(),
+            'containing_subdossier': '',
+            # 'reference': 'Client1 1.1 / 4 / 41', # reference should be updated
+        }
+
+        unchanged_indexdata = [
```
deiferni: could you put that list somewhere central? otherwise we have high duplication.

opengever/dossier/tests/test_move_items.py
```diff
@@ -54,6 +76,207 @@ def test_move_items_to_valid_target(self):
...
+    def test_move_document_indexdata_update(self, browser):
+        self.maxDiff = None
+        self.login(self.regular_user, browser=browser)
+
+        subdocument_indexdata = self.get_catalog_indexdata(self.subdocument)
+
+        with freeze(self.MOVE_TIME):
+            self.move_items((self.subdocument, ), source=self.subdossier, target=self.empty_dossier)
+
+        moved = self.empty_dossier.objectValues()[-1]
```
deiferni: personal preference: you could use `observe_children` here

opengever/document/tests/test_copy_document.py
```diff
@@ -0,0 +1,231 @@
...
+
+        # We expect some of the metadata to get modified during pasting
+        modified_metadata = {'UID': copy.UID(),
+                             # sequence numbers and such
+                             'sequence_number': 41,
+                             'id': 'document-41',
+                             'getId': 'document-41',
+                             'reference': 'Client1 1.1 / 4 / 41',
+                             # creator
+                             'listCreators': ('kathi.barfuss',),
```
deiferni: also regular user?

opengever/document/tests/test_copy_document.py
```diff
@@ -0,0 +1,231 @@
...
+            'created': paste_time_index,
+
+            # containing dossier and subdossier
+            'containing_dossier': self.empty_dossier.Title(),
+            'containing_subdossier': '',
+            'is_subdossier': 0,  # acquisition is responsible here
+        }
+
+        unchanged_indexdata = [
+            'Creator', 'Date', 'Description', 'Subject', 'Type',
```
deiferni: why is `Creator` in here?

opengever/dossier/tests/test_move_items.py
```diff
@@ -54,6 +76,207 @@ def test_move_items_to_valid_target(self):
...
+            # containing dossier and subdossier
+            'containing_dossier': self.empty_dossier.Title(),
+            'containing_subdossier': '',
+            # 'reference': 'Client1 1.1 / 4 / 41', # reference should be updated
+        }
+
+        unchanged_indexdata = [
+            'id', 'getId', 'reference', 'sequence_number', 'is_subdossier',
+            'Title', 'sortable_title', 'SearchableText', 'changed', 'start',
+            'Creator', 'Date', 'Description', 'Subject', 'Type',
```
njohner: Creator should be changed

njohner:
We now update:
* `created`
* `Creator`
* `Date`
during copy/paste operations.

For move operations there still are some indexes that are not up to date:
* 'SearchableText': I'm not actually sure we really do want to update the `SearchableText` during a move (the only thing that changed is the reference number). 
* 'is_subdossier': Comes from acquisition of the value from the containing dossier. Is not used, but should probably be set to null for documents
* 'reference': Should get updated.
* 'start'
This comes from the `ftw.copymovepatches` where we only update selected indexes when moving an object.

opengever/document/tests/test_copy_document.py
```diff
@@ -0,0 +1,231 @@
...
+            'firstname', 'getContentType', 'getIcon', 'getObjSize',
+            'getRemoteUrl', 'has_sametype_children', 'in_response_to',
+            'is_folderish', 'is_subtask', 'issuer', 'last_comment_date',
+            'lastname', 'location', 'meta_type', 'phone_office', 'portal_type',
+            'predecessor', 'preselected', 'public_trial', 'receipt_date',
+            'responsible', 'retention_expiration', 'review_state', 'task_type',
+            'title_de', 'title_fr', 'total_comments', 'trashed']
+
+        # Make sure no metadata key is in both lists of unchanged and modified metadata
+        self.assertTrue(set(unchanged_metadata).isdisjoint(modified_metadata.keys()))
```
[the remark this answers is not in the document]
njohner: Sorry, forgot to integrate this change... will do

opengever/dossier/tests/test_move_items.py
njohner: not really. It's un purpose so that when someone adds a new index, he will have to think for each case tested whether he actually wanted the index to get updated or not...

deiferni: i think we already discussed it last week, IIRC i didn't see that in one test we test indices and in the other one metadata, so it should be 👍 as is.

opengever/document/tests/test_copy_document.py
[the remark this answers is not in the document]
njohner: done

PR #5692, opened 2019-05-28
Extend listing endpoint fields with file_extension and document_type.

phgross:
So the listing endpoint supports now filtering on file_extension and document_type. Closes https://github.com/4teamwork/gever-ui/issues/159.

### DIFF
--- a/opengever/api/tests/test_listing.py
+++ b/opengever/api/tests/test_listing.py
@@ -341,6 +341,28 @@ class TestListingEndpoint(IntegrationTestCase):
             [filename for filename in filenames if filename.endswith('.docx')],
             [item['filename'] for item in browser.json['items']])
 
+    @browsing
+    def test_filter_by_document_type(self, browser):
+        self.login(self.regular_user, browser=browser)
+
+        # all documents
+        view = '@listing?name=documents&columns:list=title&columns:list=start'
+        browser.open(self.repository_root, view=view, headers=self.api_headers)
+        number_of_documents = len(browser.json['items'])
+
+
+        # filtered on document_type
+        view = ('@listing?name=documents&'
+                'columns:list=title&columns:list=start'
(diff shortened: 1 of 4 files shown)


opengever/api/tests/test_listing.py
```diff
@@ -320,6 +320,40 @@ def test_tasks_listing(self, browser):
...
+
+        # docx documents only
+        view = ('@listing?name=documents&'
+                'columns:list=filename&columns:list=start'
+                '&filters.file_extension:record:list=.docx')
+        browser.open(self.repository_root, view=view, headers=self.api_headers)
+
+        self.assertEqual(
+            [filename for filename in filenames if filename.endswith('.docx')],
+            [item['filename'] for item in browser.json['items']])
```
njohner: I'dd add an assertion that you actually filtered something here:
`self.assertNotEqual(len(filenames), len(browser.json['items']))` or something like that

opengever/api/tests/test_listing.py
```diff
@@ -320,6 +320,40 @@ def test_tasks_listing(self, browser):
...
+
+    @browsing
+    def test_filter_by_document_type(self, browser):
+        self.login(self.regular_user, browser=browser)
+
```
njohner: Here too, I'd add a test of the precondition that there are other results without the filter

opengever/api/tests/test_listing.py
```diff
@@ -384,3 +418,16 @@ def test_filter_by_start_date(self, browser):
             'start:([2016-01-01T00:00:00.000Z TO 2016-01-01T23:59:59.000Z])',
             filters,
         )
+
```
njohner: Can you also test the filter by `document_type` with solr?

phgross:
@njohner updated ...

Issue #4844, opened 2018-09-05
Improve SOLR testing, test against SOLR instead of mocking.


Rotonen:
_Harvest: `OneGov GEVER` - `Softwarepflege (Refactoring/Bug)`_

_meta: invest the assigned days_

We already have the capability of testing agains a mocked SOLR.

See 
https://github.com/4teamwork/opengever.core/blob/1cff81b94320be2512a27841f1ea4c7d0f50f1c3/opengever/base/tests/test_search.py#L185
for details.

Candidates for improved testing:
1) Livesearch
2) Tabbedview
3) Any listings
4) live search endpoint

Rotonen:
Wishlist feature: have the integration testing fixtures automatically generate mock SOLR responses.

Rotonen:
And we should probably also end up in a state where https://github.com/4teamwork/opengever.core/pull/5163 does not happen again.

njohner:
Note that I implemented some `Livesearch` tests while working on #5018 (PR incoming)

deiferni:
Up for discussion: also testing against a real Solr may be a good idea, already done in https://github.com/4teamwork/ivbe.intranet/blob/master/ivbe/intranet/testing.py#L61. Thanks @elioschmutz for the hint/example.

Rotonen:
@deiferni @elioschmutz does that solr setup/teardown approach mean we cannot split that layer to run in parallel of itself as it'll conflict on both the port allocation and the filesystem resources it uses?

elioschmutz:
Parallel tests with this approach will be tricky.

I think we need to share the solr-instance but using one core for each testing-process. But this is going beyond my solr-knowledge.

The easier way would be to put all solr-layer tests within one proccess...

Rotonen:
@elioschmutz that'll still break with a static port allocation used on the layer, if multiples of the same test happen to run on the same server at the same time. With `opengever.core` this is fairly likely.

Rotonen:
I had a brief chat with @buchi and the way to go here is to have a shared Solr installation (per version of Solr we use) deployed to the CI infrastructure and have the test layers just create new cores with unique names over the REST API of Solr.

This is thus currently pending work by operations on the CI infrastructure.

Rotonen:
The shuffle nightlies have brought to attention a further caveat of the current status - it is currently very easy to do test class level leaks across tests as the tests manipulate `self.connection`.

[...]

PR #5694, opened 2019-05-29
Fix livesearch endpoint when using Solr

buchi:
Only strip Virtual Host Monster path if we are actually using the plone.restapi
SearchHandler.


buchi:
Currently we don't have a setup for tests with VHM and Solr. Adding a VHM setup is a little bit out of scope for this simple change. Thus I would defer that for later.
In general all e2e tests should use VHM as we always use it in production.

phgross:
Siehe auch https://github.com/4teamwork/opengever.core/issues/4844

PR #5697, opened 2019-06-03
Sort BlockedLocalRolesList on reference number.

### LINKED ISSUE #5670: Adjust sorting in the blocked-local-roles tab

The dossiers in the `blocked local roles tab` are not sorted. a sorting on the `reference number` would make sense.  

Bildschirmfoto_2019-05-17_um_08_55_56-2

njohner:
Testing is a bit weak I guess. There was a test which was probably flaky because we were not sorting the list of blocked local roles. Modifying the reference number of an object and checking that the order is updated would be better, but did not find a simple way of doing this and it did not quite seem worth investing more time in this.

resolves #5670

### DIFF
--- a/opengever/sharing/browser/admin_list.py
+++ b/opengever/sharing/browser/admin_list.py
@@ -1,4 +1,6 @@
 from opengever.base.browser.helper import get_css_class
+from opengever.base.interfaces import IReferenceNumberFormatter
+from opengever.base.interfaces import IReferenceNumberSettings
 from opengever.base.utils import escape_html
 from opengever.dossier.behaviors.dossier import IDossierMarker
 from opengever.repository.interfaces import IRepositoryFolder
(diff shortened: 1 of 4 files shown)


opengever/sharing/browser/admin_list.py
```diff
@@ -42,7 +42,7 @@ def render_tree(self):
             }
 
         dossier_container_brains = api.content.find(
-            context=self.context, **query_filter)
+            context=self.context, sort_on='reference', **query_filter)
```
lukasgraf: This will not work properly. The `reference` index contains a string, but we need to respect [natural sort order](https://en.wikipedia.org/wiki/Natural_sort_order) for reference numbers (11 getting sorted after 2). So we need to use the configured `ReferenceNumberFormatter`'s sorter ([example](https://github.com/4teamwork/opengever.core/blob/68ea385193f48605acac31c9c898bea7dd076a08/opengever/repository/browser/excel_export.py#L131-L137)).

njohner: Right, of course.

Issue #5200, opened 2018-12-13
Create plone content type for Period.


deiferni:
_Harvest: `OneGov GEVER` - `Softwarepflege (Refactoring/Bug)`_

Currently handling of periods is not implemented so well. If users perform actions in the wrong order numbering of meetings will be incorrect. Manually correcting these errors is time-consuming and requires a programmer. We would like to improve the period implementation so that less errors can occur.

At the moment periods are kept in a relation SQL database. We're about to move SQL data to plone content types wherever possible. Periods are only relevant for committees and don't need to be available across admin-units for multi admin-unit meeting support. Thus periods should be converted to plone content types, the sql model can be removed. The following changes are necessary:

- create a plone content-type for `Period`, `opengever.meeting.period`
- schema fields: 
  - `title`
  - `date_from`
  - `date_to`
  - `decision_sequence_number` (only visible for `Manager`)
  - `meeting_sequence_number` (only visible for `Manager`)
- ~~create a two-state workflow. to find the relevant period for a meeting, we use the meeting's date.~~
- create a one-state workflow. to find the relevant period for a meeting, we use the meeting's date.
- instead of fetching the current period based on a SQL query, we must switch to catalog queries
- periods created within a committee belong to the committee
- migration SQL-> plone
- display hint (warning) on meeting-page when no period for meeting date exists
- display error when attempting to decide meeting without periods

out of scope:
- relation from meeting to period when holding the meeting, might be implemented when the meeting is a plone content-type
- Allow (whom? `Administrator` and `CommitteeAdministrator `?) to change the sequence numbers currently used in the period.

Part of https://github.com/4teamwork/opengever.core/issues/5106

some todos so i shall not forget:
- [x] implement tests for `Period.get_current()`
- [x] add period add form defaults (title, start/end)

phgross:
Fachliches Review: 👍

PR #5699, opened 2019-06-03
Prefix agendaitem decision numbers and meeting number by correct period title

### LINKED ISSUE #5695: Fix display of agenda item decision number on meeting overview.
_Harvest: `[3.2.2.1] OneGov GEVER` - `Softwarepflege (Refactoring/Bug)`_

The decision numbers displayed on a meeting overview are not always correct. We always use the current decision number's year.

Screenshot 2019-05-29 at 15 50 21

See the current implementation: https://github.com/4teamwork/opengever.core/blob/7b49ed79e2bdf31df58b1e1614facb01f11b0066/opengever/meeting/model/agendaitem.py#L173-L175

Split off from https://github.com/4teamwork/opengever.core/issues/5200.

Proposed changes:
- Figure out the period based on the meeting date instead of using the current period
- Periods now have a title, display the title instead of hardcoding the year. The title defaults to the year but can be something else, e.g. for periods spanning multiple years.
- Also, no longer use `select for update` when only displaying 😱

njohner:
We were displaying the year of the current period as prefix to the agendaitem decision numbers. Instead we now the title of the period matching the date of the meeting as prefix. 

This is a bit better, but ultimately we should make sure that each meeting gets assigned to a given period as soon as the decision numbers and meeting numbers are generated (when the meeting is held), as these depend on the period. This will be tackled separately though (see https://github.com/4teamwork/opengever.core/issues/5200)

I also modified the displayed meeting number, as this should also depend on the period (and until now we simply displayed the year of the meeting...).

**meeting display if there is a period matching the committee and date of the meeting**

Screen Shot 2019-06-03 at 15 59 31

**meeting display if there is no period matching the committee and date of the meeting**

Screen Shot 2019-06-03 at 16 00 14

resolves #5695

### DIFF
--- a/opengever/meeting/model/agendaitem.py
+++ b/opengever/meeting/model/agendaitem.py
@@ -166,14 +166,17 @@ class AgendaItem(Base):
             self.description = description
 
     def get_decision_number(self):
-        # Before the meeting is held, agendaitems do not have a decision number
-        # and in that case we do not want to format it with the year prefixed
+        # Before the meeting is held, agendaitems do not have a decision number and
+        # in that case we do not want to format it with the period title prefixed
         if not self.decision_number:
             return None
 
-        period = Period.query.get_current_for_update(self.meeting.committee)
-        year = period.date_from.year
-        return '{} / {}'.format(year, self.decision_number)
+        period = Period.query.get_for_meeting(self.meeting)
+        if not period:
+            return str(self.decision_number)
+
+        title = period.title
+        return '{} / {}'.format(title, self.decision_number)
 
     def get_dossier_reference_number(self):
         if self.has_proposal:
--- a/opengever/meeting/model/agendaitem.py
+++ b/opengever/meeting/model/agendaitem.py
@@ -166,9 +166,10 @@ class AgendaItem(Base):
             self.description = description
 
     def get_decision_number(self):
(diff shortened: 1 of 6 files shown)


opengever/meeting/model/agendaitem.py
```diff
@@ -166,13 +166,17 @@ def set_description(self, description):
             self.description = description
 
     def get_decision_number(self):
-        # XXX huh? what is this?
+        # Before meeting is held, agendaitems do not have a decision number and
+        # in that case we do not want to format it with the period title prefixed
         if not self.decision_number:
             return self.decision_number
```
deiferni: as you have clarified the comment, i would also adapt the code to express better what is happening.

```python
if not self.decision_number:
    return None  # or ""?
```

opengever/meeting/model/meeting.py
```diff
@@ -207,6 +207,19 @@ def generate_meeting_number(self):
 
         self.meeting_number = period.get_next_meeting_sequence_number()
 
+    def get_meeting_number(self):
+        # Before meeting is held, it will not have a meeting number.
+        # In that case we do not want to format it with the period title prefixed
+        if not self.meeting_number:
+            return self.meeting_number
```
deiferni: same here

PR #5705, opened 2019-06-04
Add pull request/review checklist.

deiferni:
We would like to make the existing pull-request checklist more visible. In our opinion the best way to do this at the moment is by creating a pull-request template for github.

We move the existing pull request checklist from the wiki (currently at https://github.com/4teamwork/opengever.core/wiki/Prozess:-PullRequest-Review-Checkliste, will be removed once merged) and enrich it with https://basecamp.com/2768704/projects/9430593/todos/320310258. I've chosen the `.github` folder as that seems to be the convention adopted by other projects, e.g. https://github.com/4teamwork/ris.

I have compressed the list a bit, so we can start with something "simple". If there is an important point i forgot please let me know.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,31 @@
+_PR-Beschreibung: muss alle notwendigen Informationen enthalten, so dass ein Aussenstehender die Änderung versteht ohne den Code anzuschauen_
+
+- _Grund, wieso diese Änderung notwendig ist_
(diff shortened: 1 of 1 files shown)


deiferni:
Ok i have amended the checklist with with:
```
- [ ] Gibt es ein neues Feature-Flag? Wurden dafür Tests mit aktiviertem und deaktivierte Flag geschrieben?
```

PR #5708, opened 2019-06-05
Avoid certain test classes being run multiple times.

njohner:
When the name of a test class TC1 is contained in the name of another test class TC2 from the same python file, the tests for TC1 are potentially run more than once by mtest (as TC1 is also matched when we try to run the test TC2). As matching of the search patterns is done on strings of the form 'testname (test.class.qualified.name)', we can do exact matching if we include the parenthesis in our search patterns.

### DIFF
--- a/bin/mtest
+++ b/bin/mtest
@@ -251,7 +251,10 @@ def run_tests(test_run_params):
 
     for testclass in testclasses:
         params.append('-t')
-        params.append(testclass)
+        # we include the parenthesis in the classname here, this
+        # essentially leads to exact matching being done later on
+        # when we use the classname to run the tests for that class
+        params.append("\\(" + testclass + "\\)")
 
     printable_params = ' '.join(["'{}'".format(param) if ' ' in param else param for param in params])
 


njohner:
Ok I think this is ready. You can see the results on the first two commits. In the first one I introduce two failing tests in problematic classes (which are also matched by another test class). You can see in the output on CI that they fail twice each. Then introduce the correction in `mtest`, to show that each test now fails just once. Finally reverted the commit introducing the failing tests, to show that tests pass.

bin/mtest
```diff
@@ -251,7 +251,10 @@ def run_tests(test_run_params):
 
     for testclass in testclasses:
         params.append('-t')
-        params.append(testclass)
+        # we include the parenthesis in the classname here, this
+        # essentially leads to exact matching being done later on
+        # when we use the classname to run the tests for that class
+        params.append("\\(" + testclass + "\\)")
```
Rotonen: I'd rather you amend the class match regex in the discovery to include the parens in the capture group.

You simply need to move the capture group parens out one level onto the other sides of the escaped parens, like this:
`classname = re.search(r'.*(\(.*\)).*', line).groups()[0]`

Rotonen: In general: use raw strings like `r"\("` if you need to represent literal backslashes.

And adding anything too subtle into the function we map across the multiprocessing pool is to be avoided as debugging that is madness (if you ever do need to do that, import the undocumented multiprocessing thread pool and use that as the pool for debugging purposes).

njohner:
Yes, I first had moved the parenthesis into the class matching, but then realised I still needed to escape it afterwards, as it's used again in a regex...

PR #5712, opened 2019-06-06
Bump python-docx to 0.8.10

Rotonen:
Tests pass, should probably get manually validated as well.


Rotonen:
@phgross per [changelog](https://github.com/python-openxml/python-docx/blob/master/HISTORY.rst) the changes are trivial and not on a critical path for us, but manual validation by someone whom knows better what to suspect would be prudent.

Rotonen:
Now superseded by https://github.com/4teamwork/opengever.core/pull/5660

PR #5717, opened 2019-06-11
Add an ftw.tesbrowser widget for filling responsible(s) in the tests

### LINKED ISSUE #5330: Can't prefill task responsible via browser.fill, must use workaround.

We currently can't fill a task responsible in tests using something like

```python
browser.fill({'Responsible': 'blah.blub'})
```

but instead must use the following workaround:

https://github.com/4teamwork/opengever.core/blob/938aff8ead15bc59766e52e20c836cfde5bbdfa7/opengever/task/tests/test_sequential_task_process.py#L205-L209

Maybe properly registering an `ftw.testbrowser` widget would help in such cases.

Rotonen:
The convolution around the `AsyncKeywordWidget`, task template wizards and the registration-at-module-init-time magic of the `ftw.testbrowser` widgets threw me off for a bit.

This PR adds easier handling for the most common use case for the responsible fields, users. It can later be further amended to handle teams and inboxes.

Closes #5330

### DIFF
--- a/opengever/testing/testbrowser_responsible_user_widget.py
+++ b/opengever/testing/testbrowser_responsible_user_widget.py
@@ -1,0 +1,74 @@
+from ftw.keywordwidget.tests.widget import AsyncKeywordWidget
+from ftw.testbrowser.widgets.base import widget
+from opengever.ogds.base.utils import get_current_org_unit
+from Products.CMFCore.MemberDataTool import MemberData
+
+
+@widget
+class OpengeverResponsibleUsersAsyncKeywordWidget(AsyncKeywordWidget):
+    """Autofill the correct admin unit for a responsible user."""
+
+    @staticmethod
+    def match(node):
+        """Match an async keyword widget responsible(s) field.
+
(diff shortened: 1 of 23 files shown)


opengever/testing/testbrowser_responsible_user_widget.py
```diff
@@ -0,0 +1,85 @@
...
+            matching_suffixes = ("responsible", "responsibles")
+            if fieldname and fieldname.split(".")[-1] in matching_suffixes:
+                return True
+        return False
+
+    def fill(self, users, auto_org_unit=True):
+        """Fill the AsyncKeywordWidget, auto.
+
+        Mostly copied over from
+        ``ftw.keywordwidget.tests.widget.AsyncKeywordWidget``.
```
deiferni: why copy the filling functionality? can't you first process the input user objects to list of strings and then do a super call?

Rotonen: Thank you. Will amend.

PR #5723, opened 2019-06-13
Solr update

buchi:
Update Solr and ftw.solr to latest versions.

ftw.solr now performs soft commits to allow near real time (NRT) searching.
This make documents available for search almost immediately after being indexed.
E.g. after uploading, documents are now visible immediately in listings in the new frontend.

Apart from that, Solr 8.1.1 comes with various bug fixes and improvements. See https://lucene.apache.org/solr/8_1_1/changes/Changes.html

Other changes:
- Tike exceptions are now properly ignored (e.g. with encrypted documents)
- Increased auto commit max time to 60s.

### DIFF
--- a/solr-conf/solrconfig.xml
+++ b/solr-conf/solrconfig.xml
@@ -35,7 +35,7 @@
        that you fully re-index after changing this setting as it can
        affect both how text is indexed and queried.
   -->
-  <luceneMatchVersion>7.3.1</luceneMatchVersion>
+  <luceneMatchVersion>8.1.1</luceneMatchVersion>
 
   <!-- <lib/> directives can be used to instruct Solr to load any Jars
        identified and use them to resolve any "plugins" specified in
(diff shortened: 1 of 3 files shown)


solr-conf/solrconfig.xml
```diff
@@ -803,8 +810,8 @@
       <str name="lowernames">false</str>
       <str name="xpath">/xhtml:html/xhtml:body//text()</str>
       <str name="fmap.content">SearchableText</str>
-      <bool name="ignoreTikaException">true</bool>
       <str name="uprefix">ignored_</str>
+      <str name="ignoreTikaException">true</str>
```
deiferni: this looks fishy, please double check that `str` is correct here.

buchi: I've tested against Solr 8.1.1. Actualy both variants work. Thus I'm keeping the bool parameter.

PR #5724, opened 2019-06-13
Fix encoding issue in OGDSUpdater's error logging.

lukasgraf:
Fix encoding issue in `OGDSUpdater`'s error logging.

When logging errors about a referenced user not being found in LDAP, the error logging blew up with an encoding error when a user with non-ASCII character in its DN was logged, because a UTF-8 bytestring was formatted into a unicode message.

This changes all the format strings to use the `repr()` of the respective values to avoid such errors in the future.

(Discovered via https://extranet.4teamwork.ch/betrieb/tracker/880)
Sentry Issue: https://sentry.4teamwork.ch/sentry/onegov-gever/issues/30789/

## Checkliste

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2019.3.0 (unreleased)
 ---------------------
 
+- Fix encoding issue in OGDSUpdater's error logging. [lgraf]
 - Provides support for some additional metadata on the search endpoint. [phgross]
 - Include file_extension in API representation of documents. [phgross]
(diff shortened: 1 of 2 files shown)


njohner:
Oh and is there an issue or an extranet ticket, a Sentry issue or such that goes with this?

lukasgraf:
@njohner amended PR body with link to Extranet issue.
As for tests, we had 4teamwork/opengever.core#92 for those. Ironically, we closed this during the spring cleaning just 10 days ago 😉
So no, can't currently test this code path unfortunately.

Issue #5725, opened 2019-06-13
Bump plone.restapi


lukasgraf:
For 2019.4 we should bump `plone.restapi`.

Things that need to be addressed as part of the bump:
- Our [custom `ChoiceFieldDeserializer`](https://github.com/4teamwork/opengever.core/blob/8d18a7ff39dd6cb31b51635d88344ee1869c2e6c/opengever/api/fields.py#L12-L27) (registered via `overrides.zcml`) should be removed. There now is [one in plone.restapi](https://github.com/plone/plone.restapi/blob/25446902caaec5b7e930f8c0d8bc198e5ddf7ed3/src/plone/restapi/deserializer/dxfields.py#L117-L130).
- The `ChoiceFieldDeserializer` in `plone.restapi` also validates the value before returning it. This currently leads to several test failures. After removing our own serializer, please run the following tests:
```
bin/test -t test_updates_responsible_and_responsible_client -t test_assign_sets_forwarding_in_open_state -t test_assign_add_corresonding_response -t test_activity_is_recorded_correctly -t test_updates_responsible_if_new_value_is_given -t test_remove_task_reminder_of_old_responsible
```
- One issue seems to be that the [`ForwardingReassignTransitionExtender`](https://github.com/4teamwork/opengever.core/blob/8d18a7ff39dd6cb31b51635d88344ee1869c2e6c/opengever/inbox/transition.py#L122-L125) uses the same schema as the [`ReassignTransitionExtender`](https://github.com/4teamwork/opengever.core/blob/master/opengever/task/transition.py#L227-L270) for tasks. That extender has the [`INewResponsibleSchema`](https://github.com/4teamwork/opengever.core/blob/8d18a7ff39dd6cb31b51635d88344ee1869c2e6c/opengever/task/transition.py#L76-L91) associated with it, which configures the `AllUsersInboxesAndTeamsSourceBinder` (used for the new responsible)  with `only_current_inbox=True`. That is correct for tasks, but not forwardings. When reassigning tasks, the org unit is not allowed to change, but forwardings can be refused and assigned to the inbox of a _different_ org unit.

- Currently, when bumping to the most recent version of `plone.restapi`, another of our GEVER tests will fail, because the status code for non-existing relations changes from 400 to 500. This will partly be addressed by https://github.com/plone/plone.restapi/pull/778
(Failing tests to check for this problem: `bin/test -t test_journal`)

PR #5726, opened 2019-06-14
Register ChoiceFieldDeserializer using overrides instead of configure ZCML

lukasgraf:
This ensures that it will be loaded with the currently pinned version of `plone.restapi` (which doesn't have it's own `ChoiceFieldDeserializer` yet) and will also not cause conflicting configurations with `plone.restapi:master`, which now [does have a `ChoiceFieldDeserializer`](https://github.com/plone/plone.restapi/blob/25446902caaec5b7e930f8c0d8bc198e5ddf7ed3/src/plone/restapi/deserializer/dxfields.py#L117-L130).

That way we make sure that GEVER tests against `plone.restapi` are green again and don't shadow problems that might be introduced in the upcoming restapi sprint.

## Checkliste

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2019.3.1 (unreleased)
 ---------------------
 
+- Register ChoiceFieldDeserializer using overrides instead of configure ZCML. [lgraf]
 - Docs: Add tasks to documented content types. [lgraf]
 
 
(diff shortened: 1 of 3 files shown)


lukasgraf:
@deiferni indeed, we can (should) drop this one once we bump `plone.restapi`.
Rright know, things go slightly kapputt when we just change the implementation the the `ChoiceFieldSerializer` from stock `plone.restapi`, the probably causes / necessary changes are tracked via #5725

PR #5727, opened 2019-06-14
Minor cleanups for cleaner test output / linter in tests

Rotonen:
So, I started hunting for quick wins in test speeds based on the output of three log files:
[2019-06-13-functionaltestperf.log](https://github.com/4teamwork/opengever.core/files/3290314/2019-06-13-functionaltestperf.log)
[2019-06-13-integrationtestperf.log](https://github.com/4teamwork/opengever.core/files/3290315/2019-06-13-integrationtestperf.log)
[2019-06-13-othertestperf.log](https://github.com/4teamwork/opengever.core/files/3290316/2019-06-13-othertestperf.log)

I quickly hit a wall as to some event ordering issue in the test server self test:
https://ci.4teamwork.ch/builds/249683/tasks/409536

The best guess is that the SQLite DB Engine gets initialized before ZCML is registered. We have an event for forcibly enabling save points for integration testing.

https://github.com/4teamwork/opengever.core/blob/1cefc45c2ff8d521dbb0c5dcd3c0bb293b948a23/opengever/ogds/base/configure.zcml#L25-L28

https://github.com/4teamwork/opengever.core/blob/7390b2e4afd830f31cec55285d91bdf490b041fa/opengever/ogds/base/events.py#L43-L63

https://github.com/4teamwork/opengever.core/blob/7390b2e4afd830f31cec55285d91bdf490b041fa/opengever/ogds/base/events.py#L14-L20

I've left a branch behind:
https://github.com/4teamwork/opengever.core/tree/jo-test-speedups

So I've just ended up cleaning the test output as much as I could. Everything remaining should be from other packages and cannot be tackled in `opengever.core`. A good example of these are various exceptions used upstream. These are mostly already sanitized to the new style of how to do exceptions in the Python 3 transition. An another one fixed by that transition is Mocker explicitly and unconditionally overriding `assertTrue` to be `failUnless`.

[2019-06-13-deprecations.log](https://github.com/4teamwork/opengever.core/files/3290361/2019-06-13-deprecations.log)

There is now an issue on `ftw.testbrowser` as well, but I could not tackle those, as something mystical happens to timezone handling with the requests driver when bumping zope.testbrowser:
https://github.com/4teamwork/ftw.testbrowser/issues/93

### DIFF
--- a/opengever/document/tests/test_checkout.py
+++ b/opengever/document/tests/test_checkout.py
@@ -331,7 +331,8 @@ class TestCheckinCheckoutManager(FunctionalTestCase):
         # We cannot freeze time due to the test browser being threaded
         oc_url = create_oc_url(
             self.doc1.REQUEST, self.doc1, {'action': 'checkout'})
-        decoded_oc_url = jwt.decode(oc_url.split(':')[-1], verify=False)
+        decoded_oc_url = jwt.decode(
+            oc_url.split(':')[-1], verify=False, algorithms=('HS256',))
 
         redirector_js = browser.login().open(
             self.doc1,
@@ -347,7 +348,7 @@ class TestCheckinCheckoutManager(FunctionalTestCase):
 
         parsed_oc_url = tokens_from_js[0]
         decoded_parsed_oc_url = jwt.decode(
-            parsed_oc_url.split(':')[-1], verify=False)
+            parsed_oc_url.split(':')[-1], verify=False, algorithms=('HS256',))
 
         # Take out the timestamps
         del decoded_oc_url['exp']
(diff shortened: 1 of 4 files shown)


njohner:
The less the deprecated methods are used, the less they will be used in the future though (copy/paste)... what is the downside of leaving the changes? Merge conflicts?

deiferni:
@njohner imho there is no real benefit to the rename and it dilutes the commit history unnecessarily. If i've seen correctly the deprecation is just stated in the docs, there are no deprecation warnings. I'm all for the rename as soon as we prevent that these method will be used in the future. Otherwise i'd suggest to omit the changes until we find a way to do so, or until we have a python version which no longer provides these methods. Renaming should be simple.

njohner:
Fine with me! If we really want to drop support, we can probably simply delete the methods on our `IntegrationTestCase` and `FunctionalTestCase` no? `del(IntegrationTestCase.assertEquals)`

deiferni:
I have hijacked this branch and modified it as suggested.

PR #5728, opened 2019-06-17
Fix validation of resolve preconditions in archive form

lukasgraf:
The archive form relied on the `DossierResolveView` already **validating dossier resolution preconditions**, and presenting error messages to the user. Because that wasn't the case any more after the refactoring of the whole dossier resolution process, the archive form was broken when preconditions weren't fulfilled.

This change fixes this by:
- Refactoring the dossier resolution code so that:
  - Raising the appropriate exceptions is done in a single method on the resolver, `raise_on_failed_preconditions()` 
  - Constructing and displaying error messages for classic views, including redirecting to the appropriate context, is done in a mixin class. That way it can be reused in the archive form.
  - Handling violated preconditions in both
    - The `DossierResolveView`, before even redirecting to the archive form
    - The `ArchiveForm` itself, in case `transition-archive` is invoked directly

This approach leads to the preconditions getting check _twice_ in some cases. We should eventually optimize this, possibly when cleaning up the archive form code next.

Fixes https://basecamp.com/2768704/projects/14549453/todos/391204514

❗️ Needs backport to `2019.2-stable`

## Review notes

The diff for `cbe0071` is a mess, as displayed by git. Easier to look at the resulting code. Just moved out the status message related stuff from the [`DossierResolveView`](https://github.com/4teamwork/opengever.core/blob/cbe0071d105973e96189ae7951034f4f701a4ff3/opengever/dossier/resolve.py#L221-L253) into the [`DossierResolutionStatusmessageMixin`](https://github.com/4teamwork/opengever.core/blob/cbe0071d105973e96189ae7951034f4f701a4ff3/opengever/dossier/resolve.py#L159-L218) class (_these are links to the new classes, after the refactoring_).

## Checkliste

- [x] Könnten Kundeninstallationen von den Änderungen betroffen sein? Müssen Policies angepasst werden?

_Ja, potentiell haben diese Änderungen Einfluss auf die Logik für den Dossier-Abschluss, sowohl bei Installation mit aktiviertem Ablagenummern-Feature wie auch bei anderen. Das extern sichtbare Verhalten sollte sich aber nicht ändern. Keine Anpassungen an Policies nötig._
 
_Nein. Durch das wiederverwenden der Fehler-Übersetzungen in der Mixin-Class ist dies nicht nötig._

### DIFF
--- a/opengever/dossier/resolve.py
+++ b/opengever/dossier/resolve.py
@@ -149,16 +149,7 @@ class LockingResolveManager(object):
             resolve_lock.release(commit=True)
 
     def execute_recursive_resolve(self):
-        # check preconditions
-        errors = self.resolver.get_precondition_violations()
-        if errors:
-            raise PreconditionsViolated(errors=errors)
-
-        # validate enddates
-        invalid_dates = self.resolver.are_enddates_valid()
-        if invalid_dates:
-            raise InvalidDates(invalid_dossier_titles=invalid_dates)
-
+        self.resolver.raise_on_failed_preconditions()
         self.resolver.resolve()
 
     def is_archive_form_needed(self):
@@ -270,6 +261,19 @@ class StrictDossierResolver(object):
         if not errors:
             self.preconditions_fulfilled = True
         return errors
+
+    def raise_on_failed_preconditions(self):
+        """Verify preconditions, and raise respective exceptions if violated.
+        """
+        # check preconditions
+        errors = self.get_precondition_violations()
+        if errors:
+            raise PreconditionsViolated(errors=errors)
+
+        # validate enddates
+        invalid_dates = self.are_enddates_valid()
+        if invalid_dates:
+            raise InvalidDates(invalid_dossier_titles=invalid_dates)
 
     def are_enddates_valid(self):
         """Check if the end dates of dossiers and subdossiers are valid.
--- a/opengever/dossier/resolve.py
+++ b/opengever/dossier/resolve.py
@@ -227,21 +227,36 @@ class DossierResolveView(BrowserView, DossierResolutionStatusmessageMixin):
 
         resolve_manager = LockingResolveManager(self.context)
 
-        if resolve_manager.is_archive_form_needed():
-            archive_url = '/'.join((self.context_url, 'transition-archive'))
-            return self.redirect(archive_url)
-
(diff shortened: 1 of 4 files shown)


opengever/dossier/resolve.py
```diff
@@ -156,44 +156,26 @@ def is_archive_form_needed(self):
...
-            return self.show_being_resolved_msg()
+    for dossier resolution errors.
 
-        except PreconditionsViolated as exc:
-            return self.show_errors(exc.errors)
-
-        except InvalidDates as exc:
-            return self.show_invalid_end_dates(titles=exc.invalid_dossier_titles)
+    This class handles producing these status messages in a classic view
+    (as opposed to the REST API, where errors are serialzed as JSON).
```
njohner: typo: serialized

opengever/dossier/tests/test_archiver.py
```diff
@@ -364,3 +364,30 @@ def test_only_give_filing_number_on_resolved_dossier(self, browser):
         statusmessages.assert_message('The filing number has been given.')
         self.assertEquals('Hauptmandant-Government-2017-1',
                           IFilingNumber(self.empty_dossier).filing_no)
+
+    @browsing
+    def test_precondition_violation_raises_error_alrady_on_resolve_view(self, browser):
```
njohner: typo: already

opengever/dossier/resolve.py
```diff
@@ -227,22 +227,37 @@ def __call__(self):
...
+        # cleaned up so we don't unnecessarily validate preconditions multiple
+        # times.
+        try:
+            resolve_manager.resolver.raise_on_failed_preconditions()
+
+        except PreconditionsViolated as exc:
+            return self.show_errors(exc.errors)
+
+        except InvalidDates as exc:
+            return self.show_invalid_end_dates(titles=exc.invalid_dossier_titles)
```
njohner: Why not move that `try` `except` block into the `if resolve_manager.is_archive_form_needed()` block? Then we would not change anything for the case where the archive form is not needed, and also avoid checking twice in that case?

lukasgraf: As discussed: That's what I attempted first, but it made the code quite unreadable.
So for now I opted for readability at the cost of performing the checks twice, plus a comment that we need to clean this up.

PR #5730, opened 2019-06-18
Add configuration for notifications for own actions and inbox actions

### LINKED ISSUE #5465: Add global configuration option for inbox notifications.

We want to implement a configuration option in the users settings so that each user can globally enable/disable inbox notifications for himself.

Proposed Implementation:
- new tab `Allgemein` in `@@personal-preferences`, see https://github.com/4teamwork/opengever.core/issues/5466
- display checkbox `Benachrichtigungen für Eingangskorb aktivieren`, in that tab, default is `True`
- store boolean value in new column on `opengever.ogds.models.user.User` (`NotificationSetting` is not built for single boolean per-user settings)
- when generating inbox notifications respect the user-settings.

Related to https://github.com/4teamwork/opengever.core/issues/5466

https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/629

### LINKED ISSUE #5466: Add global configuration option for notifications for actions by current user.

According to https://feedback.onegovgever.ch/t/benachrichtigungen-an-sich-selber-aktivierbar-machen/1360 some users would like to receive notifications for actions they just triggered in GEVER themselves. 

We want to implement a configuration option in the users settings so that each user can globally enable/disable e-mail notifications for his own actions.

Proposed Implementation:
- new tab `Allgemein` in `@@personal-preferences`, see https://github.com/4teamwork/opengever.core/issues/5465
- display checkbox `Benachrichtigung für eigene Aktionen`, in that tab, default is `False`
- store boolean value in new column on `opengever.ogds.models.user.User` (`NotificationSetting` is not built for single boolean per-user settings)
- when generating notifications respect the user-settings

Related to https://github.com/4teamwork/opengever.core/issues/5465

https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/630

njohner:
We add two new per user notification configurations. These are global settings valid for all actions and roles for a given user:
* **Notify own actions**: Some users would like to receive notifications for their own actions in GEVER. This was purposefully not so until now as most users would consider that spam. We therefore implement a user setting, defaulting to `False`, which allows users to globally set whether they want to receive notifications for their own actions or not.
* **Inbox actions notifications**: Inbox users should be able to decide whether they want to be notified for actions that concern the whole group of inbox users. Here the default is `True` and the configuration allows a user to turn such notifications off.

As the `NotificationSetting` is not build for single boolean per-user settings, we store these configurations in new columns on `opengever.ogds.models.user.User`. These configurations can be set by the user, in a new tab `General` in the `NotificationSettings` view. To keep the UI consistent with other settings from the `NotificationSettings`, we implement `Save` and `Cancel` buttons that appear when the user changes a configuration. If his configuration is different from the default there is also a `reset` button.

Note that the view does not work for zopemaster, as the user needs to be available in the db.

**General tab in the `NotificationSettings` view**
Screen Shot 2019-06-24 at 15 27 57

**When modified, `save` and `cancel` buttons appear**
Screen Shot 2019-06-24 at 15 28 01

**When setting is different from default, the `reset` button appears**
Screen Shot 2019-06-24 at 15 28 06

resolves #5466, resolves #5465

## Checkliste
- [ ] Sind UpgradeSteps `deferrable`, oder können gewisse Schritte des Upgrades konditional ausgeführt werden? Upgrade is not deferrable. I guess we could make it deferrable (would need to make sure we use `getattr(`notify_own_actions`, False)` everywhere though), but I don't think it's worth the risk/effort
- [x] Aktualisierung Dokumentation vorhanden/nötig? Muss ich noch machen


opengever/activity/browser/settings.py
```diff
@@ -206,6 +250,14 @@ def get_role_translations(self):
     def dispatchers(self):
         return notification_center().dispatchers
 
+    def get_user_configuration(self, config_name):
+        userid = api.user.get_current().getId()
+        user = User.query.filter_by(userid=userid).first()
```
lukasgraf: `.one()`

docs/public/user-manual/benachrichtigung.rst
```diff
@@ -61,4 +61,9 @@ zwei Schritte beachtet werden:
 .. |img-benachrichtigungs-einstellungen-6| image:: img/media/img-benachrichtigungs-einstellungen-6.png
 .. |img-benachrichtigungs-einstellungen-7| image:: img/media/img-benachrichtigungs-einstellungen-7.png
 
+
+Allgemeine Einstellungen für Benachrichtigungen
+-----------------------------------------------
+Im Reiter "Allgemein" befinden sich die globalen Einstellungen. Dort können Sie einstellen ob Sie auch für Eure Eigenen Aktionen benachrichtigt werden wollen.
```
lukasgraf: -> `einstellen, ob Sie auch für Ihre eigenen`

opengever/activity/model/activity.py
```diff
@@ -37,15 +38,20 @@ def __repr__(self):
...
+        notify_own_actions'setting.
         """
 
         notifications = []
         for userid in self.get_users_for_watchers():
-            if not self.is_current_user(userid):
-                notifications.append(
-                    Notification(userid=userid, activity=self))
+            if (self.is_current_user(userid) and not
+                    self.user_wants_own_action_notifications(userid)):
```
lukasgraf: 💚

opengever/activity/model/activity.py
```diff
@@ -65,6 +71,10 @@ def get_notifications_for_watcher_roles(self, roles):
     def is_current_user(self, user_id):
         return user_id == self.actor_id
 
+    def user_wants_own_action_notifications(self, userid):
+        user = User.query.filter_by(userid=userid).first()
```
lukasgraf: `.one()`

lukasgraf:
I would consider adding a help text to the UI for the setting. Something like

> Standardmässig werden keine Benachrichtigungen ausgelöst in Fällen, wo der Empfänger der Benachrichtigung dieselbe Person wäre, welche die Aktion gerade durchgeführt hat. Mit dieser Option kann das Unterdrücken dieser Benachrichtigungen ausgeschaltet werden.
>
>Bitte beachten Sie, dass nach wie vor auch die Benachrichtigung für die enstprechenden Typ-spezifischen Vorgänge und Rollen aktiviert sein muss, um für einen konkreten Vorgang eine Benachrichtigung zu erhalten.

I would also consider making the label sightly more specific:
`[ ] Benachrichtigung für eigene Aktionen` -> `[ ] Benachrichtigung für eigene Aktionen nicht unterdrücken`

docs/public/user-manual/benachrichtigung.rst
njohner: 😅

njohner:
Ok all done! There were only small corrections. Maybe someone wants to have a look at Lukas' suggestion for the new label and help text?

njohner:
Ok I've made all the required changes. As I was already working on #5465, on a branch based on this one, and realised that the selectors used in the JS were not working with several configurations in the "General" tab, and also that the `Save`, `Cancel` and `Reset` buttons were missing their icons, I ended up simply resolving the two issues in this single PR. So here is a list of the changes:
* Correct Selectors in JS to work for multiple configurations in the General Tab
* Add icons to the `Save`, `Cancel` and `Reset` buttons
* Correctly extract the value of the configurations from the request with `json.loads`
* Add configuration to deactivate inbox notifications.
* Only notify inbox users when notify_inbox_actions is set to True

PR #5737, opened 2019-06-19
Fix creating incorrect metadata changed journal entries for OC file uploads.

### LINKED ISSUE #5735: Incorrect label when checking in a document from officeconnector with rest api enabled flag..
The actual label:

Screenshot 2019-06-19 at 16 09 23

The expected label:

Screenshot 2019-06-19 at 16 10 41

Origin: https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/763

deiferni:
🚧 _it has turned out modification detection is a bit meh. investigating._

This PR fixes creation of journal entries when the OfficeConnector API upload feature is enabled.

- Add the correct field to the detection logic that determines if the file has been modified
- Fix an evil default that could not have worked outside web-forms

Fixes #5735.

## Checkliste

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2019.3.1 (unreleased)
 ---------------------
 
+- Fix journal entries for OfficeConnector API file uploads. [deiferni]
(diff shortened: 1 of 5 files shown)


deiferni:
It isn't as easy as i first thought. On hold, journal entries should be discussed.

PR #5752, opened 2019-06-25
@listing endpoint: Add support for filtering by relative path depth.

### LINKED ISSUE #5513: Extend listing endpoint with depth parameter
_Harvest: `Projekt - Umsetzung GEVER-RIS` - `GEVER API`_

In RIS, we need to display only the immediate documents of a dossier, *without* the documents of subdossiers. An additional parameter for the `@listing` endpoint would be very helpful, similar to the `depth` query string parameter of the `@search` endpoint (https://plonerestapi.readthedocs.io/en/latest/searching.html?highlight=depth#query-options).

For our use case, we cannot use the `@search` endpoint as we need specific `columns` of the documents (e.g. `pdf_url` and `checked_out`), which are only available for the `@listing` endpoint.

### Update:
`path_depth` indexer has been introduced in ftw.solr https://github.com/4teamwork/ftw.solr/pull/132

Nur Solr Variante muss `depth` Parameter unterstützen.

lukasgraf:
⚠️ Depends on 4teamwork/ftw.solr#134 ⚠️ 

This adds support for **filtering by `depth` (relative to the context)** to the `@listing` endpoint.

It does so by making use of the support for `path_depth` introduced in 4teamwork/ftw.solr#132:

- Bump `ftw.solr`
- Include `path_depth` in our `managed-schema`
- Add upgrade step that indexes `path_depth` (only) in Solr
- Add support for `depth` in `@listing` endpoint's Solr search:
  - Determine relative `depth` query option
  - Determine current context path, depth using the [same logic](https://github.com/4teamwork/ftw.solr/blob/2.4.0/ftw/solr/handlers.py#L109-L110) as in `ftw.solr`
  - Determine max absolute path depth: `current_context_depth + relative_depth`
  - Extend Solr filters with an according "lessOrEqual" filter for that absolute path depth

Fixes #5513

I noticed that the Solr-Implementation of the `@listing` endpoint does not currently filter out the searchroot - created follow-up issue #5757

Resolved questions:

- [x] Check whether upgrade step can be made deferrable
  _Yes - if `path_depth` is not reindexed, any query to the `@listing` endpoint that includes `depth` will return no results. But that shouldn't become an issue until `depth` is actually getting used by the frontend_
- [x] Check whether `stored=False` is actually correct for `path_depth` in `managed-schema`
  _Yes - atomic updates will work if a field is either `stored` or has `docValues`. The `pint` type used for the`path_depth` field has `docValues` enabled, so `stored=False` is correct._

## Checkliste

  _Includes Upgrade-Step that indexes `path_parent` to (only) Solr_

  _Yes, upgrade step is deferrable._


opengever/api/tests/test_listing.py
```diff
@@ -491,3 +512,21 @@ def test_filter_by_document_type(self, browser):
...
         self.assertIn(u'document_type:(contract)', filters)
+
+    @browsing
+    def test_filter_by_depth(self, browser):
+        self.login(self.regular_user, browser=browser)
+
+        context = self.dossier
+
+        # Guard assertion - we expect self.dossier to be on level 5
+        self.assertEqual(5, len(context.getPhysicalPath()) - 1)
```
njohner: I think it could be worth having a method somewhere that calculates the `path_depth`, and then use it here in the test and also in the listing endpoint. Because if we modify the way the depth is calculated in the listing this test will still pass...

lukasgraf:
> Could you make sure that the `path_depth` parameter does get updated properly in solr during copy/paste and move operations?

Nice catch 👍  -> 4teamwork/ftw.solr#134

versions.cfg
```diff
@@ -96,7 +96,7 @@ ftw.recipe.solr = 1.2.1
 ftw.recipe.translations = 1.2.6
 ftw.showroom = 1.5.1
 ftw.slacker = 1.0.2
-ftw.solr = 2.3.2
+ftw.solr = 2.4.0
```
buchi: Will need a newer version because of https://github.com/4teamwork/ftw.solr/pull/134

opengever/core/upgrades/20190625184855_index_path_depth_in_solr_for_all_objects/upgrade.py
```diff
@@ -0,0 +1,24 @@
...
+from plone.dexterity.interfaces import IDexterityContent
+from zope.component import getUtility
+
+
+class IndexPathDepthInSolrForAllObjects(UpgradeStep):
+    """Index path_depth in Solr for all DX objects.
+
+    This upgrade step is considered deferrable. Until it's executed and
+    path_depth has been indexed in Solr, any query to the @listing endpoint
+    *that includes a `depth` limit* will return zero results.
```
njohner: Really? My impression is that if `path_depth` is not indexed, `depth` gets set to `-1`, so results for all depths are returned  as before no?

lukasgraf: @njohner as discussed: It should behave the way I described in the docstring. The *parsing* of the `depth` query parameter is independent of whether any values for `path_depth` have been indexed in Solr. If `depth` happens to be parsed and end up being `-1`, then yes, the query isn't even restricted by depth (filter is omitted from the Solr query), and that will always work.

If the Solr query contains  *any* `path_depth` filter whatsoever though, it will return zero results if theres no indexed values for `path_depth`.

PR #5761, opened 2019-06-27
Add optional facet values and counts search to listing endpoint

### LINKED ISSUE #5738: Facetten Support im listing endpoint
_Harvest: `Projekt - Umsetzung Neues User Interface GEVER 2019/2020`-`Entwicklung`_

Der Listing Endpoint sollte vie optionalen Parameter die Abfrage von Facetten Wertebereichen ermöglichen. 

Die Abfrage des Dossier-Listings Endpoints soll zusätzlich zu den Inhalten auch Werte-Bereiche von möglichen Facetten liefern. Also bspw. die häufigsten Federführenden der aktuellen Trefferliste.

njohner:
For the new GEVER UI we need the possibility to optionally query for facet available values on the listing endpoint. 

resolves #5738 

## Checkliste

- [x] Still need to bump ftw.solr instead of installing from source once it is released

### DIFF
--- a/opengever/api/listing.py
+++ b/opengever/api/listing.py
@@ -1,12 +1,15 @@
+from collective.elephantvocabulary import wrap_vocabulary
 from DateTime import DateTime
 from DateTime.interfaces import DateTimeError
 from ftw.solr.converters import to_iso8601
 from ftw.solr.interfaces import ISolrSearch
 from ftw.solr.query import escape
 from opengever.base.behaviors.translated_title import ITranslatedTitleSupport
+from opengever.base.helpers import display_name
 from opengever.base.interfaces import ISearchSettings
 from opengever.base.solr import OGSolrDocument
 from opengever.base.utils import get_preferred_language_code
+from opengever.globalindex.browser.report import task_type_helper as task_type_value_helper
 from opengever.task.helper import task_type_helper
 from plone import api
 from plone.app.contentlisting.interfaces import IContentListingObject
@@ -19,9 +22,11 @@ from Products.CMFCore.utils import getToolByName
 from Products.ZCatalog.Lazy import LazyMap
 from Products.ZCTextIndex.ParseTree import ParseError
 from zope.component import getUtility
+from zope.component.hooks import getSite
+from zope.globalrequest import getRequest
+from zope.i18n import translate
 from ZPublisher.HTTPRequest import record
 import Missing
-
 
 DEFAULT_SORT_INDEX = 'modified'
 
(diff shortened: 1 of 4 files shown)


opengever/api/listing.py
```diff
@@ -340,24 +344,40 @@ def solr_results(self, name, term, columns, start, rows, sort_on,
...
             query=query, filters=filter_queries, start=start, rows=rows,
             sort=sort, **params)
-        return LazyMap(
-            OGSolrDocument,
-            start * [None] + resp.docs,
-            actual_result_count=resp.num_found,
-        )
+
+        # We map the index names back to the field names for the facets
+        facet_counts = {}
```
deiferni: i would add one more nesting level to this dict as with these changes we are now providing meta-information in the response, too. if i read correctly at the moment we would return name as key and count as value in the toplevel dict. i would prefer if facet counts are were all available under their own dict-key, e.g. `resp['facets']['fieldname_1']`. That way we can add more field meta-information to the response in the future without conflicts.

lukasgraf:
Necessary changes on the `ftw.solr` side have been released with `ftw.solr = 2.5.0`

njohner:
Ok so I added a comment to show where the nesting is done in the response.

Moreover, @all4manu asked me to also return the labels for the facets, as they do not want to display the tokens, but the labels. This avoids them having to send a second request to get the vocabularies or such. I've added this in a separate commit. Unfortunately I don't think we can write a test for that as long as we cannot properly test solr.

njohner:
Example response:
```
{
    "@id": "http://localhost:8080/fd/ordnungssystem/fuehrung/@listing?name=documents&facets%3Alist=review_state&facets%3Alist=sequence_number",
    "b_size": 25,
    "b_start": 0,
    "facets": {
        "review_state": {
            "document-state-draft": {
                "count": 3,
                "label": "nicht definiert"
            }
        },
        "sequence_number": {
            "2": {
                "count": 1
            },
            "33": {
                "count": 1
            },
            "35": {
                "count": 1
            }
        }
    },
    "items": [
        {
            "@id": "http://localhost:8080/fd/ordnungssystem/fuehrung/kommunikation/allgemeines/dossier-1/document-1"
        },
        {
            "@id": "http://localhost:8080/fd/ordnungssystem/fuehrung/gemeinderecht/dossier-14/document-35"
        },
        {
            "@id": "http://localhost:8080/fd/ordnungssystem/fuehrung/gemeinderecht/dossier-14/document-33"
        }
    ],
    "items_total": 3
}
```

PR #5767, opened 2019-07-01
Add linguistic index on task principal column for oracle backends.

phgross:
This fixes performance Issues on task listings when using an oracle backed for the OGDS DB (closes 4teamwork/opengever.zug#313). Because of the NLS_SORT and NLS_COMP settings we use for oracle backends (see https://github.com/4teamwork/opengever.core/blob/master/opengever/ogds/base/events.py#L6-L11), we need to make sure that oracles indexes supports linguistic sort.

Upgradestep: The PR provides a  schema migration which adds the index if it's not existing yet.

## Checkliste
  - [ ] Wurde alle Columns/Änderungen aus dem Modell in einer DB-Schema Migration nachgeführt. 
      _--> War leider nicht möglich. Stellt aber m.E. keine Problem dar._

### DIFF
--- a/opengever/core/upgrades/20190628142921_add_index_for_task_principal/upgrade.py
+++ b/opengever/core/upgrades/20190628142921_add_index_for_task_principal/upgrade.py
@@ -1,0 +1,21 @@
+from opengever.base.model import is_oracle
+from opengever.core.upgrade import SchemaMigration
+from opengever.globalindex.model.task import TaskPrincipal
+from sqlalchemy import func
+from sqlalchemy import Index
+
+
+INDEX_NAME = 'task_principals_ix'
+
+
+class AddIndexForTaskPrincipal(SchemaMigration):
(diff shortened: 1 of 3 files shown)


opengever/core/upgrades/20190628142921_add_index_for_task_principal/upgrade.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class AddIndexForTaskPrincipal(SchemaMigration):
+    """Add Index for task principal.
+    """
+
+    def migrate(self):
+        if is_oracle() and not self._has_index(INDEX_NAME, 'task_principals'):
+            task_principals_ix = Index(
+                'task_principals_ix',
```
deiferni: this should use the constant `INDEX_NAME `

phgross:
@deiferni updated.

PR #5769, opened 2019-07-02
Move notification configurations to own table

### LINKED ISSUE #5763: OGDS syncing fails probably because of user notification configurations
see https://sentry.4teamwork.ch/sentry/onegov-gever/issues/32581/events/605335/

Global notification configurations were added on the `User` model, which probably is a bad idea, as fields from the `User` model are mapped to the LDAP and updated when the OGDS is synced. As the new configuration fields are not in the LDAP schema, we get `None` as value for them (https://github.com/4teamwork/opengever.core/blob/master/opengever/ogds/base/sync/ogds_updater.py#L214), which is then not a valid value for these fields when we try to write into the OGDS.

Proposed solution: Move these configurations to a new table.

njohner:
Global notification configurations were added on the `User` model (in https://github.com/4teamwork/opengever.core/pull/5730). This has side-effects as the `User` is mapped to the LDAP schema and leads to failing `ogds_sync`. In hindsight the `User` is simply not the right place for user specific settings. We therefore migrate these notification configurations to their own table.

On the side of the implementation, I had first implemented it very similar to the notification settings, i.e. handling fallback to default value and adding a new row to the table if necessary in the `NotificationSettings` view. This leads to having to handle these things at several places and also in the tests. I finally decided that the `NotificationConfiguration` table should hold the methods to get the setting for a user (i.e. return the default when there is no entry for the user) and save a setting for a user (i.e. add a new row when necessary, update the existing row otherwise). I feel this is cleaner.

I added an Exception to allow easier and cleaner handling of the cases where we try to save configurations for a user that is not in the OGDS. The exception is handled in the `NotificationSettings` view by resetting to the default and showing an error message. I now also handle this for the settings (not only the configurations), which would raise an unhandled error before (when trying to save a setting as zopemaster for example).

closes #5763 

## Checkliste

- [x] Changelog-Eintrag vorhanden/nötig? CL nicht nötig, ist schon vorhanden im selbem release.
- [x] Aktualisierung Dokumentation vorhanden/nötig? Nicht nötig, Dokumentation wurde in https://github.com/4teamwork/opengever.core/pull/5730 ergänzt und ist auch nach dieser Änderung noch gültig.


opengever/activity/browser/settings.py
```diff
@@ -140,21 +146,23 @@ class NotificationSettings(BrowserView):
...
 
         config_name = self.request.form['config_name']
         value = json.loads(self.request.form['value'])
 
-        # Make sure that config_name is a column of the User model.
-        error_msg = "Notification configuration has to be a column on User"
-        assert config_name in User.__table__.columns, error_msg
+        # Make sure that config_name is a column of the NotificationConfiguration model.
+        error_msg = "Notification configuration has to be a column on NotificationConfiguration"
+        assert config_name in NotificationConfiguration.__table__.columns, error_msg
```
deiferni: this assertion and error message is duplicated, would it make sense to move it into `save_config_for_user`? It seems to be a precondition for that method.

opengever/activity/model/settings.py
```diff
@@ -75,3 +77,32 @@ class NotificationSetting(SettingMixin, Base):
...
     userid = Column(String(USER_ID_LENGTH), ForeignKey(User.userid))
     user = relationship(User, backref="settings")
+
+
+class NotificationConfiguration(Base):
+
+    __tablename__ = 'notification_configurations'
+
+    userid = Column(String(USER_ID_LENGTH), ForeignKey(User.userid), primary_key=True)
+    user = relationship(User, backref="configurations")
```
deiferni: We should discuss why this is not part of the `NotificationSettings` table (https://github.com/4teamwork/opengever.core/blob/db13d9ee0ab331c7d829a8ebaf10f839d6674e51/opengever/activity/model/settings.py#L67). I'm not sure if you already refer to that table in the PR description or just to the `NotificationSettings ` browser-view. At first glance it seems like it would make perfect sense to be treated like any other setting that is already stored in the `NotificationSettings` table. But maybe that is what you already tried 🤔? ☕️ ?!

deiferni: oh i just saw, this is a bit more complicated. i overlooked the attributes coming from `SettingMixin`. In that case i think i just don't like the name `NotificationConfiguration`. also quickly discussed it with @lukasgraf and we are of the opinion that `UserSettings` would be a good name, as we are actually now starting to introduce and store user-specific settings. what do you think?

njohner: For completeness, here are the original issues:
https://github.com/4teamwork/opengever.core/issues/5466
https://github.com/4teamwork/opengever.core/issues/5465

It was decided there to add these settings to the `User` model, as they would not fit the `NotificationSettings` model, which can contain several rows for each user. Of course we can change the name of the table...

njohner: If we name it `UserSettings`, should we also consider the table as a generic place to store user settings? In that case I would not put it in the `activity` module, but maybe in `base`?

njohner: After discussion, the `UserSettings` will go into `ogds`

opengever/activity/browser/settings.py
njohner: I'd thought about that and decided against it. I already thought I might be handling too much in the model itself. But I'm fine with moving the assertion there, so 👍

PR #5750, opened 2019-06-24
REST-API: Handle workspace invitations

elioschmutz:
Das GEVER-UI benötigt einen Endpoint zum Anzeigen von Teamraum-Einladungen sowie zum Annehmen und Ablehnen von Einladungen.

Dieser PR implementiert die entsprechenden Endpoints.

Issuer: https://github.com/4teamwork/gever-ui/issues/21
## Checkliste

_Zutreffendes soll angehakt stehengelassen werden._

- [x] Changelog-Eintrag vorhanden/nötig?
- [x] Aktualisierung Dokumentation vorhanden/nötig?

PR #5772, opened 2019-07-03
Remove unwanted disabling of csrf protection in @workspace-invitations endpoint

elioschmutz:
Dieser PR entfernt eine fälschlicherweise hinzugefügte Codezeile, welche den CSRF-Schutz für den @workspace-invitations Endpoint entfernt.

Ursprünglicher PR: https://github.com/4teamwork/opengever.core/pull/5750 

## Checkliste

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2019.3.1 (unreleased)
 ---------------------
 
+- Remove unwanted disabling of csrf protection in @workspace-invitations endpoint [elioschmutz]
 - @listing endpoint: Add support for filtering by relative path depth. [lgraf]
 - Add POST restapi endpint @mworkspace-invitations/{id}/{action}. [elioschmutz]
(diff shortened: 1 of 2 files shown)


lukasgraf:
@elioschmutz tests are failing

elioschmutz:
@lukasgraf ok, my fault, disabling the CSRF-protection is required.

For my understanding: Generally, protecting a REST-API against CSRF-attacks makes no sense, since the rest-endpoints are ignoring cookie-auth. Is that true? But why do we have to explicitly disable csrf-protection?

So this PR can be closed without merging.

lukasgraf:
@elioschmutz yes, for REST API endpoints can usually safely disable CSRF protection. Not because they ignore cookie-auth (they don't!), but because they require the `Accept: application/json` header, and it's pretty much impossible to force a browser into sending that in a CSRF-style attack.

So yeah, I would agree, this can be closed, and the REST API endpoint actually needs to be whitelisted, it can't work any other way.

elioschmutz:
Thanks, @lukasgraf for your explanation.

PR #5784, opened 2019-07-08
Fix upgrade step that adds linguistic index for task principal

lukasgraf:
The `self._has_index()` method that was used before is one that only exists on the `IdempotentOperations` class, which we don't currently use for Oracle. So this upgrade step (introduced in #5767) didn't quite work yet when executed on Oracle.

For this specific upgrade step, we therefore test for index presence directly using a simple SQL query.

_(Tested (by hand) on Oracle)_

## Checkliste

_(bestehende angepasst)_

### DIFF
--- a/opengever/core/upgrades/20190628142921_add_index_for_task_principal/upgrade.py
+++ b/opengever/core/upgrades/20190628142921_add_index_for_task_principal/upgrade.py
@@ -3,6 +3,7 @@ from opengever.core.upgrade import SchemaMigration
 from opengever.globalindex.model.task import TaskPrincipal
 from sqlalchemy import func
 from sqlalchemy import Index
+from sqlalchemy.sql.expression import text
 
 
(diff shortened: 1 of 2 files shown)


opengever/core/upgrades/20190628142921_add_index_for_task_principal/upgrade.py
```diff
@@ -12,8 +13,17 @@ class AddIndexForTaskPrincipal(SchemaMigration):
     """Add Index for task principal.
     """
 
+    def _has_index(self, idxname):
+        stmt = text("SELECT COUNT(*) FROM user_indexes WHERE index_name = :idxname")
```
deiferni: so using reflection here won't work?

lukasgraf: It might, but it's quote a bit more complicated.

PR #5836, opened 2019-07-24
Backend for todos

### LINKED ISSUE #5802: Backend work lightweight content type "ToDo"
Harvest: `[3.2.2.5 / 27] Projekt - OGIP 34 ToDo's` - `Entwicklung`

Todos: 
- [x] Dexterity FTI (one state Workflow) (#5808)
  - Schema: Title, Text (Plain Text mit Zeilenumbrüchen), Responsible (Person), deadline, done
- [x] Workspace Vocabularies (ActualWorkspaceMembers, PotentialWorkspaceMembers)*  Implementierung verbessern (#5836)
- [x] Rest API* (#5808)
- [x] SearchableText prüfen oder ergänzen. (#5808)
- [x] Indexer* (#5808)
- [x] Hard limit für maximal 500 Anzahl* (#5836)
- [x] ID: `todo-${global-gever-counter}` --> `todo-2383` (#5836)

njohner:
In this PR we:
* Add the standard naming and numbering behavior to `ToDo`s
* Add a hard limit of the number of `ToDo`s allowed in a single workspace
* Add two new ogds sources: `PotentialWorkspaceMembersSource` and `ActualWorkspaceMembersSource` and use them respectively for the users that can be invited into the workspace and users that can be set as responsibles on a `ToDo`
* Remove special handling of workspaces in `AllUsersInboxesAndTeamsSource`, which is not needed anymore as it was only used for tasks in workspaces, which are now not allowed anymore (see https://github.com/4teamwork/opengever.core/pull/5808).

I also improved some tests of other ogds sources, added tests for the `ToDo`'s API and a few other small fixes.

Overall the ogds sources seem more complex than necessary, have some limitations and are probably not always used correctly in our codebasis. They usually allow much more values to be used than the ones that get proposed (with the search query), because all users that might have been valid at some point need to remain valid. This does not limit correctly the users that can be set, only the ones proposed, and will need to be handled accordingly in the new frontend.

resolves #5802 

## Checkliste

- [ ] Aktualisierung Dokumentation vorhanden/nötig? We decided not to document the hard limit. The rest does not need any documentation imho.

### DIFF
--- a/opengever/ogds/base/sources.py
+++ b/opengever/ogds/base/sources.py
@@ -591,10 +591,18 @@ class PotentialWorkspaceMembersSourceBinder(object):
         return PotentialWorkspaceMembersSource(context)
 
 
-class ActualWorkspaceMembersSource(PotentialWorkspaceMembersSource):
+class ActualWorkspaceMembersSource(AssignedUsersSource):
     """Vocabulary of all users assigned to the current admin unit and
-    members of the current workspace
+    members of the current workspace.
+    The base query is not overwritten here, as this is used as source
+    for ToDo responsibles, which should remain valid even when a user's
+    permissions on a workspace are revoked (invitation deleted).
     """
+
+    @property
+    def search_query(self):
+        query = super(ActualWorkspaceMembersSource, self).search_query
+        return self._extend_query_with_workspace_filter(query)
 
     def _extend_query_with_workspace_filter(self, query):
         userids = list(get_workspace_user_ids(self.context))
--- a/opengever/ogds/base/sources.py
+++ b/opengever/ogds/base/sources.py
@@ -1,6 +1,7 @@
 from ftw.solr.interfaces import ISolrSearch
 from opengever.base.interfaces import ISearchSettings
 from opengever.base.model import create_session
+from opengever.base.query import extend_query_with_textfilter
 from opengever.contact.contact import IContact
 from opengever.contact.service import CONTACT_TYPE
 from opengever.ogds.base import _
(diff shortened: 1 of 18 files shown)


opengever/workspace/tests/test_todo.py
```diff
@@ -1,26 +1,87 @@
...
+        self.assertIsNotNone(todo)
+        self.assertIsNotNone(todo.responsible)
+        self.assertEqual(self.regular_user.id, todo.responsible)
+
+    @browsing
+    def test_reponsible_remains_valid_after_local_roles_removed(self, browser):
+        self.login(self.workspace_member, browser)
+
+        RoleAssignmentManager(self.workspace).add_or_update(
+            self.regular_user.id, ['WorkspaceGuest'], 5)
```
deiferni: i don't like the hardcoded `5` here, could you use the constant `ASSIGNMENT_VIA_INVITATION` instead?

opengever/workspace/tests/test_todo.py
```diff
@@ -1,26 +1,87 @@
...
+
+        todo = self.create_to_do(browser, self.workspace, u'Ein ToDo',
+                                 self.regular_user.id)
+
+        self.assertIsNotNone(todo)
+        self.assertEqual(self.regular_user.id, todo.responsible)
+        self.assertEqual([], getSchemaValidationErrors(IToDoSchema, todo))
+
+        RoleAssignmentManager(self.workspace).clear_by_cause_and_principal(
+            5, self.regular_user.id)
```
deiferni: same here use `ASSIGNMENT_VIA_INVITATION`

opengever/ogds/base/sources.py
```diff
@@ -489,7 +488,7 @@ def getTerm(self, value):
             user = self.base_query.filter(User.userid == value).one()
         except orm.exc.NoResultFound:
             raise LookupError(
```
deiferni: this seems to re-use pythons `LookupError`, is that desired?

njohner: This was not added in this PR. Now to give a better answer, `LookupError` is used all over the `ogds.base.sources`. This seems fine to me, as this is a lookup error, the term is not found in the vocabulary. Plone's sources also raise that error for the `getTerm` method, when the token is not found.

deiferni: alright 👍.

njohner:
Voila rebased and ready

Issue #5167, opened 2018-12-05
Aussonderung: Anbindung Austauschplattform


phgross:
_Harvest: `Projekt - GEVER SG, Optimierungen aus Pilotprojekt, Release 1.1` - `Entwicklung`_

Die Speicherung des SIP Packages als blob wurde bereits im Branch `pg_disposition_store` umgesetzt. 

Tasks:
- [x] `pg_disposition_store` PR erstellen, reviewen
- [x] add initial content for local development
- [x] FTP Test Share von Betrieb zur Verfügung stellen lassen (https://extranet.4teamwork.ch/betrieb/tracker/889)
- [x] Status Transport in Angebot führen und darstellen
- [x] Delivery via filesystem abschliessen (#5850)
- [x] Have OPS create a FTPS test-setup on our `dev` (see https://extranet.4teamwork.ch/betrieb/tracker/891/view):
  - [x] Install `vsftpd`, and [configure it in implcit SSL mode](https://serverfault.com/questions/65002/vsftpd-and-implicit-ssl) (`implicit_ssl=YES`)
  - [x] Set up an appropriate user
  - [x] ~Mount an FTPS connection to that server as a share on `dev` using [CurlFtpFS](http://curlftpfs.sourceforge.net/)~
- [x] ~Have OPS mount the FTPS connection to the _real_ share on the 805 test server~

phgross:
Kleines Konzept: https://gever.4teamwork.ch/ordnungssystem/3/2/produkt-onegov-gever-aussonderung/dossier-3324/document-29733?SearchableText=austauschplattform#overview

@deiferni hast du noch Einwände oder Vorschläge?

deiferni:
Kleinere Korrekturen/Anmerkungen auch direkt im Word im Änderungsmodus. Der wichtigste Punkt:  ich würde die SIP-Dateien nicht direkt und selber auf dem Filesystem speichern, sondern analog zu https://github.com/4teamwork/opengever.core/pull/4954 als blob.

deiferni:
the share should be available as per https://extranet.4teamwork.ch/betrieb/tracker/889

deiferni:
Unassigning `self`. Current state:
- https://github.com/4teamwork/opengever.core/pull/5837 might need to be merged/fixed 
- https://github.com/4teamwork/opengever.core/pull/5840 will help with local development
Have not started with other tasks, so any of those can be taken over.

lukasgraf:
[...]

PR #5837, opened 2019-07-24
 Rework SIP Package generation and download

phgross:
This is the first part of the work we do for #5167, a short summary of the changes this PR does.
- **Rework SIP Package generation and download.** Store SIP package as a blob locally instead of re generating them for each download. This solves time and ressources and makes it possible to provide customer specific scripts and SIP package handling. 
- **Switch SIP download button to the download view.** The export view stays still available, but is no longer available trough the UI, can be useful for debugging purposes.
- Instead of a manual trigger, generate and store the SIP Package automatically, when disposing a disposition.
- Remove local copy of the SIP package after closing a disposition.

## Checkliste

- [ ] Aktualisierung Dokumentation vorhanden/nötig? _Nicht notwendig, das sich für den Benutzer nichts ändert. Es handelt sich hier ausschliesslich um technische Anpassungen._

### DIFF
--- a/opengever/disposition/browser/ech0160.py
+++ b/opengever/disposition/browser/ech0160.py
@@ -60,7 +60,7 @@ class ECH0160DownloadView(BrowserView):
     def __call__(self):
         if not self.context.has_sip_package():
             msg = _('msg_no_sip_package_generated',
-                    default=u'No SIP Package generated for this disposition')
+                    default=u'No SIP Package generated for this disposition.')
             api.portal.show_message(msg, request=self.request, type='error')
             return self.request.RESPONSE.redirect(self.context.absolute_url())
 
--- a/opengever/disposition/browser/ech0160.py
+++ b/opengever/disposition/browser/ech0160.py
@@ -1,5 +1,8 @@
 from opengever.base.stream import TempfileStreamIterator
+from opengever.disposition import _
 from opengever.disposition.ech0160.sippackage import SIPPackage
+from plone import api
+from plone.namedfile.utils import stream_data
 from Products.Five import BrowserView
 from pyxb.utils.domutils import BindingDOMSupport
 from tempfile import TemporaryFile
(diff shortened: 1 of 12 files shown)


opengever/disposition/browser/ech0160.py
```diff
@@ -34,3 +37,38 @@ def create_zipfile(self, package):
...
+            api.portal.show_message(msg, request=self.request, type='error')
+            return self.request.RESPONSE.redirect(self.context.absolute_url())
+
+        sip_package = self.context.get_sip_package()
+        response = self.request.response
+        response.setHeader(
+            "Content-Disposition",
+            'inline; filename="%s.zip"' % self.context.get_sip_name())
+        response.setHeader("Content-type", "application/zip")
+        response.setHeader("Content-Length", sip_package.getSize())
```
deiferni: could we use `plone.namedfile.utils.set_headers` here?

opengever/disposition/disposition.py
```diff
@@ -293,3 +302,35 @@ def get_all_archivists(self):
...
         return archivists
+
+    def store_sip_package(self):
+        self._sip_package = self.generate_sip_package()
+
+    def generate_sip_package(self):
+        package = SIPPackage(self)
+        zip_file = self.create_zipfile(package)
+        zip_file.seek(0)
+        return NamedBlobFile(zip_file.read(), contentType='application/zip')
```
deiferni: I think you could pass a reference to the file instead of reading the file into memory yourself. That triggers an `plone.namedfile.interfaces.IStorage` adapter, which should be able to handle the file instance. If we are lucky it can be moved on the filesystem without reading the whole thing into memory. Not 100% sure but worth a try.

opengever/disposition/browser/ech0160.py
```diff
@@ -34,3 +37,38 @@ def create_zipfile(self, package):
...
 
         return tmpfile
+
+
+class ECH0160StoreView(BrowserView):
+    """A view which generates and store's the SIP package as a blob file.
+    """
+
+    def __call__(self):
+        self.context.store_sip_package()
```
deiferni: This view will respond to a `GET` request. do you see a way to create a form which only responds to an i.e. `POST`, as it will modify data in the database.

opengever/disposition/browser/overview.py
```diff
@@ -88,6 +88,13 @@ def get_actions(self):
         ]
 
     def sip_download_available(self):
+        if api.user.has_permission(
+            'opengever.disposition: Download SIP Package',
+            obj=self.context):
+
+            return self.context.has_sip_package()
+
```
deiferni: `return None` to have it explicit that we do something with the return value?

opengever/disposition/locales/de/LC_MESSAGES/opengever.disposition.po
```diff
@@ -273,6 +273,16 @@ msgstr "Angebot abgelehnt durch ${user}"
 msgid "msg_disposition_updated"
 msgstr "Aktualisiert durch ${user}"
 
+#. Default: "No SIP Package generated for this disposition"
+#: ./opengever/disposition/browser/ech0160.py
+msgid "msg_no_sip_package_generated"
+msgstr "Es wurde noch kein SIP Paket generiert für dieses Angebot"
```
deiferni: potentially a missing dot at the end of this sentence 🤔?

opengever/disposition/disposition.py
phgross: I current did not found a way to pass in a file in to a NamedBlobFile, we would need to register a IStorage Adapter for `__builtin__.file` what I'm not sure is the right way. For now i did no change.

opengever/disposition/browser/ech0160.py
phgross: I'll do the maintenance-actions as discussed in a separate PR, I added a checkbox to the Main Issue.

phgross:
@deiferni updated according to your feedback. The maintenance-action I'll do in a separate PR.

opengever/disposition/disposition.py
deiferni: as discussed today, if it does not work by default i'm ok with not changing the current implementation 👍.

opengever/disposition/browser/ech0160.py
deiferni: i have added a checkbox to the main issue.

PR #5840, opened 2019-07-25
Add a disposable dossier to the initial content.

deiferni:
Part of https://github.com/4teamwork/opengever.core/issues/5167.

This quality of live PR adds initial content for disposition so that we can easily create disposable content for local development. It creates a dossier with one document which has its retention period expired and which is already in closed state. It should pop up when you add a new disposition on your repository.

_I have opted to not create the disposition itself yet, however that could be amended in a later PR if deemed useful._


deiferni:
_as we discussed this today directly assigning @lukasgraf_

PR #5844, opened 2019-07-30
Fix team actor profile_url for foreign users

### LINKED ISSUE #5843: Team actor profile_url raises Exception for foreign user
https://sentry.4teamwork.ch/sentry/onegov-gever/issues/35561/events/latest/

Steps to reproduce:
 - Create task and assign it to a foreign user
 - Try to reassign the task to a team as the foreign user

phgross:
The profile_url fetches the contact folder which is not possible for a foreign user (is not permitted to see the contactfolder). Therefore I introduced a `elevate_privileges` keyword to get the contact_folder url even if the user is not allowed to access it.

In my opinion it's the right solution and part of an access as a foreign user, that some links are not accessible and answer with an unauthorized message.

Fixes #5843 

In addition to the actual change, I moved the `is_solr_feature_enabled` flag where it belongs to.

## Checkliste
- [x] Wurde etwas an der `Aufgabe` angepasst? Funktioniert das auch mit einer `Weiterleitung`? _Ja, eine Weiterleitung kann nicht direkt neu zugewiesen werden, sondern muss zuerst akzeptiert werden, demnach ist diese Situation so nicht möglich. _

### DIFF
--- a/opengever/contact/utils.py
+++ b/opengever/contact/utils.py
@@ -1,3 +1,4 @@
+from opengever.base.security import elevated_privileges
 from opengever.base.utils import get_hostname
 from opengever.contact.interfaces import IContactFolder
 from plone import api
@@ -12,7 +13,7 @@ def hostname_cache_key(m, *args, **kwargs):
 @ram.cache(hostname_cache_key)
 def get_contactfolder_url(elevate_privileges=False):
     if elevate_privileges:
-        from opengever.base.security import elevated_privileges
+
         with elevated_privileges():
             result = api.content.find(object_provides=IContactFolder)
     else:
--- a/opengever/contact/utils.py
+++ b/opengever/contact/utils.py
@@ -5,13 +5,19 @@ from plone.memoize import ram
 from zope.globalrequest import getRequest
 
 
-def hostname_cache_key(m):
+def hostname_cache_key(m, *args, **kwargs):
     return get_hostname(getRequest())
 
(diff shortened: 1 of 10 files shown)


opengever/contact/utils.py
```diff
@@ -5,13 +5,19 @@
...
+def hostname_cache_key(m, *args, **kwargs):
     return get_hostname(getRequest())
 
 
 @ram.cache(hostname_cache_key)
-def get_contactfolder_url():
-    result = api.content.find(object_provides=IContactFolder)
+def get_contactfolder_url(elevate_privileges=False):
+    if elevate_privileges:
+        from opengever.base.security import elevated_privileges
```
lukasgraf: Is this inline import needed to avoid circular dependencies?

opengever/contact/utils.py
```diff
@@ -5,13 +5,19 @@
...
+        with elevated_privileges():
```
lukasgraf: I'm wondering if it wouldn't be cleaner to just use `catalog.unrestrictedSearchResults()` vs. a regular catalog query in these cases - then we wouldn't even need elevated privileges at all. All we need is an unrestricted catalog search, right? (That would then mean renaming the kwarg to `unrestricted`)

phgross: @lukasgraf makes sense and works as expected. Added an additional commit ...

PR #5845, opened 2019-07-30
Nightly Jobs: Minor Quality of Life improvements

lukasgraf:
Two small Quality of Life improvements for nightly jobs:

- Don't require `ftw.raven` when running nightly jobs locally.
- Add short `-f` option as alias for `--force`.

#5302  Checkliste

### DIFF
--- a/opengever/nightlyjobs/cronjobs.py
+++ b/opengever/nightlyjobs/cronjobs.py
@@ -1,5 +1,6 @@
 from logging.handlers import TimedRotatingFileHandler
 from opengever.base.pathfinder import PathFinder
+from opengever.base.sentry import maybe_report_exception
 from opengever.core.debughelpers import all_plone_sites
 from opengever.core.debughelpers import setup_plone
 from opengever.nightlyjobs.runner import nightly_jobs_feature_enabled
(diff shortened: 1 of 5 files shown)


lukasgraf:
QA tests are failing because it seems our custom config is missing that bumps the line length limit from 79 to 120.

opengever/nightlyjobs/cronjobs.py
```diff
@@ -14,6 +13,14 @@
 import traceback
 
 
+try:
+    from ftw.raven.reporter import maybe_report_exception
```
deiferni: should use `pkg_resources.get_distribution('ftw.raven')` and catch `DistributionNotFound` to avoid shadowing other isses that cause an import error, e.g in the unlikely event the `maybe_report_exception` function would be moved.

deiferni: In two other modules we work with a `HAS_RAVEN` flag, this would be the third time we need it. As you are already dealing with quality of life, it could be the time to create a `compat.py` somewhere that provides these compatibility imports.

lukasgraf:
QA test config generation fixed in 4teamwork/ftw-buildouts#159 - once that one is merged, I'll rerun tests and they should pass.

PR #5847, opened 2019-08-02
Expose document actions in @actions endpoint in separate `file_actions` category.

### LINKED ISSUE #5786: Dokumentaktionen sollten als portal_actions zugänglich sein
_Harvest:  `Projekt - Umsetzung Neues User Interface GEVER 2019/2020` - `Entwicklung`_

Die Dokument Aktionen, also die Actions welche sowohl im Tooltip, Overlay und der Dokumentansicht dargestellt werden, (Kopie herunterladen, auschecken und bearbeiten, einchecken etc. ) sollten als im @actions endpoint zur Verfügung stehen und so auch via API zugänglich sein. Dabei sollen auch die Prüfungen, ob eine Aktion möglich ist, durchgeführt werden.

Erster Vorschlag: Aufnahme der Aktionen in die `portal_actions`. Um die Ansichten im klassischen UI nicht zu stören, wäre ein separate actions category sinnvoll, da dies keine Auswirkungen für das bestehende UI hat, da diese gar nicht erst geprüft und auch nicht gerendert werden.

Mögliche Alternative: Endpoint überschreiben.

An den bestehenden Konditionen sollen keine Änderungen gemacht werden.

deiferni:
This PR exposes  document/mail actions in the `@actions` endpoint. It adds a new actions category `file_actions`.

We attempt to address the concerns of **availability** and **visibility** separately.
The logic for the availability of document actions are placed in adapters implementing `IFileActions`. Most of the logic was available in `ActionButtonRendererMixin` or in one of the bumblebee things and is moved into the new classes to clearly split up availability from rendering. The `@actions` endpoint returns all **available** actions. 

**Visibility** is handled by the view subclassing `ActionButtonRendererMixin` and is considered a separate additional layer on top of availability. This means that some actions will not be shown by the view even though they would be available.

Some general remarks:
- As the actions currently only answer availabilty and we do not use the additional data they provide (title, target url) we currently do not fetch the actions while rendering the macro but instead directly rely on `FileActionAvailabilityMixin`.
- The template in https://github.com/4teamwork/opengever.core/blob/b6cb3cda15cc2ee4bf9609f9b7dc330873b619f6/opengever/base/browser/templates/macros.pt contained a lot of nested logic. This has been moved to `ActionButtonRendererMixin` and `IFileActions`. The action rendering macro is now almost flat, this should improve readability.
- I have discovered some more actions in `IBumblebeeOverlay` and `BumblebeeOverlayBaseView` and also included them in `@actions`. I had to remove some duplication and dead code (which died with https://github.com/4teamwork/opengever.core/pull/2744). Most of the rendering now happens in `VisibleActionButtonRendererMixin` and the subclasses only override certain aspects, if necessary.
- The actions endpoint does not return the url of the action, which is kind of stupid. This means that actions that have an id that is different from their url, cannot be used as returned by the actions endpoint
- We have several actions that cannot be called simply with an url. For those at least, the frontend will have to know what to do when the action is available.
- In the current frontend, we display different titles for the checkout actions depending on whether the document is already checked out. This is be handled by registering a different action for each case.

_Despite the good test coverage i have attempted to also manually verify the actions are still rendered correctly with all possible permutations. The permutation matrix is quite high as we render the actions in various places (listing tooltip, listing overlay, overview, versions, ...) and there are states (checked out, locked, closed dossier) which have an effect on the actions, too._

Closes #5786.

⚠️ Issues discovered but not addressed with this PR. Might be something to discuss in our review:
- Currently we use the request to decide whether we are on a versioned context. I would prefer if we could somehow annotate the documents-versions loaded from a repository for the time of one request. Maybe the `Versioner` in https://github.com/4teamwork/opengever.core/blob/b6cb3cda15cc2ee4bf9609f9b7dc330873b619f6/opengever/document/versioner.py could set a volatile attribute on a document after it is retrieved from the repository?
- The revert to version action was always shown, even when not possible to execute. I have changed the condition to make the action unavailable when it can't be performed. There is currently no placeholder in such cases and the action in the overlay will disappear.

## Checkliste

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2019.4.0 (unreleased)
 ---------------------
 
+- Expose document actions in @actions endpoint in separate file_actions category. [njohner, deiferni]
 - Nightly jobs: Add short -f option as alias for --force. [lgraf]
 - Nightly jobs: Don't require ftw.raven when running locally [lgraf]
 - Fix team actor profile_url for foreign users. [phgross]
--- a/opengever/api/tests/test_actions.py
+++ b/opengever/api/tests/test_actions.py
@@ -6,6 +6,7 @@ from zope.component import getMultiAdapter
 
 class TestFileActionsGet(IntegrationTestCase):
 
+    features = ('bumblebee',)
     maxDiff = None
 
     def get_file_actions(self, browser, context):
@@ -26,6 +27,9 @@ class TestFileActionsGet(IntegrationTestCase):
             {u'id': u'attach_to_email',
              u'title': u'Attach to email',
              u'icon': u''},
+            {u'id': u'open_as_pdf',
+             u'title': u'Open as PDF',
+             u'icon': u''},
             ]
         self.assertEqual(expected_file_actions,
                          self.get_file_actions(browser, self.document))
@@ -43,6 +47,9 @@ class TestFileActionsGet(IntegrationTestCase):
              u'icon': u''},
             {u'id': u'attach_to_email',
              u'title': u'Attach to email',
+             u'icon': u''},
+            {u'id': u'open_as_pdf',
+             u'title': u'Open as PDF',
              u'icon': u''},
             ]
         self.assertEqual(expected_file_actions,
@@ -92,6 +99,9 @@ class TestFileActionsGet(IntegrationTestCase):
             {u'id': u'attach_to_email',
              u'title': u'Attach to email',
              u'icon': u''},
+            {u'id': u'open_as_pdf',
+             u'title': u'Open as PDF',
+             u'icon': u''},
             ]
 
         self.assertEqual(expected_file_actions,
@@ -123,6 +133,9 @@ class TestFileActionsGet(IntegrationTestCase):
             {u'id': u'attach_to_email',
              u'title': u'Attach to email',
              u'icon': u''},
+            {u'id': u'open_as_pdf',
+             u'title': u'Open as PDF',
+             u'icon': u''},
             ]
 
         self.assertEqual(expected_file_actions,
@@ -146,12 +159,18 @@ class TestFileActionsGet(IntegrationTestCase):
             {u'id': u'cancel_checkout',
              u'title': u'Cancel checkout',
              u'icon': u''},
+            {u'id': u'open_as_pdf',
+             u'title': u'Open as PDF',
+             u'icon': u''},
             ]
         self.assertEqual(expected_manager_file_actions,
                          self.get_file_actions(browser, self.document))
 
         self.login(self.dossier_manager, browser)
         expected_dossier_manager_file_actions = [
+            {u'id': u'open_as_pdf',
+             u'title': u'Open as PDF',
+             u'icon': u''},
             ]
         self.assertEqual(expected_dossier_manager_file_actions,
                          self.get_file_actions(browser, self.document))
@@ -165,6 +184,9 @@ class TestFileActionsGet(IntegrationTestCase):
              u'icon': u''},
             {u'id': u'attach_to_email',
              u'title': u'Attach to email',
+             u'icon': u''},
+            {u'id': u'open_as_pdf',
+             u'title': u'Open as PDF',
              u'icon': u''},
             ]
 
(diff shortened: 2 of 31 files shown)


opengever/document/browser/actionbuttons.py
```diff
@@ -60,6 +63,15 @@ def is_oc_zem_checkout_action_available(self):
                 self.is_office_connector_editable() and
                 not is_officeconnector_checkout_feature_enabled())
 
+    def is_checkout_action_available(self):
+        """Check whether simple checkout action is available
+        """
+        return (self.is_document() and
+                self.has_file() and
+                self.is_checkout_and_edit_available() and
+                not self.is_office_connector_editable() and
```
phgross: Thats a change in the functionality imho. The Action `auschecken` should always be visible for documents with a file, even if the document would support editing via office connector. Otherwise replacing a document by drag n drop would no longer be supported.

docs/HISTORY.txt
```diff
@@ -10,6 +10,7 @@ Changelog
 - Include actor_id and actor_label in @notifications endpoint responses. [lgraf]
 - Bump docxcompose version to 1.0.1 [njohner]
 - Improve SIP package generation and download. [phgross]
+- Bump docxcompose version to 1.0.1 [njohner]
```

opengever/api/tests/test_actions.py
```diff
@@ -57,12 +57,39 @@ def test_checkin_without_comment_available_if_checked_out_by_current_user(self,
...
+            {u'title': u'Checkin with comment', u'id': u'checkin_with_comment', u'icon': u''}
             ]
 
         listed_file_actions = browser.json['file_actions']
         self.assertEqual(expected_file_actions, listed_file_actions)
 
+    @browsing
+    def test_checkin_only_available_for_managers_if_checked_out_by_other_user(self, browser):
+        self.login(self.regular_user, browser)
+        manager = getMultiAdapter((self.document, self.request),
```
phgross: `self.checkout_document(self.document)`

😉

njohner:
@deiferni I'm taking that back over so we can ship it today

opengever/document/interfaces.py
```diff
@@ -190,3 +190,6 @@ def is_oneoffixx_retry_action_available():
 
     def is_open_as_pdf_action_available():
         """Return whether the open aos pdf action is available."""
+
+    def is_revert_to_version_action_available():
+        """Return wheter the revert to version action is available."""
```
njohner: typo: whether

opengever/document/browser/actionbuttons.py
njohner: Ok so it is indeed exposed in the actions menu:
https://github.com/4teamwork/opengever.core/blob/master/opengever/core/profiles/default/types/opengever.document.document.xml#L132-L141

So I'll change this

njohner: @deiferni has reworked this to make one action for the discreet or normal button for the simple checkout, `oc_unsupported_file_checkout` using the availability condition `is_oc_unsupported_file_checkout_action_available` corresponds to the checkout action above. I think it's correct that this does not always show up in the `actions` endpoint, but only when it was displayed as a document action in our frontend. There is also a `checkout` action on the document, which is "always" present (this is the one showing up in the actions menu) and will be returned by the actions endpoint too (but under `object_buttons` and not `file_actions`). 
https://github.com/4teamwork/opengever.core/blob/master/opengever/core/profiles/default/types/opengever.document.document.xml#L132-L141

opengever/document/browser/actionbuttons.py
```diff
@@ -1,73 +1,157 @@
...
 
-class ActionButtonRendererMixin(object):
-    """Mixin for views that render action buttons."""
+class FileActionAvailabilityMixin(object):
+    """Mixin that delegates availability checks to an IFileActions adapter.
 
-    is_overview_tab = False
-    is_on_detail_view = False
-    overlay = None
+    Returns if an action is available. An action which is not available cannot
```
njohner: "if" should be "whether" here, to make it clearer

njohner:
Ok this is ready IMHO. Integrated or commented on all the requested changes and made a few small corrections (typos in docstrings and commit messages).

opengever/document/fileactions.py
```diff
@@ -0,0 +1,198 @@
...
+        return self.context.is_oneoffixx_creatable()
+
+    def is_revert_to_version_action_available(self):
+        manager = getMultiAdapter(
+            (self.context, self.request), ICheckinCheckoutManager)
+
+        return (self.is_versioned()
+                and self.context.has_file()
+                and not self.context.is_checked_out()
+                and manager.is_checkout_allowed())
```
njohner: This is a bit stricter than what we had previously. We were not checking that the context was not checked out nor that checkout was allowed. But it makes sense

PR #5849, opened 2019-08-05
Testing: Add disposition with SIP to fixture.

lukasgraf:
This adds a disposition to the fixture which is in state `disposition-state-disposed` and has a SIP package, ready to be delivered.

This is necessary because transports (which will be implemented as part of #5167) need to access the underlying blob on the filesystem. The blob will only have a stable filesystem location once it's committed though. Because in our `IntegrationTestCase` we only emulate `transaction.commit()`, but never actually do a hard commit, this requires the blob to have already been created as part of the fixture to be actually committed and accessible. 

## Checkliste

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -572,6 +572,7 @@ Objects
         - self.closed_meeting_dossier
         - self.decided_meeting_dossier
         - self.disposition
+        - self.disposition_with_sip
         - self.dossier
           - self.document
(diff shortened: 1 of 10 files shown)


njohner:
Ok never mind my comment, I guess I did not properly read your PR comment... it's all clear why we need this!

PR #5850, opened 2019-08-05
Implement SIP delivery via filesystem

lukasgraf:
Implements SIP delivery via filesystem.

This addresses part of #5167 _(addressing the FTPS usecase will be solved by mounting an FTPS connection using CurlFtpFS, and using filesystem delivery to drop the SIP on that mounted share)_.

### Implementation

- A new `ISIPTransport` adapter is introduced. These adapters are meant to implement specific transport mechanisms.
- For now, a `FilesystemTransport` is provided (disabled by default), which copies the SIP to a configured destination directory
- Upon disposing a disposition, a new class `DeliveryScheduler` will schedule the SIP for delivery with all enabled transports.
- Delivery status is tracked (by the `DeliveryScheduler` per transport, in annotations on the disposition
- A nightly job is provided that collects all dispositions that are eligble for transport, and invokes delivery via enabled transports

---

The delivery status will be shown on the disposition overview. If only one transport is enabled, it is displayed in a compact form (without the transport name):

delivery_status

---

If more than one transport is configured, the delivery status will be listed per transport:

delivery_status_multiple

---

- [x] Polish delivery status table on Disposition overview
  - [x] Hide transport name if only 1 enabled transport
  - [x] Translate `status_success` etc into something human readable
- [x] Status `abgeliefert` anpassen z.B. zu`in Ablieferung`

~- [ ] Add "redeliver" button (backend functionality already exists, `DeliveryScheduler.schedule_delivery(force=True)`~

_(Cancelled because the required JS work to do this properly is in no relation to the value for a "reschedule" button, that probably won't do anything useful without our intervention anyway)_

## Checkliste

### DIFF
--- a/opengever/disposition/delivery.py
+++ b/opengever/disposition/delivery.py
@@ -1,11 +1,24 @@
 from opengever.base.sentry import maybe_report_exception
+from opengever.disposition.interfaces import IDisposition
+from opengever.disposition.interfaces import IFilesystemTransportSettings
 from opengever.disposition.interfaces import ISIPTransport
+from os.path import abspath
+from os.path import expanduser
+from os.path import join as pjoin
 from persistent.mapping import PersistentMapping
+from plone.registry.interfaces import IRegistry
 from ZODB.POSException import ConflictError
 from zope.annotation import IAnnotations
+from zope.component import adapter
 from zope.component import getAdapters
+from zope.component import getUtility
 from zope.globalrequest import getRequest
+from zope.interface import implementer
+from zope.publisher.interfaces.browser import IBrowserRequest
 import logging
+import os
+import shutil
+import stat
 import sys
 
 
@@ -137,3 +150,48 @@ class BaseTransport(object):
         self.log = logging.getLogger(self.__class__.__name__)
         self.log.setLevel(logging.INFO)
         self.log.parent = parent_logger
+
+
+@implementer(ISIPTransport)
+@adapter(IDisposition, IBrowserRequest, logging.Logger)
+class FilesystemTransport(BaseTransport):
+    """Transport that copies the SIP to a filesystem location specified by
+    the IFilesystemTransportSettings.destination_directory registry entry.
+    """
+
+    def deliver(self):
+        """Delivers the SIP by copying it to the destination_directory.
+        """
+        destination_dir = self._get_destination_directory()
+        assert os.path.isdir(destination_dir)
+        sip = self.disposition.get_sip_package()
+
+        blob_path = sip._blob.committed()
+        filename = self.disposition.get_sip_filename()
+        destination_path = pjoin(destination_dir, filename)
+
+        if os.path.isfile(destination_path):
+            self.log.warn("Overwriting existing file %s" % destination_path)
+
+        shutil.copy2(blob_path, destination_path)
+
+        # Make delivered file writable for owner
+        st = os.stat(destination_path)
+        os.chmod(destination_path, st.st_mode | stat.S_IWUSR)
+
+        self.log.info("Transported %r to %r" % (filename, destination_path))
+
+    def is_enabled(self):
+        settings = self._get_settings()
+        return settings.enabled
+
+    def _get_destination_directory(self):
+        settings = self._get_settings()
+        destination_dir = settings.destination_directory
+        destination_dir = abspath(expanduser(destination_dir.strip()))
+        return destination_dir
+
+    def _get_settings(self):
+        registry = getUtility(IRegistry)
+        settings = registry.forInterface(IFilesystemTransportSettings)
+        return settings
--- a/opengever/disposition/delivery.py
+++ b/opengever/disposition/delivery.py
@@ -1,4 +1,5 @@
 from opengever.base.sentry import maybe_report_exception
+from opengever.disposition import _
 from opengever.disposition.interfaces import IDisposition
 from opengever.disposition.interfaces import IFilesystemTransportSettings
 from opengever.disposition.interfaces import ISIPTransport
@@ -27,6 +28,12 @@ TRANSPORT_STATUSES_KEY = 'opengever.disposition.delivery.transport_statuses'
 STATUS_SCHEDULED = 'status_scheduled'
 STATUS_SUCCESS = 'status_success'
 STATUS_FAILED = 'status_failed'
+
+DELIVERY_STATUS_LABELS = {
+    STATUS_SCHEDULED: _(u'label_delivery_status_scheduled', default=u'Scheduled for delivery'),
+    STATUS_SUCCESS: _(u'label_delivery_status_success', default=u'Delivered successfully'),
+    STATUS_FAILED: _(u'label_delivery_status_failed', default=u'Delivery failed'),
+}
 
 
 class DeliveryScheduler(object):
--- a/opengever/disposition/delivery.py
+++ b/opengever/disposition/delivery.py
@@ -1,0 +1,139 @@
+from opengever.base.sentry import maybe_report_exception
+from opengever.disposition.interfaces import ISIPTransport
+from persistent.mapping import PersistentMapping
+from ZODB.POSException import ConflictError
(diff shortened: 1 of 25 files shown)


lukasgraf:
@njohner could you be so kind and provide French translations for the couple message strings I touched? _(Just dump them here, I'll work them in and rebase)_

- `In Ablieferung`
- `SIP Dateiname`
- `Ablieferungsstatus`
- `Zustellung fehlgeschlagen`
- `Zustellung geplant`
- `Erfolgreich zugestellt`

njohner:
- `In Ablieferung` -> `En cours de livraison`
- `SIP Dateiname` -> `Nom du fichier SIP`
- `Ablieferungsstatus` -> `Etat de la livraison`
- `Zustellung fehlgeschlagen` -> `Distribution échouée`
- `Zustellung geplant` -> `Distribution planifiée`
- `Erfolgreich zugestellt` -> `Distribué avec succès`

opengever/disposition/tests/test_delivery.py
```diff
@@ -117,3 +124,92 @@ def test_catches_transport_exceptions_and_sets_failed_status(self):
...
+
+        self.assertEqual([
+            u"Delivering using transport 'filesystem'",
+            u"Transported %r to %r" % (self.sip_filename, self.dst_path),
+            u"Successful delivery using transport 'filesystem'"],
+            self.captured_log.pop_msgs())
+
+        self.assertEqual([self.sip_filename], os.listdir(self.tempdir))
+
+        delivery_path = os.path.join(self.tempdir, self.sip_filename)
```
njohner: you can use `self.dst_path`

opengever/disposition/tests/test_delivery.py
```diff
@@ -117,3 +124,92 @@ def test_catches_transport_exceptions_and_sets_failed_status(self):
...
+        scheduler = DeliveryScheduler(self.disposition_with_sip)
+        scheduler.deliver()
+
+        self.assertEqual([
+            u"Delivering using transport 'filesystem'",
+            u"Transported %r to %r" % (self.sip_filename, self.dst_path),
+            u"Successful delivery using transport 'filesystem'"],
+            self.captured_log.pop_msgs())
+
+        scheduler.deliver()
```
njohner: You could also delete the delivered file, and check that when you redeliver, the file is not there

opengever/disposition/tests/test_disposition.py
```diff
@@ -262,3 +266,51 @@ def test_sip_package_is_removed_on_close(self, browser):
...
+        statuses = scheduler.get_statuses()
+
+        self.assertFalse(scheduler.is_scheduled_for_delivery())
+        self.assertEqual({}, statuses)
+
+        browser.open(self.disposition, view='overview')
+        browser.click_on('disposition-transition-dispose')
+
+        self.assertTrue(scheduler.is_scheduled_for_delivery())
+        self.assertEqual({}, statuses)
```
njohner: I think that if you do `scheduler.get_statuses()` here, you will not have an empty dictionary.

opengever/disposition/delivery.py
```diff
@@ -66,20 +66,25 @@ def schedule_delivery(self, force=False):
...
-            # Only schedule if no delivery was attempted yet
-            # (Except when forcefully rescheduling)
-            if force or current_status is None:
-                self.logger.info("Scheduling delivery for transport: '%s'" % name)
-                statuses[name] = STATUS_SCHEDULED
+        # Only schedule if no delivery was attempted yet
+        # (Except when forcefully rescheduling)
+        if force or current_status is None:
+            self.logger.info("Scheduling delivery for transport: '%s'" % name)
+            statuses[name] = STATUS_SCHEDULED
```
njohner: make a `mark_scheduled` method like for the other statuses?

opengever/disposition/delivery.py
```diff
@@ -0,0 +1,139 @@
...
+                e_type, e_value, tb = sys.exc_info()
+                maybe_report_exception(self.disposition, getRequest(), e_type, e_value, tb)
+
+    def get_statuses(self, create_if_missing=False):
+        """Get the mapping of delivery statuses by transport.
+        """
+        ann = IAnnotations(self.disposition)
+        if TRANSPORT_STATUSES_KEY not in ann and create_if_missing:
+            ann[TRANSPORT_STATUSES_KEY] = PersistentMapping()
+        return ann.get(TRANSPORT_STATUSES_KEY, {})
```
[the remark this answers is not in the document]
njohner: After discussion, this is to avoid write on read, which makes sense...

opengever/disposition/testing.py
```diff
@@ -0,0 +1,49 @@
...
+from opengever.testing.helpers import CapturingLogHandler
+from zope.component import adapter
+from zope.interface import implementer
+from zope.publisher.interfaces.browser import IBrowserRequest
+import logging
+
+
+@implementer(ISIPTransport)
+@adapter(IDisposition, IBrowserRequest, logging.Logger)
+class DummyTransport(BaseTransport):
```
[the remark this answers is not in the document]
lukasgraf: Good catch, thanks! I fixed it, and added registrations for our testing transports to the test that verifies their implementation as well.

opengever/disposition/tests/test_delivery.py
```diff
@@ -0,0 +1,119 @@
...
+class TestDeliveryScheduler(LogCapturingTestCase):
+
+    def setUp(self):
+        super(TestDeliveryScheduler, self).setUp()
+        with self.login(self.records_manager):
+            self.register_transport(EnabledTransport, 't-enabled')
+            self.register_transport(DisabledTransport, 't-disabled')
+
+            scheduler = DeliveryScheduler(self.disposition_with_sip)
+            scheduler.schedule_delivery()
```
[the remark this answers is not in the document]
lukasgraf: Not in the `test_delivers_only_using_enabled_transports` though.
Doing the `scheduler.schedule_delivery()` in the setup basically makes sure we always start the state that would result if we had those transports enabled all along (which is what we want to simulate). But because (for technical reasons) the `disposition_with_sip` has to be created in the fixture, any transport we enable after that will not have been active when the respective transition happened. To abstract away that technicality, I also retrigger the `schedule_delivery()` after registering transports.

opengever/disposition/tests/test_delivery.py
```diff
@@ -0,0 +1,119 @@
...
+        scheduler.schedule_delivery(force=True)
+        self.assertEqual({u't-enabled': STATUS_SCHEDULED}, statuses)
+
+        statuses[u't-enabled'] = STATUS_SUCCESS
+        scheduler.schedule_delivery(force=True)
+        self.assertEqual({u't-enabled': STATUS_SCHEDULED}, statuses)
+
+        statuses[u't-enabled'] = STATUS_SCHEDULED
+        scheduler.schedule_delivery(force=True)
+        self.assertEqual({u't-enabled': STATUS_SCHEDULED}, statuses)
```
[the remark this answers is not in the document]
lukasgraf: As discussed: We don't actually care so much about whether the implementation really does reschedule, it's the resulting end state we're interested in.

opengever/disposition/tests/test_delivery.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def test_delivers_only_using_enabled_transports(self):
```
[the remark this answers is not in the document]
lukasgraf: Good point! Updated...

opengever/disposition/tests/test_delivery.py
lukasgraf: Good idea - done :+1:

opengever/disposition/tests/test_disposition.py
lukasgraf: Ouch. You're right 😢 Fixed.

lukasgraf:
Thanks for the dilligent review!
Fixed or commented all the points. 👍

PR #5856, opened 2019-08-07
 Fix availability checks for fileactions for non documentish types.

### LINKED ISSUE #5786: Dokumentaktionen sollten als portal_actions zugänglich sein

njohner:
with https://github.com/4teamwork/opengever.core/pull/5847 we added fileactions as a new category of `portal_actions`. The availability checks for these actions relied on a view which delegates the checks to an adapter of `IFileActions`. We did not handle the case of `portal_type`s for which no such adapter is registered. This is now handled correctly.

**Implementation details**
The `FileActionAvailabilityMixin` was implementing all the methods of the `IFileActions` interface and delegating them to the `IFileActions` adapter. Instead we now overwrite the `__getattr__` method in this mixing and delegate to `IFileActions` adapter when the attribute is a method defined by that interface. In all other cases we raise the `AttributeError` so that normal treatment of attribute search can follow its course.
Also note that when there is no adapter of `IFileActions` registered for the current `context`, but the attribute is part if `IFileActions`, we return a `lambda` function which returns `False`. This is not strictly necessary as these methods are accessed through traversal, so both callables and non-callables are handled. But as `IFileActions` defines only methods, it seems more adequate to also return methods when the adapter is not defined.

## Checkliste
- [ ] Changelog-Eintrag vorhanden/nötig? CL nicht nötig IMHO

resolves #5786

### DIFF
--- a/opengever/api/tests/test_actions.py
+++ b/opengever/api/tests/test_actions.py
@@ -13,6 +13,65 @@ class FileActionsTestBase(IntegrationTestCase):
         return browser.json['file_actions']
 
 
+class TestFileActionsGetForNonDocumentishTypes(FileActionsTestBase):
+
+    @browsing
+    def test_available_file_actions_for_plone_site(self, browser):
+        self.login(self.regular_user, browser)
+        expected_file_actions = []
+
+        self.assertEqual(expected_file_actions,
+                         self.get_file_actions(browser, self.portal))
+
+    @browsing
+    def test_available_file_actions_for_repository_root(self, browser):
+        self.login(self.regular_user, browser)
+        expected_file_actions = []
+
+        self.assertEqual(expected_file_actions,
+                         self.get_file_actions(browser, self.repository_root))
+
+    @browsing
+    def test_available_file_actions_for_repository_folder(self, browser):
+        self.login(self.regular_user, browser)
+        expected_file_actions = []
+
+        self.assertEqual(expected_file_actions,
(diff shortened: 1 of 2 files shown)


opengever/api/tests/test_actions.py
```diff
@@ -13,6 +13,57 @@ def get_file_actions(self, browser, context):
         return browser.json['file_actions']
 
 
+class TestFileActionsGetForNonDocumentishTypes(FileActionsTestBase):
+
```
phgross: I would add a test also for the plone site root.

PR #5871, opened 2019-08-16
Use async mode in keyword widgets for document and dossier keywords.

### LINKED ISSUE #5733: optimise keyword selection
The keyword selection (when adding keywords to documents) becomes very slow when there are a many keywords (a few thousand). It seems we load the whole set of keywords from the backend just once, and then we filter in the frontend, as the user types.

The widget has an `async` option, but it does not work with our current source (https://github.com/4teamwork/ftw.keywordwidget#async-option). The optimal solution would probably be to adapt our source so that we can use the async option.

I have already tested to only show the results after the user has typed several characters (`form.widget('keywords', KeywordFieldWidget, new_terms_as_unicode=True, js_config={"minimumInputLength": 3})`), but this did not really improve the UX. Other proposed solutions are to use pagination:
https://stackoverflow.com/questions/15041058/select2-performance-for-large-set-of-items

This is for issue https://extranet.4teamwork.ch/support/gemeinde-koeniz/tracker/165

njohner:
See https://github.com/4teamwork/ftw.keywordwidget/pull/37 for necessary changes in `ftw.keywordwidget`

To avoid performance issues when the number of keywords is high, we use the async mode for the keyword widgets for document and dossier keywords. For this we:
* Allow adding new terms in async mode (in `ftw.keywordwidget`)
* Use the `async` mode for document and dossier keywords

An additional difficulty was that we have the option of restricting the keywords in a `dossiertemplate`, so that only those keywords should be selectable when creating a dossier from the template. We could not solve this in the `async` mode, as a keyword search will traverse with `++widget++` onto the widget, initialising a new widget in an anonymous request. This does not allow to access the `IWizardDataStorage` and hence the template, to determine whether the vocabulary should restricted or not. We therefore had to:
* Add a new `KeywordWidgetAddableRestrictableSourceWrapper`, which allows to restrict the available keywords (the binder looks for the `IRestrictKeywords` on the request to determine whether to restrict the available keywords or not).
* Allow to use sources also when not in async mode (this was not working in `ftw.keywordwidget`)
* set the widget to `async` mode when necessary

## Checkliste

https://extranet.4teamwork.ch/support/gemeinde-koeniz/tracker/165
resolves #5733


opengever/dossier/dossiertemplate/form.py
```diff
@@ -174,6 +178,26 @@ def _create_form_class(self, parent_form_class, steptitle):
...
+                    # template. This is indicated to the source with the
+                    # IRestrictKeywords interface and the allowed keywords
+                    # are stored on the request. We also need to set the widget
+                    # in the synchronous mode, as otherwise keyword searches
+                    # will anonymously reinitialise a new source with no way of
+                    # accessing the IWizardDataStorage and hence the template.
+                    alsoProvides(self.request, IRestrictKeywords)
+                    self.request.allowed_keywords = IDossier(template_obj).keywords
+
+                    self.groups[0].fields['IDossier.keywords'].widgetFactory["input"] = ParameterizedWidget(
```
lukasgraf: The `self.groups[0]` is bound to break at some point. However, because it's not possible to retrieve a group (fieldset) by name in `z3c.form`, it's almost impossible to handle this without introducing some kind of form helper. So 👍 for now.

PR #5873, opened 2019-08-19
Implement BundleFactory to create OGGBundles

njohner:
We want to be able to generate `OGGBundles` from a folder in a filesystem, two main reasons are:
* Provide the possibility for clients to migrate content from a filesystem into Gever (partial bundles, i.e. import content into an already existing Gever repository tree)
* Provide a way to generate full bundles (complete repository tree with content), notably to easily create large repository structures for testing, for example from wikipedia content.

*Implementation:*
* I started from the WIP implementation from @lukasgraf (also took some inspiration from @tinagerber's training implementation: https://github.com/4teamwork/training_tina/pull/16)
* refactored it to better split walking the filesystem from generating the content of the bundle:
  * Walker should not need to know about the format of the bundle
  * Added `OGGBundleItemCreator` responsible for deciding what type of item (reporoot, repofolder, dossier or document) to create from a certain node returned by the walker.
  * Added distinct class for each type of `OGGBundle` item type
* Added options necessary to allow both full and partial bundles
* Added some documentation for the `bundle-factory` script

I also tried to improve the test of the ogg-bundle pipeline (`TestOggBundlePipeline`). I notably made it flat (avoid nested assertions), renamed the objects, added a comment describing the expected content of the bundle imported and made the test less sensitive to factory changes by getting the created object from their titles instead of their id.

This is for #4522

## Checkliste


opengever/bundle/tests/test_bundle_factory.py
```diff
@@ -0,0 +1,282 @@
...
+
+        document = self.find_item_by_title(documents_in_subdossier, 'Pflanzensippe.docx')
+        self.assertEqual(pjoin(input_path, root_dossier['title'], subdossier['title'], document['title']),
+                         document['filepath'])
+
+    def test_full_bundle_factory(self):
+        parser = get_parser()
+        input_path = resource_filename('opengever.bundle.tests',
+                                       'assets/basic_import_repository')
+        outpath = mkdtemp()
```
lukasgraf: Mmh, this will leave behind temporary files. Create the tempdir in `setUp` and remove it in `tearDown`? Or use the [`TEMP_DIRECTORY`](https://github.com/4teamwork/ftw.testing#temp-directory-layer) layer from ftw.testing if easily possible.

docs/public/dev-manual/oggbundle/index.rst
```diff
@@ -453,3 +453,14 @@ Die JSON-Schemas, welche die Struktur der JSON-Dateien für die Metadaten defini
...
        :language: json
+
+
+========================
+Generieren von OGGBundle
+========================
+
+Mit ``bin/bundle-factory`` kann ein ``OGGBundle`` von einem Datenverzeichnis generiert werden.
+
+-  Wenn ``--repo-nesting-depth`` gesetzt ist, wird der Skript ein ``OGGBundle`` für ein komplettes ``Ordnungssystem`` generieren. In diesem Fall wird das ``source_dir`` im ``OGGBundle`` als eine ``reporoot`` abgebildet, und alle Verzeichnisse welche ein Verschachtelungstiefe geringer als ``--repo-nesting-depth`` haben werden als ``repofolders`` abgebildet. Andere Verzeichnisse als ``dossiers`` und Dateien als ``documents``.
```
lukasgraf: der Skript -> das Skript
eine reporoot -> ein reporoot
ein Verschachtelungstiefe -> eine Verschachtelungstiefe

docs/public/dev-manual/oggbundle/index.rst
```diff
@@ -453,3 +453,14 @@ Die JSON-Schemas, welche die Struktur der JSON-Dateien für die Metadaten defini
...
+
+- Wenn ``--repo-nesting-depth`` nicht gesetzt ist (``--repo-nesting-depth=-1``), dann generiert der Skript ein ``OGGBundle`` für ein partieller Import. In diesem Fall wird das ``source_dir`` im ``OGGBundle`` nicht abgebildet, alle enthaltene Verzeichnisse werden als ``dossiers`` abgebildet und Dateien als ``documents``. Die Ordnungsposition oder das Dossier in welches das ``OGGBundle`` importiert werden soll, muss mann mit ``--import-repository-references`` und optional ``--import-dossier-reference`` spezifizieren.
```
lukasgraf: der Skript -> das Skript
ein partieller Import -> einen partiellen Import
enthaltene Verzeichnisse -> enthaltenen Verzeichnisse
mann -> man

opengever/bundle/factory.py
```diff
@@ -192,12 +198,20 @@ def __init__(self, args):
         self.users_group = args.users_group
         self.dossier_responsible = args.dossier_responsible
 
+        self.import_reference = []
+        if args.import_repository_reference is not None:
+            self.import_reference.append(args.import_repository_reference)
```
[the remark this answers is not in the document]
njohner: The `parent_reference` in the bundle is a list of lists of ints in the form `[[1, 3, 5], [472, 9]]` for subdossier with position `1.3.5 / 472.9` (see https://docs.onegovgever.ch/dev-manual/oggbundle/#parent-reference). So append is correct here. `import_repository_reference` and `import_dossier_reference` are both lists of ints and get appended to the `import_reference` which I will then use in the bundle.

opengever/bundle/tests/test_bundle_factory.py
njohner: Ah thanks! I made a `setUp` and `tearDown`as in https://github.com/4teamwork/opengever.core/blob/master/opengever/base/schemadump/tests/test_json_schema_helper.py#L13-L17
I did not want to make a new layer for two tests, seemed a bit of an overkill

docs/public/dev-manual/oggbundle/index.rst
njohner: Thanks!

docs/public/dev-manual/oggbundle/index.rst
njohner: thanks!

njohner:
Ok all done! Thanks for the review!

PR #5875, opened 2019-08-21
Add Base Response functionality to ToDos

phgross:
Implements #5854 and is part of the issue to support response in the workspace todos (frontend work see https://github.com/4teamwork/gever-ui/pull/404).

Technical details:
 - The response are stored in a Btree in the annotations of the object, and every Response has his own id (long of the current time). The storage functionality is inspired by the plone.app.discussion implementation
 - Objects/types can be marked with the marker interface `IResponseSupported`, so there API representation is extended with an additional list `responses` and additional `@response` endpoints are available (GET, POST, PATCH).
 - Tracking changes is not part of this PR
 - The validation is currently not generic, because its only one field, makes more sense to implement this with #5813 

## Checkliste


opengever/base/tests/test_response.py
```diff
@@ -0,0 +1,68 @@
...
+        self.assertNotIn(ResponseContainer.ANNOTATION_KEY, annotations.keys())
+
+        response = Response('Bitte weiter abklaeren.')
+        container = IResponseContainer(self.todo)
+        container.add(response)
+
+        self.assertEqual([response],
+                         annotations[ResponseContainer.ANNOTATION_KEY])
+
+    def test_add_expect_response_objects(self):
```
njohner: This test is actually a copy of the previous one and does not test what is advertised.

I think that `add` will allow anything right now? Maybe add a check for `IResponse` in `ResponseContainer.add`? And then add a assertion that only objects providing `IResponse` can be added, something like:
```
noLongerProvides(response, IResponse)
with self.assertRaises(...):
    container.add(response)
```

opengever/base/tests/test_response.py
```diff
@@ -0,0 +1,68 @@
...
+        self.assertEqual([response],
+                         annotations[ResponseContainer.ANNOTATION_KEY])
+
+    def test_list_returns_response_in_add_order(self):
+        self.login(self.workspace_member)
+        IResponseContainer(self.todo).list()
+
+        container = IResponseContainer(self.todo)
+
+        responses = [Response('1.'), Response('2.'), Response('3.')]
```
njohner: This could also be alphabetical. Can you use something that is not in alphabetical order like:
responses = [Response('2.'), Response('1.'), Response('3.')]

opengever/base/response.py
```diff
@@ -11,6 +11,11 @@
 from zope.interface import Interface
 
 
+class IResponseSupported(Interface):
+    """Marker interface for response
```
njohner: Marker interface for response support?

opengever/base/response.py
```diff
@@ -0,0 +1,85 @@
...
+        """Returns a list of all response."""
+
+
+@implementer(IResponseContainer)
+@adapter(Interface)
+class ResponseContainer(object):
+    """Response storage adapter.
+    """
+
+    ANNOTATION_KEY = '_responses'
```
elioschmutz: The convention is to use (or to append) the module-name: `opengever.base.responses` to prevent for conflicting keys.

opengever/base/response.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def __init__(self, context):
+        self.context = context
+
+    def add(self, response):
+        self.storage(create_if_missing=True).append(response)
+
+    def storage(self, create_if_missing=False):
```
elioschmutz: What's the benefit of using `create_if_missing`? I excpect the storage is always returning the storage, in this case a PersistentList, if I call it.

opengever/base/response.py
```diff
@@ -0,0 +1,85 @@
...
+        ann = IAnnotations(self.context)
+        if self.ANNOTATION_KEY not in ann.keys() and create_if_missing:
+            ann[self.ANNOTATION_KEY] = PersistentList()
+
+        return ann.get(self.ANNOTATION_KEY, None)
+
+    def list(self):
+        storage = self.storage()
+        if not storage:
+            return []
```
elioschmutz: If the storage is always returning the `Persistent list, this check would no longer be necessary.

opengever/api/tests/test_response.py
```diff
@@ -0,0 +1,45 @@
...
+    def test_returns_list_of_serialized_response_objects(self, browser):
+        self.login(self.workspace_member, browser=browser)
+
+        with freeze(datetime(2016, 12, 9, 9, 40)):
+            IResponseContainer(self.todo).add(
+                Response('Ich bin hier anderer Meinung!'))
+
+        with freeze(datetime(2016, 12, 24, 8, 23)):
+            with self.login(self.workspace_admin):
+                response = Response(u'Ok, Fanke f\xfcr dein Feedback')
```
njohner: 'Danke' and not 'Fanke'?

opengever/api/tests/test_response.py
```diff
@@ -0,0 +1,45 @@
...
+                IResponseContainer(self.todo).add(response)
+
+        browser.open(self.todo, method="GET", headers=self.api_headers)
+        self.assertEquals(
+            [{u'created': u'2016-12-09T09:40:00',
+              u'creator': self.workspace_member.id,
+              u'text': u'Ich bin hier anderer Meinung!'},
+             {u'created': u'2016-12-24T08:23:00',
+              u'creator': self.workspace_admin.id,
+              u'text': u'Ok, Fanke f\xfcr dein Feedback'}],
```
njohner: Danke

opengever/base/response.py
```diff
@@ -26,6 +28,9 @@ def add(self, response):
     def list(self):
         """Returns a list of all response."""
```
njohner: responses

opengever/base/response.py
```diff
@@ -26,6 +28,9 @@ def add(self, response):
...
 
+    def items(self):
+        """Returns a list of (id, response object) pairs of all response."""
```
njohner: responses

opengever/base/response.py
```diff
@@ -39,21 +44,45 @@ def __init__(self, context):
...
 
         return ann.get(self.ANNOTATION_KEY, None)
 
     def list(self):
-        storage = self.storage()
+        storage = self._storage()
+        if not storage:
+            return []
+
+        return [response for response in storage.values()]
```
njohner: Is the list comprehension useful here? would `storage.values()` also work?
otherwise maybe `list(storage.values())`?

opengever/base/response.py
```diff
@@ -39,21 +44,45 @@ def __init__(self, context):
...
+
+    def items(self):
+        storage = self._storage()
         if not storage:
             return []
 
-        return storage
+        return [item for item in storage.items()]
```
njohner: As above, would `storage.items()` work here?

opengever/base/tests/test_response.py
```diff
@@ -35,8 +35,10 @@ def test_stores_response_in_the_annotations(self):
         container = IResponseContainer(self.todo)
         container.add(response)
 
-        self.assertEqual([response],
-                         annotations[ResponseContainer.ANNOTATION_KEY])
+        storage = annotations[ResponseContainer.ANNOTATION_KEY]
+        self.assertEqual(
+            [response],
+            [response for reponse in storage.values()])
```
njohner: I'd rather you used a different variable name in the list comprehension, as you are overwriting `response` here (though it does not matter here, as this is the end of the test...)

opengever/base/tests/test_response.py
```diff
@@ -48,8 +50,10 @@ def test_add_expect_response_objects(self):
         container = IResponseContainer(self.todo)
         container.add(response)
 
-        self.assertEqual([response],
-                         annotations[ResponseContainer.ANNOTATION_KEY])
+        storage = annotations[ResponseContainer.ANNOTATION_KEY]
+        self.assertEqual(
+            [response],
+            [response for response in storage.values()])
```
njohner: Same as above

docs/HISTORY.txt
```diff
@@ -4,6 +4,7 @@ Changelog
 2019.4.0rc2 (unreleased)
 ------------------------
 
+- Add response support for ToDos. [phgross]
```
elioschmutz: ✂️ WIP in commit message...

docs/public/dev-manual/api/workspace/todos.rst
```diff
@@ -103,3 +103,111 @@ Ein ToDo löschen
...
+      GET workspaces/workspace-1/todo-1 HTTP/1.1
+      Accept: application/json
+
+
+**Beispiel-Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Accept: application/json
```
elioschmutz: `Content-Type` instead `Accept`

docs/public/dev-manual/api/workspace/todos.rst
```diff
@@ -103,3 +103,111 @@ Ein ToDo löschen
...
+Erstellung und Bearbeitung
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Antworten können mit einem POST Request auf den ``@responses`` Endpoint hinzugefügt werden.
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+      POST workspaces/workspace-1/todo-1/responses HTTP/1.1
```
elioschmutz: `@responses` instead `responses`

opengever/api/response.py
```diff
@@ -84,3 +87,22 @@ def reply(self):
         serializer = getMultiAdapter((response, self.request), ISerializeToJson)
         return serializer(
             container=self.context, response_id=self._get_response_id)
+
+
+class ResponsePost(Service):
+    """Add an Response to the current context.
```
njohner: typo: "Add a Response"

opengever/api/response.py
```diff
@@ -30,32 +38,113 @@ def get_value(self, default=None):
...
-            'added_object': json_compatible(self.context.added_object),
-            'changes': json_compatible(self.context.changes),
-        }
-
-        # Add attributes from the response schema
-        for name, field in getFields(ITaskTransitionResponseFormSchema).items():
-            if name in self.SKIPPED_FIELDS:
-                continue
+        response_url = '{}/@responses/{}'.format(
+            kwargs['container'].absolute_url(), kwargs['response_id'])
```
elioschmutz: wouldn't it be possible to store the `response_url` directly on the response-object? Then we don't have to create this url here in also don't have to call the serializer with the `container` and `response_id`... 

How do you see this point?

opengever/api/tests/test_response.py
```diff
@@ -74,3 +75,49 @@ def test_returns_the_serialized_response_objects(self, browser):
...
+    def test_adding_a_responses_requires_edit_permission(self, browser):
+        self.login(self.workspace_guest, browser=browser)
+
+        with browser.expect_http_error(401):
+            url = '{}/@responses'.format(self.todo.absolute_url())
+            browser.open(url, method="POST", headers=self.api_headers,
+                         data=json.dumps({'text': u'Angebot \xfcberpr\xfcft'}))
+
+    @browsing
+    def test_adding_a_responses(self, browser):
```
njohner: as above: `response`

opengever/api/tests/test_response.py
```diff
@@ -121,3 +121,49 @@ def test_data_is_validated(self, browser):
...
+
+        with freeze(datetime(2016, 12, 9, 9, 40)):
+            IResponseContainer(self.todo).add(Response('Test'))
+
+        with browser.expect_http_error(500):
+            url = '{}/@responses/1481272800000000'.format(self.todo.absolute_url())
+            browser.open(url, method="PATCH", headers=self.api_headers)
+
+        self.assertEquals(
+            {u'message': u'text', u'type': u'RequiredMissing'}, browser.json)
```
njohner: Could you test here that it did not overwrite the text in the response (`IResponseContainer(self.todo).list()[0].text=='Test'`)

docs/public/dev-manual/api/workspace/todos.rst
```diff
@@ -103,3 +103,111 @@ Ein ToDo löschen
...
 
       HTTP/1.1 204 No Content
+
+
+Antworten
+---------
+
+ToDo Antworten sind ebenfalls via API ersichtlich bzw. können über diese erstellt und bearbeitet werden.
+
+Die API Repräsentation eines ToDos, listet unter dem Attribut ``responses`` alle Antworten auf. Eine GET Request auf eine einzelne Aufgabe ist ebenfalls möglich und antwortet mit der gleichen Repräsentation.
```
njohner: Did you mean `Aufgabe` here or `ToDo`?

opengever/base/response.py
njohner: This is to avoid write on read...

opengever/base/response.py
phgross: The reason is to avoid writes on read.

elioschmutz: Ok 👍

opengever/base/response.py
elioschmutz: Ok 👍

phgross: see above (writes on read)

opengever/api/response.py
phgross: I was thinking about this point too, adding the response_url is not an option otherwise we would persist the complete url in the response object. So URL changes or access via ssh tunnels would lead in to wrong urls.

Storing the response_id on the response_object itself could maybe be helpful and is possible to implement. @njohner @elioschmutz should I? 🤔

elioschmutz: Ah of course, storing the full URL would be bad...

My vote: 👍 for storing the `response_id` on the response-object itself...

opengever/base/tests/test_response.py
phgross: 🙈 😅 thx

opengever/api/response.py
njohner: Yes I think that's a good idea.

phgross:
@elioschmutz @njohner updated according to your feedback.

opengever/api/tests/test_response.py
```diff
@@ -0,0 +1,178 @@
...
+        with freeze(datetime(2016, 12, 9, 9, 40)):
+            IResponseContainer(self.todo).add(Response('Test'))
+
+        with browser.expect_http_error(401):
+            url = '{}/@responses/1481272800000000'.format(self.todo.absolute_url())
+            browser.open(url, method="PATCH", headers=self.api_headers,
+                         data=json.dumps({'text': 'test'}))
+
+    @browsing
+    def test_adding_a_responses(self, browser):
```
njohner: Oh seems the method name is wrong

njohner: Should be `test_adding_a_responses`

phgross: 🙈 updated ...

njohner:
Seems the tests are stuck...

opengever/base/response.py
```diff
@@ -1,113 +1,125 @@
...
-        if self.status:
-            self.request.response.setStatus(self.status)
+        return self._storage()[long(key)]
+
+
+class IResponse(Interface):
+    """Interface and schema for the response object, an object added to
+    plone content objects."""
+
+    created = schema.Date(
```
elioschmutz: Die nachträglich hinzugefügte `response_id` fehlt noch in der Schema-Definition...

Issue #5799, opened 2019-07-09
API Support für Anträge


phgross:
_Harvest: `Projekt - Umsetzung Neues User Interface GEVER 2019/2020` - `Entwicklung`_

- GET eines Antrags (bereits teilweise unterstützt)
- POST eines Antrags (teilweise unterstütz aber wohl noch fehlende Metadaten)
- PATCH eines Antrags

Als API Benutzer will ich die Details von Anträgen fetchen und auch neue Anträge erstellen. 

Diskussion:
 - [x] Alle Proposal Felder in Plone Schema implementieren inkl. Datenmigration 
 - [x] Eventhandler für Aktualisierung der SQL Daten
 - [x] Workflow-Status Änderungen
 - [x] Display workflow transition button discreet placeholders for submit/cancel when:
    - proposal contains checked out documents (submit/cancel)
    - proposal committee is inactive (submit)
- [x] (@njohner) custom proposal API `POST` that takes additional form data so that proposals can be created like from web-rendered form. must also create proposal document when proposal is created.
- [x] (@deiferni) make sure to trigger proposal state changes remotely. Makes states via api visible. Do the following:
  - create a view to trigger the state change, that also creates a history entry
  - no longer sync history via `@@receive-proposal-history`
  - affects:
    - `ProposalRemovedFromSchedule` (synced @@receive-proposal-history), triggers `proposal-transition-unschedule`
    - `ProposalScheduled` (synced @@receive-proposal-history), triggers `proposal-transition-schedule`
    - `ProposalDecided` (synced @@receive-proposal-history, during model.Proposal.decide), triggers `proposal-transition-decide`
     -  `ProposalRejected`
  - ~~render proposal state on overview, not model state.~~

Don't forget to:
- [x] drop `Proposal.is_editable_in_dossier`
- ~~[ ] maybe add abstract base for `ProposalTransitionController` and `TaskTransitionController`~~
- [x] drop ModelProxyTabbedView for proposals 
- ~~[ ] potentially get rid of `Proposal.WORKFLOW_STATE_TO_SQL_STATE`~~
- [x] get rid of sql proposal transition forms, or explicitly use for submitted proposal
- [x] drop `ProposalModel.is_submitted`
- [x] show warning for inactive committees on proposal overview
```python
        if not model.committee.is_active():
            api.portal.show_message(
                _(u'msg_inactive_committee_selected',
                  default=u'The selected committeee has been deactivated, the'
[...]

PR #5877, opened 2019-08-21
Move remaining proposal model schema fields to proposal plone content type.

deiferni:
With https://github.com/4teamwork/opengever.core/issues/5799 we would like to provide API support for proposals. To do so we must in a first step make sure all data is stored on plone content objects as primary data source and only synced to SQL later. This PR provides only a first, small step towards API support. It adds all proposal fields that were in SQL only to the proposal plone content type. It provides upgrades that store data that were only in SQL on the plone content objects.

The following fields are moved:
- `language`, a string representing the selected lanugage
- `committee_oguid`, to store a reference to the committee. i have chosen to store the oguid in plone as that is on the abstraction level of plone content objects. unfortunuately we cannot use plones references as the referenced committee can theoretically be on a different plone site.

## Checkliste

### DIFF
--- a/opengever/core/upgrades/20190821181500_copy_proposal_language_from_model_to_plone_content/upgrade.py
+++ b/opengever/core/upgrades/20190821181500_copy_proposal_language_from_model_to_plone_content/upgrade.py
@@ -1,0 +1,45 @@
+from opengever.core.upgrade import SQLUpgradeStep
+from opengever.base.oguid import Oguid
+from sqlalchemy.sql.expression import column
+from sqlalchemy.sql.expression import table
+
+
+proposals_table = table(
+    "proposals",
+    column("id"),
+    column("admin_unit_id"),
+    column("int_id"),
+    column("submitted_admin_unit_id"),
+    column("submitted_int_id"),
+    column("language"),
+)
+
+
+class CopyProposalLanguageFromModelToPloneContent(SQLUpgradeStep):
+    """Copy proposal language from model to plone content.
+    """
+
+    def migrate(self):
+        for proposal in self.objects(
+                {'portal_type': ['opengever.meeting.proposal']},
+                'Copy language field from model to plone content.'):
+            oguid = Oguid.for_object(proposal)
+
+            row = self.execute(
+                    proposals_table.select()
+                    .where(proposals_table.c.admin_unit_id == oguid.admin_unit_id)
+                    .where(proposals_table.c.int_id == oguid.int_id)
+                ).fetchone()
+            proposal.language = row.language
+
+        for submitted_proposal in self.objects(
+                {'portal_type': ['opengever.meeting.submittedproposal']},
+                'Copy language field from model to plone content.'):
+            oguid = Oguid.for_object(submitted_proposal)
+
+            row = self.execute(
+                    proposals_table.select()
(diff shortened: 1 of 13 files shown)


deiferni:
🚧 please do hold here. i have to do some separate work, can't convert to draft again. sorry.

njohner:
Yes I've missed that feature already several times...

opengever/core/upgrades/20190821181500_copy_proposal_language_from_model_to_plone_content/upgrade.py
```diff
@@ -0,0 +1,48 @@
...
+    column("language"),
+    column("committee_id"),
+)
+
+
+class CopyProposalLanguageFromModelToPloneContent(SQLUpgradeStep):
+    """Copy proposal language from model to plone content.
+    """
+
+    deferrable = True
```
njohner: I don't think making it deferrable is a good idea. If anyone does something with a proposal before we do the upgrade, syncing of the data will overwrite the `language` in the database with the default value for the field.

opengever/core/upgrades/20190821181535_copy_proposal_committee_from_model_to_plone_content/upgrade.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+proposals_table = table(
+    "proposals",
+    column("id"),
+    column("admin_unit_id"),
+    column("int_id"),
+    column("submitted_admin_unit_id"),
+    column("submitted_int_id"),
+    column("language"),
```
njohner: You don't need the language column I think

opengever/core/upgrades/20190821181535_copy_proposal_committee_from_model_to_plone_content/upgrade.py
```diff
@@ -0,0 +1,65 @@
...
+    column("admin_unit_id"),
+    column("int_id"),
+)
+
+
+class CopyProposalCommitteeFromModelToPloneContent(SQLUpgradeStep):
+    """Copy proposal committee from model to plone content.
+    """
+
+    deferrable = True
```
njohner: As before, I don't think you can make that deferrable. Also do we want to have separate upgrade steps for every field? This will decrease performance and I don't see to much gain in readability or such...

opengever/core/upgrades/20190821181500_copy_proposal_language_from_model_to_plone_content/upgrade.py
deiferni: as discussed, imo this is borderline deferrable. we don't have many proposals in production so there is not so much benefit. i have dropped deferrability from both upgrades.

opengever/core/upgrades/20190821181535_copy_proposal_committee_from_model_to_plone_content/upgrade.py
deiferni: dropped the deferrable. separate upgrades group relevant changes and upgrades in one commit. we don't have much content in production so as discussed i'll leave it as is.

deiferni:
- typo is fixed
- squashed the one commit you mentioned, it was a leftover
- see inline comments, ready for re-review

PR #5879, opened 2019-08-21
Split proposal and submitted proposal interfaces, introduce common base for both.

deiferni:
This PR is part of https://github.com/4teamwork/opengever.core/issues/5799. In order to be able to easily modify proposal and submitted proposal fields we first want to make sure that each side has its own distinct set of fields. Currently the submitted proposal schema interface inherits from the proposal schema interface, which automatically adds all proposal fields to submitted proposal. However at the moment there are already fields which should only be available on a proposal. With https://github.com/4teamwork/opengever.core/issues/5799 we want to introduce more such fields.

- Add `IBaseProposal` schema interface class, `ISubmittedProposal` no longer inherits from `IProposal` but both inherit from `IBaseProposal`
- Change adapter registration to the correct base class. We can now easily declare for what side an adapter should be available.
- Add upgrades to drop removed field from submitted proposal objects and to reindex `object_provides`

## Checkliste


opengever/core/upgrades/20190821161339_drop_unused_fields_from_submitted_proposal/upgrade.py
```diff
@@ -0,0 +1,21 @@
+from ftw.upgrade import UpgradeStep
+from Products.CMFPlone.utils import base_hasattr
+
+
+REMOVED_FIELDS = ('relatedItems', 'predecessor_proposal')
```
njohner: As discussed, it's fine to drop this. Relations are not transported when creating a `SubmittedProposal`. In `CreateSubmittedProposal`, we use the `DexterityFieldDataCollector` to get the data, which skips relations (see `DexterityFieldDataCollector.pack`). So this field was never set on the `SubmittedProposal`.

opengever/base/tests/test_default_values_for_types.py
```diff
@@ -1335,6 +1349,62 @@ def test_z3c_add_form(self, browser):
...
+
+        persisted_values = get_persisted_values_for_obj(proposal)
+        expected = self.get_type_defaults()
+
+        self.assertDictEqual(expected, persisted_values)
+
+    def get_obj_of_own_type(self):
+        pass
+
+    def test_z3c_form_defaults_cover_all_schema_fields(self):
```
njohner: We skip this because we don't have a form right?

opengever/core/upgrades/20190821171113_reindex_proposal_and_submitted_proposal_object_provides/upgrade.py
```diff
@@ -0,0 +1,14 @@
...
+    """Reindex proposal and submitted proposal object provides.
+    """
+
+    def __call__(self):
+        for proposal in self.objects(
+                {'portal_type': ['opengever.meeting.proposal',
+                                 'opengever.meeting.submittedproposal']},
+                'Reindex object_provides for all proposal-ish.'):
+
+            proposal.reindexObject(idxs=['object_provides'])
```
njohner: Just to make sure, we do not need the implicit metadata reindexing here right?

opengever/meeting/browser/configure.zcml
```diff
@@ -182,7 +182,7 @@
       />
 
   <browser:page
-      for="opengever.meeting.proposal.IProposal"
+      for="opengever.meeting.proposal.IBaseProposal"
```
njohner: I think that can be registered for `IProposal`. It is only used from the `SubmittedProposal` side to reflect history changes into the `Proposal`

opengever/core/upgrades/20190821171113_reindex_proposal_and_submitted_proposal_object_provides/upgrade.py
deiferni: as just discussed, no. we're only interested in the index.

opengever/base/tests/test_default_values_for_types.py
deiferni: yes, exactly. there is no add form for submitted proposals.

opengever/meeting/browser/configure.zcml
deiferni: this is also used for commenting, commenting can go both ways.  I think you may have been mislead by https://github.com/4teamwork/opengever.core/blob/14d94bdb53bdfed5209005c3ecdcc0c24bb1b689/opengever/meeting/proposalhistory.py#L70, the `is_submitted` check is a state check, not a type check.

deiferni:
i have fixed the typo 😄.

PR #5880, opened 2019-08-21
Add filename and checked_out fields to the recently-touched endpoint.

njohner:
To allow to easily display an icon and a hint whether a document is checked out or not in the recently touched menu of the new UI, we add the `filename` and `checked_out` fields to the `recently-touched` endpoint.

Note that we have a discrepancy between the `last_touched` date for recently touched vs checked out documents. In one case it's timezone aware and in the other it's not. See comment in the tests.

This is for https://github.com/4teamwork/opengever.core/issues/5760

## Checkliste

### DIFF
--- a/opengever/api/tests/test_recently_touched.py
+++ b/opengever/api/tests/test_recently_touched.py
@@ -41,7 +41,9 @@ class TestRecentlyModifiedGet(IntegrationTestCase, ResolveTestHelper):
                  'icon_class': 'icon-docx',
                  'last_touched': '2018-04-30T00:00:00',
                  'target_url': self.document.absolute_url(),
-                 'title': u'Vertr\xe4gsentwurf'}]},
+                 'title': u'Vertr\xe4gsentwurf',
+                 'filename': u'Vertraegsentwurf.docx',
+                 'checked_out': ''}]},
             browser.json)
 
     @browsing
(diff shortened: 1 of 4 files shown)


opengever/api/tests/test_recently_touched.py
```diff
@@ -65,10 +67,45 @@ def test_lists_checked_out_docs_for_given_user(self, browser):
                 'icon_class': 'icon-docx is-checked-out-by-current-user',
                 'last_touched': '2018-04-30T00:00:00+02:00',
```
njohner: Here the date is timezone aware

opengever/api/tests/test_recently_touched.py
```diff
@@ -65,10 +67,45 @@ def test_lists_checked_out_docs_for_given_user(self, browser):
...
+        url = '%s/@recently-touched/%s' % (
+            self.portal.absolute_url(), self.regular_user.getId())
+        browser.open(url, method='GET', headers={'Accept': 'application/json'})
+
+        self.assertEqual(200, browser.status_code)
+        self.assertEquals(
+            {'checked_out': [],
+             'recently_touched': [{
+                'icon_class': 'icon-docx is-checked-out',
+                'last_touched': '2018-04-30T00:00:00',
```
njohner: Now here for a document checked out at the exact same time as above, but the timezone naive date... I guess in local time this is fine. If we were to consider this to be UTC, there would be a shift of 2 hours between the two time tags

njohner:
New issue opened: https://github.com/4teamwork/opengever.core/issues/5887

Issue #5744, opened 2019-06-21
ftw.flamegraph zu GEVER dependencies hinzufügen


phgross:
Bei Performance Problemen in der Produktion ist es hilfreich wenn man gleich die Funktionalitäten von ftw.flamegraph verwenden kann.

njohner:
As discussed in https://github.com/4teamwork/opengever.core/pull/5882, the problem is that if `ftw.flamegraph` is installed, it is available for all users, so that any user can see the full trace for any request he can make. We can't make it manager only, as we want the flamegraph to work for cases where `zauth` is used to impersonate another user, and it is not possible to distinguish an impersonated user from another user.
We therefore let it be for now and just close this issue.

PR #5882, opened 2019-08-22
Add ftw.flamegraph as dependency.

njohner:
To make our life easier when we have performance issues in production, we add `ftw.flamegraph`
as dependency.

This is for https://github.com/4teamwork/opengever.core/issues/5744

## Checkliste


njohner:
As @deiferni pointed out, this means that any user can get a flamegraph of any request he can make

njohner:
We finally decided against installing `ftw.flamegraph`, as this would not be ideal from the perspective of security. We therefore close this PR.

PR #5890, opened 2019-08-26
Drop custom model forms for proposals and create proposal model via event handler.

deiferni:
This PR is part of https://github.com/4teamwork/opengever.core/issues/5799. we move functionality that creates SQL records and syncs to SQL out of add and edit forms and use event handlers instead. In order to be able to easily create proposals via api endpoint and make sure the corresponding SQL records are created and synced automatically via event handler. This further decouples SQL from plone objects. Functionality and knowledge about what to sync from proposals and submitted proposals is also moved to the SQL model. The following things are addressed in this PR:

- Forms for proposals are no longer inheriting from `ModelProxyEditForm`
- Creating/Updating proposal record is now done via event handler
- Some care has to be taken when creating submitted proposals, we make sure that the Oguid for the newly created submitted proposal plone content object is also stored on the SQL record

## Checkliste

### DIFF
--- a/opengever/meeting/browser/proposalforms.py
+++ b/opengever/meeting/browser/proposalforms.py
@@ -7,10 +7,8 @@ from opengever.meeting import _
 from opengever.meeting import is_meeting_feature_enabled
 from opengever.meeting.activity.watchers import add_watcher_on_proposal_created
 from opengever.meeting.activity.watchers import change_watcher_on_proposal_edited
-from opengever.meeting.form import ModelProxyEditForm
 from opengever.meeting.interfaces import IHistory
 from opengever.meeting.proposal import IProposal
-from opengever.meeting.proposal import SubmittedProposal
 from opengever.officeconnector.helpers import is_officeconnector_checkout_feature_enabled  # noqa
 from opengever.tabbedview.helper import document_with_icon
 from plone import api
@@ -23,7 +21,6 @@ from plone.dexterity.interfaces import IDexterityFTI
 from plone.z3cform.fieldsets.utils import move
 from Products.CMFCore.interfaces import IFolderish
 from Products.CMFPlone.utils import safe_unicode
-from z3c.form import field
 from z3c.form.browser import radio
 from z3c.form.browser.checkbox import SingleCheckBoxFieldWidget
 from z3c.form.interfaces import HIDDEN_MODE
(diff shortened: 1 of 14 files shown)


opengever/meeting/model/proposal.py
```diff
@@ -179,6 +180,35 @@ class Proposal(Base):
...
+        workflow_state = proposal.workflow.default_state.name
+
+        model = cls(oguid=Oguid.for_object(proposal),
+                    workflow_state=workflow_state,
+                    physical_path=proposal.get_physical_path())
+        model.sync_with_proposal(proposal)
+        return model
+
+    def sync_with_proposal(self, proposal):
+        from opengever.meeting.model.committee import Committee
```
njohner: Why the import in the method?

opengever/meeting/browser/proposalforms.py
```diff
@@ -85,14 +88,7 @@ def updateFields(self):
     def updateWidgets(self):
         super(SubmittedProposalEditForm, self).updateWidgets()
         self.widgets['excerpts'].mode = HIDDEN_MODE
-
-        if self.context.get_state() is not self.context.load_model().STATE_PENDING:
```
njohner: Isn't that a change in behaviour? Before the issuer was sometimes editable on the submitted proposal side (though I don't understand why we should allow that) and now it's not?

opengever/meeting/model/proposal.py
deiferni: unfortunately otherwise there are cyclic imports 😞

opengever/meeting/browser/proposalforms.py
deiferni: no this is only for the submitted proposal, the state for a submitted proposal can never be `pending`, it is always one of `submitted`, `scheduled` or `decided`.

deiferni:
aab84ac: the commit message does not make much sense: "...as we can load return we need from the model instance." -> did you mean something like "...as we can get what we need from the model instance." Maybe?
- yes something seems to have happened here. it is as you suggested, i have changed the commit message to have a clear history.

53103ce: Can you comment on why inheritance order has to be changed? I see from the message it has something to do with the FTI, but really no idea what
- fixed an issue with interfaces not being provided correctly. I am also a bit at loss why that happened.

2645225: Wasn't the oguid already hardcoded? It simply changed for some reason? Or is it now the real oguid of the committee from the fixture and before it was just a random one?
- exactly, before the oguid was not the same as provided by the fixture

PR #5892, opened 2019-08-26
Handle special characters in link to advanced search.

njohner:
When searching for a term with special character, the url for the `advanced search` button was not url-encoded, which would raise a 400 on IE. We now properly encode the url.

For https://github.com/4teamwork/opengever.core/issues/5859, support issue https://extranet.4teamwork.ch/extern/opengever-kanton-zug/sprint-backlog/20437/

## Checkliste

### DIFF
--- a/opengever/base/tests/test_search.py
+++ b/opengever/base/tests/test_search.py
@@ -56,6 +56,17 @@ class TestOpengeverSearch(IntegrationTestCase):
         self.assertNotIn('OR', prefill)
         self.assertNotIn('NOT', prefill)
 
+    @browsing
+    def test_advanced_search_link_is_url_encoded(self, browser):
+        self.login(self.regular_user, browser=browser)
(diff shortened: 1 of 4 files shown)


opengever/base/tests/test_search.py
```diff
@@ -56,6 +56,16 @@ def test_prefill_from_advanced_search_omits_searchable_text_keywords(self, brows
         self.assertNotIn('OR', prefill)
         self.assertNotIn('NOT', prefill)
 
+    @browsing
+    def test_advanced_search_link_is_url_encoded(self, browser):
+        self.login(self.regular_user, browser=browser)
+        browser.open(self.portal, view='@@search?SearchableText=M\xc3\xbcller')
```
deiferni: Should we document, that other unsafe characters are also handled, like the whitespace, `{`, ...?

Issue #5775, opened 2019-07-04
@listing endpoint should allow params isSubdossier/isSubtask


all4manu:
Harvest: `Projekt - Umsetzung Neues User Interface GEVER 2019/2020` - `Entwicklung`

When querying the @listing endpoint the new gever frontend needs to be able to pass in the parameters `isSubdossier=false` and `isSubtask=false` so we can get the listing for multiple hierarchy depths but only receive the main dossiers/tasks.

njohner:
# Todos
- [x] Neuer Index `is_subtask` (catalog, solr): Index already existed in catalog. Added in solr.
- [x] Neuer Indexer registieren für Interface und ein Indexer für Aufgaben: Indexer already existed. No need for a default indexer, because the index does not have the same name as any attribute on the task, so that children of tasks will not find the field by acquisition
- [ ] Expose von `is_subdossier` und `is_subtask` im `@listing` endpoint

njohner:
This is deployed on dev

phgross:
Mit neuem Frontend getestet, i.O.

PR #5897, opened 2019-08-28
Add is_subdossier and is_subtask to listing endpoint.

njohner:
In the new gever UI, we want to display a nested structure but containing only the main dossiers, or the main tasks. For this we need to be able to filter by `is_subdossier` and `is_subtask` in the listing endpoint.

Both indexes already existed, and `is_subdossier` was also already in the solr schema. We therefore needed to
* Add `is_subtask` to solr indexes.
* Expose `is_subdossier` and `is_subtask` in the `@listing` endpoint.

Moreover because `is_subdossier` was a `BooleanIndex` while `is_subtask` a `FieldIndex`, I took the opportunity to change `is_subtask` to a `BooleanIndex`. We anyway needed to reindex it, in order to populate solr.

I also had to fix a bug in `CatalogContentListingObject` which breaks the acquisition chain when getting an attribute from the object. For this I overwrite the `__getattr__` method in `OpengeverCatalogContentListingObject`. The same issue had already been fixed for `RealContentListingObject` in `OpengeverRealContentListingObject` (https://github.com/4teamwork/opengever.core/blob/master/opengever/base/contentlisting.py#L178-L192).

This is for https://github.com/4teamwork/opengever.core/issues/5775

## Checkliste
- [x] Sind UpgradeSteps `deferrable`, oder können gewisse Schritte des Upgrades konditional ausgeführt werden? Deferrable should be fine here, only filtering by `is_subtask` will not work correctly until solr is reindexed.


opengever/api/tests/test_listing.py
```diff
@@ -342,9 +343,29 @@ def test_tasks_listing(self, browser):
...
+        all_tasks = browser.json['items']
+
+        # only subtasks
+        view = view + '&filters.is_subtask:record:list=true'
+        browser.open(self.dossier, view=view, headers=self.api_headers)
+        subtasks = browser.json['items']
+
+        self.assertTrue(len(all_tasks) > len(subtasks) > 0)
+        self.assertItemsEqual(
+            [item for item in all_tasks if item.get("is_subtask")], subtasks)
```
phgross: I'd check here for the real objects, if something is wrong in the `is_subtask` method or `index`, the test will still pass even a subtask is missing or a main_task is wrongly listed.

opengever/api/tests/test_listing.py
```diff
@@ -409,6 +430,25 @@ def test_filter_by_depth(self, browser):
...
+        all_dossiers = browser.json['items']
+
+        # only subdossiers
+        view = view + '&filters.is_subdossier:record:list=true'
+        browser.open(self.leaf_repofolder, view=view, headers=self.api_headers)
+        subdossiers = browser.json['items']
+
+        self.assertTrue(len(all_dossiers) > len(subdossiers) > 0)
+        self.assertItemsEqual(
+            [item for item in all_dossiers if item.get("is_subdossier")],
```
phgross: Same as above.

opengever/api/tests/test_listing.py
njohner: I added a check on the real objects. I also left the check on the items, which checks that the filtering works as expected (filters on the `is_subtask` index), whereas the check on the objects actually checks that the index itself works as expected.

opengever/api/tests/test_listing.py
njohner: Same as above too.

opengever/api/tests/test_listing.py
phgross: 👍

PR #5904, opened 2019-09-03
Always raise viewlet(-manager) rendering errors in dev mode.

deiferni:
This PR patches a tuple of exceptions that should be re-raised in `BaseOrderedViewletManager` so that all exceptions are re-raised. The patch is only applied for local development, i.e. when the env variable `IS_DEVELOPMENT_MODE` is enabled.

With the patch active rendering of a page is aborted when a viewlet or viewlet manager fails to render. Without this patch a placeholder markup snippet is rendered indicating rendering errors, e.g. `error while rendering plone.failing.viewlet`. That snippet will then be cached happily by `plone.app.caching` and can remain in place even after the error has been fixed. After spending way too much time attempting to debug a "failing" viewlet i'd prefer to have this patch for local development.

## Checkliste

### DIFF
--- a/opengever/base/monkey/patches/development.py
+++ b/opengever/base/monkey/patches/development.py
@@ -1,0 +1,19 @@
+from opengever.base.monkey.patching import MonkeyPatch
+import os
+
+
+class PatchBaseOrderedViewletManagerExceptions(MonkeyPatch):
+    """Patch exceptions so that all are raised during viewlet rendering.
+
+    Only apply the patch in development mode.
+    """
+    def __call__(self):
+        if not os.environ.get('IS_DEVELOPMENT_MODE', False):
+            return
+
+        from plone.app.viewletmanager.manager import BaseOrderedViewletManager
+
+        new_exceptions = (Exception,)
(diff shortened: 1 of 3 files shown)


opengever/base/monkey/patches/development.py
```diff
@@ -0,0 +1,19 @@
...
+import os
+
+
+class PatchBaseOrderedViewletManagerExceptions(MonkeyPatch):
+    """Patch exceptions so that all are raised during viewlet rendering.
+
+    Only apply the patch in development mode.
+    """
+    def __call__(self):
+        if not os.environ.get('IS_DEVELOPMENT_MODE', False):
```
lukasgraf: This should also be `AND`ed with whether the instance is in debug mode or not, IMHO.
See https://github.com/4teamwork/opengever.core/blob/master/opengever/base/monkey/patches/exception_formatter.py#L44-L45 for example.

lukasgraf: Nevermind, I had a brain fart.

PR #5921, opened 2019-09-05
Use base responses for tasks

elioschmutz:
Dieser PR verwendet die neue Response Implementation (PR #5893 ) für Tasks.

⚠️ die JSON-Repräsentation von Responseobjekten für Tasks ändert mit diesem PR. Muss dies irgendwo speziell vermerkt werden, ausser in den Dokus?

Dieser PR migriert die alten Responses. Die alten Responses werden dabei nicht gelöscht. Es wird ein Backup unter einem neuen Key erstellt. Ansonsten haben wir keine Möglichkeit der Nachverfolgung falls etwas schief geht.

## Checkliste

Related Issue #5813


opengever/base/response.py
```diff
@@ -53,12 +53,13 @@ def __init__(self, context):
     def __len__(self):
         return len(self.list())
 
-    def add(self, response):
+    def add(self, response, response_id=None):
```
njohner: Ok I understand this is needed for the migration, but it's a bit misleading, because we might actually not use that `response_id` in the end, if it is already taken. Why not pass the an optional `creation_time` argument instead? Makes more sense to me to be able to set the response creation time than its id.

opengever/base/response.py
```diff
@@ -187,6 +187,9 @@ def add_change(self, field_id, before, after, field_title=''):
     def add_related_item(self, relation):
         self.related_items.append(relation)
 
+    def set_text(self, text):
+        text = safe_unicode(text)
+        self.text = text if text else ''
```
njohner: `else u''`

elioschmutz:
@njohner thanks for your review! I won't go into specific feedbacks because I think a lot of the points have been misunderstood or will change anyway. 

# General Feedback
In general, the main problem is the fact, that we have an implementation for the old ui and an implementation for the new ui. In this PR i do not change the old implementation. The only thing I'm doing is changing the Response-Container to the new implementation. With this change, we get access to the restapi for adding a comment response object which is the initially goal of this task.

But beside the container-implementation, the old ui comes with a lot of extra code for handling this responses in the old ui. I.e. all the Description-Objects or a lot of extra response-fields like the `rendered_text` attribute (also this false or inconsistent written fields like `added_object` or `relatedItems` are only used by the old ui. This code is very hard to understand or to update and have a lot of side-effects if touching it.

With my implementation I do not really touch this old implementation. But I had to update some parts of the code because the old implementation was not really decoupled and everything was interlocked together.

# Options
In my point of view, we have this options:

1. We can now rethink the whole task response implementation. We rewrite every task-transition form, every task description and the handling of this responses with tasks. We also have to rethink how this response object should look like, i.e. we now have a `response_type`. A transition could be a special `response_type` or we could introduce a `response_sub_type`. Also changing the implementation to not use `rendered_text`, `added_objects` or `relatedItems` would come with a great impact to the current implementation.

Most of the work will be remove as soon as we move to the new UI.

2. We do not reimplement the whole implementation. We just create the new response-Object `TaskResponse` with the current fields and implementation to remove the task-specific field from the generic `Response` object. But we do not touch the whole task-response mechanism.
Using this option would save a lot of extra-work.

Perhaps we have to do another "cleanup"-migration if we remove the old ui to remove fields like `rendered_text`. But this way seems to be much more efficient.

# Conclusion

We can rethink a really big topic and rewrite a lot of code to make it correctly right now. But, everything we are refactoring on the old implementation will be removed as soon as we change to the new UI.

I do not change the current behavior or implementation of the old implementation. I just make it compatible with the new implementation.

I agree with introducing a new response-object `TaskResponse` as proposed by @njohner . But I do not agree to completely refactor the whole implementation of task responses (Attributes, Functions, Description objects...).

I prefer option 2 with an additional refactoring as soon as we move completely to the new UI.

# Further steps

@4teamwork/gever please take a look at this PR that we can discuss it tomorrow at the meeting.

opengever/task/response_description.py
```diff
@@ -38,6 +38,35 @@ def get(cls, response=None, transition=None):
 
         return description(response)
 
+    @classmethod
+    def get_added_objects(cls, response):
```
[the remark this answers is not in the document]
elioschmutz: This is required due to the current implementation of the description objects. This method will be called by another `classmethod`. See https://github.com/4teamwork/opengever.core/pull/5921/files/a13a1525e276a4d4dd7991c90f8ce3af9d586d9d..03aebf88466f7cf6991c16b4a65c3109dd1e9e2a#diff-efd7c639c1dd8573fc2e139471e5fec9R21

elioschmutz:
@phgross Unfortunately you and @njohner where missing at the meeting.

I discussed it in the Daily with @lukasgraf but we need feedback from someone who knows the the task response code.

@phgross Could you please give feedback to this discussion? Which way shall we go?

phgross:
As discussed I would go with Option 2.

opengever/base/response.py
elioschmutz: 👍 I updated the function to generate the `response_id` based on the `creation` time of the response.

Passing the `response_id` is now obsolete.

opengever/core/upgrades/20190919104620_migrate_old_task_responses_to_new_responses_implementation/upgrade.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def __call__(self):
+        self.install_upgrade_profile()
+        for task in self.objects({'object_provides': ITask.__identifier__},
+                                 'Migrate task responses to new responses implementation'):
+            self.migrate_responses(task)
+
+    def migrate_responses(self, task):
+        container = IResponseContainer(task)
+        old_container = OldIResponseContainer(task)
```
phgross: Is it really necessary, to use the "old" responsecontainer interface for the migration? Because, as soon we would remove the interface, a bigger upgrade which combines multiple releases would not longer be possible.

elioschmutz: @phgross no, it is not necessary to keep the respnose-container. I could directly access the annotations-storage. But we definitely have to keep the old Response class.

Should I update it anyway?

phgross: So the upgradestep will fail, as soon the ReponseClass has been removed, right?

elioschmutz: yes...

elioschmutz: There is a comment for each deprecated class: https://github.com/4teamwork/opengever.core/pull/5921/files#diff-d47977b135b2e68e74d84ba829f80b7cR49

PR #5924, opened 2019-09-09
Handle creation of new proposals over the REST-API.

njohner:
The Add form for proposals contained some logic and specific steps needed for the creation of a `Proposal`, such as the creation of a `Document` (the proposal document). Because of that, creation of a `Proposal` through the REST-API was not possible (or rather broken).

In this PR we make it possible to create proposals through the REST-API. For that we:
* Move proposal creation logic from the AddForm to an event handler
* Add a custom deserializer for proposals (`DeserializeProposalFromJson`) 
    * Make sure a template is set for the proposal document
    * Store the template information on the request so that it can be retrieved in the event handler
* Adapt proposal builders, fixture, tests and example content as we now always need to specify a template in order to create a `Proposal`

These changes are necessary to allow working with `Proposal`s in the new UI.

This is part of https://github.com/4teamwork/opengever.core/issues/5799

## Checkliste

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2019.4.0rc3 (unreleased)
 ------------------------
 
+- Handle creation of new proposals over the REST-API. [njohner]
 - Introduce plone proposal workflow, provide api support for proposal workflow transitions. [deiferni]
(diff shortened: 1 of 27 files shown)


deiferni:
So this one is dead as it is replaced with https://github.com/4teamwork/opengever.core/pull/5950?

Issue #5922, opened 2019-09-09
UnicodeDecodeError in listing endpoint, when using a textfilter.


phgross:
Steps to reproduce:
 - Use the textfilter in the new UI and type in umlauts.

https://sentry.4teamwork.ch/sentry/onegov-gever/issues/38711/

```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 2: ordinal not in range(128)
(2 additional frame(s) were not displayed)
...
  File "ZPublisher/Publish.py", line 48, in call_object
    result=apply(object,args) # Type s<cr> to step into published object.
  File "apps/eggs/plone.rest-1.2.0-py2.7.egg/plone/rest/service.py", line 23, in __call__
    return self.render()
  File "apps/eggs/plone.restapi-4.3.0-py2.7.egg/plone/restapi/services/__init__.py", line 21, in render
    content = self.reply()
  File "src/opengever.core/opengever/api/listing.py", line 276, in reply
    depth, facets)
  File "src/opengever.core/opengever/api/listing.py", line 377, in solr_results
    pattern.format(term=escape(t)) for t in term.split()]

UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 2: ordinal not in range(128)
```

PR #5927, opened 2019-09-11
Fix unicode error in listing endpoint.

njohner:
Solr test actually catches that encoding issue, even if it's a not so useful mock test.

For https://github.com/4teamwork/opengever.core/issues/5922

## Checkliste


deiferni:
hijacked and rebased. :shipit:.

PR #5932, opened 2019-09-13
Validate task deadline modification through the rest-api

elioschmutz:
Validate task deadline modification through the rest-api if the user uses the same deadline as already set on the task.

Issuer: #5814 

## Checkliste


opengever/base/transition.py
```diff
@@ -119,3 +120,8 @@ def _deserialize_values(self, transition_params, collect_errors=False):
                 errors.append({'error': error, 'message': error.message})
 
         return values, errors
+
+    def _validate_field(self, name, field, value, errors):
```
phgross: Is there a reason, why you don't use a regular validator, as we uses for the responsible team (see `NoTeamsInProgressStateValidator`)

So we would not need to introduce an additional validation method.

elioschmutz: 👍 thanks for the hint! I updated the pr to use a regular validator.

PR #5940, opened 2019-09-18
Sync proposal workflow state when committee-side proposal state changes.

deiferni:
This PR is part of #5799. We always keep the proposal plone object workflow state in sync with its submitted side. So when the state on the submitted/committee side of the proposal changes the proposal state in the dossier is kept in sync. The states were introduced in https://github.com/4teamwork/opengever.core/pull/5919.

With this PR we:
- use all proposal states in the plone workflow, actions and guards are updated. the new states are only reachable internally.
- amend the transition controller for proposals with guards for all states
- refactor the way a submitted proposals communicates with a proposal. instead of multiple separate requests we now use only one request when a state change is involved.

## Checkliste


deiferni:
I have fixed the commit message. :shipit:.

PR #5942, opened 2019-09-18
Render discreet workflow transition buttons and show warning/info messages on proposal.

deiferni:
This PR is part of #5799. We introduce discreet workflow transition buttons for cancel and submit, when those transitions are not available due to proposal state or committee state. We also introduce info/warning messages.

With this PR we:
- Render discreet submit and cancel button when the proposal document is checked out, and also show an info message on the proposal overview.
- Render a discreet submit button when the committee has been deactivated and also show a warning message on the proposal overview.

Screenshot 2019-09-19 at 18 14 01

## Checkliste


opengever/meeting/browser/templates/proposalinfoviewlet.pt
```diff
@@ -9,4 +11,8 @@
         </tal:block>
     </dd>
   </dl>
+  <dl class="portalMessage warning" tal:condition="show_committee_inactive_warning">
+    <dt i18n:translate="" i18n:domain="plone">Warning</dt>
+    <dd i18n:translate="message_committee_inactive_warning">The committee is no longer active.</dd>
```
phgross: Is it "hard work" to include the committee title? Imho it would help to understand the message. 

`Das Komitee Finanzkommission wurde deaktiviert.`

deiferni: you're right. i have amended the committee title and updated the screenshot. the committee name is "Deaktiviert" so don't be confused 😉.

deiferni:
re-requesting review, best-case this can be merged tomorrow.

PR #5952, opened 2019-09-25
Extend @solrsearch endpoint

buchi:
- Return an object instead of a list
- Return total item count
- Add support for contentlisting attributes (e.g. getURL)
- Return snippets (aka highlighting)
- Return facets


opengever/api/solrsearch.py
```diff
@@ -1,45 +1,147 @@
...
+    '@id',
+    '@type',
+    'title',
+    'description',
+    'review_state',
+])
+
+# Field name -> (Solr field, accessor)
+FIELD_MAPPING = {
+    "@id": ("path", "getURL"),
```
njohner: These would actually be good candidates to use `namedtuple`s (https://docs.python.org/2/library/collections.html#collections.namedtuple). Just saying because in the search/listing endpoints we have these mappings all over the place, and it does not make the code very readable

opengever/api/solrsearch.py
```diff
@@ -1,45 +1,147 @@
...
+        else:
+            filters = []
+
+        if 'start' in params:
+            start = safe_int(params['start'])
+            del params['start']
+        else:
+            start = 0
+
+        if 'rows' in params:
```
njohner: I guess you could also use 
```
rows = min(safe_int(params.get('rows'), 25), 1000)
if rows in params:
    del params['rows']
```
Not sure whether it would be more readable, so as you prefer.

opengever/api/solrsearch.py
```diff
@@ -1,45 +1,147 @@
...
+        resp = solr.search(
+            query=query, filters=filters, start=start, rows=rows, sort=sort,
+            **params)
+
+        docs = OGSolrContentListing(resp)
+        items = []
+        for doc in docs:
+            item = {}
+            for field in requested_fields:
+                if field.startswith("_"):
```
njohner: Can you add a comment why we skip these?

opengever/api/solrsearch.py
njohner: But that's really just a suggestion, it can remain as is too...

njohner:
Is there a ticket for this PR?

opengever/api/solrsearch.py
buchi: I'll leave it as is for now.

opengever/api/solrsearch.py
buchi: IMO this doesn't make a difference in readability. But performance wise subscription is faster. Thus I prefer my variant.

opengever/api/solrsearch.py
buchi: Seems pretty obvious to me. It's a common pattern to prefix private attributes with an underscore. As we basically allow access to all attributes, we want to make sure that we at least don't leak any attributes that were marked explicitly as private.
I've added a comment in the code.

PR #5964, opened 2019-10-02
Support the bundle-import of mails in the msg format.

phgross:
Currently the bundle import of GEVER had no support for importing *.msg mails. With this changes it now also support the import of those mails, similar than GEVER does when uploading them via dnd (quickupload):
 - Converting the *.msg to an *.eml with `msgconvert` and stores it as `message`
 - Stores the `*.msg` as original_mesasge

For https://github.com/4teamwork/opengever.koeniz/issues/78

## Checkliste

### DIFF
--- a/opengever/bundle/loader.py
+++ b/opengever/bundle/loader.py
@@ -4,6 +4,7 @@ from collections import OrderedDict
 from datetime import datetime
 from jsonschema import FormatChecker
 from jsonschema import validate
+from opengever.document.document import MAIL_EXTENSIONS
 from pkg_resources import resource_filename as rf
 import codecs
 import json
@@ -212,7 +213,7 @@ class ItemPreprocessor(object):
         if filepath is None:
             return False
 
-        return os.path.splitext(filepath)[-1] in ('.eml', '.p7m',)
+        return os.path.splitext(filepath)[-1] in MAIL_EXTENSIONS
 
     def _strip_extension_from_title(self):
         """Strip extension from title if present.
--- a/opengever/bundle/sections/fileloader.py
+++ b/opengever/bundle/sections/fileloader.py
@@ -4,6 +4,7 @@ from collective.transmogrifier.utils import defaultMatcher
 from collective.transmogrifier.utils import traverse
 from ftw.mail.mail import IMail
 from mimetypes import guess_type
+from opengever.base.transforms.msg2mime import Msg2MimeTransform
 from opengever.bundle.sections.bundlesource import BUNDLE_KEY
 from opengever.bundle.sections.bundlesource import BUNDLE_PATH_KEY
 from opengever.document.document import IDocumentSchema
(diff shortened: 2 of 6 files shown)


opengever/bundle/sections/fileloader.py
```diff
@@ -167,12 +168,35 @@ def add_namedblob_file(self, abs_filepath, field, obj):
...
             mimetype = 'application/octet-stream'
 
-        with open(abs_filepath, 'rb') as f:
-            namedblobfile = field._type(
-                data=f.read(),
-                contentType=mimetype,
-                filename=self.get_filename(abs_filepath))
-            setattr(obj, field.getName(), namedblobfile)
+        # Special handling for outlook mails (*.msg)
+        if mimetype == 'application/vnd.ms-outlook' and field.getName() == 'message':
```
deiferni: we  already have an `is_mail` method on this class, we should also check if we are on a mail content type.

opengever/bundle/loader.py
```diff
@@ -212,7 +212,7 @@ def _is_mail_type(self, filepath):
         if filepath is None:
             return False
 
-        return os.path.splitext(filepath)[-1] in ('.eml', '.p7m',)
+        return os.path.splitext(filepath)[-1] in ('.eml', '.p7m', '.msg')
```
deiferni: could you re-use https://github.com/4teamwork/opengever.core/blob/34c7ace4c267395e4b9ef7b9ef4ba295521f8049/opengever/document/document.py#L52

opengever/bundle/sections/fileloader.py
```diff
@@ -167,12 +168,35 @@ def add_namedblob_file(self, abs_filepath, field, obj):
...
+                    data=Msg2MimeTransform().transform(f.read()),
+                    contentType='message/rfc822',
+                    filename=filename)
+                setattr(obj, field.getName(), namedblobfile)
+
+                # original_message
+                f.seek(0)
+                namedblobfile = IOGMail['original_message']._type(
+                    data=f.read(),
+                    contentType='message/rfc822',
```
deiferni: shouldn't this remain message/rfc822? do you have to specify the mime-type here? If you can i would prefer you'd omit it so it is chosen by mime-type registry based on filename.

opengever/bundle/sections/fileloader.py
```diff
@@ -167,12 +168,35 @@ def add_namedblob_file(self, abs_filepath, field, obj):
...
+
+                # eml using Msg2Mime transform
+                namedblobfile = field._type(
+                    data=Msg2MimeTransform().transform(f.read()),
+                    contentType='message/rfc822',
+                    filename=filename)
+                setattr(obj, field.getName(), namedblobfile)
+
+                # original_message
+                f.seek(0)
```
deiferni: if you assign the output of `f.read()` there is no need to reset the file position and you don't need to read the file twice.

opengever/bundle/sections/fileloader.py
phgross: Hmm what I really want to know here is, if there is a `*.msg` file send as `message` (because the special handling, is only necessary when the msg is send as primary file).

an additional check for `is_mail` does not help here imho.

phgross:
@deiferni updated ...

PR #5979, opened 2019-10-09
Downpin ftw.recipe.solr to fix log4j config.

### LINKED ISSUE #5978: Also downgrade ftw.recipe.solr after downgrading Solr
Starting Solr on a fresh buildout from `master` currently fails:

```
bin/solr fg

log4j:WARN Continuable parsing error 20 and column 16
log4j:WARN Document root element "Configuration", must match DOCTYPE root "null".
log4j:WARN Continuable parsing error 20 and column 16
log4j:WARN Document is invalid: no grammar found.
log4j:ERROR DOM element is - not a <log4j:configuration> element.
log4j:WARN No appenders could be found for logger (org.eclipse.jetty.util.log).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
```

This is because the [setup of the log4j config needed to be updated in `ftw.recipe.solr`](https://github.com/4teamwork/ftw.recipe.solr/pull/4) to be compatible with Solr 8.x. That change in the setup of the log4j config is only compatible with Solr >= 7.4 though, but we downgraded back to 7.3.x

We therefore need to also downgrade `ftw.recipe.solr` to `1.2.1`.

`solrconfig.xml` also [contains some changes](https://github.com/4teamwork/opengever.core/commit/fb9ed78bb#diff-4a472ce962966d67e0691fe91257e660R88) that were introduced with the bump to Solr 8.1.x that have not been reverted yet it seems.

deiferni:
Solr configuration in newer recipe versions are already optimized for solr 8.x. Fortunately "only" log4j initialization was affected, so solr itselft worked correctly but could not log. Downpinning will  fix that issues. Re-running buildout should fix any issues for affected deployments as it updates configuration.

Fixes #5978.


lukasgraf:
Hijacked to fix Solr Testserver Selftest failure.

Issue #5905, opened 2019-09-03
Missing metadata/properties in REST API GET of document


buchi:
Harvest: `Projekt - Umsetzung Neues User Interface GEVER 2019/2020` - `Entwicklung`

The following metadata or properties are needed by the frontend and should be returned by the REST API GET endpoint for documents:
- `checked_out`
- `is_locked`
- `containing_dossier`
~- `containing_subdossier`~
- `dossier_review_state`
- `trashed`
- `is_shadow_document`
- `get_current_version_id` ->  `current_version`

Used for the document detail view.

Notes:
- Consider whether we should namespace these properties (additional properties which are not DX fields)

phgross:
Fachliches Review: 👍

PR #5986, opened 2019-10-11
Provide additional attributes in the API representation of documents and mails

phgross:
For #5905 

We also wondered during the Scrum meeting if it would make sense to namespace the additional attributes or use the `metadata_fields` behavior. I decided against it for the following reasons:
 - The document representation already contains "ready-only" metadata which is not part of a schema on the top level
 - Non schema keys are ignored by the restapi on Patch and Post requests
 - `metadata_fields` is currently not supported for the representation of a context

## Checkliste

- [ ] Aktualisierung Dokumentation vorhanden/nötig? Nein, ich bin der Meinung das die zusätzlichen Werte nicht zusätzlich erwähnt werden müssen.

### DIFF
--- a/opengever/api/tests/test_document.py
+++ b/opengever/api/tests/test_document.py
@@ -48,6 +48,40 @@ class TestDocumentSerializer(IntegrationTestCase):
         self.assertEqual(browser.status_code, 200)
         self.assertEqual('.msg', browser.json.get(u'file_extension'))
 
+    @browsing
+    def test_contains_additional_metadata(self, browser):
+        self.login(self.regular_user, browser)
+
+        # The helpers used for the api are already tested, so no
+        # need to repeat these here.
+        self.checkout_document(self.subdocument)
+
+        browser.open(self.subdocument, headers={'Accept': 'application/json'})
+
+        self.assertEqual(self.regular_user.id, browser.json['checked_out'])
+        self.assertFalse(browser.json['is_locked'])
+        self.assertEqual(u'Vertr\xe4ge mit der kantonalen Finanzverwaltung',
+                         browser.json['containing_dossier'])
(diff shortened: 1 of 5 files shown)


opengever/api/tests/test_document.py
```diff
@@ -48,6 +48,40 @@ def test_document_serialization_contains_file_extension(self, browser):
...
+        self.assertEqual(u'2016', browser.json['containing_subdossier'])
+        self.assertFalse(browser.json['trashed'])
+        self.assertFalse(browser.json['is_shadow_document'])
+        self.assertFalse(0, browser.json['current_version_id'])
+
+    @browsing
+    def test_additional_metadata_for_mails(self, browser):
+        self.login(self.regular_user, browser)
+
+        browser.open(self.subdocument, headers={'Accept': 'application/json'})
```
njohner: `self.mail` probably?

opengever/document/tests/test_base.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def test_is_trashed(self):
+        self.login(self.regular_user)
+        self.assertFalse(self.base_document.is_trashed)
+
+        self.trash_documents(self.base_document)
+        self.assertTrue(self.base_document.is_trashed)
+
+
+class TestBaseDocumentMails(TestBaseDocument):
```
njohner: That's nice!

phgross:
The `containing_dossier` and `containing_subdossier` indexers a registered and implemented for every IDexterityContent, so the indexers work also for documents and mails. As discussed it does not make much sense to register a specific one for documents which does the same as the general one.

njohner:
Sorry, didn't realise I should have another look...

Issue #5974, opened 2019-10-08
XLS Report, Zip Export, PDF Export should also handle relative_paths


phgross:
The listing actions `Als Zip-Datei exportieren`, `Auswahl exportieren` and `Auswahl drucken (PDF)` should also work with realtive paths passed in the `paths` parameter.

Views:
- `zip_selected` (Als Zip-Datei exportieren, Documents Tab)
- `document_report` (Auswahl exportieren, Documents Tab)
- `pdf-dossier-listing` (Auswahl drucken, Dossiers Tab)

Should fix https://github.com/4teamwork/gever-ui/issues/479

lukasgraf:
Sentry Issue for ZIP-Export:  https://sentry.4teamwork.ch/sentry/onegov-gever/issues/9124/

phgross:
Fachliches Review: 👍

PR #5994, opened 2019-10-17
Add support for relative paths for views that take paths:list parameter

lukasgraf:
This adds support for pseudo-relative paths for these four views that take `paths:list` parameters:

- `zip_selected` (Als Zip-Datei exportieren, Documents Tab)
- `document_report` (Auswahl exportieren, Documents Tab)
- `dossier_report` (Auswahl exportieren, Dossiers Tab)
- `pdf-dossier-listing` (Auswahl drucken, Dossiers Tab)

Pseudo-relative in this context means paths that don't start with the Plone site (so are relative to it), but still have a trailing slash).

This is required for these views because the new gever-ui sends paths that aren't prefixed with the Plone site.

Addresses #5974

## Checkliste

### DIFF
--- a/opengever/base/utils.py
+++ b/opengever/base/utils.py
@@ -4,6 +4,7 @@ from plone import api
 from Products.CMFCore.utils import getToolByName
 from xml.sax.saxutils import escape
 from zope.component import getMultiAdapter
+from zope.component.hooks import getSite
 import hashlib
 import json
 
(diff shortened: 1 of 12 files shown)


opengever/testing/integration_test_case.py
```diff
@@ -669,9 +669,24 @@ def get_allowed_roles_and_users_for(self, obj):
...
+        """
         return {
             'paths:list': ['/'.join(obj.getPhysicalPath()) for obj in objects]}
 
+    def make_pseudorelative_path_param(self, *objects):
+        """Build a paths:list request parameter with pseudo-relative paths.
+
+        Pseudo-relative in this context means that the paths don't start with
+        the Plone site (they are relative to it), but they still start with
+        a leading slash.
```
njohner: How about saying here that this is what we get from the new frontend?

opengever/base/utils.py
```diff
@@ -174,3 +175,35 @@ def make_persistent(data):
...
+    """If request contains a paths:list param, rewrite these paths so
+    they're always absolute (start with the Plone site).
+
+    (Modifies the request in place)
+
+    This is required for a couple views that use a paths:list parameter and
+    break with the new gever-ui, which sends "pseudo-relative" paths.
+
+    Pseudo-relative in this context means paths that don't start with the
+    Plone site (so are relative to it), but still have a trailing slash).
```
njohner: I think you mean `leading slash` and not `trailing slash`?

lukasgraf:
Updated, thanks for the review :)

PR #5997, opened 2019-10-18
Fix mail deserialization for mails uploaded through tus-upload.

njohner:
When uploading mails through drag&drop in the new `gever-ui`, we use the `tus-upload` endpoint. 

**Problem:**
This endpoint was not working for E-mails, the reason being that the deserialized the `message` field in our custom `DeserializeMailFromJson`, to check whether we need to move the message to the `original_message` field, and then deserialize the field again (standard deserialization to fill the value into the object). Problem is that deserialization of the field consumes the file in the `tus-upload`, so the second deserialization was failing. 

**Solution:**
We invert the order, first we deserialize the field to write the data to the object, then, if necessary, we move that file to the `original_message` field.

For https://github.com/4teamwork/gever-ui/issues/539

## Checkliste

### DIFF
--- a/opengever/api/tests/test_deserializer.py
+++ b/opengever/api/tests/test_deserializer.py
@@ -1,0 +1,225 @@
+from ftw.testbrowser import browsing
+from opengever.testing import IntegrationTestCase
+import os
+from base64 import b64encode
+from opengever.testing.assets import path_to_asset
+
+
+def _base64_str(s):
+    if not isinstance(s, bytes):
+        s = s.encode("utf-8")
+    s = b64encode(s)
+    if not isinstance(s, str):
+        s = s.decode("utf-8")
+    return s
+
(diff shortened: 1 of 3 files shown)


opengever/api/tests/test_deserializer.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def _prepare_metadata(filename, content_type, _type=None):
+    return "filename {},content-type {}{}".format(
+        _base64_str(filename), _base64_str(content_type), ', @type ' + _base64_str(_type) if _type else ''
+    )
+
+
+msg_mimetype = u'application/vnd.ms-outlook'
+eml_mimetype = u'message/rfc822'
```
deiferni: i'd also add a test for p7m mails here.

PR #6001, opened 2019-10-21
Implement schema-intent concept

lukasgraf:
Implements the schema-intent concept according to #5973:
- New intent-aware `@schema` endpoint
- Customized, intent-aware`@sources`, `@querysources` and `@vocabularies` endpoints

Todo:
- [x] Slap the `IDuringContentCreation` interface on the request during add-intent scenarios
- [x] Tests. Especially one that tests default values with `restricted_vocab.py`
- [x] Check whether the expansion support for `@types` needs to be added to `@schema` as well
- [x] Verify that `@types` responses aren't broken in non-backwards compatible ways
~- [ ] Documentation~

Obviously the frontend also needs to be adapted to use these new endpoints in their intent-aware ways.


opengever/api/schema/vocabularies.py
```diff
@@ -0,0 +1,61 @@
...
+            self.intent = 'edit'
+            portal_type = None
+            vocab_name = self.params[0]
+        elif len(self.params) == 2:
+            # Add intent
+            # - context is the container where the object will be created
+            # - portal_type is the type of object to be created
+            # - schemata need to be determined via portal_type
+            self.intent = 'add'
+            portal_type = self.params[0]
```
njohner: portal_type is unused here

njohner:
Current stand:
- Added `IDuringContentCreation` interface in the case of `add` intent
- Added tests for vocabularies, sources and querysources endpoints.

Then there were some issues with fields of the type type of the keywords which have one more level of nesting (`field.value_type` to get to the vocabulary):
- Querying such fields would fail. I fixed this in the commit `Fix queryources and sources endpoint for keywords vocabulary.`
- For the `@schema` endpoint, these fields do not get treated correctly:
  - I wrote a failing test that shows at least part of the problem
  - In the `@id` in the schema, the name of the field is missing. This is because the `IJsonSchemaProvider` adapts the `field.value_type` (https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/types/adapters.py#L192-L195), which has no knowledge of the field name (https://github.com/4teamwork/opengever.core/blob/lg-schema-intent/opengever/api/schema/adapters.py#L79-L80), hence missing in the url
  - In the response of the schema endpoint, for such fields, the `@id` is nested one level deeper in the json response (`properties/fieldname/items/querysource/@id` instead of `properties/fieldname/vocabulary/@id`)

opengever/api/schema/adapters.py
```diff
@@ -38,6 +55,88 @@ def additional(self):
...
         return result
 
+# These IJsonSchemaProviders below are customized so that we can retain
+# a link to the parent field. We do this so that we can use the parent field's
+# name to render URLs to sources on anonymous inner value_type Choice fields.
+
+
+@adapter(ICollection, Interface, IOpengeverBaseLayer)
+@implementer(IJsonSchemaProvider)
+class GEVERCollectionJsonSchemaProvider(CollectionJsonSchemaProvider):
```
njohner: 😅 When I was looking at this issue I was thinking to myself `I don't want to overwrite all  these adapters, there must be some easier way...`.

lukasgraf:
This one is ready for a final review. We decided to not duplicate efforts for documentation, and create it when this finally moves upstream into `plone.restapi`. Also, the high-level concept is already documented quite well in the referenced issue.

lukasgraf:
Updated

PR #6002, opened 2019-10-22
Allow adding favorites by UID parameter via favorites endpoint.

phgross:
The oguid parameter is still supported. In addition the `listing` endpoint now also supports the UID as a column.

Used for https://github.com/4teamwork/gever-ui/issues/490

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,9 @@ Changelog
 2019.4.0rc4 (unreleased)
 ------------------------
 
+- Allow adding favorites by UID parameter via favorites endpoint. [phgross]
+- Add UID to listing endpoints supported fields. [phgross]
+- Allow adding favorites with UID parameter via favorites endpoint. [phgross]
 - Bump ftw.keywordwidget version to fix missing titles on terms. [njohner]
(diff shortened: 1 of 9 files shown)


docs/HISTORY.txt
```diff
@@ -4,6 +4,7 @@ Changelog
 2019.4.0rc4 (unreleased)
 ------------------------
 
+- Allow adding favorites by UID parameter via favorites endpoint. [phgross]
```
njohner: I'd say `with` instead of `by`

opengever/api/favorites.py
```diff
@@ -99,15 +106,24 @@ def reply(self):
...
-        oguid = Oguid.parse(data.get('oguid'))
-        obj = oguid.resolve_object()
+        if data.get('oguid'):
+            oguid = Oguid.parse(data.get('oguid'))
+            obj = oguid.resolve_object()
+        elif data.get('uid'):
+            uid = data.get('uid')
+            catalog = api.portal.get_tool('portal_catalog')
+            brains = catalog(UID=uid)
+            if len(brains):
```
njohner: isn't `brains` a list? I'd use `if brains` or `if len(brains) > 0`

Issue #5901, opened 2019-09-02
Listing endpoint überarbeiten und catalog Support entfernen.


phgross:
Harvest: `[3.2.2.1] OneGov GEVER` - `Softwarepflege (Refactoring/Bug)`

Follow up aus vergangenen sprint plannings: der listing endpoint benötigt refactoring.

Zu dem kann der catalog Support entfernen werden., der Listing Endpoint muss neu nur noch für Deployments mit aktiviertem Solr funktionieren

Todos

- [ ] Refactoring
- [x] Remove catalog implementation

njohner:
For refactoring part, after quick discussion with @lukasgraf I'd propose
- use `namedtuple`s for the items in `FIELDS` or maybe drop `FIELDS` altogether and write this in a more object-oriented way. This would allow to group the methods used for certain fields together, like translations, dealing with dates, and so on...
- Maybe only keep the mapping in `FIELDS` when it is non-standard?
- extract some of the functionality to methods and add docstrings.

njohner:
Done, but wait to test with new solr tests.

PR #6009, opened 2019-10-29
Remove catalog support for @listing endpoint

elioschmutz:
Issue #5901 

This PR makes the `@listing` endpoint only available with an activated solr-feature.

## Checkliste


docs/HISTORY.txt
```diff
@@ -4,6 +4,7 @@ Changelog
 2019.4.0rc5 (unreleased)
 ------------------------
 
+- The @listing endpoint is no longer working without solr. [elioschmutz]
```
njohner: I'd rather have the CL entry say what was done, not what is working or not:
`- Remove catalog support for @listing endpoint. [elioschmutz]`

elioschmutz: 👍

Issue #6011, opened 2019-10-29
Webactions API: Patch fails due to userid (is unicode, should use bytestring)


sebastianmanger:
Harvest: `[3.2.2.1] OneGov GEVER` - `Softwarepflege (Refactoring/Bug)`

**Description**

Patching of existing webaction via API (`@webactions/<action_id>`) fails due to an invalid owner: `BadRequest: [('owner', WrongType(u'webaction.manager', <type 'str'>, 'owner'))]`

When `patch`ing a webaction, the add-method of the WebActions storage class (`opengever/webactions/storage.py:95`) returns a unicode string, which is not valid as the field is defined as `ASCIILine` (see `opengever/webactions/schema.py:190`).

This seems to be caused by this line: `api.user.get_current().getId()`:

```
(Pdb) api.user.get_current().getId()
u'webaction.manager'
(Pdb) api.user.get('webaction.manager').getId()
'webaction.manager'
```

**Context**
This problem occured when implementing a feature for RIS: https://github.com/4teamwork/ris/issues/1561 

The RIS-Application needs to create and update webactions per API. 

**Fix**
I'll provide a PR to solve this issue on the `ris` branch. The GEVER team will discuss about how to approach this, or if a fix is relevant at all. If no unicode-strings are set as user-id on production environments, this is not an issue.
Until this is clarified by GEVER, the mentioned feature for RIS will not be shipped so that no invalid data can be created when adding webactions.

sebastianmanger:
Note: Webactions were already created (before this was discovered) on https://ris.onegovgever.ch/kr/

PR #6012, opened 2019-10-29
Cast the userid to bytestring so that validation to ASCIILine passes

sebastianmanger:
In some test cases, a unicode string is returned which raises an error when patching webactions with valid data due to invalid, but generated unicode data for the owner field.

This is a quick, but not complete fix so that the RIS team can finish implementing changes. Changelog entry, tests and more may follow depending on how the GEVER team decieds to approach this issue.

Relates to  #6011 (but does not close it).

### DIFF
--- a/opengever/webactions/storage.py
+++ b/opengever/webactions/storage.py
@@ -92,7 +92,9 @@ class WebActionsStorage(object):
         new_action = PersistentMapping(action_data)
         self._actions[action_id] = new_action
 
-        userid = api.user.get_current().getId()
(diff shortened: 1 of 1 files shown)


opengever/webactions/storage.py
```diff
@@ -92,7 +92,7 @@ def add(self, action_data):
         new_action = PersistentMapping(action_data)
         self._actions[action_id] = new_action
 
-        userid = api.user.get_current().getId()
+        userid = str(api.user.get_current().getId())
```
njohner: Maybe just add a comment in the code that this is a workaround and why

PR #6020, opened 2019-11-01
Update the usersettings-serializer: A manager has always seen all scr…

elioschmutz:
Issue: https://github.com/4teamwork/gever-ui/issues/594
Required for frontend issue: https://github.com/4teamwork/gever-ui/pull/655

The `user_settings` provides the seen tours of a user. But pure Plone-Users (like the `zopemaster`) are not stored in the OGDS and the usersettings are based on users in the OGDS.

So we're not able to store seen tours for pure plone users. A pure plone user would always see every available tour on the frontend.

To fix this issue, this PR extends the user_settings serializer to mark all tours as seen if we call it as a pure plone user.

## Checkliste

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2019.4.0rc5 (unreleased)
 ------------------------
 
+- Update the usersettings-serializer: A pure plone user has always seen all screens. [elioschmutz]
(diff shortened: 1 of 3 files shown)


elioschmutz:
@lukasgraf could you take look at this PR? Thanks

Issue #5980, opened 2019-10-09
Reordering objects via plone.restapi creates unicode ids, should use bytestring ids.


deiferni:
This issue is very similar to https://github.com/4teamwork/opengever.core/issues/5957. When reordering objects in `plone.restapi` in https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/deserializer/mixins.py#L36-L43 unicode ids are used to update ordering via the `OFS.interfaces.IOrderedContainer` interface. We are of the opinion those ids should be `ascii` instead. The ids used for ordering are later output in a call to `objectIds` where bytestrings are expected. The result of a call to `objectIds` may be used for traversal by the client where unicode paths will fail. Thus we should use bytestring ids to call the reordering methods as they perform no internal conversion. Next steps to take here would be:
- decide how/where to fix this, can we fix it in `plone.restapi`, should we provide a migraiton in `plone.restapi`?
- should we override the functionality here until we can fix it upstream?
- should we provide our own migration to fix or does `plone.restapi` provide such migrations?

njohner:
We fix it in `opengever.core`, then @lukasgraf will discuss the issue with Timo at the ploneconf.

deiferni:
we have discussed this a moment ago. we will create issue  and probably also PR to fix in `plone.restapi`. timo will also start a discussion about the uprgade path that we want to take here. iirc we tend to not upgrade by default as an upgrade would have to inspect all orderable content items but we expect that most content won't actually be affected.

deiferni:
I have attempted to reproduced this locally with `plone5.2` and `python3.7` and `python2.7` with the following steps:

- have used https://github.com/4teamwork/ftw.testing as it already provides base plone5.2 and python3.7 buildout.cfg files.
- setup a new plone site, create one parent folder and some pages in it.
- inspect `objectIds` of folder in debug shell
- reorder content via rest api, move one item down
```js
{
    "ordering": {"obj_id": "page3", "delta": 1}
}
```
- inspect `objectIds` again

plone5.2 + python3.7, no issues as python3 string implementation defaults to unicode

plone5.2 + python2.7: the list returned by a call to `objectIds` will contain mixed unicode and bytestrings. Calls to `keys` and `objectIds` are affected, the `ExtednedPathIndex` is not affected as it relies on `getPhysicalPath`.

```python
from AccessControl.SecurityManagement import newSecurityManager
from zope.component.hooks import setSite

[...]

PR #6030, opened 2019-11-04
Fix an issue with ids not being stored as bytestring when reordering.

deiferni:
This PR fixes an issue where ids of reordered content are converted to unicode on python 2. We now convert input ids when necessary and make sure the ordering stored on a container contains only bytestring object ids.

When changing object ordering via PATCH request we used to incorrectly store ids of reordered resources as unicode instead of a bytestring. This lead to mixed types being stored in the ordering annotations and subsequently mixed types being returned when calling `objectIds` of a container.

Affected content types are todo and todolist (or their possible content types respectively).
We also discovered that with newer versions of `OFS.ObjectManager` non-ascii ids are allowed. Thus we switch agenda item template id input validation to also use the `utf-8` encoding.

See https://github.com/plone/plone.restapi/issues/827 and  https://github.com/plone/plone.restapi/pull/828 for upstream issue and PR.

Closes https://github.com/4teamwork/opengever.core/issues/5980.

## Checkliste

### DIFF
--- a/opengever/core/upgrades/20191031155315_fix_todo_and_todolist_ordering_object_id_type/upgrade.py
+++ b/opengever/core/upgrades/20191031155315_fix_todo_and_todolist_ordering_object_id_type/upgrade.py
@@ -1,0 +1,36 @@
+from ftw.upgrade import UpgradeStep
+from persistent.list import PersistentList
+from zope.annotation.interfaces import IAnnotations
+
+
+ORDER_KEY = 'plone.folder.ordered.order'
+
+
+def safe_utf8(to_utf8):
+    if isinstance(to_utf8, unicode):
+        to_utf8 = to_utf8.encode('utf-8')
+    return to_utf8
+
+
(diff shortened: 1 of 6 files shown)


opengever/core/upgrades/20191031155315_fix_todo_and_todolist_ordering_object_id_type/upgrade.py
```diff
@@ -0,0 +1,36 @@
...
+def safe_utf8(to_utf8):
+    if isinstance(to_utf8, unicode):
+        to_utf8 = to_utf8.encode('utf-8')
+    return to_utf8
+
+
+class FixTodoAndTodolistOrderingObjectIdType(UpgradeStep):
+    """Fix todo and todolist ordering object id type.
+
+    The id is persistet in an objects parent so we query for all types that may
```
njohner: `persistet` -> `persisted`
`objects` -> `object's`

Issue #6022, opened 2019-11-01
API @actions should send 'trashed' action id in 'file_actions' property


all4manu:
In order for the new UI to render the 'move to trash' actions in the context as well as for each item in listings, we need to add the 'trashed' action id to the 'file_actions' property in the @actions endpoint.

See https://github.com/4teamwork/gever-ui/issues/525.

PR #6031, opened 2019-11-04
Add trash_document and untrash_document file_actions

lukasgraf:
Add `trash_document` and `untrash_document` file_actions.

Because [`ActionInfo._checkPermissions()`](https://github.com/zopefoundation/Products.CMFCore/blob/master/Products/CMFCore/ActionInformation.py#L262-L281) only checks permissions on the actual context for action categories that start with `object`,`workflow` or `document`, our permissions for our own `file_actions` category were always checked against the Plone site root. Because so far we only used `View` to protect those actions, this seemed to sort of work. But when using permissions like `"opengever.trash: Trash content"` this failure now surfaces.

🐒I therefore patched `_checkPermissions()` to also consider our `file_actions` category a context based one. IMHO this is even preferrable to renaming the action category, because we don't know what subtle magic around action category names might be lurking around somewhere else, and renaming it to something like `object_file_actions` might lead to unforeseen side effects.

Addresses #6022 

## Checkliste

### DIFF
--- a/opengever/base/monkey/patches/action_info.py
+++ b/opengever/base/monkey/patches/action_info.py
@@ -1,0 +1,40 @@
+from opengever.base.monkey.patching import MonkeyPatch
+from Products.CMFCore.utils import _checkPermission
+
+
+class PatchActionInfo(MonkeyPatch):
+    """We patch the _checkPermissions() method of the ActionInfo object
+    in order to also consider our 'file_actions' category one that should
+    have its actions' permissions checked on the context.
+
+    Without this, the permissions would be checked on the Plone Site instead.
+    """
+
+    def __call__(self):
+
+        def _checkPermissions(self, ec):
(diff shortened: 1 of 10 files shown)


opengever/base/monkey/patches/action_info.py
```diff
@@ -0,0 +1,41 @@
...
+                    context = ec.contexts['portal']
+
+            for permission in self._permissions:
+                if _checkPermission(permission, context):
+                    return True
+            return False
+
+        from Products.CMFCore.ActionInformation import ActionInfo
+        locals()['__patch_refs__'] = False
+        original_checkPermissions = ActionInfo._checkPermissions
```
elioschmutz: `original_checkPermissions` is never used

lukasgraf: Good catch, thanks :)

PR #6033, opened 2019-11-06
 Support combined notation for task responsible

phgross:
Currently the api for task creation/editing only supports a single value for the `responsible_client` and `responsible` field. But the `schema´ or `querysources` endpoint returnes a combined value for the users containing `responsible_client` and `responsible`. This difference leads to bad situation where we would need to split the values in the frontend or need to use a separate endpoint to fetch the responsible values.

I was also thinking about adding a different endpoint or moving the ":" splitting to the frontend, but I think thats the cleanest way. I see the following advantages:
 - The value exposed by the schema endpoint is valid and can be used for a POST or PATCH request. The default implementation on the frontend side works also for tasks.
 - The special handling for combined values are only in the backend and not spread over different endpoints and sources
 - The change does not have any impact on other featueres/modules or on the old frontend, the sources works still the same.
 - No data migration necessary 

Additional fixes for task add/edit support in the frontend:
 - Also pass in the data to the deserialization instead of refetching it from the request.
 - Ignore colons when query one of the contact sources

For https://github.com/4teamwork/gever-ui/issues/124 

## Checkliste
 the forwarding are currently not supported by the rest api, so no need to invest time.
- [ ] Aktualisierung Dokumentation vorhanden/nötig?
 _I'm not sure if I should provide documentation about this "feature" or not, what you think._

### DIFF
--- a/opengever/api/task.py
+++ b/opengever/api/task.py
@@ -1,7 +1,13 @@
 from opengever.api.response import ResponsePost
 from opengever.api.response import SerializeResponseToJson
+from opengever.ogds.base.actor import ActorLookup
+from opengever.ogds.models.team import Team
 from opengever.task.interfaces import ICommentResponseHandler
+from opengever.task.task import ITask
 from opengever.task.task_response import ITaskResponse
+from plone.restapi.deserializer import json_body
+from plone.restapi.deserializer.dxcontent import DeserializeFromJson
+from plone.restapi.interfaces import IDeserializeFromJson
 from plone.restapi.interfaces import ISerializeToJson
 from zExceptions import Unauthorized
 from zope.component import adapter
@@ -14,6 +20,54 @@ from zope.interface import Interface
 class SerializeTaskResponseToJson(SerializeResponseToJson):
 
     model = ITaskResponse
+
+
+@implementer(IDeserializeFromJson)
+@adapter(ITask, Interface)
+class TaskDeserializeFromJson(DeserializeFromJson):
+    """A task specific deserializer which allows to pass in the
+    responsible_client and responsible value in a combined string.
+    In the smae way as it is exposed by the APIs querysoure endpoint.
+    """
+
+    def __call__(self, validate_all=False, data=None, create=False):
+        if data is None:
+            data = json_body(self.request)
+
+        self.update_reponsible_field_data(data)
+
+        super(TaskDeserializeFromJson, self).__call__(
+            validate_all=validate_all, data=data, create=create)
+
+    def update_reponsible_field_data(self, data):
+        """Extract responsible_client when a combined value is used (client,
+        responsible separated by a colon).
+        """
+
+        if not data.get('responsible'):
+            return
+
+        if isinstance(data['responsible'], dict):
+            responsible = data['responsible']['token']
+        else:
+            responsible = data['responsible']
+
+        # Skip values without the orgunit prefix
+        if ':' not in responsible:
+            return
+
+        if ActorLookup(responsible).is_inbox():
+            responsible_client = responsible.split(':', 1)[1]
+
+        elif ActorLookup(responsible).is_team():
+            team = Team.query.get_by_actor_id(responsible)
+            responsible_client = team.org_unit.unit_id
+
+        else:
+            responsible_client, responsible = responsible.split(':', 1)
+
+        data['responsible'] = responsible
+        data['responsible_client'] = responsible_client
 
 
 class TaskResponsePost(ResponsePost):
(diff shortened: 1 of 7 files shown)


opengever/api/task.py
```diff
@@ -16,6 +22,60 @@ class SerializeTaskResponseToJson(SerializeResponseToJson):
...
+            responsible = data['responsible']
+
+        # Skip values without the orgunit prefix
+        if ':' not in responsible:
+            return
+
+        if ActorLookup(responsible).is_inbox():
+            client = responsible.split(':', 1)[1]
+
+            responsible_client = client
```
njohner: why not directly `responsible_client = responsible.split(':', 1)[1]`?

opengever/api/task.py
```diff
@@ -16,6 +22,60 @@ class SerializeTaskResponseToJson(SerializeResponseToJson):
...
+
+        super(TaskDeserializeFromJson, self).__call__(
+            validate_all=validate_all, data=data, create=create)
+
+    def update_reponsible_field_data(self, data):
+        """Extract responsible_client when a combined value is used (client,
+        responsible separated by a colon).
+        """
+
+        if not data['responsible']:
```
njohner: Shouldn't that be something like `if not 'responsible' in data` or `if not data.get('responsible', None)`?

opengever/api/task.py
```diff
@@ -16,6 +22,60 @@ class SerializeTaskResponseToJson(SerializeResponseToJson):
...
+            responsible = responsible
+
+        elif ActorLookup(responsible).is_team():
+            team = Team.query.get_by_actor_id(responsible)
+            responsible_client = team.org_unit.unit_id
+            responsible = responsible
+
+        else:
+            client, user = responsible.split(':', 1)
+            responsible_client = client
```
njohner: Same here, why not `responsible_client, responsible = responsible.split(':', 1)`?

opengever/api/task.py
```diff
@@ -16,6 +22,60 @@ class SerializeTaskResponseToJson(SerializeResponseToJson):
...
+
+        # Skip values without the orgunit prefix
+        if ':' not in responsible:
+            return
+
+        if ActorLookup(responsible).is_inbox():
+            client = responsible.split(':', 1)[1]
+
+            responsible_client = client
+            responsible = responsible
```
njohner: ✂️

opengever/api/tests/test_task.py
```diff
@@ -166,3 +166,146 @@ def test_comment_will_be_synced(self):
...
+                'token': self.secretariat_user.id,
+                'title': u'Finanzamt: J\xfcrgen K\xf6nig'
+            }
+        })
+
+        with self.observe_children(self.dossier) as children:
+            browser.open(self.dossier, json.dumps(self.data),
+                         method="POST", headers=self.api_headers)
+
+        self.assertEqual(1, len(children['added']))
```
njohner: Maybe test that on the added object, the responsible is set as expected

opengever/api/tests/test_task.py
```diff
@@ -166,3 +166,146 @@ def test_comment_will_be_synced(self):
...
+        self.data.update({
+            "responsible": "fa:{}".format(self.regular_user.id),
+            "issuer": self.secretariat_user.id,
+        })
+
+        with self.observe_children(self.dossier) as children:
+            browser.open(self.dossier, json.dumps(self.data),
+                         method="POST", headers=self.api_headers)
+
+        self.assertEqual(1, len(children['added']))
```
njohner: Same as above

opengever/api/tests/test_task.py
```diff
@@ -166,3 +166,146 @@ def test_comment_will_be_synced(self):
...
+        self.assertEqual('inbox:fa', task.issuer)
+
+    @browsing
+    def test_supports_teams(self, browser):
+        self.login(self.regular_user, browser=browser)
+
+        self.data.update({
+            "responsible": "team:1",
+            "issuer": self.regular_user.id,
+        })
```
njohner: Missing the last part of the test here

opengever/ogds/base/tests/test_sources.py
```diff
@@ -225,6 +225,12 @@ def test_search_for_orgunit(self):
         self.assertTermKeys([u'org-unit-1:hans', u'org-unit-1:hugo', u'org-unit-1:john'],
                             result)
 
+    def test_search_for_inbox(self):
+        result = self.source.search('inbox:org-unit-1')
+
+        self.assertEquals(1, len(result), 'Expect 3 items')
```
njohner: `Expect 1 item`

opengever/api/task.py
phgross: 🙈 no reason ... 😅

phgross:
@njohner updated ...

opengever/ogds/base/sources.py
```diff
@@ -233,6 +233,8 @@ def getTermByToken(self, token):
 
     def search(self, query_string):
         self.terms = []
+        # Ignore
```
njohner: That comment is a bit unhelpful

phgross:
@njohner oh thanks 👍 ...

Issue #5988, opened 2019-10-15
Document with mail as file should not be allowed


njohner:
_Harvest: `Projekt - Umsetzung Neues User Interface GEVER 2019/2020` - `Entwicklung`_

Currently it is possible to create a document with a mail as file. It will soon also be possible to add an E-mail with a normal file as message. 

The problem is that the validator (https://github.com/4teamwork/opengever.core/blob/master/opengever/document/document.py#L128-L129) does not get called when the document is created through the REST-API.

Check whether we need an upgrade step (on lab and our gever). This needs an upgrade step to fix broken mails / documents. (trash / delete the concerned documents, and generate mails instead).

### Info Sprint Abschluss: 
 Ausstehend nur noch review Anpassungen.

PR #6034, opened 2019-11-06
Disallow mail upload as documents via API.

phgross:
Because we still wont an detailed error message (including the mail-in address) directly placed by the field, I duplicate the check from the simplefieldvalitor and register it also as a schema invariant. Simplefieldvalidator are z3c form bsed validation and therefore not called by plone.restapi.

An migration/upgradestep is not necessary, I doublechecked our gever installation, there were only 2 which i migrated manually (delete and recreate). Other productive deployment should not be affected by this.

For #5988

## Checkliste
- [ ] Gibt es neue Funktionalität mit einem `Dokument`? Funktioniert das auch mit einem `Mail`? _Nur bei einem Dokument relevant._ 

Checked by:
```
from opengever.maintenance import dm
dm()
from plone import api
cat = api.portal.get_tool('portal_catalog')

MAIL_EXTENSIONS = ['.eml', '.msg', '.p7m']
wrong = []

for brain in docs:
    doc = brain.getObject()
    if doc.get_file() and doc.get_file_extension() in MAIL_EXTENSIONS:
        wrong.append(doc)
```

### DIFF
--- a/opengever/api/tests/test_document.py
+++ b/opengever/api/tests/test_document.py
@@ -2,6 +2,7 @@ from ftw.testbrowser import browsing
 from opengever.document.interfaces import ICheckinCheckoutManager
 from opengever.testing import IntegrationTestCase
 from zope.component import getMultiAdapter
+import json
 
 
 class TestDocumentSerializer(IntegrationTestCase):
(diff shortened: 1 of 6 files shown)


opengever/api/tests/test_document.py
```diff
@@ -83,6 +84,22 @@ def test_additional_metadata_for_mails(self, browser):
...
+    def test_does_not_allow_mails_as_file(self, browser):
+        self.login(self.regular_user, browser)
+
+        with browser.expect_http_error(code=400, reason='Bad Request'):
+            data = {'@type': 'opengever.document.document',
+                    'file': {'data': 'foo bar', 'filename': 'test.eml'}}
+            browser.open(self.dossier, data=json.dumps(data), method='POST',
+                         headers=self.api_headers)
+        self.assertEqual(
+            u"[{'message': 'error_mail_upload_api', 'error': Invalid(u'error_mail_upload_api',)}]",
```
deiferni: this looks weird, why do you have json data as string here? shouldn't `browser.json` decode the whole body if it is correct json?

opengever/document/document.py
```diff
@@ -101,6 +106,15 @@ def title_or_file_required(data):
                             default=u'Either the title or the file is '
                             'required.'))
 
+    @invariant
+    def no_eml(data):
```
deiferni: i would not call this method `…eml` as the check is broader, could you change to e.v. `disallow_email_extensions` or something alike.

opengever/api/tests/test_document.py
phgross: This is a general Problem, I open a separate Issue for this (https://github.com/4teamwork/opengever.core/issues/6047)

PR #6054, opened 2019-11-13
No longer add journal entry for document file modifications.

njohner:
Newer OfficeConnector versions upload the modified file each time it gets saved. As these file modifications get journalised, this clogs up the journal with uninformative entries. With this change, we no longer journalise file modifications. Instead, document checkin/checkout entries serve as markers to who modified a document and when (although a checkout/checkin cycle with no modifications cannot be distinguished in the journal from one with modifications, but document versions would allow to track changes if necessary).

For https://github.com/4teamwork/opengever.core/issues/6003

## Checkliste

### DIFF
--- a/opengever/journal/handlers.py
+++ b/opengever/journal/handlers.py
@@ -348,9 +348,8 @@ def document_modified(context, event):
         # and is handled elsewhere - don't journalize it twice.
         return
 
-    # we need to distinguish between "metadata modified", "file modified",
-    # "file and metadata modified" and "public_trial modified"
-    file_changed = False
+    # we need to distinguish between "metadata modified"
+    # and "public_trial modified"
     metadata_changed = False
     public_trial_changed = False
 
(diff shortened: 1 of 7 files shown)


opengever/journal/handlers.py
```diff
@@ -348,56 +348,28 @@ def document_modified(context, event):
...
                 metadata_changed = True
 
-    if context.REQUEST.get('form.widgets.file.action',
-                           u'nochange') == u'nochange':
-        file_changed = False
-
-    if not file_changed and not metadata_changed and not public_trial_changed:
-        # the event shouldn't be fired in this case anyway..
+    if not metadata_changed and not public_trial_changed:
+        # We end up here when only the file has changed
```
phgross: I would add a hint/comment here, that the file_changed event needs no journalization because of the checkout/checkin cycle.

Issue #5848, opened 2019-08-02
Get rid of non-api office connector features.


deiferni:
Harvest: `[3.2.4 / 2] Projekt - teamraum 2019` - `Entwicklung`

We currently support non-api office connector features:
- non api based checkout and edit
- non api based upload

We have discussed that we would like to get rid of these features and make office connector via api the one and only default. Getting rid of that would mean to:
- Get rid of logic and different code paths for api and zem
- Possibly also adapt OfficeConnector and remove logic there
- Get rid of registry flags and adapt https://gever.4teamwork.ch/vorlagen/opengever-dossier-templatefolder/document-22875#overview

❗️ The estimate only refers to the `opengever.core` side - OfficeConnector changes will need to be initiated/tracked by the person doing this, but are not part of the estimate.

njohner:
I don't see what to adapt in https://gever.4teamwork.ch/vorlagen/opengever-dossier-templatefolder/document-22875, the `api_enabled` flag is not mentioned is the checklist.

njohner:
Created issue in OfficeConnector: https://github.com/4teamwork/OfficeConnector/issues/336

njohner:
This is done, but we wait to make sure all clients have the new version of OC installed

phgross:
Fachliches review 👍

PR #5126, opened 2018-11-20
Implement feature flagged RESTAPI payloads for Office Connector

Rotonen:
Amended the fixtured shadow document to have a doctype.

Made unshadowing documents trigger on file upload instead of unlocking as there was no convenient way to trip it up otherwise via the RESTAPI. Should just work as expected.

Added a feature flag for producing RESTAPI using payloads for Office Connector. The Office Connector to use this does not exist yet, but I've also implemented user story tests, including Oneoffixx tests, which show how this will work.

Closes (in tandem with the OC change): https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/439

PR #6058, opened 2019-11-15
Always use API for OfficeConnector.

njohner:
In this PR we remove support for non-API based OfficeConnector actions (checkout, upload, checkin, lock and unlock). This makes API-based OfficeConnector the one and only default.

For https://github.com/4teamwork/opengever.core/issues/5848

## Checkliste

Introduced in GEVER in https://github.com/4teamwork/opengever.core/pull/5126
Introduced in OC in https://github.com/4teamwork/OfficeConnector/pull/307

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2019.5.1 (unreleased)
 ---------------------
 
+- Always use API for OfficeConnector. [njohner]
 - Refactor solrsearch and listing endpoints. [njohner]
 - Add tests for solrsearch and listing endpoints. [njohner]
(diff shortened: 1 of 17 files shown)


njohner:
With this change, https://github.com/4teamwork/opengever.ftw/pull/84 will not need merging anymore. But we indeed need to be carefull:
- Make sure internally everybody has the latest OC version.
- Inform clients that older versions of OC will not work after update (1.9.0 introduced the API, but 1.9.2 would be better as several bug fixes happened in 1.9.1 and 1.9.2)

deiferni:
As just discussed we'll merge and ship this for `2020.1`

Issue #5934, opened 2019-09-16
API Support für Anträge - Verlauf/ProposalHistory


deiferni:
_Harvest: `Projekt - Umsetzung Neues User Interface GEVER 2019/2020` - `Entwicklung`_

Als API Benutzer will den Verlauf eines Antrags sehen. 

- [x] custom proposal API `GET` that contains proposal history. suggested to implement similar to task responses, see example: https://github.com/4teamwork/opengever.core/blob/master/opengever/api/task.py#L13-L22
- [x] BaseResponse für Anträge und eingereichte Anträge verwenden, ProposalHistory durch BaseResponse ersetzen.

njohner:
After having a good look at the code I think this will require quite some refactoring. The main things we need to handle are:
- The `message` field. We have one class per record type (subclassing `BaseHistoryRecord`), each defining a `message`, which is a translated string and gets used in the `proposaloverview.pt`. If we use `BaseResponse`s instead, we do not want to store the messages on the response objects (as they are stored in the annotations). I'd simply propose a mapping from response type to message (similar to what is done for activities).
- The `css_class` will have to be handled similarly to the `message` field above.
- Syncing the history between the proposal and the submitted proposal. I think this could be handled using a `needs_syncing` attribute on the response and an event handler for `ResponseAddedEvent`, although it's not straightforward and we have to be careful not to end up in a syncing loop. A possible solution would be:
  - We make a class for the proposal responses with correct defaults for the syncing depending on the response type, something like:
    ```
    class ProposalResponse(BaseResponse):
        def __init__(self, type, text, needs_syncing=None):
            if needs_syncing is not None:
                 self.needs_syncing = needs_syncing
            else:
                if type == `reopened`:
                    self.needs_syncing = True
                else:
                    self.needs_syncing = False
    ```
  - In the event handler we specifically set `needs_syncing` to False to make sure, we don't enter a syncing loop:
    ```
    def sync_proposal_response(context, event):
        if event.response.needs_syncing:
            obj = ... // find the object where to sync to
            IResponseContainer(obj).add(ProposalResponse(response.type, response.text, False))
    ```
- Some `HistoryRecord`s also generate an activity when added, this can also be handled in the event handler

With that I think we can delete the whole `ProposalHistory` class and the various classes for `HistoryRecord`s, which should make this much leaner...

njohner:
[...]

PR #6069, opened 2019-11-19
Fix ProposalHistory migration to Response.

njohner:
It seems that records are missing the text attribute in certain cases. I think the origin of the missing text is in the upgrade from SQL to plone: https://github.com/4teamwork/opengever.core/blob/e3046d8adbc8f5c06f9dd9d788a41b152ad7cfce/opengever/core/upgrades/20170704112234_migrate_sql_proposal_history_to_plone/upgrade.py

We do not seem to set the text there...

For https://github.com/4teamwork/opengever.core/issues/5934

### DIFF
--- a/opengever/core/upgrades/20191108143533_migrate_proposal_history_to_proposal_responses/upgrade.py
+++ b/opengever/core/upgrades/20191108143533_migrate_proposal_history_to_proposal_responses/upgrade.py
@@ -6,9 +6,11 @@ from opengever.meeting.proposalhistory import IProposalResponse
 from opengever.meeting.proposalhistory import ProposalResponse
 from zope.annotation.interfaces import IAnnotations
 from zope.schema import getFields
+import logging
 
 
 annotation_key = 'object_history'
+log = logging.getLogger('ftw.upgrade')
 
 
 class MigrationResponseContainer(ResponseContainer):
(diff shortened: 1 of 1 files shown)


opengever/core/upgrades/20191108143533_migrate_proposal_history_to_proposal_responses/upgrade.py
```diff
@@ -50,15 +52,21 @@ def migrate_records_to_responses(self, context):
...
-            history_type = record.pop('history_type')
-            userid = record.pop('userid')
-            text = record.pop('text')
+            created = record.pop('created', None)
+            history_type = record.pop('history_type', None)
+            userid = record.pop('userid', None)
+            if created is None or history_type is None or userid is None:
+                log.warning("Incomplete record on {}: {}".format(
+                    context.absolute_url(), record))
+
```
deiferni: `continue` here?

opengever/core/upgrades/20191108143533_migrate_proposal_history_to_proposal_responses/upgrade.py
```diff
@@ -50,15 +52,21 @@ def migrate_records_to_responses(self, context):
...
+            # Text can be None, because it seems we did not set it during the
+            # last migration from sql model to the annotations...
```
deiferni: sorry 😞.

njohner: We could blame the reviewer :-D

PR #6070, opened 2019-11-19
Fix fallback to default sorting index in listing endpoint.

njohner:
The fallback to the default sorting index in the `listing` endpoint was broken. This PR fixes that bug and adds tests that sorting parameters get set properly.

## Checkliste

For https://sentry.4teamwork.ch/sentry/onegov-gever/issues/35027/


njohner:
Ah yes thanks, that was the plan...

Issue #6026, opened 2019-11-04
@solrsearch endpoint should return facets label/title


all4manu:
We ran into an issue with the new gever-ui when querying the `@solrsearch` endpoint. In order to keep the backend calls to a minimum, we use the facet information returned from the endpoint to build a list of facets the user can further refine the results with.

Currently, the endpoint only returns the token values for each item per facet. This puts us in a difficult position regarding translations. We expected to receive the facet information in the same way the `@listing` endpoint returns them: 

```json
{
  facets: {
    facet_name: {
      token_value: {
        count: 1,
        label: "Translated value",
      },
    },
  }
}
```

This would make the translation situation much easier in the frontend for the following cases:

- Some facets (i.e. document_type) are actually dynamic, which means they can be customized for different GEVER installations.
- Some facets (i.e. users) aren't finite lists, so we cannot translate them in a static file.
- If the items for a particular facet change (i.e. new ones are added), there is no risk that we forget to update the frontend.

phgross:
Fachliches Review 👍

PR #6071, opened 2019-11-19
Also return facet labels in solrsearch endpoint.

njohner:
The `solrsearch` endpoint returned the facets directly as received from solr, without any of cleanup done in the `SolrResponse` and in the `@listing` endpoint, i.e. in the form:
```
"facet_counts": {
    "facet_fields": {
        "document_type": [
            "contract",
            1,
            "request",
            1
        ]
    },
    "facet_heatmaps": {},
    "facet_intervals": {},
    "facet_queries": {},
    "facet_ranges": {}
}
```
This is hard to use in the frontend and labels for the facets are missing, which poses challenges for the translations in the frontend.

In this PR, we fix the `facet_counts` in the response to match that of `@listing` endpoint:
- We use the facets from `SolrResponse`, which is a dictionary of facets and counts (instead of the list with every second element being a facet or a count)
- We add the labels to the facets for certain fields, as done in the `@listing` endpoint

Response now looks like:
```
    "facet_counts": {
        "document_type": {
            "contract": {
                "count": 1,
                "label": "Vertrag"
            },
            "request": {
                "count": 1,
                "label": "Antrag"
            }
        },
        "responsible": {
            "committee:161": {
                "count": 2,
                "label": "Kommission für Rechtsfragen"
            }
        }
    }
```
For https://github.com/4teamwork/opengever.core/issues/6026

Note that we should extract the method preparing the facets for the response to a separate function used in both the `listing` and `solrsearch` endpoints, but this will be done when refactoring the `listing` endpoint (https://github.com/4teamwork/opengever.core/issues/5901)

## Checkliste

### DIFF
--- a/opengever/api/solrsearch.py
+++ b/opengever/api/solrsearch.py
@@ -1,5 +1,6 @@
 from ftw.solr.interfaces import ISolrSearch
 from ftw.solr.query import make_query
+from opengever.api.listing import FACET_TRANSFORMS
 from opengever.base.interfaces import ISearchSettings
 from opengever.base.solr import OGSolrContentListing
 from plone import api
@@ -143,6 +144,13 @@ class SolrSearchGet(Service):
             "rows": rows,
         }
 
-        if resp.facets:
-            res['facet_counts'] = resp.facets
+        facet_counts = {}
+        for field, facets in resp.facets.items():
+            facet_counts[field] = {}
+            transform = FACET_TRANSFORMS.get(field)
+            for facet, count in facets.items():
+                facet_counts[field][facet] = {"count": count}
+                if transform:
+                    facet_counts[field][facet]['label'] = transform(facet)
+        res['facet_counts'] = facet_counts
         return res
--- a/opengever/api/solrsearch.py
+++ b/opengever/api/solrsearch.py
@@ -143,6 +143,6 @@ class SolrSearchGet(Service):
             "rows": rows,
         }
 
-        if 'facet_counts' in resp.body:
-            res['facet_counts'] = resp.body['facet_counts']
+        if resp.facets:
+            res['facet_counts'] = resp.facets
         return res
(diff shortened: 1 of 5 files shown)


opengever/api/tests/test_solrsearch.py
```diff
@@ -197,6 +196,10 @@ def test_returns_facets(self, browser):
 
         self.assertIn(u'facet_counts', browser.json)
 
+    @skip("Just a reminder to test that facets also return translated labels")
+    def test_returns_facet_labels(self):
+        pass
```
deiferni: i do not understand why this is not tested yet, are we waiting for solr testing things? if so, could you amend an issue so we don't forget to update?

opengever/api/solrsearch.py
```diff
@@ -143,6 +144,13 @@ def reply(self):
             "rows": rows,
         }
 
-        if 'facet_counts' in resp.body:
-            res['facet_counts'] = resp.body['facet_counts']
+        facet_counts = {}
+        for field, facets in resp.facets.items():
```
deiferni: you could put transforming facets into a separate function, so you could test it in an unit test (i.e. `dict` in, transformed `dict` out, that should be easily testable)

all4manu:
A frontend question: It seems that I don't get labels for every facet. More specifically - in the current UI's version of the search feature (see lab) - I only get labels for `document_type`, `responsible` and `review_state`.

For the following facets I don't receive labels: `Creator`, `Subject`, `file_extension` and `portal_type`. It is true that I don't need labels for every facet, but I thought I'd list them all here.

Do I have to change my query parameters or is there something missing on the backend side?

njohner:
@all4manu it's the same in the `@listing` endpoint, we only give back labels when the labels are different from the facet names. I could change that, I'm not sure what is better from an API point of view. Let's discuss at the daily.

opengever/api/tests/test_solrsearch.py
njohner: Yes I really don't feel like testing that I did put translated labels in my faked solr response. Makes absolutely no sense...

I've made an issue with some of the files that will need to be looked at when we have solr testing: https://github.com/4teamwork/opengever.core/issues/6072

opengever/api/solrsearch.py
njohner: Yes I'd keep that for the refactoring (which I'm doing now)

PR #6092, opened 2019-11-28
Fix styling bug in tabbedview after showing bumblebee tooltip.

njohner:
We fix a styling issue in the documents tab, where after opening the bumblebee tooltip, the highlighting of the corresponding row does not get removed when the tooltip is closed again.

It also happens that the highlighted row is in front of other open menus (search menu for example), but I think that is fine, as long as the highlighting is not left behind...

**With open search and tooltip, the row color and text is above the search menu.**
Screenshot 2019-11-28 at 13 06 31

**Once the tooltip is closed, the color is removed and does not overlap the search menu anymore**
Screenshot 2019-11-28 at 13 03 16

For https://github.com/4teamwork/opengever.core/issues/5898

## Checkliste

### DIFF
--- a/opengever/base/browser/resources/tooltip.js
+++ b/opengever/base/browser/resources/tooltip.js
@@ -69,22 +69,28 @@
     $(target).closest('.x-grid3-row').addClass('bumblebee-tooltip-active');
   }
 
-  function deactivateTableCells(target) {
+  function deactivateOtherTableCells(target) {
     $(target).closest('.x-grid3-row').siblings().removeClass('bumblebee-tooltip-active');
+  }
+
+  function deactivateTableCell(target) {
+    $(target).closest('.x-grid3-row').removeClass('bumblebee-tooltip-active');
   }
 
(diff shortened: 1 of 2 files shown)


opengever/base/browser/resources/tooltip.js
```diff
@@ -69,22 +69,29 @@
...
   }
 
-  function hideBackdrop() {
+  function hideBackdrop(event) {
     tooltipHideTimer = setTimeout(function() {
       $('body').removeClass('bumblebee-tooltip-open');
     }, settings.hide.delay / 2);
+    var target = event.originalEvent.target;
+    deactivateTableCell(target);
+    $(target).closest('.x-grid3-row').removeClass('bumblebee-tooltip-active');
```
all4manu: Isn't this already in `deactivateTableCell` and should be called only once?

njohner: 😅 I moved it to a function and forgot to delete. Sorry

PR #6114, opened 2019-12-06
Implement Solr testing (stage 2)

lukasgraf:
This implements testing against a **real Solr instance**.

With this change we introduce a new `ContentFixtureWithSolrLayer` that allows to run tests against a real Solr instance (with the regular GEVER content fixture).

We
- take care of **Solr isolation** between tests by taking snapshots using the Solr replication API
- ensure **fixture caching** works by introducing a `cache_label` that allows for keeping distinct cache subdirectories (to isolate incompatible caches from each other) 
- put the **Solr tests on their own layer**. With that, we can avoid `bin/mtest` booting more than one Solr on Jenkins, and run the Solr tests in their own `test-solr.cfg` (and skip that layer in `bin/mtest`)

## Checkliste


opengever/testing/tests/test_solr.py
```diff
@@ -0,0 +1,41 @@
...
+        result = browser.json
+        self.assertEqual(1, result['items_total'])
+        self.assertEqual(1, len(result['items']))
+        self.assertEqual(random_title, result['items'][0]['title'])
+
+
+# Create additional copies of the above test that would fail if isolation
+# didn't work between tests (i.e. if created objects in Solr would leak
+# between test runs instead of getting reset to the layer snapshot).
+for i in range(2):
```
jone: If understand correctly, 2 copies means that we end up with 3.
I think having one copy (resulting with 2 tests) should be enough for proving isolation.
Why did you choose to make 2 copies?

opengever/testing/tests/test_solr.py
```diff
@@ -0,0 +1,41 @@
+from ftw.builder import Builder
+from ftw.builder import create
+from ftw.testbrowser import browsing
+from opengever.core.testing import OPENGEVER_SOLR_INTEGRATION_TESTING
+from opengever.testing import IntegrationTestCase
+import random
+
+
+class TestSolrIsolation(IntegrationTestCase):
```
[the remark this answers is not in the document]
lukasgraf: Good idea - I'll look into it 👍

opengever/testing/tests/test_solr.py
lukasgraf: No particular reason - you're right, 2 in total are be enough.

lukasgraf:
We're shipping this even though we've seen some behavior (when running tests locally)
that suggests that there *might* be some isolation issues.

In particular, when running tests from the `ContentFixtureWithSolrLayer` and `ContentFixtureLayer` together, we've seen failures that indicate that there might be layer isolation issues. But because those are very complex and time consuming to track down, we've decided to run with this for now, and keep an eye on it for now.

Depending on whether issues crop up when running tests locally, we'll have to look at this again and possibly deactivate these tests again. Running the tests on CI (Jenkins) should work though, because those two layers are never run together.

PR #6120, opened 2019-12-11
Make sure that owner has correct type in webactions.

njohner:
For some webactions, `owner` was set to a unicode string, which then does not pass validation when patching the webaction. We were not validating that field when adding a webaction as it is generated in the backend and not user input. I now:
- Added validation of all the fields when creating a new webaction
- Cast the `userid` to correct type before setting it to the `owner` field of the webaction
- Switch `owner` and `groups` field to unicode
- Update existing webactions to make sure correct type is being used

I switch to unicode as some userids and groups can be unicode

For https://github.com/4teamwork/opengever.core/issues/6011

## Checkliste

### DIFF
--- a/opengever/core/upgrades/20191211081852_make_webaction_owners_and_groups_unicode/upgrade.py
+++ b/opengever/core/upgrades/20191211081852_make_webaction_owners_and_groups_unicode/upgrade.py
@@ -1,0 +1,18 @@
+from ftw.upgrade import UpgradeStep
+from opengever.webactions.storage import get_storage
+from Products.CMFPlone.utils import safe_unicode
+
+
+class MakeWebactionOwnersAndGroupsUnicode(UpgradeStep):
+    """Make webaction owners and groups unicode.
+    """
+
+    deferrable = True
+
+    def __call__(self):
+        storage = get_storage()
+        for action in storage._actions.values():
+            action['owner'] = safe_unicode(action['owner'])
+            if 'groups' in action:
+                action['groups'] = map(safe_unicode,
+                                       action['groups'])
--- a/opengever/testing/builders/webactions.py
+++ b/opengever/testing/builders/webactions.py
@@ -1,5 +1,6 @@
 from ftw.builder import builder_registry
 from opengever.webactions.interfaces import IWebActionsStorage
+from Products.CMFPlone.utils import safe_unicode
 from zope.component import getMultiAdapter
 from zope.component.hooks import getSite
 from zope.globalrequest import getRequest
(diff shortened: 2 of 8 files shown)


opengever/webactions/storage.py
```diff
@@ -105,7 +106,9 @@ def add(self, action_data):
         new_action['action_id'] = action_id
         new_action['created'] = now
         new_action['modified'] = now
-        new_action['owner'] = userid if userid else 'Anonymous'
+        # Some userids are unicode so we need to make sure that
```
lukasgraf: This sentence started, but it never

opengever/webactions/storage.py
```diff
@@ -108,7 +108,7 @@ def add(self, action_data):
         new_action['modified'] = now
         # Some userids are unicode so we need to make sure that
         owner_field = IWebActionMetadataSchema.get("owner")
-        new_action['owner'] = owner_field._type(userid) if userid else 'Anonymous'
+        new_action['owner'] = owner_field._type(userid) if userid else u'Anonymous'
```
lukasgraf: This is equivalent to `unicode(userid)`, which would attempt to decode `userid` as `ascii` if it happens to be a bytestring. Since `owner` and `group` fields now are unicode, with the intention of also supporting IDs with non-ASCII characters, we should make sure to also actually support them. So `safe_unicode(userid)` would be the way to go here IMHO (`from Products.CMFPlone.utils import safe_unicode`).

opengever/core/upgrades/20191211081852_make_webaction_owners_and_groups_unicode/upgrade.py
```diff
@@ -0,0 +1,18 @@
...
+class MakeWebactionOwnersAndGroupsUnicode(UpgradeStep):
+    """Make webaction owners and groups unicode.
+    """
+
+    def __call__(self):
+        storage = get_storage()
+        owner_field = IPersistedWebActionSchema.get("owner")
+        groups_field = IPersistedWebActionSchema.get("groups")
+        for action in storage._actions.values():
+            action['owner'] = owner_field._type(action['owner'])
```
lukasgraf: Same here, `safe_unicode` should be used for owner field and the group field's values.

opengever/testing/builders/webactions.py
```diff
@@ -30,7 +30,7 @@ def titled(self, title):
         return self
 
     def owned_by(self, owner):
-        self.owner = owner
+        self.owner = unicode(owner)
```
lukasgraf: Same here, `safe_unicode`.

PR #6123, opened 2019-12-11
Replace @participations endpoint with @invitations endpoint

deiferni:
For #5911. With this PR we replace the `@participations` endpoint with an `@invitations` endpoint that no longer accepts an existing user-id but always requires the e-mail for invitations. It adapts the existing storage so that it requires an email and then attempts to find the corresponding user via pas_search. The user may also be `None`.

Current state:
- Missing description in `IWorkspaceInvitationSchema`, maybe add when implementing mails?
- the UI will break completely with these changes 🙈

## Checkliste

- [ ] Aktualisierung Dokumentation vorhanden/nötig? **yes but as in https://github.com/4teamwork/opengever.core/pull/6121 not done yet.**


opengever/workspace/exceptions.py
```diff
@@ -0,0 +1,7 @@
+class DuplicatePendingInvitation(Exception):
+    """Tried to invite an e-mail address that already has a pending invitation.
+    """
+
+
+class MultipleUsersFound(Exception):
+    """Found multiple users for the invitaion."""
```
njohner: invitaion -> invitation

opengever/workspace/participation/storage.py
```diff
@@ -66,9 +82,9 @@ def iter_invitations_for_context(self, context):
             if data['target_uuid'] == uuid:
                 yield self.get_invitation(iid)
 
-    def iter_invitations_for_recipient(self, userid):
+    def iter_invitions_for_recipient_email(self, email):
```
njohner: You reintroduced the typo 😅

opengever/workspace/tests/test_participation_storage.py
```diff
@@ -124,24 +139,26 @@ def test_iter_invitations_for_context(self):
     def test_iter_invitations_for_recipient(self):
         self.login(self.workspace_admin)
         storage = getUtility(IInvitationStorage)
-        foo1 = self.add_invitation(recipient='foo')
-        foo2 = self.add_invitation(recipient='foo')
-        bar = self.add_invitation(recipient='bar')
+        foo1 = self.add_invitation(recipient_email='foo@example.com')
+        foo2 = self.add_invitation(target=self.workspace_folder,
```
njohner: Does it make sense an invitation on a `workspace_folder`?

opengever/api/invitations.py
```diff
@@ -136,3 +142,25 @@ def validate_data(self, data):
...
+            raise BadRequest(errors)
+
+        inviter_id = api.user.get_current().getId()
+        storage = getUtility(IInvitationStorage)
+        iid = storage.add_invitation(self.context,
+                                     data['recipient_email'],
+                                     inviter_id,
+                                     data['role'])
+
+        self.request.response.setStatus(201)
```
njohner: Hum why not use `ManageParticipants._add`, which does more or less precisely that?

opengever/workspace/invitation.py
```diff
@@ -1,16 +1,52 @@
...
+    preceded and followed by at least one char.
+
+    Everything else is validated by actually sending the mail.
+    """
+    EMAIL_REGEX = re.compile('^[^@]+@[^@]+$')
+
+    def validate(self, value):
+        """Validate if the retention period of all selected dossiers is
+        expired and the dossiers are not part of another disposition.
+        """
```
njohner: Comment is probably quite wrong

opengever/workspace/invitation.py
```diff
@@ -1,16 +1,52 @@
...
+        super(RecipientEmailValidator, self).validate(value)
+
+        if u'@' not in value:
```
njohner: Do you need that first part? The regex coming after this also checks there is an `@` symbol, seems enough to me

njohner:
@deiferni I'm taking this over.

opengever/workspace/invitation.py
```diff
@@ -1,16 +1,52 @@
...
+                  default=u'The provided email address is invalid, should '
+                           'contain @.'))
+
+        if not self.EMAIL_REGEX.match(value):
+            raise Invalid(
+                _(u'error_invalid_email',
+                  default=u'The provided email address is invalid.'))
+
+
+validator.WidgetValidatorDiscriminators(
```
deiferni: this, of course, can't work in the API 🤦‍♂as there are no widgets.

opengever/api/invitations.py
deiferni: It is still user-id based. I'm expecting the `ManageParticipants` view to go away eventually, but i agree the functions are not the best way to handle it. But i think `add` and `_add` should be removed instead at the moment?

opengever/workspace/invitation.py
deiferni: i have added it to provide a bit nicer error messages.

opengever/workspace/tests/test_participation_storage.py
deiferni: I'm not sure. it was currently possible. i left it in.

deiferni:
@njohner i have addressed or implemented your feedback. Please feel free to take this over tomorrow, hopefully it can be used more or less as is?

njohner:
I've corrected some failing tests and added the mailing part. I've started working on the view to accept the invitations.

deiferni:
ok, changes LGTM 👍

PR #6139, opened 2019-12-17
Enable the solr flag in the policytemplate.

phgross:
We enable solr for every new deployment, so it makes sense to configure this default. 

## Checkliste


deiferni:
(i have hijacked this branch and rebased)

PR #6161, opened 2019-12-20
Invitations for external users via E-mail

njohner:
In this PR we modify the invitations to work over E-mails, so that invitations can be sent to external users.

Basically this works as follows:
- An E-mail is sent when a new invitation is added containing a link to the `@@my-invitations/accept` view.
- The view itself handles 3 different scenarios:
  - The user is already signed in, then the invitation is accepted for that user
  - The user is not signed in:
    - The E-mail address of the invitation corresponds to an existing user. In that case the the user gets redirected to the portal `login` and from there back to the `@@my-invitations/accept` view.
    - The E-mail address does not correspond to any existing user. In that case the user is redirected to the portal `registration` and from there back to the `@@my-invitations/accept`
Moreover, when the user is redirected back to the `@@my-invitations/accept` view after logging in or registering, he does not yet have a valid GEVER session and we need raise `Unauthorized` so that the CAS authentication happens...

The payload for accepting and declining an invitation is signed using `itsdangerous`.
We also sign the payload in the accept url in the E-mail. This prevents guessing a valid invitation id as possible security issue...

Finally, when an invitation with a given E-mail is accepted by a user, we map all other invitations with that same E-mail address to that user, so that he will see those in the list of his invitations.

The whole process was tested locally with a portal and works correctly.

For https://github.com/4teamwork/opengever.core/issues/5911

## Checkliste
- [ ] Aktualisierung Dokumentation vorhanden/nötig? Documentation should be actualised when everything is done

# Things that still need doing
- store user in OGDS after registration. In `MyWorkspaceInvitations.accept`, when `payload.get('new_user')`, the user should be written into the `ogds`
- Also use signed payloads for the REST-API:
  - Adapt the POST endpoint creating a new invitation to not return anything
  - Adapt the `WorkspaceInvitationsPost` for accepting and declining invitations
  - Probably also adapt `InvitationsDelete` and `InvitationsPatch`
  - This also means adapting `MyWorkspaceInvitations.get_invitations` to return signed urls for the invitations that the frontend can use
- `InvitationRoleAssignment` should store the invitation `iid`. For now it stores the target (i.e. the context), which does not seem to make too much sense to me. It's anyway stored on the context, why would it need to also store it in the assignment??
- Add `decline` URL in E-mail?


lukasgraf:
I added creation of OGDS entry upon accepting an invitation, fixed the flow for an existing, but not logged in user, and fixed test flakyness.

PR #6182, opened 2020-01-06
Configure Solr replication handler

buchi:
Allows to run Solr in a master/slave replication setup.

Replication is disabled by default. It can be enabled using the Solr Config API by setting the `solr.replication.masterUrl` and `solr.replication.enable` properties on the slave.

Example:
```
curl -X POST \
  -H 'Content-type:application/json' \
  -d '{"set-user-property": {"solr.replication.masterUrl": "http://solr-master:8983/solr/core1/replication"}}' \
 http://solr-slave:8983/api/cores/core1/config

curl -X POST \
  -H 'Content-type:application/json' \
  -d '{"set-user-property": {"solr.replication.enable": "true"}}' \
   http://solr-slave:8983/api/cores/core1/config
```

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2019.5.1 (unreleased)
 ---------------------
 
+- Configure Solr replication handler. [buchi]
 - Implement @role-inheritance serivce endpoint for workspace folders. [deiferni]
 - Include permissions.zcml of Products.CMFEditions. [lgraf]
 - Add action to create new invitations to workspaces. [deiferni]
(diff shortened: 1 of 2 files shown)


lukasgraf:
We would then need to configure `ftw.solr` on both instances (srv13 and srv14) so that they use `solr-master:8983` instead of `localhost`8983` though, right? And that would mean that the Solr master would need to listen on all interfaces (`0.0.0.0` instead of just `localhost`). That would require some additional support in [`ftw.recipe.solr`](https://github.com/4teamwork/ftw.recipe.solr)

buchi:
Correct. We will make an entry in `/etc/hosts` for `solr-master` and use that for the Solr URL.
This also allows to switch easy to the slave in case of a failure of the master server.

ftw.recipe.solr already supports to listen on all interfaces with the host option: https://github.com/4teamwork/ftw.recipe.solr#supported-options

PR #6190, opened 2020-01-08
Workspace folder participation, implement @role_inheritance views.

deiferni:
This PR implements the a `@role_inheritance` `POST` and `GET` service endpoint for workspace folders as specified in https://github.com/4teamwork/opengever.core/issues/6167.

- The `GET` service endpoint allows a user who can `View` to query if role inheritance on context is blocked
- The `POST` endpoint allows a workspace admin to set if inheritance on context is blocked or not. When blocking the parents local role assignments are copied, when unblocking the current folders assignments are removed.

_Creating a draft, but could be merged separately as it  fully implements the `@role_inheritance` endpoint as specified in https://github.com/4teamwork/opengever.core/issues/6167. Checkboxes ticked off accordingly. Didn't get to work on this today unfortunately, so not 100% finsihed._

### DIFF
--- a/docs/public/dev-manual/api/workspace/role_inheritance.rst
+++ b/docs/public/dev-manual/api/workspace/role_inheritance.rst
@@ -1,0 +1,87 @@
+.. _role_inheritance:
+
+Ordnerberechtigungen
+====================
+
+Teamraum-Ordner besitzen grundsätzlich die gleichen Rechte, wie der Teamraum selbst. Die Berechtigungen können jedoch explizit unterbrochen und neu gesetzt werden. Dazu wird der ``@role-inheritance`` Endpoint verwendet.
+
+Vererbungsstatus überprüfen:
+----------------------------
+Ein GET Request auf den ``@role-inheritance`` Endpoint gibt zurück, ob die Vererbung momentan unterbrochen ist oder nicht:
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+       GET /workspace-1/folder-1/@role-inheritance HTTP/1.1
+       Accept: application/json
+
+**Beispiel-Response**:
+
+   .. sourcecode:: http
+
+    HTTP/1.1 200 OK
+    Content-Type: application/json
+
+    {
+      "blocked": false
+    }
+
+Vererbung unterbrechen:
+-----------------------
+Mit einem POST Request kann die Rollenvererbung unterbrochen werden.
+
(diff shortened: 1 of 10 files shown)


opengever/api/configure.zcml
```diff
@@ -422,6 +422,14 @@
       permission="zope2.View"
       />
 
+  <plone:service
+      method="POST"
+      name="@role_inheritance"
```
lukasgraf: Should use hyphens instead of underscores. See
https://plonerestapi.readthedocs.io/en/latest/conventions.html#attribute-names-in-uris

docs/public/dev-manual/api/workspace/role_inheritance.rst
```diff
@@ -0,0 +1,55 @@
+.. _role_inheritance:
+
+Ordnerberechtigungen
+====================
+
+Teamraum-Ordner besitzen grundsätzlich die gleichen Rechte, wie der Teamraum selbst. Die Berechtigungen können jedoch explizit unterbrochen und neu gesetzt werden. Dazu wird der ``@role_inheritance`` verwendet.
```
lukasgraf: verwendet -> Endpoint verwendet

docs/public/dev-manual/api/workspace/role_inheritance.rst
```diff
@@ -0,0 +1,55 @@
...
+
+    HTTP/1.1 200 OK
+    Content-Type: application/json
+
+    {
+      "blocked": false
+    }
+
+Vererbung anpassen:
+-------------------
```
lukasgraf: IMHO this should also mention the behavior when changing `blocked` from `true` to `false`, or the other way. Especially because removing the inheritance-block seems to destroy local role assignments:

```python
        if blocked:
            self._copy_assigments_from_parent_with_local_roles()
        else:
            self._delete_local_role_assignments()
```

opengever/api/configure.zcml
elioschmutz: 👍

docs/public/dev-manual/api/workspace/role_inheritance.rst
elioschmutz: 👍

docs/public/dev-manual/api/workspace/role_inheritance.rst
elioschmutz: 👍

PR #6199, opened 2020-01-15
Await solr to be restored when tearing down test layers

siegy22:
Instead of waiting for solr to have restored a backup when setting up the
testlayer, properly tear it down by waiting for solr to be finished restoring.

A symptom of this issue was, that solr logged:
"Restore in progress. Cannot run multiple restore operationsfor the same core":

See https://ci.4teamwork.ch/builds/302050/tasks/496604

This needs to be backported to 2019.4.x, 2019.5.x and 2019.6.x as we test against these versions in RIS.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2019.5.1 (unreleased)
 ---------------------
 
+- When tearing down test layer, wait for solr to be torn down properly. [siegy]
 - Configure Solr replication handler. [buchi]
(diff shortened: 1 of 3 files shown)


lukasgraf:
@siegy22 I will backport this change to the branches you mentioned, and ping you as soon as I'm done.

lukasgraf:
@siegy22 backports done to the following branches:
- [x] `2019.6-stable`
- [x] `2019.5-stable`
- [x] `2019.4-stable`

PR #6202, opened 2020-01-17
Edit submitted document in proposal

elioschmutz:
Issuer: #6013 

Dieser PR implementiert das Freigeben und Bearbeiten von eingereichten Anhängen:

screen1

## Checkliste


opengever/meeting/tests/test_proposal.py
```diff
@@ -692,6 +694,55 @@ def test_submit_additional_document(self):
...
+
+        browser.visit(submitted_document, view="@@unlock_submitted_document_form")
+        browser.find_button_by_label('Unlock').click()
+
+        statusmessages.assert_message('Document has been unlocked')
+
+        self.assertFalse(lockable.locked(), "Submitted document should be unlocked")
+        self.assertIsNotNone(
+            SubmittedDocument.query.get_by_target(submitted_document),
+            "Submitted document should be unlinked")
```
njohner: Shouldn't the message be "should not be unlinked"?

opengever/meeting/tests/test_proposal.py
```diff
@@ -692,6 +694,55 @@ def test_submit_additional_document(self):
...
+        with self.assertRaises(NotFound):
+            browser.visit(self.document, view="@@unlock_submitted_document_form")
+
+    @browsing
+    def test_unlock_submitted_document_returns_no_content_if_ajax_load_is_true(self, browser):
+        self.login(self.administrator, browser)
+        ori_document = self.subdocument
+        with self.observe_children(self.submitted_proposal) as children:
+            with self.login(self.dossier_responsible):
+                self.proposal.submit_additional_document(ori_document)
```
njohner: Can't you use `self.submitted_document` for this test instead of first submitting an additional document?

elioschmutz: Unfortunately there is no such variable...

My implementation is a copy of the other tests in this file...

njohner: Yes but in your other test you use the already existing submitted document:
```
    @browsing
    def test_unlock_submitted_document_creates_history_entry(self, browser):
        self.login(self.committee_responsible, browser)

        submitted_document = SubmittedDocument.query.get_by_source(
            self.proposal, self.document).resolve_submitted()
```

I think there is one, it's just not assigned to an attribute of the integration test case.

elioschmutz: You're totally right!

But, I can't get the test passing when using the already submitted document. I can't figure out why it's not working.

How to reproduce it:

1. create a new proposal with the builder factory and directly submit the proposal as it is done in the test-fixture: https://github.com/4teamwork/opengever.core/blob/master/opengever/testing/fixtures.py#L1021 (or just use the already created self.proposal)
2. try to unlock the submitted version of `self.document`  (as you proposed)
3. run the test `test_unlock_submitted_document_creates_history_entry `=> the test will fail 

As soon as I try to test it with the already submitted document, I have to commit the transaction in the `unlock_submitted_document_form`. Otherwise the document is still locked for the assertion.

I already used too much time to debug it. I can't figure out why I have to commit the transaction with the directly submitted document and not with the additional submitted proposals. Btw. TTW it's working.

Do you have any pointer for me? or should we stop debugging and shipping it?

PR #6208, opened 2020-01-22
Speed up validation of dossier resolution preconditions.

njohner:
This is for https://extranet.4teamwork.ch/support/gever-st-gallen/tracker/908

It seems that checking the preconditions for resolving a dossier can be very slow. After inspecting the code I found:
- A bug in the recursion in `ResolveConditions.check_end_dates`, which led to visiting certain objects several times and hence calculating the `earliest_possible_enddate` several times per object in the main dossier.
- A bug in `ResolveConditions.is_all_checked_in`, which only checked documents in the main dossier and within to a depth of 2.

I corrected these two bugs and optimised `DossierContainer.has_valid_enddate` to only calculate `earliest_possible_end_date` when necessary, which is probably almost never in most cases.

This should speed up the precondition calculation quite a bit and I also ensured that the preconditions are only calculated once in the `DossierResolveView`.

## Checkliste

### DIFF
--- a/opengever/dossier/resolve.py
+++ b/opengever/dossier/resolve.py
@@ -227,33 +227,38 @@ class DossierResolveView(BrowserView, DossierResolutionStatusmessageMixin):
 
         resolve_manager = LockingResolveManager(self.context)
 
-        # Validate preconditions early. This is so we don't redirect to the
-        # archive form (if filing number feature enabled) in a case where
-        # it will fail anyway because of violated preconditions.
-        #
-        # XXX: This will validate preconditions *twice* though (the second
-        # time via resolve_manager.resolve()). This should eventually be
-        # cleaned up so we don't unnecessarily validate preconditions multiple
-        # times.
+        # If filing number feature is enabled, we redirect to an additional
+        # archive form that needs to be filled out first. The actual resolving
+        # will then be triggered from that form.
+        if resolve_manager.is_archive_form_needed():
+            # Validate preconditions early. This is so we don't redirect to the
+            # archive form (if filing number feature enabled) in a case where
+            # it will fail anyway because of violated preconditions.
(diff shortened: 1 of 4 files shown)


opengever/dossier/resolve.py
```diff
@@ -227,34 +227,39 @@ def __call__(self):
...
+                return self.show_invalid_end_dates(titles=exc.invalid_dossier_titles)
+
             archive_url = '/'.join((self.context_url, 'transition-archive'))
             return self.redirect(archive_url)
 
-        # All good, proceed with resolving the dossier.
         try:
             resolve_manager.resolve()
 
+        except PreconditionsViolated as exc:
```
deiferni: i'm not sure why these exceptions are needed here, could you explain/check?

njohner: Before we were first checking the preconditions (`resolve_manager.resolver.raise_on_failed_preconditions()`) and showing the error messages for `PreconditionsViolated` and `InvalidDates` if necessary, then resolving the dossier (`resolve_manager.resolve()`), which calls the `raise_on_failed_preconditions` again, but we could rely on the fact that this would not raise `PreconditionsViolated` or `InvalidDates`, so we only had to check for `AlreadyBeingResolved`. Now that we don't first call `raise_on_failed_preconditions`, we need to check for all possible exceptions.

deiferni: uhm, ok. would it help to put this into a comment in the code 🤔? what do you think?

njohner: hum. I think it only bothers you because it was added here and the exceptions were uncaught before. But if you just read the code it's pretty clear that we are trying to resolve the dossier and showing error messages if the dossier cannot be resolved...

Issue #6204, opened 2020-01-21
Unicode Fehler bei nicht-ASCII Zeichen in Feld "Bearbeitungsinformation"


lukasgraf:
_Harvest: `[3.2.2.1] OneGov GEVER` - `Softwarepflege (Refactoring/Bug)`_

Nicht-ASCII Zeichen im Feld "Bearbeitungsinformation" (`public_trial_statement `) führen zu einem Unicode Fehler beim Darstellen des Dokuments.

### Steps to reproduce

- Das im Sentry-Issue angegebene Dokument öffnen

https://sentry.4teamwork.ch/sentry/onegov-gever/issues/58646/

Siehe auch: https://basecamp.com/2768704/projects/14549453/todos/408873015

njohner:
It seems the field is saved as unicode on objects older than 2016, exact date being hard to determine. This has nothing to do with the version bump, as it is already broken on the previous versions. See for example ska-arch/ordnungssystem2/leitung/dossier-2158/dossier-2159/document-8631 on zug PROD which still runs `2019.2.6`.

Proposed solution is an upgrade step to fix the older objects. This should be fairly cheap as it only affects older objects, of which there are rather few.

njohner:
newest object with wrong field type dates from 2015/07/24 and the first one with correct type is dates from 2015/07/28. So it seems it was fixed with update from GEVER 3.1 to GEVER 3.4: https://github.com/4teamwork/opengever.zug/pull/181

Here a commit that hints towards encoding issue with the public_trial_statement:
https://github.com/4teamwork/opengever.core/pull/473/files

njohner:
Fixed the one object they reported with:
```
from opengever.maintenance import dm; dm()
from plone import api
portal = api.portal.get()
from zope.component.hooks import setSite; setSite(portal)

obj_url = "ska-arch/ordnungssystem2/oeffentlichkeit/allgemeines-und-uebergreifendes/dossier-148/document-409"
obj = portal.unrestrictedTraverse(obj_url)
obj.public_trial_statement = obj.public_trial_statement.encode('utf-8')
```

njohner:
PR with proposed upgrade step to fix this for Zug

lukasgraf:
Upgrade-Step needs backport to `2019.6-stable` branch.

njohner:
Backported and released `2019.6.2`

PR #6215, opened 2020-01-24
Add upgrade step to correct public_trial_statement type.

njohner:
On some older objects public_trial_statement is saved as unicode on the object which will lead to errors if it contains non ASCII characters. We did not determine in which commit this was fixed and even then we would still need to find out for each deployment when that fix was introduced. Nevertheless we only fix objects older than 2016 as this seems to be a problem only for Zug, and it seems the issue was fixed there mid 2015.

for https://github.com/4teamwork/opengever.core/issues/6204

## Checkliste

### DIFF
--- a/opengever/core/upgrades/20200124171823_correct_public_trial_statement_type/upgrade.py
+++ b/opengever/core/upgrades/20200124171823_correct_public_trial_statement_type/upgrade.py
@@ -1,0 +1,39 @@
+from Acquisition import aq_base
+from DateTime import DateTime
+from ftw.upgrade import UpgradeStep
+from Products.CMFDiffTool.utils import safe_utf8
+import pkg_resources
+
+
+class CorrectPublicTrialStatementType(UpgradeStep):
+    """Correct public_trial_statement type.
+
+    On some older objects public_trial_statement is saved as unicode
+    on the object which will lead to errors if it contains non ASCII
+    characters.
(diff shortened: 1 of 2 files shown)


njohner:
I could not test yet that the test to see whether we are on Zug deployment works. On the other hand I wonder whether we shouldn't simply make a maintenance script for that

opengever/core/upgrades/20200124171823_correct_public_trial_statement_type/upgrade.py
```diff
@@ -0,0 +1,36 @@
...
+        # we only run this upgrade for Zug for now.
+        try:
+            pkg_resources.get_distribution('opengever.zug')
+        except pkg_resources.DistributionNotFound:
+            return
+
+        query = {'portal_type': ['opengever.document.document', 'ftw.mail.mail'],
+                 'created': {'query': DateTime('2016-01-01'),
+                             'range': 'max'}}
+        for obj in self.objects(query, 'Correct public_trial_statement type'):
```
lukasgraf: I would prefer if this check could be made a bit more robust. Something along these lines:
```python
pts_value = getattr(aq_base(obj), 'public_trial_statement', None)
if pts_value and isinstance(pts_value, unicode):
    obj.public_trial_statement = safe_utf8(pts_value)
```

PR #6218, opened 2020-01-28
Fix JS ordering issue: define overlayhelpers.js position.

njohner:
https://github.com/4teamwork/opengever.core/pull/6202 made an ordering issue in the `jsregistry.xml` surface. Indeed not all installations seem to have `plone.app.jquerytools.overlayhelpers.js` positioned before `opengever.base/prepoverlay.js`, but functions from `overlayhelpers.js` are used in `prepoverlay.js`. Here we correct that ordering issue by ensuring the jquerytools are loaded early, placing them and the things they depend on at the top of the registry.

## Checkliste

- [ ] Sind UpgradeSteps `deferrable`, oder können gewisse Schritte des Upgrades konditional ausgeführt werden? Not deferrable

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 
 - Add 'filing_no' field to Solr schema. [lgraf]
 - Add upgrade step to correct public_trial_statement type. [njohner]
+- Fix JS ordering issue: define overlayhelpers.js position. [njohner]
 - Fix volatile related proposal documents. [elioschmutz]
 - Add a button to create a task from a proposal. [elioschmutz]
 - Allow to unlock and edit submitted documents in a submitted proposal. [elioschmutz]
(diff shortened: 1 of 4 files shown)


deiferni:
Mental note, discussed with @lukasgraf yesterday. Maybe it would be better to set the order for all items again. We've done this before with https://gist.github.com/bierik/ece24ad316424d6d4909efd272940fb4

njohner:
Had a look at it and discussed with @bierik . This requires to find out all the new dependencies that we have added and check on what they depend themselves. But there does not seem to be any good way to determine on what a js file depends, except looking at the code and trying to find out from where the used functions come. I don't really feel like doing that for all the new dependencies, especially with my poor knowledge of javascript, that would probably cost me a day and still be full of guesses and errors.

njohner:
For safety I add `jquery` in the same position it is in a newly setup system, making sure `plone.app.jquerytools.js` and `plone.app.jquerytools.overlayhelpers.js` and inserted just after.

PR #6236, opened 2020-02-04
Fix js ordering issue.

njohner:
Seems that the fix of the JS ordering issue in https://github.com/4teamwork/opengever.core/pull/6218 does not work properly for new deployments. I was sure I had tested both the upgrade and creation of a new deployment, but it seems it is not working for new deployments. I now really fix the ordering of the first 6 entries in the registry.

As this is pretty hard to test because we don't really know what the order is on different installation, it would be nice if someone could test the upgrade step, as well as setting up a new deployment with this fix.

## Checkliste


njohner:
Just had a look at `js_graph.py` and there are a lot of jsfiles that are not referenced in there. Here is the list:
```
'unorm', 'notifications', 'settings', 'inline_validation', 'contact', 'ogds_base', 'SelectAutocomplete', 'meetingtemplate', 'proposal', 'disposition', 'download_confirmation', 'cancel_checkout_confirmation', 'officeconnector', 'edit_note', 'tree', 'init_tree', 'advanced_search', 'statefilter', 'startredirect', 'bumblebee-cable', 'bumblebee', 'swfobject', 'formwidget-autocomplete', 'modernizr', 'jquery-migrate.min', 'keywordwidget', 'logoutClient', 'logout', 'generateToc', 'Favorites', 'config', 'vue.min', 'axios.min', 'sharing', 'dropdown', 'bumblebee-teamraum', 'teamraum', 'colorpicker', 'ftw.showroom.handlebars.js'
```

opengever/core/upgrades/20200204091707_define_overlayhelper_order_in_js_registry/jsregistry.xml
```diff
@@ -0,0 +1,51 @@
+<object name="portal_javascripts" meta_type="JavaScripts Registry">
+
+  <javascript
+      cacheable="True"
```
deiferni: do you really need to set these attributes for reordering? if not, please omit them and just define `id` and `insert-` as with the three last entries.

njohner: Indeed it seems we only need the `id` and `insert`. Updated accordingly

PR #6240, opened 2020-02-05
Only return create_task in actions endpoint on dossiers and tasks.

njohner:
We only return `create_task` in the `@actions` endpoint on dossiers and tasks. Before the action was available everywhere where the user has the `Add task` permission and it was then specifically disabled on some different content_types in the tabs. We now only return the action when it actually should be displayed so that the frontend can rely on the presence of the action in the `@actions` endpoint.

Note that we decided to not display the `new_task_from_document` action  (in the new UI) on documents inside tasks, although the `create_task` action is available on the document listing inside tasks.

## Checkliste
- [ ] Sind UpgradeSteps `deferrable`, oder können gewisse Schritte des Upgrades konditional ausgeführt werden? I guess we could make it deferrable, but there really isn't any benefits to it

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -8,6 +8,7 @@ Changelog
 - Implement the workspace client to make requests to a teamraum from GEVER. [elioschmutz]
 - Implement the workspace client authentication flow. [elioschmutz]
 - Handle deployments with no repository in navigation endpoint. [njohner]
+- Only return create_task in actions endpoint on dossiers and tasks. [njohner]
 - Add documentation for sharing endpoint. [njohner]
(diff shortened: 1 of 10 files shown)


njohner:
After reflection, I'm not sure we need this. We already have a `create_task` action which does exactly that in the classic UI. 😓

njohner:
So tasks are allowed in the following portal_types:
- `opengever.dossier.businesscasedossier`
- `opengever.meeting.meetingdossier`
- `opengever.task.task`
Forbidden everywhere else. The `opengever.task: Add task` permission is actually available on many other portal_types, though adding a task there will throw an error. 
So for the action we both check the permission and that we are in either a dossier or a task.

PR #6245, opened 2020-02-06
Handle path depth when filtering a table.

njohner:
In the InboxDocuments tab, the depth is set to 1 to avoid listing documents in forwardings. This depth restriction was not handled when the query is sent to solr (i.e. when documents are filtered), leading to more documents being displayed when applying a filter.

For https://basecamp.com/2768704/projects/14549453/todos/410114932

## Checkliste

### DIFF
--- a/opengever/tabbedview/catalog_source.py
+++ b/opengever/tabbedview/catalog_source.py
@@ -78,8 +78,21 @@ class GeverCatalogTableSource(FilteredTableSourceMixin, CatalogTableSource):
             elif key == 'sort_on' or key == 'sort_order':
                 continue
             elif key == 'path':
+                path_query = value.get('query')
                 filters.append(u'path_parent:{}'.format(
-                    escape(value.get('query'))))
+                    escape(path_query)))
+
+                depth = value.get('depth', -1)
+                try:
+                    depth = int(depth)
+                except ValueError:
+                    depth = -1
+                if depth > 0:
+                    # path starts with /, so splitting gives context_depth + 1
+                    context_depth = len(path_query.split("/")) - 1
+                    max_path_depth = context_depth + depth
+                    filters.append(u'path_depth:[* TO {}]'.format(max_path_depth))
+
             elif isinstance(value, (list, tuple)):
                 filters.append(u'{}:({})'.format(
                     key, escape(' OR '.join(value))))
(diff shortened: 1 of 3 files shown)


opengever/tabbedview/catalog_source.py
```diff
@@ -78,8 +78,21 @@ def solr_results(self, query):
...
+                depth = value.get('depth', -1)
+                try:
+                    depth = int(depth)
+                except ValueError:
+                    depth = -1
+                if depth > 0:
+                    # path starts with /, so splitting gives context_depth + 1
+                    context_depth = len(path_query.split("/")) - 1
+                    max_path_depth = context_depth + depth
+                    filters.append(u'path_depth:[* TO {}]'.format(max_path_depth))
```
elioschmutz: Are you aware of that?:

[Plone-REST-API](https://plonerestapi.readthedocs.io/en/latest/searching.html)
> A search invoked on a container will by default include that container itself as part of the search results. This is the same behavior as displayed by ZCatalog, which is used internally. If you add the query string parameter depth=1 to your search, you will only get immediate children of the container, and the container itself also won’t be part of the results. See the Plone docs on [searching for content within a folder](https://docs.plone.org/develop/plone/searching_and_indexing/query.html#searching-for-content-within-a-folder). for more details.

njohner: We are already excluding the search root in the solr query (we are setting `path_parent` on line 82)

elioschmutz: What I mean is, only if you use 1 as the depth, the parent will be excluded:

Bildschirmfoto 2020-02-06 um 13 53 25

I don't think, that this will be the same with solr. Should we handle this like the plone catalog?

njohner: The default for the search root exclusion in the `GeverCatalogTableSource` is `True`. And in `solr_results` it is always excluded. So while I think you are right that when `exclude_search_root` is set to `False` in the query, we will not handle it properly in `solr_results`, I think that does not really happen in GEVER and it is a separate bug.

elioschmutz: Ok 👍

njohner: I'll make an issue to have a closer look at this

njohner:
We had the same issue with documents in template folders

PR #6254, opened 2020-02-10
Add linked workspace

elioschmutz:
Issuer: #6154 

Dieser PR implementiert einen neuen Endpoint `@create-linked-workspace`

Der Endpoint erstellt ein neuer Arbeitsraum und verknüpft diesen mit dem Dossier, auf welchem der Request abgesetzt wurde.

## Checkliste

### DIFF
--- a/opengever/api/configure.zcml
+++ b/opengever/api/configure.zcml
@@ -12,6 +12,7 @@
   <include package="opengever.trash" file="permissions.zcml" />
   <include package="opengever.webactions" file="permissions.zcml" />
   <include package="opengever.workspace" file="permissions.zcml" />
+  <include package="opengever.workspaceclient" file="permissions.zcml" />
 
   <include package=".schema" />
 
@@ -565,6 +566,14 @@
       for="Products.CMFPlone.interfaces.IPloneSiteRoot"
       factory=".globalindex.GlobalIndexGet"
       permission="zope2.View"
+      />
+
+  <plone:service
+      method="POST"
+      name="@create-linked-workspace"
+      for="opengever.workspaceclient.interfaces.ILinkedWorkspacesMarker"
+      factory=".linked_workspaces.LinkedWorkspacesPost"
+      permission="opengever.workspaceclient.UserWorkspaceClient"
       />
 
 </configure>
--- a/opengever/api/configure.zcml
+++ b/opengever/api/configure.zcml
@@ -571,7 +571,7 @@
   <plone:service
       method="POST"
       name="@create-linked-workspace"
-      for="opengever.workspaceclient.interfaces.ILinkedWorkspacesMarker"
+      for="opengever.dossier.behaviors.dossier.IDossierMarker"
       factory=".linked_workspaces.LinkedWorkspacesPost"
       permission="opengever.workspaceclient.UserWorkspaceClient"
       />
--- a/opengever/api/linked_workspaces.py
+++ b/opengever/api/linked_workspaces.py
@@ -15,6 +15,9 @@ class LinkedWorkspacesPost(Service):
         if not is_workspace_client_feature_available():
             raise NotFound
 
+        if self.context.is_subdossier():
+            raise NotFound
+
         # Disable CSRF protection
         alsoProvides(self.request, IDisableCSRFProtection)
 
--- a/opengever/api/linked_workspaces.py
+++ b/opengever/api/linked_workspaces.py
@@ -1,0 +1,24 @@
+from opengever.workspaceclient import is_workspace_client_feature_available
+from opengever.workspaceclient.interfaces import ILinkedWorkspaces
+from plone.protect.interfaces import IDisableCSRFProtection
+from plone.restapi.deserializer import json_body
+from plone.restapi.services import Service
+from zExceptions import NotFound
+from zope.interface import alsoProvides
+
+
+class LinkedWorkspacesPost(Service):
+    """API Endpoint to add a new linked workspace.
+    """
+
+    def reply(self):
+        if not is_workspace_client_feature_available():
+            raise NotFound
+
+        # Disable CSRF protection
+        alsoProvides(self.request, IDisableCSRFProtection)
+
+        # Validation will be done on the remote system
+        data = json_body(self.request)
+
+        return ILinkedWorkspaces(self.context).create(**data)
--- a/opengever/api/tests/test_linked_workspaces.py
+++ b/opengever/api/tests/test_linked_workspaces.py
@@ -1,0 +1,78 @@
+from ftw.testbrowser import browsing
+from opengever.workspaceclient.tests import FunctionalWorkspaceClientTestCase
+from plone import api
+from zExceptions import NotFound
+from zExceptions import Unauthorized
+import json
+
+
+def fix_publisher_test_bug(browser, obj):
+    # We need to make a GET request before using the `@create-linked-workspace`
+    # endpoint. This is a workaround to fix a weird issue where we get
+    # a silent `ConflictError`. ZPublisher will then automatically
+    # republish the request although the previous POST request has
+    # already been handled successfully, effectively creating the new
+    # workspace twice
+    #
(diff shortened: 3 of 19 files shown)


opengever/api/tests/test_linked_workspaces.py
```diff
@@ -0,0 +1,57 @@
...
+class TestLinkedWorkspaces(FunctionalWorkspaceClientTestCase):
+
+    @browsing
+    def test_create_linked_workspace(self, browser):
+        browser.login()
+
+        workspaces_before_add = self.workspace_root.listFolderContents()
+        with self.workspace_client_env():
+            # We need to make a GET request before using the `@create-linked-workspace`
+            # endpoint. This is a workaround to fix a weired issue where we get
```
njohner: weired -> weird

opengever/api/tests/test_linked_workspaces.py
```diff
@@ -0,0 +1,57 @@
...
+            # a silent `ConflictError`. ZPublisher will automatically republish
```
njohner: ZPublisher will then automatically republish the request although the previous POST request has already been handled successfully, effectively creating the new workspace twice

opengever/api/tests/test_linked_workspaces.py
```diff
@@ -0,0 +1,57 @@
...
+            # This issue only appears in a testing environment.
+            browser.open(self.dossier, method='GET', headers={'Accept': 'application/json'})
+            browser.open(
+                self.dossier.absolute_url() + '/@create-linked-workspace',
+                data=json.dumps({"title": "My linked workspace"}),
+                method='POST',
+                headers={'Accept': 'application/json',
+                         'Content-Type': 'application/json'},
+            )
+        workspaces_after_add = self.workspace_root.listFolderContents()
```
njohner: I think that test would be easier to read if you used `with self.observe_children(self.workspace_root) as children`... When you would not need `workspaces_before_add` and `workspaces_after_add`

opengever/workspaceclient/tests/test_client.py
```diff
@@ -35,3 +36,15 @@ def test_search_for_objects(self):
         self.assertEqual(1, response.get('items_total'))
         self.assertEqual(self.workspace.absolute_url(),
                          response.get('items')[0].get('@id'))
+
+    def test_create_workspace(self):
+        with self.workspace_client_env() as client:
+            response = client.create_workspace(title=u"My new w\xf6rkspace")
```
njohner: Here too `observe_children` seems more natural

elioschmutz:
@njohner 👍 I updated the PR.

Thats true, it should not be available for subdossiers. Please, see my last commit which excludes subdossiers. I'm really not sure if this is the right way. I decided to use a behavior for the LinkedWorkspaces-adapter in the previous PR. A dossier is providing this behavior. So I can register Views to the behavior marker interface (`ILinkedWorkspacesMarker`). That means, that this View should be type independent. But now, I have to implement a type specific check in the REST-Endpoint View. That seems to be a bad pattern...

The only way to get rid of this bad pattern is to directly registering the view to the `IDossierMarker` interface than the behavior interface. But if I do so, having the behavior seems to be unnecessary.

What do you think? Should I let it how it is? Or should I remove the behavior and implement it only for Dossiers?

elioschmutz:
@njohner thanks for your clarifications!

I updated the implementation with the following changes (see my last two commits):
- remove the behavior and use an adapter instead
- register the rest-service for dossiers only
- the adapter will raise a `CompoentLookupError` if trying to lookup it with a subdossier

⚠️ I do not change the store itself. I think, the store should not check for subdossiers. it just stores workspaces for a context in its annotations. thats all. The `ILinkedWorkspaces` adapter is the public interface to use everywhere in GEVER to interact with linked workspaces. So preventing it there seems to be the right place for me. Do you agree?

opengever/workspaceclient/tests/test_linked_workspaces.py
```diff
@@ -3,6 +3,7 @@
 from opengever.workspaceclient.interfaces import ILinkedWorkspaces
 from opengever.workspaceclient.tests import FunctionalWorkspaceClientTestCase
 from plone import api
+from zope.component.interfaces import ComponentLookupError
```
njohner: You are not using this

opengever/workspaceclient/tests/test_linked_workspaces.py
```diff
@@ -88,3 +89,9 @@ def test_create_workspace_will_store_workspace_in_the_storage(self):
             workspace = children['added'].pop()
             self.assertEqual([workspace.absolute_url()],
                              [ws.get('@id') for ws in manager.list()])
+
+    def test_subdossiers_do_not_provided_linked_workspaces(self):
+        subdossier = create(Builder('dossier').within(self.dossier))
+
+        with self.assertRaises(TypeError):
```
njohner: hum why isn't that `ComponentLookupError`?

opengever/api/linked_workspaces.py
```diff
@@ -0,0 +1,27 @@
...
+class LinkedWorkspacesPost(Service):
+    """API Endpoint to add a new linked workspace.
+    """
+
+    def reply(self):
+        if not is_workspace_client_feature_available():
+            raise NotFound
+
+        if self.context.is_subdossier():
+            raise NotFound
```
njohner: I think now that https://github.com/4teamwork/opengever.core/pull/6235 is merged, you should check the permission here

elioschmutz: Sure? Currently, a 401 will be raised because the `WorkspaceClient` is protected already:

Bildschirmfoto 2020-02-14 um 20 03 44

Is it necessary to check it also in the rest-service itself? If so, I'll update it.

opengever/workspaceclient/tests/test_linked_workspaces.py
elioschmutz: This is because I call the interface to get the adapter. This comes from the [`zope.component adapter_hook`](https://github.com/zopefoundation/zope.component/blob/3.9/src/zope/component/tests.py#L380).

But it's true, it's weird. I changed it to call the `getAdapter` function which is then raising the proper exception.

opengever/workspaceclient/tests/test_linked_workspaces.py
elioschmutz: 👍 now I'm using it, because I changed the test implementation. See my next comment.

elioschmutz:
@njohner could you take a final look?

The permission-check makes me think because I’m not sure if we really have to explicitly validate the permission in the view itself. I think it depends on the intention of this permission. I understand it as a general permission to protect the use of the `WorkspaceClient` itself, means, requests to the teamraum deployment. Because I use the `Client` in each view, it is protected anyway. But this would mean that it is theoretically possible to have a linked workspace view which is available also for users without the permission. But this seems to be ok, because the permission protects the client, not the linked workspaces.

What do you think?

njohner:
Isn't it protected by the wrong permission now? You use `Add portal content`, but you basically check whether the user is allowed to add a portal content in the dossier, whereas he is actually trying to add a new workspace somewhere completely different. I'd simply change that permission to the relevant one, which for now is the `opengever.workspaceclient: Use Workspace Client`. I think that makes more sense to me.

elioschmutz:
@njohner 👌 updated

opengever/api/configure.zcml
```diff
@@ -567,4 +567,12 @@
       permission="zope2.View"
       />
 
+  <plone:service
+      method="POST"
+      name="@create-linked-workspace"
+      for="opengever.dossier.behaviors.dossier.IDossierMarker"
+      factory=".linked_workspaces.LinkedWorkspacesPost"
+      permission="cmf.AddPortalContent"
```
njohner: It's still on `cmf.AddPortalContent`.

elioschmutz: oh man 🙈 just updated

PR #6264, opened 2020-02-12
Add dossier title to task activity summaries.

njohner:
To give more context to the users we add the dossier title in task activities. 
As we want to add the information about the dossier only for certain types of activities, adding a new column in the `Activity` model that would be empty for most activities did not seem to make too much sense, especially that then `Notification` serialization would also need to support cases with or without that field and so on. 

Instead we simply add the dossier title in the `summary` field for the task activities. I chose to add it in the `summary` and not the `description` field as the `description` is only displayed in the notification mails, but not in the badges and digests.

The tabular view (`mynotifications` tab) also displays notifications, but it does neither display the summary nor the description, hence the dossier will not appear there. As already said, I don't think it would make too much sense to add a column here that would be mainly empty.

**Badge notification**
Screenshot 2020-02-14 at 12 03 09

**New task created E-mail**
Screenshot 2020-02-14 at 12 04 14

For https://github.com/4teamwork/opengever.sg/issues/376

## Checkliste

### DIFF
--- a/opengever/activity/browser/views.py
+++ b/opengever/activity/browser/views.py
@@ -99,7 +99,7 @@ class NotificationView(BrowserView):
             data.append({
                 'title': notification.activity.title,
                 'label': notification.activity.label,
-                'summary': notification.activity.summary,
+                'summary': notification.activity.summary.splitlines(),
                 'created': notification.activity.created.astimezone(
                     pytz.UTC).isoformat(),
                 'link': resolve_notification_url(notification),
--- a/opengever/activity/mailer.py
+++ b/opengever/activity/mailer.py
@@ -103,6 +103,10 @@ class Mailer(object):
         if data.get('description'):
             data['description'] = data.get('description').splitlines()
 
+        # Break (potential) summary up into a list of lines
+        if data.get('summary'):
+            data['summary'] = data.get('summary').splitlines()
+
         html = self.prepare_html(data)
         msg.attach(MIMEText(html.encode('utf-8'), 'html', 'utf-8'))
         return msg
--- a/opengever/task/activities.py
+++ b/opengever/task/activities.py
@@ -5,6 +5,7 @@ from opengever.activity.model.notification import Notification
 from opengever.activity.roles import TASK_OLD_RESPONSIBLE_ROLE
 from opengever.activity.roles import TASK_REMINDER_WATCHER_ROLE
 from opengever.base.model import get_locale
+from opengever.globalindex.model.task import Task
 from opengever.ogds.base.actor import Actor
 from opengever.ogds.base.actor import SYSTEM_ACTOR_ID
 from opengever.task import _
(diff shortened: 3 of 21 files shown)


jone:
@njohner is it difficult to also include the dossier title in mail subject?

njohner:
@jone Easy way would be to add it to the `title` field of the activity, but then it will be in the title of mails and badges and in digests in the title of individual notifications. In that case it would make sense to remove it from the summary and put it in the title instead.
If we want it only for mail titles then we would need to add a new translatable column in the activity table.

jone:
@njohner I have had a chat with the customer: it would help the users the most when the dossier title is visible in the mail subject.
Thoughts:
- I'd prefix the dossier-title in github-style (`[Dossier title] Task title`)
- The dossier title should be cropped so that the task title is visible even with a very long dossier title in average environments (Outlook, GMail, ..)
- I think would be more consistent to have the dossier title in all appearances (app notifications, digests)

opengever/task/activities.py
```diff
@@ -5,16 +5,51 @@
...
 
-class TaskAddedActivity(BaseActivity):
+class BaseTaskActivity(BaseActivity):
+
+    @property
+    def dossier_title(self):
+        """If the task is in a subdossier, return its title, otherwise
+        return the title of the main dossier (or inbox for a forwarding)"""
+
+        if isinstance(self.context, Task):
```
deiferni: a more object-oriented option would be to also implement the getters on `Task`. That way the activity does not have to know things about the type of the object passed to it and you can get rid of the isinstance check here.

opengever/activity/browser/views.py
```diff
@@ -99,7 +99,7 @@ def dump_notifications(self, notifications):
             data.append({
                 'title': notification.activity.title,
                 'label': notification.activity.label,
-                'summary': notification.activity.summary,
+                'summary': notification.activity.summary.splitlines(),
```
deiferni: couldn't the summary be `None`?

opengever/activity/mailer.py
```diff
@@ -103,6 +103,10 @@ def prepare_mail(self, subject=u'', to_userid=None, to_email=None,
         if data.get('description'):
             data['description'] = data.get('description').splitlines()
 
+        # Break (potential) summary up into a list of lines
+        if data.get('summary'):
+            data['summary'] = data.get('summary').splitlines()
```
deiferni: maybe we could build this so the model returns a list of summary items (potentially empty) that can then be handled by mailer and ui and ... instead of splitting the string in various places?

njohner:
Superseded by https://github.com/4teamwork/opengever.core/pull/6270

opengever/task/activities.py
njohner: Included in https://github.com/4teamwork/opengever.core/pull/6270

PR #6269, opened 2020-02-17
Allow administrators to deactivate dossiers.

njohner:
Administrators cannot deactivate dossiers although they can for example activate a deactivated dossier. This is confusing for the clients and leads to support issues. I don't think that this is an intended feature so I would simply give that permission to administrators.

For https://extranet.4teamwork.ch/support/gemeinde-muri-guemligen/tracker/9

## Checkliste

- [x] Sind UpgradeSteps `deferrable`, oder können gewisse Schritte des Upgrades konditional ausgeführt werden? Not deferrable


opengever/core/upgrades/20200217115450_allow_administrators_to_deactivate_dossiers/upgrade.py
```diff
@@ -0,0 +1,9 @@
+from ftw.upgrade import UpgradeStep
+
+
+class AllowAdministratorsToDeactivateDossiers(UpgradeStep):
+    """Allow administrators to deactivate dossiers.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
```
deiferni: i think you also need to `update_workflow_security` here

njohner: Ah sorry, you're right

njohner: Actually I don't think we need to. This is a guard expression, not a permission stored on the object.

deiferni: oh yes, you're right.

njohner:
Same here, client deemed this as not urgent, so no backport needed

PR #6270, opened 2020-02-17
Add dossier title to task activity titles.

njohner:
To give more context to the users we add the dossier title in task activities. 
As we want to add the information about the dossier only for certain types of activities, adding a new column in the `Activity` model that would be empty for most activities did not seem to make too much sense, especially that then `Notification` serialization would also need to support cases with or without that field and so on. 

Instead we simply add the dossier title in the `title` field for the task activities in the form `[dossier title] task title`. Adding it in the title will make it visible in the badges, notification mails and digests.

The tabular view (`mynotifications` tab) also displays the notification title, so there too, the dossier name will now appear for task activities.

While I was at it I made a few further changes:
- The notification mail template displayed the title of the context thrice (once in the mail subject, once as title in the mail content and a third time as link in the mail content) for normal activities and four times for `TaskAddedActivity`. This is now cut down to twice (once in the subject and once as linked title in the mail content) for normal activities.
- For `TaskAddedActivity`, the description already displayed the dossier title, in the style `Dossier: dossier title`, but this was actually not always the dossier title but the title of the parent, which could also be a main task. We now correctly display the dossier title there (or inbox) and added the main task title when a subtask was added.
- I did leave the dossier title and task title in the description field of the `TaskAddedActivity`, which means that it now appears twice in a combined style (`[dossier title] task title`), once in the subject of the E-mail and once as title for the E-mail content. And then each field also appears separately in the E-mail content in the description of the task (see screenshots below)

**Badge notification**
Screenshot 2020-02-18 at 17 45 55

**TaskCreatedActivity E-mails**
*As it was previously, task title appears 3 times and main task title is shown as `Dossiertitle`*
Screenshot 2020-02-18 at 17 44 56

*As it is now, dossier and task title appear twice combined (with dossier title cropped) and then Dossier title appears once uncropped, task title appears once on its own and main task title appears correctly labelled*
Screenshot 2020-02-18 at 17 43 21

For https://github.com/4teamwork/opengever.sg/issues/376
Supersedes https://github.com/4teamwork/opengever.core/pull/6264

## Checkliste

### DIFF
--- a/opengever/activity/tests/test_digest.py
+++ b/opengever/activity/tests/test_digest.py
@@ -75,7 +75,7 @@ class TestDigestMail(IntegrationTestCase):
         browser.open_html(str(messages[0].get_payload()[0]))
 
         self.assertEquals('Oct 16, 2017', browser.css('table p').text[0])
-        self.assertEquals(['Daily Digest for B=C3=A4rfuss K=C3=A4thi'],
+        self.assertEquals(['= Daily Digest for B=C3=A4rfuss K=C3=A4thi ='],
                           browser.css('h1').text)
         self.assertEquals(['Bitte =C3=84nderungen nachvollziehen'],
                           browser.css('h2 a').text)
--- a/opengever/activity/tests/test_mail.py
+++ b/opengever/activity/tests/test_mail.py
@@ -56,11 +56,18 @@ class TestEmailNotification(IntegrationTestCase):
     @browsing
     def test_subject_is_title(self, browser):
         self.login(self.dossier_responsible, browser)
+
+        # We change the dossier title to ASCII-only, makes it easier
+        # to test the mail subject
+        self.dossier.title = "Dossier title"
+
         self.create_task_via_browser(browser)
         process_mail_queue()
 
         mail = email.message_from_string(Mailing(self.portal).pop())
-        self.assertEquals('GEVER Activity: Test Task', mail.get('Subject'))
+
+        self.assertEquals(
+            'GEVER Activity: [Dossier title] Test Task', mail.get('Subject'))
 
     @browsing
     def test_notification_summary_is_split_into_paragraphs(self, browser):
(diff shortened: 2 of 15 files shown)


opengever/task/activities.py
```diff
@@ -12,9 +12,41 @@
...
+
+        container_title = safe_unicode(subdossier_title or dossier_title)
+        return container_title
+
+    @property
+    def title(self):
+        task_title = super(BaseTaskActivity, self).title
+
+        if len(self.dossier_title) > self.CONTAINER_TITLE_MAX_LENGHT:
+            cropped_dossier_title = self.dossier_title[:self.CONTAINER_TITLE_MAX_LENGHT - 3] + "..."
```
lukasgraf: This should technically be `+ u"..."`, otherwise we're concatenating unicode and bytestrings, which leads to some implicit decoding as ASCII (this doesn't fail here because "..." doesn't contain non-ASCII characters, but it's not ideal).

opengever/activity/tests/test_digest.py
```diff
@@ -75,7 +75,7 @@ def test_mail_contains_date_activity_summaries(self, browser):
         browser.open_html(str(messages[0].get_payload()[0]))
 
         self.assertEquals('Oct 16, 2017', browser.css('table p').text[0])
-        self.assertEquals(['Daily Digest for B=C3=A4rfuss K=C3=A4thi'],
+        self.assertEquals(['= Daily Digest for B=C3=A4rfuss K=C3=A4thi ='],
```
lukasgraf: Where do the surrounding `=` come from? This doesn't seem quite right. That string decodes like this:

```python
>>> print '= Daily Digest for B=C3=A4rfuss K=C3=A4thi ='.decode('quoted-printable')
= Daily Digest for Bärfuss Käthi
```

opengever/activity/tests/test_mail.py
```diff
@@ -152,7 +152,7 @@ def test_task_title_is_linked_to_resolve_notification_view(self, browser):
         process_mail_queue()
 
         raw_mail = Mailing(self.portal).pop()
-        link = ('<p><a href=3D"http://nohost/plone/@@resolve_notification?notification=\n_id=3D1">'
+        link = ('<p><a href=3D"http://nohost/plone/@@resolve_notification?notificati=\non_id=3D1">'
```
lukasgraf: This doesn't seem right (neither does the way it was before though)...

PR #6273, opened 2020-02-19
Implement @copy-document-to-workspace endpoint

elioschmutz:
This PR impements the `@copy-document-to-workspace` endpoint according to the specs in https://github.com/4teamwork/opengever.core/issues/6197#issuecomment-586218068 .

## Checkliste

### DIFF
--- a/opengever/api/configure.zcml
+++ b/opengever/api/configure.zcml
@@ -584,4 +584,12 @@
       permission="zope2.View"
       />
 
+  <plone:service
+      method="POST"
+      name="@copy-document-to-workspace"
+      for="opengever.dossier.behaviors.dossier.IDossierMarker"
+      factory=".linked_workspaces.CopyDocumentToWorkspacePost"
+      permission="opengever.workspaceclient.UserWorkspaceClient"
+      />
+
 </configure>
--- a/opengever/api/configure.zcml
+++ b/opengever/api/configure.zcml
@@ -573,7 +573,7 @@
       name="@create-linked-workspace"
       for="opengever.dossier.behaviors.dossier.IDossierMarker"
       factory=".linked_workspaces.LinkedWorkspacesPost"
-      permission="opengever.workspaceclient.UserWorkspaceClient"
+      permission="opengever.workspaceclient.UseWorkspaceClient"
       />
 
   <plone:service
@@ -589,7 +589,7 @@
       name="@copy-document-to-workspace"
       for="opengever.dossier.behaviors.dossier.IDossierMarker"
       factory=".linked_workspaces.CopyDocumentToWorkspacePost"
-      permission="opengever.workspaceclient.UserWorkspaceClient"
+      permission="opengever.workspaceclient.UseWorkspaceClient"
       />
 
 </configure>
--- a/opengever/api/linked_workspaces.py
+++ b/opengever/api/linked_workspaces.py
@@ -1,7 +1,10 @@
 from functools import wraps
 from opengever.base.sentry import maybe_report_exception
+from opengever.document.document import IDocumentSchema
 from opengever.workspaceclient import is_workspace_client_feature_available
 from opengever.workspaceclient.interfaces import ILinkedWorkspaces
+from plone import api
+from plone.app.uuid.utils import uuidToObject
 from plone.protect.interfaces import IDisableCSRFProtection
 from plone.restapi.deserializer import json_body
 from plone.restapi.services import Service
@@ -98,6 +101,9 @@ class LinkedWorkspacesService(Service):
         if not is_workspace_client_feature_available():
             raise NotFound
 
+        if self.context.is_subdossier():
+            raise BadRequest
+
         return super(LinkedWorkspacesService, self).render()
 
 
@@ -106,9 +112,6 @@ class LinkedWorkspacesGet(LinkedWorkspacesService):
     """
     @request_error_handler
     def reply(self):
-        if self.context.is_subdossier():
-            raise BadRequest
-
         response = ILinkedWorkspaces(self.context).list()
 
         # The response id contains the url of the workspace-client request.
@@ -125,9 +128,6 @@ class LinkedWorkspacesPost(LinkedWorkspacesService):
 
     @request_error_handler
     def reply(self):
-        if self.context.is_subdossier():
-            raise BadRequest
-
         # Disable CSRF protection
         alsoProvides(self.request, IDisableCSRFProtection)
 
@@ -135,3 +135,44 @@ class LinkedWorkspacesPost(LinkedWorkspacesService):
         data = json_body(self.request)
 
         return ILinkedWorkspaces(self.context).create(**data)
+
+
+class CopyDocumentToWorkspacePost(LinkedWorkspacesService):
+    """API Endpoint to copy a document to a linked workspace.
+    """
+    @request_error_handler
+    def reply(self):
+        # Disable CSRF protection
+        alsoProvides(self.request, IDisableCSRFProtection)
+
+        workspace_uid, document = self.validate_data(json_body(self.request))
+
+        return ILinkedWorkspaces(self.context).copy_document_to_workspace(
+            document, workspace_uid)
+
+    def validate_data(self, data):
+        workspace_uid = data.get('workspace_uid')
+        if not workspace_uid:
+            raise BadRequest("Property 'workspace_uid' is required")
+
+        document_uid = data.get('document_uid')
+        if not document_uid:
+            raise BadRequest("Property 'document_uid' is required")
+
+        document = uuidToObject(document_uid)
+        if not document or not IDocumentSchema.providedBy(document):
+            raise BadRequest("The document does not exist")
+
+        if not self.obj_contained_in(document, self.context):
+            raise BadRequest(
+                "Only documents within the current main dossier are allowed")
+
+        return workspace_uid, document
+
+    def obj_contained_in(self, obj, container):
+        catalog = api.portal.get_tool('portal_catalog')
+        results = catalog.searchResults(
+            path={'query': '/'.join(container.getPhysicalPath())},
+            UID=obj.UID())
+
+        return len(results)
--- a/opengever/api/tests/test_linked_workspaces.py
+++ b/opengever/api/tests/test_linked_workspaces.py
@@ -1,4 +1,3 @@
-from contextlib import contextmanager
 from ftw.builder import Builder
 from ftw.builder import create
 from ftw.testbrowser import browsing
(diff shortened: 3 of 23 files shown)


opengever/api/configure.zcml
```diff
@@ -584,4 +584,12 @@
       permission="zope2.View"
       />
 
+  <plone:service
+      method="POST"
+      name="@copy-document-to-workspace"
+      for="opengever.dossier.behaviors.dossier.IDossierMarker"
+      factory=".linked_workspaces.CopyDocumentToWorkspacePost"
+      permission="opengever.workspaceclient.UserWorkspaceClient"
```
njohner: Shit there is a typo in the permission ID 😓 . It should read `UseWorkspaceClient` not `UserWorkspaceClient`.
Because that is in the zcml only, can we simply correct it there?

opengever/api/linked_workspaces.py
```diff
@@ -135,3 +139,37 @@ def reply(self):
...
+
+        document_uid = data.get('document_uid')
+        if not document_uid:
+            raise BadRequest("Property 'document_uid' is required")
+
+        document = uuidToObject(document_uid)
+        if not document or not IDocumentSchema.providedBy(document):
+            raise BadRequest("The document does not exist")
+
+        if self.context.get_main_dossier().absolute_url() not in document.absolute_url():
```
njohner: I'm not sure I'm a big fan of testing containment with URLs

njohner: Why not a catalog search for the UID in the dossier?

opengever/api/tests/test_linked_workspaces.py
```diff
@@ -180,3 +180,153 @@ def assertStatusCode(test_with, raised_error):
...
+                with self.assertRaises(Unauthorized) as cm:
+                    browser.open(
+                        self.dossier.absolute_url() + '/@copy-document-to-workspace',
+                        data=json.dumps(payload),
+                        method='POST',
+                        headers={'Accept': 'application/json',
+                                 'Content-Type': 'application/json'},
+                    )
+
+        self.assertEqual("You are not authorized to access this resource.", str(cm.exception))
```
njohner: I guess here you also don't have access to the dossier, and that might actually be raising. On the other hand we never set permissions on documents...
Anyway that test seems a bit weak, as it is now you effectively remove all the permissions of the user not? I'd rather that you block the local roles on the dossier, assert that it raises, then give the user the minimal local roles necessary to make it work and show that it works in that case.

opengever/workspaceclient/client.py
```diff
@@ -81,3 +83,61 @@ def lookup_url_by_uid(self, uid):
...
+                   **additional_metadata):
+        """Creates a new file within the folder of the given url or path and
+        extends the created file with some additional metadata.
+
+        :param url_or_path: Location where to create the new document
+        :param file: Readable IO which holds the content of the file
+        :param size: The size of the file
+        :param content_type: The content type of the document
+        :param filename: The filename of the document
+        :parapm additional_metadata: Additional metadatadata for the new object
```
njohner: parapm -> param

opengever/workspaceclient/client.py
```diff
@@ -81,3 +83,61 @@ def lookup_url_by_uid(self, uid):
...
+        document_url = created_document.headers['Location']
+
+        # It's not possible to directly update the metadata of an object through
+        # a TUS-request. We perform another patch-request to update the newly
+        # created document with some additional metadata.
+        return self.patch(document_url, json=additional_metadata,
+                          headers={'Prefer': 'return=representation'})
+
+    def _tus_metadata_string(self, metadata):
+        """We can pass metadata in the request header of a TUS-reqeust.
```
njohner: TUS-reqeust -> TUS-request

opengever/workspaceclient/tests/test_linked_workspaces.py
```diff
@@ -10,6 +11,31 @@
 import transaction
 
 
+@contextmanager
+def auto_commit_after_request(client):
+    """This contextmanager injects a sessino hook for the current client
```
njohner: sessino -> session

opengever/workspaceclient/tests/test_linked_workspaces.py
```diff
@@ -10,6 +11,31 @@
...
+    session to commit the transaction after each request.
+
+    This is required in some functional non-browser tests where we perform
+    multiple subrequests in one function call.
+
+    This is not required for browser-tests or production environments because
+    the transaction will be committed automatically as soon as a request have
```
njohner: -> as soon as a request is finished

opengever/api/tests/test_linked_workspaces.py
```diff
@@ -16,14 +15,14 @@
 
 
 def fix_publisher_test_bug(browser, obj):
-    # We need to make a GET request before using the `@create-linked-workspace`
-    # endpoint. This is a workaround to fix a weird issue where we get
+    # We need to make a GET request before using a post-endpoint endpoint.
```
njohner: post-endpoint endpoint?

opengever/api/tests/test_linked_workspaces.py
```diff
@@ -330,3 +329,39 @@ def test_copy_document_without_file_to_workspace(self, browser):
...
+                )
+
+            workspace_document = children['added'].pop()
+
+            self.assertEqual(workspace_document.absolute_url(), browser.json.get('@id'))
+            self.assertEqual(workspace_document.title, document.title)
+
+            self.assertItemsEqual(
+                manager._serialized_document_schema_fields(document),
+                manager._serialized_document_schema_fields(workspace_document))
```
njohner: Check the document file content as well

opengever/core/profiles/default/actions.xml
```diff
@@ -525,6 +525,15 @@
       <property name="visible">True</property>
     </object>
 
+    <object name="copy_documents_to_workspace" meta_type="CMF Action" i18n:domain="opengever.core">
```
njohner: I'm not too sure what our strategy is here, I thought we do not want anything having to do with workspaces in the classic UI. This would mean having a new category `folder_actions` that is only used in the frontend. Here you say in the description `Gever-UI action` but it will actually appear in the classic UI too.

opengever/api/linked_workspaces.py
```diff
@@ -110,10 +110,8 @@ class LinkedWorkspacesGet(LinkedWorkspacesService):
     """
     @request_error_handler
     def reply(self):
-        if self.context.is_subdossier():
-            raise BadRequest
-
-        response = ILinkedWorkspaces(self.context).list()
+        context = self.context.get_main_dossier()
```
njohner: I think we had decided to only list linked workspaces on main dossiers (see https://github.com/4teamwork/opengever.core/pull/6254#pullrequestreview-358361693). See with @jone if you feel this should also be on subdossiers

elioschmutz: Thats true. The intention of this adjustment is to have the linked workspaces of the main dossier available on subdossiers, not to link or list them on subdossiers.

Currently, we can only copy documents to a linked workspace if we are in the main dossier. But we also want to copy documents if we are within a subdossier. Otherwise this seems to be very user unfriendly.

To make this possible, we need a list of linked workspaces on the main dossier if we call the view from a subdossier.

With this adjustments, I implement point 2 of your comment in  https://github.com/4teamwork/opengever.core/pull/6254#pullrequestreview-358361693
> Forwarding documents to a workspace could also be allowed on subdossiers, or not, depending on what is easier to implement
I understand the decisions like: we don't want to let 

What do you think? @njohner @jone 

I think we should do this.

opengever/core/profiles/default/actions.xml
elioschmutz: I don't know the current state of discussion. Of course, we can just use a new category and use this cateogry in the frontend.

But we never use the actions directly in the old ui nor in the new ui. We always whitelist actions for a view or type. So the new action will not appear in the old frontend because it is not whitelistet in the tabbedviews. (Yes, it is listed in the `folder_contents` view, but this view is not used by the user and anyway, there are listed many not related actions for a specific type).

If we change to a new category, we would have to copy all existing actions to the new category to make the new UI works again.

I think, using the same action category for all actions is enough at the moment.

But we can discuss this topic in the team that we have a clear strategy.

What do you think? @njohner

opengever/api/configure.zcml
elioschmutz: np, I'll update it in a new commit 👍

opengever/api/tests/test_linked_workspaces.py
elioschmutz: Sure. I think I can just remove this test because I test some internals of plone. I do not change or validate any permission in my code, so I don't need to test it. Plone will raise an unauthorized error automatically if the user does not have permission to access the document.

Do you agree if I remove the test?

opengever/workspaceclient/linked_workspaces.py
```diff
@@ -64,3 +69,47 @@ def create(self, **data):
...
+            for fieldname in schema:
+                fieldnames.append(fieldname)
+        return fieldnames
+
+    def _serialized_document_schema_fields(self, document):
+        """Serializes all document schema fields.
+        """
+        serializer = getMultiAdapter((document, self.context.REQUEST), ISerializeToJson)
+        document_repr = serializer()
+        whitelisted = self._form_fields(document)
```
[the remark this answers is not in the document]
elioschmutz: There are several attributes we don't want in our patch-request: i.e. the `type` or the `id`.

We only want the form-fields which contains the important metadata.

opengever/api/tests/test_linked_workspaces.py
njohner: 👍

opengever/core/profiles/default/actions.xml
njohner: Right, ok I got messed up. The issue is actually the other way round, when some actions always appear in the `folder_actions` but are only whitelisted in certain places, then the new UI has no way of knowing that. But here it is fine, they are only in the `folder_actions` when necessary. so 👍

opengever/api/linked_workspaces.py
elioschmutz: @all4manu FYI

njohner: We discussed this again and bottom line is we want everything only on the main dossier for now. Once we have all the actions and we see how everything works in the new UI we can always discuss again whether we want some things also on the subdossiers.

njohner: So I'd drop that commit for now...

PR #6283, opened 2020-02-24
Add smoke tests for rewrite rules and VHost configs.

lukasgraf:
Implements some basic smoke test for our rewrite rules and VHost configurations.

### Usage

Run

```bash
bin/zopepy rewrite_rule_smoketests/smoketests.py
```

to execute the smoke tests. Some tests will require credentials, and will therefore be skipped.

skipped_tests

In order to provide those credentials (one pair per cluster), run

```bash
bin/zopepy rewrite_rule_smoketests/smoketests.py --prompt-credentials
```

The credentials (username and password) will then be used to log into the cluster's corresponding portal, get a session, and store those session cookies to `~/.opengever/rewrite_rule_smoketests/`.

The password itself will never be stored.

Addresses #6259 

## Checkliste

### DIFF
--- a/rewrite_rule_smoketests/auth.py
+++ b/rewrite_rule_smoketests/auth.py
@@ -1,0 +1,96 @@
+from config import SESSION_COOKIES_FILE
+from os.path import isfile
+import json
+
+
+API_POST_HEADERS = {
+    'Accept': 'application/json',
+    'Content-Type': 'application/json',
+}
+
+# Cookie params that the CookieJar doesn't allow us to set
+UNSUPPORTED_COOKIE_PARAMS = [
+    '_rest',
+    'path_specified',
+    'domain_specified',
+    'port_specified',
+    'domain_initial_dot',
+]
+
+
+def login_to_portal(browser, cluster, username, password):
+    """Performs a login to the portal, using given username and password.
+
+    Returns the username of the logged in user, and the cookies received in
+    the response. These cookies contain the portal session, and can be
+    stored and loaded to avoid logging in all the time.
+    """
+    if cluster.new_portal:
+        payload = json.dumps({
+            'username': username,
+            'password': password,
+            'remember_me': False,
+            'invitation': ""})
+
+        login_api_url = '/'.join((cluster.portal_url, 'api', 'login'))
+        browser.open(login_api_url,
+                     method='POST',
+                     headers=API_POST_HEADERS, data=payload)
+
+        expected_response = {
+            'username': username,
+            'state': 'logged_in',
+            'invitation_callback': ''}
+
+        if not expected_response == browser.json:
+            print "Login failed: %r" % browser.contents
+
+    else:
+        login_url = '/'.join((cluster.portal_url, 'login'))
+
+        browser.open(login_url)
+        browser.fill({'Benutzername': username, 'Passwort': password}).submit()
+        # TODO: Check for login success, and maybe raise a LoginFailed
+        # exception here so we can distinguish this case.
+
+    # Extract and clean up cookies from response
+    cookies = dict(browser.cookies)
+    for cookie_name, cookie_data in cookies.items():
+        for param_to_remove in UNSUPPORTED_COOKIE_PARAMS:
+            cookies[cookie_name].pop(param_to_remove, None)
+
+    return username, cookies
+
+
+def load_session_cookies():
+    """Load stored session cookies from file.
+    """
+    if not isfile(SESSION_COOKIES_FILE):
+        return {}
+
+    with open(SESSION_COOKIES_FILE, 'r') as cookies_file:
+        return json.load(cookies_file)
+
+
+def store_session_cookies(session_cookies):
+    """Store session cookies to file, overwriting existing ones.
+    """
+    with open(SESSION_COOKIES_FILE, 'w') as cookies_file:
+        json.dump(session_cookies, cookies_file,
+                  sort_keys=True, indent=4, separators=(',', ': '))
+
+
+def login_and_store(browser, cluster, username, password):
+    """Log in, and store session cookies to file (keyed by cluster URL).
+
+    If cookies already exist for the given cluster, they get updated.
+    Cookies for other clusters are left untouched.
+    """
+    logged_in_user, cookies = login_to_portal(
+        browser, cluster, username, password)
+
+    session_cookies = load_session_cookies()
+    cookie_set = {cluster.url: {'username': username,
+                                'cookies': cookies}}
+    session_cookies.update(cookie_set)
+    store_session_cookies(session_cookies)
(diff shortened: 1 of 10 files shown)


rewrite_rule_smoketests/registry.py
```diff
@@ -0,0 +1,23 @@
+"""Registry that keeps track of tests that get registered via decorators.
+"""
+
+tests_by_type = {
+    'cluster': [],
+    'admin_unit': [],
+    'logged_in': [],
```
njohner: If I understand well, `logged_in` is a supplemental information for a test that is also in one of the two other categories, but is not a category by itself (tests in there do not get run if they are not in one of the two other categories)

njohner: It's clear from the code, but maybe a comment somewhere or in a sentence in the README would help.

rewrite_rule_smoketests/smoketests.py
```diff
@@ -0,0 +1,198 @@
...
+
+@on_cluster
+@on_admin_unit
+def test_canonical_https_redirect(browser, entity):
+    """Tests for presence of canonical HTTPS redirect.
+
+    All URLs that start with a http:// scheme should be redirected to their
+    https equivalent.
+    """
+    expect_trailing_slash = False
```
njohner: Does it actually matter that we have the trailing slash in some cases? 
If we're not sure whether we need it or not, then this is simply the conservative assertion, which is fine for now...

rewrite_rule_smoketests/smoketests.py
```diff
@@ -0,0 +1,198 @@
...
+        expected += '/'
+
+    assert_equal(301, browser.status_code)
+    assert_equal(expected, browser.headers['Location'])
+
+
+@on_admin_unit
+def test_teamraum_redirect(browser, admin_unit):
+    if not admin_unit.is_dedicated_teamraum:
+        return
```
njohner: Instead of skipping for non `teamraum` admin_units, we could make sure that we do not have the redirect in those cases

rewrite_rule_smoketests/smoketests.py
```diff
@@ -0,0 +1,198 @@
...
+    """
+    browser.allow_redirects = False
+    browser.open(admin_unit.front_page_url)
+
+    cluster = admin_unit.cluster
+
+    if cluster.gever_ui_is_default:
+        # When the new UI is the default, we won't see a CAS redirect
+        # performed by Plone for Anonymous requests. Instead, the frontend
+        # app's index.html be served, and the frontend will trigger the CAS
```
njohner: be served -> is served

rewrite_rule_smoketests/auth.py
```diff
@@ -0,0 +1,96 @@
...
+        browser.open(login_api_url,
+                     method='POST',
+                     headers=API_POST_HEADERS, data=payload)
+
+        expected_response = {
+            'username': username,
+            'state': 'logged_in',
+            'invitation_callback': ''}
+
+        if not expected_response == browser.json:
```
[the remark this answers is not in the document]
lukasgraf: Yep. Though *it is* the response from the portal's JSON API. So it should at least be reasonably stable (at least not as brittle as asserting HTML).

rewrite_rule_smoketests/auth.py
```diff
@@ -0,0 +1,96 @@
...
+            print "Login failed: %r" % browser.contents
+
+    else:
+        login_url = '/'.join((cluster.portal_url, 'login'))
+
+        browser.open(login_url)
+        browser.fill({'Benutzername': username, 'Passwort': password}).submit()
```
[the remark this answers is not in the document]
lukasgraf: It most certainly is :) This was also one of those "lets cross this bridge when we come to it" cases for me. Should be a failure that's trivial to spot, and relatively simple to fix.

rewrite_rule_smoketests/auth.py
```diff
@@ -0,0 +1,96 @@
...
+        # TODO: Check for login success, and maybe raise a LoginFailed
+        # exception here so we can distinguish this case.
+
+    # Extract and clean up cookies from response
+    cookies = dict(browser.cookies)
+    for cookie_name, cookie_data in cookies.items():
+        for param_to_remove in UNSUPPORTED_COOKIE_PARAMS:
+            cookies[cookie_name].pop(param_to_remove, None)
```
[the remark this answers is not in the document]
lukasgraf: Hmm - unless I'm missing something here, that wouldn't work. That would remove the param from the `cookie_data`, which is a copy of the dict's value, but still leave it in the original `cookies` which we want to clean up.

rewrite_rule_smoketests/config.py
```diff
@@ -0,0 +1,70 @@
...
+        # We want a homogenous URL style to start building other URLs from
+        # in order to get predictable results.
+        assert not cluster.url.endswith('/'), \
+            "Cluster URLs must not end with trailing slash."
+
+
+def get_config():
+    check_config(CLUSTERS_TO_TEST)
+
+    # Link admin units to their containing cluster
```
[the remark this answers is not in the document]
lukasgraf: Great idea - it fit's much better there!

rewrite_rule_smoketests/registry.py
lukasgraf: Yep, it's exactly as you described. I moved these pieces around in the last minute, and it became a bit more entangled than I intended. I added a comment that should make it pretty clear.

rewrite_rule_smoketests/smoketests.py
lukasgraf: It most likely doesn't make a difference. But as you say, I erred on the conservative side, if just to keep track of these little oddities. I don't want to start smoothing over things by normalizing them quite yet 😉

rewrite_rule_smoketests/smoketests.py
lukasgraf: It's not quite that easy unfortunately. Depending on several factors (whether the new UI is the default, and whether the user is logged in or not), we might still get a (legitimate) 302 redirect from the admin unit URL.

So, to easily have the test not fail in these cases, and not make it too complex, one could basically assert on the exact inverse: Expect that there is NO redirect to the `/workspaces` URL. However, tests like that (same for example with asserting that something is NOT in a list) tend to be quite dangerous - because they could still pass, even if the behavior is broken. For example if the teamraum redirect target changed, but the teamraum redirect still gets triggered erroneously, that test would be green.

So it's usually better to assert on the behavior you expect to see, instead of asserting on the absence of fauly behavior. But in this case this would pretty much mean speccing out all the other desired behaviors. Redirect to portal, delivery of frontend `index.xml`, serving of the GEVER front page. And that would replicate assertions in this test that are already covered elsewhere.

rewrite_rule_smoketests/auth.py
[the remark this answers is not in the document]
njohner: Guess it depends whether `cookie_data` is mutable...

Issue #6282, opened 2020-02-24
Unauthorized if using the `WorkspaceClient`


elioschmutz:
_Harvest: `[3.2.4 / 2] Projekt - teamraum 2019`_

The `WorkspaceClient` handles authentication tokens for each user using `ftw.tokenauth`. If a token is expired, it will automatically renew the token which works fine: https://github.com/4teamwork/opengever.core/blob/master/opengever/workspaceclient/session.py#L16

`ftw.tokenauth` is cleaning up all expired tokens after issuing a new token. If there is no token in the token-storage of `ftw.tokenauth` it will not raise a 401 with the `token expired` messagee. See: See: https://github.com/4teamwork/ftw.tokenauth/issues/14

That means, we're no longer able to renew an expired token as soon as a new token (for the current user in another thread or for a new user) is issued. We end up in an anonymous api call which will raise a default 401.

We need to fix this either in `ftw.tokenauth` or in GEVER.

elioschmutz:
Entscheid: Bei einem 401 grundsätzlich immer ein neues Token anfordern.

jone:
✅ ist für mich ok; ich habe den Fall jeztt nicht reproduziert da schwierig.

PR #6285, opened 2020-02-26
Always obtain a new access token on a 401 http error in the workspace client session

elioschmutz:
Issuer: #6282 

This PR fixes an issue where the user got a 401 if the users access token is expired.

Please see the details in the issue #6282

### DIFF
--- a/opengever/workspaceclient/session.py
+++ b/opengever/workspaceclient/session.py
@@ -40,27 +40,19 @@ class FtwTokenAuthSession(requests.Session):
 
         response = super(FtwTokenAuthSession, self).request(method, url,
                                                             *args, **kwargs)
-        if self.token_has_expired(response):
-            # We got an 'Access token expired' response => refresh token
+        if response.status_code == 401:
+            # We try to refresh the access token
             self.obtain_token()
+
             # Re-dispatch the request that previously failed
+            # (only once, using a single, non-recursive request, in order to
+            # avoid an infinite loop)
             response = super(FtwTokenAuthSession, self).request(method, url,
                                                                 *args, **kwargs)
 
         response.raise_for_status()
 
         return response
-
-    def token_has_expired(self, response):
-        status = response.status_code
-        content_type = response.headers.get('Content-Type')
-
-        if status == 401 and content_type == 'application/json':
-            body = response.json()
-            if body.get('error_description') == 'Access token expired':
-                return True
-
-        return False
 
     def obtain_token(self):
         """Acquires a fresh authorization token from the workspace and sets it
--- a/opengever/workspaceclient/tests/test_session.py
+++ b/opengever/workspaceclient/tests/test_session.py
@@ -8,6 +8,7 @@ from opengever.workspaceclient.tests import FunctionalWorkspaceClientTestCase
 from plone import api
 from plone.app.testing import TEST_USER_ID
 from requests import HTTPError
+from zExceptions import Unauthorized
 import requests_mock
 import transaction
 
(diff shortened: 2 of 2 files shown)


opengever/workspaceclient/session.py
```diff
@@ -40,9 +40,10 @@ def request(self, method, path_or_url, *args, **kwargs):
 
         response = super(FtwTokenAuthSession, self).request(method, url,
                                                             *args, **kwargs)
-        if self.token_has_expired(response):
-            # We got an 'Access token expired' response => refresh token
+        if response.status_code == 401:
```
lukasgraf: I think with this approch, this does need a retry counter that is able to break cycles. 401 responses are also possible for users that are authenticated, but unauthorized.

I see that there is a test for token renewal 401 loops, but I don't quite understand how the loop is prevented here. Is it because below, we redispatch to `super(FtwTokenAuthSession, self).request` instead of `self.request`?

elioschmutz: @lukasgraf exactly... we don't need a retry counter.  We just have to obtain a new token once if we get a 401. If the user still have no access, we can raise. The user won't get permission if we retry it more than once.

The loop is prevented because it is not a recursive function. We only obtain the token once if we get a 401. If it fails, it will raise.

lukasgraf: Ah, I see - I somehow assumed the superclass of `FtwTokenAuthSession` was some similar `BaseSession` implementation whose `request` method then would have been recursive (didn't see the base class in the diff). But it's just a plain `requests.Session`, so it's a regular non-recursive request. 👍 

It's not very obvious though from the code that this is the mechanism that prevents cycles 😉 Could you add a quick comment that explains this, so that someone in the future doesn't have the glorious idea to make this recursive? 😉 Something like
```
# Re-dispatch the request that previously failed
# (Once, using a single, non-recursive request, in order to avoid an infinite loop)
```

elioschmutz: @lukasgraf 👍 updated

opengever/workspaceclient/tests/test_session.py
```diff
@@ -71,6 +72,62 @@ def extract_jwt_token(session):
...
+            del storage._access_tokens[token]
+        transaction.commit()
+
+        # Making a request with a not exisitng token will auto-renew the jwt-token
+        response = session.request.get('/').json()
+        self.assertNotEqual(jwt_token, extract_jwt_token(session),
+                            "JWT should have been renewed")
+        self.assertEqual(self.portal.absolute_url(), response.get('@id'))
+
+    def test_http_error_401_will_not_end_in_an_infinite_loop_of_token_renewal(self):
```
lukasgraf: I still do not quite understand this test. It mocks the `@@oauth2-token` to return a `401`. But this view should not ever return a `401` because
- [It's `Public`](https://github.com/4teamwork/ftw.tokenauth/blob/master/ftw/tokenauth/oauth2/browser/configure.zcml#L7-L11), and therefore accessible to `Anonymous`
- Requests to that view should *not include* any tokens anyway. They need to be made with a different session (or none at all), to avoid sending invalid / expired tokens when trying to get a new one. This is what this sentence in the `ftw.tokenauth` README refers to:
  > Care needs to be taken to not include an expired token (or any Authorization header for that matter) with the requests to the token endpoint when obtaining a new token.
And `FtwTokenauthSession. obtain_token ()` does this correctly, from what I can tell.

The problem with "default 401 vs 401 with invalid token JSON message" should instead be happening with just regular requests, no? Requests to the `@linked-workspaces` endpoint for example.

So, I'm not sure that this test ever [failed / would fail](https://dzone.com/articles/make-your-test-fail-first). You do also mock a `401` response onto a GET for `self.portal.absolute_url()` though. But I'd appreciate it if you could make sure that the test really does fail if you build an infinite loop. I mean, if the test actually works, then it would actually create an infinite loop (on Jenkins for example) in the case of incorrect core, and then eventually fail with `RuntimeError: maximum recursion depth exceeded`.

elioschmutz: @lukasgraf you're right! I missed the `real_http=True` param of the Mocker. Now, it's working as expected. Here the loop, if I create a recursion:

Bildschirmfoto 2020-02-26 um 11 53 01

PR #6292, opened 2020-03-02
Smoketests: Test that POST requests never get answered with index.html.

lukasgraf:
We had a regression in our rewrite rules where POST requests (from Bumblebee) got answered with the `index.html` for the new frontend.

This adds a failing test to ensure this doesn't happen.

The current plan to fix this is to change the RewriteRules in a way that the `index.html` _never_ gets served in response to any request with a method other than `GET`.

_(No changelog entry, since it's a testing only change)._

## Checkliste

- [ ] Changelog-Eintrag vorhanden/nötig?

### DIFF
--- a/rewrite_rule_smoketests/smoketests.py
+++ b/rewrite_rule_smoketests/smoketests.py
@@ -183,6 +183,22 @@ def test_gever_front_page_logged_in_old_ui(browser, admin_unit):
     assert_equal(admin_unit.url, logo.attrib['href'])
 
 
+@on_admin_unit
+def test_anonymous_post_requests_always_hit_gever(browser, admin_unit):
+    """This is a regression test for a case, where simple POST requests got
+    answered with the new frontend's index.html.
+    """
+    target_url = '/'.join((admin_unit.front_page_url, 'doesnt-exist'))
+
+    browser.raise_http_errors = False
+    browser.open(target_url, method='POST')
+
+    assert_equal(404, browser.status_code)
+
+    # Assert we hit Plone by checking for the presence of the logo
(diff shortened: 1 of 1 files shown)


rewrite_rule_smoketests/smoketests.py
```diff
@@ -183,6 +183,28 @@ def test_gever_front_page_logged_in_old_ui(browser, admin_unit):
...
+@on_admin_unit
+def test_anonymous_post_requests_always_hit_gever(browser, admin_unit):
+    """This is a regression test for a case, where simple POST requests got
+    answered with the new frontend's index.html.
+    """
+    target_url = '/'.join((admin_unit.front_page_url, 'doesnt-exist'))
+
+    driver = browser.get_driver()
+
+    cookie_data = {'name': 'geverui', 'value': '0'}
```
deiferni: I'm not sure why we set a cookie here, shouldn't `POST` hit plone in any case?

rewrite_rule_smoketests/smoketests.py
```diff
@@ -183,6 +183,28 @@ def test_gever_front_page_logged_in_old_ui(browser, admin_unit):
...
+    cookie = create_cookie(**cookie_data)
+    driver.requests_session.cookies.set_cookie(cookie)
+
+    browser.raise_http_errors = False
+    browser.open(target_url, method='GET')
```
deiferni: this is a `GET` request?

lukasgraf:
Absolutely - I had a case of brain fog and check in the wrong test. Will update.

lukasgraf:
@deiferni updated.

PR #6294, opened 2020-03-02
Register demo.teamraum.ch to rewrite rule tests

elioschmutz:
This PR registers demo.teamraum.ch as a new Cluster for rewrite rule tests.

### DIFF
--- a/rewrite_rule_smoketests/config.py
+++ b/rewrite_rule_smoketests/config.py
@@ -48,4 +48,13 @@ CLUSTERS_TO_TEST = [
         gever_ui_is_default=False,
     ),
 
+    Cluster(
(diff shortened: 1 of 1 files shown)


elioschmutz:
@njohner I thought that we currently execute this tests locally...

njohner:
They are checking the rewrite rules so they need to be executed on the corresponding server. From what I remember there is currently no option to test certain clusters on certain servers. But I might be wrong though. Have a look at the code or maybe @lukasgraf can give input here.

elioschmutz:
Yeah, but for testing the rules, it is just requesting the defined cluster url. If they are public, you can test it from your local machine. I think you mean testing rules of a gever inside a private network...

PR #6298, opened 2020-03-04
Add API endpoint to create document from template.

deiferni:
This PR adds an api endpoint `@document-from-template` to create documents based on a document template. It also adds a vocabulary `opengever.dossier.DocumentTemplatesVocabulary` that contains the available templates.

The endpoint is very simple and mirrors the functionality currently available in the old UI. It just allows the user to select a template and a title for the document which will be created as a copy of the template.

I have made the following disputable decisions while implementing:
- the vocabulary namespace is `opengever.dossier` as the dossier might in the future restrict the templates that are available. currently it does not though, so we migth also choose `opengever.templatefolder...` as namespace for the vocabulary. If you think so please speak up.
- currently the helper `get_validation_errors` does not seem handle vocabulary-based schema fields at all. Thus i have not implemented a schema/interface with two fields but instead just hardcoded the two fields into the service endpoint. I can invest some more time to fix `get_validation_errors` and implement a "proper" schema if desired.

For https://github.com/4teamwork/gever-ui/issues/944.

## Checkliste

- [x] Gibt es neue Funktionalität mit einem `Dokument`? ~~Funktioniert das auch mit einem `Mail`?~~

### DIFF
--- a/docs/public/dev-manual/api/templatefolder.rst
+++ b/docs/public/dev-manual/api/templatefolder.rst
@@ -1,0 +1,33 @@
+.. _templatefolder:
+
+Dokumente ab Vorlage erstellen
+==============================
+
+In einem Dosser kann über den Endpoint ``@document-from-template`` ein neues
+Dokument ab Vorlage erstellt werden.
+
+Der Endpoint steht auf einem Dossier zur Verfügung. Der Endpoint ist mit der
+Berechtigung ``opengever.document: Add document`` geschützt, kann also verwendet
+werden wenn der Benutzer Dokumente hinzufügen kann.
+
+Der Endpoint erwartet zwei Parameter:
+
+- ``template`` das zu verwendende template aus dem Vokabular ``opengever.dossier.DocumentTemplatesVocabulary``
+- ``title`` Titel des zu erstellenden Dokumentes
+
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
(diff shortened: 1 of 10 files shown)


opengever/api/templatefolder.py
```diff
@@ -0,0 +1,44 @@
...
+from plone.restapi.services import Service
+from zExceptions import BadRequest
+from zope.component import getUtility
+from zope.component import queryMultiAdapter
+from zope.interface import alsoProvides
+from zope.schema.interfaces import IVocabularyFactory
+
+
+class DocumentFromTemplatePost(Service):
+    """API Endpoint to crate a document in a dossier from a template.
```
lukasgraf: crate -> create

docs/public/dev-manual/api/templatefolder.rst
```diff
@@ -0,0 +1,33 @@
...
+       POST /ordnungssystem/fuehrung/dossier-23/@document-from-template HTTP/1.1
+       Accept: application/json
+
+       {
+        "template": {"token": "1234567890"},
+        "title": "Document title"
+       }
+
+
+Als Reponse wir die JSON-Räpresentation des neu erstellen Dokuments geliefert,
```
lukasgraf: wir -> wird, Räpresentation -> Repräsentation

deiferni: Wunderschön bärndütsch, Räpresentation 😂

PR #6299, opened 2020-03-05
Smoketest: Switch dev.onegovgever.ch to classic UI default.

lukasgraf:
Since [`dev.onegovgever.ch` has been switched to serve the classic UI by default](https://git.4teamwork.ch/misc/puppet/commit/88923ee96c0bb3daf1e7907e234ecb96a51a4733) (instead of the new UI like before), the smoketests, which are currently failing because of that, need to be updated accordingly.


lukasgraf:
Thanks 😉

PR #6318, opened 2020-03-12
Extend Admin und OrgUnit Schema with `hidden` flag

njohner:
- We extend admin_unit and org_unit with a `hidden` flag
- We only show non-hidden org_units in the orgunit selector
- We do not allow to choose a responsible from hidden orgunits when adding a task or a forwarding TTW
- We allow to choose a responsible from hidden orgunits when adding a task or a forwarding over the REST-API

### Implementation
- We had two `OGDSService` classes doing essentially the same thing.
  - We make sure that one of them does everything
  - We delete the other
- We add the new `hidden` column in the models
- We make sure that the flag is taken into account in all methods in the `OGDSService`
- For allowing hidden orgunits over the REST-API but not TTW, we make use of the fact that in `AllUsersInboxesAndTeamsSource`, `getTerm` also considers `hidden` and `disabled` orgunits, whereas the `search_query` does not, i.e. the widget will only show users from visible and enabled orgunits, but useres from hidden and disabled orgunits are still valid.

For https://github.com/4teamwork/opengever.core/issues/6128, https://github.com/4teamwork/opengever.core/issues/6129 and https://github.com/4teamwork/opengever.core/issues/6130

### Open points:
- [ ] `OGDSService.has_multiple_org_units`  also counts disabled (and hidden orgunits), not sure why

## Checkliste


opengever/ogds/models/service.py
```diff
@@ -108,10 +117,12 @@ def all_groups(self, active_only=True):
             query = query.filter_by(active=True)
         return query.all()
 
-    def _query_org_units(self, enabled_only=True):
+    def _query_org_units(self, enabled_only=True, visible_only=True):
         query = OrgUnit.query
         if enabled_only:
             query = query.filter_by(enabled=enabled_only)
+        if visible_only:
+            query = query.filter_by(hidden=False)
```
elioschmutz: Bei `enabled_only` verwendest du das Argument als Wert, beim `visible_only` direkt `False`

Kannst du das konsistent machen?

opengever/ogds/models/tests/test_service.py
```diff
@@ -100,17 +123,69 @@ def test_falsy_multiple_org_units(self):
...
+        self.org_unit_b.enabled = False
+        self.org_unit_c.enabled = False
+        self.org_unit_d.enabled = False
+        self.assertTrue(self.service.has_multiple_org_units())
+
+    def test_has_multiple_org_units_counts_hidden_org_units(self):
+        self.org_unit_b.hidden = True
+        self.org_unit_c.hidden = True
+        self.org_unit_d.hidden = True
+        self.assertTrue(self.service.has_multiple_org_units())
```
elioschmutz: Duplication of the previous test

opengever/ogds/base/sources.py
```diff
@@ -155,6 +156,8 @@ def search_query(self):
               be queried.
             - only_active_orgunits: Filter out actors that don't have an org
               unit at all, or an inactive one (true by default).
+            - only_visible_orgunits: Filter out actors that don't have an org
+              unit at all, or an hidden one (true by default).
```
elioschmutz:
```suggestion
              unit at all, or a hidden one (true by default).
```

opengever/inbox/tests/test_forwarding.py
```diff
@@ -100,25 +100,24 @@ def test_users_are_available_as_responsible(self, browser):
 
     @browsing
     def test_hidden_orgunits_are_not_available_as_responsible_client(self, browser):
+        """responsibles from a hidden orgunit are valid, but the widget does
```
elioschmutz:
```suggestion
        """Responsibles from a hidden orgunit are valid, but the widget does
```

opengever/ogds/base/tests/test_vocabularies.py
```diff
@@ -140,7 +140,7 @@ def test_contains_only_enabled_org_units(self):
              (u'arch', u'Staatsarchiv')],
             voca_factory(self.portal))
 
-    def test_contains_only_visible_org_units(self):
+    def test_contains_hidden_org_units(self):
```
elioschmutz:
```suggestion
    def test_contains_all_org_units(self):
```

opengever/task/tests/test_task.py
```diff
@@ -289,24 +289,25 @@ def test_cannot_set_responsible_client_for_disabled_orgunit(self, browser):
             browser.css('#formfield-form-widgets-responsible_client .error').text)
 
     @browsing
-    def test_cannot_set_responsible_client_for_disabled_orgunit2(self, browser):
-        self.get_org_unit().hidden = True
+    def test_cannot_set_responsible_client_for_hidden_orgunit(self, browser):
+        """responsibles from a hidden orgunit are valid, but the widget does
```
elioschmutz:
```suggestion
        """Responsibles from a hidden orgunit are valid, but the widget does
```

opengever/core/upgrades/20200320101520_add_hidden_flag_to_admin_and_org_units/upgrade.py
```diff
@@ -0,0 +1,34 @@
...
+            self.insert_default_value(tablename)
+            self.make_column_non_nullable(tablename)
+
+    def add_column(self, tablename):
+        self.op.add_column(
+            tablename,
+            Column('hidden', Boolean, default=False, nullable=True))
+
+    def insert_default_value(self, tablename):
+        contact_table = table(
```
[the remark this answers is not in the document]
njohner: 😅

opengever/ogds/models/service.py
njohner: Yes, I wanted to avoid `hidden= not visible_only`. I'll update the one with `enabled`

opengever/ogds/base/tests/test_vocabularies.py
njohner: No I think I'd rather leave as is. It does not contain `disabled` org units for example (which is the test that comes just before). So this test, tests that the hidden ones appear.

opengever/ogds/models/tests/test_service.py
njohner: No, one os for `disabled` and the other one for `hidden` orgunits.

PR #6320, opened 2020-03-13
Add team and user listing enpoints to API.

deiferni:
With this PR we add two more endpoints to list users and teams from ogds, `@user-listing` and `@team-listing`. They return the same structure as the already existing `@listing` endpoint, which uses solr as a data source. In contrary to what was originally specified in https://github.com/4teamwork/gever-ui/issues/780 we have decided to implement separate endpoints rather than marrying sql based queries into the solr based endpoint, reasoning:

- query building for sql and solr is entirely different
- result handling for sql queries is also different
- endpoints can also be available if there is no solr
- we can register the endpoints for other content types, currently it is only available on an `IContactFolder`
- we've discussed to maybe split `@listing` and provide one separate endpoint for each content type that currently has to be specified via `name` parameter. the  changes in this PR can be seen as a first step in that direction
- only common thing is same subset of url query string parameters, i've factored these out into their own method so that it can be unified if desired

This PR provides a base service for sql/ods based listings `OgdsListingBaseService`. I have been able to reuse some code that was already available for the `ftw.tabbedview` based listings and thus is already battle tested. Most of it comes from `SqlTableSource` https://github.com/4teamwork/opengever.core/blob/d2a7d1c5927d1c35bd1170cc98ad7432708b65d7/opengever/tabbedview/sqlsource.py#L59 or the concrete tables sources for the already existing listings `UsersListingTableSource` https://github.com/4teamwork/opengever.core/blob/d2a7d1c5927d1c35bd1170cc98ad7432708b65d7/opengever/tabbedview/browser/users.py#L110 and `TeamsListingTableSource` https://github.com/4teamwork/opengever.core/blob/d2a7d1c5927d1c35bd1170cc98ad7432708b65d7/opengever/contact/browser/tabs.py#L85. I've decided to create a separate implementation as marrying the new API based listings and the `tabbedview` listings would IMO be overkill at the moment. The `tabbedview` based listings will go away eventually and i'd rather keep the new implementation separate.

As a preparation for https://github.com/4teamwork/gever-ui/issues/780 we also add endpoints `@user` and `@team`. This is currently a stub for the UI so that navigation and the "this feature is not yet available" page work correctly. They are also only available for `IContactFolder`, maybe we need to change this to easily allow access to user information on any context. I was not sure though, so i kept it simple for now.

Things that might need to be discussed for this PR:
- Separate endpoints OK?
- Strategy to split `@listing` OK? I can create issues for that if desired.
- Does limiting endpoint availability to `IContactFolder` make sense? My reasoning was that customers could block inheritance on `IContactFolder` if access to these endpoints is undesired for the whole user base for some reason.
- I'm not 100% happy with the state filter implementation. The current implementation has been guided by what filters are available/implemented in the UI. I can create issues to change it, or adapt this PR if the reviewer dislikes the solution strongly. I would have something that would make more sense for a boolean filter in mind, e.g.: 
  - `/@team-listing?filters.state:record:=true`, only active
  - `/@team-listing?filters.state:record:=false`, only inactive
  - `/@team-listing`, all states

The endpoints implemented here are for https://github.com/4teamwork/gever-ui/issues/780.

## Checkliste

### DIFF
--- a/docs/public/dev-manual/api/listings.rst
+++ b/docs/public/dev-manual/api/listings.rst
@@ -253,3 +253,91 @@ Optionale Parameter:
 
     GET /ordnungssystem/fuehrung/dossier-23/@listing?name=documents&facets:list=creator HTTP/1.1
     Accept: application/json
+
+
+Auflistungen User und Teams
+===========================
+
+Mit den Endpoints ``@ogds-user-listing`` und ``@team-listing`` können Benutzer und
+Teams aus dem ogds aufgelistet werden. Diese beiden Endpoints liefern
+inhaltlich die gleiche Struktur wie der ``@listing`` Endpoint, unterstützen
+aber nur ein Subset der Parameter. Im Moment ist es nicht möglich die
+``columns`` anzugeben, sondern es werden immer alle vom vom Modell
+untertstützten Attribute zurückgegeben. Des weiteren ist der ``depth``
+Paremeter nicht implementiert, ``facets`` werden ebenfalls nicht unterstützt.
+Dies weil die Datenquelle eine SQL-Datenbank und nicht Solr ist.
+
+
+Beispiel: Auflistung aller Benutzer:
+
+  .. sourcecode:: http
+
+    GET /kontakte/kontakte/@ogds-user-listing HTTP/1.1
+    Accept: application/json
+
+  .. sourcecode:: http
+
+    HTTP/1.1 200 OK
+    Content-Type: application/json
+
+    {
+      "@id": "http://localhost:8080/fd/kontakte/@ogds-user-listing",
+      "b_size": 25,
+      "b_start": 0,
+      "facets": {},
+      "items": [
+        {
+          "@id": "http://localhost:8080/fd/kontakte/@ogds-user/sandro.ackermann",
+          "@type": "virtual.ogds.user",
+          "active": true,
+          "department": null,
+          "directorate": null,
+          "email": "sandro.ackermann@example.com",
+          "firstname": "Sandro",
+          "lastname": "Ackermann",
+          "phone_office": null,
+          "title": "Ackermann Sandro",
+          "userid": "sandro.ackermann"
+        },
+      ],
+      "items_total": 1
+    }
+
+
+Optionale Parameter:
+--------------------
+Folgende Parameter werden im Moment unterstützt:
+
+- ``b_start``: Das erste zurückzugebende Element
+- ``b_size``: Die maximale Anzahl der zurückzugebenden Elemente
+- ``sort_on``: Sortierung nach einem indexierten Feld
+- ``sort_order``: Sortierreihenfolge: ``ascending`` (aufsteigend) oder ``descending`` (absteigend)
+- ``search``: Filterung nach einem beliebigen Suchbegriff
+- ``filters``: Einschränkung nach einem bestimmten Wert eines Feldes
+
+
+Filtern:
+--------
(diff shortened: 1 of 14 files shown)


docs/public/dev-manual/api/listings.rst
```diff
@@ -253,3 +253,91 @@ Optionale Parameter:
...
+- ``sort_on``: Sortierung nach einem indexierten Feld
+- ``sort_order``: Sortierreihenfolge: ``ascending`` (aufsteigend) oder ``descending`` (absteigend)
+- ``search``: Filterung nach einem beliebigen Suchbegriff
+- ``filters``: Einschränkung nach einem bestimmten Wert eines Feldes
+
+
+Filtern:
+--------
+Im Moment ist für beide Endpoinst nur ein Filter nach Status (aktiv/inaktiv)
+implementiert. Mit ``filters.state:record:list`` können die gewünschten Stati
```
lukasgraf: Stati -> Status 😉

opengever/api/teamlisting.py
```diff
@@ -0,0 +1,52 @@
...
+    default_sort_on = 'title'
+    model_class = Team
+    default_state_filter = tuple()
+
+    def fill_item(self, item, model):
+        item = super(TeamListingGet, self).fill_item(item, model)
+        item['org_unit_title'] = model.org_unit.title
+        item['@type'] = 'virtual.ogds.team'
+        # currently a dummy link but required by the UI
+        item['@id'] = '{}/team/{}'.format(
```
lukasgraf: Shouldn't this point to the `@team` Service? It's also not really a dummy anymore I think.

opengever/api/userlisting.py
```diff
@@ -0,0 +1,60 @@
...
+
+    default_sort_on = 'lastname'
+    model_class = User
+    default_state_filter = tuple()
+
+    def fill_item(self, item, model):
+        item = super(UserListingGet, self).fill_item(item, model)
+        item['@type'] = 'virtual.ogds.user'
+        item['title'] = model.fullname()
+        # currently a dummy link but required by the UI
```
lukasgraf: Not a dummy link any more, right?

lukasgraf:
> I have been able to reuse some code that was already available for the ftw.tabbedview based listings and thus is already battle tested.

Yay! :tada:

> Most of it comes from `SqlTableSource`

Oh dear.

> - Separate endpoints OK?

👍
 
> - Strategy to split `@listing` OK? I can create issues for that if desired.

👍 

> - Does limiting endpoint availability to `IContactFolder` make sense? My reasoning was that customers could block inheritance on `IContactFolder` if access to these endpoints is undesired for the whole user base for some reason.

👍
 
> - I'm not 100% happy with the state filter implementation. The current implementation has been guided by what filters are available/implemented in the UI. I can create issues to change it, or adapt this PR if the reviewer dislikes the solution strongly.


I agree in principle, but I think it's fine for now.

opengever/api/teamlisting.py
deiferni: you're right, it points to a stub, but the link itself is valid an could remain as is.

deiferni:
Thanks for the review, have updated the PR with all your suggestions.

The naming originated by me having an `@user-listing` and an `@ogds-user` endpoint, i wanted to consistently name them. Somehow i decided to drop `ogds` rather than adding it to the listing endpoint name.  The plone.restapi endpoint is called `@users` so that would no shadow the `@user` endpoint but you're right it is confusing nevertheless. I have now consistently prefixed the user related endpoints with `ogds`.

deiferni:
@all4manu fyi: the listing views are merged now.

PR #6322, opened 2020-03-13
OfficeOnline: Use collaborative checkout / checkins

lukasgraf:
This introduces the concept of **collaborative checkouts** and applies it to the **OfficeOnline** (WOPI) integration.

> ℹ️  This branch is currently deployed to https://lab.onegovgever.ch/ for easier review 

With this change,
- Documents may be checked in collaborators other than the user that initially checked out the document (_addresses #6229_)
- Collaborators are written to version and journal comment on checkin (_addresses #6230_)

**Caveat**
- The version and journal comments get written as a fixed-language string (in the site's default language at the time of checkin). 

## Checkliste

### DIFF
--- a/opengever/document/locales/fr/LC_MESSAGES/opengever.document.po
+++ b/opengever/document/locales/fr/LC_MESSAGES/opengever.document.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
-"POT-Creation-Date: 2019-11-04 13:11+0000\n"
+"POT-Creation-Date: 2020-03-13 14:31+0000\n"
 "PO-Revision-Date: 2017-09-03 08:50+0000\n"
 "Last-Translator: Jacqueline Sposato <jacqueline.sposato@gmail.com>\n"
 "Language-Team: French <https://translations.onegovgever.ch/projects/onegov-gever/opengever-document/fr/>\n"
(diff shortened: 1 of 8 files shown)


opengever/document/tests/test_checkout.py
```diff
@@ -73,6 +73,95 @@ def test_force_checkin_clears_lock(self):
...
+
+        # List of collaborators starts out with the user that initially
+        # checked out the document
+        self.assertEqual([self.regular_user.getName()],
+                         manager.get_collaborators())
+
+        # Other collaborator can checkin
+        manager.add_collaborator(self.dossier_responsible.getName())
+        self.assertEqual(
+            [self.regular_user.getName(), self.dossier_responsible.getName()],
```
njohner: shouldn't that be `self.regular_user..getId())`?

opengever/document/tests/test_checkout.py
```diff
@@ -73,6 +73,95 @@ def test_force_checkin_clears_lock(self):
         manager.checkin(comment="Force checkin")
         self.assertFalse(IRefreshableLockable(self.document).locked())
 
+    def test_collaborative_checkin(self):
+        self.login(self.regular_user)
+        manager = getMultiAdapter((self.document, self.portal.REQUEST),
+                                  ICheckinCheckoutManager)
+
+        manager.checkout(collaborative=True)
+        self.assertEqual(self.regular_user.getName(),
```
njohner: shouldn't that be `self.regular_user..getId())`?

opengever/document/locales/fr/LC_MESSAGES/opengever.document.po
```diff
@@ -442,6 +437,11 @@ msgstr "Disponble sous forme numérique"
 msgid "label_document_checked_out_by"
 msgstr "En cours de traitement"
 
+#. Default: "Collaborators: ${collaborators}"
+#: ./opengever/document/checkout/manager.py
+msgid "label_document_collaborators"
+msgstr "Collaborateurs: ${collaborators}"
```
njohner: `Edité par: ${collaborators}` would be more consistent with the German verison

opengever/document/tests/test_checkout.py
njohner: Ok seems to be the same thing, sorry

opengever/document/tests/test_checkout.py
njohner: as above 👍

opengever/document/locales/fr/LC_MESSAGES/opengever.document.po
lukasgraf: Thanks! Updated 👍

lukasgraf:
@njohner I applied all your suggestions. I still changed `getName()` to `getId()`, I think it's less confusing. I'm still testing on lab for a last time, and will then self-merge.

lukasgraf:
Problem with the "site-language not being respected" (or so I thought) was me not running buildout on `lab` after adding new translations. But that's required, because for new style deployments we need to precompile `.mo` files during buildout time 🤦‍♂

PR #6332, opened 2020-03-18
Add a new profile to setup a cas auth plugin for the new portal

elioschmutz:
The new Ianus-Portal provides a differen CAS-Auth URL than the old portal. In the PR: https://github.com/4teamwork/opengever.core/pull/6296 I already tried to permanently use the new url. But that was declined because we still use the old portal for several new deployments.

The URL is part of the plugin and will be set during a profile-hook. If we want to use the new cas-url, we would have to manually adjust it after installation or have to create a new fake-profile in the deployment repositories which will be installed after the `casauth` profile to get an additional hook to reset the URL.

This PR adds a brand new profile `casauth_ianus_portal` which already sets the URL to the new endpoint. We can now just use this new profile instead the old `casauth` to get a deployment with the correct url.

## Checkliste

### DIFF
--- a/opengever/base/casauth.py
+++ b/opengever/base/casauth.py
@@ -53,11 +53,11 @@ def get_cas_server_url():
     return url
 
 
-def build_cas_server_url():
+def build_cas_server_url(cas_path):
     """Build the URL to the CAS server, based on the GEVER cluster URL.
 
     This URL should only be constructed once during setup, in order to
     initially configure the ftw.casauth plugin.
     """
     base_url = get_cluster_base_url()
-    return urljoin(base_url, 'portal')
+    return urljoin(base_url, cas_path)
--- a/opengever/setup/casauth.py
+++ b/opengever/setup/casauth.py
@@ -7,21 +7,29 @@ from Products.PluggableAuthService.interfaces.plugins import IChallengePlugin
 SESSION_DEFAULT_TIMEOUT = 24 * 60 * 60
 
 
-def setup_cas(site):
+def setup_cas_gever_portal(site):
+    setup_cas(site, cas_path='portal')
+
+
+def setup_cas_ianus_portal(site):
+    setup_cas(site, cas_path='portal/cas')
+
+
+def setup_cas(site, cas_path):
     """Install and configure CAS authentication plugin.
        Rename session cookie to avoid conflicts with multiple clients.
     """
-    install_cas_auth_plugin()
+    install_cas_auth_plugin(cas_path)
     rename_session_cookie(site)
 
 
-def install_cas_auth_plugin():
+def install_cas_auth_plugin(cas_path):
     acl_users = api.portal.get_tool('acl_users')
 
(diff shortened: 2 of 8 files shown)


opengever/base/casauth.py
```diff
@@ -53,11 +53,15 @@ def get_cas_server_url():
     return url
 
 
-def build_cas_server_url():
+def build_cas_server_url(new_portal=False):
```
deiferni: a naming suggestion, i'd prefer other names to new and old. might be best to go with project names according to github so `ianus_portal` and `gever-portal`.

opengever/setup/casauth.py
```diff
@@ -7,21 +7,29 @@
...
+def setup_cas_new_portal(site):
+    setup_cas(site, new_portal=True)
+
+
+def setup_cas(site, new_portal=False):
     """Install and configure CAS authentication plugin.
        Rename session cookie to avoid conflicts with multiple clients.
     """
-    install_cas_auth_plugin()
+    install_cas_auth_plugin(new_portal)
```
deiferni: `new_portal` is a keyword argument, so please pass the `new_portal` parameter as a keyword argument, too.

opengever/setup/casauth.py
```diff
@@ -7,21 +7,29 @@
 SESSION_DEFAULT_TIMEOUT = 24 * 60 * 60
 
 
-def setup_cas(site):
+def setup_cas_old_portal(site):
+    setup_cas(site, new_portal=False)
```
deiferni: on second thought, why not inject the path here? so just pass on the path as an argument. that way one can see the different configuration in the outermost method/handler.

elioschmutz: 👍

opengever/setup/profiles.zcml
```diff
@@ -39,7 +39,20 @@
...
   <profilehook:hook
       profile="opengever.setup:casauth"
-      handler=".casauth.setup_cas"
+      handler=".casauth.setup_cas_gever_portal"
+      />
+
+  <genericsetup:registerProfile
+      name="casauth_ianus_portal"
+      title="opengever.setup: casauth ianus portal"
+      directory="profiles/casauth"
```
deiferni: oh, i just spotted this, this seems to point to the same directory as the existing casauth profile. is this on purpose? i don't recall other profiles sharing directories. maybe it is technically possible, but i fear it is very confusing for developers. i'd rather copy the existing profile and use a different directory here.

elioschmutz: Ok... basically, I just need another hook. Thats because I use the same profile. Would it help if I add a comment instead copying the whole profile folder?

deiferni: ok as just discussed by phone:
- add a separate directory, a copy
- migration path should work as we can uninstall the old profile and install the new one. the action remains the same, or should be updated as the new one uses the same name.

PR #6334, opened 2020-03-18
Do no longer send activity mails from the current user.

elioschmutz:
An activity mail should always be sent by the system itself and not under the users email address. It is possible that the user is registered by with a different email address than the portal domain. This will end up in spam issues.

Issuer: #6258  (https://github.com/4teamwork/opengever.core/issues/6258#issuecomment-600501160)

## Checkliste

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -9,6 +9,7 @@ Changelog
 - Extend the document serialization with `checked_out_fullname`. [elioschmutz]
 - Add a new profile to setup a cas auth plugin for the ianus portal. [elioschmutz]
 - Fix encoding issue in query-source `query` parameter. [deiferni]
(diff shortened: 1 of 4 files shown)


docs/HISTORY.txt
```diff
@@ -7,6 +7,7 @@ Changelog
 - Extend the document serialization with `checked_out_fullname`. [elioschmutz]
 - Add a new profile to setup a cas auth plugin for the ianus portal. [elioschmutz]
 - Fix encoding issue in query-source `query` parameter. [deiferni]
+- Do no longer send activity mails from the current user due to spam issues when the users email address does not match the portal domain. [elioschmutz]
```
njohner: Same thing here: `the users email` -> `the user's email`

PR #6336, opened 2020-03-18
Also return @id in globalindex endpoint.

njohner:
We update the `@globalindex` endpoint to also return the url to the tasks in the `@id` attribute.
This is needed so that the new frontend can link the tasks in the listing.

For https://github.com/4teamwork/gever-ui/issues/962

## Checkliste


njohner:
Sorry for the tests but I already force-pushed 3 times and they are just not running

PR #6337, opened 2020-03-18
Extend the document serialization with `checked_out_fullname`.

elioschmutz:
This PR extends the document serialization with the `checked_out_fullname` property.

Issuer: https://github.com/4teamwork/gever-ui/issues/982

## Checkliste

### DIFF
--- a/opengever/api/document.py
+++ b/opengever/api/document.py
@@ -1,10 +1,11 @@
 from ftw import bumblebee
+from opengever.api.serializer import GeverSerializeToJson
+from opengever.base.helpers import display_name
 from opengever.base.interfaces import IReferenceNumber
 from opengever.document.behaviors import IBaseDocument
 from opengever.document.interfaces import ICheckinCheckoutManager
 from plone.restapi.deserializer import json_body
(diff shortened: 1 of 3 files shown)


all4manu:
There is something a bit strange on my machine:



"Gross" or "Abeggle"?

Can this be ignored as a local phenomenon or are there two data tables where we can get a user's full name?

PS: The UI queries the `@users` endpoint for detailed information about a user.

lukasgraf:
@all4manu that's a phenomenon in the sense that someone changed the last name for that test account to "Abeggle" at some point (in the LDAP). However, that was months ago, I'm curious as to where you even get "Gross" from.

Do I understand that correctly that the yellow message fetches its info from document `GET` -> `checked_out_fullname` and the other label from `@users`? They should be in sync.

opengever/api/document.py
```diff
@@ -1,10 +1,11 @@
 from ftw import bumblebee
+from opengever.api.serializer import GeverSerializeToJson
```
lukasgraf: Unused import?

PR #6356, opened 2020-03-26
Backdate AddHiddenFlagToAdminAndOrgUnit upgrade step.

njohner:
The current admin unit is fetched every time the metadata of an object is updated, because of the reference number indexer. This will lead to failing upgrade steps if we do not update the admin_units table first. So we predate this upgrade step to just before the oldest upgrade that has not been executed on all client installations yet.

### DIFF
--- a/opengever/core/upgrades/20170601103050_add_hidden_flag_to_admin_and_org_units/upgrade.py
+++ b/opengever/core/upgrades/20170601103050_add_hidden_flag_to_admin_and_org_units/upgrade.py
@@ -11,6 +11,8 @@ class AddHiddenFlagToAdminAndOrgUnits(SchemaMigration):
 
     def migrate(self):
         for tablename in ['admin_units', 'org_units']:
+            if self.has_column(tablename, "hidden"):
+                continue
             self.add_column(tablename)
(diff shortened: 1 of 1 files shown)


opengever/core/upgrades/20170601103050_add_hidden_flag_to_admin_and_org_units/upgrade.py
```diff
@@ -32,3 +34,7 @@ def insert_default_value(self, tablename):
     def make_column_non_nullable(self, tablename):
         self.op.alter_column(tablename, 'hidden',
                              existing_type=Boolean, nullable=False)
+
+    def has_column(self, table_name, column_name):
+        table = self.metadata.tables.get(table_name)
```
deiferni: just one question out of carefulness, this works when the column does not exist, right? it inspects the database state, and not some metadata definition for `hidden` which would already exist when the new code with the `hidden` `Column` is loaded?

njohner: yes. Just tested.

PR #6357, opened 2020-03-26
Prevent attempts to edit exclusively checked out documents in Office Online.

lukasgraf:
This will prevent attempts to edit exclusively checked out documents (by OfficeConnector, usually) with OfficeOnline (because that would lead to conflicting changes / data loss).

Measures to prevent this are added at three levels:

### WOPIView

This view will be used by the OfficeOnline client to start (an manage) an editing session - it's the implementation of the **WOPI protocol**. This is the last line of defense where we need to prevent conflicting checkouts.

In this view we reject the Office Online client's attempt to`LOCK` the document if it's already checked out _exclusively_. The LOCK request will be answered with `400 Bad Request`, which will lead to the following behavior in the Office Online client:

oo_cant_open_for_editing

Trying to open the document in read only mode doesn't really worked (haven't looked into why):

oo_cant_open_overloaded

But this path fails in a _reasonably_ graceful way in my opinion, given that the user should never really be able to end up in this situation.

### `office_online_edit` view

This view is visited by the user to launch (or join) an Office Online editing session and it renders the frame in which the editing session will run.

Immediately when invoking this view, we will check whether the document is already checked out for exclusive editing. If so, a portal status message is displayed, and the user is redirected back to the context:

error_message_checked_out_for_exclusive_editing

This will be a bit clunky in the new frontend:
- User is on the new frontend, and opens a link to the `/office_online_edit` view.
- Rewrite rules will always dispatch this view to the backend
- The view will add a cookie for the status message, and redirect to the document
- The request to the document will end up on the new frontend again, because of the user's `geverui=1` cookie. The frontend doesn't know / care about Plone statusmessage cookies though, so will do nothing out of the ordinary.
- For the user, this will look like a quick page refresh, except that the "Edit in Office Online" action won't be shown anymore afterwards.

But, as discussed during daily, we think we can currently live with that.

### File action buttons on documents

The "Edit in Office Online" button is disabled immediately after clicking the button that launches OfficeConnector:

office_online_disabled

## Checkliste


opengever/document/static/officeconnector.js
```diff
@@ -61,11 +61,18 @@
...
         var checkoutButton = $(event.currentTarget);
+        var officeOnlineButton = $('#office-online-edit');
         var documentURL = checkoutButton.data("document-url");
         if(checkoutButton.hasClass('oc-loading')) {
             return false;
         }
         checkoutButton.addClass("oc-loading");
+
+        // Disable "Edit in Office Online" button
+        var disabledOfficeOnlineButton = $('<span class="function-edit-inactive discreet" />');
```
bierik: I'd rather do:
``` javascript
officeOnlineButton.removeClass('function-edit').addClass('function-edit-inactive discreet');
```

lukasgraf: @bierik that's what I tried first, but that's [not enough](https://github.com/4teamwork/opengever.core/blob/master/plonetheme/teamraum/resources/css/gever/partials/_action-buttons.scss#L153-L159). I would need to add a `a.function-edit-inactive` selector to the SCSS, and also kill the `href` attribute of the `<a />`, otherwise it's still clickable.

all4manu: @lukasgraf I think replacing the anchor tag with a span is the way to do it. There is not such a thing as a disabled link - it would not make sense in HTML.

lukasgraf: @all4manu thanks, that was my thinking as well. Replacing the element is a bit clunky, but bastardizing the link felt worse 😉

lukasgraf:
@deiferni sigh, you're right of course. Will have a look at this tomorrow. Thanks again for the thorough review!

lukasgraf:
As discussed: Will ship this as-is for now, and address the other direction (disabling OC when edited in OfficeOnline) in a separate PR.

PR #6358, opened 2020-03-27
Bump ftw.solr version to 2.8.2.

njohner:
This fixes a rounding error in DateTime fields which led to indexing in solr failing for certain objects.

This means we will need to run a solr sync for all deployments. (added to the SaaS checklist https://github.com/4teamwork/opengever.ftw/issues/110)


njohner:
Sure, I had more info info in the commit, but can add it in the CL as well

njohner:
voilà...

PR #6360, opened 2020-03-30
Merge base.helper and base.helpers files.

njohner:
There is no reason to have two different helper modules here, this only clutters the repository.
This is just some cleanup, no associated issue and also did not add a changelog for that.


njohner:
Oh didn't think of that. Ok we'll see...

PR #6367, opened 2020-04-02
Hide OC actions when user is in EMM environment

njohner:
In this PR we add functionality to hide buttons that would triggere Office Connector when the User is on an EMM environment. 
Implementation:
- For this we add a configuration for ip ranges which are blacklisted for OfficeConnector actions. 
- When the client IP address is in the blacklisted ranges:
  - we do not display the OC actions (checkout and edit, edit...)
  - The `@actions` endpoint also does not return the OC actions
  - We omit the `edit_after_creation` field when 

For https://github.com/4teamwork/opengever.sg/issues/404

## Checkliste (Must have)

- [ ] Aktualisierung Dokumentation (API/Deployment/...) aktualisiert? _Falls im PR nicht vorhanden erachtet es der PR-Ersteller als unnötig._
- [ ] Jira Link + Backlink im Jira / Github Issue Link. _Link muss im PR-Body vorhanden sein._

## Checkliste (optional)

### DIFF
--- a/opengever/officeconnector/helpers.py
+++ b/opengever/officeconnector/helpers.py
@@ -32,9 +32,10 @@ def is_client_ip_in_office_connector_disallowed_ip_ranges():
     request = getRequest()
     client_ip = request.getClientAddr()
     if not client_ip:
-        # trusted proxies are not set up correctly so we cannot
-        # check whether office connector should be disallowed for
-        # this request.
+        # If for some reason we cannot determine the client_ip, we
+        # cannot check whether office connector should be disallowed
+        # for this request. This could be a problem with the configuration
+        # of trusted proxies.
         log_msg_to_sentry('Cannot determine client IP.', request=request)
         return False
 
--- a/opengever/officeconnector/helpers.py
+++ b/opengever/officeconnector/helpers.py
@@ -1,5 +1,7 @@
 from Acquisition import aq_inner
 from Acquisition import aq_parent
+from opengever.base.ip_range import is_in_ip_range
+from opengever.base.sentry import log_msg_to_sentry
 from opengever.document.behaviors import IBaseDocument
 from opengever.officeconnector.interfaces import IOfficeConnectorSettings
 from plone import api
@@ -8,6 +10,7 @@ from Products.PluggableAuthService.interfaces.plugins import IAuthenticationPlug
 from ua_parser.user_agent_parser import ParseUserAgent
 from zExceptions import Forbidden
 from zExceptions import NotFound
+from zope.globalrequest import getRequest
 import json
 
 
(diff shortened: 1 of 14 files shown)


opengever/officeconnector/helpers.py
```diff
@@ -25,6 +28,23 @@ def is_officeconnector_checkout_feature_enabled():
         )
 
 
+def is_client_ip_in_office_connector_disallowed_ip_range():
+    request = getRequest()
+    client_ip = request.getClientAddr()
+    if not client_ip:
```
lukasgraf, line 34: I'm quite certain that if `trusted-proxies` is not set, there will still be a source IP. It will just be the one of the reverse proxy (HAProxy or Apache in our case) that dispatches the request to the GEVER backend.

opengever/officeconnector/interfaces.py
```diff
@@ -28,6 +29,18 @@ class IOfficeConnectorSettings(Interface):
         u'These are subtractive from the default list.',
         default=[])
 
+    office_connector_disallowed_ip_range = schema.TextLine(
```
lukasgraf, line 32: I would make that plural (`...ranges`), since it should also be able to be a list of ranges, list of IPs, or a mix of them. And add a quick test for one comma separated case.

lukasgraf:
@njohner looks good in a preliminary review, just some minor things.

lukasgraf:
I couldn't really deploy this one on `lab` though unfortunately, because it requires an upgrade step, running that one could make it tricky to backtrack if its contents were still to change.

opengever/officeconnector/helpers.py
njohner: That's a check you have in `ftw.tokenauth`. I figured it makes sense to check that we were able to determine the client_ip. If not, obviously we do not want to deactivate OC buttons.
But I can change the comment to something less categorical?

lukasgraf: @njohner oh wow, ok :) Hmm, maybe I misremember and it's actually true that `getClientAddr` returns `None` if `trusted-proxies` is not set at all. And it's just that if it's set, but incorrectly, it might return the IP of the reverse proxy.

Either way, 👍 for your suggestion to just make the comment less categorical.

njohner:
Ok this should be ready.

njohner:
I put the new things and corrections into new commits

opengever/base/configuration.py
```diff
@@ -82,6 +82,7 @@ def get_settings(self):
         settings['oneoffixx_settings'] = self.get_oneoffixx_settings()
         settings['user_settings'] = self.get_user_settings()
         settings['sharing_configuration'] = self.get_sharing_configuration()
+        settings['office_connector_disallowed_ip_range'] = api.portal.get_registry_record('office_connector_disallowed_ip_range', interface=IOfficeConnectorSettings)  # noqa
```
lukasgraf: I wouldn't disclose the actual IP range here in the `@config` endpoint. For one, that endpoint is public, and so is accessible for Anonymous users.

And in my opinion a client has no business knowing about this, this information should stay in the backend. Exposing the derived information, `is_emm_environment`, should be enough.

Issue #6350, opened 2020-03-24
TabbedView fulltext filtering isn't batched when using Solr


lukasgraf:
When Solr is enabled, the TabbedView fulltext filtering **hard-limits results to 50 entries**, with no batching, and cuts off any results beyond that.

Extranet Ticket: https://extranet.4teamwork.ch/extern/opengever-kanton-zug/sprint-backlog/20477/

![Screenshot 2020-03-24 at 16 00 33](https://user-images.githubusercontent.com/405124/77440478-a49fd580-6de8-11ea-9e5e-14950fcca511.png)

### Steps to reproduce

- Set up GEVER from `master`
- Enable Solr
- Create more than 50 dossiers (e.g. by importing a bundle)
- Visit the repository root
- Switch to the `Dossiers` tab (classic UI)
- Filter for something that will produce more than 50 results

### Expected behavior
- All results are accessible, preferrably batched, and respecting the user's dynamic batch size settings

### Actual behavior
- Only the first 50 results are shown, no batching, and the user interface suggests that there's only exactly 50 results.

---

### First analysis

This seems to happen mainly for two reasons:
- Batching support simply isn't implemented for Solr results in the table source. Solr results are [hard limited to 50 entries](https://github.com/4teamwork/opengever.core/blob/999f5cff4b88844e63bf426398b91368a8117897/opengever/tabbedview/catalog_source.py#L124-L127) and no batching aware datastructure (like a `plone.batching.Batch` or a `LazyMap`, with .actual_result_count`) is being returned
- For some reason, even if that is corrected, a custom sort method gets applied to the result batch. The [`default_custom_sort` will eliminate a `Batch` wrapper](https://github.com/4teamwork/ftw.table/blob/master/ftw/table/catalog_source.py#L26-L46) and turn the batch items into a plain list, destroying any batching information. I don't quite understand yet why custom sorting is being run on batched results anyway - that can't work properly.

In addition:

- TabbedView queries via Solr are [hacked into the existing `GeverCatalogTableSource`](https://github.com/4teamwork/opengever.core/blob/999f5cff4b88844e63bf426398b91368a8117897/opengever/tabbedview/catalog_source.py#L18-L31) using a conditional, instead of a dedicated `SolrTableSource`. That leads to inheriting a lot of behavior, much of which doesn't make sense for Solr.

deiferni:
classic ui, close.

PR #6382, opened 2020-04-23
Always use Solr in table source

buchi:
Always use Solr for tabbed view listings. Before that change Solr was used only if a filter text (SearchableText) was provided.

Enable Solr by default. This mainly affects testing as in most production environments Solr is enabled anyway. Because of this change several tests needed to be converted to `SolrIntegrationTests`.

Add two additional fields to the Solr index that were missing: `retention_expiration` and `email2`.

Due to running tests with Solr the following bugs were discovered and fixed:

Bugfixes for tabbed view listings:
- Exclude search root when using Solr
- Respect `trashed` parameter
- Support date range queries
- Fix contact lookup by email2
- Use same title format in OGDS sources with Solr

Other bugfixes:
- Fix a bug in the search view not respecting the batch size when solr is deactivated
- Fix Solr backup status check
- Fix indexing of documents `SolrIntegrationTests`

Jira: https://4teamwork.atlassian.net/browse/GEVER-207

## Checkliste (Must have)

_Alles muss gemacht/angehakt werden._

- [x] Aktualisierung Dokumentation (API/Deployment/...) aktualisiert? _Falls im PR nicht vorhanden erachtet es der PR-Ersteller als unnötig._
- [x] Jira Link + Backlink im Jira / Github Issue Link. _Link muss im PR-Body vorhanden sein._

## Checkliste (optional)

_Nur Zutreffendes soll angehakt stehengelassen werden._


njohner:
I hope I fixed most of the failing tests, but am not quite through. I also found two bugs that I corrected:
- `GeverCatalogTableSource` was not respecting the `trashed` parameter when solr is enabled
- Fix a bug in the search view not respecting the batch size when solr is deactivated. I also changed the default batch size here when solr is used to be the same as it was before without solr.

Moreover one test seems to give different results when run on CI or locally (`opengever.advancedsearch.tests.test_search.TestSearchWithContent. test_search_documents `). When executed locally the search finds 7 documents, because it also searches the content of the files, whereas on ci, it only finds the ones where the metadata matches the query.

lukasgraf:
@buchi I'm not quite done yet with the review, but when testing locally I noticed a pretty serious issue: Pagination seems to be broken for listings, at least in some cases. I believe this might be related to the issue described in #6350 (though that originally only applied to *filtered* listings, because only these were done via Solr). I remember batch sizes being hard coded to 50 somewhere in `solr_results`, it looks like this will now also affect listings.

**Steps to reproduce**
- Create a dossier with more than 50 documents
- Visit the "Documents" tab

**Expected behavior**
- First batch of 50 documents should be listed
- Pagination bar should be visible
- The "Select All" link on the top left should refer to the total amount (e.g. 58), not just the batch size (50)

**Actual behavior**
- Only the first batch page seems to be listed
- No pagination bar is rendered
- The amount in parentheses in the "Select All" is 50 (the batch size)

opengever/base/tests/test_search.py
```diff
@@ -12,6 +13,66 @@
...
+    @browsing
+    def test_batch_size_is_set_to_25(self, browser):
+        self.login(self.regular_user, browser=browser)
+
+        browser.open(self.portal, view='@@search')
+
+        self.assertEqual(25, len(browser.css('dl.searchResults dt')))
+
+    @browsing
+    def test_sorting_on_relevance_is_handled_correctly(self, browser):
```
lukasgraf, line 36: I don't quite see how this tests verifies what it claims in its name.

opengever/tabbedview/catalog_source.py
```diff
@@ -135,12 +139,14 @@ def solr_results(self, query):
...
 
         sort = query.get('sort_on', None)
-        if sort:
+        if sort and sort in solr.manager.schema.fields:
             sort_order = query.get('sort_order', 'ascending')
             if sort_order in ['descending', 'reverse']:
                 sort += ' desc'
             else:
                 sort += ' asc'
+        else:
```
lukasgraf, line 148: Would it make sense to log or [write to Sentry](https://github.com/4teamwork/opengever.core/blob/master/opengever/base/sentry.py#L55) here? Because the way I understand it, every time we end up here we basically have a mismatch where the user is told by the UI that results are sorted by a particular column, but they really aren't. Otherwise these subtle bugs are quite hard to catch I think.

opengever/core/upgrades/20200427154326_always_use_solr_for_table_source/upgrade.py
```diff
@@ -0,0 +1,42 @@
...
+from ftw.solr.interfaces import ISolrIndexHandler
+from ftw.upgrade import ProgressLogger
+from ftw.upgrade import UpgradeStep
+from opengever.contact.contact import IContact
+from opengever.dossier.behaviors.dossier import IDossierMarker
+from zope.component import getMultiAdapter
+from zope.component import getUtility
+
+
+class AlwaysUseSolrForTableSource(UpgradeStep):
```
lukasgraf, line 11: We might need to check how exactly this plays out when running upgrades via `update_deployment`. I vaguely remember Solr getting stopped by that script before running updates, but I'm not certain.

opengever/tabbedview/catalog_source.py
```diff
@@ -22,11 +22,11 @@ class GeverCatalogTableSource(FilteredTableSourceMixin, CatalogTableSource):
...
-        if 'SearchableText' in query:
-            registry = getUtility(IRegistry)
-            settings = registry.forInterface(ISearchSettings)
-            if settings.use_solr:
-                return self.solr_results(query)
+
+        registry = getUtility(IRegistry)
+        settings = registry.forInterface(ISearchSettings)
+        if settings.use_solr:
+            return self.solr_results(query)
```
lukasgraf: If `settings.use_solr` is `False` for some reason, this would fall through to regular catalog results. Is this intentional? Because while catalog results still work for pure listings, the `SearchableText` and `Title` are gone in deployments where Solr has been activated.

opengever/base/tests/test_search.py
buchi: Me neither. This test was already there and it seems that @njohner just duplicated it to run it once with Solr and once without Solr. I would just drop the tests without Solr.

opengever/core/upgrades/20200427154326_always_use_solr_for_table_source/upgrade.py
buchi: I think Solr must be running during upgrades. If not, anything that causes reindexing would lead to an outdated Solr index.

opengever/tabbedview/catalog_source.py
buchi: Yes, it's intentional. Currently it still can be used without Solr as long as it hasn't been activated using the `activate_solr.py` script which removes the `SearchableText` index.
I'm planning to remove the non-Solr path when cutting the catalog indexes.

buchi:
Pagination simply wasn't implemented yet. I've implemented that now.

It doesn't work in combination with grouping. But it also doesn't work correctly when using the catalog. Thus I'm leaving that out as it would exceed the scope.

opengever/tabbedview/catalog_source.py
buchi: Basically I don't like the idea of explicitly logging to Sentry in Code. Python has a logging system that fits well for this purpose. What goes to Sentry should be a configuration setting by defining different log levels for the various loggers.
When developing locally I prefer to have it logged to the console than to Sentry or nowhere because Sentry is not configured.
Thus I prefer to add a `logger.warning()`.

lukasgraf: I'm fine with just Python logging for now. But in my opinion that would be level `error`, since it's either a misconfiguration in Solr (because a field is missing from the schema that should be there), or a misconfiguration in tabbed view (if we present the user with an option to sort on a field that shouldn't be sortable).

deiferni: sorry, i just stumbled upon this, my two cents here:
- if i understand correctly if this happens we have a software bug and we would like to be aware of this, no? especially when it happens in production early after we have our new solr listings.
- if we just `log`, even with `error` level as suggested by @lukasgraf realistically we won't see the error in production, we only see it if we get notified in sentry
- the code is in gever, we currently have the strategy to explicitly log to sentry in some cases, so IMO it would be justifiable to do it here, too
- in the long run i agree, having log messages in sentry would make perfect sense. iirc we discussed it recently but it won't be in production before or at the same time we swap the listings to solr.

thus i would suggest to also temporarily add the explicit logging to sentry and get rid of again it once we have proper sentry logging configuration set up. this could be in addition to `logger`. what do you think?

lukasgraf:
Pagination looks good to me 👍

opengever/tabbedview/catalog_source.py
buchi: Sorry for disagreeing again. For me error level is something fatal to the involved operation while warning level is something that can cause potential application oddities (https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels).
I've now added warnings for unknown sort and filter fields.

deiferni: the log-level is fine 👍. we can merge this IMO but i'd like to discuss adding explicit sentry logging from https://github.com/4teamwork/opengever.core/blob/master/opengever/base/sentry.py#L55, if i understood correctly @lukasgraf is of the same opinion. we can amend this in a follow-up PR though, if we agree it is necessary.

lukasgraf: I also don't want to block the merge because of this, so, fine for now.

PR #6384, opened 2020-04-24
Add @allowed-roles-and-principals endpoint and extend @users endpoint

tinagerber:
Jira: https://4teamwork.atlassian.net/browse/GEVER-15

The RIS needs a way to list only those contracts for which a user has view permission. Therefore a new endpoint is implemented:

The @allowed-roles-and-principals endpoint (callable on any context) returns the information which roles, groups or users are allowed to view the object. The new endpoint is protected with a separate permission.

Furthermore, the response of the /@users/<user_id> endpoint must contain the information which "roles and principals" the user actually has (which global roles he has, which groups he is in, and which is his user ID).

With these two pieces of information about an object and the user, an external system can independently answer the question of whether a user will have view permission to a particular object in GEVER

## Checkliste (Must have)

_Alles muss gemacht/angehakt werden._

- [x] Aktualisierung Dokumentation (API/Deployment/...) aktualisiert? _Falls im PR nicht vorhanden erachtet es der PR-Ersteller als unnötig._
- [x] Jira Link + Backlink im Jira / Github Issue Link. _Link muss im PR-Body vorhanden sein._


opengever/api/allowed_roles_and_principals.py
```diff
@@ -0,0 +1,22 @@
...
+
+class AllowedRolesAndPrincipalsGet(Service):
+    """API Endpoint (callable on every context) which returns the information which roles,
+    groups or users are allowed to view an object.
+
+    GET /(path)/@allowed-roles-and-principals HTTP/1.1
+    """
+
+    def reply(self):
+        catalog = api.portal.get_tool('portal_catalog')
```
lukasgraf, line 13: Hmm. Why are you using the information from the catalog here, instead of the `Products.CMFPlone.CatalogTool. allowedRolesAndUsers` function? If the information in the catalog is stale for some reason, this endpoint would then return incorrect information.

opengever/api/tests/test_allowed_roles_and_principals.py
```diff
@@ -0,0 +1,34 @@
...
+                u'Editor',
+                u'Manager',
+                u'Reader',
+                u'principal:fa_inbox_users',
+                u'principal:fa_users',
+                u'principal:kathi.barfuss'],
+            u'@id': u'http://nohost/plone/ordnungssystem/fuhrung/vertrage-und-vereinbarungen'
+                    u'/dossier-1/@allowed-roles-and-principals'}
+
+        self.assertEquals(excepted_json, browser.json)
```
lukasgraf, line 34: Could you please extend this test so that it modifies the local roles for a user on `self.dossier`, and then does a second request to verify that the new entry is present in `allowed_roles_and_principals`?

So, [add a new local role](https://github.com/4teamwork/opengever.core/blob/master/opengever/task/tests/test_localroles.py#L173-L175) `[Reader]` for a different user `self.dossier_responsible` on `self.dossier`. That should result in `principal:robert.ziegler` showing up in `allowed_roles_and_principals`.

Something like:
```python
        RoleAssignmentManager(self.dossier).add_or_update_assignment(
            SharingRoleAssignment(self.dossier_responsible.getId(), ['Reader']))
```

opengever/api/allowed_roles_and_principals.py
tinagerber: 👍 On the first try I had a problem with the DelegatingIndexer class, but I think I have now figured out how to use the function

tinagerber:
@lukasgraf I have made the desired changes, can you look at it again?

PR #6389, opened 2020-04-29
Drastically improve performance when resolving large dossiers.

deiferni:
Resolving large dossiers with a lot of subdossiers and high nesting depth could be very slow. I'm adding the following performance improvements:

- Only resolve each subdossier once. Due to a bug in https://github.com/4teamwork/opengever.core/blob/c22e75c215c274b5c69b186e42dad14bc888c9ab/opengever/dossier/resolve.py#L371-L373 subdossiers could be resolved multiple times as each dossier dit not only recurse into its immediate children but instead into all subdossiers.
- Don't sort queries when we're not interested in order of objects
- Only reindex dossiers where the date actually changes
- Improve `has_sametype_children` indexer performance for dossiers by not loading dossier child document objects

I created a dossier with the same structure as for the customer issue reported in https://4teamwork.atlassian.net/browse/GEVER-199. The dossier contains about 1650 subdossiers and about 18'000 documents. I measured the time necessary to resolve the dossier with `ftw.flamegraph`:

- Time required before optimisation: Samples taken: 21244683, sample time: 7592855.20029 ms, duration: 30149126.673 ms
- Time after optimisation: Samples taken: 316332, sample time: 9619.41814423 ms, duration: 319907.74703 ms

So we managed to improve performance to resolve this example dossier from about 8 hours to approximately 5 minutes. This means it is about 100 times faster 🚀.

This is an addition to https://github.com/4teamwork/opengever.core/pull/6208 where we already optimised the pre-resolve checks.

The following flamegraph shows where we have further optimisation potential, most of the time is now spent in reindexing dossier metadata (about 60% request time):
Screenshot 2020-04-29 at 09 05 41
The reference number indexer and adapter lookup in general seem to need a lot of time:
Screenshot 2020-04-29 at 09 14 29

The `.svg` is attached to https://4teamwork.atlassian.net/browse/GEVER-199. Unfortunately i could not create a pre-optimization flamegraph as the instance crashed while it was generated.

## Checklist (Must have)

- [x] Changelog entry

## Further documentation

I'm putting my hacked together scripts to generate the large dossier in here, hoping they might be useful for someone in the future, or just for documentation purpose. apologies for the hacks:

1) printing large dossier information on console:

```python
from opengever.maintenance import dm
from plone import api

dm()

level = 0
dossier = plone.restrictedTraverse('/path/to/very/large/dossier')

def _recurse_dossiers(dossier, level):
    path = '/'.join(dossier.getPhysicalPath())
    document_brains = catalog({
        'path': {'query': path, 'depth': 1},
        'object_provides': [
            'opengever.document.behaviors.IBaseDocument',
            ],
        })
    print "{} - {} - {}".format(level, len(document_brains), path)
    subdossier_brains = dossier.get_subdossiers(depth=1)
    for brain in subdossier_brains:
        _recurse_dossiers(brain.getObject(), level + 1)
```

2) parse printed console text and create structure on local filesystem

```python

import os
import shutil

txt = """
<output from 1) goes in here>
"""

outpath = "/local/path/to/very/large/dossier-data/"
docpath = "/local/path/to/a/test-document/word.docx"

seen = set()
total_docs = 0
total_dossiers = 0

for line in txt.splitlines():
    if not line:
        continue

    level, nof_docs, path = line.split(' - ')
    segments = path.replace('/path/to/very/large/', '')

    if segments in seen:
        continue
    seen.add(segments)

    dirpath = outpath + segments
    if not os.path.exists(dirpath):
        os.makedirs(dirpath)
    total_dossiers += 1

    print level, dirpath
    for index in range(int(nof_docs)):
        shutil.copy(docpath, dirpath + '/{}.docx'.format(index))
    total_docs += int(nof_docs)

    print "{} documents copied".format(nof_docs)

print "created {} dossiers, {} documents".format(total_dossiers, total_docs)
```

3) create bundle from local filesystem
`bin/create-bundle /local/path/to/very/large/dossier-data/ /local/path/to/very/large/dossierbundle --import-repository-reference 2 --dossier-responsible david.erni`

4) load bundle into gever
`bin/instance import /local/path/to/very/large/dossierbundle/Sample-2020-04-22_15-31-11.oggbundle`


opengever/dossier/indexers.py
```diff
@@ -85,6 +86,23 @@ def blocked_local_roles(obj):
     return bool(getattr(aq_inner(obj), '__ac_local_roles_block__', False))
 
 
+@indexer(IDossierMarker)
+def has_sametype_children(obj):
```
njohner, line 90: Wouldn't it be faster in that case to simply make a catalog query?

opengever/dossier/base.py
```diff
@@ -329,15 +344,15 @@ def earliest_possible_end_date(self):
...
-            'use_changed_for_end_date', IDossierResolveProperties)
-        if use_changed:
-            date_name = 'changed'
-        else:
-            date_name = 'document_date'
+        earliest_possible_end_date = max_date(
+            self.earliest_possible_end_date_dossiers(),
+            self.earliest_possible_end_date_documents())
+
+        self._v_earliest_possible_end_date = earliest_possible_end_date
```
njohner, line 351: Is this used anywhere?

opengever/dossier/indexers.py
deiferni: yes, but wrong. catalog queryies during reindexing, well we've been there before 😄 and you fixed it in https://github.com/4teamwork/opengever.core/pull/5636.

opengever/dossier/base.py
deiferni: no it should not be here, thanks. removed it.
it was a leftover of an attempt at caching the result of this function. i dropped it again due to test isolation issue, also the performance gain from caching the values seems to be minimal as all brains are in memory already.

PR #6390, opened 2020-04-29
Add is_subdossier to catalog metadata.

deiferni:
It is already an index, but unfortunately not in metadata. We need it for some still catalog based api listings, namely when displaying item summaries for a parent in e.g. a `GET` request on the parent.

I've noticed that documents and other objects within dossiers seem to index the parents value via acquisition. I've disabled that behavior and registered one generic indexer which does return `None` and specific indexers for content types where we should access the `is_subdossier` method on the object.

I have only added an upgrade step for dossiers (and templatedossiers) as those are the only types where the new metadata field is truly relevant. This is a trade-off leaning towards upgrade-step performance.

it will slightly change the indexers behavior:
- the index will contain `False` for all objects unless it is a dossier and a subdossier, then it will be `True`, this is tue to a call to `bool` in the `BooleanIndex`
- the metadata will contain `False` for main dossiers, `True` for subdossiers and `None` for all other objects

See https://4teamwork.atlassian.net/browse/GEVER-122

## Checklist (Must have)

- [x] Changelog entry_

## Checklist (optional)

### DIFF
--- a/opengever/base/indexes.py
+++ b/opengever/base/indexes.py
@@ -92,3 +92,8 @@ def has_sametype_children(obj):
     # portal_type attribute.
     return any(obj.portal_type == getattr(aq_base(child), "portal_type", None)
                for child in obj.objectValues())
+
+
+@indexer(IDexterityContent)
+def is_subdossier(obj):
+    return None
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 ------------------------
 
 - Improve performance when resolving large dossiers. [deiferni]
+- Add is_subdossier to catalog metadata. [deiferni]
 - Add Bumblebee auto refresh feature to policy template. [2e12]
 - Task GET API: Also include info about containing dossier. [mbaechtold]
 - Fix `task_type_helper` to respect the current language for the ram-cache. [elioschmutz]
--- a/opengever/base/configure.zcml
+++ b/opengever/base/configure.zcml
@@ -381,6 +381,11 @@
       name="has_sametype_children"
       />
 
+  <adapter
+      factory=".indexes.is_subdossier"
+      name="is_subdossier"
+      />
+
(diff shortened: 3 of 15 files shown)


buchi:
Can you provide more details where this additional metadata is used? Adding more metadata to the catalog conflicts with the goal to reduce catalog indexes and metadata.
If it's just used for the parent object then this is not justified IMO. The parent object is loaded anyway, why not get it from the object?

deiferni:
yeah, it is used wherever we list things from the catalog which are not solr-based yet. one example is the summary of `items` that is displayed for a `GET` of a folderish. we display many items there, fetching the objects would probably hit performance too much.

but i'm aware of the conflict, i'm hoping this is temporary. Also it is "just" a boolean, so the memory consumption should be limited. Also, it is already a catalog index, i'm just adding it to metadata.

opengever/base/indexes.py
```diff
@@ -92,3 +92,8 @@ def has_sametype_children(obj):
     # portal_type attribute.
     return any(obj.portal_type == getattr(aq_base(child), "portal_type", None)
                for child in obj.objectValues())
+
+
+@indexer(IDexterityContent)
+def is_subdossier(obj):
```
lukasgraf, line 98: Wouldn't it make more sense to only have an indexer for dossiers (and template dossiers), but:
- First check whether the AQ-unwrapped object provides `IDossierMarker` or `IDossierTemplateMaker`
- If not, return `None`
- And only then check `is_subdossier()` (on the AQ-wrapped object)

deiferni: yes, i think you're missing acquisition. if we don't have a default indexer for all non-dossiers the default attribute-based indexer for e.g. a document would find the `is_subdossier` method from the parent dossier via acquisition and then index the dossiers value. that is what's happening right now already, but i'd argue that is an undesired side-effect and should not continue in the future.

PR #6393, opened 2020-05-01
Add @watchers API endpoint

tinagerber:
It should be possible to observe tasks as a watcher in order to stay informed.
This PR provides the API support for watchers. 

With the `@watcher` endpoint, it is possible to query all watchers of a task and their roles and add new watchers with the role `regula_watcher`. In addition, a user can delete his role as `regular_watcher` of a task.
Jira: https://4teamwork.atlassian.net/browse/GEVER-22

## Checklist (Must have)

- [x] Changelog entry

## Checklist (optional)

### DIFF
--- a/docs/public/dev-manual/api/watchers.rst
+++ b/docs/public/dev-manual/api/watchers.rst
@@ -1,0 +1,114 @@
+Beobachter
+==========
+
+Der ``@watchers`` Endpoint dient dazu, für Aufgaben und Weiterleitungen Beobachter zu registrieren.
+
+
+Auflistung
+----------
+
+Ein Beobachter kann verschiedene Rollen haben, beispielsweise die Rollen Auftraggeber (``task_issuer``), Auftragnehmer (``task_responsible``) oder Beobachter (``regular_watcher``). Mittels eines GET-Requests können alle Beobachter und alle Rollen einer Aufgabe abgefragt werden.
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+       GET /task-1/@watchers HTTP/1.1
+       Accept: application/json
+
+
+**Beispiel-Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+      {
+        "@id": "https://example.org/ordnungssystem/fuehrung/dossier-23/task-1/@watchers",
+        "watchers_and_roles": {
+          "rolf.ziegler": [
+            "regular_watcher",
+            "task_responsible"
+          ],
+          "peter.mueller": [
+            "task_issuer"
+          ]
+        }
+      }
+
+
+Beobachter als erweiterbare Komponente
+--------------------------------------
+
+Die Beobachter können als Kompomente einer Aufgabe direkt über den ``expand``-Parameter eingebettet werden, so dass keine zusätzliche Abfrage nötig ist.
+
+**Beispiel-Request**:
+
+  .. sourcecode:: http
+
+    GET /task-1?expand=watchers HTTP/1.1
+    Accept: application/json
+
+**Beispiel-Response**:
+
+  .. sourcecode:: http
+
+    HTTP/1.1 200 OK
+    Content-Type: application/json
+
+    {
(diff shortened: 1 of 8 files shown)


tinagerber:
Update: There was an error in the specification, so I have to revise this PR again.

tinagerber:
Now the PR is ready for the review.

opengever/api/watchers.py
```diff
@@ -0,0 +1,79 @@
...
+
+        result['watchers']['watchers_and_roles'] = self.get_watchers_and_roles()
+        return result
+
+    def get_watchers_and_roles(self):
+        watchers_and_roles = dict()
+        for subscription in self.center.get_subscriptions(self.context):
+            userid = subscription.watcher.actorid
+
+            if userid in watchers_and_roles:
```
lukasgraf, line 39: This would the ideal situation for a [`defaultdict`](https://docs.python.org/2/library/collections.html#collections.defaultdict):

```python
from collections import defaultdict

watchers_and_roles = defaultdict(list)
# ...
    watchers_and_roles[userid].append(subscription.role)
```

_(Basically, you can just start appending, not need to initialize the list - if the key is not in there, the defaultdict will create it for you)_

opengever/api/watchers.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def reply(self):
+        watchers = Watchers(self.context, self.request)
+        return watchers(expand=True)['watchers']
+
+
+class WatchersPost(Service):
+
+    def reply(self):
+
```
lukasgraf, line 56: ✂️

opengever/api/watchers.py
```diff
@@ -0,0 +1,79 @@
...
+        self.extract_data()
+        self.center = notification_center()
+        self.center.add_watcher_to_resource(self.context, self.userid, WATCHER_ROLE)
+        self.request.response.setStatus(204)
```
lukasgraf, line 60: Just a comment: Technically, a [successful POST](https://plonerestapi.readthedocs.io/en/latest/content.html#creating-a-resource-with-post) in `plone.restapi` should return `201 Created`, the JSON representation of the created resource in the body, and the URL to the new resource in the `Location:` header.

However, _the URL to the new resource_ would imply that we also would support something like `GET task-1/@watchers/robert.ziegler`. That doesn't make a whole lot of sense right now though. So I would stay pragmatic and leave it as is for now (204 and no body content) 👍

opengever/api/watchers.py
```diff
@@ -0,0 +1,79 @@
...
+        data = json_body(self.request)
+        self.userid = data.get("userid", None)
+        if not self.userid:
+            raise BadRequest("Property 'userid' is required")
+        if not api.user.get(self.userid):
+            raise BadRequest("userid '{}' does not exist".format(self.userid))
+
+
+class WatchersDelete(Service):
+    def reply(self):
```
lukasgraf, line 73: Currently, this will behave a bit strangely if someone tries to DELETE with a body `{"userid": "someone_else"}`. The endpoint will not complain, but completely ignore the body and instead remove the current user's watcher. So I would recommend to also override `extract_data()` in this endpoint, and just do a check that the request body is  == empty string. If not, raise a `BadRequest('DELETE does not take any parameters')` or something like that.

docs/public/dev-manual/api/watchers.rst
```diff
@@ -0,0 +1,112 @@
...
+
+
+**Beispiel-Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+      [
```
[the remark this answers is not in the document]
tinagerber: 👍 I changed it in the docs.

opengever/api/watchers.py
tinagerber: 👍 Thanks, I didn't know `defaultdict`.

opengever/api/watchers.py
tinagerber: 👍 Good point, I changed it.

tinagerber:
@lukasgraf Thanks for the review. I have made the desired changes, can you look at this again?

PR #6387, opened 2020-04-24
Task GET API: Also include info about containing dossier.

mbaechtold:
The response of a GET request of a task now includes the information about the containing dossier of the task. This is achieved by using a custom serializer for the task.

Belongs to the Jira task https://4teamwork.atlassian.net/browse/GEVER-245

## Checklist (Must have)

- [x] Changelog entry
- [x] Documentation updated (notably for API and deployment)
- [x] Link to issue (Jira or GitHub) and backlink in issue (Jira)

## Checklist (optional)

- [x] New functionality  for `task` also works for `forwarding`

PR #6395, opened 2020-05-04
Return main dossier instead of containing dossier.

mbaechtold:
On the detail view of a task (in the new GEVER UI), the  linked dossier must be the main dossier, not the parent of the task. This is mainly relevant for tasks in subdossiers.

This is a follow-up of https://github.com/4teamwork/opengever.core/pull/6387 and belongs to https://4teamwork.atlassian.net/browse/GEVER-246.

## Checklist (Must have)
- [ ] Changelog entry

### DIFF
--- a/opengever/api/tests/test_task.py
+++ b/opengever/api/tests/test_task.py
@@ -142,7 +142,7 @@ class TestTaskSerialization(IntegrationTestCase):
             browser.json)
 
     @browsing
-    def test_containing_dossier_for_task(self, browser):
+    def test_containing_dossier_for_task_within_dossier(self, browser):
         self.login(self.regular_user, browser=browser)
         browser.open(self.task, method="GET", headers=self.api_headers)
         self.maxDiff = None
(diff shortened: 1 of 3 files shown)


opengever/api/tests/test_task.py
```diff
@@ -157,6 +157,28 @@ def test_containing_dossier_for_task(self, browser):
             browser.json['containing_dossier']
         )
 
+    @browsing
+    def test_containing_dossier_for_task_within_subdossier(self, browser):
+        self.login(self.regular_user, browser=browser)
+        task_in_subdossier = create(Builder('task')
+                                    .within(self.subdossier)
+                                    .having(responsible_client='fa')
```
deiferni, line 165: having would accept `**kwargs`, but maybe you prefer it like this. just a hint, no changes necessary 😉.

mbaechtold: I know and I was hesitant. But I have switched to `**kwargs` now. You have convinced me 😉

mbaechtold:
@deif, the test case https://github.com/4teamwork/opengever.core/blob/40db37948d5b731aaa066bd39f1fb0f4d32749ad/opengever/api/tests/test_task.py#L177 has been added in https://github.com/4teamwork/opengever.core/pull/6387 where we test with tasks in inboxes.

Can inboxes be nested too? This case is not tested yet. Should we?

deiferni:
there can be multiple inboxes in an `InboxContainer`. but the inboxes themselves can't be nested. the test should be good as is 👍 ,

PR #6400, opened 2020-05-04
Add livechat to documentation.

njohner:
With this PR we integrate live chat in our online documentation. I also fixed the line endings in some files from CRLF to LF.

This can be tested directly on the online documentations, which I have pushed to from my local branch: https://docs.onegovgever.ch/

**Screenshot of our online documentation. Chat is available and posts into slack: https://4teamwork.slack.com/archives/C013Q3K9HJL/p1588603048000700**
Screenshot 2020-05-04 at 16 41 13

For: https://4teamwork.atlassian.net/browse/GEVER-274

## Checklist (Must have)

- [x] Changelog entry


docs/public/dev-manual/api/scanin.rst
```diff
@@ -53,3 +53,8 @@ Beispiel: Upload eines Dokuments in die persönliche Ablage des Benutzers hugo.b
   Content-Type: application/octet-stream
 
   [FILE_DATA]
+
+.. disqus::
```
deiferni, line 57: could we somehow put this in a general footer and make sure it is included automatically? pasting it everywhere is quite redundant.

docs/public/_static/livechat.html
```diff
@@ -0,0 +1,7 @@
+<script type="text/javascript">
```
deiferni, line 1: this seems duplicated, too. but maybe we have to live with it as the docs are separated. correct?

njohner:
I agree it would have been nicer to have two commits, but I obviously scripted adding the live chat correctly everywhere (as the path to the file is not always the same either...) and that changed the line endings as well...

docs/public/dev-manual/api/scanin.rst
lukasgraf: You can, in theory. However, that will involve messing with the alabaster theme which our  [`opengever/sphinx/gevertheme`](https://github.com/4teamwork/opengever.core/tree/master/opengever/sphinx/gevertheme) inherits from. In the end, you'd have to overwrite it in a way so that you can inject a snippet into [alabaster's `layout.html`](https://github.com/bitprophet/alabaster/blob/master/alabaster/layout.html#L87-L124) in an appropriate place.

Not impossible, but probably quite challenging to do in a way where the overwrite isn't too sensitive to upstream changes in the alabaster theme.

lukasgraf:
@njohner @deiferni ok, footerizing it actually isn't that hard. This seems to work:

```jinja
{% extends 'alabaster/layout.html' %}

{% block livechat_footer %}
        <h1>Hello World</h1>
{% endblock %}
```
(create as `opengever/sphinx/gevertheme/layout.html`, rebuild docs).

docs/public/_static/livechat.html
njohner: Yeah it's duplicated though it doesn't really have to. It's more for consistency. `disqus.js` is also duplicated. I think it's ok to keep each documentation directory separate, as it also gets built separately. Having them self-sufficient makes sense IMO.

njohner:
We could try. That would work for `disqus` I think, but not for the live chat, as the path to the file is not always the same. Maybe there is a better way to do that though.
Also for now we do not have `disqus` in the main navigation points. Putting it in the footer would add it there, not that I mind...

lukasgraf:
@njohner you would then ditch `livechat.html` and include the HTML directly in the customized `layout.html (instead of `<h1>Hello World</h1>`). The theme customization is on HTML level, not RST.

For that reason it would also not easily work for disqus actually, because that's a RST directive (which includes some dynamic magic to set a `shortname` under which discussion threads on a particular page are grouped together).

njohner:
Ok so adding the chat in the layout works fine. I updated the PR accordingly. Downside is
that the chat is everywhere, also in intern documentation. Upside is that the chat is everywhere, so we won't forget to add it anywhere in the future.

PR #6403, opened 2020-05-05
Show proposal template corresponding to the committee.

2e12:
The proposal templates weren't filtered anymore by the selected committee. This should by repaired in this PR.

In commit 31b96c4ea3b395224d1307b5c11dccfb3898898e the field 'committee' was renamed to 'committee_oguid'. This had the consequence that a select field was renamed what caused that form.widgets.proposal_template wasn't updated anymore when changing the value in the select box.

Also the vocabulary filtered not correctly also due the switch to oguid.

Jira Issue -> https://4teamwork.atlassian.net/browse/GEVER-268

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/meeting/tests/test_vocabularies.py
+++ b/opengever/meeting/tests/test_vocabularies.py
@@ -137,8 +137,8 @@ class TestProposalTemplatesForCommitteeVocabulary(IntegrationTestCase):
             [term.value for term in factory(context=self.dossier)])
 
         self.committee.allowed_proposal_templates = [IUUID(baubewilligungen)]
-        self.request.form['form.widgets.committee'] = [
-            unicode(self.committee.load_model().committee_id)]
+        self.request.form['form.widgets.committee_oguid'] = [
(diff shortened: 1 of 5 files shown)


opengever/meeting/tests/test_vocabularies.py
```diff
@@ -136,7 +136,7 @@ def test_reduce_allowed_templates_with_committee_settings(self):
              self.recurring_agenda_item_template],
             [term.value for term in factory(context=self.dossier)])
 
-        self.committee.allowed_proposal_templates = [IUUID(baubewilligungen)]
+        self.committee.allowed_proposal_templates = [Oguid(baubewilligungen)]
```
deiferni, line 139: i don't think that is correct, the vocabulary for the allowed templates still expects UIDs, see https://github.com/4teamwork/opengever.core/blob/04fcdf52cae62e3ccacfa6c0ff3b3d5a4d9c9ebf/opengever/meeting/vocabulary.py#L302

PR #6409, opened 2020-05-06
Drop sorting by sortable_author for solr and avoid handling sorting parameters as fields.

deiferni:
In https://github.com/4teamwork/opengever.core/pull/6382 and https://github.com/4teamwork/opengever.core/pull/6404 we have introduced explicit logging for potential issues in solr listing. We now have already found some cases where we need special handling/fixes. This PR addresses the following issues

- Pop sorting parameters from solr query to avoid attempting to handle them as schema fields
- No longer sort by `sortable_author` in solr as no such field exists, instead `document_author` should be used

Furthermore we should also log to sentry when we encounter unknown fields as more issues may appear.

For https://4teamwork.atlassian.net/browse/GEVER-328

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.3.0rc2 (unreleased)
 ------------------------
 
+- Drop sorting by sortable_author for solr and avoid handling sorting parameters as fields. [deiferni]
 - Add live chat to online documentation. [njohner]
 - Bump ftw.monitor and ftw.contentstats to get performance metrics. [lgraf]
 - Merge notification settings for tasks. [elioschmutz]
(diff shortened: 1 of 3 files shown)


docs/HISTORY.txt
```diff
@@ -4,6 +4,8 @@ Changelog
 2020.3.0rc2 (unreleased)
 ------------------------
 
+- Bump ftw.solr to 2.8.5 to ensure solr maintenance scripts are run as system user. [njohner]
```
mbaechtold, line 7: I suppose this is due to a rebase gone awry? Or do I need to call you N. now? 😂

deiferni: thanks, you were too fast, i didn't get to look at it myself after the rebase 😓

deiferni:
@mbaechtold i'm not sure what you mean with the strangeness of commit order 😓, please elaborate.

deiferni:
_it may have been a temporary rebase artifact, i rebased twice as many changes were going in_

njohner:
Sometimes the order of the commit is shown differently in github than in it actually is (if you look at git log in the console). This can be very confusing to review...

PR #6412, opened 2020-05-08
Watchers GET API: Also include info about referenced_users and referenced_watcher_roles.

tinagerber:
The frontend also needs the full name of a watcher and a translation for the roles.
Therefore the watchers GET API has been extended with the lists` referenced_users` and `referenced_watcher_roles`

## Checklist (Must have)

- [x] Changelog entry

## Checklist (optional)


opengever/api/watchers.py
```diff
@@ -29,14 +31,33 @@ def __call__(self, expand=False):
...
+        referenced_users = []
         watchers_and_roles = defaultdict(list)
+        portal_url = api.portal.get().absolute_url()
         for subscription in self.center.get_subscriptions(self.context):
-            watchers_and_roles[subscription.watcher.actorid].append(subscription.role)
-        return watchers_and_roles
+            userid = subscription.watcher.actorid
+            watchers_and_roles[userid].append(subscription.role)
+            roles.add(subscription.role)
+            referenced_users.append(
```
njohner, line 42: It seems to me that a userid can be in several subscriptions, and in that case it will appear several times in the `referenced_users`. That does not make sense. Instead you can simply determine the referenced users once that loop is finished:
```
referenced_users = [ 
                  {
                    '@id': "{}/@users/{}".format(portal_url, userid),
                    'id': userid,
                    'fullname': api.user.get(userid).getProperty('fullname')
                    } for user in watchers_and_roles]
```

opengever/api/tests/test_watchers.py
```diff
@@ -53,10 +86,42 @@ def test_get_watchers_for_inbox_forwarding(self, browser):
...
+                },
+                {
+                    u'@id': u'http://nohost/plone/@users/kathi.barfuss',
+                    u'fullname': u'B\xe4rfuss K\xe4thi',
+                    u'id': u'kathi.barfuss'
+                },
+                {
+                    u'@id': u'http://nohost/plone/@users/kathi.barfuss',
+                    u'fullname': u'B\xe4rfuss K\xe4thi',
+                    u'id': u'kathi.barfuss'
```
njohner, line 103: As per my comment below, `kathi.barfuss` should only appear once here

opengever/api/watchers.py
tinagerber: 🙈 Yes, it's friday.

PR #6413, opened 2020-05-08
Fix an issue where solr facet labels have not been transformed correctly

elioschmutz:
Jira: https://4teamwork.atlassian.net/browse/GEVER-290

This PR:
- Added a feature for the `@solrsearch` endpoint to provide the same facets like the `@listing` endpoint. This means, a facet can now be requested by entering the exact solr index name like `Creator` or a self defined `ListingField`-Name like `creator` (https://github.com/4teamwork/opengever.core/blob/master/opengever/api/solr_query_service.py#L200).
- Fixes an issue where facet values have not been tranformed if requesting a facet based on a `ListingField` (i.e.: `/@listing?name=documents&columns=title&facets=creator`)

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/api/tests/test_listing.py
+++ b/opengever/api/tests/test_listing.py
@@ -8,7 +8,6 @@ from opengever.base.solr import OGSolrDocument
 from opengever.testing import IntegrationTestCase
 from opengever.testing.integration_test_case import SolrIntegrationTestCase
 from plone.uuid.interfaces import IUUID
-from unittest import skip
 
 
 class TestListingEndpointWithoutSolr(IntegrationTestCase):
@@ -78,10 +77,6 @@ class TestListingEndpointWithSolr(IntegrationTestCase):
                         msg="facet=true is needed to get facet counts back")
         self.assertEqual(1, params['facet.mincount'])
         self.assertEqual(['start'], params['facet.field'])
-
-    @skip("Just a reminder to test that facets also return translated labels")
-    def test_facet_labels(self):
-        pass
 
     @browsing
     def test_excludes_searchroot(self, browser):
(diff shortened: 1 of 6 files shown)


elioschmutz:
@njohner Thanks for the review.

Unfortunately we got another issue with my implementation. There are multiple `ListingField` instances with the same index value. Using a solr facet with i.e. the `responsible` would not transform the index label value properly because the `responsible_fullname` `ListingField` would be taken by the mapping and there is no transform registered: See https://github.com/4teamwork/opengever.core/blob/master/opengever/api/solr_query_service.py#L215

In general, the `SolrQueryBaseService` implementation seems to be optimized for the `@listing` endpoint and not for the `@solrsearch` endpoint. It's not easy to adapt the solr-facet implementation to the current implementation. I think the best option is to extract the transforms in its own mapping which I've done in this PR.

So I'd be glad if you take another look at it. Thanks

elioschmutz:
@njohner 🏃 pls take another look... I think it's the right direction now 😄

opengever/api/tests/test_listing.py
```diff
@@ -305,6 +300,27 @@ def test_search_filter(self, browser):
             [self.taskdocument.absolute_url()],
             [item['@id'] for item in browser.json['items']])
 
+    @browsing
+    def test_facet_labels_are_transformed_properly(self, browser):
+        self.login(self.regular_user, browser=browser)
+
+        view = ('@listing?name=documents&columns:list=title'
+                '&facets:list=creator&facets:list=responsible')
```
njohner, line 308: So if I read that correctly, this is one solr index (responsible) and one listing field (Creator) right? If not I would use one of each type in the test

elioschmutz: No, the `responsible` is a field where the field-name and the solr-index name are equal and the `creator` is a field where the field-name is `creator` and the solr-index is `Creator`.

But I'll add another field to test the fallback-functionality with the `Creator` field 👍

elioschmutz: Hmm... ok, even this behavior is completely different from te @solrsearch endpoint.

In the @solrsearch endpoint we have a blacklist of solr-indexes, in the @listing endpoint we have a whitelist of not mapped indexes.

I'll add an additional test for the @solrsearch endpoint to test if it's possible to query a not mapped facet.

PR #6414, opened 2020-05-11
Implement @possible-watchers endpoint

elioschmutz:
Jira: https://4teamwork.atlassian.net/browse/GEVER-320

This PR implements the `@possible-watchers` endpoint.

The Endpoint returns users without a watcher-role on the current context. The results are sorted by last and first name. The current user will be returned on the first position if he is part of the result-set.

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/public/dev-manual/api/watchers.rst
+++ b/docs/public/dev-manual/api/watchers.rst
@@ -140,3 +140,82 @@ Mittels DELETE-Requests kann die Rolle ``regular_watcher`` eines Beobachters von
       HTTP/1.1 204 No content
 
 
+Liste von möglichen Beobachtern
+-------------------------------
+Der ``@possible-watchers``-Endpoint liefert eine Liste von Benutzern welche als Beobachter für den aktuellen Kontext hinzugefügt werden können.
+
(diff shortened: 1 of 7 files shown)


docs/public/dev-manual/api/watchers.rst
```diff
@@ -140,3 +140,75 @@ Mittels DELETE-Requests kann die Rolle ``regular_watcher`` eines Beobachters von
...
+            "token": "rolf.ziegler"
+          },
+          { "...": "..." },
+        ],
+        "items_total": 17
+      }
+
+Resultate filtern
+~~~~~~~~~~~~~~~~~
+Mit dem ``query``-Parameter können die Resultate entsprechend gefiltert werden:
```
mbaechtold, line 183: It would be nice to explain how the filtering must be used. How does it work? What can be filtered?

mbaechtold: Thanks for extending the documentation 👍 I'm just curious: what happens if `params=Mueller Peter` (i.e. surname AND given name is provided)?

elioschmutz: The string gets splitted and each word will be used for each attribute with wildcards.

Take a look at our base implementation: https://github.com/4teamwork/opengever.core/blob/master/opengever/base/query.py#L29

For further information see the base-class of the user source: https://github.com/4teamwork/opengever.core/blob/master/opengever/ogds/base/sources.py#L509

mbaechtold: Thank you very much 👍

PR #6416, opened 2020-05-12
Add new filename column to Favorites.

njohner:
With this PR we add the `filename` field to the `Favorite`s, so that the Gever-UI can determine which icons to show for favourite documents.

Some technical hurdles have arisen because the filename and title of documents get synced. 

## Situation prior to this PR
We currently have a bit of a mess, with title and filename getting synced in an event handler for documents (and an other one for mails), and the title of favorites also getting updated in an event handler (for all portal types). We already now have no control on the order in which these two happen. Adding another event handler for syncing the filename in the favorites would make things worse. We already had two bugs linked to these event handlers:

- revert does not notify an ObjectModifiedEvent -> title and filename do not get synced. We could register sync_title_and_filename_handler for ObjectRevertedToVersion

- Favorite title updater (update_favorites_title) is registered on ObjectModifiedEvent, i.e. same event as sync_title_and_filename_handler, so we do not have any guarantee of the order in which things happen. Consequence is that editing a document and deleting the title in the form will throw an error (update_favorites_title gets run with title=None which it can’t handle https://sentry.4teamwork.ch/sentry/onegov-gever/issues/10316)

## Chosen implementation
To get better control on the order in which things happen and to be sure that a document's title and filename are always in sync, I refactored the syncing to happen in setters of the title and filename. One complication is that when both the title and filename are modified in the same request, syncing will happen twice and care has to be taken to make sure the correct values are used.

The chosen implementation naturally fixes both bugs listed above. We could make an upgrade step to sync filenames with titles for documents that became out of sync because of the revert issue, but I'd rather avoid this.

For https://4teamwork.atlassian.net/browse/GEVER-138

## Checklist (Must have)

- [x] Changelog entry

## Checklist (optional)
- [x] New functionality  for `document` also works for `mail`
- DB-Schema migration


opengever/base/tests/test_filename_normalizer.py
```diff
@@ -19,6 +20,22 @@ def test_maximal_filename_length(self):
         filename = u''.join('a' for i in range(2*self.max_length))
         self.assertEqual(self.max_length, len(self.normalizer.normalize_name(filename)))
 
+    def test_filename_in_favorites_can_handle_long_filename(self):
+        filename = u''.join('a' for i in range(2*self.max_length))
+        self.assertTrue(
+            len(self.normalizer.normalize_name(filename)) < FILENAME_LENGTH)
+
+    def test_filename_in_favorites_can_handle_long_filename_with_extension(self):
+        filename = u''.join('a' for i in range(2*self.max_length)) + ".jpeg"
```
deiferni, line 29: my linter wants this to be `2 * self.max_length`

opengever/base/tests/test_filename_normalizer.py
```diff
@@ -19,6 +20,22 @@ def test_maximal_filename_length(self):
         filename = u''.join('a' for i in range(2*self.max_length))
         self.assertEqual(self.max_length, len(self.normalizer.normalize_name(filename)))
 
+    def test_filename_in_favorites_can_handle_long_filename(self):
```
deiferni, line 23: thanks for improving the tests here. i think we should also add a test for a file with dots in the filename.

opengever/base/tests/test_filename_normalizer.py
```diff
@@ -104,13 +121,13 @@ def test_unmapped_extensions_remain_unchanged(self):
             self.assertEqual(extension,
                              self.normalizer.normalize_extension(extension))
 
-    def test_split_filename_extension(self):
+    def test_split_filename_extension_works_for_length_up_to_4_characters(self):
         self.assertEqual((u'filename', u'txt'),
                          self.normalizer.split_filename_extension(u'filename.txt'))
         self.assertEqual((u'filename', u'jpeg'),
                          self.normalizer.split_filename_extension(u'filename.jpeg'))
-        self.assertEqual((u'my.filename', ''),
```
deiferni, line 112: what happens if there are more than 4 characters in the extension? shouldn't we document the behavior in a test 🤔 ?

opengever/document/tests/test_checkout.py
```diff
@@ -307,6 +307,18 @@ def test_reverting_to_older_version_does_not_revert_metadata(self):
...
+        self.assertEqual(u'Testdokum\xe4nt', self.document.title)
+        self.assertEqual(u'Testdokumaent.txt', self.document.get_filename())
+
+        self.document.title = "New title"
+        create_document_version(self.document, 3)
+        self.assertEqual(u'New title', self.document.title)
+        self.assertEqual(u'New title.txt', self.document.get_filename())
+
+        self.manager.revert_to_version(0)
+        self.assertEqual("New title.docx", self.document.get_filename())
```
deiferni, line 318: i'm confused by this test. i would have expected a revert back to `Testdokumaent.txt`. but now we still have `New title` but the file extension has changed. can you help me understand what is going on here?

opengever/document/document.py
```diff
@@ -245,6 +251,7 @@ def title(self, value):
         if self.title == value:
             return
         self.__dict__['title'] = value
+        self._v_title = value
```
deiferni, line 256: can you add a test for this case? or is there already one and that's why you've seen this problem 😄?

opengever/base/tests/test_favorite.py
```diff
@@ -285,6 +286,85 @@ def test_titles_of_favorites_get_updated(self, browser):
...
+
+        fav2 = create(Builder('favorite')
+                      .for_object(self.document)
+                      .for_user(self.regular_user))
+
+        self.assertEquals('GEVER Weeklies', fav1.title)
+        self.assertEquals(u'Vertr\xe4gsentwurf', fav2.title)
+
+        browser.open(self.document, view='edit')
+        browser.fill({u'Title': u''})
```
deiferni, line 307: hmm maybe this test is a bit more readable when you explicitly change the title here to something else and then use that newly entered value in the assertion? i was a bit confused by the indirection.

deiferni: ah you wanna change the filename of the file. hmm. ok.

opengever/document/document.py
```diff
@@ -235,12 +236,13 @@ def sync_title_and_filename(self):
...
-            self.__dict__["file"].filename = normalizer.normalize(basename, extension=ext)
-        elif title:
-            # use the title as filename
-            self.__dict__["title"] = title
-            self.__dict__["file"].filename = normalizer.normalize(title, extension=ext)
+            title = basename
+
+        self.__dict__["title"] = title
+        new_filename = normalizer.normalize(title, extension=ext)
+        if not self.get_filename() == new_filename:
```
deiferni, line 243: `not blah == foo` -> `blah != foo`

opengever/base/tests/test_filename_normalizer.py
njohner: It's documented just below with `my.jpegs`, 5 characters in the extension. Extension is empty. I just changed it to use 5 characters which is the lower limit for the behaviour, rather than the 8 characters that we were testing here

opengever/base/tests/test_filename_normalizer.py
njohner: oh my linter stopped working, sorry

njohner: Hum, my linter doesn't care for some reason. But I'll update this

opengever/document/tests/test_checkout.py
njohner: No it should not revert to `Testdokumaent.txt`. Title and filename should always be in sync and when there is a title, it is used to determine the filename. So in this case, reverting to an older version should not change the title (we never revert metadata), so the filename should remain the same as the title. Content and extension are adapted to the reverted file though.

opengever/document/document.py
njohner: No I've seen the problem because I've manually tested for it (I saw that one coming).
I've added an api test for that, as we need to change both the title and the filename in the same request, and the file cannot be easily changed in the edit form using the testbrowser.

opengever/base/tests/test_favorite.py
njohner: I actually want to set the title to empty string, then it gets synced back to the filename, so to its prior value

opengever/base/tests/test_filename_normalizer.py
njohner: done

opengever/document/tests/test_checkout.py
njohner: I added a comment in the test.

PR #6417, opened 2020-05-12
Fix activity bug when creating tasks with tasktemplates.

tinagerber:
If new tasks were created with the task templates and the standard process was not triggered, no watchers were created in the database for the main task. 

The `IDuringTaskTemplateFolderTriggering` inteface prevents a `TaskAddedActivity` from being triggered when creating tasks from task templates. Therefore no watchers are created.

Now a `TaskAddedActivity` is also triggered for the main task, so the watchers are also created.

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/tasktemplates/tests/test_activities.py
+++ b/opengever/tasktemplates/tests/test_activities.py
@@ -3,7 +3,6 @@ from ftw.builder import create
 from ftw.testbrowser import browsing
 from opengever.activity.model import Activity
 from opengever.testing import IntegrationTestCase
-from plone import api
 
 
 class TestTaskTemplateActivites(IntegrationTestCase):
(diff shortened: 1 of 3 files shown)


opengever/tasktemplates/tests/test_activities.py
```diff
@@ -72,6 +71,5 @@ def test_does_not_record_activity_for_planed_task(self, browser):
         main_task = self.dossier.objectValues()[-1]
         seq1, seq2 = main_task.objectValues()
 
-        self.assertItemsEqual([main_task, seq1],
-                              [activity.resource.oguid.resolve_object()
-                               for activity in Activity.query.all()])
+        self.assertNotIn(seq2, [activity.resource.oguid.resolve_object()
```
deiferni, line 74: you have dropped the "positive" test in favour of a "negative" one. those are a bit brittle, there may be different reasons why `q2` is not in the recorded activity. Could you rewrite this using a "positive" assertion and assert on the activities that must be generated, or at least re-add and rewrite the test that explicitly tests for a list of generated acitivites?

tinagerber:
@deiferni I changed the test, can you look at it again?

PR #6420, opened 2020-05-12
Add minimal support for creating .oggbundle from repository excel files.

deiferni:
This PR adds minimal support for creating an `*.oggbundle` from a repository file that we use to initialize content when policies are setup initially. It does not yet add support for all attributes from the excel but just what i needed for a migration for a customer in  https://4teamwork.atlassian.net/browse/GEVER-219.

Corresponding Pull-Request with content generated from the new bundle factory can be found at https://github.com/4teamwork/opengever.kktsg/pull/17.

In this PR we:
- refactor the existing filesystem based bundle import a bit so that other types of nodes can be plugged in
- add support for creating an excel based bundle, consider the following points:
  - support is added to the `create-bundle` command, in addition to the path parameter to a folder you can also pass in the path to an excel file
  - creating a full repository bundle from the excel for basic attributes has been implemented
  - local roles are currently not set however
  - imports for a partial repository are not supported yet
  - not all command line arguments make sense when an excel is specified, in that case the factory aborts
  - only `dotted` reference numbers can be used, we attempt to detect `grouped_by_three` and complain
- the loader is not changed, that way currently released `opengever.core` versions can import bundles created by the new factory

## Checklist (Must have)

- [x] Changelog entry


opengever/bundle/xlsx.py
```diff
@@ -95,3 +95,47 @@ def is_repo(self, repo_depth):
     @property
     def title(self):
         return self.item['effective_title']
+
+    @property
+    def title_fr(self):
+        return self.item['effective_title_fr'] or None
```
njohner, line 101: Is that because you get empty string by default and you want to return `None` instead?

opengever/bundle/factory.py
```diff
@@ -174,8 +174,8 @@ def __call__(self, node):
                 return OGGBundleRepoRoot(node, self.user_group)
             else:
                 return OGGBundleDossier(node, self.responsible, self.import_reference)
-            return OGGBundleRepoFolder(node)
         elif node.is_repo(self.repo_depth) and not self.content_only:
+            return OGGBundleRepoFolder(node)
```
njohner, line 178: Ok I guess that belongs in another commit, but not tragic

opengever/bundle/xlsx.py
```diff
@@ -95,3 +95,47 @@ def is_repo(self, repo_depth):
...
+
+    @property
+    def valid_from(self):
+        return self.item['valid_from'] or None
```
njohner, line 105: For all these I'm a bit lost why you do it this way... shouldn't `get` be enough here?

opengever/bundle/xlsx.py
```diff
@@ -0,0 +1,97 @@
...
+        self.parent_reference_number = self.parse_parent_reference_number(refnum_to_guid)
+        self.guid = self.make_guid()
+        self.parent_guid = self.parse_parent_guid(refnum_to_guid)
+
+        refnum_to_guid[self.reference_number] = self.guid
+
+    def parse_reference_number(self):
+        reference_number = self.item.get('reference_number', None)
+        if reference_number:
+            level = len(reference_number.split('.'))
```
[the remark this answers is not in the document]
deiferni: i think the excel is always in the dotted format, we create correct numbers during bundle loading only.

opengever/bundle/xlsx.py
deiferni: yeah, exactly.

opengever/bundle/xlsx.py
deiferni: it can be empty string or `None`. I want to make sure it is `None`. Didn't see a more "elegant" way of doing it.

opengever/bundle/xlsx.py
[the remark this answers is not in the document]
deiferni: ah no you're right, it isn't, example in https://github.com/4teamwork/opengever.duerrenaesch/blob/45c4a3967371ae3fa87acbd4d8bb2da9f21525d6/opengever/duerrenaesch/profiles/default_content/opengever_repositories/ordnungssystem.xlsx. the issue is i don't have the correct plone site when generating the bundle. you can generate the bundle on another plone site than where it is loaded into. i will add very dumb heuristics that abort when it detects a `grouped_by_three` repository. we can implement command-line support later.

opengever/bundle/xlsx.py
```diff
@@ -0,0 +1,97 @@
...
+
+        return level, reference_number, reference_number_prefix
+
+    def parse_parent_reference_number(self, refnum_to_guid):
+        # there is only a parent reference number for sub-repofolders, and None
+        # for the first repo-folders and the repo-root.
+        if self.level <= 1:
+            return None
+
+        parent = self.reference_number[:self.reference_number.rfind('.')]
```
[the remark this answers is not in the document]
deiferni: yeah, more elegant your way. amended.

opengever/bundle/xlsx.py
```diff
@@ -0,0 +1,97 @@
...
+        suffix = uuid4().hex[:8]
+        guid = u'{}-{}-{}'.format(
+            self.reference_number or 'ROOT',
+            self.title,
+            suffix
+        )
+        # Transliterate Umlauts to ae, oe, ue, after that safe-encode to ASCII
+        return normalizer.normalize(guid).encode('ascii', errors='replace')
+
+    def parse_parent_guid(self, refnum_to_guid):
```
lukasgraf, line 80: This doesn't really parse the parent guid, it looks it up. `lookup_parent_guid()`?

opengever/bundle/tests/test_xlsx_walker.py
```diff
@@ -0,0 +1,210 @@
...
+            ], items)
+
+    def test_xlsx_walker_iter_and_note_initialization(self):
+        walker = XLSXWalker(assets.get_path('basic_repository.xlsx'))
+        items = list(walker)
+
+        self.assertEqual(8, len(items))
+
+        root = items[0]
+        self.assertIsNone(root.creation_date)
```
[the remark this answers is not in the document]
deiferni: added unit test for all nodes.

opengever/bundle/xlsx.py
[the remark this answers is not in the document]
deiferni: i have added a simple check for grouped_by_three. not elegant but should do the job.

opengever/bundle/factory.py
deiferni: i couldn't easily move it. sorry 😓.

deiferni:
ok i have amended, new unit-tests in the following commits:
- https://github.com/4teamwork/opengever.core/pull/6420/commits/1db45058aee6e4e75c54db9d37aef42d101feaa9
- https://github.com/4teamwork/opengever.core/pull/6420/commits/00836284154a5de4c6e2ed1d97ea11917ac58426

also amended the last two commits that add better exceptions, and complain for grouped by three.

deiferni:
@lukasgraf i'm pulling you in for a quick re-review, see https://github.com/4teamwork/opengever.core/pull/6420#issuecomment-628585320 for affected commits.

deiferni:
as just discussed @lukasgraf will take over and merge this tomorrow. thanks.

PR #6421, opened 2020-05-13
Skip unknown attributes in POST @invitation endpoint.

elioschmutz:
Jira: https://4teamwork.atlassian.net/browse/GEVER-357

The `@invitation` endpoint is currently hard-validating the json-body and will raise an error if there is an unknown attribute. That's not the same behavior like the default plone restapi for adding resources.

This PR refactors the endpoint to extract all relevant attributes from the body.

**Side-Info**
The frontend provides a generic add form which uses a predefined schema for a type. Since an invitation is a kind of type in the frontend, it also sends the `@type` property to the server for adding a new invitation.

Bildschirmfoto 2020-05-13 um 10 16 37

This will raise an error on the backend because `@type` is not part of the `IWorkspaceInvitationSchema`:

`BadRequest: [(u'@type', UnknownField(u'@type'))]`
https://sentry.4teamwork.ch/sentry/onegov-gever/issues/67036/

Of course, we could fix this in the frontend by not sending the `@type` property. But the API should be robust enough to handle this kind of request.

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/api/tests/test_validation.py
+++ b/opengever/api/tests/test_validation.py
@@ -1,0 +1,26 @@
+from opengever.api.exceptions import UnknownField
+from opengever.api.validation import get_validation_errors
+from opengever.testing import FunctionalTestCase
+from plone.supermodel.model import Schema
+
+
+class TestSchema(Schema):
+    """
+    """
+
+
+class TestSchemaValidation(FunctionalTestCase):
+
+    def test_allow_unknown_fields(self):
+        data = {'lorem_ipsum_field': u'james.bond'}
+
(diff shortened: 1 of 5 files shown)


elioschmutz:
@njohner just updated the PR. Could you take another look?

opengever/api/tests/test_validation.py
```diff
@@ -0,0 +1,25 @@
...
+from opengever.testing import FunctionalTestCase
+from plone.supermodel.model import Schema
+
+
+class TestSchema(Schema):
+    """
+    """
+
+
+class TestSchemaValidation(FunctionalTestCase):
```
njohner, line 12: missing new line

PR #6424, opened 2020-05-13
Add watchers, resources and subscriptions to tasks and forwarding in fixtures

tinagerber:
Since the `activity` feature is not enabled when creating tasks (and forwardings) in the `OpengeverContentFixture`, no subscriptions, resources and watchers are created in the database for the tasks.

Currently it is not possible to enable the feature, so the required content is now created manually for each task.

Now subscriptions, resources and watchers for tasks no longer need to be created first in each test

## Checklist (Must have)

- [x] Changelog entry

## Checklist (optional)

### DIFF
--- a/opengever/testing/fixtures.py
+++ b/opengever/testing/fixtures.py
@@ -263,6 +263,7 @@ class OpengeverContentFixture(object):
             'meeting_user',
             u'Herbert',
             u'J\xe4ger',
+            email='herbert@jager.com',
             user_settings={'_seen_tours': '["*"]'},
             )
 
--- a/opengever/testing/fixtures.py
+++ b/opengever/testing/fixtures.py
@@ -9,21 +9,25 @@ from datetime import timedelta
 from ftw.builder import Builder
 from ftw.builder import create
 from ftw.builder import ticking_creator
+from ftw.builder.builder import original_create
 from ftw.bumblebee.interfaces import IBumblebeeUserSaltStore
 from ftw.bumblebee.tests.helpers import asset as bumblebee_asset
 from ftw.testing import freeze
 from ftw.testing import staticuid
 from ftw.tokenauth.pas.storage import CredentialStorage
+from opengever.activity.model import Watcher
+from opengever.activity.roles import TASK_ISSUER_ROLE
+from opengever.activity.roles import TASK_RESPONSIBLE_ROLE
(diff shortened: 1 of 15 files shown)


opengever/testing/fixtures.py
```diff
@@ -70,6 +74,23 @@ def configure_jwt_plugin(self):
         jwt_plugin.use_keyring = False
         jwt_plugin._secret = JWT_SECRET
 
+    def get_watcher(self, actorid):
+        watcher = Watcher.query.get_by_actorid(actorid)
+        if watcher:
+            return watcher
+        return original_create(Builder('watcher').having(actorid=actorid))
```
deiferni, line 81: i'd add a comment why we use `original_create` here

opengever/testing/fixtures.py
```diff
@@ -70,6 +74,23 @@ def configure_jwt_plugin(self):
...
+
+    def create_subscriptions(self, obj):
+        oguid = Oguid.for_object(obj)
+        resource = original_create(Builder('resource').oguid(oguid.id))
+        responsible_watcher = self.get_watcher(obj.responsible)
+        issuer_watcher = self.get_watcher(obj.issuer)
+        original_create(Builder('subscription').having(resource=resource,
```
deiferni, line 88: and here

opengever/testing/fixtures.py
```diff
@@ -70,6 +74,23 @@ def configure_jwt_plugin(self):
         jwt_plugin.use_keyring = False
         jwt_plugin._secret = JWT_SECRET
 
+    def get_watcher(self, actorid):
+        watcher = Watcher.query.get_by_actorid(actorid)
+        if watcher:
+            return watcher
+        return original_create(Builder('watcher').having(actorid=actorid))
+
+    def create_subscriptions(self, obj):
```
deiferni, line 83: i'd rename this to `create_task_subscriptions` as it accesses task attributes and can only work for tasks.

opengever/testing/fixtures.py
```diff
@@ -70,6 +74,23 @@ def configure_jwt_plugin(self):
         jwt_plugin.use_keyring = False
         jwt_plugin._secret = JWT_SECRET
 
+    def get_watcher(self, actorid):
```
deiferni, line 77: `get_or_create_watcher` would indicate that we also might create records. when i read `get` i usually assume an error or a return value of `None` for nonexisting actorids. what do you think?

PR #6425, opened 2020-05-14
Improve performance while determining leaf node state.

mbaechtold:
The previous implementation was not performing well for repository folders containing lots of folders, because `contentItems` is not lazy. It should only be used for small number of items (see https://docs.plone.org/develop/plone/content/listing.html).

Furthermore, the exit condition has been improved due to the fact that a a repository folder cannot contain other repository folders and items of other content types at the same time.

Belongs to the Jira issue [GEVER-375](https://4teamwork.atlassian.net/browse/GEVER-375) (which contains the proposed solution and a flame graph).

## Checklist

- [x] Changelog entry

### DIFF
--- a/opengever/repository/repositoryfolder.py
+++ b/opengever/repository/repositoryfolder.py
@@ -149,9 +149,13 @@ class RepositoryFolder(content.Container, RepositoryMixin):
     def is_leaf_node(self):
         """ Checks if the current repository folder is a leaf-node.
         """
-        for id, obj in self.contentItems():
+        for obj in self.objectValues():
+            # A repository folder cannot contain other repository folders and
+            # items of other content types at the same time.
             if obj.portal_type == self.portal_type:
                 return False
+            else:
+                return True
         return True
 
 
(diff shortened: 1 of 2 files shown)


opengever/repository/repositoryfolder.py
```diff
@@ -149,9 +149,13 @@ def _prefix_with_reference_number(self, title):
     def is_leaf_node(self):
         """ Checks if the current repository folder is a leaf-node.
         """
-        for id, obj in self.contentItems():
+        for obj in self.objectValues():
```
mbaechtold, line 152: Any idea how the performance improvement could be tested❓

buchi: Testing performance at this level is difficult and doesn't make sense IMO. 
I'm not sure yet how we want to track this.

mbaechtold: That's what I thought. So is it safe to merge this change?

PR #6426, opened 2020-05-14
Improve performance while determining repositoryfolder emptiness.

mbaechtold:
`listFolderContents` honors the currently logged-in user roles, so the computed value might be misleading. Furthermore it is slow for large folders (see https://docs.plone.org/develop/plone/content/listing.html#getting-folder-objects-filtered).

Belongs to [GEVER-373](https://4teamwork.atlassian.net/browse/GEVER-373) (which contains the proposed solution and a flame graph).

## Checklist

- [x] Changelog entry

### DIFF
--- a/opengever/repository/deleter.py
+++ b/opengever/repository/deleter.py
@@ -27,4 +27,4 @@ class RepositoryDeleter(object):
         return self.is_repository_empty()
 
     def is_repository_empty(self):
-        return len(self.repository.listFolderContents()) == 0
+        return self.repository.objectCount() == 0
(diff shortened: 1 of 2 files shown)


opengever/repository/deleter.py
```diff
@@ -27,4 +27,4 @@ def is_deletion_allowed(self):
         return self.is_repository_empty()
 
     def is_repository_empty(self):
-        return len(self.repository.listFolderContents()) == 0
+        return self.repository.objectCount() == 0
```
mbaechtold, line 30: Any idea how the performance improvement could be tested❓

buchi: Testing performance at this level is difficult and doesn't make sense IMO.
I'm not sure yet how we want to track this.

PR #6429, opened 2020-05-15
Assign permission to role "ServiceKeyUser".

mbaechtold:
Among various possible solutions, this was deemed the most appropriate one (as discussed extensively between Lukas, Elio and Martin). The creation of an additional role was considered "overkill". The chosen role seems the most appropriate at the moment.

Belongs to the Jira issue [GEVER-339](https://4teamwork.atlassian.net/browse/GEVER-339)

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/core/profiles/default/rolemap.xml
+++ b/opengever/core/profiles/default/rolemap.xml
@@ -54,6 +54,10 @@
       <role name="Manager" />
     </permission>
 
+    <permission name="opengever.api: View AllowedRolesAndPrincipals" acquire="False">
+      <role name="ServiceKeyUser" />
+    </permission>
+
     <!-- CMFEDITIONS -->
     <permission name="CMFEditions: Save new version" acquire="True">
       <role name="Administrator" />
--- a/opengever/core/upgrades/20200515162243_assign_permission_to_service_key_user_role/upgrade.py
+++ b/opengever/core/upgrades/20200515162243_assign_permission_to_service_key_user_role/upgrade.py
@@ -1,0 +1,9 @@
+from ftw.upgrade import UpgradeStep
+
+
+class AssignPermissionToServiceKeyUserRole(UpgradeStep):
+    """Assign permission to ServiceKeyUser role.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.3.0rc3 (unreleased)
 ------------------------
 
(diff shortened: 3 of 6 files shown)


opengever/core/upgrades/20200515162243_assign_permission_to_service_key_user_role/upgrade.py
```diff
@@ -0,0 +1,9 @@
+from ftw.upgrade import UpgradeStep
+
+
+class AssignPermissionToServiceKeyUserRole(UpgradeStep):
+    """Assign permission to ServiceKeyUser role.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
```
mbaechtold, line 9: Could this upgrade step be made deferrable?

opengever/core/profiles/default/rolemap.xml
```diff
@@ -54,7 +54,7 @@
       <role name="Manager" />
     </permission>
 
-    <permission name="opengever.api: View AllowedRolesAndPrincipals" acquire="False">
+    <permission name="opengever.api: View AllowedRolesAndPrincipals" acquire="True">
```
mbaechtold, line 57: I have discussed this with @lukasgraf prior to the implementation. But then an existing test (`test_get_returns_allowed_roles_and_principals`) failed in commit 45d03e099784891cf208e005f4456f4fc864a3b0 

So I'm not sure what to do. Should we really set `acquire="True"`?

opengever/core/upgrades/20200515162243_assign_permission_to_service_key_user_role/upgrade.py
njohner: I guess it could, but I think I wouldn't. This will not be a time consuming upgrade, so there is no gain in making it deferrable.

opengever/core/profiles/default/rolemap.xml
njohner: It will pass if you login with `self.service_user`. That's the user with the `ServiceKeyUser` role. `Manager` does not have the permission. Then you can set `acquire="False"`

opengever/core/upgrades/20200515162243_assign_permission_to_service_key_user_role/upgrade.py
mbaechtold: Thanks

opengever/core/profiles/default/rolemap.xml
mbaechtold: I was too confused... Thanks, I have changed the user.

PR #6430, opened 2020-05-18
Fix upgrade populating filename column in favorites table.

njohner:
The upgrade step to populate the favorites table was broken because filename is `Missing.Value` for non document type brains, which will make the update of the table fail when trying to set the filename value. I probably only had favorited documents locally when I executed the upgrade. 

For https://4teamwork.atlassian.net/browse/GEVER-138

## Checklist (Must have)

- [x] Changelog entry: No changelog, same release as the PR that introduced this upgrade

## Checklist (optional)

- [x] New functionality  for `document` also works for `mail`
- DB-Schema migration

### DIFF
--- a/opengever/core/upgrades/20200508163417_populate_filename_column_in_favorites/upgrade.py
+++ b/opengever/core/upgrades/20200508163417_populate_filename_column_in_favorites/upgrade.py
@@ -30,21 +30,14 @@ class PopulateFilenameColumnInFavorites(SQLUpgradeStep):
             .where(favorites_table.c.admin_unit_id == current_admin_unit_id)
             .distinct()
         ).fetchall()
-        plone_uids = [row[0] for row in rows]
+        favorite_uids = [row[0] for row in rows]
 
-        for plone_uid in plone_uids:
-            brains = self.catalog_unrestricted_search({'UID': plone_uid})
-            if len(brains) != 1:
-                LOG.error(
-                    'Could not find a unique brain for the UID={}'.format(plone_uid)
-                )
-                continue
-            brain = brains[0]
-
-            filename = brain.filename
+        query = {"object_provides": "opengever.document.behaviors.IBaseDocument",
+                 "UID": favorite_uids}
+        for brain in self.catalog_unrestricted_search(query=query):
+            filename = brain.filename or u""
             self.execute(
                 favorites_table.update()
                 .values(filename=filename)
-                .where(favorites_table.c.plone_uid == plone_uid)
-                .where(favorites_table.c.admin_unit_id == current_admin_unit_id)
-            )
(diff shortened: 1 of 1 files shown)


opengever/core/upgrades/20200508163417_populate_filename_column_in_favorites/upgrade.py
```diff
@@ -30,9 +30,14 @@ def migrate(self):
             .where(favorites_table.c.admin_unit_id == current_admin_unit_id)
             .distinct()
         ).fetchall()
-        plone_uids = [row[0] for row in rows]
+        favorite_uids = set(row[0] for row in rows)
 
-        for plone_uid in plone_uids:
+        documents = self.catalog_unrestricted_search(
```
lukasgraf, line 35: Ugh. This queries the catalog for _all_ documents, and constructs brains for them. It should be possible to query a list (set?) of UIDs, by doing something like
```python
catalog(object_provides='opengever.document.behaviors.IBaseDocument', UID=favorite_uids)
```

opengever/core/upgrades/20200508163417_populate_filename_column_in_favorites/upgrade.py
```diff
@@ -46,5 +51,4 @@ def migrate(self):
                 favorites_table.update()
                 .values(filename=filename)
```
lukasgraf, line 52: I think this still needs to account for the fact that documents without files will have a `brain.filename == Missing.Value`.  =>
```python
filename = brain.filename or ''
```?

njohner: The filename indexer defaults to empty string: https://github.com/4teamwork/opengever.core/blob/master/opengever/document/indexers.py#L214-L219
So that should not be necessary. But for an upgrade step, maybe better safe than sorry... I'll update just in case

opengever/core/upgrades/20200508163417_populate_filename_column_in_favorites/upgrade.py
njohner: Yes True, sorry. I somehow thought that passing a very large list of UIDs in the query might not be optimal, but I think I was just mixing up with `path` queries where we used to make one query for each path in the background...
So it's much better as you propose.

lukasgraf: Yes, this is certainly also an area of concern. You're right, we did have exactly these issues with `path` - however, that used to perform badly because of an implementationd detail of how the `ExtendedPathIndex` works, AFAIR. An UIDIndex type should be much more capable of handling set operations, and I would expect the number of favorites to be somewhat manageable.

PR #6432, opened 2020-05-18
Remove assigned_client index and metadata

buchi:


opengever/core/upgrades/20200518084137_cleanup_catalog_indexes_and_metadata/upgrade.py
```diff
@@ -0,0 +1,59 @@
...
+        for del_index in del_indexes:
+            del names[del_index]
+
+        # Rebuild the schema
+        schema = {}
+        for i, name in enumerate(names):
+            schema[name] = i
+
+        self.schema = schema
+        self.names = tuple(names)
```
mbaechtold, line 48: What does this do? It seems not be used anywhere.

opengever/core/upgrades/20200518084137_cleanup_catalog_indexes_and_metadata/upgrade.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class CleanupCatalogIndexesAndMetadata(UpgradeStep):
+    """Cleanup catalog indexes and metadata.
+    """
+
+    def __call__(self):
+        self.remove_metadata()
+
+    def remove_indexes(self):
```
mbaechtold, line 21: This seems not to be called.

opengever/core/upgrades/20200518084137_cleanup_catalog_indexes_and_metadata/upgrade.py
buchi: oops. should be `catalog._catalog` not `self`.

opengever/core/upgrades/20200518084137_cleanup_catalog_indexes_and_metadata/upgrade.py
buchi: fixed.

buchi:
Indeed. `Task` does not have an `assigned_client` field, so passing it to `Builder` shouldn't have an effect.

PR #6433, opened 2020-05-18
Remove catalog indexes, metadata for new sites.

mbaechtold:


mbaechtold:
This will be included in https://github.com/4teamwork/opengever.core/pull/6436/, since we do not need to remove the index `assigned_client` in the profile hooks because the index will not be created anymore at all (it's an index of GEVER, so its in our control).

PR #6436, opened 2020-05-18
Remove catalog indexes and metadata of "plone.app.discussion".

mbaechtold:
This also introduces a profile hook which removes unneeded catalog indexes and metadata when a fresh GEVER is installed.


mbaechtold:
Thanks for the hint. I have removed them.

PR #6441, opened 2020-05-19
Remove catalog metadata "getRemoteUrl".

mbaechtold:


opengever/base/browser/navigation.py
```diff
@@ -146,3 +152,53 @@ def minute_of(stamp):
             previous = brain.modified
 
         return str(newest.millis())
+
+
```
buchi, line 156: We should add a comment why this customization was done. E.g.
```
# Plone's default implementation requires the `getRemoteUrl` metadata from the
# catalog. As we removed this metadata, we need a custom implementation that
# does not depend on it and thus doesn't support external links.
```

mbaechtold: Good idea. I have added a docstring to the class.

PR #6452, opened 2020-05-19
Notify added watchers.

tinagerber:
If a user is added as a watcher, he should be notified about it.

This is achieved as follows:

- If a watcher is added to a resource with the PloneNotificationCenter, a WatcherAddedEvent is triggered. 
- A handler of the event creates either a TaskWatcherAddedActivity or a ForwardingWatcherAddedActivity, depending on whether the watcher observes a task or a forwarding. 
- The WatcherAddedActivity then creates an activity and a notifications entry in the database.

The new notification setting can be found in the general tab.
Bildschirmfoto 2020-05-19 um 14 13 33

Jira: https://4teamwork.atlassian.net/browse/GEVER-306

## Checklist (Must have)

- [x] Changelog entry

## Checklist (optional)

- New translations


opengever/activity/center.py
```diff
@@ -222,10 +224,12 @@ def _get_oguid_for(self, item):
...
         return item
 
-    def add_watcher_to_resource(self, obj, actorid, role):
+    def add_watcher_to_resource(self, obj, actorid, role=WATCHER_ROLE):
         oguid = self._get_oguid_for(obj)
-        return super(PloneNotificationCenter, self).add_watcher_to_resource(
+        super(PloneNotificationCenter, self).add_watcher_to_resource(
             oguid, actorid, role)
+        if role == WATCHER_ROLE:
+            notify(WatcherAddedEvent(oguid, actorid))
```
njohner, line 232: Is there a reason that we go to the trouble to fire an event and add a handler for this, instead of directly notifying the added watcher here?

opengever/core/upgrades/20200506092908_add_added_as_watcher_notification_default/upgrade.py
```diff
@@ -0,0 +1,36 @@
...
+from opengever.core.upgrade import SchemaMigration
+from sqlalchemy.schema import Sequence
+from sqlalchemy.sql.expression import column
+from sqlalchemy.sql.expression import table
+import json
+
+SETTING = {
+     'kind': 'added-as-watcher',
+     'badge_notification_roles': [WATCHER_ROLE],
+     'mail_notification_roles': []
```
lukasgraf, line 11: Aren't digest notification roles missing here (and below)?

opengever/task/activities.py
```diff
@@ -130,6 +130,61 @@ def before_recording(self):
         self.center.add_task_issuer(self.context, self.context.issuer)
 
 
+class TaskWatcherAddedActivity(BaseTaskActivity):
+    """Activity representation for commenting a task.
```
lukasgraf, line 134: for commenting a task -> for a watching being added to a task.

opengever/activity/center.py
tinagerber: There was a problem with circular imports.

opengever/task/activities.py
```diff
@@ -130,6 +130,61 @@ def before_recording(self):
...
+    @property
+    def description(self):
+        return {}
+
+    @property
+    def label(self):
+        msg = _('label_task_watcher_added', u'Added as watcher of the task')
+        return self.translate_to_all_languages(msg)
+
+    def add_activity(self):
```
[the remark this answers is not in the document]
lukasgraf: I agree that this should not be handled in `add_activity`. Instead, the right place would be to override [`create_notifications `](https://github.com/4teamwork/opengever.core/blob/f860c6d79fda56777f576b087444f3dcd151fd34/opengever/activity/model/activity.py#L39-L56) of the base `Activity` class.

It's not so much that notifications shouldn't be _dispatched_ to users other than the newly added watcher - they shouldn't even be _created_ for anyone else. And that needs to be customized by overriding  `create_notifications` for this case I think. The dispatching should then all fall into place I believe.

So, similarly like the original `create_notifications` handles the special case of not creating a notification for the actor herself, the overridden `create_notifications` on this class would only create one for `self.watcherid`.

_(I think `add_activity` then wouldn't even need to be overridden anymore)_

opengever/activity/center.py
lukasgraf: Mhh, that's cheating though, just adding indirection by introducing new events to "solve" this doesn't sound like a good idea :D
With circular imports, the usual strategy to solve them usually is:
- Find out what the contested resource is (what are several modules fighting about, trying to import it)?
- Extract that thing (class, function, ...) out into it's own module (`.py`), and change the two fighting importeurs to import from there

elioschmutz: ℹ️ that was my idea! I'm currently on a phonecall... will answer later

opengever/task/activities.py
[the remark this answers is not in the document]
njohner: Yes I wanted to write `create_notifications ` in my comment above, not `dispatch_notifications`, I've corrected accordingly. But this cannot be overriden as is, because it's in the `Activity` model, which is the same for all activities. That's why I proposed to add a new parameter to `create_notifications` in `Activity`, which seems to be the easiest path.

opengever/core/upgrades/20200506092908_add_added_as_watcher_notification_default/upgrade.py
tinagerber: No, strangely enough `digest_notification_roles` are set to NULL by default in the database.

opengever/task/activities.py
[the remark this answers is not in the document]
lukasgraf: @njohner ah, I see. My mistake. So we had the same idea, except you thought it through 😉

opengever/activity/center.py
elioschmutz: Ok... yes we have big troubles with circular imports. So I suggested to use the event pattern.

The problem here is that an `Activity` object (not the SQL object) has a strong relation to the notification center. If you want to dispatch an activity, you create an instance of an activity and you call `record` on the object. The activity instance creates it's own instance of the notification center and will call `add_activity` on the notification center.

Now we want to record a new activity within the notification center itself. Means, the activity needs the notification center which will need the activity.

The first implementation of @tinagerber was to use a local import in the `add_watcher_to_resource` to fix the circular import. For me, this was not an option, because it tells us, that something is wrong.

Using the event pattern seemed to solve the problem and it felt right if considering the current activity/notification implementation. 

If not, then I see the following options:

1. Handling it in the API endpoint instead in the notification-center (preffered)
Creating the activity directly in the `WatchersPost` seems to be the best place to create the `WatcherAddedActivity`. @tinagerber and me discussed this option. I have in mind, that @lukasgraf was not satisfied with this solution because it's not close enough to the notification implementation. @lukas, pleas tell me if i'm wrong... for me, it's still the best option.

2. Big refactoring
We would have to make a big refactoring. I don't know what we have to touch, but I think a lot! And we have to think a lot. We would need to decouple the notification center from the activities to achieve our goal. I think, that's not possible in reasonable time.

3. Using the local import

@tinagerber @lukasgraf @njohner so what do you think? Do you have a better option? Or do you see an easy way to refactor it without touching the whole notification system?

tinagerber: @lukasgraf  and I discussed this again today and decided to leave it that way for the moment and add a comment. The second option would probably be the nicest, but this would go beyond the scope of this issue.

njohner: Yes I'm fine with the event. Certainly not adding activity in the API endpoints, that's just not the place.

PR #6454, opened 2020-05-19
Add version constraints for tests with Plone 4.3

buchi:
Currently setuptools is unpinned in tests which results in picking an
incompatible version that no longer supports Python 2.7.
Installation itself currently does not fail, but some eggs (e.g.
zope.interface) are not installed correctly and tests fail because of
that. This does not manifest when using shared eggs and the affected
eggs were installed with an older setuptools before.

By extending from test-versions-plone-4.cfg we include our known
version constraints for Plone 4.3/Python2.7.

## Checklist (Must have)

- [ ] Changelog entry


deiferni:
backported to `2020.2-stable`

PR #6455, opened 2020-05-19
Remove searchable text

elioschmutz:


opengever/dossier/tests/test_indexer.py
```diff
@@ -164,52 +165,27 @@ def test_keywords_field_is_indexed_in_Subject_index(self):
...
-            u'900',
-            index_data_for(self.dossier).get('SearchableText'),
-            )
-
-        self.assertIn(
-            u'9001',
-            index_data_for(self.dossier).get('SearchableText'),
-            )
+        self.assertIn(u'qpr', indexed_value)
+        self.assertIn(u'900', indexed_value)
```
buchi, line 179: This doesn't make much sense if the following assertion is met. Maybe add whitespaces if you're expecting `'900'` as a word?

opengever/dossier/tests/test_indexer.py
```diff
@@ -292,14 +268,15 @@ def test_returns_filing_number_for_dossiers_with_only_filing_prefix_information(
...
         IFilingNumber(self.dossier).filing_no = self.filing_no
         self.dossier.reindexObject()
+        self.commit_solr()
 
-        searchable_text_data = index_data_for(self.dossier).get('SearchableText')
+        indexed_value = solr_data_for(self.dossier, 'SearchableText')
 
         for segment in self.searchable_filing_no:
-            self.assertIn(segment, searchable_text_data)
+            self.assertIn(segment, indexed_value.lower())
```
buchi, line 282: Solr's SearchableText should contain the full `searchable_filing_no`. Thus it should be possible to assert containment of the whole filing number.
Isn't `searchable_filing_number` a string? Thus a segment would be a character, which makes this assertion weak.

opengever/dossier/tests/test_indexer.py
elioschmutz: I'm not sure if I understand your suggestion correctly. I assert the whole fillling_no now...

opengever/dossier/tests/test_indexer.py
elioschmutz: I'm not sure if I understand your suggestion correctly. I assert the whole external reference now...

opengever/dossier/tests/test_indexer.py
buchi: My point was if you do `assertIn(u'9001', indexed_value)` then `assertIn(u'900', indexed_value)` is already covered by the other assertion. There's no way that this assertion can fail while the `9001` assertion succeeds. Asserting the whole external reference is better anyway.

elioschmutz: ah ok. Thanks for the clarification

PR #6456, opened 2020-05-22
Catalog cleanup

buchi:
Remove a first batch of unneeded catalog indexes and metadata. This mainly reduces memory usage and slightly reduces indexing time.

The following _indexes_ have been removed:
- assigned_client
- client_id
- commentators
- Date
- Description
- effective
- effectiveRange
- expires
- getRawRelatedItems
- in_reply_to
- meta_type
- SearchableTex
- total_comments

The following _metadata_ columns have been removed:
- assigned_client
- author_name
- commentators
- CreationDate
- Date
- effective
- EffectiveDate
- ExpirationDate
- expires
- getObjSize
- getRemoteUrl
- last_comment_date
- location
- meta_type
- ModificationDate
- total_comments

There is an upgrade step that removes the indexes and metadata for existing installations. For indexes and metadata added by Plone there is a hook that removes those after site setup.

`CreationDate` and `ModificationDate` are now computed from `creation` and `modified` in `IContentListings`

As all the `effective` and `expires` indexes are removed, it's no longer possible to filter out inactive content. But that was never used so far.

Because of the removal of `getRemoteUrl` the `CatalogNavigationTabs` had to be customized and it's no longer supported to add Link content objects with external URLs.

 Catalog based filtering was removed in `UsersContactsInboxesSource` and `AllEmailContactsAndUsersSource`.

The `@livesearch` API endpoint now requires Solr.

The `folder_contents` View no longer displays the size (was always 0 KB anyway) and the modification date (uses the `Date` metadata column), but it still works.

Jira:
https://4teamwork.atlassian.net/browse/GEVER-154
https://4teamwork.atlassian.net/browse/GEVER-374

## Checklist (Must have)

- [x] Changelog entry

## Checklist (optional)


opengever/core/upgrades/20200518084137_cleanup_catalog_indexes_and_metadata/upgrade.py
```diff
@@ -0,0 +1,90 @@
...
+        self.remove_indexes()
+        self.remove_metadata()
+
+    def remove_indexes(self):
+        catalog = self.getToolByName('portal_catalog')
+        for index in INDEXES_TO_REMOVE:
+            if index in catalog._catalog.indexes:
+                catalog._catalog.delIndex(index)
+
+    def remove_metadata(self):
```
lukasgraf, line 57: The hook that drops metadata on site setup has been rewritten to use `catalog._catalog.delColumn(column)` instead of this custom code - can this be done for the upgrade step as well?

opengever/core/hooks.py
```diff
@@ -140,6 +140,7 @@ def remove_unused_catalog_metadata(site):
         'EffectiveDate',
         'ExpirationDate',
         'expires',
+        'getObjSize',
```
lukasgraf, line 144: `getObjSize` is also a Solr field.
Removing the catalog metadata column means that in a call like `reindexObject(idxs=['someOtherIndex'])` this won't be updated in Solr any more, whereas it would have been before (since `ftw.solr` [builds the intersection of names between `idxs` and the Solr schema, if `idxs` is specified](https://github.com/4teamwork/ftw.solr/blob/master/ftw/solr/handlers.py#L75-L79). 

Could this lead to cases where `getObjSize` now doesn't get updated in Solr anymore because we relied on it happening as a side effect before? Or is this basically ununsed, and could therefore also be removed from Solr?

buchi: Not sure I understand you correctly. IMO `reindexObject(idxs=['someOtherIndex'])` would have updated the metadata in the catalog but not the field in Solr. 

I guess we are not using this field in Solr anyway. But I didn't check any Solr fields yet.

opengever/core/upgrades/20200518084137_cleanup_catalog_indexes_and_metadata/upgrade.py
buchi: We could, but then we would iterate multiple times through all metadata tuples.
For large systems this could make the upgrade slower and we wouldn't have a progress indicator.
I'm not sure about the impact however, as after the first iteration everything should be in RAM.

Using `delColumn` in the hook is no problem as the amount of items is small on site setup.

lukasgraf: Ah, I see the reasoning now. WFM 👍

opengever/core/hooks.py
lukasgraf: Ah, you're right, I was thinking about a slightly different case. The Solr field would indeed not have been updated already in the case I described. 👍

PR #6457, opened 2020-05-22
Add attachments in mail serializer.

njohner:
In order to be able to work with Mails in the GEVER-UI we needed several changes in the backend:
- return the attachments in the GET of `Mail`s, to allow representing attachments in the detailview
- A new endpoint `extract-attachments` to extract the attachments of an `Mail` as a separate document

There are also some new requirements for `Mail`s, that will allow a more modern representation of mails in GEVER:
- Each attachment should only be extracted once
- For each attachment the `Mail` should store a link to the extracted document
- deleting of attachments from the `Mail` itself should not be allowed anymore (mail messages should be immutable)
This allows to represent each attachment as either a link to the extracted document or as a link to extracting the attachment.

On the technical side, information about attachments are already stored on `Mail`s in a dictionary:
- This is sadly not in a `PersistentMapping`, I chose to leave it as is, as otherwise we would need to migrate all existing mails. 
- We now also store the information about the extracted document in that mapping.
- This information now needs to be updated when an extracted document is moved
- I also handle the cases where an extracted document is deleted. The attachment can then be extracted again
- The case where the `Mail` is deleted is partially handled, the `IExtractedFromMail` interface is removed from the document, but the extracted document will still be broken, because the `Mail` is also in the RelatedItems, and these fields do not support object deletion properly. But this is a general problem we have in GEVER, not specific to this situation.

Some more edge-cases and comments:
- When an attachment gets extracted and then moved to a position a user cannot see. @all4manu what happens then in the GEVER-UI? No link?
- E-mails with documents that have been extracted prior to this change will be confusing (documents will not show up the same in the GEVER-UI, and users might extract them again)

For https://4teamwork.atlassian.net/browse/GEVER-198

## Checklist (Must have)
- [x] Changelog entry

## Checklist (optional)
- New translations

### DIFF
--- a/docs/public/dev-manual/api/extract_attachments.rst
+++ b/docs/public/dev-manual/api/extract_attachments.rst
@@ -1,0 +1,37 @@
+
+Anhänge speichern
+=================
+
+E-Mail-Anhänge können in OneGov GEVER als separate Dokumente gespeichert werden (nur einmal pro Anhang). Dies wird in der REST API mit einem ``POST`` Request auf den ``@extract-attachments`` Endpoint ermöglicht. Der ``positions`` Parameter erlaubt auszuwählen, welche Anhänge extrahiert werden sollen und entspricht der ``position``, die in der Response eines ``GET`` Request auf eine E-Mail für jeden Anhang enthalten ist. Wenn ``positions`` nicht angegeben wird, dann werden alle Anhänge extrahiert, welche noch nicht gespeichert wurden. Die Dokumente werden im selben Gefäss erstellt, in welchem sich die E-Mail befindet.
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+       POST http://localhost:8080/fd/ordnungssystem/bildung/dossier-25/document-78/@extract-attachments HTTP/1.1
+       Accept: application/json
+
+       {
+        "positions": [4, 5]
+       }
+
+
+**Beispiel-Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+      [
+          {
+              "extracted_document_title": "Ein Dokument",
+              "extracted_document_url": "http://localhost:8080/fd/ordnungssystem/bildung/dossier-25/document-79",
+              "position": 4
+          },
+          {
+              "extracted_document_title": "Ein weiteres Dokument",
+              "extracted_document_url": "http://localhost:8080/fd/ordnungssystem/bildung/dossier-25/document-80",
+              "position": 5
+          }
+      ]
--- a/opengever/api/mail.py
+++ b/opengever/api/mail.py
@@ -1,11 +1,13 @@
 from ftw.mail.mail import IMail
+from opengever.api.document import SerializeDocumentToJson
 from opengever.base.transforms.msg2mime import Msg2MimeTransform
-from opengever.mail.mail import initialize_title
 from opengever.mail.mail import initialize_metadata
+from opengever.mail.mail import initialize_title
 from plone.namedfile.file import NamedBlobFile
 from plone.restapi.deserializer import json_body
 from plone.restapi.deserializer.dxcontent import DeserializeFromJson
 from plone.restapi.interfaces import IDeserializeFromJson
+from plone.restapi.interfaces import ISerializeToJson
 from zope.component import adapter
 from zope.interface import implementer
 from zope.interface import Interface
@@ -42,3 +44,13 @@ class DeserializeMailFromJson(DeserializeFromJson):
             initialize_metadata(context, None)
 
         return context
+
+
+@implementer(ISerializeToJson)
+@adapter(IMail, Interface)
+class SerializeMailToJson(SerializeDocumentToJson):
+
+    def __call__(self, *args, **kwargs):
+        result = super(SerializeMailToJson, self).__call__(*args, **kwargs)
+        result[u'attachments'] = self.context.get_attachments()
+        return result
--- a/opengever/api/mail.py
+++ b/opengever/api/mail.py
@@ -5,6 +5,7 @@ from opengever.mail.exceptions import AlreadyExtractedError
 from opengever.mail.exceptions import InvalidAttachmentPosition
 from opengever.mail.mail import initialize_metadata
 from opengever.mail.mail import initialize_title
+from plone.app.uuid.utils import uuidToURL
 from plone.namedfile.file import NamedBlobFile
 from plone.protect.interfaces import IDisableCSRFProtection
 from plone.restapi.deserializer import json_body
@@ -57,8 +58,16 @@ class SerializeMailToJson(SerializeDocumentToJson):
 
     def __call__(self, *args, **kwargs):
         result = super(SerializeMailToJson, self).__call__(*args, **kwargs)
-        result[u'attachments'] = self.context.get_attachments()
+        result[u'attachments'] = self.get_attachments()
         return result
+
+    def get_attachments(self):
+        attachments = self.context.get_attachments()
+        for attachment in attachments:
+            if attachment.get('extracted'):
+                attachment['extracted_document_url'] = uuidToURL(
+                    attachment.get('extracted_document_uid'))
+        return attachments
 
 
 class ExtractAttachments(Service):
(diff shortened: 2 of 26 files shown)


opengever/mail/tests/test_extract_attachments.py
```diff
@@ -23,6 +24,25 @@ def setUp(self):
                            .within(self.dossier)
                            .with_message(MESSAGE_TEXT))
 
+    @browsing
+    def test_extract_attachment_view_content(self, browser):
+        browser.login().open(self.mail, view='extract_attachments')
+        table = browser.css('table').first
+        self.assertEqual(
+            [['', 'Already extracted', 'Type', 'Filename', 'Size'],
+             ['', 'No', '', u'B\xfccher.txt', '1 KB']],
```
mbaechtold, line 33: Should we also add a test for an extracted e-mail?

docs/public/dev-manual/api/extract_attachments.rst
```diff
@@ -0,0 +1,37 @@
...
+Anhänge speichern
+=================
+
+E-Mail-Anhänge können in OneGov GEVER als separates Dokument gespeichert werden (nur einmal für jeder Anhang). Dies wird mit in der REST API mit einer ``POST`` Request auf den ``@extract-attachments`` Endpoint ermöglicht. Der ``positions`` Parameter erlaubt auszuwählen welche Anhänge extrahiert werden sollen und entspricht der ``position`` die in der Response eines ``GET`` Request auf ein E-Mail für jeder Anhang enthalten ist. Wenn ``positions`` nicht angegeben wird, dann werden alle Anhänge, die noch nicht gespeichert wurden, extrahiert. Die Dokumente werden im selben Gefäss wo sich der E-Mail befindet erstellt.
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+       POST http://localhost:8080/fd/ordnungssystem/bildung/dossier-25/document-78 HTTP/1.1
```
mbaechtold, line 11: The endpoint `@extract-attachments`  is missing in the url.

docs/public/dev-manual/api/extract_attachments.rst
```diff
@@ -0,0 +1,37 @@
+
+Anhänge speichern
+=================
+
+E-Mail-Anhänge können in OneGov GEVER als separates Dokument gespeichert werden (nur einmal für jeder Anhang). Dies wird mit in der REST API mit einer ``POST`` Request auf den ``@extract-attachments`` Endpoint ermöglicht. Der ``positions`` Parameter erlaubt auszuwählen welche Anhänge extrahiert werden sollen und entspricht der ``position`` die in der Response eines ``GET`` Request auf ein E-Mail für jeder Anhang enthalten ist. Wenn ``positions`` nicht angegeben wird, dann werden alle Anhänge, die noch nicht gespeichert wurden, extrahiert. Die Dokumente werden im selben Gefäss wo sich der E-Mail befindet erstellt.
```
mbaechtold, line 5: E-Mail-Anhänge können in OneGov GEVER als separate Dokumente gespeichert werden (nur einmal pro Anhang). Dies wird in der REST API mit einem ``POST`` Request auf den ``@extract-attachments`` Endpoint ermöglicht. Der ``positions`` Parameter erlaubt auszuwählen, welche Anhänge extrahiert werden sollen und entspricht der ``position``, die in der Response eines ``GET`` Request auf eine E-Mail für jeden Anhang enthalten ist. Wenn ``positions`` nicht angegeben wird, dann werden alle Anhänge extrahiert, welche noch nicht gespeichert wurden. Die Dokumente werden im selben Gefäss erstellt, in welchem sich die E-Mail befindet.

opengever/api/mail.py
```diff
@@ -54,3 +57,25 @@ def __call__(self, *args, **kwargs):
         result = super(SerializeMailToJson, self).__call__(*args, **kwargs)
         result[u'attachments'] = self.context.get_attachments()
         return result
+
+
+class ExtractAttachments(Service):
+
+    def reply(self):
+
+        # Disable CSRF protection
```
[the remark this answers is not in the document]
njohner: Agreed that the comment is not very useful, and we have it copy pasted in the whole API. I'll update it at least here...

Basically every POST request should include an `X-CSRF-TOKEN` in its headers that was emitted by plone for a given user. This is achieved using javascript in forms created by plone, but cannot work for REST-API requests. So basically all POST endpoints must disable the CSRF protection.

docs/public/dev-manual/api/extract_attachments.rst
njohner: Thanks! corrected...

docs/public/dev-manual/api/extract_attachments.rst
njohner: Thanks!

opengever/api/mail.py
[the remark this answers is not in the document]
mbaechtold: Thanks for explanation 👍

opengever/api/mail.py
```diff
@@ -42,3 +44,13 @@ def __call__(self, validate_all=False, data=None, create=False):
...
         return context
+
+
+@implementer(ISerializeToJson)
+@adapter(IMail, Interface)
+class SerializeMailToJson(SerializeDocumentToJson):
+
+    def __call__(self, *args, **kwargs):
+        result = super(SerializeMailToJson, self).__call__(*args, **kwargs)
+        result[u'attachments'] = self.context.get_attachments()
```
elioschmutz, line 60:
```suggestion
        result[u'items'] = self.context.get_attachments()
```

See discussion here: https://github.com/4teamwork/opengever.core/pull/6462#discussion_r431034923

opengever/mail/mail.py
```diff
@@ -213,6 +232,11 @@ def extract_attachment_into_parent(self, position):
             IRelatedDocuments(doc).relatedItems = [RelationValue(iid)]
             doc.reindexObject()
 
+        # mark attachment as extracted
+        info['extracted'] = True
+        info['extracted_document_url'] = doc.absolute_url()
```
elioschmutz, line 243: Why do we need to track the URL and the UID? I think the UID is enough, we can lookup the document and get the URL from it.

Then we also don't need to track moved documents because the url is generated.

opengever/mail/subscribers.py
```diff
@@ -13,3 +19,35 @@ def resolve_mail_author(mail, event):
...
+
+        for info in related_obj.get_attachments():
+            if info.get("extracted_document_uid") == uid:
+                write_info = related_obj._get_attachment_info(
+                    info.get("position"), write_modus=True)
+                return write_info
+    raise SourceMailNotFound("Source Mail not found when Extracted attachment moved.")
+
+
+def extracted_attachment_moved(doc, event):
```
elioschmutz, line 42: ✂️  This subscriber is not required if we only track the documents UID.

opengever/api/mail.py
njohner: I'd tend to disagree here, because they are not children of the E-mail. For comparison, in the case of a user, this would be analogous to his `addresses` or something like that.

opengever/mail/mail.py
njohner: We need the URL because the Gever-UI needs it. But I guess we could get the URL for the UID in the serializer.

opengever/mail/mail.py
```diff
@@ -148,17 +150,30 @@ def get_extraction_parent(self):
...
+            return self.attachment_infos
+        return tuple(info for info in self.attachment_infos if not info.get('extracted'))
 
     def has_attachments(self):
         """Return whether this mail has attachments."""
         return len(self.get_attachments()) > 0
 
+    def _get_attachment_info(self, position, write_modus=False):
+        """Return the attachment info for attachment at given position.
+        If write_modus is True, return the actual persistent mapping"""
```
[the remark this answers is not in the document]
njohner: Solution  1 is problematic here, as this is not a persistent mapping.
Solution 2 should make it cleaner.

opengever/api/mail.py
njohner: As discussed I'll leave this under `attachments`

opengever/mail/mail.py
```diff
@@ -148,17 +150,30 @@ def get_extraction_parent(self):
...
+        for info in self._attachment_infos:
+            if info['position'] == position:
+                if not write_modus:
+                    return dict(info)
+                self._p_changed = True
+                return info
```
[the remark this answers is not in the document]
njohner: Has been refactored

opengever/mail/mail.py
njohner: We now only store the UID. URL gets added in the API endpoint

njohner:
I incorporated the changes as discussed.

PR #6460, opened 2020-05-25
Rename ogds-user endpoint to ogds-users.

tinagerber:
There are naming conventions for the REST API that state that the plural should be used for endpoint names.
https://plonerestapi.readthedocs.io/en/latest/conventions.html#singluar-vs-plural
 Therefore the endpoint `@ogds-user` is renamed to `@ogds-users`.

Jira: https://4teamwork.atlassian.net/browse/GEVER-418

## Checklist (Must have)

- [x] Changelog entry


njohner:
I added the `api-change` label. This is a breaking change and needs to be communicated. I'm not completely sure what the process is TBH. but I remember the label is how we find the corresponding changes...

tinagerber:
Ok, I added a docs changelog entry. 
We don't need a PR for the UI, because the endpoint is only used indirectly via the `@ogds-user-listing` endpoint.

PR #6462, opened 2020-05-26
Add @ogds-groups API endpoint and extend @sharing endpoint with ogds_summary

tinagerber:
A new detail view for groups shall be implemented, which shows which users belong to a group.

To implement this, the new endpoint `@ogds-groups` is needed to provide the necessary information.

To be able to show a detail view in the sharing view as well, it must be known where to find the required information for a group.Therefore the `@sharing` endpoint has been extended with `ogds_summary`, for OGDS groups the dict contains the JSON summary of the group, for users the JSON summary of the user.

Jira: https://4teamwork.atlassian.net/browse/GEVER-418

## Checklist (Must have)

- [x] Changelog entry


opengever/api/ogdsgroups.py
```diff
@@ -0,0 +1,37 @@
...
+from zExceptions import BadRequest
+from zope.component import queryMultiAdapter
+from zope.interface import implements
+from zope.publisher.interfaces import IPublishTraverse
+
+
+class OGDSGroupsGet(Service):
+    """API Endpoint that returns a single group from ogds.
+
+    GET /@groups/group.id HTTP/1.1
```
elioschmutz, line 13:
```suggestion
    GET /@ogds-groups/group.id HTTP/1.1
```

opengever/api/serializer.py
```diff
@@ -130,6 +130,22 @@ def add_additional_metadata(self, data):
...
+@adapter(Group, IOpengeverBaseLayer)
+class SerializeGroupModelToJson(SerializeSQLModelToJsonBase):
+
+    content_type = 'virtual.ogds.group'
+
+    def add_additional_metadata(self, data):
+        """Add the team members, group summary and org_unit_title"""
+
+        data['users'] = []
+        for user in self.context.users:
```
elioschmutz, line 145: There could be thousands of users in a group. The users listing needs to be batched or limited.

opengever/api/serializer.py
```diff
@@ -229,7 +231,11 @@ def __call__(self, *args, **kwargs):
 
     @property
     def get_url(self):
-        base_url = self.request.URL.rsplit("/@")[0]
+        try:
+            base_url = get_contactfolder_url()
```
elioschmutz, line 235: Sure? Not only contacts are stored in the OGDS. There are tasks, notifications, favorites and so on... this serializer is the base for all of those objects.

Perhaps you can add a contact specific base class.

opengever/sharing/browser/sharing.py
```diff
@@ -220,23 +222,30 @@ def role_settings(self):
...
         as this was determined from the acl_users which wrongly identifies
         inactive groups (groups that were deleted from the LDAP) as users.
         Instead we check using the ogds service.
         """
-        if ogds_service().fetch_group(item['id']):
-            return '{}/@@list_groupmembers?{}'.format(
+        ogds_summary = None
+        url = None
+        group = ogds_service().fetch_group(item['id'])
+        user = ogds_service().fetch_user(item['id'])
```
elioschmutz, line 234:
```suggestion
        obds_service = ogds_service()
        group = ogds_service.fetch_group(item['id'])
        user = ogds_service.fetch_user(item['id'])
```

opengever/api/serializer.py
tinagerber: The serializer is only for ogds-user, team and ogds-groups the base, so I renamed it. But the name is not so well chosen yet.

opengever/api/serializer.py
tinagerber: Then we should probably also use batching for `SerializeTeamModelToJson`.
Can I still call the list `users` then? It seems that all other lists that use batching are called `items`.

opengever/api/serializer.py
tinagerber: I have now implemented a separate base class for users, teams and groups, but I still haven't come up with a good name: `SerializeContactModelToJsonSummaryBase`

tinagerber:
@elioschmutz I have now implemented the batching for the groups the way I think it should be, can you look at it again?

opengever/api/serializer.py
elioschmutz: Yes, I think it makes sense to also update the `SerializeTeamModelToJson` 👍 

And yes, you could still call it `users`. Technically. But I think it's better to use `items` because it's the convention.

But honestly speaking, I'd be glad if someone else, @deiferni or @njohner, could approve my proposal.

opengever/api/serializer.py
```diff
@@ -139,13 +140,15 @@ class SerializeGroupModelToJson(SerializeSQLModelToJsonBase):
...
-        data['users'] = []
-        for user in self.context.users:
-            user_serializer = queryMultiAdapter(
-                (user, self.request), ISerializeToJsonSummary)
-            data['users'].append(user_serializer())
+        data['users_total'] = batch.items_total
+        data['users'] = users
+
+        if batch.links:
+            data['batching'] = batch.links
```
elioschmutz, line 151: Sorry, I haven't seen that you already implemented it. 👍 i like it. But as I already wrote in the previous comment, I'd be glad if someone else could prove the naming.

opengever/api/serializer.py
elioschmutz: Suggestion:

You could use a mixin instead of a introducing a new base class:

```python
class ContactFolderRestrictionMixin(object):
    @property
    def get_url(self):
        """
        """
```

Then you can just extend the new classes:

```python
class SerializeTeamModelToJsonSummary(ContactFolderRestrictionMixin, SerializeSQLModelToJsonSummaryBase):
```

opengever/api/serializer.py
deiferni: IMO we should strictly stick to `items` as that is the generic way that child (summaries) are returned in the rest api and so it provides a consistent way to access them regardless of content type.

I also totally agree about batching, `items` should be batched. You can have a look at the implementation for e.g. dx folderish content at https://github.com/plone/plone.restapi/blob/b2a83a45585b8f2f8e4cfc28f378d4affc12cb2e/src/plone/restapi/serializer/dxcontent.py#L168-L171. 

For SQL batching we already have `SQLHypermediaBatch` which may help here https://github.com/4teamwork/opengever.core/blob/94148564c8ee5bbb3e56212927d3c8a4f78ee6de/opengever/api/batch.py.

opengever/api/serializer.py
```diff
@@ -17,6 +17,7 @@
 from plone import api
 from plone.dexterity.interfaces import IDexterityContainer
 from plone.dexterity.interfaces import IDexterityContent
+from plone.restapi.batching import HypermediaBatch
```
elioschmutz, line 20: use `from opengever.api.batch import SQLHypermediaBatch`

opengever/api/serializer.py
elioschmutz: As discussed, use `items`instead of `users`

opengever/api/serializer.py
```diff
@@ -130,6 +133,24 @@ def add_additional_metadata(self, data):
...
 
 
+@implementer(ISerializeToJson)
+@adapter(Group, IOpengeverBaseLayer)
+class SerializeGroupModelToJson(SerializeSQLModelToJsonBase):
+
+    content_type = 'virtual.ogds.group'
+
+    def add_additional_metadata(self, data):
+        batch = HypermediaBatch(self.request, self.context.users)
```
deiferni, line 143: you should use `SQLHypermediaBatch` to profit from limited and offset queries. otherwise you will query for all items and also create potentially expensive ORM objects for all of them. You can have a look at the task implementation for an example on how to do that. https://github.com/4teamwork/opengever.core/blob/94148564c8ee5bbb3e56212927d3c8a4f78ee6de/opengever/api/globalindex.py#L48-L65. you'll probably have to build a query for each contact class, or implement a factory method in the query helper classes.

i can show you what i mean in a quick pair programming sessions if it is not clear from my comment here 😅.

opengever/api/serializer.py
```diff
@@ -130,6 +133,24 @@ def add_additional_metadata(self, data):
...
+        users = [queryMultiAdapter((user, self.request), ISerializeToJsonSummary)()
+                 for user in list(batch)]
+
+        data['users_total'] = batch.items_total
+        data['users'] = users
```
deiferni, line 148: as already mentioned i strongly tend towards the generic `items` here

opengever/api/serializer.py
deiferni: @elioschmutz i have given my thoughts/opinion, looks like our comments overlapped 😄 . but also looks like we agree 😉.

elioschmutz: 👌 perfect! Thanks a lot!

opengever/api/serializer.py
tinagerber: If I use a mixin, I have to define things twice (`self.endpoint`, `self.id_attribute_name`). I also need `self.context` in `get_url`. I'm not sure if this makes sense.

elioschmutz: No, then go for the subclass 👍

tinagerber:
I have now implemented batching for teams and groups. @elioschmutz @deiferni Can you take another look at it?

opengever/api/serializer.py
```diff
@@ -98,6 +102,7 @@ def __call__(self, *args, **kwargs):
 
         data['@type'] = self.content_type
         data['@id'] = self.request.URL
+        self.batch_items(data)
```
deiferni, line 105: `add_batch_items`  `add_batched_items`?

PR #6464, opened 2020-05-26
Make sure SQL references are up to date in watchers test.

njohner:
`test_delete_watchers_for_inbox_forwarding` was flaky, failing systematically in cached mode and sometimes failing on CI (e.g. https://ci.4teamwork.ch/builds/335741/tasks/561628)

Problem was that the notification center uses the reference to the `Subscription`s on a `Resource` to get the `Subscription`s (https://github.com/4teamwork/opengever.core/blob/master/opengever/activity/center.py#L104-L112). This reference is not always up to date during tests, and the `Resource` first needs to be refreshed.

thanks @deiferni for the help!

## Checklist (Must have)
- [x] Changelog entry. No changelog
- [x] Documentation updated (notably for API and deployment).
- [x] Link to issue (Jira or GitHub) and backlink in issue (Jira). No issue

### DIFF
--- a/opengever/api/tests/test_watchers.py
+++ b/opengever/api/tests/test_watchers.py
@@ -1,6 +1,8 @@
+from collections import defaultdict
 from ftw.testbrowser import browsing
 from opengever.activity import notification_center
 from opengever.activity.roles import WATCHER_ROLE
+from opengever.base.model import create_session
 from opengever.testing import IntegrationTestCase
 import json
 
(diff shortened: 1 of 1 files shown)


opengever/api/tests/test_watchers.py
```diff
@@ -284,62 +285,70 @@ def test_post_watchers_for_inbox_forwarding(self, browser):
 class TestWatchersDelete(IntegrationTestCase):
     features = ('activity', )
 
+    def get_watchers_and_roles(self, context):
+        center = notification_center()
+        watchers_and_roles = defaultdict(list)
+        resource = center.fetch_resource(context)
+        center.session.refresh(resource)
```
deiferni, line 292: a detail to some degree, but i'd use `create_session` here instead of accessing centers attribute as a foreigner.

PR #6471, opened 2020-05-28
Fix MergeNotificationSettings upgradestep to use a fixed configuration

elioschmutz:
The `MergeNotificationSettings` relies on the main notification configuration which leads to issues when upgrading multiple upgradesteps changing the notification defaults.

This PR uses a static configuration for the `MergeNotificationSettings` upgradestep` to fix this issue.

I manually updated the dev.onegovgever.ch ogds with:

```sql
update notification_defaults set digest_notification_roles='["workspace_member_role"]' where id=173;
update notification_defaults set digest_notification_roles='["workspace_member_role"]' where id=174;
```

to add the missing digest configuration.

### DIFF
--- a/opengever/core/upgrades/20200430153116_merge_notification_settings/upgrade.py
+++ b/opengever/core/upgrades/20200430153116_merge_notification_settings/upgrade.py
@@ -231,6 +231,8 @@ class MergeNotificationSettings(SchemaMigration):
                     default_settings.get('mail_notification_roles', []))
             setattr(setting, 'badge_notification_roles',
                     default_settings.get('badge_notification_roles', []))
+            setattr(setting, 'digest_notification_roles',
+                    default_settings.get('digest_notification_roles', []))
 
     def migrate_custom_settings(self):
         # Get a list of distinct userids of users that have any settings stored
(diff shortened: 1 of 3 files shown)


opengever/core/upgrades/20200430153116_merge_notification_settings/upgrade.py
```diff
@@ -12,27 +20,225 @@
...
     """Merge notification settings.
     """
     def migrate(self):
+        self.notification_settings = NotificationSettings(NOTIFICATION_CONFIGURATION)
         self.migrate_default_settings()
         self.migrate_custom_settings()
 
     def migrate_default_settings(self):
+        # Remove all notification defaults
         NotificationDefault.query.delete()
```
njohner, line 222: This is fine even if the model has changed in the meantime right?

opengever/core/upgrades/20200430153116_merge_notification_settings/upgrade.py
```diff
@@ -12,27 +20,225 @@
...
+        # Re-add the new notification defaults
+        for item in self.notification_settings.configuration:
+            setting = NotificationDefault(kind=item.get('id'))
+            self.session.add(setting)
+
+            default_settings = item.get('default_settings')
+            setattr(setting, 'mail_notification_roles',
+                    default_settings.get('mail_notification_roles', []))
+            setattr(setting, 'badge_notification_roles',
+                    default_settings.get('badge_notification_roles', []))
```
njohner, line 233: As discussed, `digest_notification_roles ` is missing here

PR #6475, opened 2020-05-29
Allow notifying users and groups when creating a new task.

njohner:
We add a new `informed_principals` field to the `TaskAddForm`, listing principals that will receive a notification for the newly created `Task`. This allows avoiding having to create separate `Task`s just to inform certain users about a new `Task`.

This is implemented by temporarily adding watchers to the `Task` before creation and removing them directly after, so that they just get informed of the creation itself. As they get added in the `WATCHER_ROLE` we need to avoid generating a `WatcherAddedActivity`. Note that we refactor the generation of the `WatcherAddedActivity` a bit, as I found it a bit confusing that they such an activity only gets generated when I add the watcher through the `PloneNotificationCenter` but not if I used the `NotificationCenter`. 

We also only show the field when the `activity` feature is enabled, as it otherwise does not make much sense.

Finally there was a bug, preventing to set the issuer and informed_principals when adding a forwarding.

This is for https://4teamwork.atlassian.net/browse/GEVER-31
Frontend PR: https://github.com/4teamwork/gever-ui/pull/1120

## Checklist (Must have)

- [x] Changelog entry
- [x] Documentation updated (notably for API and deployment): Screenshot would need an update

## Checklist (optional)

- New translations

### DIFF
--- a/opengever/task/activities.py
+++ b/opengever/task/activities.py
@@ -6,6 +6,7 @@ from opengever.activity.base import BaseActivity
 from opengever.activity.model.notification import Notification
 from opengever.activity.roles import TASK_OLD_RESPONSIBLE_ROLE
 from opengever.activity.roles import TASK_REMINDER_WATCHER_ROLE
+from opengever.activity.roles import WATCHER_ROLE
 from opengever.base.model import get_locale
 from opengever.ogds.base.actor import Actor
 from opengever.ogds.base.actor import SYSTEM_ACTOR_ID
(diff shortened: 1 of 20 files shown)


mbaechtold:
@njohner, this is not a review but a friendly reminder asking you to rebase onto the latest `master` and then to update the `docs/HISTORY.txt`

opengever/task/activities.py
```diff
@@ -128,6 +129,14 @@ def before_recording(self):
         self.center.add_task_responsible(self.context,
                                          self.context.responsible)
         self.center.add_task_issuer(self.context, self.context.issuer)
+        for principal in self.context.informed_principals:
+            self.center.add_watcher_to_resource(
+                self.context, principal, omit_watcher_added_event=True)
```
lukasgraf, line 134: As discussed:
Not a huge fan of abusing a temporary watcher to trigger the notification, but doing it properly would require a substantial refactoring that would bust the scope of this story, so we agreed to do it later.

_(The ideal way in my opinion would be to just create a (role-independent) notification in `create_notifications()` for  TaskAddedActivities. The problem with that is that the `create_notifications` method only exists on the very generic and abstract `Activity` SQL model. In order to take activity-specific knowledge into account it would be better suited on the domain-specific activity Python classes, like `TaskAddedActivity` in this case, and should probably be moved there. But as stated above, this will have to wait.)_

njohner: agreed...

PR #6477, opened 2020-06-02
Add last_login info to ogds user.

tinagerber:
An administrator should have the possibility to display on which day a user has logged in for the last time.

Since there should be a cross-client view for the users, the information must be stored in the OGDS.

Therefore the `users` model is extended with the attribute `last_login`.

The `@caslogin` endpoint didn't fire the right events, this could be corrected directly in the `ftw.casauth` package.

Furthermore the `@ogds-user-listing` is extended with the possibility to filter by `last_login` date.

Jira: https://4teamwork.atlassian.net/browse/GEVER-39

## Checklist (Must have)

- [x] Changelog entry

## Checklist (optional)

- DB-Schema migration


tinagerber:
@deiferni I still have to write tests and update the documentation, but could you, if you have time, take a first look at it?

opengever/ogds/models/user.py
```diff
@@ -50,6 +54,7 @@ class User(Base):
     country = Column(String(20))
 
     import_stamp = Column(String(26))
+    last_login = Column(Date())
```
deiferni, line 57: - you do not need to call it, you can pass in a reference to the class
- you must add an index with `index=True` here as we will query against this column and not having an index is potentially expensive, see https://docs.sqlalchemy.org/en/13/core/constraints.html#indexes

docs/HISTORY.txt
```diff
@@ -5,6 +5,11 @@ Changelog
 ------------------------
 
 - Support placeholders in the target url of the webactions. [mbaechtold]
+- Extend @ogds-user-listing endpoint with last_login filter. [tinagerber]
```
deiferni, line 8: IMO the changelog is a bit too granular, it is basically on the same level as your commits. we try to provide a summary in the CL, which may also help to write the customer facing release notes. so something like "add last login information to user" would suffice.

opengever/api/login.py
```diff
@@ -0,0 +1,74 @@
...
+    expire_clipboard(request)
+    mtool.createMemberArea(member_id=userid)
+    return member
+
+
+def set_login_times(member, membership_tool):
+    # The return value indicates if this is the first logged login time.
+
+    first_login = False
+    default = DateTime('2000/01/01')
```
deiferni, line 27: so just so i understand this correctly, that's the default which is somehow hardcoded in plone, too? otherwise i'd prefer `None`.

opengever/api/login.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def fire_login_events(first_login, user):
+    if first_login:
+        notify(UserInitialLoginInEvent(user))
+    else:
+        notify(UserLoggedInEvent(user))
+
+
+def expire_clipboard(request):
```
deiferni, line 48: unsure if this is necessary, why was it added to casauth 🤔 ?

opengever/api/login.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class LoginPost(Login):
+
+    def reply(self):
+        result = super(LoginPost, self).reply()
+        if 'token' in result:
+            data = json_body(self.request)
+            userid = data["login"]
+            if six.PY2:
```
deiferni, line 60: we are currently always on python2, i would not add such conditionals here. i don't think gever will support multiple python versions at the same time.

opengever/core/upgrades/20200527133213_add_last_login_column_to_users/upgrade.py
```diff
@@ -0,0 +1,13 @@
...
+from sqlalchemy import Date
+
+
+class AddLastLoginColumnToUsers(SchemaMigration):
+    """Add last_login column to users.
+    """
+
+    def migrate(self):
+        self.op.add_column(
+            'users', Column('last_login', Date(), nullable=True)
```
deiferni, line 12: also here, add `index`

opengever/api/login.py
tinagerber: Yes, that's the default in PlonePAS

opengever/api/login.py
tinagerber: I have no idea, why we need this. Maybe @buchi can tell us.

buchi: Plone does this by default: https://github.com/plone/Products.PlonePAS/blob/5.x/src/Products/PlonePAS/tools/membership.py#L663

We want the same behavior as when the user logs in using Plone's login form.

tinagerber:
@deiferni I have now added some tests and made the desired changes.

tinagerber:
@deiferni I have now dropped the commit that affects the @login endpoint, can you do another review?

mbaechtold:
@tinagerber, this is not a review but a friendly reminder asking you to rebase onto the latest `master` and then to update the `docs/HISTORY.txt`

opengever/ogds/base/handlers.py
```diff
@@ -0,0 +1,8 @@
+from opengever.ogds.models.user import User
+
+
+def user_logged_in(event):
+    ogds_user = User.query.get_by_userid(event.object.getId())
+    if ogds_user:
+        login_time = event.object.getProperty('login_time')
```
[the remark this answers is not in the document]
tinagerber: I don't think that's possible, but to calm your paranoia, I've added an extra control. 😉

tinagerber:
@deiferni I have made the desired changes and added a test with an actual login. Can you review it again?

PR #6478, opened 2020-06-02
Use Solr for dossier overview

buchi:
Use Solr for dossier overview. This includes getting documents and the dossier navigation.

Jira: https://4teamwork.atlassian.net/browse/GEVER-374

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/dossier/browser/navigation.py
+++ b/opengever/dossier/browser/navigation.py
@@ -1,17 +1,28 @@
+from ftw.solr.interfaces import ISolrSearch
+from ftw.solr.query import make_filters
 from opengever.base.browser.navigation import make_tree_by_url
+from opengever.base.solr import OGSolrContentListing
 from opengever.dossier.base import DOSSIER_STATES_OPEN
+from opengever.dossier.behaviors.dossier import IDossierMarker
+from opengever.dossier.dossiertemplate.behaviors import IDossierTemplateMarker
 from plone import api
 from Products.Five import BrowserView
+from zope.component import getUtility
 import json
 
 
-class JSONNavigation(BrowserView):
+class DossierJSONNavigation(BrowserView):
     """Return a JSON-navigation structure for a dossier (context) and its
     subdossiers.
 
     The navigation starts at the current context, i.e. the dossier.
 
     """
+
+    filters = {
+        'object_provides': IDossierMarker.__identifier__,
+        'review_state': DOSSIER_STATES_OPEN,
+    }
 
     def __call__(self):
         response = self.request.response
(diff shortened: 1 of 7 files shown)


opengever/dossier/browser/navigation.py
```diff
@@ -30,16 +35,33 @@ def _context_as_node(self):
...
+                'depth': -1,
+            },
+            review_state=DOSSIER_STATES_OPEN,
+        )
+        fieldlist = ['UID', 'Title', 'Description', 'path']
+        resp = solr.search(
+            filters=filters, start=0, rows=100000, sort='sortable_title asc',
+            fl=fieldlist)
+
+        def contentlisting_object_to_node(obj):
```
njohner, line 53: I'd rather have that as a `staticmethod` on the class

opengever/dossier/browser/navigation.py
```diff
@@ -30,16 +35,33 @@ def _context_as_node(self):
...
-            self._brain_to_node,
-            self.context.get_subdossiers(sort_on='sortable_title',
-                                         review_state=DOSSIER_STATES_OPEN))
-        if nodes:
-            nodes = [self._context_as_node()] + nodes
-        return make_tree_by_url(nodes)
+        solr = getUtility(ISolrSearch)
+        filters = make_filters(
+            trashed=False,
+            object_provides=IDossierMarker.__identifier__,
```
njohner, line 41: I think this change will break navigation for `DossierTemplate` for which this is also registered, but has different filters in its `get_subdossiers` method: https://github.com/4teamwork/opengever.core/blob/master/opengever/dossier/dossiertemplate/dossiertemplate.py#L136-L148

buchi: Good catch!
I've amended that.

opengever/dossier/browser/navigation.py
buchi: What's your reasoning? Personally I prefer nested functions for simple functions, that are used only at one place, because of readability. You don't have to scroll to lookup the code.
The downside is that you can't test it separately. But that's not an issue for functions that are as simple as this.

deiferni: i'm jumping in here to second @njohner, for me from an object orientation standpoint nested functions are a bit of an anti-pattern:
- as you already say not testable
- not reusable outside the method it is defined in
- cannot be overridden in potential subclasses

buchi: That's why it's implemented as a nested function 😉. It's not meant to be used outside or be overridden. For me readability still brings more value than potential customizability that still can be added if really needed.

buchi:
@njohner Added test for `DossierTemplateJSONNavigation`.

PR #6485, opened 2020-06-07
Fix p7m attachment extraction from mails.

deiferni:
When extracting p7m attachments for mails they were created as a document by mistake. This PR makes sure we use the same special handling when extracting p7ms as during D&D upload and during mail-in unwrapping. We make sure the extracted part is correctly created as another mail and not as a a document.

https://4teamwork.atlassian.net/browse/GEVER-467

## Checklist (Must have)

- [x] Changelog entry


deiferni:
backported to `2020.2-stable`

PR #6486, opened 2020-06-08
Add support for pivot queries in listing-stats endpoint.

njohner:
In the new frontend, we use the `listing-stats` endpoint to show the total number of items in the different tabs, without having to query the corresponding listing. But there are several places where we have hidden filters, for example:
- only show dossiers for current responsible on the dashboard
- only show `opengever.document.document` in template folder and not all `BaseDocument`
- only count main dossiers and tasks in the respective tabs... 

So to get be able to display the correct total item counts in such situations, we add support for pivot queries in the `listing-stats` endpoint. This allows to define queries as request parameter for which statistics will be calculated for each pivot. I also added support for `depth`, which we will need to get the correct counts for main dossiers and tasks.

For https://4teamwork.atlassian.net/browse/GEVER-456, https://4teamwork.atlassian.net/browse/GEVER-91 and https://4teamwork.atlassian.net/browse/GEVER-216

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/public/dev-manual/api/listing_stats.rst
+++ b/docs/public/dev-manual/api/listing_stats.rst
@@ -3,7 +3,7 @@
 Statistik zu Auflistungen
 =========================
 
-Mit dem ``@listing-stats`` Endpoint können Statistiken zu Auflistungen unter :ref:`listings` abgefragt werden.
+Mit dem ``@listing-stats`` Endpoint können Statistiken zu Auflistungen unter :ref:`listings` abgefragt werden. Spezifische Statistiken können mit dem ``queries`` Parameter abgefragt werden.
 
 
 **Beispiel-Request**:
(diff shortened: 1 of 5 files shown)


opengever/api/listing_stats.py
```diff
@@ -116,8 +130,29 @@ def _fetch_solr_stats(self, pivot):
...
+        if query.startswith("depth:"):
+            try:
+                depth = int(query.rsplit(":", 1)[1])
+            except (ValueError, IndexError):
+                raise BadRequest("Could not parse depth query: {}".format(query))
+
+            context_depth = get_path_depth(self.context)
+            max_path_depth = context_depth + depth
+            return "{{!tag=q1}}path_depth:[* TO {}]".format(max_path_depth)
+        return "{{!tag=q1}}{}".format(query)
```
deiferni, line 150: i think you should at least use `escape` from `ftw.solr.query` to prevent evil things when using strings from the request

docs/public/dev-manual/api/listing_stats.rst
```diff
@@ -3,7 +3,7 @@
 Statistik zu Auflistungen
 =========================
 
-Mit dem ``@listing-stats`` Endpoint können Statistiken zu Auflistungen unter :ref:`listings` abgefragt werden.
+Mit dem ``@listing-stats`` Endpoint können Statistiken zu Auflistungen unter :ref:`listings` abgefragt werden. Spezifische statistiken können mit dem ``queries`` Parameter abgefragt werden.
```
tinagerber, line 6:
```suggestion
Mit dem ``@listing-stats`` Endpoint können Statistiken zu Auflistungen unter :ref:`listings` abgefragt werden. Spezifische Statistiken können mit dem ``queries`` Parameter abgefragt werden.

File: opengever/api/listing_stats.py (hunk 1 above)
Comment (njohner) [reply opengever/api/listing_stats.py:150]: True, thanks!

**Comment by njohner** [comment]
All done

File: opengever/api/listing_stats.py [hunk 3]
```diff
@@ -36,8 +39,18 @@ class ListingStats(object):
...
+        queries = self.request.form.get("queries", [])
+        if isinstance(queries, basestring):
+            queries = [queries]
+        self.facet_queries = [self._escape_query(query) for query in queries]
+
         self.solr = getUtility(ISolrSearch)
 
+    @staticmethod
+    def _escape_query(query):
+        return ":".join(escape(safe_unicode(el)) for el in query.split(":"))
```
deiferni, line 52: i think you are potentially mixing unicode and non-unicode strings here. i think you should use `u":".join(...)`. would it make sense to amend a test wit non-ascii characters to make sure we are not running into encoding issues?

njohner: Thanks. I added a specific test for non-ascii characters. It was also a problem in the `listing` endpoint, so I also fixed it there.

PR #6488, opened 2020-06-08
Don't send mails with From-Addresses other than our own, and set Reply-To header

lukasgraf:
- Don't send mails with `From:` addresses other than our own
- Instead set a `Reply-To:` header containing the human actor's address
- Make this new behavior the default, but configurable

This avoids sending activity mails with `From:` addresses of OGDS users, which would effectively be spoofing, and instead uses the Plone site's sender address as configured in the mail settings.

In addition, a `Reply-To:` header with the user's address is set instead for mails that are triggered based on the action of a human actor (in the hope that out-of-office replies will be sent there, still reaching the human actor).

In contrast to what's been outlined in the story / done before in #6334, I chose to still use the **user's full name** in the `From:` header (just like GitHub does for example), and just not use their email address. From what I can tell, this shouldn't be a problem with DKIM / DMARC.

So, mails triggered by a human actor will contain headers like this:
```
From: Peter Muster <noreply@gever.example.org>
Reply-To: Peter Muster <peter@muster.example.com>
```

This new behavior is the default, but setting the registry flag `IOGMail.send_with_actor_from_address` to `True` the legacy behavior can be enabled, where the actor's email address is directly used in the `From:` header.

---

Unfortunately, the `Reply-To:` header might not be used for out-of-office replies from what I can tell. [RFC 3834](https://tools.ietf.org/html/rfc3834) specifies behavior for auto-responders, and it says

> Use of the From field as the destination for automatic responses has
> some of the same problems as use of Reply-To.  In particular, the
> From field may list multiple addresses, while automatic responses
> should only be sent to a single address.  In general, the From and
> Reply-To addresses are used in a variety of ways according to
> differing circumstances, and for this reason Personal or Group
> Responders **cannot reliably assume that an address in the From or
> Reply-To field is an appropriate destination for the response.  For
> these reasons the From field SHOULD NOT be used as a destination for
> automatic responses.**

The recommendation seems to be for auto-responders to base these sorts of bounces on the Return-Path (so the MAIL FROM from the SMTP envelope, usually), and I've seen examples where this is the behavior seen in the wild ([MS Exchange example](https://stackoverflow.com/questions/3178429/out-of-office-replies-are-sent-to-from-address-not-reply-to))

For this reason, I made the behavior configurable, so we can enable the legacy behavior when desired by the customer.

Jira: https://4teamwork.atlassian.net/browse/GEVER-171

## Checklist (Must have)

- [x] Changelog entry

## Checklist (optional)

### DIFF
--- a/opengever/activity/mailer.py
+++ b/opengever/activity/mailer.py
@@ -85,14 +85,20 @@ class Mailer(object):
             data = {}
 
         msg = MIMEMultipart('related')
+
         actor = ogds_service().fetch_user(from_userid) if from_userid else None
+        noreply_gever_address = api.portal.get().email_from_address
 
         if actor:
+            # Set From: header to full name of actor, but 'noreply' address
+            # of the GEVER deployment. Sending mails with the From-address of
+            # the actor would lead to them getting rejected in modern mail
+            # setup, e.g. when DKIM is being used.
             msg['From'] = make_addr_header(actor.fullname(),
-                                           actor.email, 'utf-8')
+                                           noreply_gever_address, 'utf-8')
         else:
             msg['From'] = make_addr_header(
-                self.default_addr_header, api.portal.get().email_from_address, 'utf-8')
+                self.default_addr_header, noreply_gever_address, 'utf-8')
 
         if to_userid:
             to_email = ogds_service().fetch_user(to_userid).email
--- a/opengever/activity/mailer.py
+++ b/opengever/activity/mailer.py
@@ -2,6 +2,7 @@ from email.header import Header
 from email.mime.multipart import MIMEMultipart
 from email.mime.text import MIMEText
 from opengever.activity.error_handling import NotificationErrorHandler
+from opengever.base.interfaces import IOGMailSettings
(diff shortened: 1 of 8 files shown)


opengever/activity/mailer.py
```diff
@@ -85,14 +85,25 @@ def prepare_mail(self, subject=u'', to_userid=None, to_email=None,
...
         if actor:
+            # Set From: header to full name of actor, but 'noreply' address
+            # of the GEVER deployment. Sending mails with the From-address of
+            # the actor would lead to them getting rejected in modern mail
+            # setup, e.g. when DKIM is being used.
             msg['From'] = make_addr_header(actor.fullname(),
-                                           actor.email, 'utf-8')
+                                           noreply_gever_address, 'utf-8')
+
+            # Set the Reply-To header to the actual user's address, in order
```
buchi, line 100: IMO this comment is confusing. The Reply-To header should be used this way independent of any autoresponder. Further in your PR description you say that autoresponders send responses to the return path and not the Reply-To address.

PR #6492, opened 2020-06-09
Fix and polish new info at feature.

njohner:
This is a followup PR to https://github.com/4teamwork/opengever.core/pull/6475. In this PR we fix a bug and add some polish according to feedback from the PO.

**Bugfix:**
- We fix a bug with groups selected in the `informed_principals` field. Groups were not handled correctly in the activities. Specifically we had to:
  - Add `OGDSGroupActor` class.
  - Make sure that `Watcher.get_user_ids` returns the representatives for a group, so that all representatives of the group get notified when a task is added

**Polish:**
- Reorder fields in task add form.
- Show informed principals in task added history.
- Add watcher role to task added mail notification settings, so that by default users get notified by mail of a task creation when they are in the `informed_principals`. For this upgrade, we only add the `WATCHER_ROLE` if the user hadn't changed the setting for `TASK_RESPONSIBLE_ROLE`. If the user does not want to be informed as `TASK_RESPONSIBLE_ROLE, he probably does not want to be informed as `WATCHER_ROLE` either...

**task added history when `informed_principals` is set**
Screenshot 2020-06-09 at 11 51 50

For https://4teamwork.atlassian.net/browse/GEVER-31
Frontend PR: https://github.com/4teamwork/gever-ui/pull/1129

## Checklist (Must have)

- [x] Changelog entry: No changelog necessary, in the same release as original implementation https://github.com/4teamwork/opengever.core/pull/6475
- [x] Documentation updated (notably for API and deployment).

## Checklist (optional)

- DB-Schema migration
- New translations

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.3.0rc5 (unreleased)
 ------------------------
 
+- Add OGDSGroupActor class. [njohner]
 - Set Reply-To header from mails sent on behalf of users. [lgraf]
 - Avoid sending mails with From-Addresses other than our own. [lgraf]
 - Fix bug with setting issuer and informed_principals on forwardings. [njohner]
--- a/opengever/activity/model/watcher.py
+++ b/opengever/activity/model/watcher.py
@@ -34,7 +34,7 @@ class Watcher(Base):
         representatives = [user.userid for user in
                            actor_lookup.lookup().representatives()]
 
-        if actor_lookup.is_inbox() or actor_lookup.is_team():
+        if actor_lookup.is_inbox():
             return [userid for userid in representatives
                     if UserSettings.get_setting_for_user(
                         userid, 'notify_inbox_actions')]
--- a/opengever/activity/model/watcher.py
+++ b/opengever/activity/model/watcher.py
@@ -1,6 +1,5 @@
 from opengever.base.model import Base
 from opengever.base.model import USER_ID_LENGTH
-from opengever.ogds.base.actor import Actor
 from opengever.ogds.base.actor import ActorLookup
 from opengever.ogds.models.user_settings import UserSettings
 from sqlalchemy import Column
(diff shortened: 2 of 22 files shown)


opengever/ogds/base/actor.py
```diff
@@ -332,6 +341,29 @@ def representatives(self):
         return [self.user]
 
 
+class OGDSGroupActor(Actor):
+
+    def __init__(self, identifier, group=None):
+        super(OGDSGroupActor, self).__init__(identifier)
+        self.group = group
+
+    def corresponds_to(self, group):
```
lukasgraf, line 350: `corresponds_to` is implemented differently for the other group-like actors (`TeamActor` and `InboxActor`), where it checks for containment. Not saying that makes much sense, given the method name `corresponds_to` , I just noticed its different.

docs/HISTORY.txt
```diff
@@ -4,6 +4,7 @@ Changelog
 2020.3.0rc5 (unreleased)
 ------------------------
 
+- Add OGDSGroupActor class. [njohner]
```
lukasgraf, line 7: Could you please add some tests for the `OGDSGroupActor` to [`test_actor.py`](https://github.com/4teamwork/opengever.core/blob/master/opengever/ogds/base/tests/test_actor.py) ? Should be pretty straight forward.

opengever/activity/model/watcher.py
```diff
@@ -34,7 +34,7 @@ def get_user_ids(self):
         representatives = [user.userid for user in
                            actor_lookup.lookup().representatives()]
 
-        if actor_lookup.is_inbox() or actor_lookup.is_team():
+        if actor_lookup.is_inbox():
```
lukasgraf, line 37: Just to make sure I understand: We previously also considered inbox notification settings for team members, when deciding whether they should get the notification or not. Which was wrong, that setting should only be considered for inbox members. So team members now get the notification unconditionally. Correct?

njohner: Yes I think that was simply wrong. For example when you create a team task, all team members will unconditionally be informed (according to their settings as TASK_RESPONSIBLE of course). @phgross what do you think?

opengever/ogds/base/actor.py
njohner: right thanks

docs/HISTORY.txt
njohner: Yes sorry!

njohner:
Ok should be all done here. I'll wait for feedback from philippe and then I'll merge

PR #6495, opened 2020-06-09
Mba/505 delete action

mbaechtold:
Belongs to the Jira issue https://4teamwork.atlassian.net/browse/GEVER-91

image

## Checklist (Must have)

- [x] Changelog entry

## Checklist (optional)

- [ ] New functionality  for `document` also works for `mail`
- [ ] New functionality  for `task` also works for `forwarding`
- DB-Schema migration
- New feature flag:
  - [ ] Tests with activated and deactivated feature
- New translations
  - [ ] Correct i18n-domain was used (Copy-Paste errors are common here)


opengever/base/browser/is_delete_available.py
```diff
@@ -30,5 +32,7 @@ def __call__(self):
             any(interface.providedBy(self.context) for interface in self.authorized_interfaces),
             # Whitelisted by interface provided by the acquired parent
             any(interface.providedBy(parent) for interface in self.authorized_parent_interfaces),
+            # Document templates can be deleted.
+            ITemplateFolder.providedBy(parent) and IBaseDocument.providedBy(self.context)
```
njohner, line 36: Actually I think you could just add `ITemplateFolder` to the `authorized_parent_interfaces`. You could also add `ITemplateDossier`. It seems everything in the template area can be deleted

njohner:
I hate delete buttons, we have too many of them. For some reason for `MeetingTemplate` we actually show the `folder_delete` button instead of the `folder_delete_confirmation`...

opengever/base/browser/is_delete_available.py
mbaechtold: I have changed it. Thanks for sharing your background knowledge.

mbaechtold:
I have rephrased the intent of the pull request: this change is required for the new GEVER UI to be able to show the delete action on templates. As a side effect, the delete action is shown on the legacy UI too.

mbaechtold:
@njohner, do you want to do the code review or should I ask in Slack so somebody else can have a look?

docs/HISTORY.txt
```diff
@@ -4,8 +4,13 @@ Changelog
 2020.3.0rc5 (unreleased)
 ------------------------
 
+- Make available the delete action for templates. [mbaechtold]
 - Set Reply-To header from mails sent on behalf of users. [lgraf]
 - Avoid sending mails with From-Addresses other than our own. [lgraf]
+- Add special handling for signed/multipart message attachments. [deiferni]
```
njohner, line 10: Careful with the rebase of the backport...

mbaechtold: Thanks. This is not the first time it happened. I need to check better after rebasing.

njohner: It's always when we make a backport. It's a bit of a pain. Worst case it wouldn't be the first time we merge a PR with changelog screwed up 😄

PR #6496, opened 2020-06-10
Always use configured solr port

2e12:
In case 'SOLR_PORT' is already set in env it can't be overwritten by the configured port and in order to fix this setdefault() is replaced.

**Background**
For some reason I have defined a Solr port as env variable global. This had the consequence that my Solr did not work in the tests. In an exhausting debugging session we found out that "setdefault" was used to set the correct port. But here the value is only set if the value is not yet available. This had the consequence that a completely wrong port was used.

Was solved as part of https://4teamwork.atlassian.net/browse/GEVER-462.

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -20,6 +20,7 @@ Changelog
 - Enable API endpoint `@document-from-template` for tasks. [mbaechtold]
 - Support combined notation for task responsible in workflow transitions. [elioschmutz]
 - Bump docxcompose to 1.1.2 to fix issues with external image references and drawing properties. [buchi]
+- Always use configured solr port in tests. [2e12]
 
 
 2020.3.0rc4 (2020-06-05)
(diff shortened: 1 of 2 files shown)


deiferni:
so for the record i quickly double-checked the chnages with @jone and @siegy22, they confirmed my suspicion. @2e12 please get back to me on monday so we can finish and 🚢 .

docs/HISTORY.txt
```diff
@@ -20,7 +20,7 @@ Changelog
 - Enable API endpoint `@document-from-template` for tasks. [mbaechtold]
 - Support combined notation for task responsible in workflow transitions. [elioschmutz]
 - Bump docxcompose to 1.1.2 to fix issues with external image references and drawing properties. [buchi]
-
+- Always use configured solr port in tests. [2e12]
```
deiferni, line 23: it looks like you removed a newline here. we always have two newlines at the end of a section.

PR #6497, opened 2020-06-10
Define a set of columns that get synchronized in user and group model and remove import_stamp

tinagerber:
This is a followup PR to #6477 

Currently the last_login field is overwritten with None every time the LDAP is synchronized, because the field is not mapped to LDAP.

This has been corrected by the OGDSUpdater no longer synchronizing all fields of the user models, but only those that are in a set (column_names_to_sync). The same applies to the group model.

Furthermore the column import_stamp of the user model was deleted, because it is not used anymore.

Jira: https://4teamwork.atlassian.net/browse/GEVER-39

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/ogds/base/tests/test_ogds_updater.py
+++ b/opengever/ogds/base/tests/test_ogds_updater.py
@@ -5,12 +5,16 @@ from opengever.ogds.base.interfaces import IOGDSUpdater
 from opengever.ogds.base.tests.ldaphelpers import FakeLDAPPlugin
 from opengever.ogds.base.tests.ldaphelpers import FakeLDAPSearchUtility
 from opengever.ogds.base.tests.ldaphelpers import FakeLDAPUserFolder
+from opengever.ogds.models.group import Group
 from opengever.ogds.models.service import ogds_service
+from opengever.ogds.models.user import User
 from opengever.testing import FunctionalTestCase
 from plone import api
 
 
 FAKE_LDAP_USERFOLDER = FakeLDAPUserFolder()
+BLACKLISTED_USER_COLUMNS = {'userid', 'last_login'}
+BLACKLISTED_GROUP_COLUMNS = set()
 
 
 class TestOGDSUpdater(FunctionalTestCase):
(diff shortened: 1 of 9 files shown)


opengever/ogds/base/tests/test_ogds_updater.py
```diff
@@ -50,6 +54,15 @@ def test_flags_users_not_present_in_ldap_as_inactive(self):
         self.assertTrue(ogds_service().fetch_user('hugo.boss').active)
         self.assertTrue(ogds_service().fetch_user('new.user').active)
 
+    def test_unmapped_user_columns(self):
+        unmapped = []
+        for column in User.__table__.columns:
+            if column.name not in BLACKLISTED_USER_COLUMNS.union(User.column_names_to_sync):
+                unmapped.append(column.name)
+        self.assertEqual([], unmapped,
+                         'The columns are neither in the column_names_to_sync list of '
```
njohner, line 63: I'd go for something more explicit here, maybe `A new column was added to the User model, it should either be added to the column_names_to_sync list of the user model if it needs syncing with the LDAP, or in the BLACKLISTED_USER_COLUMNS list here.`

opengever/ogds/base/tests/test_ogds_updater.py
```diff
@@ -59,6 +72,15 @@ def test_imports_groups(self):
         updater.import_groups()
         self.assertIsNotNone(ogds_service().fetch_group('og_mandant1_users'))
 
+    def test_unmapped_group_columns(self):
+        unmapped = []
+        for column in Group.__table__.columns:
+            if column.name not in BLACKLISTED_GROUP_COLUMNS.union(Group.column_names_to_sync):
+                unmapped.append(column.name)
+        self.assertEqual([], unmapped,
```
njohner, line 80: Same here

opengever/ogds/base/tests/test_ogds_updater.py
```diff
@@ -50,6 +54,15 @@ def test_flags_users_not_present_in_ldap_as_inactive(self):
...
         self.assertTrue(ogds_service().fetch_user('new.user').active)
 
+    def test_unmapped_user_columns(self):
+        unmapped = []
+        for column in User.__table__.columns:
+            if column.name not in BLACKLISTED_USER_COLUMNS.union(User.column_names_to_sync):
+                unmapped.append(column.name)
+        self.assertEqual([], unmapped,
+                         'The columns are neither in the column_names_to_sync list of '
+                         'the user model nor in the BLACKLISTED_USER_COLUMNS list.')
```
njohner, line 64: To be more complete you could make sure that:
- `column_names_to_sync ` and `BLACKLISTED_USER_COLUMNS` are disjoint
- union of `column_names_to_sync ` and `BLACKLISTED_USER_COLUMNS` is equal to `User.__table__.columns`
So that you can also make sure that we did not remove a column but forgot to remove it from the `column_names_to_sync ` or `BLACKLISTED_USER_COLUMNS`

So basically test that:
- `assertEqual(set(), set(column_names_to_sync).intersection(BLACKLISTED_USER_COLUMNS))`
- `assertEqual(set(), set(User.__table__.columns).difference(set(column_names_to_sync).union(BLACKLISTED_USER_COLUMNS)), "a new column was added in the user model...."`
- `assertEqual(set(), set(column_names_to_sync).union(BLACKLISTED_USER_COLUMNS).difference(set(User.__table__.columns)), "a column was deleted from user model but is still in ...."`

PR #6499, opened 2020-06-10
Add minimalistic log handler for sentry.

deiferni:
Explicitly log to sentry for two `ftw.solr` modules we want to monitor well at the moment.

The package `ftw.solr` currently only logs potentially critical issues in https://github.com/4teamwork/ftw.solr/blob/6f1db34678b4271e3b2640b85e4f11b6c8795fa3/ftw/solr/connection.py and https://github.com/4teamwork/ftw.solr/blob/6f1db34678b4271e3b2640b85e4f11b6c8795fa3/ftw/solr/handlers.py. With our currenty sentry integration only exceptions end up in sentry.

With this PR we add a very simple log-handler and specifically register it for the two modules. It also logs the messages (for certain levels) to sentry. Some examples:
- https://sentry.4teamwork.ch/sentry/onegov-gever/issues/68017
- https://sentry.4teamwork.ch/sentry/onegov-gever/issues/67593
- https://sentry.4teamwork.ch/sentry/onegov-gever/issues/67556
- https://sentry.4teamwork.ch/sentry/onegov-gever/issues/68016 _(for this one i hacked `ftw.solr` to also consider 200 repsones non-ok. wont' appear in production like this, of course)_

I have decided to write my first suggestion based on a log-handler integration as that way we do not have to modify `ftw.solr` yet.

This can be considered a stopgap measure for the moment as we are moving many things to solr and thus want to be specially aware of certain problems we might encounter. Not intended to stay around for long but rather a quick way to get logging to sentry for potentially critical solr errors. Should go away again once we have proper logging for non-exceptions to sentry.

I have not found a good way to easily write a test for this functionality. Input welcome. I have decided not to invest much time into maybe being able to also log a stack trace. If we switch to another sentry integration soon that will come for free. For now it provides awareness of certain issues which should hopefully be good enough for now.

For https://4teamwork.atlassian.net/browse/GEVER-182

Todo:
- [ ] delete sentry errors again once the PR is merged

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/base/sentry.py
+++ b/opengever/base/sentry.py
@@ -29,6 +29,39 @@ else:
 log = logging.getLogger('opengever.base.sentry')
 
 
+class SentryHandler(logging.Handler):
+    """Minimalistic log handler to also log records to sentry.
+
+    For simplicity contains a hard-coded mapping of log-levels we are
+    currently interested in.
+
+    Not intented to stay around for long but rather a quick way to get logging
+    to sentry for potentially critical solr errors *now*. Should go away again
+    once we have proper logging for non-exceptions to sentry.
+    """
+
+    logging_level_to_sentry_level = {
+        logging.ERROR: 'error',
+        logging.WARNING: 'warning',
+        logging.CRITICAL: 'critical',
+        logging.FATAL: 'fatal',
+    }
+
+    def emit(self, record):
+        level = self.logging_level_to_sentry_level.get(record.levelno)
+        if not level:
+            return
+
+        message = record.getMessage()
+        log_msg_to_sentry(message, level=level)
+
+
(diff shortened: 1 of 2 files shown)


opengever/base/sentry.py
```diff
@@ -29,6 +29,39 @@ def maybe_report_exception(*args, **kwargs):
 log = logging.getLogger('opengever.base.sentry')
 
 
+class SentryHandler(logging.Handler):
+    """Minimalistic log handler to also log records to sentry.
+
+    For simplicity contains a hard-coded mapping of log-levels we are
+    currently interested in.
+
+    Not intented to stay around for long but rahter a quick way to get logging
```
njohner, line 38: rahter -> rather

PR #6502, opened 2020-06-12
Add new endpoints @assigned-users and @transfer-task

tinagerber:
The possibility should be created to transfer all tasks of one user to another. This means all tasks where the user is the responsible and/or the issuer.

This required some backend work:
- The new endpoint `@assigned-users` provides a vocabulary of all users of the current admin unit. The frontend needs this list to get all possible users to which the task could be transferred.
- To transfer a task, the new endpoint `@transfer-task` is required. The endpoint is protected by a new permission `opengever.api.TransferAssignment`, which allows only administrators and managers to use the endpoint.
- A new header `X-GEVER-SuppressNotifications` is introduced, with which it is possible to suppress the triggering of a notification.

Currently, only local tasks can be transferred, not cross-client tasks.

Jira: https://4teamwork.atlassian.net/browse/GEVER-35

## Checklist (Must have)

- [x] Changelog entry

## Checklist (optional)

- New translations

### DIFF
--- a/opengever/api/transfer.py
+++ b/opengever/api/transfer.py
@@ -1,0 +1,89 @@
+from opengever.inbox.forwarding import IForwarding
+from opengever.ogds.models.user import User
+from opengever.task.activities import TaskChangeIssuerActivity
+from opengever.task.activities import TaskReassignActivity
+from opengever.task.interfaces import ISuccessorTaskController
+from opengever.task.localroles import LocalRolesSetter
+from opengever.task.task import ITask
+from opengever.task.util import add_simple_response
+from plone.protect.interfaces import IDisableCSRFProtection
+from plone.restapi.deserializer import json_body
+from plone.restapi.services import Service
+from zExceptions import BadRequest
+from zope.event import notify
+from zope.interface import alsoProvides
+from zope.lifecycleevent import ObjectModifiedEvent
+
+
+class TransferTaskPost(Service):
+
+    def extract_data(self):
+        data = json_body(self.request)
+        self.old_userid = data.get("old_userid")
+        self.new_userid = data.get("new_userid")
+
+        if not self.old_userid:
+            raise BadRequest("Property 'old_userid' is required")
+        if not self.new_userid:
+            raise BadRequest("Property 'new_userid' is required")
+        if not User.query.get_by_userid(self.old_userid):
+            raise BadRequest("userid '{}' does not exist".format(self.old_userid))
+        if not User.query.get_by_userid(self.new_userid):
+            raise BadRequest("userid '{}' does not exist".format(self.new_userid))
+        if self.old_userid == self.new_userid:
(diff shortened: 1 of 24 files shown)


opengever/api/transfer.py
```diff
@@ -0,0 +1,87 @@
...
+from zope.event import notify
+from zope.interface import alsoProvides
+from zope.lifecycleevent import ObjectModifiedEvent
+
+
+class TransferTaskPost(Service):
+
+    def extract_data(self):
+        data = json_body(self.request)
+        self.old_userid = data.get("old_userid", None)
```
lukasgraf, line 22: `None` is implied with `dict.get()` in Python, no need to specify it.

opengever/api/transfer.py
```diff
@@ -0,0 +1,87 @@
...
+        self.new_userid = data.get("new_userid", None)
+
+        if not self.old_userid:
+            raise BadRequest("Property 'old_userid' is required")
+        if not self.new_userid:
+            raise BadRequest("Property 'new_userid' is required")
+        if not api.user.get(self.old_userid):
```
lukasgraf, line 29: This should check for the existence of the User in OGDS, not a Plone User (which will try to look it up in LDAP). Same below.

If it's no too much effort, a test case for this would be nice _(it should be possible to transfer tasks for a user that is not in LDAP (any more), but in OGDS (as inactive), since that's the most common use case)_

opengever/api/transfer.py
```diff
@@ -0,0 +1,87 @@
...
+
+class TransferTaskPost(Service):
+
+    def extract_data(self):
+        data = json_body(self.request)
+        self.old_userid = data.get("old_userid", None)
+        self.new_userid = data.get("new_userid", None)
+
+        if not self.old_userid:
+            raise BadRequest("Property 'old_userid' is required")
```
lukasgraf, line 26: Also raise `BadRequest` if `old_userid == new_userid`? Should be a quick win to prevent rather silly things from happening 😉

opengever/task/activities.py
```diff
@@ -231,6 +234,29 @@ def _is_ignored_transition(self):
...
+class TaskChangeIssuerActivity(TaskTransitionActivity):
+
+    def before_recording(self):
+        change = ResponseDescription.get(response=self.response).get_change('issuer')
+        if not change or change.get('before') == self.context.issuer:
+            return
+
+        self.center.add_task_issuer(self.context, self.context.issuer)
+        self.center.remove_task_issuer(self.context, change.get('before'))
+        self.center.add_watcher_to_resource(
```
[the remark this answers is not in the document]
tinagerber: Well, when I think about it, it doesn't really make sense to add the old issuer, since no notifications are triggered anyway.

tinagerber:
@lukasgraf I updated the PR, can you have another look?

PR #6505, opened 2020-06-15
Allow filtering for all checked out documents.

tinagerber:
We want to be able to filter for all checked out documents. For this we must be able to make a query of this type: `http://localhost:8080/fd/@listing?name=documents&facets:list=checked_out&filters.-checked_out:record=`
Therefore the following two things had to be implemented:
- Ability to filter for empty strings
- Possibility to negate a filter query

Jira: https://4teamwork.atlassian.net/browse/GEVER-41

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/api/solr_query_service.py
+++ b/opengever/api/solr_query_service.py
@@ -196,7 +196,6 @@ DEFAULT_FIELDS = set([
 REQUIRED_RESPONSE_FIELDS = set(['UID'])
 
 FIELDS_WITH_MAPPING = [
-    ListingField('checked_out', 'checked_out', transform=display_name),
     ListingField('bumblebee_checksum', 'bumblebee_checksum', sort_index=DEFAULT_SORT_INDEX),
     ListingField('checked_out', 'checked_out', transform=display_name),
     ListingField('checked_out_fullname', 'checked_out', 'checked_out_fullname'),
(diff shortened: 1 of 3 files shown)


opengever/api/solr_query_service.py
```diff
@@ -196,7 +196,6 @@ def listing_to_solr_filter(self, value):
 REQUIRED_RESPONSE_FIELDS = set(['UID'])
 
 FIELDS_WITH_MAPPING = [
-    ListingField('checked_out', 'checked_out', transform=display_name),
```
lukasgraf, line 199: :trollface:

opengever/api/solr_query_service.py
```diff
@@ -129,7 +129,14 @@ def listing_to_solr_filter(self, value):
             value = u' OR '.join(value)
         else:
             value = escape(safe_unicode(value))
-        return u'{}:({})'.format(escape(self.index), value)
+
+
```
lukasgraf, line 136: I would still add a comment to describe why we escape the solr field name. Something like
```
        # Escaping the Solr field name is done for security reasons
        # (to prevent attempts to circumvent the security filter by injection
        # of a maliciously crafted field name)
```

PR #6510, opened 2020-06-19
Update PR template to include API breaking changes.

njohner:
Following the discussion in https://4teamwork.slack.com/archives/C062G6FGT/p1592497331009400?thread_ts=1592466050.000300&cid=C062G6FGT, `api-change` label is not enough to track breaking changes made to the API. I propose to add a new checkbox in the PR checklist, so that we at least inform the SM.

I also took the liberty to clean the PR checklist up a bit, so that it does not become longer than it already is...

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -16,27 +16,29 @@ Definition of Done: https://4teamwork.atlassian.net/wiki/spaces/CHX4TW/pages/917
 _Everything has to be done/checked. Checked but not present means the author deemed it unnecessary._
 
 - [ ] Changelog entry
-- [ ] Documentation updated (notably for API and deployment)
 - [ ] Link to issue (Jira or GitHub) and backlink in issue (Jira)
 
 
-## Checklist (optional)
+## Checklist (if applicable)
 
 _Only applicable should be left and checked._
 
-- [ ] New functionality  for `document` also works for `mail`
-- [ ] New functionality  for `task` also works for `forwarding`
+- API change:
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -16,27 +16,27 @@ Definition of Done: https://4teamwork.atlassian.net/wiki/spaces/CHX4TW/pages/917
...
 
 ## Checklist (optional)
 
 _Only applicable should be left and checked._
 
-- [ ] New functionality  for `document` also works for `mail`
-- [ ] New functionality  for `task` also works for `forwarding`
+- API change:
+  - [ ] Documentation is updated
+  - [ ] If breaking, Scrum master is informed
```
lukasgraf, line 28:
```
- [ ] If breaking:
    - [ ] api-change label added
```

lukasgraf: @njohner actually, I just realized that like this the API Change related points get moved from the mandatory to the optional section... 🤔

njohner: The optional part is not optional. It's `if applicable`. Because for example, introducing a new functionality for documents that breaks E-mails is not an option either 😄 .
So maybe we should just rename the section...

lukasgraf: 👍 for renaming to something along the lines of "if applicable". However, then I would propose to move the API changes stuff to the very top of that section, so it doesn't get missed that easily.

PR #6511, opened 2020-06-19
Also normalise filename of original_message when present.

tinagerber:
When an e-mail with the file extension "msg" is uploaded, two blob files are created, one with the original file, one that has been converted to "eml".

If the file name of an email has been changed, only the file name of the converted message has been changed, but not the file name of the original message. 

However, when the email is added as a favorite, the filename of the original message is written to the database.
This can cause errors because the filename of the original message was not truncated.

Now the filename of the original message is also truncated when the filename of an email changes, this solves the problem.

Jira: https://4teamwork.atlassian.net/browse/GEVER-477

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -10,6 +10,7 @@ Changelog
 - Extend the @favorites endpoint to let it return already resolved favorites. [elioschmutz]
 - Use correct response type for proposal comment responses. [njohner]
 - Add expandable endpoint @tasktree for getting task hierarchy. [buchi]
+- Also normalise filename of original_message when present. [tinagerber]
 
 
 2020.3.0 (2020-06-18)
--- a/opengever/core/upgrades/20200507073857_update_filename_of_mails_with_file_extension_msg/upgrade.py
+++ b/opengever/core/upgrades/20200507073857_update_filename_of_mails_with_file_extension_msg/upgrade.py
@@ -1,0 +1,13 @@
+from ftw.upgrade import UpgradeStep
+
+
+class UpdateFilenameOfMailsWithFileExtensionMsg(UpgradeStep):
+    """Update filename of mails with file extension msg.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
+        query = {'portal_type': 'ftw.mail.mail', 'file_extension': '.msg'}
+        for mail in self.objects(query, 'Update filename'):
+            mail.update_filename()
+            mail.reindexObject(idxs=['UID', 'filename'])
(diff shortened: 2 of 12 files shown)


opengever/mail/mail.py
```diff
@@ -338,6 +338,9 @@ def update_filename(self):
 
         new_filename = u'{}{}'.format(normalized_subject, ext)
         if self.message.filename != new_filename:
+            if self.original_message:
+                ext = u'.msg'
```
njohner, line 342: I'd use `ext = os.path.splitext(self. original_message.filename)[-1]` here

docs/HISTORY.txt
```diff
@@ -4,7 +4,7 @@ Changelog
 2020.4.0 (unreleased)
 ---------------------
 
-- Nothing changed yet.
+- Fix update filename for emails. [tinagerber]
```
njohner, line 7: I think the CL and commit message / title could be a bit clearer. Maybe something like:
`- Also normalise filename of original_message when present. [tinagerber]`
And for the commit:
title: Also normalise filename of original_message when present.
message: When updating filename of an email, we should update both the filename of message and of the original message if it exists.

tinagerber:
@njohner I wrote a test, created an upgrade step and implemented the other change requests. I have also backdated the upgrade step so that it is performed before the schema migration of the favorites. Can you rereview?

opengever/core/upgrades/20200507073857_update_filename_of_mails_with_file_extension_msg/upgrade.py
```diff
@@ -0,0 +1,13 @@
...
+class UpdateFilenameOfMailsWithFileExtensionMsg(UpgradeStep):
+    """Update filename of mails with file extension msg.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
+        query = {'portal_type': 'ftw.mail.mail', 'file_extension': '.msg'}
+        for mail in self.objects(query, 'Update filename'):
+            mail.update_filename()
+            mail.reindexObject(idxs=['UID'])
```
njohner, line 13: `filename` is available in the matadata and in solr, so you need to make sure these two are updated. As it is, only the metadata gets updated, but not solr. So you have to explicitely add `filename` in `idxs`, i.e.
`idxs=['UID', 'filename']`

njohner:
Ok no backport needed. Instead we will ship a fix of https://github.com/4teamwork/opengever.core/blob/master/opengever/core/upgrades/20200508163417_populate_filename_column_in_favorites/upgrade.py and backport that.

tinagerber:
@njohner Should I reverse the backdating or leave it at that?

PR #6512, opened 2020-06-19
Include task tree in task serialization

buchi:
Used to display a task hierarchy in the frontend if a task has subtasks.
See: https://github.com/4teamwork/gever-ui/pull/1158

JIRA: https://4teamwork.atlassian.net/browse/GEVER-419

## Checklist (Must have)

- [x] Changelog entry

## Checklist (optional)

### DIFF
--- a/opengever/api/task.py
+++ b/opengever/api/task.py
@@ -1,6 +1,11 @@
+from Acquisition import aq_parent
+from ftw.solr.interfaces import ISolrSearch
+from ftw.solr.query import make_filters
 from opengever.api.response import ResponsePost
 from opengever.api.response import SerializeResponseToJson
 from opengever.api.serializer import GeverSerializeFolderToJson
+from opengever.base.browser.navigation import make_tree_by_url
+from opengever.base.solr import OGSolrContentListing
 from opengever.ogds.base.actor import ActorLookup
 from opengever.ogds.models.team import Team
 from opengever.task.interfaces import ICommentResponseHandler
@@ -14,6 +19,7 @@ from plone.restapi.interfaces import ISerializeToJsonSummary
 from zExceptions import Unauthorized
 from zope.component import adapter
 from zope.component import getMultiAdapter
+from zope.component import getUtility
 from zope.interface import implementer
 from zope.interface import Interface
 
@@ -25,6 +31,7 @@ class SerializeTaskToJson(GeverSerializeFolderToJson):
     def __call__(self, *args, **kwargs):
         result = super(SerializeTaskToJson, self).__call__(*args, **kwargs)
         result[u'containing_dossier'] = self._get_containing_dossier_summary()
+        result[u'task_tree'] = self.task_tree()
         return result
 
     def _get_containing_dossier_summary(self):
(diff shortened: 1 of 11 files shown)


opengever/api/task.py
```diff
@@ -25,6 +31,7 @@ class SerializeTaskToJson(GeverSerializeFolderToJson):
     def __call__(self, *args, **kwargs):
         result = super(SerializeTaskToJson, self).__call__(*args, **kwargs)
         result[u'containing_dossier'] = self._get_containing_dossier_summary()
+        result[u'task_tree'] = self.task_tree()
```
deiferni, line 34: have you considered using expansion https://plonerestapi.readthedocs.io/en/latest/expansion.html here? 
we are doing so for e.g. dossier hierarchy as discussed in https://4teamwork.atlassian.net/browse/GEVER-555. that way the UI can request the tree when it is needed and otherwise skip the potentially expensive? serialization.

opengever/api/task.py
```diff
@@ -35,6 +42,36 @@ def _get_containing_dossier_summary(self):
...
+        filters = make_filters(
+            path={
+                'query': '/'.join(main_task.getPhysicalPath()),
+                'depth': -1,
+            },
+            object_provides=ITask.__identifier__,
+        )
+        fieldlist = ['Title', 'portal_type', 'path', 'review_state']
+        resp = solr.search(
+            filters=filters, start=0, rows=1000, sort='created asc',
```
deiferni, line 62: can we avoid hardcoding the maximum rows?

lukasgraf: I understand the sentiment / agree in theory, but there's really no good alternative here I think. Having the frontend deal with paginated responses, just to find out that the DOM becomes unusably slow and impossible to navigate when 1000+ items are rendered doesn't really improve the situation.

deiferni: yeah i was fearing as much, but still i had to ask. what worries me is that we silently hide these tasks. admittedly it may be unlikely to have so many subtasks, though. can we create a follow-up to make sure this is addressed. maybe a hint for the user that there are too many subtasks to display would suffice?

opengever/api/task.py
buchi: Yes. Currently the the handling of expansion in the frontend is somewhat inflexible as we don't know the type before fetching the context and need to add the expansion parameters before that. But I agree that from the API point of view, it's better to use an expansion here and will amend that.

opengever/api/task.py
buchi: I agree that it's not a terribly beautiful solution. But for me this falls into the category 'good enough'. I really do not expect that we will ever have a case that hits this limit. Thus for me handling this, is a waste of time. @deiferni if you disagree, let's make a call to discuss it further.

deiferni: no, i was just wondering/asking if there was an easy way around the hardcoded limit. but it seems to be difficult and deemed unnecessary.
"good enough" works for for me, no further objections 👍

buchi:
Added docs for `@tasktree` endpoint.

PR #6513, opened 2020-06-19
Fix sort order within task template folder.

mbaechtold:
Until now, reordering the task templates within a task template folder looked to the user like it was not working. Although the items can be ordered with drag'n'drop, the ordering was lost after a browser refresh. 

This was due to the fact that Solr does not have an index for `getObjPositionInParent`. We can fix this by sorting the items by their position in the parent, just like Plone would do.

Belongs to Jira issue https://4teamwork.atlassian.net/browse/GEVER-565

image

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.4.0 (unreleased)
 ---------------------
 
+- Fix sort order within task template folder. [mbaechtold]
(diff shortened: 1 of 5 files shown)


mbaechtold:
Thanks for your comment, Lukas.

Since @buchi was involved while analyzing the problem and finding a solution, I'd like to have his opinion on this one. What do you think?

mbaechtold:
Danke vielmals für die Mithilfe und das Code Review!

PR #6514, opened 2020-06-22
Add API endpoint `@trigger-task-template` to create tasks in a dossier from a template.

deiferni:
The endpoint `@trigger-task-template` allows the user to create tasks from task-templates. The tasks are created in a dossier.

The main goal of this PR is to expose triggering task templates via API so we can re-implement the classic UI wizard in the new ui. We won't implement it as a wizard though so all data can be passed to the endpoint at once.

Challenges i had:
- `plone.restapi` does not help me validate arbitrary schema based input, as far as i could see i have to do some manual validation. if someone has a better idea about how this should be done please let me know.
- i had to split the schema in two, one part is "automatically" validated via `get_validation_errors`, the other part is only used for its sources and for manual validation.
- the vocabulary returning selectable template folders is not so useful for the UI as it only returns UIDs as tokens, we might wanna swap that out against a listing/solr query thing eventually. we have left it in for now.
- the existing wizard and the task template vocabulary are strongly coupled (the vocabulary only works inside of the wizard) so i have decided against reusing it for the moment.
- some of the tests are copied from existing functionality and adapted to the endpoint. the method names are not the best but i'd like to keep them so we can easily cross-reference in the future.

Jira: https://4teamwork.atlassian.net/browse/GEVER-101

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/public/dev-manual/api/trigger_task_template.rst
+++ b/docs/public/dev-manual/api/trigger_task_template.rst
@@ -1,0 +1,42 @@
+.. _trigger_task_template:
+
+Standardabläufe auslösen
+========================
+
+In einem Dossier kann mit dem Endpoint ``@trigger-task-template`` ein
+bestehender Standardablauf ausgelöst werden.
+
+Der Endpoint erwartet folgende Parameter:
+
+- ``tasktemplatefolder``: Das zu verwendende Template aus dem Vokabular ``opengever.dossier.DocumentTemplatesVocabulary``
+- ``tasktemplates``: zu verwendende Aufgabenvorlagen und optional Auftragnehmer der zu erstellenden Aufgabe
+- ``related_documents``: Optionale Verweise auf Dokumente/Mails
+- ``start_immediately``: Erste Aufgabe nach Erstellung sofort starten
+
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+        POST /ordnungssystem/fuehrung/dossier-23/@trigger-task-template HTTP/1.1
+        Accept: application/json
+
+        {
+            "tasktemplatefolder": "67a25fc941354568950439f08f7af3ed",
+            "tasktemplates": [
+                {
+                    "@id": "http://localhost:8080/fd/vorlagen/tasktemplatefolder-1/tasktemplate-1",
+                    "responsible": "stv:david.erni"
+                }
+            ],
+            "related_documents": [
+                {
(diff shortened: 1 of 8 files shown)


opengever/api/templatefolder.py
```diff
@@ -188,6 +188,8 @@ def _validate_tasktemplates(self, tasktemplatefolder, data):
         if not data:
             errors.append('At least one tasktemplate is required')
 
+        valid_templates = set(tasktemplatefolder.objectValues())
```
lukasgraf, line 191: This probably destroys BTree lazyness and wakes up all the buckets, even though the containment test below could abort early. Probably completely irrelevant for the amounts of data we're talking here, I'm more just thinking out loud. Wouldn't something like `if template not in tasktemplatefolder` work?

docs/public/dev-manual/api/trigger_task_template.rst
```diff
@@ -0,0 +1,44 @@
...
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+        POST /ordnungssystem/fuehrung/dossier-23/@trigger-task-template HTTP/1.1
+        Accept: application/json
+
+        {
+            "tasktemplatefolder": {
+                "token": "67a25fc941354568950439f08f7af3ed"
```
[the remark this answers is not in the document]
deiferni: ok, adapted. have also added a test for the flat notation.

opengever/api/templatefolder.py
deiferni: `in` directly on the would not work as it would attempt to find `template` in the keys of the `tasktemplatefolder`. i am working with object instances here. i could use `template.getId()` but that would not guard agains duplicate ids within different parents 🤔 .

I would suggest to change to a catalog-based check with `getFolderContents()` and the `UID`.

deiferni:
@lukasgraf pulling you in for a quick re-review. have amended a commit for an issue discovered by @all4manu, would appreciate you to have a quick look.
also have amended changes as discussed in your suggestions.

opengever/api/templatefolder.py
lukasgraf: What about `if template not in tasktemplatefolder.objectValues()`? That should already be a lazy `TreeSet` that supports containment test if I remember correctly. Otherwise I'd just leave it as is, no need to involve the catalog here IMHO.

deiferni: i have already rewritten it 😓

deiferni: ok pushed again 😄

PR #6515, opened 2020-06-23
Extend the @favorites endpoint to let it return already resolved favorites.

elioschmutz:
This PR implements a `resolve` query parameter for the `@favorites` endpoint which will return already resolved favorites if possible.

We need resolved favorites for the new content selector. Navigation and selection of favorite items is not fully provided without this changes.

⚠️ In the spec we defined to only return resolved favorites if we use the `resolve` parameter. But this seems to be wrong and not useful. I introduced a new property on the serialized favorite instead to indicate if the current favorite is resolved or not. The frontend can then filter for resolved favorites if necessary.

Jira: https://4teamwork.atlassian.net/browse/GEVER-473

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/base/model/favorite.py
+++ b/opengever/base/model/favorite.py
@@ -98,7 +98,21 @@ class Favorite(Base):
         params.update(kwargs)
         return cls(**params)
 
-    def serialize(self, portal_url):
+    def serialize(self, portal_url, resolve=False):
+        """Serializes a favorite.
+
+        Favorites are unresolved by default. The `target_url` contains a link
+        to resolve a favorite. Resolving a favorite means, we lookup the favorite
+        object by its oguid.
+
+        The serializer can also return already resolved favorites. The target url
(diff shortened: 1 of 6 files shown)


opengever/base/model/favorite.py
```diff
@@ -98,7 +98,21 @@ def create(cls, userid, obj, **kwargs):
...
+
+        Favorites are unresolved by default. The `target_url` contains a link
+        to resolve a favorite. Resolving a favorite means, we lookup the favorite
+        object by its oguid.
+
+        The serializer can also return already resolved favorites. The target url
+        will then contain the url to the object itself.
+
+        If the favorite cannot be resolved, i.e. because it does not exist on the
+        current admin-unit, the unresolved favorite will returned instead.
```
njohner, line 112: will BE returned instead

PR #6516, opened 2020-06-23
Fix deadline of task templates no longer shown in tabular listing.

mbaechtold:
Since we switched the tabular listing view to Solr, the deadline of task templates was no longer shown. There is already a Solr field called `deadline`, which is a `pdate`. But the deadline of the task templates is an integer. So we store the value in a Solr field called `period` and render it in the tabular listing.

Belongs to the Jira issue https://4teamwork.atlassian.net/browse/GEVER-561

The screenshot depicts the affected column in the table listing of task templates.

image

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/core/upgrades/20200623100315_index_task_template_deadline_as_period/upgrade.py
+++ b/opengever/core/upgrades/20200623100315_index_task_template_deadline_as_period/upgrade.py
@@ -1,0 +1,18 @@
+from ftw.upgrade import UpgradeStep
+
+
+class IndexTaskTemplateDeadlineAsPeriod(UpgradeStep):
+    """Index task template deadline as period (Solr only).
+    """
+    deferrable = True
+
+    def __call__(self):
+        self.install_upgrade_profile()
+
+        # `period` is not an index in the catalog, only a field in Solr. In order
+        # avoid reindexing the whole objects, we must pick any index that exists
+        # for all objects and is fast to compute, `UID` in this case.
+        self.catalog_reindex_objects(
+            {"portal_type": "opengever.tasktemplates.tasktemplate"},
+            idxs=["UID", "period"]
+        )
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.4.0 (unreleased)
 ---------------------
 
+- Fix deadline of task templates no longer shown in tabular listing. [mbaechtold]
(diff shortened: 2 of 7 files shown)


opengever/core/upgrades/20200623100315_index_task_template_deadline_as_period/upgrade.py
```diff
@@ -0,0 +1,14 @@
...
+class IndexTaskTemplateDeadlineAsPeriod(UpgradeStep):
+    """Index task template deadline as period (Solr only).
+    """
+    deferrable = True
+
+    def __call__(self):
+        self.install_upgrade_profile()
+        self.catalog_reindex_objects(
+            {'portal_type': 'opengever.tasktemplates.tasktemplate'},
+            idxs=['period']
```
njohner, line 13: Careful here, if I'm not mistaken, because `period` is not an index in the catalog, you will end up reindexing the whole object, as if `idxs` was empty. So what you need to do here is add a cheap index, for example `UID`. See for example: https://github.com/4teamwork/opengever.core/blob/master/opengever/core/upgrades/20190418160803_add_has_same_type_children_metadata_column/upgrade.py#L16-L18

@lukasgraf, @deiferni Can one of you confirm?

mbaechtold: You are right, `period` is not an index in the catalog, only a field in Solr. 

I did not know about that quirk in the catalog, thanks for pointing it out 👍 

I have improved the upgrade step according to the example you have linked in your comment.

Nevertheless, I'd also like the confirmation from @deiferni or @lukasgraf.

deiferni: correct, the catalog will filter out invalid indexes which will result in an empty list.

bit of background, the filtering happens in:
https://github.com/zopefoundation/Products.CMFCore/blob/862811d730e7c80a734d9a48c7d96b8674eb9994/Products/CMFCore/CatalogTool.py#L355
then this will later on cause all indexes to be used in 
https://github.com/zopefoundation/Products.ZCatalog/blob/304d9d1b868577b6aa5b57c26889aed27590d571/src/Products/ZCatalog/Catalog.py#L352-L355

mbaechtold: Thanks for linking the relevant code, very kind of you 👍

PR #6517, opened 2020-06-23
Add API expansion "main-dossier".

mbaechtold:
This API expansion enables us to fetch the data of the main dossier of any given object.

This is required for an enhancement of the new content selector widget in GEVER UI. See https://4teamwork.atlassian.net/browse/GEVER-555 where @elioschmutz, @deiferni any myself discussed possible solutions.

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/public/dev-manual/api/operations.rst
+++ b/docs/public/dev-manual/api/operations.rst
@@ -114,6 +114,79 @@ children des Objekts).
     .. literalinclude:: examples/example_get.py
 
 
+Erweiterbare Komponente (Expansion)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Eine erweiterbare Komponente (auch "Expansion" genannt) ist ein Mechanismus, um in einem GET-Request
+einer Ressource weitere Information anzufordern, z.B. die Navigation, Breadcrumbs, etc. Damit kann verhindert
+werden, dass ein weiterer Request abgesetzt werden muss, um diese Angaben abzuholen. Weitere Informationen
+zu diesem Mechanismus findet man unter https://plonerestapi.readthedocs.io/en/latest/expansion.html.
+
+Die erweiterbaren Komponenten können über den Parameter `expand` in die Response eingebettet werden.
+
+Folgende erweiterbaren Komponente stehen zur Verfügung und sind in den entsprechenden Kapiteln beschrieben:
+
+- :ref:`navigation`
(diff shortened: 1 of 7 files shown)


docs/public/dev-manual/api/operations.rst
```diff
@@ -114,6 +114,62 @@ children des Objekts).
     .. literalinclude:: examples/example_get.py
 
 
+Hauptdossier als erweiterbare Komponente
```
mbaechtold, line 117: I'm not sure if this is the best place for the documentation. Any ideas? Maybe we should put all expansions in one dedicated chapter?

njohner: Hum, all the other expansions have their own chapter. On the other hand, it would make sense to have a paragraph about expansion here too. How about adding a few sentences about what expansions are, use main_dossier as example and add a few pointers to the other expansions (breadcrumbs, navigation and maybe listing-stats)?

mbaechtold: Thank you very much for the review, @njohner. I have updated the documentation. Could you pleas have another look (even though you already approved this pull request)?

opengever/api/dossier.py
```diff
@@ -22,3 +26,27 @@ def __call__(self, *args, **kwargs):
...
+        self.context = context
+        self.request = request
+
+    def __call__(self, expand=True):
+        return {
+            'main-dossier': self._get_main_dossier_summary()
+        }
+
+    def _get_main_dossier_summary(self):
+        if not hasattr(self.context, 'get_main_dossier'):
```
njohner, line 45: So I just realised, you actually make use of acquisition here. That's why this works on all content types. But that is a bit sneaky. Can you instead use `get_main_dossier` from `from opengever.dossier.utils import get_main_dossier`? that is a bit more explicit. It will also return `None` when the context is not in a dossier

mbaechtold: Thanks for the hint. The previous solution based on the attribute was proposed by @deiferni (maybe I did not understand it correctly). Just pinging him so he might chime in.

I have changed the implementation and I'm using the helper function now.

mbaechtold:
Oooops, the test got lost during a rebase. Luckily my IDE kept a copy of it. I have added the tests now, @njohner.

elioschmutz:
@mbaechtold 👌 works perfectly! Here is the frontend-implementation: https://github.com/4teamwork/gever-ui/pull/1165

PR #6518, opened 2020-06-23
Add additional data in GET of Proposals and ProposalResponses

njohner:
In this PR we add some additional data in the GET of proposals and proposal responses. `Proposal`s indeed include important data depending on their status which is not part of their Schema (the meeting it is scheduled in, the excerpt and the decision number). Similarly additional data is required to represent certain `ProposalResponse`s correctly, which is not in the response schema (titles and urls of meeting and successor proposal).

For https://4teamwork.atlassian.net/browse/GEVER-313

## Checklist (Must have)
- [x] Changelog entry

### DIFF
--- a/opengever/api/proposal.py
+++ b/opengever/api/proposal.py
@@ -2,7 +2,8 @@ from opengever.api.response import SerializeResponseToJson
 from opengever.api.serializer import GeverSerializeFolderToJson
 from opengever.base.oguid import Oguid
 from opengever.meeting.model import Meeting
-from opengever.meeting.proposal import IBaseProposal
+from opengever.meeting.proposal import IProposal
+from opengever.meeting.proposal import ISubmittedProposal
 from opengever.meeting.proposalhistory import IProposalResponse
 from plone.restapi.interfaces import ISerializeToJson
 from plone.restapi.interfaces import ISerializeToJsonSummary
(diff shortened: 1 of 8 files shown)


opengever/api/proposal.py
```diff
@@ -1,13 +1,44 @@
...
+        if meeting:
+            result[u'meeting'] = {
+                'title': meeting.title,
+                '@id': meeting.get_url(view=None)
+                }
+
+        decision_number = self.context.load_model().get_decision_number()
+        if decision_number:
+            result[u'decision_number'] = decision_number
+        return result
```
elioschmutz, line 88: Shouldn't the `excerpt`, `meeting`, 'successor_proposals` and the `decision_number` always be included in the response but with `None`? This would make the response more transparent, i.e. b because we know, that there can be a meeting but currently we don't have one...

opengever/api/tests/test_proposal.py
```diff
@@ -73,3 +73,17 @@ def test_submitted_proposal_contains_a_list_of_responses(self, browser):
              u'response_type': u'submitted',
              u'text': u''},
             responses[0])
+
+    @browsing
+    def test_decided_proposal_contains_meeting_and_decision_number(self, browser):
```
elioschmutz, line 93: and the excerpt...

opengever/api/proposal.py
njohner: I was actually unsure what the behaviour should be here. But fine with me to always send them back. Will amend.

njohner:
@elioschmutz I've updated according to your comments and added a some more tests (specifically realised that the `decided_proposal` is actually the `SubmittedProposal`, so wrote another test for the proposal on the dossier side...).

As for documentation, we do not have complete responses for all the `portal_type` GETs, so I don't think we need to update them. For proposals we have an entry for `POST` because it's a bit special (https://docs.onegovgever.ch/dev-manual/api/proposals/), but nothing on the GET and IMHO that's ok...

PR #6519, opened 2020-06-24
Allow trashed as field in @listing endpoint.

tinagerber:
To decide in the frontend which folderActions are available for documents, you have to know if the documents are in the trash or not. Therefore the listing endpoint must provide this information.

Jira: https://4teamwork.atlassian.net/browse/GEVER-45

## Checklist (Must have)

- [x] Changelog entry


tinagerber:
@deiferni The field is now also represented in the tests.

PR #6522, opened 2020-06-25
Make "populate_filename_column_in_favorites" UpgradeStep more robust:

lukasgraf:
**Background**
This upgrade step attempts to poplate the filename SQL column for favorites, which is limited to 105 characters. If an object (specifically mails) is encountered that for some reason still has a longer filename indexed in the catalog, this upgrade step will fail.

**Fix**
This change makes it more robust by truncating any filename to the max length before attempting to write it to SQL.

**Jira** (related to): https://4teamwork.atlassian.net/browse/GEVER-477

❗ Needs to be backported to `2020.3.x`

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.4.0 (unreleased)
 ---------------------
 
+- Make "populate_filename_column_in_favorites" UpgradeStep more robust. [lgraf]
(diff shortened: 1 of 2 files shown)


lukasgraf:
FYI: Tested this fixed upgrade step like that on a local upgrade (2020.2.6 -> `master`) of a PROD copy where the issue surfaced.

PR #6523, opened 2020-06-25
Track date when content within a dossier changes.

mbaechtold:
We need to track the date (not datetime) of the moment, when a content of a dossier is changed (or the dossier itself). 

We add a new property to the dossier for this purpose and index it in Solr, but not in the catalog since the feature will only be implemented in the new GEVER UI.

Deif and I thought hard about the naming and `touched` seemed the most appropriate (`changed` and `modified` are already in use, `updated` seemed confusing).

Belongs to the Jira issue https://4teamwork.atlassian.net/browse/GEVER-40

## Checklist (Must have)

- [x] Changelog entry


opengever/dossier/handlers.py
```diff
@@ -149,7 +152,11 @@ def purge_reference_number_mappings(copied_dossier, event):
     prefix_adapter.purge_mappings()
 
 
-def update_dossier_touched_date(dossier, event):
+def update_dossier_touched_date(obj, event):
     today = date.today()
-    if IDossier(dossier).touched != today:
-        IDossier(dossier).touched = today
+    while obj and not IPloneSiteRoot.providedBy(obj):
```
njohner, line 157: Did you check how many times every object will be checked here? I sort of fear that this will not be how you expect for deeply nested dossiers. For example:
- let's say you modify `dossier1/dossier2/dossier3/document`.
- you end up in this handler, which recursively walk up the path, hence set `touched` for `dossier1`, `dossier2` and `dossier3`
- but setting `touched` on `dossier3` will fire an `ObjectModified` envent, which will trigger this handler, which will recursively check `touched` for `dossier2` and `dossier1`, etc...

opengever/dossier/handlers.py
```diff
@@ -149,7 +152,11 @@ def purge_reference_number_mappings(copied_dossier, event):
...
+        if (IDossierMarker.providedBy(obj)
+                and IDossier(obj).touched != today
+                and (not (IObjectAddedEvent.providedBy(event) and obj == event.object))):
+            IDossier(obj).touched = today
```
njohner, line 161: If my suspicion above is correct, you could simply add a `break` statement here. This will then automatically propagate up in the next event...

opengever/dossier/handlers.py
```diff
@@ -159,4 +159,5 @@ def update_dossier_touched_date(obj, event):
                 and IDossier(obj).touched != today
                 and (not (IObjectAddedEvent.providedBy(event) and obj == event.object))):
             IDossier(obj).touched = today
+            obj.reindexObject(idxs=['touched'])
```
njohner, line 162: Add a `UID` here, to make sure you are not reindexing the full object

opengever/core/upgrades/20200612160606_update_solr_field_touched_for_all_dossiers/upgrade.py
```diff
@@ -0,0 +1,50 @@
...
+        return self.solr_search.search(
+            query=u'object_provides:"opengever.dossier.behaviors.dossier.IDossierMarker"'
+        ).num_found
+
+    def update_touched_date(self, dossier):
+        modified = self.solr_search.search(
+            query=u'path_parent:"{0}"'.format(dossier['path']),
+            sort='modified desc',
+            rows=1
+        ).docs[0]['modified']
```
njohner, line 44: I think that will wail if the dossier is empty?

opengever/core/upgrades/20200612160606_update_solr_field_touched_for_all_dossiers/upgrade.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+
+    def __call__(self):
+        self.solr_search = getUtility(ISolrSearch)
+        self.solr_connection = getUtility(ISolrConnectionManager).connection
+
+        dossiers = self.query_dossiers()
+        for index, dossier in enumerate(dossiers.docs, 1):
+            self.update_touched_date(dossier)
+            if index % 1000 == 0:
```
njohner, line 21: I'd also log something here, because that might be a lot of objects to update, although it should be rather fast I guess.

opengever/core/upgrades/20200612160606_update_solr_field_touched_for_all_dossiers/upgrade.py
```diff
@@ -0,0 +1,50 @@
...
+from ftw.upgrade import UpgradeStep
+from zope.component import getUtility
+
+
+class UpdateSolrFieldTouchedForAllDossiers(UpgradeStep):
+    """Update Solr field "touched" for all dossiers.
+    """
+
+    def __call__(self):
+        self.solr_search = getUtility(ISolrSearch)
```
njohner, line 15: I was wondering whether we should check if solr is enabled, and if not, simply return. On the other hand, I think it now is enabled for all our deployments

opengever/dossier/tests/test_touched.py
```diff
@@ -0,0 +1,39 @@
...
+                         .within(self.branch_repofolder))
+        self.assertIsNone(IDossier(dossier).touched)
+
+    @browsing
+    def test_touched_date_is_not_updated_on_the_same_day(self, browser):
+        self.login(self.administrator, browser=browser)
+
+        # Multiple modifications on the same day result in the same touched
+        # date. This is mostly implicit by the usage of a date field instead
+        # of a datetime field.
```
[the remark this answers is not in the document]
mbaechtold: True. I have renamed the test case.

opengever/dossier/handlers.py
mbaechtold: I could not prove your assumption. Each dossier in the path is only affected once.

opengever/dossier/handlers.py
mbaechtold: See above.

opengever/dossier/handlers.py
mbaechtold: Done.

opengever/core/upgrades/20200612160606_update_solr_field_touched_for_all_dossiers/upgrade.py
mbaechtold: True. I have made this part more robust.

opengever/core/upgrades/20200612160606_update_solr_field_touched_for_all_dossiers/upgrade.py
mbaechtold: Sure, logging is always nice. Done.

opengever/core/upgrades/20200612160606_update_solr_field_touched_for_all_dossiers/upgrade.py
mbaechtold: It won't hurt. I have added a check.

opengever/core/upgrades/20200612160606_update_solr_field_touched_for_all_dossiers/upgrade.py
```diff
@@ -0,0 +1,68 @@
...
+                log.info("Committing items {}/{} to Solr.".format(index, dossiers.num_found))
+                self.solr_connection.commit()
+
+        # Commit the remaining items to Solr.
+        self.solr_connection.commit()
+
+    @property
+    def is_solr_enabled(self):
+        registry = getUtility(IRegistry)
+        settings = registry.forInterface(ISearchSettings)
```
mbaechtold, line 40: @njohner is this the correct way to check if Solr is enabled? The name `ISearchSettings` suggests something else and the description of `ISearchSettings.use_solr` is `Enables Solr for search forms.`. 🤔

PR #6526, opened 2020-06-29
Fix logout issue by introducing custom @@logout view

lukasgraf:
This change fixes a logout issue where the user gets logged out of CAS, but the Plone session is never invalidated.

It does so by
- Introducing a new `@@logout` view that
  - **Invalidates the Plone session**, and then...
  - either **redirects to the CAS** logout or the Plone `logged_out` page
- Wires up that custom view for the logout action
- **Drops the cross-tab logout JavaScripts**, because they interfere with logout by hijacking clicks, currently only work in Firefox, and could be considered an anti-feature anyway

This view also makes sure that **logout works for Zope level users** (like `zopemaster`) by making sure `CookieAuthHelper` cookies also get cleared if the Plone session plugin's cookie name has been customized.

Jira: https://4teamwork.atlassian.net/browse/GEVER-558 

## Checklist (Must have)

- [x] Changelog entry


opengever/base/tests/test_logout.py
```diff
@@ -26,8 +56,58 @@ def tearDown(self):
...
         self.login(self.regular_user, browser=browser)
+        browser.open(self.portal)
+
+        browser.raise_http_errors = False
+        browser.find('Log out').click()
 
-        browser.open(view='logout_url')
-        self.assertEquals('http://nohost/portal/logout', browser.contents)
+        cas_server_url = 'http://nohost/portal'
+        self.assertEquals('/'.join((cas_server_url, 'logout')), browser.url)
```
buchi, line 113: `assertEquals` is deprecated.

lukasgraf: I thought I caught them all... Thanks, fixed 👍

PR #6530, opened 2020-06-30
Fix contenttree.js: IE cannot parse ES6 syntax.

njohner:
In https://github.com/4teamwork/opengever.core/pull/6461/ we introduced a modified version of the javascript for the contenttree widget, to optimise its use with large amounts of items. This change introduced some ES6 syntax which is not supported by IE. This breaks gever on IE.

For https://4teamwork.atlassian.net/browse/GEVER-621

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.4.0 (unreleased)
 ---------------------
 
+- Fix contenttree.js so that it is also supported by IE. [njohner]
 - Also allow replacing concrete responsibles with interactive responsibles when triggering task templates. [deiferni]
 - Remove cross-tab logout functionality. [lgraf]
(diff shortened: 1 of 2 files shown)


njohner:
True, that's why I put it in two separate commits. But I can keep only the first one.

lukasgraf:
Yeah, I assumed so. Sorry, I know you had the best intentions, but I would advise against it in this case. It probably would be fine, but it just has (a slight) potential to complicate things a lot down the line, which I just don't think is worth it right now.

njohner:
I've egg-patched on `sia`, `ipet` and `gev01p` for all deployments except `quarten` and `ogquarten` which are not up to date. I've left a copy of the original file on each server `/apps/eggs/opengever.core-2020.3.1-py2.7.egg/opengever/base/browser/resources/contenttree.js_BAK`

PR #6531, opened 2020-06-30
Separate conditions for links from document to proposal/submitted proposal/meeting.

deiferni:
Add separate conditions to display the link from a document to its proposal, submitted proposal and meeting (if any). The visibility of these items can be different depending on how gever is setup. We now check separately if the user can view the proposal or the submitted proposal. In both cases we display link to the proposal/submitted porposal and the meeting.

I have added displaying a link to the submitted proposal, so this is not strictly a bugfix. I think it may be the best approach and help the user to navigate to all the different content that is generated automatically. The following cases are displayed differently _(sorry about screenshot alignment)_:

- User can only see meeting, not case dossier
Screenshot 2020-06-30 at 12 21 08
- User can see case dossier and meeting
Screenshot 2020-06-30 at 12 20 10
- User can only see case dossier (meeting is not linked, only name appears)
Screenshot 2020-06-30 at 12 19 42
- if the user can se neither, we don't display any links (as before)

For https://4teamwork.atlassian.net/browse/GEVER-590

- [x] Changelog entry

### DIFF
--- a/opengever/document/base.py
+++ b/opengever/document/base.py
@@ -71,33 +71,59 @@ class BaseDocumentMixin(object):
 
         return None
 
+    def get_submitted_proposal(self, check_security=True):
+        """Return the submitted  proposal to which this document belongs.
+
+        Return submitted proposal when self is an excerpt document in the
+        meeting dossier.
+        """
+        relations = list(getUtility(ICatalog).findRelations({
+                'to_id': getUtility(IIntIds).getId(aq_inner(self)),
+                'from_attribute': 'excerpts'}))
+
+        # We expect that there are 0 or 1 relation, because this document
+        # cannot be the excerpt of multiple proposals.
+        if relations:
+            relation = relations[0]
+            submitted_proposal = relation.from_object
+            assert(ISubmittedProposal.providedBy(submitted_proposal))
+            if check_security:
+                if api.user.has_permission('View', obj=submitted_proposal):
+                    return submitted_proposal
+            else:
+                return submitted_proposal
+
+        return None
+
     def get_proposal(self):
         """Return the proposal to which this document belongs.
 
-        This may return a "proposal" or a "submitted proposal".
+        Handles the following cases:
+        - the document is inside a proposal, return the parent
+        - the document is an excerpt in the meeting dossier, find proposal via
+          submitted porposal relation
+        - the document is an excerpt in the case dossier
+
         """
         if self.is_inside_a_proposal():
             return aq_parent(aq_inner(self))
 
-        # Find submitted proposal when self is an excerpt document in the
-        # meeting dossier.
-        for relation in getUtility(ICatalog).findRelations({
-                'to_id': getUtility(IIntIds).getId(aq_inner(self)),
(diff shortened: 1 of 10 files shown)


opengever/meeting/tests/test_excerpt.py
```diff
@@ -154,96 +162,75 @@ class TestExcerptOverview(IntegrationTestCase):
...
         self.login(self.regular_user, browser)
         browser.open(excerpt1, view='tabbedview_view-overview')
-        fields = dict(zip(
-            browser.css('.documentMetadata th').text,
-            map(attrgetter('innerHTML'), browser.css('.documentMetadata td')),
-        ))
-        self.assertItemsEqual(expected_fields, fields.keys())
+        fields = meeting_fields(browser)
+        self.assertItemsEqual(expected_fields, fields.keys(),
+            msg='The dossier responsible cannot view the submitted proposal')
```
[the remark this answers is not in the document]
njohner: ok now as `committee_responsible` 😅

opengever/meeting/tests/test_excerpt.py
```diff
@@ -228,9 +228,57 @@ def test_no_link_to_proposal_visible_if_no_access_to_dossier(self, browser):
...
+            self.meeting.model.get_url(),
+            fields['Meeting'].css('a').first.get('href'))
+
+    @browsing
+    def test_no_link_to_submitted_proposal_visible_if_no_access_to_committee(self, browser):
+        self.login(self.committee_responsible, browser)
+        agenda_item = self.schedule_proposal(self.meeting, self.submitted_proposal)
+        agenda_item.decide()
+        excerpt1 = agenda_item.generate_excerpt('excerpt 1')
+        agenda_item.return_excerpt(excerpt1)
```
njohner, line 244: how about using  the `decided_proposal` instead?

opengever/document/base.py
```diff
@@ -259,6 +259,7 @@ def containing_subdossier_title(self):
 
         return dossier.title
 
+
```
njohner, line 262: commit message should be "add ONE line"

opengever/document/base.py
```diff
@@ -71,33 +71,56 @@ def get_parent_inbox(self):
 
         return None
 
+    def get_submitted_proposal(self, check_security=True):
+        # Find proposal when self is an excerpt document in the
```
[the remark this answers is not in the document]
deiferni: this should be a proper docstring in any case, i have converted it

opengever/document/base.py
deiferni: 😂  💛 , i have clarified the commit message.

opengever/meeting/tests/test_excerpt.py
deiferni: it was my first attempt, but then i figured the fixture is missing the documents etc. i gave up attempting to properly register all of them. if you don't insist on it i would propose to not change that, too for this bugfix.

deiferni:
@njohner implemented all your comments, but one. hope that is fine for you.

PR #6532, opened 2020-06-30
Enhance "@config" with "private_folder_url".

mbaechtold:
Expose the path to the user's private folder in the `@config` API endpoint. This serves as a feature flag for the GEVER UI too.

The reason for this addition is the fact, that the GEVER UI does not get the path to the private folder right in every case. Especially when the user id is an e-mail address, the id of the private folder is normalized (see the tests for an example).

Belongs to the Jira issue https://4teamwork.atlassian.net/browse/GEVER-611

## Checklist (Must have)
- [x] Changelog entry

## Checklist (if applicable)

- API change:

### DIFF
--- a/docs/public/dev-manual/api/config.rst
+++ b/docs/public/dev-manual/api/config.rst
@@ -56,6 +56,7 @@ GEVER-Mandanten abgefragt werden.
           "is_emm_environment": false,
           "max_dossier_levels": 5,
           "max_repositoryfolder_levels": 3,
+          "private_folder_url": "http://localhost:8080/fd/private/kathi.barfuss",
           "recently_touched_limit": 10,
           "root_url": "http://localhost:8080/fd",
           "sharing_configuration": {
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -7,6 +7,7 @@ Changelog
 - Make creating favorites more robust in case of workflow issues. [deiferni]
 - Improve response history for (automatically) opened subtasks in sequential task templates. [mbaechtold]
 - Fix contenttree.js so that it is also supported by IE. [njohner]
+- Expose the url to the user's private folder in the `@config` API endpoint. Serves as feature flag too. [mbaechtold]
 - Also allow replacing concrete responsibles with interactive responsibles when triggering task templates. [deiferni]
 - Remove cross-tab logout functionality. [lgraf]
 - Add @@logout view to clear Plone session and redirect to CAS logout if necessary. [lgraf]
--- a/opengever/api/config.py
+++ b/opengever/api/config.py
@@ -2,6 +2,7 @@ from ftw.bumblebee.config import bumblebee_config
 from opengever.base import utils
 from opengever.base.interfaces import IGeverSettings
 from opengever.officeconnector.helpers import is_client_ip_in_office_connector_disallowed_ip_ranges
+from opengever.private import get_private_folder_url
 from plone.restapi.services import Service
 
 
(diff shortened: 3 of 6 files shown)


opengever/private/tests/test_utils.py
```diff
@@ -0,0 +1,99 @@
...
+        # Make sure the private folder has been created. This is for a better
+        # understanding of what is happening here.
+        self.assertEqual(
+            ['jane.bond'],
+            [
+                home_folder_id
+                for home_folder_id
+                in self.portal.private.objectIds()
+                if home_folder_id.startswith('jane')
+            ]
```
njohner, line 34: I think you could replace that with `self.assertIn('jane.bond', self.portal.private.objectIds())`

opengever/private/tests/test_utils.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class TestPrivateFolderUtils(IntegrationTestCase):
+
+    @browsing
+    def test_user_having_home_folder(self, browser):
+        # Enable member area creation in Plone.
+        membership_tool = api.portal.get_tool('portal_membership')
+        enable_opengever_private()
+        self.assertTrue(membership_tool.getMemberareaCreationFlag())
```
njohner, line 17: I'd add an assertion that there is no home folder with the name that we will be creating:
`self.assertNotIn('jane.bond', self.portal.private.objectIds())`

opengever/private/tests/test_utils.py
```diff
@@ -0,0 +1,99 @@
...
+
+        # Create a user and its private folder (this is not triggered by
+        # authenticating the user, it must be done manually).
+        jane = create(Builder('user').named("Jane", "Bond"))
+        self.login(jane, browser)
+        membership_tool.createMemberarea()
```
njohner, line 23: Maybe put that in a method? `create_user_and_private_folder`

opengever/private/utils.py
```diff
@@ -0,0 +1,14 @@
...
+def get_private_folder():
+    membership_tool = api.portal.get_tool('portal_membership')
+    home_folder = membership_tool.getHomeFolder()
+    return home_folder
+
+
+def get_private_folder_url():
+    home_folder = get_private_folder()
+    if not home_folder:
+        return ''
```
deiferni, line 13: i would use `None` to indicate it is not there, not empty string.

opengever/private/tests/test_utils.py
```diff
@@ -0,0 +1,99 @@
+from ftw.builder.builder import Builder
+from ftw.builder.builder import create
+from ftw.testbrowser import browsing
+from opengever.private import enable_opengever_private
+from opengever.private.utils import get_private_folder_url
+from opengever.testing import IntegrationTestCase
+from plone import api
+
+
+class TestPrivateFolderUtils(IntegrationTestCase):
```
njohner, line 10: I'm not a 100% sure what we are testing in this class? are we testing the `get_private_folder_url` works correctly? or are we testing the name normalisation that happens when creating a private folder?

opengever/private/utils.py
```diff
@@ -0,0 +1,14 @@
+from plone import api
```
deiferni, line 1: this seems like it is one of the main api functions for this package? have you considered moving it to `__init__.py` so it can be imported `from opengever.private`.

_may be personal taste so feel free to merge without addressing this comment 😉_

docs/public/dev-manual/api/config.rst
```diff
@@ -56,6 +56,7 @@ GEVER-Mandanten abgefragt werden.
           "is_emm_environment": false,
           "max_dossier_levels": 5,
           "max_repositoryfolder_levels": 3,
+          "private_folder": "",
```
deiferni, line 59: i would prefer to have a non-empty example in the docs

mbaechtold: Done.

opengever/private/tests/test_utils.py
mbaechtold: We're testing `get_private_folder_url`. I have moved the tests to `test_folder.py`.

opengever/private/tests/test_utils.py
mbaechtold: Done.

opengever/private/tests/test_utils.py
mbaechtold: Done.

opengever/private/utils.py
mbaechtold: Nope, not even thought of it but I like you suggestion so I have move the functions.

opengever/private/utils.py
mbaechtold: Done.

opengever/private/tests/test_utils.py
mbaechtold: This code only exists once after implementing the requested changes so I would rather like to leave as is.

PR #6533, opened 2020-06-30
Robustify adding favorites with invalid workflows.

deiferni:
When for some reason workflow state cannot be retrieved we still want to be able to create the favorite. Thus make the code a bit more robust for this non-critical information.

Also fix an upgrade that would attempt to write `Missing.Value` from the catalog into an sql column which won't work due to missing serializers.

- ✅   i have manually verified that favorites can now be added
- ✅   i have run an upgrade from https://github.com/4teamwork/opengever.core/commit/91a89d17b (the commit before https://github.com/4teamwork/opengever.core/commit/04fcdf52cae62e3ccacfa6c0ff3b3d5a4d9c9ebf was merged to https://github.com/4teamwork/opengever.core/tree/deif/617-fix-favorites-workflow-state to make sure the upgrade would now pass and doesn't break other things

Sample sentry message from my local machine: https://sentry.4teamwork.ch/sentry/onegov-gever/issues/68240/

For https://4teamwork.atlassian.net/browse/GEVER-617

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.4.0 (unreleased)
 ---------------------
 
+- Make creating favorites more robust in case of workflow issues. [deiferni]
 - Improve response history for (automatically) opened subtasks in sequential task templates. [mbaechtold]
 - Fix contenttree.js so that it is also supported by IE. [njohner]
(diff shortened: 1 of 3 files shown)


deiferni:
amended example sentry message at https://sentry.4teamwork.ch/sentry/onegov-gever/issues/68240/

PR #6535, opened 2020-07-01
Improve "touched" date feature.

mbaechtold:
This is a follow-up of the pull request https://github.com/4teamwork/opengever.core/pull/6523 and belongs to the Jira issue https://4teamwork.atlassian.net/browse/GEVER-40

The product owner identified some shortcomings while testing the first iteration of the new feature. 

This pull requests fixes the following shortcomings:

1.  Set touched date upon dossier creation: Until now, new dossiers did not get the touched date.
2. Update touched date upon workflow changes: When the workflow state of tasks, dossiers or proposals changes, we also need to "touch" the containing dossiers.

This pull request also touches the upgrade step which indexes the "touched" date because we need to run it again on dev.onegovgever.ch due to issue 2.


mbaechtold:
Thanks for the review.

@njohner, you touched a point. I also thought about removing that line in the handler. But then I was not sure what would be the proper way to set a default on newly created object (schema or handler) and opted to put a default value on the schema. But you are right, the handler is hard to read and I will switch to your proposed variant. I need to fix some tests anyway.

As we discussed in Slack, we don't need to handle `dispositions` as they are never put into dossiers.

opengever/dossier/tests/test_touched.py
```diff
@@ -90,3 +90,73 @@ def test_moving_content_touches_all_dossiers_in_path(self, browser):
             self.assertEqual("2020-06-12", str(IDossier(self.dossier).touched))
             self.assertEqual("2020-06-12", str(IDossier(self.subdossier).touched))
             self.assertEqual("2020-06-12", str(IDossier(self.subdossier2).touched))
+
+    @browsing
+    def test_modifying_proposal_touches_containing_dossier(self, browser):
```
mbaechtold, line 95: To the reviewer: this test is not related to the changes in this commit but it does not hurt to have another test ;-)

PR #6538, opened 2020-07-03
Add API endpoint to logout current user.

njohner:
We add an API endpoint to log out the current user by expiring his cookies (__ac and deployment specific cookie, e.g. __ac_fd). This is necessary because logout in the frontend should also log the user out of the backend.

I'm actually really not sure this is the right approach. It works, but maybe that should be the job of the portal or something?
I haven't added the documentation yet, as I'm more and more doubtful that this is a good solution.

For https://4teamwork.atlassian.net/browse/GEVER-558
PR in the frontend: https://github.com/4teamwork/gever-ui/pull/1186

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:

### DIFF
--- a/opengever/api/logout.py
+++ b/opengever/api/logout.py
@@ -1,0 +1,34 @@
+from Products.CMFCore.utils import getToolByName
+from plone.restapi.services.auth.logout import Logout
+
+
+class GeverLogout(Logout):
+    """Log out current user by trying to invalidate the JWT and expiring
+    his cookies (__ac and deployment specific cookie, e.g. __ac_fd).
+    This is necessary because logout in the frontend should also log the user
+    out of the backend.
+    """
+
+    def reply(self):
+        # Handles logout by invalidating the JWT
+        # This sets the response status accordingly if it fails
+        resp = super(GeverLogout, self).reply()
+
+        # Now we try to expire the user's cookies
+
+        # We do not use logout of PAS here, as it redirects to logout view.
+        pas = getToolByName(self.context, 'acl_users')
+        pas.resetCredentials(self.request, self.request['RESPONSE'])
+
+        # Also expire any __ac cookie that might have been issued by the
+        # CookieAuthHelper on the Zope root (e.g. for zopemaster)
+        self.request.response.expireCookie('__ac', path='/')
+
+        # if either the JWT token was invalidated or any cookie was deleted
+        # logout is considered successful, otherwise we return the error
+        # from the JWT logout.
+        for cookie_name, cookie in self.request.response.cookies.items():
+            if cookie.get('value') == 'deleted' and cookie_name in self.request.cookies:
+                self.request.response.setStatus(204)
+                return
+        return resp
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.5.0 (unreleased)
 ---------------------
 
+- Overwrite logout API endpoint to also expire the user's cookies. [njohner]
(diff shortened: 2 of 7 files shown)


tinagerber:
https://plonerestapi.readthedocs.io/en/latest/conventions.html#naming-convention-for-attribute-names-in-uris

njohner:
right thanks...

opengever/api/logout_user.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+
+    def reply(self):
+        # We do not use logout of PAS here, as it redirects to logout view.
+        pas = getToolByName(self.context, 'acl_users')
+        pas.resetCredentials(self.request, self.request['RESPONSE'])
+
+        # Also expire any __ac cookie that might have been issued by the
+        # CookieAuthHelper on the Zope root (e.g. for zopemaster)
+        self.request.response.expireCookie('__ac', path='/')
```
elioschmutz, line 18: can't we somehow reuse the `@@logout` view? instead of the redirect, they have to do the same thing

elioschmutz:
@njohner 👍 but I'm not sure either, if this is the right approach. So perhaps someone else should take another look to it...

opengever/api/logout_user.py
```diff
@@ -0,0 +1,19 @@
...
+        return
```
deiferni, line 19: should we do a 204 no-content status? or does that happen already by returning `None`?

opengever/api/logout_user.py
njohner: I'd say no. It's basically a single line of code that is shared, and if we could work around the redirect it would be 2 lines of code. So I'd rather simply duplicate that here.

njohner:
Ok I've now overwritten the `@logout` endpoint instead. I've also improved the tests to actually test that logout works and added a few new ones. I've also added some documentation for the endpoint.
As there was no real opposition, I'd propose we that we go with this for now. Once we only have the new UI, we will probably not need to invalidate the session cookies anymore and we can then remove that customisation.

opengever/api/logout.py
```diff
@@ -0,0 +1,34 @@
...
+    """Log out current user by trying to invalidate the JWT and expiring
+    his cookies (__ac and deployment specific cookie, e.g. __ac_fd).
+    This is necessary because logout in the frontend should also log the user
+    out of the backend.
+    """
+
+    def reply(self):
+        # Handles logout by invalidating the JWT
+        # This sets the response status accordingly if it fails
+        resp = super(GeverLogout, self).reply()
```
buchi, line 15: I see one problem now:
The frontend does not know if there's a `__ac` cookie, thus it always has to call the `@logout` endpoint. However if there's no `__ac` cookie, it will get a 501 response because by default we do not store JWT tokens and thus it's not possible to invalidate those on the server side. Usually this is not needed as the Client can just delete the token client-side to logout.
This can't be done with the `__ac` cookies as those should be `httponly` by default and thus the server needs to delete 'em.

Not sure how we want to deal with that. Maybe add a URL parameter `invalidate=false`. And if that's given, skip the JWT invalidation? 

`__ac` cookies can't be invalidated individually, just deleted.

njohner: Yes currently the frontend gets a 501 but ignores it. I mean it deletes it's own session token and then basically tries to logout from the backend just in case, and that might fail, but it doesn't really care.
With the current implementation I would also return a 501 even if we skip deleting the JWT token if there is no `__ac` or customized (`__ac_fd`) cookie to delete. I sort of think this is OK, it's the backend answering that it could not logout.

PR #6539, opened 2020-07-03
Add UserSnap API key to registry.

njohner:
For the integration of Usersnap in the frontend, we add a registry entry with the API key. This also gets returned in the `@config` endpoint and will serve as feature flag (usersnap is shown when the API key is defined).

I took the opportunity to update the API documentation with many of the missing features and configs returned by the `@config` endpoint.

For https://4teamwork.atlassian.net/browse/GEVER-516

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking: It's non breaking!
    - [ ] api-change label added

### DIFF
--- a/opengever/base/interfaces.py
+++ b/opengever/base/interfaces.py
@@ -372,6 +372,14 @@ class IGeverUI(Interface):
         default=False)
 
 
+class IUserSnapSettings(Interface):
+
+    api_key = schema.TextLine(
+        title=u"API key",
+        description=u"API key for the usersnap widget.",
+        default=u"")
+
+
 class IDocPropertyProvider(Interface):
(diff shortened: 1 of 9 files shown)


opengever/base/interfaces.py
```diff
@@ -372,6 +372,14 @@ class IGeverUI(Interface):
         default=False)
 
 
+class IUserSnap(Interface):
```
deiferni, line 375: i would prefer a name indicating that this is a registry setting interface, like `IUserSnapSettings`

elioschmutz:
@njohner whats the state of this issue? can we merge it?

PR #6540, opened 2020-07-06
Fix loading of more items in contenttree widget

buchi:
Loading more items did not work for top level items. E.g. when selecting related items in a task and if there are more than 100 items. The problem was that top level items do not have a parent and the previous implementation simply appended the additionally loaded items to the parents children.

JIRA: https://4teamwork.atlassian.net/browse/GEVER-655

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- [x] JS tested with IE9

### DIFF
--- a/opengever/base/browser/resources/contenttree.js
+++ b/opengever/base/browser/resources/contenttree.js
@@ -71,10 +71,14 @@ if(jQuery) (function($){
                 o.multiSelect = false;
             }
 
-            function loadTree(c, t, r, b_start) {
+            function loadTree(c, t, r, b_start, extend) {
                 $(c).addClass('wait');
                 $.post(o.script, {href: t, rel: r, b_start: b_start || 0}, function(data) {
-                    $(c).removeClass('wait').append(data);
+                    if (extend) {
+                        $(c).removeClass('wait').append($(data).contents());
+                    } else {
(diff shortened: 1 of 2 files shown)


opengever/base/browser/resources/contenttree.js
```diff
@@ -107,14 +111,25 @@ if(jQuery) (function($){
...
+                            $(this).data('bstart'),
+                            true
+                        );
+                    } else {
+                        loadTree(
+                            ul,
+                            escape(o.rootUrl),
+                            '0',
+                            $(this).data('bstart'),
+                            true
```
deiferni, line 130: if i properly understand the PR description shouldn't you pass `false` here? you are always passing `true` as the newly introduced param, that doesn't look right 🤔.

buchi: `false` is implicitly used [here](https://github.com/4teamwork/opengever.core/blob/f6b29f9deabcb69427311512fec2273db278ae13/opengever/base/browser/resources/contenttree.js#L102) when loading the first batch. In the "load more" case we always want to extend the current tree. This was wrong before but not visible because of styling.

deiferni: ahh, right. thanks!

PR #6544, opened 2020-07-07
Fix filtering on values containing spaces in listing endpoint.

njohner:
Filtering by `document_author` was not working in the new frontend when the author name contained a space. We fix this here by escaping spaces in filters when creating the solr query in the `listing` endpoint. We do not fix this in `ftw.solr` as the `escape` function defined there is used for the search query, for which spaces do not need to be escaped. Instead we implement our own escaping function specifically for filter queries.

For https://4teamwork.atlassian.net/browse/GEVER-649

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added

### DIFF
--- a/opengever/api/solr_query_service.py
+++ b/opengever/api/solr_query_service.py
@@ -1,9 +1,10 @@
 from collective.elephantvocabulary import wrap_vocabulary
+from copy import copy
 from DateTime import DateTime
 from DateTime.interfaces import DateTimeError
 from ftw.solr.converters import to_iso8601
 from ftw.solr.interfaces import ISolrSearch
-from ftw.solr.query import escape
+from ftw.solr.query import SPECIAL_CHARS
 from opengever.base.behaviors.translated_title import ITranslatedTitleSupport
 from opengever.base.helpers import display_name
 from opengever.base.solr import OGSolrContentListing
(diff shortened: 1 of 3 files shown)


opengever/api/solr_query_service.py
```diff
@@ -26,6 +26,20 @@
 import Missing
 
 
+TO_ESCAPE = [el for el in SPECIAL_CHARS]
```
deiferni, line 29:
```
from copy import copy
TO_ESCAPE = copy(SPECIAL_CHARS)
```
?

opengever/api/solr_query_service.py
```diff
@@ -26,6 +26,20 @@
...
+TO_ESCAPE.append(' ')
+
+
+def filter_escape(term):
+    """Copy of ftw.solr.query.escape, but using the above defined TO_ESCAPE
+    instead of SPECIAL_CHARS, additionnaly including a white space. For queries,
```
deiferni, line 35: additionally

opengever/api/solr_query_service.py
```diff
@@ -26,6 +26,20 @@
...
+    whitespaces should not be escaped, but for filters yes.
```
deiferni, line 36: but for filters yes -> but for filters they should be
?

PR #6550, opened 2020-07-09
Fix update mail filename upgrade step.

njohner:
Filename column has to be added to favorites before fixing mail filenames, i.e. https://github.com/4teamwork/opengever.core/blob/master/opengever/core/upgrades/20200508162410_add_filename_column_to_favorites/upgrade.py should be executed before https://github.com/4teamwork/opengever.core/blob/master/opengever/core/upgrades/20200507073857_update_filename_of_mails_with_file_extension_msg/upgrade.py which should itself be executed (although this is not mandatory) before https://github.com/4teamwork/opengever.core/blob/master/opengever/core/upgrades/20200508163417_populate_filename_column_in_favorites/upgrade.py

For https://4teamwork.atlassian.net/browse/GEVER-168

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

  - [ ] Make it deferrable if possible: I don't make it deferrable so it gets executed in between the two others

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.4.0 (2020-07-02)
 ---------------------
 
+- Fix update mail filename upgrade step. [njohner]
 - Improve check if solr has started to prevent an issue during the testserver startup. [sebastianmanger]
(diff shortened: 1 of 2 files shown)


njohner:
This was backported directly

PR #6555, opened 2020-07-10
Add @role-assignment-reports endpoint to list, add and delete role assignment reports.

tinagerber:
An administrator wants to know what local roles a user or group has.
This PR makes it possible to create role assignment reports.

With the API endpoint `@role-assignment-reports` a report can be created. This is stored in the `RoleAsssignmentReportsStorage` with the status "in progress".
In a nightly job, all local roles of a Principal are collected and the report is completed with them. The report then has the status "ready".

Jira: https://4teamwork.atlassian.net/browse/GEVER-33

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:

### DIFF
--- a/docs/public/dev-manual/api/docs_changelog.rst
+++ b/docs/public/dev-manual/api/docs_changelog.rst
@@ -5,6 +5,12 @@ Changelog
 
 Im Folgenden sind (substantielle) Änderungen an der Dokumentation aufgeführt.
 
+2020-07-10
+----------
+
+- Kapitel "Rollenzuweisungs-Berichte" hinzugefügt
+
+
 2020-06-11
 ----------
 
--- a/opengever/api/permissions.zcml
+++ b/opengever/api/permissions.zcml
@@ -4,5 +4,6 @@
 
   <permission id="opengever.api.ViewAllowedRolesAndPrincipals" title="opengever.api: View AllowedRolesAndPrincipals" />
   <permission id="opengever.api.TransferAssignment" title="opengever.api: Transfer Assignment" />
+  <permission id="opengever.api.ManageRoleAssignmentReports" title="opengever.api: Manage Role Assignment Reports" />
 
 </configure>
--- a/opengever/api/role_assignment_reports.py
+++ b/opengever/api/role_assignment_reports.py
@@ -1,0 +1,113 @@
+from opengever.base.interfaces import IRoleAssignmentReportsStorage
+from plone.app.uuid.utils import uuidToObject
+from plone.protect.interfaces import IDisableCSRFProtection
+from plone.restapi.batching import HypermediaBatch
+from plone.restapi.deserializer import json_body
+from plone.restapi.services import Service
+from zExceptions import BadRequest
+from zope.interface import alsoProvides
+from zope.interface import implements
+from zope.publisher.interfaces import IPublishTraverse
+
+
+class RoleAssignmentReportsGet(Service):
+    """API Endpoint which returns a list of all role assignment reports
+
+    GET /@role-assignment-reports HTTP/1.1
+
+    or a single report
+
+    GET /@role-assignment-reports/report_1 HTTP/1.1
+    """
+
+    implements(IPublishTraverse)
+
+    def __init__(self, context, request):
+        super(RoleAssignmentReportsGet, self).__init__(context, request)
+        self.params = []
+
+    def publishTraverse(self, request, name):
+        # Consume any path segments after /@role-assignment-reports as parameters
+        self.params.append(name)
+        return self
+
+    def reply(self):
+        storage = IRoleAssignmentReportsStorage(self.context)
+        # a single report
+        if len(self.params) == 1:
+            reportid = self.params[0]
+            try:
+                result = storage.get(reportid)
+            except KeyError:
+                raise BadRequest("Invalid reportid '{}'".format(reportid))
+            for item in result['items']:
+                item['url'] = uuidToObject(item['UID']).absolute_url()
+        # all reports
+        elif len(self.params) == 0:
+            result = {'items': storage.list()}
+            for report in result['items']:
+                del report['items']
+                report['@id'] = '/'.join([self.context.absolute_url(), '@role-assignment-reports',
+                                          report['reportid']])
+        else:
+            raise BadRequest("Too many parameters. Only principalid is allowed.")
+
+        batch = HypermediaBatch(self.request, result['items'])
+        result['items'] = [item for item in batch]
+        result['items_total'] = batch.items_total
+        result['@id'] = batch.canonical_url
+        if batch.links:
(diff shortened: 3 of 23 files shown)


opengever/api/permissions.zcml
```diff
@@ -4,5 +4,6 @@
 
   <permission id="opengever.api.ViewAllowedRolesAndPrincipals" title="opengever.api: View AllowedRolesAndPrincipals" />
   <permission id="opengever.api.TransferAssignment" title="opengever.api: Transfer Assignment" />
+  <permission id="opengever.api.ManageRoleAssignmentReports" title="opengever.api: Manage Role Assignment Reports" />
```
njohner, line 7: Can you be a bit more explicit in the commit message, maybe:
`Add new permission to manage role assignment reports`

docs/public/dev-manual/api/docs_changelog.rst
```diff
@@ -5,6 +5,12 @@ Changelog
 
 Im Folgenden sind (substantielle) Änderungen an der Dokumentation aufgeführt.
 
+2020-07-10
+----------
+
+- Kapitel "Rollenzuweisungs-Berichte" hinzugefügt
```
njohner, line 11: ouch right, I think there are several other recent additions that we forgot in here...

opengever/api/role_assignment_reports.py
```diff
@@ -0,0 +1,115 @@
...
+from zope.interface import implements
+from zope.publisher.interfaces import IPublishTraverse
+
+
+class RoleAssignmentReportsGet(Service):
+    """API Endpoint which returns a list of all role assignment reports
+
+    GET /@role-assignment-reports HTTP/1.1
+
+    or
```
njohner, line 22: or of a single report

opengever/api/role_assignment_reports.py
```diff
@@ -0,0 +1,115 @@
...
+                                          report['reportid']])
+        else:
+            raise BadRequest("Too many parameters. Only principalid is allowed.")
+
+        batch = HypermediaBatch(self.request, result['items'])
+        result['items'] = [item for item in batch]
+        result['items_total'] = batch.items_total
+        if batch.links:
+            result['batching'] = batch.links
+
```
njohner, line 63: I think the `@id` is missing no?

opengever/api/role_assignment_reports.py
```diff
@@ -0,0 +1,115 @@
...
+
+class RoleAssignmentReportsPost(Service):
+
+    def reply(self):
+        self.extract_data()
+        # Disable CSRF protection
+        alsoProvides(self.request, IDisableCSRFProtection)
+        storage = IRoleAssignmentReportsStorage(self.context)
+        storage.add(self.principalid)
```
njohner, line 79: POST normally returns the serialised object. Minimum here would be to return the `report_id` but I'd rather return the serialized report

opengever/api/role_assignment_reports.py
```diff
@@ -0,0 +1,115 @@
...
+        self.request.response.setStatus(204)
+        return super(RoleAssignmentReportsPost, self).reply()
+
+    def extract_data(self):
+        data = json_body(self.request)
+        self.principalid = data.get("principalid", None)
+        if not self.principalid:
+            raise BadRequest("Property 'principalid' is required")
+        principal = ActorLookup(self.principalid).lookup()
+        if not(isinstance(principal, (OGDSGroupActor, OGDSUserActor, PloneUserActor))):
```
njohner, line 89: Actually I think you have to allow principals that do not exist. Role assignments are stored as simple strings, so there can be assignments for principals that do no longer exist. We've had that situation a lot with these reports. So instead I would add the report anyway, but return a warning maybe? No idea how we would organise the response though.

opengever/api/role_assignment_reports.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def reply(self):
+        if len(self.params) != 1:
+            raise BadRequest("reportid is required")
+        self.reportid = self.params[0]
+        # Disable CSRF protection
+        alsoProvides(self.request, IDisableCSRFProtection)
+        storage = IRoleAssignmentReportsStorage(self.context)
+        storage.delete(self.reportid)
+        self.request.response.setStatus(204)
```
njohner, line 114: So here we would need to handle unknown `report_id`s, once the `delete` method does not silently skip those

opengever/api/tests/test_role_assignment_reports.py
```diff
@@ -0,0 +1,172 @@
...
+                 u'modified': u'2016-08-31T20:01:33+00:00',
+                 u'principalid': u'kathi.barfuss',
+                 u'reportid': u'report_1',
+                 u'state': u'in progress'},
+                {u'@id': u'http://nohost/plone/@role-assignment-reports/report_0',
+                 u'modified': u'2016-08-31T20:01:33+00:00',
+                 u'principalid': u'jurgen.fischer',
+                 u'reportid': u'report_0',
+                 u'state': u'ready'}],
+             u'items_total': 2}, browser.json)
```
njohner, line 29: The `@id` is missing here

opengever/api/tests/test_role_assignment_reports.py
```diff
@@ -0,0 +1,172 @@
...
+        self.login(self.administrator, browser=browser)
+        with freeze(datetime(2020, 4, 18)):
+            browser.open(self.portal.absolute_url() + '/@role-assignment-reports',
+                         method='POST',
+                         headers=self.api_headers,
+                         data=json.dumps({"principalid": self.meeting_user.getId()}))
+
+        self.assertEqual(browser.status_code, 204)
+
+        browser.open(self.portal.absolute_url() + '/@role-assignment-reports/report_2',
```
njohner, line 89: So because the POST does not return anything you have to guess the report_id to get the report here 😉 .

tinagerber:
@njohner Thanks for the review. I have made the desired adjustments, can you review again?

opengever/api/role_assignment_reports.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def reply(self):
+        storage = IRoleAssignmentReportsStorage(self.context)
+        if len(self.params) == 1:
+            reportid = self.params[0]
+            try:
+                result = storage.get(reportid)
+            except KeyError:
+                raise BadRequest("Invalid reportid '{}'".format(reportid))
+            self.serialize_report(result, summary=True)
```
njohner, line 37: huh, I think there is some mixup here. `Summary` serializers contain less information (subset) from the complete serialization. So when you GET an object, ou use the normal serialization. When you get a container object, usually each item contained in the container will be get it's summary in the `items`...

tinagerber:
@njohner I've now dropped the function.

PR #6536, opened 2020-07-02
Add teaser viewlet to promote the new frontend.

tinagerber:
We want to have the possibility to promote the new frontend.. Therefore a viewlet is implemented which shows a teaser about the new frontend. 
If the teaser is clicked away or the new frontend is visited, the teaser is no longer displayed.

Bildschirmfoto 2020-07-02 um 12 30 59

Mobile view:

Bildschirmfoto 2020-07-02 um 12 32 32

Jira: https://4teamwork.atlassian.net/browse/GEVER-552

Definition of Done: https://4teamwork.atlassian.net/wiki/spaces/CHX4TW/pages/917562/

## Checklist (Must have)

_Everything has to be done/checked. Checked but not present means the author deemed it unnecessary._

- [x] Changelog entry
- [x] Link to issue (Jira or GitHub) and backlink in issue (Jira)

PR #6557, opened 2020-07-14
Fix javascript ordering.

njohner:
This is a follow-up to In https://github.com/4teamwork/opengever.core/pull/6536

`teaser.js` got inserted at the beginning of the registry instead of just before `opengever.base.js`. This led to new deployments and the testserver being broken. The insertion did not work as expected because `teaser.js` got inserted before `opengever.base.js`, then the `insert-before` dose not find the position of the element it should insert before and defaults to beginning of the registry: https://github.com/plone/Products.ResourceRegistries/blob/master/Products/ResourceRegistries/tools/BaseRegistry.py#L842-L844

Simply making sure `opengever.base.js` appears before `teaser.js` in opengever/core/profiles/default/jsregistry.xml is therefore enough.

For https://4teamwork.atlassian.net/browse/GEVER-552

## Checklist (Must have)

- [x] Changelog entry: No changelog entry, as this is in the same version as the commit that introduced the issue

### DIFF
--- a/opengever/core/profiles/default/jsregistry.xml
+++ b/opengever/core/profiles/default/jsregistry.xml
@@ -307,9 +307,9 @@
       cookable="True"
       expression=""
       enabled="True"
-      id="++resource++opengever.base/teaser.js"
+      id="++resource++opengever.base/base.js"
       inline="False"
-      insert-before="++resource++opengever.base/base.js"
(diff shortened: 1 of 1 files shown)


njohner:
As discussed, the existing upgrade works, only the order for new deployments is wrong.

PR #6559, opened 2020-07-14
Handle wildcard in date filters in listing endpoint.

njohner:
Date filters in the `@listing` endpoint did not handle wildcards. We now correctly handle wildcards in these filters.
Note that we have also changed the error handling. The status before was that:
- If no range was provided the code would simply fail with `UnboundLocalError` (https://sentry.4teamwork.ch/sentry/onegov-gever/?query=is%3Aunresolved+UnboundLocalError%3A+local+variable+%27date_from%27+referenced+before+assignment)
- if parsing failed, it would be silently ignored and return no results.

Now when parsing fails it will return an appropriate `Bad Request` error

Not sure whether we should consider that braking. It feels more like fixing a bug to me.

For https://4teamwork.atlassian.net/browse/GEVER-624

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,7 +4,7 @@ Changelog
 2020.6.0 (unreleased)
 ---------------------
 
-- Nothing changed yet.
+- Handle wildcard in date filters in listing endpoint. [njohner]
 
 
 2020.5.0 (2020-07-14)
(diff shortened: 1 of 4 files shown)


njohner:
Note to the reviewer. I've force pushed and there should be 3 new tests now, but it does not show in the diff in github (at least for me). So if you see only two new tests, maybe checkout the branch or something? I have no idea what is going on with github right now

PR #6560, opened 2020-07-17
Handle errors in solrsearch endpoint.

njohner:
Since https://github.com/4teamwork/opengever.core/pull/6499, errors when querying solr will be logged to sentry.
This notably happens for invalid queries that solr cannot parse. While it would be better to avoid actually sending invalid queries to solr, it is complex and would first need us to establish what is the behaviour we expect and what type of user queries we want to support. This will be done in a further step (https://4teamwork.atlassian.net/browse/GEVER-752)

For now, when the search fails on the solr side, we need to at least raise an error in the solrsearch endpoint, so that
the gever-ui can handle such errors. This changes the behaviour of when a query becomes invalid in the `LiveSearch`. Before it would suddenly show no results when a query becomes invalid (try typing "test OR" on dev.onegovgever.ch/fd), now it will keep the last set of results when the query becomes invalid (see https://github.com/4teamwork/gever-ui/blob/master/src/components/Livesearch.vue#L104-L109).

For https://4teamwork.atlassian.net/browse/GEVER-687

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -8,6 +8,7 @@ Changelog
 - Allow workspace members to trash and untrash documents in workspaces. [tinagerber]
 - Handle wildcard in date filters in listing endpoint. [njohner]
 - Handle multiple content interfaces in @navigation endpoint. [njohner]
+- Handle errors in solrsearch endpoint. [njohner]
 
 
 2020.5.0 (2020-07-14)
--- a/opengever/api/solrsearch.py
+++ b/opengever/api/solrsearch.py
@@ -4,6 +4,7 @@ from opengever.api.solr_query_service import SolrQueryBaseService
 from opengever.base.interfaces import ISearchSettings
 from plone import api
 from zExceptions import BadRequest
+from zExceptions import InternalError
 
 BLACKLISTED_ATTRIBUTES = set([
     'getDataOrigin',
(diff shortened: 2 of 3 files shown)


deiferni:
or hm, one comment, would it make sense to use a custom exception for this, i.e. an `InternalError` subclass?

njohner:
@lukasgraf I did ponder whether I should pass on the status code from the request to solr, or always return a 400 or a 500. I decided 500. Here are my thoughts as to why: let's say the user does type `test AND`, I mean this is not fundamentally a user input error or a `BadRequest` as such. It's more that the backend cannot handle it, because it decides to interpret this as a complex query, which then leads to an error. We could had a smarter parsing of this in the backend. So I'd argue that this is a not a `BadRequest`, but and `InternalServerError`.

lukasgraf:
Ok, I see the reasoning here. An incomplete query is not necessarily an invalid query, it's just... incomplete. Works for me 👍
When revisiting this, we could consider having a parameter `livesearch=1` that the frontend could set in a context where it might be sending incomplete queries. That probably would allow the backend to better recognize what could be going on.

njohner:
> or hm, one comment, would it make sense to use a custom exception for this, i.e. an `InternalError` subclass?

@deiferni  I'd tend to maybe raise specific errors once we have implemented the behavior we really want from that endpoint (with https://4teamwork.atlassian.net/browse/GEVER-752). For now it would be something very generic like `SolrSearchFailed` which would not yield much more information.

PR #6563, opened 2020-07-17
Fix createContentInContainer to respect behaviors.

njohner:
We fix `createContentInContainer` to respect behaviors. This has been fixed in `plone.dexterity` in version 2.2.0 (https://github.com/plone/plone.dexterity/pull/17) but never made its way into our `MonkeyPatch` (see https://github.com/plone/plone.dexterity/pull/17). Note that it was fixed several times in `plone.dexterity`, but never quite correctly. The current implementation (https://github.com/plone/plone.dexterity/blob/master/plone/dexterity/utils.py#L133-L150) is still fragile, relying on `getattr` to decide whether a certain field is part of a given schema. This has two problems:
- This does not really test whether the field is part of a given behavior, but only whether the field exist in the storage of that behavior (so it could be part of another behavior using the same schema)
- If at any point an attribute was persisted in the wrong storage (for example `relatedItems` stored on the object instead of in the annotations), this implementation will think that this is the storage where that field should indeed be saved.

So we actually write our own implementation of `createContentInContainer`, which uses the behaviors themselves (the schemata) as source of truth.

This bug in `createContentInContainer` leads to problems for PHVS and their custom behaviors, which are stored in the annotations (https://github.com/4teamwork/opengever.phvs/blob/master/opengever/phvs/configure.zcml#L87-L94) and which were stored on the newly created object when saving a document as pdf (`SavePDFUnderForm`). This leads to the values not getting indexed correctly.

Moreover there was a second issue this time in `SavePDFUnderForm`, where we did not correctly get the data over the behaviors on the source document. This data is then used to created the new PDF document, which would also lead to missing data on the created document (although not in deployments with no custom behavior).

For https://4teamwork.atlassian.net/browse/GEVER-678

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 ---------------------
 
 - Fix id normalization when setting up a repository. [tinagerber]
+- Fix createContentInContainer to respect behaviors. [njohner]
 - Allow workspace members to trash and untrash documents in workspaces. [tinagerber]
 - Handle wildcard in date filters in listing endpoint. [njohner]
 - Handle multiple content interfaces in @navigation endpoint. [njohner]
--- a/opengever/api/scanin.py
+++ b/opengever/api/scanin.py
@@ -78,7 +78,7 @@ class ScanIn(Service):
                 obj = createContentInContainer(
                     private_folder,
                     'opengever.private.dossier',
-                    title='Scaneingang')
+                    title=u'Scaneingang')
                 return obj
 
         return self.error(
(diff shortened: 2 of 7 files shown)


njohner:
So, I actually had to fix this as behaviors are not handled correctly in `plone.dexterity.utils.createContent`. I've also added a test. I tried to write another test for the `SavePDFUnderForm`, but the only behaviour on documents that is stored in the annotations is `relatedItems`, which does not get copied over when saving a document as PDF.

I decided against writing an upgrade step in `opengever.core`, instead I'll make one in `opengever.phvs` and other policies, if others are concerned by this bug. I tried to see whether we had other instances where this bug could have led to data being stored in the wrong storage, but I could not spot any such instance in `opengever.core`.

njohner:
@deiferni Ok this should be ready now. I've tested for PHVS and written an upgrade step there https://github.com/4teamwork/opengever.phvs/pull/157.

The things that you have not reviewed yet are the last 3 commits.

PR #6564, opened 2020-07-20
Add `watchers` field to solr and reindex existing content.

deiferni:
With this PR we add the field `watchers` to solr so that it can be queried via the `@solrsearch` endpoint and to allow a user to view "watched" content. Currently it is only possible to watch tasks, but the implementation is generic so that new content can be added if desired.

We add the following new functionality:
- `watchers` field in solr and indexers for plone content. there is one generic indexer which returns an empty list and one task (and forwarding) specific indexer which queries for watchers. Only users that have set themselves manually as watchers (i.e. that have the `WATCHER_ROLE`) are kept in this index.
- add `IDisableCSRFProtection` to existing endpoints so that reindex operations are successful
- add an upgrade-step to reindex existing tasks
- also add the `watchers` field to the `@listing` endpoint and adapt docs

Watchers could then be queried as follows: `http://localhost:8080/fd/@solrsearch?q.raw=watchers:jonas.baumann`

For https://4teamwork.atlassian.net/browse/GEVER-29.

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
- New functionality:


opengever/activity/center.py
```diff
@@ -93,17 +93,17 @@ def remove_watchers_from_resource_by_role(self, oguid, role):
             # for more information about this issue.
             self.session.expire(resource)
 
-    def get_watchers(self, oguid):
-        """Returns a read-only tuple of watchers for a given oguid.
+    def get_watchers(self, oguid, role=None):
+        """Returns watchers for a given resouce identified by oguid.
```
lukasgraf, line 97: `resouce` -> `resource`

opengever/activity/center.py
```diff
@@ -93,17 +93,17 @@ def remove_watchers_from_resource_by_role(self, oguid, role):
...
+
+        Allow to filter by role to return only watchers of a certain role.
```
lukasgraf, line 99: `Allow` -> `Allows`

opengever/activity/tests/test_notification_center.py
```diff
@@ -142,12 +142,45 @@ def test_get_watchers_returns_a_list_of_resource_watchers(self):
...
+        watchers = self.center.get_watchers(
+            Oguid('fd', '123'), role=WATCHER_ROLE)
+        self.assertItemsEqual([peter, marie], watchers)
+
+    def test_get_watchers_returns_empty_list_when_no_watchers_are_present(self):
+        create(Builder('resource').oguid('fd:123'))
+
+        self.assertEquals([], self.center.get_watchers(Oguid('fd', '123')))
+
+    def test_get_watchers_returns_empty_list_when_no_role_does_not_exist(self):
```
lukasgraf, line 173: -> `...when_role_does_not_exist`

opengever/activity/center.py
```diff
@@ -234,23 +234,41 @@ def _get_oguid_for(self, item):
             return Oguid.for_object(item)
         return item
 
+    def _reindex_object(self, obj_or_oguid):
```
lukasgraf, line 237: `_reindex_watchers`? _(To avoid the impression that the whole object is getting reindexed)_

deiferni: i like it, it clarifies what happens in the method. amended 👍 .

deiferni:
_thanks, amended. will wait with merging for UI feedback for now_

PR #6567, opened 2020-07-22
Allow workspace members to delete documents in workspaces.

tinagerber:
With this change, WorkspaceMember and WorkspaceAdmin can delete documents in workspaces and workspace folders.
Furthermore, it is now only possible to delete a document if it is in the trash.

Jira: https://4teamwork.atlassian.net/browse/GEVER-573

## Checklist (Must have)

- [x] Changelog entry


opengever/core/profiles/default/workflows/opengever_mail_workflow/definition.xml
```diff
@@ -186,4 +186,19 @@
       <guard-permission>Manage portal content</guard-permission>
     </guard>
   </transition>
+  <variable
+      variable_id="review_history"
```
deiferni, line 190: what happens to existing mails? what if there are already trashed or removed mails? could the missing review_history lead to an error?

tinagerber: I have tested this extensively locally and found no errors. It probably behaves as if it were the first transition.

tinagerber:
@deiferni I made the changes and also discovered and fixed another bug (see last commit). Can you review it again?

PR #6568, opened 2020-07-23
Fix @workflow when no permissions in target state.

deiferni:
Fix executing workflow transitions via @workflow service when the executing user has no permission in target state on the context for which the transition is executed.

This is done by accessing `review_history` with elevated privileges, if necessary.

There is an issue in the `WorkflowTransition` service when the user performs a transition into a state where his access is revoked. The service will raise a `WorkflowException` when attempting to access `review_history` while rendering the response. This will prevent the transition from being executed.

We work around this issue when we catch a `WorkflowException` and attempt to read the history with elevated privileges.

For https://4teamwork.atlassian.net/browse/GEVER-755.

I am of the opinion that this is a `plone.restapi` issue, as the issue surfaces whenever `review_history` is protected with a `guard-permission`, but the executing user no longer has those permission in the target state. Example in https://github.com/4teamwork/opengever.core/blob/47a8673419f7c56ab542a9577c1269fbac7c0195/opengever/document/upgrades/20160914162259_add_shadow_document_state/workflows/opengever_document_workflow/definition.xml#L334-L335. At the moment `plone.restapi` would no longer be able to execute these transitions but instead fail with a traceback.

Sentry: https://sentry.4teamwork.ch/sentry/onegov-gever/issues/53693

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/api/tests/test_workflow.py
+++ b/opengever/api/tests/test_workflow.py
@@ -72,3 +72,40 @@ class TestWorkflowPost(IntegrationTestCase):
              u'review_state': u'document-state-removed'},
             browser.json
         )
+
+    @browsing
+    def test_calling_endpoint_without_transition_gives_400(self, browser):
+        self.login(self.regular_user, browser=browser)
+
+        with browser.expect_http_error(code=400):
+            browser.open(
+                self.dossier.absolute_url() + '/@workflow',
+                method='POST', headers=self.api_headers)
+
+        self.assertEqual('Bad Request', browser.json['error']['type'])
+        self.assertIn(u"Invalid transition 'None'",
+                      browser.json['error']['message'])
+
+    @browsing
+    def test_calling_workflow_with_additional_path_segments_results_in_404(self, browser):
+        self.login(self.regular_user, browser=browser)
+
+        with browser.expect_http_error(code=404):
+            browser.open(
+                self.dossier.absolute_url() + '/@workflow/dossier-transition-resolve/invalid',
+                method='POST', headers=self.api_headers)
+
+        self.assertEqual('NotFound', browser.json['type'])
+
+    @browsing
+    def test_invalid_transition_results_in_400(self, browser):
+        self.login(self.regular_user, browser=browser)
+
+        with browser.expect_http_error(code=400):
+            browser.open(
+                self.dossier.absolute_url() + '/@workflow/invalid-transition',
+                method='POST', headers=self.api_headers)
+
+        self.assertEqual('Bad Request', browser.json['error']['type'])
+        self.assertIn(u"Invalid transition 'invalid-transition'",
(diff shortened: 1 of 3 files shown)


opengever/api/transition.py
```diff
@@ -40,12 +41,30 @@ def check_action_available(self):
...
+        service will raise a `WorkflowException` when attempting to access
+        `review_history` while rendering the response. This will prevent the
+        transition from being executed.
+
+        We work around this issue when we catch a `WorkflowException` and
+        attempt to read the history with elevated privileges.
+        """
+        try:
+            return super(GEVERWorkflowTransition, self).reply()
+        except WorkflowException:
```
njohner, line 57: This is fairly dangerous, because it could hide the fact that the transition failed. It works only because `WorkflowException` are caught in `WorkflowTransition` (https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/services/workflow/transition.py#L69). I almost wonder if we would not be better copying the whole endpoint code here.

opengever/api/tests/test_workflow.py
```diff
@@ -0,0 +1,74 @@
...
+            self.document.absolute_url() + '/@workflow/document-transition-remove',
+            method="POST", headers=self.api_headers)
+
+        self.assertDictContainsSubset(
+            {u'title': u'document-state-removed',
+             u'actor': u'kathi.barfuss',
+             u'action': u'document-transition-remove',
+             u'review_state': u'document-state-removed'},
+            browser.json
+        )
```
njohner, line 74: Can we add a test that a transition that cannot be executed correctly returns an error with the corresponding status code?

opengever/api/transition.py
deiferni: yes that was our first solution but then i figured that is also a bit dangerous due to possible changes upstream we won't catch. i'm quite undecided though and can be persuaded in all directions 😏.
but maybe you're right, we better copy and try to fix it upstream.

deiferni:
@njohner @tinagerber i have amended what was requested. you can hijack and merge tomorrow if you find it acceptable 😉 .

opengever/api/transition.py
```diff
@@ -40,12 +41,62 @@ def check_action_available(self):
...
+        """Access review history with elevated privileges, if necessary.
+
+        There is an issue in the `WorkflowTransition` service when the user
+        performs a transition into a state where his access is revoked. The
+        service will raise a `WorkflowException` when attempting to access
+        `review_history` while rendering the response. This will prevent the
+        transition from being executed.
+
+        We work around this issue when we catch a `WorkflowException` and
+        attempt to read the history with elevated privileges.
```
njohner, line 53: Hum, so now it seems you decided to copy the whole endpoint and fix it here. So I guess the docstring needs adapting.

PR #6573, opened 2020-07-30
Add mail-in address to inbox serializer.

njohner:
With this PR we:
- Add the mail-in address to the inbox serializer. 
- Add the inbox_id to the inbox serializer. 
- We also clean-up the query for inbox documents which still used an outdated filter `isWorkingCopy` which is not used in gever anymore (https://github.com/4teamwork/opengever.core/commit/1e7c8a1140247d2c65919925d9e715362ff92c50)

For https://4teamwork.atlassian.net/browse/GEVER-463
## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added

### DIFF
--- a/opengever/api/inbox.py
+++ b/opengever/api/inbox.py
@@ -1,0 +1,18 @@
+from ftw.mail.interfaces import IEmailAddress
+from opengever.api.serializer import GeverSerializeFolderToJson
+from opengever.inbox.inbox import IInbox
+from plone.restapi.interfaces import ISerializeToJson
+from zope.component import adapter
+from zope.interface import implementer
+from zope.interface import Interface
+
+
+@implementer(ISerializeToJson)
+@adapter(IInbox, Interface)
+class SerializeInboxToJson(GeverSerializeFolderToJson):
+
+    def __call__(self, *args, **kwargs):
+        result = super(SerializeInboxToJson, self).__call__(*args, **kwargs)
+        result[u'email'] = IEmailAddress(self.request).get_email_for_object(
+            self.context)
+        return result
--- a/opengever/api/inbox.py
+++ b/opengever/api/inbox.py
@@ -1,6 +1,7 @@
 from ftw.mail.interfaces import IEmailAddress
 from opengever.api.serializer import GeverSerializeFolderToJson
 from opengever.inbox.inbox import IInbox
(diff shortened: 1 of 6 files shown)


opengever/api/inbox.py
```diff
@@ -0,0 +1,20 @@
...
+
+@implementer(ISerializeToJson)
+@adapter(IInbox, Interface)
+class SerializeInboxToJson(GeverSerializeFolderToJson):
+
+    def __call__(self, *args, **kwargs):
+        result = super(SerializeInboxToJson, self).__call__(*args, **kwargs)
+        result[u'email'] = IEmailAddress(self.request).get_email_for_object(
+            self.context)
+        result[u'inbox_id'] = get_current_org_unit().inbox().id()
```
deiferni, line 19: this will always return the currently selected org-unit, regardless of context. i think you want to use `IResponsibleOrgUnit.responsible_org_unit` here

njohner: Ok I've done the following now: 
- Use the `responsible_org_unit` if it is set on the inbox
- Otherwise get the current orgunit and use that. (`responsible_org_unit` is not always set on the inbox).

There are some weird cases right now in local development and on Dev, that, according to @phgross should not exist. There we have two orgunits sharing a single inbox. The classic UI will `get_current_org_unit` to decide which tasks to list, so for a single plone object, you will see different task listings depending on the cookie storing the currently selected org_unit. That is not very nice for a REST-API... I created a story to clean that up (https://4teamwork.atlassian.net/browse/GEVER-825)

opengever/api/inbox.py
```diff
@@ -0,0 +1,24 @@
...
+class SerializeInboxToJson(GeverSerializeFolderToJson):
+
+    def __call__(self, *args, **kwargs):
+        result = super(SerializeInboxToJson, self).__call__(*args, **kwargs)
+        result[u'email'] = IEmailAddress(self.request).get_email_for_object(
+            self.context)
+
+        # org_unit does not have to be configured on an inbox. If it isn't
+        # we default to the current org_unit.
+        orgunit = self.context.get_responsible_org_unit() or get_current_org_unit()
```
deiferni, line 22: i'd cover this special case with a test.

PR #6574, opened 2020-07-30
Fix docs only partially indexed in solr.

deiferni:
With this PR we provide a fix for documents that are only partially indexed in solr.

We think affected documents may have been not indexed in solr at all due to a bug with date serialization in `ftw.solr`. The bug was only fixed in https://github.com/4teamwork/ftw.solr/pull/158.

Subsequent edits of the document by users via plone or reindex operations during upgrades would have partially indexed or restored the document in solr, but only some of the fields as we use atomic updates.

We've seen that all affected documents do not have the created date set as we never manually reindex it, as it never changes. Thus apply a heuristic to detect broken solr documents. We assume that broken documents do not have the field 'created' set to a value in solr.

I have decided against writing better catalog queries or only reindexing solr for now, as we expect the amount of affected documents to be low, dozends or at most hundreds. The extra time required to also reindexing the catalog in such cases should be minimal.

✅   additionally, i have manually verified that on my machine broken solr entries are fixed with the upgrade

For https://4teamwork.atlassian.net/browse/GEVER-787.

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/core/upgrades/20200730094255_fix_docs_only_partially_indexed_in_solr/upgrade.py
+++ b/opengever/core/upgrades/20200730094255_fix_docs_only_partially_indexed_in_solr/upgrade.py
@@ -1,0 +1,88 @@
+from ftw.solr.interfaces import ISolrConnectionManager
+from ftw.solr.interfaces import ISolrSettings
+from ftw.upgrade import UpgradeStep
+from plone.registry.interfaces import IRegistry
+from zope.component import getUtility
+import logging
+
+
+logger = logging.getLogger('opengever.core')
+
+
+class FixDocsOnlyPartiallyIndexedInSolr(UpgradeStep):
+    """Fix docs only partially indexed in solr.
+
+    We think affected documents may have been not indexed in solr at all due to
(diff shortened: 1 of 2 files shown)


opengever/core/upgrades/20200730094255_fix_docs_only_partially_indexed_in_solr/upgrade.py
```diff
@@ -0,0 +1,89 @@
...
+        registry = getUtility(IRegistry)
+        settings = registry.forInterface(ISolrSettings)
+        return settings.enabled
+
+    def __call__(self):
+        if not self.is_solr_enabled:
+            return
+
+        self.solr_connection = getUtility(ISolrConnectionManager).connection
+        import pudb; pudb.set_trace()  # noqa
```
buchi, line 42: pdb leftover

deiferni: thanks 🙈, sorry i was doing some last minute checks and forgot to remove it.

PR #6575, opened 2020-07-30
Add @resolve-oguid endpoint, amend @notifications and @globalindex.

deiferni:
With this PR we add the endpoint `@resolve-oguid` which allows to get the a serialized object via oguid. Then endpoint will proxy the change to another remote internally, if necessary.

It also includes the following other changes:
- add `oguid` to `@notifications` and `@globalindex` endpoint
- amend `@globalindex` endpoint with the following changes
  - Support search queries
  - translate `task_type`
  - add correct batching information
  - add issuer_fullname
- extend `@config` with admin-unit and org-unit

I have verified that the api-change does not have an impact in https://4teamwork.slack.com/archives/C3RUACNN7/p1595930231095400.

For https://4teamwork.atlassian.net/browse/GEVER-439.

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [x] api-change label added

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,7 +4,11 @@ Changelog
 2020.6.1 (unreleased)
 ---------------------
 
+- Add @resolve-oguid endpoint. [deiferni]
+- Include oguid in @notifications endpoint. [deiferni]
+- Extend @globalindex endpoint, avoid duplicate tasks, add batching information. [deiferni]
 - Add upgrade to fix docs only partially indexed in solr. [deiferni]
+- Extend @config with admin-unit and org-unit. [njohner]
 
 
 2020.6.0 (2020-07-29)
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.6.1 (unreleased)
 ---------------------
 
+- Handle search queries in GlobalIndexGet endpoint. [njohner]
 - Add @resolve-oguid endpoint. [deiferni]
 - Include oguid in @notifications endpoint. [deiferni]
 - Extend @globalindex endpoint, avoid duplicate tasks, add batching information. [deiferni]
--- a/docs/public/dev-manual/api/resolve_oguid.rst
+++ b/docs/public/dev-manual/api/resolve_oguid.rst
@@ -1,0 +1,49 @@
+.. resolve_oguid:
+
+Resolve Oguid
+=============
+
+Mit dem ``@resolve-oguid`` kann die serialisierte Repräsentation eines Inhalts
+im Mandatenverbund abgefragt werden. Der Endpoint leitet den Request an eine
+andere Admin-Unit weiter, sollte sich der Inhalt nicht auf der aktuellen
+Admin-Unit befinden.
+
+Parameter
+---------
+
+Die Oguid muss als Query-String Parameter mitgegeben werden.
+
+================ ===========================================================
+Parameter        Beschreibung
+================ ===========================================================
+``oguid``        Die Oguid des zu serialisierenden Inhalts. Diese setzt sich
+                 aus <admin_unit_id>:<int_id> zusammen.
+================ ===========================================================
+
+Des weiteren sind alle Query-String Parameter erlaubt, die bei einem GET Request
+auf einen einzelnen Inhalt möglich sind.
+
+  .. sourcecode:: http
+
+    GET /@resolve-oguid?oguid=fd:123456789 HTTP/1.1
+    Accept: application/json
+
+  .. sourcecode:: http
+
+    HTTP/1.1 200 OK
+    Content-Type: application/json
+
+    {
+      "@id": "http://localhost:8080/rk/ordnungssystem/uebergeordnete-erlasse/dossier-1/task-1",
+      "@type": "opengever.task.task",
+      "UID": "1546308893054eb8b3c21fc8be129607",
+      "changed": "2020-07-27T12:28:47+00:00",
+       "...": "..."
+    }
+
+
+Remote Requests
+---------------
+
+Wurde der Request intern an eine andere Admin-Unit weitergeleitet so ist dies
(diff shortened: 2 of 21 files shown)


opengever/api/tests/test_resolve_oguid.py
```diff
@@ -0,0 +1,192 @@
...
+    @browsing
+    def test_resolve_oguid_respects_security(self, browser):
+        self.login(self.administrator)
+        oguid = Oguid.for_object(self.protected_document)
+
+        self.login(self.regular_user, browser)
+        url = '{}/@resolve-oguid?oguid={}'.format(self.portal.absolute_url(), str(oguid))
+
+        with browser.expect_http_error(code=401):
+            browser.open(url, method='GET', headers=self.api_headers)
```
njohner, line 70: I guess it's difficult to add a test if a remote request would also respect the security?

opengever/api/tests/test_resolve_oguid.py
```diff
@@ -0,0 +1,192 @@
...
+            mocker.get(remote_url, text=json.dumps(response))
+
+            browser.open(local_url, method='GET', headers=self.api_headers)
+            self.assertEqual(response, browser.json)
+
+    @browsing
+    def test_resolve_remote_oguid_proxies_query_string_parameteres_to_remote(self, browser):
+        self.setup_remote_admin_unit()
+        remote_url, local_url = self.get_resolve_urls(someparam='foo')
+
```
njohner, line 125: For readability I'd add a test here that the remote url contains `someparam=foo`.

opengever/globalindex/model/task.py
```diff
@@ -500,6 +500,23 @@ def restrict(self):
                 TaskPrincipal.principal.in_(principals))
         return self.filter(Task.task_id.in_(principal_query))
 
+    def avoid_duplicates(self, admin_unit_id=None):
+        """Avoid duplicates and only list one task if a task has a successor.
+
+        If a task has a successor task, list only the task that is on a
+        specified admin unit. The default admin unit is the local one, so
+        the query will only return tasks that are on the local admin unit.
```
njohner, line 508: `so the query will only return tasks that are on the local admin unit` -> `so for tasks with a successor, the query will only return tasks that are on the local admin unit`

opengever/api/resolve_oguid.py
```diff
@@ -72,6 +72,8 @@ def _get_remote_serialized_object(self, oguid, params):
                    'X-OGDS-AUID': get_current_admin_unit().id()}
         url = '/'.join([target_unit.site_url, '@resolve-oguid'])
 
+        self.request.response.setHeader('X-GEVER-RemoteRequest', url)
```
njohner, line 75: I'd propose that we add a test that the `X-GEVER-RemoteRequest` header is not already on the request adding the header here and proxying the request to a remote host. If it is, we return an error. That should never happen but if it does, we might be in an infinite loop of request proxying.

docs/public/dev-manual/api/resolve_oguid.rst
```diff
@@ -0,0 +1,49 @@
...
+Die Oguid muss als Query-String Parameter mitgegeben werden.
+
+================ ===========================================================
+Parameter        Beschreibung
+================ ===========================================================
+``oguid``        Die Oguid des zu serialisierenden Inhalts. Diese setzt sich
+                 aus <admin_unit_id>:<int_id> zusammen.
+================ ===========================================================
+
+Des weiteren sind alle Query-String Parameter erlaubt, die bei eime GET Request
```
njohner, line 23: `eime` -> `einem`

njohner:
I'm highjacking this one so that we can move forward with the frontend before @all4manu goes on vacation

opengever/api/tests/test_resolve_oguid.py
deiferni: yes, we always only mock the remote requests. but i think it is already tested as resolving the object via oguid will always happen in the local and not in the remote part. so this test already covers what would happen in the remote request.

docs/HISTORY.txt
```diff
@@ -4,7 +4,12 @@ Changelog
 2020.6.1 (unreleased)
 ---------------------
 
+- Handle search queries in GlobalIndexGet endpoint.
```
deiferni, line 7: you should take credits here 😉  with a `[njohner]`

opengever/api/globalindex.py
```diff
@@ -76,3 +86,36 @@ def reply(self):
...
+    def extend_query_with_textfilter(self, query, text):
+        """Extends the given `query` with text filters.
+        This is a copy from opengever.tabbedview.sqlsource.SqlTableSource
+        """
+        if len(text):
+            if isinstance(text, str):
+                text = text.decode('utf-8')
+
+            # remove trailing asterisk
+            if text.endswith('*'):
```
deiferni, line 99: as we decode to unicode i would use unicode strings for all the strings below this line

opengever/api/tests/test_globalindex.py
```diff
@@ -119,6 +119,23 @@ def test_filter_by_value_list(self, browser):
...
+    @browsing
+    def test_handles_search_queries(self, browser):
+        self.login(self.regular_user, browser=browser)
+
+        view = '@globalindex'
+        browser.open(self.portal, view=view.format(''), headers=self.api_headers)
+        self.assertEqual(15, browser.json['items_total'])
+        all_tasks = browser.json.get('items')
+
+        search_term = 'Vertrag'
```
deiferni, line 131: i would prefer a test with some non-ascii characters, as we still have unicode issues from time to time 😞

njohner:
@deiferni I've updated according to your comments, cleaned up the commits and added the documentation for searching in the globalindex endpoint. I'll merge when the tests are green except if you have more comments

deiferni:
i can't approve my own PR, so :shipit:!

PR #6576, opened 2020-07-31
Define update chain to sync solr documents

bierik:
RIS stores documents in the GEVER solr. To search for documents in RIS, the documents have to be synced to a remote solr.

A StatelessScriptUpdateProcessor with the name `sync.chain` provides a script that is using a JavaScript Script to sync the documents to a remote solr.

The update of `ftw.recipe.solr` provides a configoverlay option to the ftw.recipe.solr to configure the update chain.

Belongs to the spike: https://4teamwork.atlassian.net/browse/DJANGO-449

## Checklist (Must have)

- [x] Changelog entry


README.rst
```diff
@@ -226,6 +226,43 @@ Then run the `activate_solr` maintenance script:
...
+            }
+        }
+    }
+
+When the ``sync.chain`` UpdateRequestProcessorChain is activated, the ``remoteCoreURL`` and ``portalTypes`` option has to be set in the ``buildout.cfg``. The ``portalTypes`` options is a comma separated list of portal_types to sync.
+This is done by using the ``jvm-opts`` option:
+
+.. code::
+    [solr]
+        jvm-opts = -Xms512m -Xmx512m -Xss256k -DremoteCoreURL=http://localhost:8984/solr/ris  -DportalTypes=opengever.document.document,opengever.dosser
```
njohner, line 260: `opengever.dosser` ? maybe `opengever.dossier.businesscasedossier`?

bierik: 👍

bierik:
@njohner Tests are now passing

solr-conf/solrconfig.xml
```diff
@@ -1074,6 +1074,19 @@
     </highlighting>
   </searchComponent>
 
+  <updateRequestProcessorChain name="sync.chain">
+    <processor class="solr.processor.StatelessScriptUpdateProcessorFactory">
```
buchi, line 1078: I assume that the `StatelessScriptUpdateProcessorFactory` is run before the `RunUpdateProcessorFactory` which actually updates the document in the local Solr. This could lead to unwanted delays. I think it would be better to register this processor as `post-processor`.

README.rst
```diff
@@ -226,6 +226,43 @@ Then run the `activate_solr` maintenance script:
     $ bin/instance run src/opengever.maintenance/opengever/maintenance/scripts/activate_solr.py
 
 
+Activating Solr sync
+~~~~~~~~~~~~~~~~~~~~
+
+Solr is able to sync every document to a remote solr.
```
buchi, line 232: This sound as if this is a Solr core functionality and could be mistaken with replication. Thus I would propose something more explict:
"The custom Solr sync allows to propagate document updates to another Solr. This can be enabled for specific portal types."

solr-conf/sync-solr.js
```diff
@@ -0,0 +1,336 @@
...
+
+  portalType = extractValueFromDoc(doc, "portal_type");
+  if (portalType.isAtomic === false) {
+    if (portalTypesToSync.indexOf(portalType.value) === -1) {
+      return;
+    }
+  }
+
+  // List of values sent to remote solr
+  updateDocument = fillPayload(updateDocument, doc, "SearchableText");
```
buchi, line 257: As only a subset of fields is sent to the remote Solr, it would make sense to make the fields configurable so this feature can be used in a more generic way. But for now this is good enough.

buchi (commented):
I have two things we need to discuss before merging:

Automatic testing seems to be hard. But we should at least have a documentation how to setup this locally and how to test it manually. E.g. a few curl commands that add a document to the local Solr and verify that it was also added to the remote Solr. Solrs could be easily setup with Docker.

What happens if the updating the remote Solr fails? E.g. because of a network failure or if the remote Solr is not available. It's important that local updates still work and there's no delay (see also my comment about making this a post-processor). Can you verify this and document it?

README.rst
bierik: 👍

solr-conf/solrconfig.xml
bierik: The order of the processor in the chain makes no difference when an error occurs. The whole processor is wrapped in a try catch block instead to prevent the further chain from crashing.

bierik:
@buchi I updated the readme with instructions for manual tests. May you have another look please.

buchi (commented):
@bierik I've tested the Solr Sync locally. Basically it works but I've noticed an issue with indexing of documents in the remote Solr. It seems that the SearchableText of documents is not indexed at all. Did this work in your tests?

This may be because ftw.solr does not explicitly commit when updating the SearchableText of documents, but relies on Solr's autoCommit.

Another note: As you do not index the field `trashed` on the remote Solr, it doesn't know if a document was trashed in GEVER. Not sure if this is relevant for RIS.

Everything seems to work if the remote Solr is not available.

buchi:
@bierik Bitte noch den Typo fixen und das Changelog. Der Eintrag ist noch im falschen Release.

PR #6577, opened 2020-08-03
Bump `ftw.solr` to also count missing `created` field as out of sync, add basic solr maintenance smoke tests.

deiferni:
Solr documents where the created field value is missing will now be treated as out of sync with portal catalog. This fixes an issue we had in production where documents that were only inserted partially in solr would not be picked up by `diff` and `sync`.

With this PR we also add some basic smoke tests for the maintenance view (and indirectly console-commands) of `ftw.solr` against a real solr, as `ftw.solr` currently has none. We also document the way we think incomplete documents were created and make sure that they appear in a solr `diff`.

For further reference:
- For a detailed description of the changes to `ftw.solr` https://github.com/4teamwork/ftw.solr/pull/173 for a detailed description of the changes
- For the root cause analysis https://4teamwork.atlassian.net/browse/GEVER-787

For https://4teamwork.atlassian.net/browse/GEVER-795

**Todo**
- [x] cut a release of https://github.com/4teamwork/ftw.solr/pull/173 and bump version in here

## Checklist (Must have)

- [x] Changelog entry


deiferni:
_already requesting a review despite the missing version bump, you'll have to trust me on that one 😏 _

PR #6578, opened 2020-08-05
Reindex SearchableText when filing_no is set.

njohner:
The `SearchableText` was not reindexed when the `filing_no` is set by the `Archiver`, although the  `filing_no`  is part of the `SearchableText`. Apart from this correction, I also realised that the advanced search did not work for filing numbers, as this referenced the `searchable_filing_number` field, which is not indexed in solr. I therefore now use the `filing_no` in the advanced search form instead.

The upgrade step reindexing the `SearchableText` will find dossiers only for deployments having this behavior (i.e. only Zug), as we query for the `IFilingNumberMarker` interface.

For https://4teamwork.atlassian.net/browse/GEVER-802

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/core/upgrades/20200805094934_reindex_searchable_text_to_include_filing_no/upgrade.py
+++ b/opengever/core/upgrades/20200805094934_reindex_searchable_text_to_include_filing_no/upgrade.py
@@ -1,0 +1,18 @@
+from ftw.upgrade import UpgradeStep
+from opengever.dossier.behaviors.filing import IFilingNumberMarker
+from opengever.dossier.behaviors.filing import IFilingNumber
+
+
+class ReindexSearchableTextToIncludeFilingNo(UpgradeStep):
+    """Reindex searchable text to include filing no.
+    """
+
+    deferrable = True
+
+    def __call__(self):
+        query = {'object_provides': IFilingNumberMarker.__identifier__}
+        for dossier in self.objects(query, 'Reindex searchable text to include filing no.'):
+            if IFilingNumber(dossier).filing_no:
+                # SearchableText is not in the catalog, so to avoid reindexing the
+                # full object, we also reindex the UID.
+                dossier.reindexObject(idxs=['UID', 'SearchableText'])
--- a/opengever/dossier/archive.py
+++ b/opengever/dossier/archive.py
@@ -320,7 +320,7 @@ class Archiver(object):
 
     def _recursive_update_prefix(self, dossier, prefix):
         IDossier(dossier).filing_prefix = prefix
-        dossier.reindexObject(idxs=['filing_no', 'searchable_filing_no'])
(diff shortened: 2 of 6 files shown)


opengever/core/upgrades/20200805094934_reindex_searchable_text_to_include_filing_no/upgrade.py
```diff
@@ -0,0 +1,16 @@
...
+    """Reindex searchable text to include filing no.
+    """
+
+    deferrable = True
+
+    def __call__(self):
+        query = {'object_provides': IFilingNumberMarker.__identifier__}
+        for dossier in self.objects(query, 'Reindex searchable text to include filing no.'):
+            if IFilingNumber(dossier).filing_no:
+                dossier.reindexObject(idxs=['SearchableText'])
```
deiferni, line 16: a word of warning here, `SearchableText` is now only in solr, we have removed the catalog index. this will lead plone catalog logic to pick all indexes instead. so best include a cheap one, like `UID`, too.

opengever/dossier/archive.py
```diff
@@ -320,7 +320,7 @@ def update_prefix(self, prefix):
 
     def _recursive_update_prefix(self, dossier, prefix):
         IDossier(dossier).filing_prefix = prefix
-        dossier.reindexObject(idxs=['filing_no', 'searchable_filing_no'])
+        dossier.reindexObject(idxs=['filing_no', 'searchable_filing_no', 'SearchableText'])
```
deiferni, line 323: this is a bit out of context, but have we moved the `filing_no` indexes to solr? or are they added dynamically? this code seems to be executed regardless of whether the filing profile is installed. so maybe plone catalog logic will potentially also reindex all indexes here. maybe we should also pick a cheap index to reindex, too, to guard against unnecessary full reindexes.

opengever/core/upgrades/20200805094934_reindex_searchable_text_to_include_filing_no/upgrade.py
njohner: Yes thanks!

opengever/dossier/archive.py
njohner: Ok I checked and no, this is only executed when the behavior is installed. The form is registered under `transition-archive` (https://github.com/4teamwork/opengever.core/blob/master/opengever/dossier/configure.zcml#L278-L283), and this transition is executed only when the context has the marker interface corresponding to the filing number: https://github.com/4teamwork/opengever.core/blob/master/opengever/dossier/resolve.py#L233-L252. So this should be safe

deiferni: ok, thanks for double checking 👍 !

PR #6581, opened 2020-08-10
Add bundle import support for workspace content

lukasgraf:
This implements import of workspace content via OGGBundle, according to the requirements in the Jira story.

The newly importable types are:
- `opengever.workspace.root`
- `opengever.workspace.workspace`
- `opengever.workspace.folder`

Workspace roots can either
- be supplied as part of the bundle, and be reference via `parent_guid` from workspaces
- or omitted from the bundle, in which case it is assumed exactly one workspace root already exists, and workspaces (without a `parent_guid`) will then be imported into that existing root

Jira: https://4teamwork.atlassian.net/browse/GEVER-81

##  Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- New functionality:


opengever/bundle/tests/test_teamraum_oggbundle_rootless_pipeline.py
```diff
@@ -0,0 +1,160 @@
...
+from zope.annotation import IAnnotations
+import pytz
+
+
+FROZEN_NOW = datetime(2016, 12, 20, 9, 40)
+
+
+class TestTeamraumOggBundleRootlessPipeline(IntegrationTestCase):
+
+    def remove_existing_workspace_content(self):
```
deiferni, line 25: you could move this to a common base-class for the teamraum bundle tests

opengever/bundle/tests/test_teamraum_oggbundle_rootless_pipeline.py
```diff
@@ -0,0 +1,160 @@
...
+        self.login(self.manager)
+
+        self.remove_existing_workspace_content()
+
+        # Create the 'bundle_guid' index. In production, this will be done
+        # by the "bin/instance import" command in opengever.bundle.console
+        add_guid_index()
+
+        # load pipeline
+        transmogrifier = Transmogrifier(api.portal.get())
```
deiferni, line 56: and maybe then also factor out and move up to baseclass the part where you run the bundle import

docs/public/dev-manual/oggbundle/index.rst
```diff
@@ -111,8 +119,30 @@ workspaceroots.json
...
+Falls auf der Installation, in welche ein OGGBundle mit Teamräumen importiert wird, bereits ein Teamraum-Root existiert, kann diese Datei weggelassen werden. Beim Import wird dann vorausgesetzt, dass genau ein Teamraum-Root bereits existiert, und die Teamräume werden in dieses Teamraum-Root importiert.
+
+In diesem Fall dürfen die Teamräume keine ``parent_guid`` gesetzt haben.
+
 JSON Schema: :ref:`workspaceroots.schema.json <workspaceroots_schema_json>`
 
+workspaces.json
+~~~~~~~~~~~~~~~
+
+Diese Datei beinhaltet ein oder mehrere Teamräume.
```
deiferni, line 131: ein -> einen

docs/public/dev-manual/oggbundle/index.rst
```diff
@@ -111,8 +119,30 @@ workspaceroots.json
...
+Die Teamräume werden über die ``parent_guid`` einem Teamraum-Root zugeordnet, welches ebenfalls im Bundle enthalten ist.
+
+Alternativ kann die ``parent_guid`` für Teamräume, und die Definition eines Workspace-Roots im ``workspaceroots.json`` weggelassen werden - die Teamräume werden dann in ein bereits existierendes Workspace-Root importiert.
+
+JSON Schema: :ref:`workspaces.schema.json <workspaces_schema_json>`
+
+workspacefolders.json
+~~~~~~~~~~~~~~~~~~~~~
+
+Diese Datei beinhaltet ein oder mehrere Teamraum-Ordner.
```
deiferni, line 142: einen

docs/public/dev-manual/oggbundle/index.rst
```diff
@@ -199,6 +229,35 @@ Berechtigungen
...
+Beispiel:
+
+.. code::
+
+  {
+  "guid": "9999-0000-0000-0000",
+  ...
+
+    "_permissions": {
+      "read": [
```
deiferni, line 243: should we mention that the possible permissions are type-specific?

opengever/bundle/schemas/workspaceroots.schema.json
```diff
@@ -0,0 +1,141 @@
...
+            ]
+        },
+        "permission": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "block_inheritance": {
+                    "type": "boolean"
+                },
+                "read": {
```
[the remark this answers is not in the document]
lukasgraf: Absolutely, as you noticed below 😉 I was debating wheter it would be more readable in the commit history to start with the standard permission schema (incorrect in this case) and show in the diff how it gets changed (which is what I did), or check in a "retouched" schema that omits them, even though the schemadump code at that time would produce them.

Aparently it's a confusing evolution either way 😉

docs/public/dev-manual/oggbundle/index.rst
```diff
@@ -217,6 +217,16 @@ Zusätzlich kann mit einem **block\_inheritance** Flag spezifiziert werden, ob d
...
 Berechtigungen werden an einen oder mehrere “Principals” vergeben, dies entspricht einem Benutzer oder einer Gruppe.
 
+--------------
+
+Für **Teamräume** gibt es separate Rollen welche auf unterschiedlichen Stufen gesetzt werden können.
+
+Auf der Ebene des Teamraum-Roots können folgende Rollen vergeben werden:
+
+-  ``workspaces_creator`` (Teamräume erstellen)
+-  ``workspaces_user`` (Teamräume auflisten)
```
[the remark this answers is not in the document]
lukasgraf: Good point, I'll keep this in mind when we start integrating with the export side, and doing first test runs.

opengever/bundle/tests/test_teamraum_oggbundle_rootless_pipeline.py
lukasgraf: I would rather not, to be honest. I thought about it, but the method does slightly different things in both test cases (need to remove workspace roots for one, but keep the root for the rootless tests in the other).

It was also quite intentional that I mechanically duplicated the existing test cases, with changing them as little as possible, comments and all, because this all needs to be cleaned up in a more fundamental way. Now starting to refactor one set of test classes, but not the other ones, would obfuscate their common origin IMHO. Like this at least it's painfully obvious that there's major code smells there.

PR #6582, opened 2020-08-13
Add one nesting level to generated policies

deiferni:
With this PR we add one more nesting level to all policies generated from the saas policy template.

We want to always have three levels of nesting for generated policies. The template only covers the simple case of a saas policy where the two inner package-names will be identical. Making it customizable or overriding it for potential teamraum deployments will be implemented separately, but is prepared by already using the adminunit-id for the innermost python package name. That way say a teamraum policy generator could easily override it by stetting a different admin-unit id, e.g. `tr`.

Testing is prepared in such a way that we could have multiple tests for each main policy profile with its own layer in `opengever.[customer].testing.py`. See https://github.com/4teamwork/opengever.ftw/pull/121 for an example of an implementation of two layers with separate tests for each policy profile (teamraum and GEVER).

I have tested the changes manually by:
- creating a policy from the new template
- buildouting it with `development.cfg` symlinked to `buildout.cfg`
- installing the policy
- running tests of the newly generated policy

Jira: https://4teamwork.atlassian.net/browse/GEVER-843

## Checklist (Must have)

- [x] Changelog entry


elioschmutz (approved):
👍

PR #6584, opened 2020-08-14
Add missing inboxes for multi orgunit setups in the examplecontent fd profile and testing fixture

elioschmutz:
This PR updates the example content profile and the testing fixture to add separate inboxes for each org-unit within an inbox container.

This is the expected setup for a multi-orgunit setup.

Jira: https://4teamwork.atlassian.net/browse/GEVER-490

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/examplecontent/profiles/default_content/content_creation/01_default_content.json
+++ b/opengever/examplecontent/profiles/default_content/content_creation/01_default_content.json
@@ -1,0 +1,42 @@
+[
+    {
+        "_path": "private",
+        "_type": "opengever.private.root",
+        "title_de": "Meine Ablage",
+        "title_fr": "Mon dépôt",
+        "_placefulworkflow": ["opengever_private_policy",
+                              "opengever_private_policy"]
+    },
+    {
+        "_path": "eingangskorb",
+        "_type": "opengever.inbox.container",
+        "title_de": "Eingangskorb",
+        "title_fr": "Boîte de réception",
+        "_placefulworkflow": ["opengever_inbox_policy",
+                              "opengever_inbox_policy"],
+        "_ac_local_roles": {
+            "og_demo-ftw_users": [
+                "Reader"
+            ]
+        }
+    },
+    {
+        "_path": "vorlagen",
+        "_type": "opengever.dossier.templatefolder",
+        "title_de": "Vorlagen",
+        "title_fr": "Modèles"
+    },
+    {
+        "_path": "kontakte",
+        "_type": "opengever.contact.contactfolder",
+        "title_de": "Kontakte",
+        "title_fr": "Contacts",
+        "_transitions": "publish"
+    },
+    {
+        "_path": "workspaces",
+        "_type": "opengever.workspace.root",
+        "title_de": "Teamräume",
+        "title_fr": "Espace partagé"
+    }
+]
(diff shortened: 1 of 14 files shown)


opengever/examplecontent/profiles/default_content/content_creation/01_default_content.json
```diff
@@ -14,6 +14,7 @@
         "title_fr": "Boîte de réception",
         "_placefulworkflow": ["opengever_inbox_policy",
                               "opengever_inbox_policy"],
+        "_block-local-roles": true,
```
njohner, line 17: why do we want to block the local roles?

opengever/testing/fixtures.py
```diff
@@ -927,6 +927,8 @@ def create_inbox_container(self):
         policy_config = pwf_tool.getWorkflowPolicyConfig(self.inbox_container)
         policy_config.setPolicyIn(inbox_policy_id, update_security=False)
         policy_config.setPolicyBelow(inbox_policy_id, update_security=False)
+        self.inbox_container.__ac_local_roles_block__ = True
+        self.set_roles(self.inbox_container, self.secretariat_user.getId(), ['Reader'])
```
njohner, line 931: Same here. I don't think that we block local roles inheritance on the inboxes in production

opengever/examplecontent/profiles/default_content/content_creation/01_default_content.json
elioschmutz: Only the explicitly defined users/groups should have access to the inbox. Thus we need to block the localroles...

opengever/testing/fixtures.py
elioschmutz: We have to. Otherwise it seems to be missconfigured. Also if we have multiple org-units, it's necessary to block the local roles on each inbox.

elioschmutz: Each inbox-group have read access on the container, if we don't block it on the inbox-level, the inbox-group of inbox-a would have read access on inbox-b

opengever/examplecontent/profiles/default_content/content_creation/01_default_content.json
elioschmutz: I think @phgross knows it best, how to configure the inboxes properly... perhaps he can disclose the secret...

elioschmutz: @phgross do you agree, that inbox-containers and inboxes need to block the localroles?

phgross: I can try 😅 ...The local_roles need to be blocked on every inbox, because the access should on a specific inbox, should only possible for users of the inbox group. Without a block of the local-roles inheritance, the local roles of the inbox container would allow every inbox users to access each inbox.

phgross: The inbox container does not need a block.

elioschmutz: @phgross thanks!

@njohner I'll remove the block for containers 👍

elioschmutz:
@njohner 👍 I updated the pr. I do no longer block the local roles for inbox containers.

njohner (approved):
LGTM

PR #6586, opened 2020-08-17
Reindex modified containers after bundle import.

njohner:
In the oggbundle import pipeline, object reindexing is deactivated to avoid objects being reindexed several times during the import.
Every object is indexed just once in the pipeline, but containers into wich objects are added were not reindex afterwards, leading to a discrepency in the modified date on the object and in the catalog. This in turn leads to problems with solr diff / sync.

We also add an upgrade step, which identifies the concerned objects by their symptom of having a different modified date in the catalog than on the object, and hence in solr...

For https://4teamwork.atlassian.net/browse/GEVER-190

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

### DIFF
--- a/opengever/bundle/sections/reindex_containers.py
+++ b/opengever/bundle/sections/reindex_containers.py
@@ -1,0 +1,48 @@
+from collective.transmogrifier.interfaces import ISection
+from collective.transmogrifier.interfaces import ISectionBlueprint
+from collective.transmogrifier.utils import traverse
+from opengever.bundle.sections.bundlesource import BUNDLE_KEY
+from plone import api
+from zope.annotation import IAnnotations
+from zope.interface import classProvides
+from zope.interface import implements
+import logging
+import transaction
+
+
+log = logging.getLogger('opengever.bundle.reindex_containers')
+log.setLevel(logging.INFO)
+
+
+class ReindexContainersSection(object):
+    """Reindex specific indexes for containers into which objects
+    were added during bundle import.
+    """
+
+    classProvides(ISectionBlueprint)
+    implements(ISection)
+
+    def __init__(self, transmogrifier, name, options, previous):
+        self.previous = previous
+        self.transmogrifier = transmogrifier
+        self.site = api.portal.get()
+        self.bundle = IAnnotations(transmogrifier)[BUNDLE_KEY]
+        self.indexes = ['modified']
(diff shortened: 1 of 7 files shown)


opengever/bundle/sections/reindex_containers.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def __iter__(self):
+        for item in self.previous:
+            yield item
+
+        n_containers = len(self.bundle.containers_to_reindex)
+        log.info("Reindexing {} containers after bundle import...".format(n_containers))
+
+        for obj in self.bundle.containers_to_reindex:
+            obj.reindexObject(idxs=self.indexes)
```
lukasgraf, line 37: This will also update all metadata for all those containers. I think we'll want to use  `portal_catalog.reindexObject(..., update_metadata=False)` instead. That won't reindex into Solr though. That might could be addressed by manually using the `ISolrIndexHandler` like [we already did in other places](https://github.com/4teamwork/opengever.core/blob/master/opengever/core/upgrades/20200427154326_always_use_solr_for_table_source/upgrade.py#L40-L41).

opengever/bundle/sections/reindex_containers.py
```diff
@@ -0,0 +1,43 @@
...
+from collective.transmogrifier.interfaces import ISectionBlueprint
+from opengever.bundle.sections.bundlesource import BUNDLE_KEY
+from zope.annotation import IAnnotations
+from zope.interface import classProvides
+from zope.interface import implements
+import logging
+import transaction
+
+
+log = logging.getLogger('opengever.bundle.reindex_parent')
```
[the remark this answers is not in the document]
njohner: Oh yes, right

opengever/bundle/sections/reindex_containers.py
njohner: `modified` in the metadata is also wrong, so I think we actually need to reindex all the metadata

njohner:
Ok amended according to your comments. I'm still reindexing the whole of the metadata, because modified is wrong there too.
I've also modified the upgrade step to only query for content types that might have been imported and are containerish.

lukasgraf (approved):

lukasgraf:
Rebased and merged on your behalf.

PR #6587, opened 2020-08-18
Complement @role-assignment-reports responses with type, principal label, title and available roles.

tinagerber:
To implement the role assignment reports in the frontend, the `@role-assignment-reports` had to be complemented with some information:

- A single report has been supplemented with the information `@type`, `available_roles`, and `principal_label`
- For each item of the report the `title` of the element is now also provided
- To maintain local consistency, `reportid` and `principalid` were renamed to `report_id` and `principal_id`.

Jira: https://4teamwork.atlassian.net/browse/GEVER-547

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:


opengever/api/role_assignment_reports.py
```diff
@@ -47,15 +48,20 @@ def reply(self):
...
             try:
                 result = storage.get(report_id)
             except KeyError:
                 raise BadRequest("Invalid report_id '{}'".format(report_id))
             for item in result['items']:
+                available_roles.update(item['roles'])
                 obj = uuidToObject(item['UID'])
                 item['title'] = obj.Title()
                 item['url'] = obj.absolute_url()
+            result['available_roles'] = [{'id': role, 'title': translate(
```
njohner, line 62: For consistency with the `@watchers` endpoint, I'd rather call that `referenced_roles`

docs/HISTORY.txt
```diff
@@ -5,6 +5,7 @@ Changelog
 ---------------------
 
 - Extend the @config endpoint with the current inbox_folder_url. [elioschmutz]
+- Complete @role-assignment-reports with type, principal label, title and available roles. [tinagerber]
```
njohner, line 8: I'd rephrase that:
`- Add type, principal label, title and available roles fields to the @role-assignment-reports.`
or
`- Complement @role-assignment-reports responses with type, principal label, title and available roles.`
Or something along these lines. I don't think `complete` is the verb you want here.

njohner (commented):
LGTM, just a few minor comments. Also:
- I think the endpoint is not documented?
- Commit message of last commit: "Make test more robust." -> "Make role assignment report test more robust."
- PR title should be updated similary to changelog.

opengever/api/role_assignment_reports.py
tinagerber: In the `@sharing` endpoint it is named `available_roles`, so I named it that way, but `referenced_roles` is probably more appropriate.

njohner: right, but in the `@sharing` it has a different meaning I think. It actually is the list of `available_roles`, i.e. the ones that are managed and that you can therefore assign to principals. Here it really is the list of roles that were referenced

tinagerber:
@njohner I have implemented the desired changes.
You can find the documentation here: https://github.com/4teamwork/opengever.core/blob/35a20855e1abb7b9eee616d6f15c18dc418dd4d4/docs/public/dev-manual/api/role_assignment_reports.rst
Can you rereview?

njohner (approved):
LGTM

njohner:
Also notice that I've added the API-change label

PR #6589, opened 2020-08-18
Make the portal_url configurable through the portal_registry.

elioschmutz:
This PR adds a new registry entry to define the portal_url.

By default, the portal url refers to `[admin-unit-url]/portal`(i.e. https://mygever.com/portal). In rare cases where the portal is not running under the same domain like the gever, we have to define it separately.

In Jira, there is a suggestion to reuse the cas-auth url from the auth-plugin to guess the portal url. This could have been one possible solution. But guessing will lead once again into problems in the future. So I introduced a new setting to override the default `portal_url`.

Currently, the `portal_url` will be exposed in the @config endpoint and will be used to generate the invitation url for teamraum invitations.

Jira: https://4teamwork.atlassian.net/browse/GEVER-860

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/base/interfaces.py
+++ b/opengever/base/interfaces.py
@@ -431,3 +431,13 @@ class IRoleAssignmentReportsStorage(Interface):
     def delete(report_uid):
         """Delete the corresponding role assignment report.
         """
+
+
+class IPortalSettings(Interface):
+
+    portal_url = schema.TextLine(
+        title=u'Portal url',
+        description=u'This url will be exposed in the @config endpoint and will '
+                    u'be used to generate the workspace invitation url. '
+                    u'Leave it blank to use an auto generated portal url to '
+                    u'[cluster-base-url]/portal')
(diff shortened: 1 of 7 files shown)


opengever/base/interfaces.py
```diff
@@ -431,3 +431,13 @@ def list():
...
+
+
+class IPortalSettings(Interface):
+
+    portal_url = schema.TextLine(
+        title=u'Portal url',
+        description=u'This url will be exposed in the @config endpoint and will '
+                    u'be used to generate the workspace invitation url. '
+                    u'Leave it blank to use an auto generated portal url to '
+                    u'[cluset-base-url]/portal')
```
njohner, line 443: cluset -> cluster ?

njohner (approved):
LGTM

PR #6594, opened 2020-08-23
Wopi fixes

buchi:
Two fixes required to make WOPI validator tests pass.

1. Unghostify blob before reading its serial, which is used as file version. If the blob is a ghost, its serial is 0. This seems to happen in a ZEO setup when no attributes were accessed before on the blob. 

2. Implement proof key validation (https://wopi.readthedocs.io/en/latest/scenarios/proofkeys.html)
According to the documentation proof key validation is optional, but the validator always runs those tests, which results in test failures.
Proof key validation provides additional security as every request to the WOPI endpoints is digitally signed. The signature is calculated from the access token, the URL and a timestamp. Requests with a timestamp older than 20 minutes are considered invalid.
In order to implement proof key validation I've refactored the WOPI discovery. Now discovery is rerun every 24 hours because the proof keys that are provided by WOPI discovery may change from time to time.

Jira: https://4teamwork.atlassian.net/browse/GEVER-852

Bildschirmfoto 2020-08-24 um 09 49 17

## Checklist (Must have)

- [x] Changelog entry


opengever/document/tests/test_fileactions.py
```diff
@@ -73,8 +74,10 @@ def test_editable_if_enabled_and_supported_document(self):
         settings = registry.forInterface(IWOPISettings)
         settings.enabled = True
         settings.discovery_url = u'http://localhost/hosting/discovery'
-        discovery._EDITABLE_EXTENSIONS = {
-            'http://localhost/hosting/discovery': set(['docx', 'xlsx', 'pptx'])
+        discovery._WOPI_DISCOVERY = {
```
deiferni, line 77: there is a lot of repetition, you could factor this out into a test-helper.

opengever/wopi/browser/wopi.py
```diff
@@ -110,6 +112,12 @@ def __call__(self):
         # Disable CSRF protection - WOPI client can't supply a CSRF token
         alsoProvides(self.request, IDisableCSRFProtection)
 
+        if not self.validate_proof():
+            self.request.response.setStatus(500)
```
deiferni, line 116: this is a server error because discovery would somehow have failed if validation fails here, correct?

deiferni (approved):
LGTM 👍 , just some minor comments/suggestion.

I'm not sure how wopi/discovery actually works so it's hard to judge from that point 😅 .

opengever/wopi/browser/wopi.py
buchi: This is because of Microsoft's dumb specification: https://wopi.readthedocs.io/en/latest/scenarios/proofkeys.html?highlight=500%20Internal%20Server%20Error 🤦 
But of course the semantic correct error would be a 400 Bad Request.

opengever/document/tests/test_fileactions.py
buchi: will do that

opengever/wopi/browser/wopi.py
deiferni: yeah that was my suspicion, oh well 🙈 .

opengever/document/tests/test_fileactions.py
buchi: @deiferni added helper to setup WOPI discovery for tests

PR #6595, opened 2020-08-23
Add API endpoints for accepting and completing remote tasks

lukasgraf:
This change introduces three new endpoints:

- `POST @remote-workflow` on the Plone site root
- `POST @accept-remote-task` on dossiers
- `POST @complete-successor-task` on tasks

These are endpoints required in the backend to facilitate the process of accepting inter-admin-unit tasks, and finally resolving them, via the new frontend.

The underlying idea is that the frontend should need to know as little as possible about the detailed mechanics of inter-admin-unit processes, remote requests, and authenticating to non-local admin units.

These endpoints therefore proxy or internally dispatch (via server-to-server requests) the necessary actions, allowing the frontend to perform all actions from a single, current admin unit.

Using these endpoints, the processes of accepting and completing inter-admin-unit tasks can be implemented as follows.

*(In these examples, `rk` is always considered the current, local admin unit, and `fd` the remote AU, where a task has been assigned to us)*

### Accepting a task

-----
accept_choices

*(Refresher what this looks like in the classic UI)*

-----

#### 1) Accepting the task on the remote unit where it physically resides.

This is the simplest case. In this case the `@remote-workflow` endpoint can be used to invoke a **workflow transition on a remote object** (referenced by `remote_oguid`) from the current admin unit in order to accept that task remotely.

```http
    POST /rk/@remote-workflow/some-transition

    {
      "remote_oguid": "fd:12345",
      "text": "I accept this task"
    }
```

#### 2) Accept the remote task into an existing local dossier.

This can be accomplished using the `@accept-remote-task` endpoint on the local target dossier where the successor task should be copied to (including the documents attached to the remot predecessor task).

```http
    POST /rk/dossier-17/@accept-remote-task

    {
      "task_oguid": "fd:12345",
      "text": "I'm accepting this task into this dossier."
    }
```

#### 3) Accept the remote task into a new local dossier

Since the frontend will handle the creation of the new dossier, this case can simply be reduced to case 2) from a backend perspective.

### Completing the successor task

For cases 2) and 3), where a successor task is created locally, the completion of that successor will involve also completing the predecessor on the remote side, and optionally delivering selected documents. This can be done using the `@complete-successor-task` on the local successor task:

```http
    POST /rk/task-42/@complete-successor-task

    {
      "transition": "task-transition-in-progress-resolved",
      "documents": [1423795951],
      "text": "I finished this task."
    }

```

Jira: https://4teamwork.atlassian.net/browse/GEVER-270

(Also: https://4teamwork.atlassian.net/browse/GEVER-269)

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
- New functionality:
     (out of scope, separate story for forwardings)

### DIFF
--- a/opengever/api/accept_remote_task.py
+++ b/opengever/api/accept_remote_task.py
@@ -1,17 +1,13 @@
+from opengever.api.remote_task_base import RemoteTaskBaseService
 from opengever.base.exceptions import MalformedOguid
 from opengever.base.oguid import Oguid
 from opengever.task.browser.accept.utils import accept_task_with_successor
 from plone.protect.interfaces import IDisableCSRFProtection
-from plone.restapi.deserializer import json_body
-from plone.restapi.interfaces import ISerializeToJson
-from plone.restapi.services import Service
 from zExceptions import BadRequest
-from zope.component import queryMultiAdapter
 from zope.interface import alsoProvides
-import json
 
 
-class AcceptRemoteTaskPost(Service):
+class AcceptRemoteTaskPost(RemoteTaskBaseService):
     """Accepts a task on a remote admin unit by creating a successor copy
     in the dossier the endpoint was invoked on.
 
@@ -34,8 +30,8 @@ class AcceptRemoteTaskPost(Service):
     *must* be on a different admin unit than the current one.
     """
 
-    def _format_exception(self, exc):
-        return '%s: %s' % (exc.__class__.__name__, str(exc))
+    required_params = ('task_oguid', )
+    optional_params = ('text', )
 
     def is_remote(self, oguid):
         return not oguid.is_on_current_admin_unit
@@ -44,23 +40,11 @@ class AcceptRemoteTaskPost(Service):
         # Disable CSRF protection
         alsoProvides(self.request, IDisableCSRFProtection)
 
-        # Extract and validate parameters
-        json_data = json_body(self.request)
+        # Extract and validate parameters, assign defaults
+        params = self.extract_params()
 
-        raw_task_oguid = json_data.pop('task_oguid', None)
-        if not raw_task_oguid:
-            raise BadRequest(
-                'Required parameter "task_oguid" is missing in body')
-
-        response_text = json_data.pop('text', u'')
-
-        # Reject any unexpected parameters
-        unexpected_params = json_data.keys()
-        if unexpected_params:
-            raise BadRequest(
-                'Unexpected parameter(s) in JSON '
-                'body: %s. Supported parameters are: '
-                '["task_oguid", "text"]' % json.dumps(unexpected_params))
+        raw_task_oguid = params['task_oguid']
+        response_text = params.get('text', u'')
 
         # Validate that Oguid is well-formed and refers to a task that is
         # actually remote. Turn any Python exceptions into proper
@@ -68,7 +52,7 @@ class AcceptRemoteTaskPost(Service):
         try:
             task_oguid = Oguid.parse(raw_task_oguid)
         except MalformedOguid as exc:
-            raise BadRequest(self._format_exception(exc))
+            raise BadRequest(self.format_exception(exc))
 
         if not self.is_remote(task_oguid):
             raise BadRequest(
@@ -87,7 +71,5 @@ class AcceptRemoteTaskPost(Service):
         self.request.response.setStatus(201)
         self.request.response.setHeader("Location", successor.absolute_url())
 
-        serializer = queryMultiAdapter((successor, self.request), ISerializeToJson)
-        serialized_successor = serializer()
-        serialized_successor["@id"] = successor.absolute_url()
+        serialized_successor = self.serialize(successor)
         return serialized_successor
--- a/opengever/api/accept_remote_task.py
+++ b/opengever/api/accept_remote_task.py
@@ -1,0 +1,93 @@
+from opengever.base.exceptions import MalformedOguid
+from opengever.base.oguid import Oguid
+from opengever.task.browser.accept.utils import accept_task_with_successor
+from plone.protect.interfaces import IDisableCSRFProtection
+from plone.restapi.deserializer import json_body
+from plone.restapi.interfaces import ISerializeToJson
+from plone.restapi.services import Service
+from zExceptions import BadRequest
+from zope.component import queryMultiAdapter
+from zope.interface import alsoProvides
+import json
+
+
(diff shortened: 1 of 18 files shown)


lukasgraf:
@all4manu breaking change compared to what you saw last:
The field for the comment / response is now always called `text`, in all three endpoints *(it was named `response_text` before for the accept endpoint)*.

Other than that, this should be pretty much exactly what we discussed.

opengever/base/oguid.py
```diff
@@ -71,6 +73,32 @@ def get_url(self):
         from opengever.base.browser.resolveoguid import ResolveOGUIDView
         return ResolveOGUIDView.url_for(self, admin_unit=admin_unit)
 
+    def get_remote_url(self, public=False):
+        """
```
deiferni, line 77: delete empty docstring

opengever/base/oguid.py
```diff
@@ -71,6 +73,32 @@ def get_url(self):
...
+        """
+        if self.admin_unit_id == get_current_admin_unit().id():
+            raise NonRemoteOguid('Not a remote OGUID. Use get_url() instead')
+
+        admin_unit = ogds_service().fetch_admin_unit(self.admin_unit_id)
+
+        if public:
```
deiferni, line 84: i did not find a place where we use `public=True`, can we get rid of the parameter (or test it separately)?

opengever/api/remote_workflow.py
```diff
@@ -0,0 +1,110 @@
...
+        except MalformedOguid as exc:
+            raise BadRequest(self._format_exception(exc))
+
+        try:
+            remote_obj_url = remote_obj_oguid.get_remote_url()
+        except (UnsupportedTypeForRemoteURL, NonRemoteOguid) as exc:
+            raise BadRequest(self._format_exception(exc))
+
+        # Extract and forward path parameters (i.e., the transition)
+        path_params = '/'.join(self.path_params)
```
deiferni, line 80: this service mixes how it receives relevant parameters, the workflow-name is in the request path, the oguid in the json body. you did this to keep the interface of `@workflow` consistent, correct?

opengever/api/accept_remote_task.py
```diff
@@ -0,0 +1,93 @@
...
+        return not oguid.is_on_current_admin_unit
+
+    def reply(self):
+        # Disable CSRF protection
+        alsoProvides(self.request, IDisableCSRFProtection)
+
+        # Extract and validate parameters
+        json_data = json_body(self.request)
+
+        raw_task_oguid = json_data.pop('task_oguid', None)
```
deiferni: should we consistently name the parameter `remote_oguid`? it may be confusing for the api-user if the name is different 🤔 , also it would allow to extract oguid validation into a central place.

opengever/api/configure.zcml
```diff
@@ -861,4 +861,13 @@
       layer="opengever.base.interfaces.IOpengeverBaseLayer"
       />
 
+  <plone:service
+      method="POST"
+      name="@accept-remote-task"
+      for="opengever.dossier.behaviors.dossier.IDossierMarker"
+      factory=".accept_remote_task.AcceptRemoteTaskPost"
+      permission="cmf.ModifyPortalContent"
```
deiferni, line 869: shouldn't this be `cmf.AddPortalContent` as we are adding new content locally?

opengever/task/browser/complete.py
```diff
@@ -44,6 +41,7 @@
 from zope.schema.vocabulary import SimpleVocabulary
 import AccessControl
 import json
+from opengever.task.browser.complete_utils import complete_task_and_deliver_documents
```
deiferni, line 44: unsorted imports

deiferni (approved):
LGTM 👍 , thanks for the good and reviewable commits 🎉 .

There are some mostly optional comments. Only the permission in the accept-service needs a double-check.

opengever/base/oguid.py
lukasgraf: Oops. I think I'll write one instead 😉

opengever/base/oguid.py
lukasgraf: That's correct. Currently, this is only used for server-to-server requests, therefore the `site_url` is the appropriate one. I didn't want to default to `site_url` though without making it obvious that this is what happens. Otherwise someone might use that method in the future in a context where `public_url` would be appropriate.

I'll add a separate test for this.

opengever/api/remote_workflow.py
lukasgraf: Correct. This mirrors the fact that the `@workflow` endpoint receives the transition as a path-parameter.

opengever/api/accept_remote_task.py
lukasgraf: I thought about it. However, after some back and forth I decided against it, because the (`@remote-workflow`) vs. (`@accept-remote-task`, `@complete-successor-task`) endpoints implement two different concepts.

The first has a somewhat generic underlying base: `remote_oguid` could (in theory) refer to an object of any type, and could (and should) be used the same way in potentially other `@remote-<...>` endpoints.

The latter two are tightly bound to tasks though (as indicated by their names as well). They are highly custom, they'll be limited in number (maybe one more in the future for forwardings), and they can't possibly work for anything else than tasks. Because of that, I see some value in the property name being specific (and more descriptive, by referring to a `task_...`) instead of generic.

In terms of API usage, any code that invokes these two endpoints most likely will be as custom as the behavior they trigger, and can't be (shouldn't be, IMHO) any generalized code that just happens to invoke those.

opengever/api/configure.zcml
lukasgraf: True 👍

lukasgraf:
@deiferni FYI, the failing tests were indeed for the reason you suspected. For Role Assignment Reports,`report['items']` order is dependent on a catalog query with imlicit ordering by `modified`, which can get flaky of its limited precision of one second. Will create a follow up PR that makes that order stable _(instead of changing the tests to not assert on a specific order)_.

Also confirmed by the fact that I was able to re-trigger the tests for the previously failing commits in this PR, and they now passed 😉

opengever/api/accept_remote_task.py
deiferni: i was also going back and forth before writing the comment, so it probably does not matter too much and we need to decide to go one way. i'm fine with the way it is right now, too 👍 .

PR #6596, opened 2020-08-24
Fix filtering with exclusion filters if the field has a mapping.

tinagerber:
If a query was submitted with an excluding filter (e.g. -keywords=bla), the correct ListingField was not found until now, because only the field name "keywords" is known. 

Therefore fields with mapping are now duplicated and the field name and index are preceded by a minus sign, so that the fields can also be used as exclusion filters.

Jira: https://4teamwork.atlassian.net/browse/GEVER-699

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 ---------------------
 
 - Add policy template for teamraum policies. [njohner]
+- Fix filtering with exclusion filters if the field has a mapping. [tinagerber]
 - Make the portal_url configurable through the portal_registry. [elioschmutz]
 - Include OGUID in all API content GET responses. [lgraf]
 - Extend the @config endpoint with the current inbox_folder_url. [elioschmutz]
--- a/opengever/api/solr_query_service.py
+++ b/opengever/api/solr_query_service.py
@@ -1,5 +1,6 @@
 from collective.elephantvocabulary import wrap_vocabulary
 from copy import copy
+from copy import deepcopy
 from DateTime import DateTime
 from DateTime.interfaces import DateTimeError
 from ftw.solr.converters import to_iso8601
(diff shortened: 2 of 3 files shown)


lukasgraf (changes requested):
This seems a bit backwards to me, to be honest.
Instead of populating `FIELDS_WITH_MAPPING` with double the number of fields, half of them "fake" fields that don't really exist, shouldn't the code be fixed in those places where the field lookup happens, to remove any operators like `-` from the field name before the lookup happens?

elioschmutz:
@lukasgraf I discussed this with @tinagerber 

The problem here is, that the `FIELDS_WITH_MAPPING` contains instances of `ListingField`. It's not enough to just lookup it without `-`. We have to know in the ListingFiled-Instance, if it's negated or not because we need this information later for the request. Adding a property, i.e. `negate` to the `ListingField` would not work because it would update it for all fields with the same name. The instances are made at boot-time, so they're shared between all requests.

Duplicate the fields with a negated name and index seemed to be the best way to solve this problem.

Perhaps you have a good idea how to solve it without refactoring the whole implementation?

lukasgraf:
@elioschmutz I see. I would go with this approach for now then, but at least add a comment about why it's necessary to duplicate the fields with negated copies.

A `negate` property on `ListingField` would be just as wrong in my opinion 😉 These are field definitions, including alias mappings for alternative names. A *Field* however is never negated. The *query* against that field is. So for a future refactoring this should be solved on the query parsing side IMHO. They field is always `keywords`. But the query might contain a negation operator `-`.

So when parsing the incoming query, we currently seem to be doing a naive approach where the query is split into <QueryField> <QueryValue>. Whereas it should split into something like <QueryOperator><QueryField><QueryValue>. Then, <QueryField> could be *mapped* according to `MAPPED_FIELDS` like before, and afterwards the approprate Solr Query could be constructed from operator, mapped field name and query value.

tinagerber:
@lukasgraf I added now a comment.

lukasgraf (approved):
Thanks! LGTM! 👍

PR #6597, opened 2020-08-24
Role Assignment Reports: Ensure stable sort order for report items:

lukasgraf:
Because the `items` of role assignment reports are based on **catalog queries**, their order is dependent on the order of those catalog results. Because the catalog query didn't include an explicit `sort_on` parameter, it defaulted to sorting on `modified`.

But since the precision of the modified index is **limited to one second**, this can lead to an undefined sort order if two items have been created within the same second (as may happen in our test fixture).

The order of the reports' items therefore wasn't stable in all cases, and has lead to some flaky tests, which is fixed by explicitly ordering by path.

## Checklist (Must have)

- [x] Changelog entry


deiferni (approved):
LGTM 👍

PR #6598, opened 2020-08-25
Bring @resolve-oguid error responses in line with REST API style

lukasgraf:
This brings the format of error responses for the `@resolve-oguid` endpoint in the line with the response format used in the other parts of the REST API:

Return `400 Bad Request` with a JSON body including error details, instead of returning `400` with a custom status line reason text.

In addition, this allowed to also homogenize some details about how the two proxying endpoints `@resolve-oguid` and `@remote-workflow` behave, and factor out their commonalities into a base class.  

Follow up to: https://github.com/4teamwork/opengever.core/pull/6595

Jira: https://4teamwork.atlassian.net/browse/GEVER-270

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [x] api-change label added
    *(deemed unnecessary, will coordinate with frontend guys directly)*

### DIFF
--- a/opengever/api/proxy_base.py
+++ b/opengever/api/proxy_base.py
@@ -1,0 +1,52 @@
+from opengever.ogds.base.utils import get_current_admin_unit
+from plone import api
+from plone.restapi.services import Service
+from urlparse import parse_qs
+from zExceptions import InternalError
+import requests
+
+
+class ProxyingEndpointBase(Service):
+    """Base class for endpoints that proxy request to remote admin units.
+    """
+
+    def extract_query_string_params(self):
+        parsed = parse_qs(self.request['QUERY_STRING'])
+        qs_params = dict((k, v if len(v) > 1 else v[0])
+                         for k, v in parsed.iteritems())
+        return qs_params
+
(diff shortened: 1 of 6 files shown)


opengever/api/proxy_base.py
```diff
@@ -0,0 +1,65 @@
...
+    """Base class for endpoints that proxy request to remote admin units.
+    """
+
+    implements(IPublishTraverse)
+
+    def __init__(self, context, request):
+        super(ProxyingEndpointBase, self).__init__(context, request)
+        self.path_params = []
+
+    def publishTraverse(self, request, name):
```
deiferni, line 21: This will cause `@resolve-oguid` to accept arbitrary path parameters, even though unused. IMO it would make more sense to leave it in the workflow endpoint as it is only used there at the moment, otherwise we generalize this  early.

deiferni (approved):
Thx, LGTM 👍 .

One optional comment for discussion, you can :shipit: as is if you are strongly leaning the other way.

PR #6599, opened 2020-08-25
Fix translated review state for meeting content

lukasgraf:
Some meeting content like proposals have a custom workflow implementation that's layered on top of the Plone workflow.
For these, we need to use `obj.get_state()` to access the real workflow state that's in play, and translate that.

Jira: https://4teamwork.atlassian.net/browse/GEVER-808

## Checklist (Must have)

- [x] Changelog entry


deiferni (approved):
LGTM 👍

PR #6600, opened 2020-08-25
Filter out owner role

tinagerber:
In this PR, the owner role is filtered out of the role assignment reports and "Rollenzuweisungsbericht" is renamed to "Berechtigungsreport" in the API documentation.

Jira: https://4teamwork.atlassian.net/browse/GEVER-547

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/base/nightly_role_assignment_reports.py
+++ b/opengever/base/nightly_role_assignment_reports.py
@@ -77,7 +77,9 @@ class NightlyRoleAssignmentReports(object):
             sharing_assignments = [assignment for assignment in assignments
                                    if assignment.cause == ASSIGNMENT_VIA_SHARING]
             if sharing_assignments:
-                items.append({'UID': obj.UID(), 'roles': sharing_assignments[0].roles})
+                roles = filter(lambda r: r != 'Owner', sharing_assignments[0].roles)
+                if roles:
+                    items.append({'UID': obj.UID(), 'roles': roles})
 
         report_data = {'state': STATE_READY, 'items': items}
         self.storage.update(reportid, report_data)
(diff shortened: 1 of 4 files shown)


opengever/base/nightly_role_assignment_reports.py
```diff
@@ -76,8 +76,11 @@ def run_job(self, job, interrupt_if_necessary):
             assignments = RoleAssignmentManager(obj).get_assignments_by_principal_id(principalid)
             sharing_assignments = [assignment for assignment in assignments
                                    if assignment.cause == ASSIGNMENT_VIA_SHARING]
-            if sharing_assignments:
-                items.append({'UID': obj.UID(), 'roles': sharing_assignments[0].roles})
+            for ass in sharing_assignments:
+                roles = ass.roles
```
deiferni, line 80: 😂

deiferni (approved):
LGTM 👍 , but:

- no test was updated? is this functionality untested? could you provide a simple test?

tinagerber:
I realized that the earlier variant (`items.append({'UID': obj.UID(), 'roles': sharing_assignments[0].roles})`) was correct after all, since there cannot be more than one assignment: Since in the init function of `SharingRoleAssignment` the reference is set to null and in the `add_or_update` method of `RoleAssignmentStorage` only a new entry is created if the reference is different, a new role assignment via sharing always updates the existing entry. 
Therefore, I have only filtered out the owner.

lukasgraf:
@tinagerber sounds good - sorry, I didn't catch that when we looked at it together 🤦

lukasgraf (approved):
Changelog seems to have suffered some rebase-damage, otherwise LGTM 👍

PR #6601, opened 2020-08-25
Fix empty bytestring overriding unicode in some cases. Fix existing content.

deiferni:
With this PR we fix our default_value patches not correctly detecting the difference between `''` and `u''` in some cases. The issue has surfaced due to the change introduced in https://github.com/4teamwork/opengever.core/pull/6563/ where we set values via their field during `createContentInContainer`. This will perform a type-check and fail if the type is not correct.

The issue appeared in production when creating a dossier from a templatedossiers with a sub-template dossiers with only the description default value. See https://sentry.4teamwork.ch/sentry/onegov-gever/issues/68884/ for corresponding sentry issue. See https://4teamwork.atlassian.net/browse/GEVER-895 for steps to reproduce.

**To tackle the issue i have:**
- added type checks to our default value tests
- fixed where the schema interfaces did not specify `default` and `missing_value`
- fixed some cases where we use incorrect types in our business logic
- changed defaults used in our default value tests, so that they document which type is used
- in some cases inconsistent types are still used when creating content via `plone.restapi`, but the tests now document where this happens
- i could get rid of some existing workarounds in tests which indicates the changes go the right way 🎉 

**To fix the issues in production:**
- there is an upgrade-step making sure dossier templates now use the correct type for description
- i have not provided an upgrade step for all other potentially affected types as the risk to touch almost all content exceeds the value of having consistent types. it should™️  not be an issue in production in general.
- i have manually tested the upgrade and confirmed it eliminates the issue for affected content
- i have verified that our existing tests would be able to pick up the issue in https://github.com/4teamwork/opengever.core/blob/396fb99a13cb241a650c21dcf163cb0026da7505/opengever/dossier/tests/test_dossier_template.py#L434-L461 ⚠️ this change is not in this PR as we will fix the affected data instead

**Know problems:**
- when creating content via `plone.restapi` the description is still incorrectly set to a bytestring: https://4teamwork.atlassian.net/browse/GEVER-918
- `IDossierTemplate` seems to inherit unused fields from `IDossier`, namely `touched` and `responsible`, this may become an issue when we implement dossier templates in the new ui: https://4teamwork.atlassian.net/browse/GEVER-917

_The PR contains very granular commits, where i change tests so that the issue surfaces, then attempt to fix a single issue with a single commit. I can squash, if desired._

Jira: https://4teamwork.atlassian.net/browse/GEVER-895

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -8,6 +8,7 @@ Changelog
 - Introduce POST @accept-remote-task endpoint for dossiers. [lgraf]
 - Introduce POST @remote-workflow endpoint. [lgraf]
 - Role Assignment Reports: Ensure stable sort order for report items. [lgraf] 
+- Fix dossier template description, ensure unicode. [deiferni]
 - Add policy template for teamraum policies. [njohner]
 - Fix filtering with exclusion filters if the field has a mapping. [tinagerber]
 - Make the portal_url configurable through the portal_registry. [elioschmutz]
--- a/docs/schema-dumps/ftw.mail.mail.schema.json
+++ b/docs/schema-dumps/ftw.mail.mail.schema.json
@@ -57,8 +57,7 @@
             "type": "string",
             "title": "Bearbeitungsinformation",
             "description": "Datum Gesuch, Gesuchsteller, Datum Entscheid, Verweis auf GEVER-Gesuchdossier",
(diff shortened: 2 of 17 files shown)


lukasgraf:
LGTM 👍 🎉

PR #6602, opened 2020-08-25
Update policy generator

elioschmutz:
This PR updates the policy generator with the newest knowledge after generating a teamraum policy.

- removes unused questions for the teamraum deplyoment
- sets useful default values for some teamraum deployment questions
- do ask for the full base-ldap name instead only the last part of it
- adds a useful help message for the base-ldap name
- add the admin-unit id to the client-policy in the buildout.cfg to properly load the client policy
- adds an additional questions for the `ogds_name`
- adds an additional questions for the `bumblebee_app_id`
- adds an additional questions for the `bumblebee_secret`
- adds an additional questions for the `workspace_secret`

Jira: https://4teamwork.atlassian.net/browse/GEVER-675

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/policytemplates/policy_template/opengever.+package.name+/opengever/+package.name+/+adminunit.id+/profiles/ldap/ldap_plugin.xml.bob
+++ b/opengever/policytemplates/policy_template/opengever.+package.name+/opengever/+package.name+/+adminunit.id+/profiles/ldap/ldap_plugin.xml.bob
@@ -25,7 +25,7 @@
                 <item value="cn"/>
             </property>
             <property id="users_base" type="str">
-                <item value="ou=Users,ou={{{deployment.ldap_ou}}},ou=OneGovGEVER,dc=4teamwork,dc=ch"/>
+                <item value="ou=Users,ou={{{deployment.ldap_ou}}}"/>
             </property>
             <property id="users_scope" type="int">
                 <item value="2"/>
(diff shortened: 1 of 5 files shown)


opengever/policytemplates/policy_template/opengever.+package.name+/opengever/+package.name+/+adminunit.id+/profiles/ldap/ldap_plugin.xml.bob
```diff
@@ -25,7 +25,7 @@
                 <item value="cn"/>
             </property>
             <property id="users_base" type="str">
-                <item value="ou=Users,ou={{{deployment.ldap_ou}}},ou=OneGovGEVER,dc=4teamwork,dc=ch"/>
+                <item value="ou=Users,ou={{{deployment.ldap_ou}}}"/>
```
deiferni, line 28: so this will now expect the full OU path? i'm not sure if we want to go that way, the convenience so far was that the template would know the path 🤔. Could you provide a description for the question that provides a longer example of what is expected, so we can ask using the question mark `?` input.

deiferni (approved):
LGTM 👍 . One comment.

Also a bit more context in the PR description would be nice, specifically addressing the motivation of this PR (i figured it out from the commit messages).

elioschmutz:
@deiferni I update the PR description.

To the `ldap_ou`. I thought that it's easier to understand and is more flexible if we can just write the full base-ldap ou instead of just the last part.

The full OU is visible i.e. in the ldaptain:

Bildschirmfoto 2020-08-27 um 12 26 49

We could also just add a better help text for the previous template... as you want.

deiferni (approved):
I'm fine with the flexibility when there is a good description 😉 , it just occurred to me that we might have different OUs for the teamraum deployments. We can always revert later if we figure out it is not needed.

PR #6603, opened 2020-08-27
Use correct client-policy in production template.

deiferni:
This is a follow-up for https://github.com/4teamwork/opengever.core/pull/6582. The `client-policy` used in the production `buildout.cfg` was not correct yet and is fixed with this PR to also include the additional nesting level.

✅   &nbsp;I have manually tested that the new policy is now correct.

Jira: https://4teamwork.atlassian.net/browse/GEVER-751

## Checklist (Must have)

- [ ] ~~Changelog entry~~


elioschmutz:
@deiferni There is already an open PR for this: https://github.com/4teamwork/opengever.core/pull/6602

deiferni:
🙈

PR #6604, opened 2020-08-27
Add @transfer-dossier Endpoint

elioschmutz:
This PR implements the `@transfer-dossier` endpoint.

This endpoint provides the functionality to transfer the responsible of a user to another user. With the option: `include_subdossiers`, all subdossiers of the specified user will also be transfered.

Jira: https://4teamwork.atlassian.net/browse/PHX-14

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/docs/public/dev-manual/api/dossiers.rst
+++ b/docs/public/dev-manual/api/dossiers.rst
@@ -1,0 +1,65 @@
+.. _dossiers:
+
+Dossiers
+========
+
+Dossiers können über die REST-API gem. Kapitel :ref:`operations` bedient werden. Zusätzlich stehen folgende Funktionen für Dossiers zur Verfügung.
+
+
+Dossier übertragen
+------------------
+
+Der Dossier-Verantwortliche kann über den Endpoint ``@transfer-dossier`` an eine neue Person übertragen werden. Dabei wird überprüft, ob ``old_userid`` der aktuelle Dossier-Verantwortliche ist. Ist dies der Fall, wird der Benutzer mit der User-ID ``new_userid`` als neuer Verantwortlicher gesetzt.
+Benachrichtigungen, die normalerweise bei einer Änderung ausgelöst werden, werden unterdrückt. Dieser Endpoint wird mit einer Berechtigung geschützt: ``opengever.api.TransferAssignment``
+Die Berechtigung ist standardmässig den Rollen `Administrator` und `Manager` zugewiesen.
+
+Standardmässig werden auch Subossiers, welche explizit auf den früheren Benutzer gesetzt waren, auf den neuen Verantwortlichen übertragen.
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+      POST /dossier-1/@transfer-dossier HTTP/1.1
+      Accept: application/json
+      Content-Type: application/json
+
+      {
+        "old_userid": "john.doe",
+        "new_userid": "robert.ziegler"
+      }
+
+
+**Beispiel-Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 204 No content
+
+Übertragen von Subdossiers unterbinden
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Über die Option ``recursive`` kann gesteuert werden, ob Subdossiers auch übertragen werden sollen, oder nicht.
+
+Standardmässig ist diese Option aktiviert.
+
+
+**Beispiel-Request**:
+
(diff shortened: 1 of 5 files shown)


opengever/api/tests/test_transfer_dossier.py
```diff
@@ -93,3 +94,97 @@ def test_transfer_dossier_with_identical_userids_raises_bad_request(self, browse
...
+                     headers=self.api_headers, data=json.dumps(
+                         {"old_userid": self.dossier_responsible.getId(),
+                          "new_userid": self.meeting_user.getId(),
+                          "include_subdossiers": False}))
+
+        self.assertEqual(self.meeting_user.getId(), IDossier(self.dossier).responsible)
+        self.assertEqual(self.dossier_responsible.getId(), IDossier(self.subdossier).responsible)
+
+    @browsing
+    def test_transfer_dossier_is_only_alloed_for_not_resolved_dossiers(self, browser):
```
lukasgraf, line 159: `alloed` -> `allowed`

opengever/api/tests/test_transfer_dossier.py
```diff
@@ -93,3 +94,97 @@ def test_transfer_dossier_with_identical_userids_raises_bad_request(self, browse
...
+        browser.open(self.dossier.absolute_url() + '/@transfer-dossier', method='POST',
+                     headers=self.api_headers, data=json.dumps(
+                         {"old_userid": self.dossier_responsible.getId(),
+                          "new_userid": self.meeting_user.getId()}))
+
+        self.assertEqual(self.meeting_user.getId(), IDossier(self.dossier).responsible)
+        self.assertEqual(self.meeting_user.getId(), IDossier(self.subdossier).responsible)
+
+    @browsing
+    def test_transfer_dossier_changes_responsible_only_if_its_set_to_the_current_user(self, browser):
```
lukasgraf, line 128: `current_user` -> `old_user`

opengever/api/transfer.py
```diff
@@ -95,14 +98,39 @@ def reply(self):
...
+        self.transfer_dossier(self.context, old_userid, new_userid, include_subdossiers)
+
         self.request.response.setStatus(204)
         return super(TransferDossierPost, self).reply()
+
+    def transfer_dossier(self, dossier, old_userid, new_userid, include_subdossiers=True):
+        if include_subdossiers:
+            catalog = api.portal.get_tool('portal_catalog')
+            dossiers_to_transfer = [brain.getObject() for brain in catalog(
+                path="{}".format('/'.join(self.context.getPhysicalPath())),
```
lukasgraf, line 126: The `"{}".format(...)` is unnecessary

opengever/api/transfer.py
```diff
@@ -95,14 +98,39 @@ def reply(self):
...
+                portal_type=self.context.portal_type,
```
lukasgraf, line 127: I would change this to querying `object_provides` for `IDossierMarker` - we *know* we want to find all dossiers, so it's unnecessary to get the portal type dynamically from the context. Also `IDossierMarker` would work for cases where dossiers of different types are nested (special dossiers in ZG).

opengever/api/transfer.py
```diff
@@ -95,14 +98,39 @@ def reply(self):
...
+                responsible=old_userid,
+                review_state=DOSSIER_STATES_OPEN)]
+        else:
+            dossiers_to_transfer = [self.context]
+
+        for dossier in dossiers_to_transfer:
+            IDossier(dossier).responsible = new_userid
+            dossier.reindexObject(idxs=['responsible'])
+            notify(ObjectModifiedEvent(dossier))
```
lukasgraf, line 136: Could you add a comment here describing why `ObjectModifiedEvent` needs to be fired? I assume it has to do with journalization, but it's not quite clear from the code.

lukasgraf (changes requested):
Some minor details, otherwise LGTM 👍

docs/public/dev-manual/api/dossiers.rst
```diff
@@ -0,0 +1,65 @@
+.. _dossiers:
+
+Dossiers
+========
+
+Dossier können über die REST-API gem. Kapitel Kapitel :ref:`operations` bedient werden. Zusätzlich stehen folgende Funktionen für Dossiers zur Verfügung.
```
lukasgraf, line 6: Dossier -> Dossiers, Kapitel Kapitel -> Kapitel

PR #6605, opened 2020-08-27
Provide create_forwarding action in API for documents in inboxes.

deiferni:
With this PR we provide an action `create_forwarding` for documents in inboxes. This helps the UI to create forwardings not only from listing but from the overview of a single document.

- The action will only appear in inboxes as only they manage the `opengever.inbox: Add Forwarding` permission.
- The action is only exposed via rest-API but not in the classic ui as that would require a custom JS-Integration and is out of scope for this issue.

## Checklist (Must have)

- [x] Changelog entry


lukasgraf (approved):
LGTM 👍

PR #6606, opened 2020-08-28
Pass query string to location url on maindossier and parentdossier redirect

bierik:
This PR preserves the query string for the `redirect_to_parent_dossier` and `redirect_to_main_dossier` view.

The GEVER-UI needs to redirect to parent contents. The business-logic should come from the backend. Because we need this paths in listings, we would have to index all the possible paths in the solr catalog and return it when needed. We can't do/don't want to do this.

Instead, we can just use the redirect views on a context to get the new parent context. Because the GEVER-UI adds a lot of query parameters to get all required information form the backend, we have to preserve the query string on redirection.

Bildschirmfoto 2020-08-31 um 10 37 14

Jira: https://4teamwork.atlassian.net/browse/PHX-1

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/document/browser/redirect_to_parent_dossier.py
+++ b/opengever/document/browser/redirect_to_parent_dossier.py
@@ -3,10 +3,18 @@ from Products.Five import BrowserView
 
 class RedirectToParentDossier(BrowserView):
     """Redirects to the containing subdossier of a document.
+
+    Caution: this view will be used by GEVER-UI/RIS/VM
     """
     def __call__(self):
         dossier = self.context.get_parent_dossier()
         assert dossier, ('the redirect view should only be called when a '
                          'parent dossier is available')
 
-        return self.request.RESPONSE.redirect(dossier.absolute_url())
+        url_parts = [dossier.absolute_url()]
+        qs = self.request['QUERY_STRING']
+        if qs:
+            url_parts.append(qs)
+        redirect_url = '?'.join(url_parts)
+
+        return self.request.RESPONSE.redirect(redirect_url)
--- a/opengever/dossier/browser/redirect_to_maindossier.py
+++ b/opengever/dossier/browser/redirect_to_maindossier.py
@@ -5,7 +5,10 @@ from Products.Five.browser import BrowserView
 
 
 class RedirectToMainDossier(BrowserView):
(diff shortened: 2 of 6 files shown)


opengever/dossier/browser/redirect_to_maindossier.py
```diff
@@ -15,4 +18,9 @@ def __call__(self):
             api.portal.show_message(msg, request=self.request, type='error')
             return self.request.RESPONSE.redirect(self.context.absolute_url())
 
-        return self.request.RESPONSE.redirect(main_dossier.absolute_url())
+        return self.request.RESPONSE.redirect(
+            "?".join([
+                main_dossier.absolute_url(),
+                self.request["QUERY_STRING"],
```
lukasgraf, line 24: This will also append an unnecessary `?` at the end of the URL if `self.request["QUERY_STRING"]` is the empty string. Could you please change this so it only includes `?` in the URL if there's an actual query string?

E.g. like this:
```python
url_parts = [main_dossier.absolute_url()]
qs = self.request["QUERY_STRING"]
if qs:
    url_parts.append(qs)
redirect_url = '?'.join(url_parts)
```
_(untested)_

opengever/document/browser/redirect_to_parent_dossier.py
```diff
@@ -3,10 +3,14 @@
...
+    Caution: this view will be used by GEVER-UI/RIS/VM
     """
     def __call__(self):
         dossier = self.context.get_parent_dossier()
         assert dossier, ('the redirect view should only be called when a '
                          'parent dossier is available')
 
-        return self.request.RESPONSE.redirect(dossier.absolute_url())
+        self.request.RESPONSE.redirect(
+            "?".join([dossier.absolute_url(), self.request["QUERY_STRING"]])
```
lukasgraf, line 15: Same here, this will lead to an unnecessary `?` at the end of the URL if `self.request["QUERY_STRING"]` is the empty string.

lukasgraf (approved):
Just a minor point, otherwise LGTM 👍

PR #6607, opened 2020-08-31
Add missing translations for dossier export.

2e12:
A French translation isn't available when exporting the dossier cover, therefore it is added.

Example files:
[dossiercover.pdf](https://github.com/4teamwork/opengever.core/files/5274165/dossiercover.pdf)
[dossierdetails.pdf](https://github.com/4teamwork/opengever.core/files/5274166/dossierdetails.pdf)

Jira: https://4teamwork.atlassian.net/browse/GEVER-832

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/latex/dossiercover.py
+++ b/opengever/latex/dossiercover.py
@@ -6,10 +6,12 @@ from ftw.pdfgenerator.utils import provide_request_layer
 from ftw.pdfgenerator.view import MakoLaTeXView
 from opengever.base.interfaces import IReferenceNumber
 from opengever.dossier.behaviors.dossier import IDossier
+from opengever.latex import _
 from opengever.ogds.base.utils import get_current_admin_unit
 from opengever.repository.repositoryroot import IRepositoryRoot
 from zope.component import adapter
 from zope.component import getAdapter
+from zope.i18n import translate
 from zope.interface import Interface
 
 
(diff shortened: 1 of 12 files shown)


opengever/latex/dossiercover.py
```diff
@@ -35,21 +37,55 @@ class DossierCoverLaTeXView(MakoLaTeXView):
...
-            'referencenr': self.convert_plain(self.get_referencenumber()),
-            'title': self.convert_plain(self.context.Title()),
-            'description': self.convert(self.get_description()),
-            'responsible': self.convert_plain(self.get_responsible()),
-
-            'start': self.convert_plain(self.context.toLocalizedTime(
-                    str(IDossier(self.context).start)) or '-'),
-            'end': self.convert_plain(self.context.toLocalizedTime(
-                    str(IDossier(self.context).end)) or '-'),
+        return self.get_dossier_metadata()
```
njohner, line 40: I'm not a fan of this method that just proxies to another method. What's the point?

opengever/latex/dossiercover.py
```diff
@@ -35,21 +37,55 @@ class DossierCoverLaTeXView(MakoLaTeXView):
...
+
+    def get_metadata_config(self):
```
njohner, line 42: The naming is a bit weird to me. How is that a config?

opengever/latex/dossiercover.py
```diff
@@ -35,21 +37,55 @@ class DossierCoverLaTeXView(MakoLaTeXView):
...
+
+    def get_dossier_metadata(self):
+        args = {}
+        metadata = self.get_metadata_config()
+        for key, item in metadata.items():
+            item['label'] = translate(item['label'], context=self.request)
+            if item['label'] == '':
+                args[key] = item['value']
+            else:
+                args[key] = u'{}: {}'.format(item['label'], item['value'])
```
njohner, line 88: I find that whole thing pretty convoluted. First you set labels for all the metadata, then when it's an empty string you get rid of it.
Moreover I feel like you are doing the template's job here. Why not simply add the few missing labels to the original missing `args` dictionary you had? You did not make the code less complex and you did not make the code shorter, so what did you gain?
I'd either simply add 'responsible_label', `start_label` and so on to `args` or I'd go maybe for something like:
```
    translatable_labels = {
        'responsible_label': _('label_responsible', default='Responsible')
    }
    def get_render_arguments(self):
        args = self.get_dossier_metadata()
        args.update(self.get_metadata_labels())
  
    def get_translated_labels(self):
        args = {}
        for key, label in self.translatable_labels.items():
              args[key] = translate(label, context=self.request)
        return args
```

But again the simplest solution of adding the needed labels in `args` might even be easier to read

opengever/latex/templates/dossiercover.tex
```diff
@@ -15,16 +15,16 @@
 %% \end{tabularx}
 \vspace{0.5cm}
 \rule{\linewidth}{.11em}
-Version Ordnungssystem: ${repositoryversion}\newline
-\textbf{Aktenzeichen: ${referencenr}}\newline
+${repositoryversion}\newline
+\textbf{${referencenr}}\newline
```
njohner, line 19: and here you would have:
`\textbf{${referencenr_label}: ${referencenr}}\newline`

opengever/latex/tests/test_default_layout.py
```diff
@@ -83,94 +43,69 @@ def test_before_render_hook(self):
...
+            'page_label': 'Page',
+            'client_title': 'Hauptmandant',
+            'location': 'Bern',
             'show_organisation': False,
-            'location': FAKE_LOCATION,
+            'member_phone': '012 345 6789',
+            'show_logo': False
             }
 
         self.assertEqual(layout.get_render_arguments(), args)
```
njohner, line 67: `assertDictEqual` would be better

njohner (changes requested):
Over LGTM, but I think you took a bit of a convoluted approach, see my comment. Also, try to make your commit message titles a bit more explicit. You also have some typos in the commit messages, notably `is'nt` should be `isn't`

2e12:
@njohner Thank you for your feedback.I tried to make it as similar as possible to [dossierdetails.py,](https://github.com/4teamwork/opengever.core/blob/master/opengever/latex/dossierdetails.py) which is why the code is the way it is. But I can understand your feedback and I will adapt my changes accordingly.

njohner (approved):
LGTM

PR #6608, opened 2020-08-31
Allow to query by token in @querysources API endpoint.

deiferni:
With this PR we allow the `@querysources` API endpoint to be queried by a `token` parameter in addition to the existing `query` parameter. The token can be used to retrieve the representation of one already known token. It will return an empty list if no such token exists.

- I have added some very basic docs for our customizations to vocabularies and sources as specified in https://plonerestapi.readthedocs.io/en/latest/vocabularies.html
- I mostly used the description i found in https://github.com/4teamwork/opengever.core/issues/5973 and https://github.com/4teamwork/opengever.core/issues/5968 for this
- I have amended tests so they also cover the possible error-states (no or too many query string parameters provided)
- I have also update the schema docs. this is a leftover of https://github.com/4teamwork/opengever.core/pull/6601 and a should have done it there 😓 

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 
 - Adjust the policy generator for easier policy generation. [elioschmutz]
 - Provide create_forwarding action in API for documents in inboxes. [deiferni]
+- Allow to query by token in @querysources API endpoint. [deiferni]
 
 
 2020.8.0 (2020-08-26)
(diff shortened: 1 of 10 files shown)


lukasgraf (approved):
LGTM 👍 
Thanks for the docs regarding the schema intent! 🙈  🎉

PR #6609, opened 2020-08-31
Escape solr queries when building literal queries.

deiferni:
With this PR we fix escaping special characters when building solr literal queries. This lead to query strings like `Eingangskorb:Blabla` failing and thus the whole request not returning any e.g. issuers for a task in autocomplete.

## Checklist (Must have)

- [x] Changelog entry


lukasgraf (approved):
LGTM 👍

PR #6610, opened 2020-08-31
Prevent deadlock when reassigning inter-admin-unit tasks

lukasgraf:
This solves an edge case which can lead to a deadlock when reassigning inter-admin-unit tasks. If the new responsible doesn't have an entry in the `watchers` SQL table yet, both the predecessor and successor sides would try to add it for the first time, and therefore **deadlock on the to-be-inserted row for the new watcher**.

This is prevented by creating the new watcher (if required) at the very beginning of this process, immediately committing the transaction, and starting a new one.

For a more detailed description of the problem, please see the Jira issue linked below.

Testing the effect of this change unfortunately is near impossible - it's not enough to fake a sort-of-inter-admin-unit scenario, because actual HTTP requests, across multiple real Zope instances were responsible for creating the deadlock in the first place.

Jira: https://4teamwork.atlassian.net/browse/GEVER-946

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/base/txnutils.py
+++ b/opengever/base/txnutils.py
@@ -1,0 +1,23 @@
+from zope.globalrequest import getRequest
+import itertools
+
+
+def txn_is_dirty():
+    """Determines whether the current transaction has changes or not.
+    """
+    return bool(registered_objects())
+
+
+def registered_objects():
+    """Returns a list of objects changed in this transaction.
+
+    Lifted from plone.protect.auto.
+    """
+    app = getRequest().PARENTS[-1]
+    return list(itertools.chain.from_iterable([
+        conn._registered_objects
+        # skip the 'temporary' connection since it stores session objects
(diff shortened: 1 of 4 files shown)


opengever/base/txnutils.py
```diff
@@ -0,0 +1,23 @@
...
+import itertools
+
+
+def txn_is_dirty():
+    """Determines whether the current transaction has changes or not.
+    """
+    return bool(_registered_objects())
+
+
+def _registered_objects():
```
deiferni, line 11: I would not prefix this with the `_` "protected" underscore. For me this indicates it is not part of the "public" api of the `txnutils` module but only a helper function. But it is and it should/can be used stand-alone. I'd remove the prefix.

deiferni (approved):
LGTM 👍 

just one "public" interface naming suggestion.

PR #6611, opened 2020-08-31
Consider cookie when figuring out current orgunit in AllUsersInboxesAndTeamsSource.

deiferni:
This PR adds an additional fallback for the `AllUsersInboxesAndTeamsSource` when attempting to figure out the currently selected orgunit. For requests from the UI via API we don't have the form prefilling that is used to get the current orgunit for the forwarding add-form. Instead we must consider the cookie value for the orgunit currently selected.

The current mechanism only considers request form parameters https://github.com/4teamwork/opengever.core/blob/2f2f92f4dc4576ffe254195a79e94d59c440dd82/opengever/ogds/base/sources.py#L301-L318, but the cookie, which is always provided by the UI, is not considered. I have added the fallback last to avoid it from shadowing the edit form, when the value must be retrieved from `context`.

I have also fixed a related issue that popped up in some functional tests by using a better functional fixture (separate commits for clarity).

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -8,6 +8,7 @@ Changelog
 - Provide create_forwarding action in API for documents in inboxes. [deiferni]
 - Allow to query by token in @querysources API endpoint. [deiferni]
 - Fix escaping solr literal queries. [deiferni]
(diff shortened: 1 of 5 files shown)


lukasgraf (approved):
LGTM 👍

PR #6612, opened 2020-09-01
Fix forwarding requiring task_type in API, fix forwarding task_type translations

deiferni:
With this PR we fix multiple related issues with `opengever.inbox.forwarding` when creating and reading via API:

- forwarding no longer requires a `task_type` to be provided, and will no longer ignore the `task_type` provided via `POST`
- as `task_type` is now a proper field for forwardings we get rid of the property, we also provide a migration to set the `task_type` on existing content to ensure all objects are in the same state
- forwarding `task_type` representation in GET is fixed to be consistent with GET of a task, the task_type is now an object, also providing the translation
- forwarding `task_type` in `@globalindex` is now also translated

I have also amended missing test for forwarding defaults (where possible) and have amended default tests for creating a task via API.

I have not unified the various distributed task_type helpers 😠 .

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [x] api-change label added
- New functionality:
- New translations

### DIFF
--- a/opengever/inbox/forwarding.py
+++ b/opengever/inbox/forwarding.py
@@ -4,6 +4,7 @@ from datetime import datetime
 from ftw.keywordwidget.widget import KeywordFieldWidget
 from opengever.base.role_assignments import RoleAssignmentManager
 from opengever.inbox import _
+from opengever.inbox import FORWARDING_TASK_TYPE_ID
 from opengever.ogds.base.sources import AllUsersInboxesAndTeamsSourceBinder
 from opengever.ogds.base.utils import get_current_org_unit
 from opengever.ogds.base.utils import get_ou_selector
@@ -21,6 +22,8 @@ from z3c.form.interfaces import HIDDEN_MODE
 from zope import schema
 from zope.i18n import translate
 from zope.interface import implements
+
+
 
 
 class IForwarding(ITask):
@@ -79,7 +82,7 @@ class Forwarding(Task):
         """Provide a marker string, which will be translated
         in the tabbedview helper method.
         """
-        return 'forwarding_task_type'
+        return FORWARDING_TASK_TYPE_ID
 
     def set_static_task_type(self, value):
         """Marker set function"""
@@ -89,7 +92,7 @@ class Forwarding(Task):
     task_type = property(get_static_task_type, set_static_task_type)
 
     def get_task_type_label(self, language=None):
-        label = _('forwarding_task_type', default=u'Forwarding')
+        label = _(FORWARDING_TASK_TYPE_ID, default=u'Forwarding')
         if language:
             return translate(
                 label,
--- a/opengever/inbox/forwarding.py
+++ b/opengever/inbox/forwarding.py
@@ -22,8 +22,21 @@ from z3c.form.interfaces import HIDDEN_MODE
 from zope import schema
 from zope.i18n import translate
 from zope.interface import implements
+from zope.interface import provider
+from zope.schema.interfaces import IContextSourceBinder
+from zope.schema.vocabulary import SimpleTerm
+from zope.schema.vocabulary import SimpleVocabulary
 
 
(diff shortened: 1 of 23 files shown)


opengever/inbox/forwarding.py
```diff
@@ -58,6 +66,15 @@ class IForwarding(ITask):
         required=False,
     )
 
+    # task type only allows one value, make it not required with a default
+    task_type = schema.Choice(
+        title=task_mf(u'label_task_type', default=u'Task Type'),
+        required=False,
+        default=FORWARDING_TASK_TYPE_ID,
+        missing_value=FORWARDING_TASK_TYPE_ID,
```
lukasgraf, line 74: This is incorrect IMHO. The missing value is intended to be used for the case where the value is actually _absent_. It should always be falsy. Omitting the `missing_value` kwarg here (and letting the `zope.schema.Choice` standard value, `None`, play out), would be the right thing to to here I'd say.

opengever/inbox/forwarding.py
```diff
@@ -92,7 +109,7 @@ def set_static_task_type(self, value):
     task_type = property(get_static_task_type, set_static_task_type)
```
lukasgraf, line 109: Ugh. Wouldn't it make more sense to make this a proper [readonly](https://github.com/4teamwork/opengever.core/blob/master/opengever/dossier/behaviors/dossier.py#L90) field on the forwarding type instead?

opengever/inbox/locales/de/LC_MESSAGES/opengever.inbox.po
```diff
@@ -18,6 +18,10 @@ msgstr ""
 msgid "Documents"
 msgstr "Dokumente"
 
+#: ./opengever/inbox/forwarding.py
+msgid "Forwarding"
```
lukasgraf, line 22: I'm honestly not quite sure what the motivation was to remove the indirection via the `forwarding_task_type` message ID for translations, and instead directly rely on a direct translation of the term `Forwarding`.

Maybe I'm missing something, but it looks to me like it makes it harder to preserve the context in which a translation is used (and possibly change it separately in the future, independent of other places where the term `Forwarding` might be used in contexts other than the "task" type).

lukasgraf (changes requested):

opengever/inbox/forwarding.py
deiferni: will amend 👍

opengever/inbox/forwarding.py
deiferni: yeah, i know. i had tried the `readonly` initially but it "did not work"™️. but i can give it another shot with all the changes implemented now.

opengever/inbox/locales/de/LC_MESSAGES/opengever.inbox.po
deiferni: you are correct, the forwarding content type and the task type are two different things and it makes more sense to keep them separate even though the translation currently is the same. will amend 👍.

deiferni:
@lukasgraf amended with the changes you suggested. i could get rid of the property, too. i have written a migration to make sure the `task_type` is set on all existing forwardings. while this would not be strictly necessary it ensures all
objects are in the same state and all have the correct attribute in their `__dict__` and don't acquire it via field default. that's why we have the default value patches in the first place 😉. 'd love to hear your thoughts about the approach though.

lukasgraf (approved):
LGTM 👍

deiferni:
For the record, we've decided against `readonly=True` as this would, for them moment, lead to the default value not being set as per https://github.com/4teamwork/opengever.core/blob/40fdc9b7c76bbfd3e1ab6bae5cf72289646b94bc/opengever/base/default_values.py#L308-L309.

PR #6614, opened 2020-09-02
Add @type to @globalindex items.

deiferni:
With this PR we add the `@type` attribute to items we return. We use the existing `task_type` column to figure out the portal type ss we don't store an e.g. `portal_type` in the tasks table. The `task_type` is distinct and we can use it to figure out the portal type. This helps the UI as it might display forwardings and tasks differently when retrieved from the `@globalindex` API service endpoint.

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
- New functionality:


lukasgraf (approved):
LGTM 👍

PR #6615, opened 2020-09-02
Fix resolving favorites that don't exist.

tinagerber:
If an attempt is made to resolve a favorite that does not exist, an exception is raised. This exception has not been handled so far. Now the favorite is returned unresolved.

Jira: https://4teamwork.atlassian.net/browse/GEVER-868

## Checklist (Must have)

- [x] Changelog entry


lukasgraf (changes requested):
Could you add a regression test for this please? 😉

tinagerber:
@lukasgraf I have added a test, not sure if it is what you wanted 🤷‍♀️

lukasgraf (approved):

PR #6616, opened 2020-09-02
Add support for transferring inter-admin-unit tasks.

lukasgraf:
This adds support for also transferring inter-admin-unit tasks using the `@transfer-tasks` endpoint.

For changing the `responsible`, the existing "reassign task" mechanism, which involves an actual workflow transition, is reused. For changing the `issuer`, that mechanism is emulated, and supported by a new `IssuerChangeResponseSyncerSender` that syncs the change to the remote side (without triggering an actual WF transition though, because there is none).

The situation for forwardings doesn't really change with the support for inter-admin-unit tasks, because forwardings can never be in a state where the predecessor forwarding is still open - therefore, no syncing of changes is required. 

## Checklist (Must have)

- [x] Changelog entry

Jira: https://4teamwork.atlassian.net/browse/GEVER-541

## Checklist (if applicable)

- API change:
  _(Not needed, checked)_

- New functionality:

### DIFF
--- a/opengever/api/transfer.py
+++ b/opengever/api/transfer.py
@@ -1,12 +1,15 @@
+from opengever.activity.model.watcher import Watcher
+from opengever.base.model import create_session
+from opengever.base.sentry import log_msg_to_sentry
+from opengever.base.txnutils import registered_objects
+from opengever.base.txnutils import txn_is_dirty
 from opengever.dossier.base import DOSSIER_STATES_OPEN
 from opengever.dossier.behaviors.dossier import IDossier
 from opengever.dossier.behaviors.dossier import IDossierMarker
 from opengever.inbox.forwarding import IForwarding
 from opengever.ogds.models.user import User
 from opengever.task.activities import TaskChangeIssuerActivity
-from opengever.task.activities import TaskReassignActivity
-from opengever.task.interfaces import ISuccessorTaskController
-from opengever.task.localroles import LocalRolesSetter
+from opengever.task.response_syncer import sync_task_response
 from opengever.task.task import ITask
 from opengever.task.util import add_simple_response
 from plone import api
@@ -17,6 +20,11 @@ from zExceptions import BadRequest
 from zope.event import notify
 from zope.interface import alsoProvides
 from zope.lifecycleevent import ObjectModifiedEvent
+import logging
+import transaction
+
+
+logger = logging.getLogger('opengever.api')
 
 
 class ExtractOldNewUserMixin(Service):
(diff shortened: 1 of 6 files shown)


opengever/api/transfer.py
```diff
@@ -62,41 +72,100 @@ def reply(self):
...
+
+        self.context.issuer = new_issuer
+        notify(ObjectModifiedEvent(self.context))
+        TaskChangeIssuerActivity(self.context, self.context.REQUEST, issuer_response).record()
+
+        sync_task_response(
+            self.context, self.request, 'issuer-change',
+            transition, text=None, new_issuer=new_issuer)
+
+    def create_watcher_and_commit_if_needed(self, new_responsible):
```
deiferni, line 135: This is an exact a copy of what we already have in `AssignTaskForm`? If so i would prefer to have this critical path extracted into e.g. a method and reused by both the `AssignTaskForm` and `TransferTaskPost`. Maybe a function in `opengever.activity` would be a good place to put it?

opengever/task/response_syncer/issuer_change.py
```diff
@@ -0,0 +1,79 @@
...
+    def raise_sync_exception(self, task, transition, text, **kwargs):
+        raise ResponseSyncerSenderException(
+            'Could not transfer issuer for task on remote '
+            'admin unit {} ({})'.format(
+                task.admin_unit_id,
+                task.physical_path))
+
+    def _is_synced_transition(self, transition):
+        return transition in [
+            'task-transition-change-issuer',
```
deiferni, line 37: Would it make sense to extract this state into one of the constants in https://github.com/4teamwork/opengever.core/blob/abed7d9520e21478f312ab18079f40cd5d0fbe3c/opengever/task/__init__.py ?

opengever/testing/test_case.py
```diff
@@ -76,6 +78,44 @@ def allowed(obj):
         children['added'] = set(children['after']) - set(children['before'])
         children['removed'] = set(children['before']) - set(children['after'])
 
+    @contextmanager
+    def observe_responses(self, obj):
```
deiferni, line 82: 💛

deiferni (approved):
LGTM 👍 , just two remarks. I also like the new `observe_[thingy]` context managers 🎉 .

opengever/api/transfer.py
lukasgraf: I deliberately did not factor that out, because that would move the code far away from the place where it's being used, and make it much less obvious that something rather unusual is being done in these two places. Having the `transaction.commit(); transaction.begin()` (which should raise a couple flags when spotted) right there in the same file, next to the context where it's being used, trumps DRY / code reuse by far here IMHO.

In other words: I don't trust people enough to responsibly use that function, that's why I don't want to factor it out. It _should_ not be reusable with minimal effort. If we actually do require this pattern in a third place, then we should consider it, but for now it's strictly tied to the "reassign" mechanics (and documented as such).

opengever/task/response_syncer/issuer_change.py
lukasgraf: I don't think so. It's a pseudo-transition, as mentioned in the class below for the receiving side. And it has specifically been invented to facilitate that issuer-change via the `@transfer-task` endpoint. I see zero chance of this being reused anytime soon.

PR #6618, opened 2020-09-03
Don't show workspace actions for non-open dossiers or when the user can only view.

deiferni:
Don't display actions to transfer documents from and to a teamraum when the dossier is not open or when the user does not have permissions to modify the dossier or add content to the dossier.

Jira: https://4teamwork.atlassian.net/browse/GEVER-953

## Checklist (Must have)

- [x] Changelog entry


lukasgraf (approved):
LGTM 👍

PR #6619, opened 2020-09-03
Add task `@predecessor` and  @`successors` expansion.

deiferni:
With this PR we add `@predecessor` and  @`successors` expansion to tasks. This helps the UI to display predecessors/successors for inter-admin-unit tasks.

I've seen that currenlty the `@globalindex` endpoint had implemented its own serialization, bypassing the `ISerializeToJson` and `ISerializeToJsonSummary` adapters introduced by `plone.restapi`. With this PR i have refactored the existing functionality to provide such adapters.

Decisions/thoughts:
- After some back and forth i've decided to put all functionality into `opengever.api.task` but of course that can be changed.
- After some more back and forth i've decided to nest the tasks as `items` and `item` respectively. This could be changed to the explicit `successors` and `predecessors` but i did not like the repetition.
- Task serialization as used by `@globalindex` was not changed, this means that some columns are deliberately not exposed. I'm not sure why that is the case in the first place, but to keep this a pure refactoring I kept the behavior.

Jira: https://4teamwork.atlassian.net/browse/PHX-9

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
- New functionality:

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.8.1 (unreleased)
 ---------------------
 
+- Add @successors and @predecessor expansion for tasks. [deiferni]
 - Add @share-content endpoint to share content in workspace. [tinagerber]
 - Add @actual-workspace-members endpoint. [tinagerber]
 - Add support for transferring inter-admin-unit tasks. [lgraf]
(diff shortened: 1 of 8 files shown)


deiferni:
Notes to the reviewer:
- It may be easier to review the whole diff, not each commit 
- This one is important for @all4manu, you may ☠️  this PR tomorrow if there are changes necessary

lukasgraf (approved):
LGTM 👍

PR #6621, opened 2020-09-07
Revert adding missing value for public_trial_statement.

njohner:
In this PR we revert part of the changes introduced in https://github.com/4teamwork/opengever.core/pull/6601, namely we remove the `missing_value` for the `public_trial_statement` field. This led to issues when submitting proposals, for which the `public_trial_statement` had a value of `None` (on the proposal document), which was now invalid.

For some unknown reason, the `public_trial_statement` default/missing value does not always get persisted on the object anymore, so that we had to manually fix the default values tests. After discussion with @deiferni and @lukasgraf we deemed this as acceptable for now.

For https://4teamwork.atlassian.net/browse/GEVER-959

This should fix the following sentry issues:
- https://sentry.4teamwork.ch/sentry/onegov-gever/issues/69172
- https://sentry.4teamwork.ch/sentry/onegov-gever/issues/69124

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 
 - Add @successors and @predecessor expansion for tasks. [deiferni]
 - Don't show workspace actions for non-open dossiers or when the user can only view. [deiferni]
+- Revert adding missing value for public_trial_statement. [njohner]
 - Add @share-content endpoint to share content in workspace. [tinagerber]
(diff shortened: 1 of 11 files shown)


njohner:
For info, checks were green. I then added the changelog entry...

deiferni (commented):
As discussed 👍 .

deiferni (approved):
_ehem, approved i mean_

PR #6623, opened 2020-09-08
Correctly handle query strings for oguid on remote admin units in ResolveOGUIDView.

njohner:
In the `resolve_oguid` view we constructed invalid urls when trying to resolve an OGUID on a remote client with query string parameters. This typically happened when the portal appends a ticket to the url (see https://sentry.4teamwork.ch/sentry/onegov-gever/issues/13998/), so that the view was called with an URL of the type `.../adminunit/@@resolve_oguid/remote-adminunit:intid?ticket=myticket`. In that case it would redirect to `.../remote-adminunit/@@resolve_oguid?oguid=remote-adminunit:intid?ticket=myticket`. The second query string separator `?` is interpreted as literal question mark and the view tries to resolve the oguid `remote-adminunit:intid?ticket=myticket`.

For https://4teamwork.atlassian.net/browse/GEVER-888

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 ---------------------
 
 - Bump ftw.monitor to get bin/dump-perf-metrics script. [lgraf]
+- Correctly handle query strings for oguid on remote admin units in ResolveOGUIDView. [njohner]
 - Add @successors and @predecessor expansion for tasks. [deiferni]
 - Don't show workspace actions for non-open dossiers or when the user can only view. [deiferni]
 - Add @share-content endpoint to share content in workspace. [tinagerber]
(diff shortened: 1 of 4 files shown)


docs/HISTORY.txt
```diff
@@ -1,7 +1,7 @@
 Changelog
 =========
 
-2020.8.1 (unreleased)
+2020.9.0 (unreleased)
```
deiferni, line 4: best also update the number in `setup.py` then

deiferni (approved):
LGTM 👍

docs/HISTORY.txt
njohner: Right. I was just bothered by having twice the 2020.8.1 in the changelog, once released and once unreleased

deiferni: Yeah, sorry i think i did not update the default suggestion during `fullrelease`, usually we bump to the next MINOR. Thinking of it, the default next-version could probably easily be changed 🤔.

PR #6624, opened 2020-09-08
Fix ogds listing endpoints for undefined sort orders.

njohner:
When sorting according to a non-unique column, response of ogds listings could contain duplicate items and other items could be missing. This is because the sort order of successive queries is not guaranteed to be identical (except when specifically sorting on a key with a unique value for each item).

With this PR we fix this issue by always having a secondary sort by a unique key for the `OGDSUserListingGet`, `TeamListingGet` and in `GlobalIndexGet`. We also refactor the global index endpoint to use the same base class as the other two OGDS listings.

For https://4teamwork.atlassian.net/browse/GEVER-968

## Checklist (Must have)

- [x] Changelog entry

API changes are non breaking, except that the ordering of tasks will now be different when an invalid `sort_on` is requested. Documentation does not need to be updated IMHO

**Problem analysis:**
Iteration over items of an `SQLHypermediaBatch` iterates over its batch (https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/batching.py#L24-L26), which corresponds to an `SQLBatch`. Iterating over the items in a batch will simply use the `_getitem_` method of the sequence (see https://github.com/plone/plone.batching/blob/master/plone/batching/batch.py#L154 as `SQLBatch` inherits from `BaseBatch`). In our cases, that sequence is an ORM query, which makes a query for every item we are getting (https://github.com/sqlalchemy/sqlalchemy/blob/rel_1_1_18/lib/sqlalchemy/orm/query.py#L2528-L2552), i.e. a query for every item in the batch when iterating over the batch, changing the `OFFSET` and with a `LIMIT` of 1 (which are set in the `slice` method https://github.com/sqlalchemy/sqlalchemy/blob/rel_1_1_18/lib/sqlalchemy/orm/query.py#L2554-L2594). The queries for example for users in the admin panel sorted by `department` made to the SQL in GEVER look like:
```
2020-09-11 16:48:00,616 INFO sqlalchemy.engine.base.Engine SELECT users.userid AS users_userid, users.active AS users_active, users.firstname AS users_firstname, users.lastname AS users_lastname, users.directorate AS users_directorate, users.directorate_abbr AS users_directorate_abbr, users.department AS users_department, users.department_abbr AS users_department_abbr, users.email AS users_email, users.email2 AS users_email2, users.url AS users_url, users.phone_office AS users_phone_office, users.phone_fax AS users_phone_fax, users.phone_mobile AS users_phone_mobile, users.salutation AS users_salutation, users.description AS users_description, users.address1 AS users_address1, users.address2 AS users_address2, users.zip_code AS users_zip_code, users.city AS users_city, users.country AS users_country, users.last_login AS users_last_login 
FROM users ORDER BY department ASC 
 LIMIT %(param_1)s OFFSET %(param_2)s
2020-09-11 16:48:00,617 INFO sqlalchemy.engine.base.Engine {'param_1': 1, 'param_2': 19}
```

Looking at what such queries to SQL actually give back:
```
opengever=# SELECT users.userid FROM users ORDER BY department ASC LIMIT 1;
    userid     
---------------
 sabine.ammann
(1 row)

opengever=# SELECT users.userid FROM users ORDER BY department ASC LIMIT 1 OFFSET 1;
    userid     
---------------
 sabine.ammann
(1 row)

opengever=# SELECT users.userid FROM users ORDER BY department ASC LIMIT 1 OFFSET 2;
    userid     
---------------
 sabine.ammann
(1 row)
```

That behaviour is rather strange but we also see it clearly when we simply query with a limit:
```
opengever=# SELECT users.userid FROM users ORDER BY department ASC LIMIT 1;
    userid     
---------------
 sabine.ammann
(1 row)

opengever=# SELECT users.userid FROM users ORDER BY department ASC LIMIT 2;
     userid      
-----------------
 martin.saalbach
 sabine.ammann
(2 rows)

opengever=# SELECT users.userid FROM users ORDER BY department ASC LIMIT 3;
     userid      
-----------------
 martin.saalbach
 carmen.albrecht
 sabine.ammann
(3 rows)
```
`sabine.ammann` always comes last, for any `LIMIT` used. This comes from sorting according to a column that is mainly `null` for many items. Probably sorting stops as soon as `LIMIT` items with `null` have been found and somehow (probably because of the sorting algorithm), the first item (here `sabine.ammann`) always ends up at the end of that query's results. So when retrieving with an offset to take the last of the results we always get `sabine.ammann`.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,7 +4,8 @@ Changelog
 2020.10.0 (unreleased)
 ----------------------
 
-- Nothing changed yet.
+- Fix globalindex endpoint for undefined sort orders. [njohner]
+- Fix ogds listing endpoints for undefined sort orders. [njohner]
 
 
 2020.9.0 (2020-09-10)
@@ -13,8 +14,6 @@ Changelog
 - Order groups and teams in User serializer by title. [elioschmutz]
 - Bump ftw.monitor to get bin/dump-perf-metrics script. [lgraf]
 - Correctly handle query strings for oguid on remote admin units in ResolveOGUIDView. [njohner]
-- Fix globalindex endpoint for undefined sort orders. [njohner]
-- Fix ogds listing endpoints for undefined sort orders. [njohner]
 - Add @successors and @predecessor expansion for tasks. [deiferni]
 - Don't show workspace actions for non-open dossiers or when the user can only view. [deiferni]
 - Add @share-content endpoint to share content in workspace. [tinagerber]
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -13,6 +13,7 @@ Changelog
 - Order groups and teams in User serializer by title. [elioschmutz]
 - Bump ftw.monitor to get bin/dump-perf-metrics script. [lgraf]
 - Correctly handle query strings for oguid on remote admin units in ResolveOGUIDView. [njohner]
+- Fix globalindex endpoint for undefined sort orders. [njohner]
 - Fix ogds listing endpoints for undefined sort orders. [njohner]
 - Add @successors and @predecessor expansion for tasks. [deiferni]
 - Don't show workspace actions for non-open dossiers or when the user can only view. [deiferni]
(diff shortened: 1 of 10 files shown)


deiferni (commented):
Please talk to @njohner you seem to touch the same thing https://github.com/4teamwork/opengever.core/pull/6624

njohner:
> Please talk to @njohner you seem to touch the same thing #6624

Probably not here that you wanted to post that comment ;-)

lukasgraf (approved):
LGTM! 👍

PR #6626, opened 2020-09-09
Order groups and teams by title

elioschmutz:
This PR orders the groups and the teams coming from `assigned_groups` and `assigned_teams` by title.

This ensures a user friendly default sort order for teams and groups.

Jira: https://4teamwork.atlassian.net/browse/PHX-40

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations


njohner (approved):
LGTM

PR #6627, opened 2020-09-09
Fix id of workflow policy for workspaces

buchi:
I've previously renamed the id to be consistent with other policies but missed some occurrences.
This has broken the upgrade step.

JIRA: https://4teamwork.atlassian.net/browse/GEVER-82

## Checklist (Must have)

- [x] Changelog entry


deiferni (approved):
LGTM 👍

PR #6628, opened 2020-09-09
Fix document workflow definition in upgrade step

buchi:
The upgrade step contained the wrong document workflow:
`opengever_document_workflow` instead of `opengever_workspace_document` 🤦 

Also use PlacefulWorkflowPolicyActivator from ftw.upgrade instead of
custom implementation.

JIRA: https://4teamwork.atlassian.net/browse/GEVER-82

## Checklist (Must have)

- [x] Changelog entry


opengever/core/upgrades/20200827114625_update_workspace_workflow/upgrade.py
```diff
@@ -87,12 +88,16 @@ def set_workspace_policy(self):
             full_objects=True,
         )
         for root in workspace_roots:
-            if '.wf_policy_config' not in root.objectIds():
```
deiferni, line 90: can you `touch` the upgrade, so it is re-run for dev? that way we don't have to intervene manually.

deiferni (approved):
LGTM apart from comment.

PR #6629, opened 2020-09-09
Do not manage the "Manage properties" permission

buchi:
Fixes ZMI access to the Properties tab.

By default the "Manage properties" permission is assigned to the "edit" action group.
As we do not manage the "edit" action group in workspace workflows we have to inherit the "Manage properties" permission.

Jira: https://4teamwork.atlassian.net/browse/GEVER-82

## Checklist (Must have)

- [x] Changelog entry


deiferni (approved):
LGTM 👍

PR #6631, opened 2020-09-11
Overwrite @groups endpoint to handle the OGDS

njohner:
For Teamraum deployments we want to allow administrators to manage groups so that groups can be used to give access to workspaces. Plone already has a `@groups` endpoint, but we needed to extend it to support our OGDS as well as allow Administrators to use the endpoint. For that we:
- Overwrite the `@groups` POST, PATCH and DELETE endpoints
- Customize the endpoints to be allowed for Administrators
- Customize the endpoints to handle the OGDS
- Restrict the endpoints to handle workspace roles
- Restrict the endpoints to only manage local groups (source groups)
- Modify the `sync_ogds` to avoid overwriting the source groups managed by the `@groups` endpoint

Another difficulty that arose in that groups are enumerated and managed over the plugins, so we need to make sure that our deployments are configured correctly to avoid writing to the LDAP. For that we:
- Add a check in the `sync_ogds` to make sure group management happens over the `source_groups` plugin and that this is also enabled for enumeration
- Add upgrade step to setup current deployments accordingly
- Modify policy templates and example content accordingly

Note that we will probably still need to implement a GET or a vocabulary or a listing endpoint in order to be able to implement the frontend part of the story. But this will be done separately and should be driven by the frontend's needs.

For https://4teamwork.atlassian.net/browse/GEVER-530

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking: Not sure, whether we want to consider this breaking. I guess we do now allow less things done with the `@groups` endpoint, but on the other hand I'd anyway prefer if nobody was using it before now...
    - [ ] api-change label added
- DB-Schema migration
- [x] Change could impact client installations, client policies need to be adapted:
        I think that with the upgrade step and adaptation of `deploy.py` both already existing and newer installations should be covered

### DIFF
--- a/docs/public/dev-manual/api/users.rst
+++ b/docs/public/dev-manual/api/users.rst
@@ -118,6 +118,9 @@ sortiert zurückgegeben.
 Gruppen
 =======
 
+Gruppendetails
+--------------
+
 Details über Gruppen können mit dem ``@ogds-groups`` Endpoint abgefragt werden. Der Endpoint steht nur auf Stufe Kontaktordner zur Verfügung und erwartet eine Einschränkung auf eine Gruppe via Gruppen-ID. Die URL setzt sich somit folgendermassen zusammen:
 
 ``http://example.org/kontakte/@ogds-groups/stv_benutzer``
@@ -159,3 +162,13 @@ Nachnamen sortiert zurückgegeben.
         ],
         "items_total": 11
       }
+
+Gruppen erstellen, löschen und modifizieren
+-------------------------------------------
+
+Gruppen erstellen, modifizieren und löschen kann über den ``@groups`` Endpoint gemacht werden und ist in der `plone.restapi Dokumentation <https://plonerestapi.readthedocs.io/en/latest/groups.html>`_ beschrieben. Dieser Endpoint wurde für GEVER folgendermassen angepasst:
+
+- Die Gruppen Daten werden korrekt im OGDS abgespiegelt.
+- Er steht auch für Administratoren zur Verfügung.
+- Er wurde eingeschränkt um nur die Administration von gewissen Rollen zu erlauben: ``workspace_guest``, ``workspace_member`` und ``workspace_admin``.
+- Gruppennamen darf nicht länger als 255 Zeichen lang sein
--- a/opengever/api/group.py
+++ b/opengever/api/group.py
@@ -1,0 +1,73 @@
+# -*- coding: utf-8 -*-
+from plone.restapi.deserializer import json_body
+from plone.restapi.interfaces import ISerializeToJson
+from plone.restapi.services import Service
+from Products.CMFCore.utils import getToolByName
+from zExceptions import BadRequest
+from zope.component import queryMultiAdapter
+from zope.component.hooks import getSite
+from zope.interface import alsoProvides
+
+import plone.protect.interfaces
+
+
+class GeverGroupsPost(Service):
+    """Copy of plone.restapi.services.groups.add.GroupsPost
+    """
+
+    def reply(self):
+        portal = getSite()
+        data = json_body(self.request)
+
+        groupname = data.get("groupname", None)
+
+        if not groupname:
+            raise BadRequest("Property 'groupname' is required")
+
+        email = data.get("email", None)
+        title = data.get("title", None)
+        description = data.get("description", None)
+        roles = data.get("roles", None)
+        groups = data.get("groups", None)
+        users = data.get("users", [])
+
+        properties = {"title": title, "description": description, "email": email}
+
+        gtool = getToolByName(self.context, "portal_groups")
+        regtool = getToolByName(self.context, "portal_registration")
+
+        if not regtool.isMemberIdAllowed(groupname):
+            raise BadRequest("The group name you entered is not valid.")
+
+        already_exists = gtool.getGroupById(groupname)
+        if already_exists:
+            raise BadRequest("The group name you entered already exists.")
+
+        # Disable CSRF protection
+        if "IDisableCSRFProtection" in dir(plone.protect.interfaces):
+            alsoProvides(self.request, plone.protect.interfaces.IDisableCSRFProtection)
+
+        success = gtool.addGroup(
+            groupname,
+            roles,
+            groups,
+            properties=properties,
+            title=title,
+            description=description,
+        )
+        if not success:
+            raise BadRequest(
+                "Error occurred, could not add group {}.".format(groupname)
+            )
+
+        # Add members
+        group = gtool.getGroupById(groupname)
+        for userid in users:
+            group.addMember(userid)
+
+        self.request.response.setStatus(201)
+        self.request.response.setHeader(
+            "Location", portal.absolute_url() + "/@groups/" + groupname
+        )
+        serializer = queryMultiAdapter((group, self.request), ISerializeToJson)
+        return serializer()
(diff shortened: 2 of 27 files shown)


opengever/api/tests/test_group.py
```diff
@@ -142,3 +145,167 @@ def test_group_creation_fails_when_assigning_user_not_in_ogds(self, browser):
...
+                method='PATCH',
+                headers=self.api_headers)
+
+        self.assertEqual({u'message': u'Can only modify local groups.',
+                          u'type': u'BadRequest'},
+                         browser.json)
+        self.assertEqual(u'Test Group', self.ogds_group.title)
+
+    @browsing
+    def test_cannot_update_group_with_dissallowed_roles(self, browser):
```
lukasgraf, line 282: -> `disallowed`

opengever/api/group.py
```diff
@@ -247,16 +251,24 @@ def _get_group(self, group_id):
...
 
         portal_groups = getToolByName(self.context, "portal_groups")
-        self.group = self._get_group(self._get_group_id)
+        self.group_id = self._get_group_id
+        self.ogds_group = Group.query.get(self.group_id)
+        self.group = self._get_group(self.group_id)
 
         self.check_preconditions()
 
         delete_successful = portal_groups.removeGroup(self._get_group_id)
```
lukasgraf, line 275: I'm wondering why this doesn't require elevated privileges, like PATCH does...

opengever/ogds/base/sync/ogds_updater.py
```diff
@@ -248,7 +249,9 @@ def import_groups(self):
         # to clear out memberships from the `groups_users` association table
         # before importing them, so that memberships from groups that have
         # been deleted in LDAP get removed from OGDS.
-        for group in session.query(Group):
+        #
+        # Local groups are ignored as they should not get synced to the LDAP.
+        for group in session.query(Group).filter(Group.is_local != true()):
```
lukasgraf, line 304: I believe it should be possible to just replace the SQLAlchemy expression `true()` with Python `True` here, and let SQLAlchemy deal with type casting from Python to SQL.

opengever/ogds/base/sync/ogds_updater.py
```diff
@@ -299,6 +302,13 @@ def import_groups(self):
                             u"Skipping duplicate group '{}'!".format(groupid))
                         continue
 
+                if group.is_local:
+                    # We avoid overwriting local groups.
+                    logger.warn(
+                        u"Skipping group '{}'! "
```
lukasgraf, line 358: -> `Skipping LDAP group ...` to make it even more clear what's happening?

docs/public/dev-manual/api/users.rst
```diff
@@ -159,3 +162,13 @@ Nachnamen sortiert zurückgegeben.
         ],
         "items_total": 11
       }
+
+Gruppen erstellen, löschen und modifizieren
+-------------------------------------------
+
+Gruppen erstellen, modifizieren und löschen kann über den `@groups` Endpoint gemacht werden und ist in der `plone.restapi Dokumentation <https://plonerestapi.readthedocs.io/en/latest/groups.html>`_ beschrieben. Dieser Endpoint wurde für GEVER angepasst so dass:
```
lukasgraf, line 169: Double backticks around `@groups` and other literals below instead of single ones ;)

docs/public/dev-manual/api/users.rst
```diff
@@ -159,3 +162,13 @@ Nachnamen sortiert zurückgegeben.
...
+
+- Die Gruppen Daten werden korrekt im OGDS abgespiegelt.
```
lukasgraf, line 171: If you lead in with `so dass`, the sentence order for these bullet points would need to be so that the verbs are at the end of the sentence. I would simply change the line above to `Dieser Endpoint wurde für GEVER folgendermassen angepasst:`, then it works 😉

docs/public/dev-manual/api/users.rst
```diff
@@ -159,3 +162,13 @@ Nachnamen sortiert zurückgegeben.
...
+- Er steht auch für Administratoren zur Verfügung.
+- Er wurde eingeschränkt um nur die Administration von gewissen Rollen zu erlauben: `workspace_guest`, `workspace_member` und `workspace_admin`.
+- Gruppennamen kann nicht länger als 255 Charakter lang sein
```
lukasgraf, line 174: kann -> darf, Charakter -> Zeichen

opengever/core/upgrades/20200921135157_manage_groups_with_source_groups_plugin/upgrade.py
```diff
@@ -0,0 +1,38 @@
...
+        should all be source groups.
+        We now also need to find the created source groups, hence we need to
+        activate enumeration for the source_groups plugin.
+        """
+        acl_users = getToolByName(self.portal, 'acl_users')
+        plugins = acl_users.plugins
+        group_management_plugins = plugins.getAllPlugins('IGroupManagement')
+
+        if 'ldap' in group_management_plugins.get('active'):
+            plugins.deactivatePlugin(IGroupManagement, 'ldap')
```
lukasgraf, line 28: As discussed, we decided to not deactivate GroupManagement for the LDAP plugin, right?

lukasgraf (changes requested):
LGTM 👍 
Just some minor nitpicks, and the changes we discussed to not deactivate GroupManagement for the LDAP plugin, but instead change the upgrade step to move `source_groups` to the top, and assert on its position where appropriate.

opengever/api/group.py
```diff
@@ -150,17 +153,31 @@ def _get_group(self, group_id):
...
+            self.ogds_group.title = title
+        if description:
+            self.ogds_group.description = description
+        if users:
+            self.ogds_group.users = map(get_sql_user, users)
 
     def reply(self):
         data = json_body(self.request)
-        self.group = self._get_group(self._get_group_id)
+        self.group_id = self._get_group_id
```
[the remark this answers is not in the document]
njohner: Yes I agree, but `_get_group_id` is defined in the plone restapi as a property for this endpoint and I wanted to rather use group_id, as I also do for the `Delete` endpoint. I'll change `_get_group_id` to `get_group_id` and make it a method instead of a property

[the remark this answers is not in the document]
lukasgraf: @njohner ah, I see, didn't realize that. Then I would actually leave it as-is, for ease of diffing changes with upstream in the future.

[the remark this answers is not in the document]
njohner: Ok I removed the assignment to `self.group_id` instead

opengever/api/group.py
njohner: It seems that only the `addMember` and `removeMember` from groups need elevated privileges

njohner: I've corrected the POST endpoint accordingly

opengever/core/upgrades/20200921135157_manage_groups_with_source_groups_plugin/upgrade.py
njohner: I changed the upgrade step to activate the `source_groups` plugin for group enumeration and management and that it comes first for group management.
I also check those conditions in the `@groups` endpoint and in the `sync_ogds`

njohner:
Should be ready now for re-review.

lukasgraf (approved):
LGTM 👍

PR #6633, opened 2020-09-11
Improve ordering for assigned_groups by using the group_id as the fallback for the group title

elioschmutz:
This PR makes the following changes:

- Fix changelog of the previous merged PR
- Group titles can be `null`. We use now the group_id as the fallback row for ordering. Otherwise, all groups without a title would be unordered at the end of the response.

Jira: https://4teamwork.atlassian.net/browse/PHX-40

Follow-UP of #6626 

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations


njohner:
@elioschmutz I'd leave the sorting case-sensitive for now, as we have an open story for that and if I understood @lukasgraf correctly, we could set it to case-insensitive correctly if we want to. (see https://4teamwork.atlassian.net/browse/GEVER-982)

elioschmutz:
@njohner ok, thanks. I updated the PR

elioschmutz:
@lukasgraf I also requested your review because you wanted to take a look on it as well...

docs/HISTORY.txt
```diff
@@ -6,12 +6,12 @@ Changelog
 
 - Fix globalindex endpoint for undefined sort orders. [njohner]
 - Fix ogds listing endpoints for undefined sort orders. [njohner]
+- Order groups and teams in User serializer by title. [elioschmutz]
```
lukasgraf, line 9: This CL entry only speaks of the serializer. It doesn't address the fact that the ordering for `assigned_groups` in the core `OGDSService` now has been changed.

opengever/ogds/models/service.py
```diff
@@ -72,7 +73,10 @@ def assigned_org_units(self, userid=None, omit_current=False):
     def assigned_groups(self, userid):
         query = Group.query.join(Group.users)
         query = query.filter(User.userid == userid)
-        query = query.order_by(Group.title)
+
+        # Order by title or groupid if no title is given.
+        query = query.order_by(func.coalesce(Group.title, Group.groupid))
```
lukasgraf, line 78: I'm about worried about compatibility with Oracle and performance for `COALESCE`. This will now `ORDER BY COALESCE(title, groupid)` every single time `assigned_groups` is used somewhere - which is a lot. And we definitely don't have an index over `COALESCE(group.title, group.groupid)` that could be used.

For example, just to test whether the user is in the OrgUnit's `users_group`, which happens all the time, we definitely don't need this ordering. Is it really necessary to even do this sorting unconditionally at such a low level (inside the service)?

Either way, could you please test:
- that this works (functionally) on Oracle
- if there is a performance difference to the previous query on systems with large amounts of groups (for both PostgreSQL and Oracle)

lukasgraf (changes requested):

opengever/ogds/models/service.py
elioschmutz: Postgres (~4500 groups, sg):
- without ordering: ~2.5ms
- order by title: ~8-10ms
- COALESCE: ~25-50ms

lukasgraf: Oracle (1168 Groups, ZG):

- without ordering: ~170ms
- order by title: ~175ms
- COALESCE: ~190ms

(Without filtering by user, but via SQLAlchemy ORM).

lukasgraf (approved):
LGTM 👍

PR #6634, opened 2020-09-11
Reuse already running solr instance if available

elioschmutz:
Currently, the `bin/testserver` will fail on startup if a solr is already running. So it's not possible to run a local dev environment in addition with a running testserver.

A running solr is blocking all of its cores. Thus, running two solr servers on different ports with the same configuration is not possible. The first running server is blocking the core of the second one. This happens if you try to run the testserver with an `opengever.core` checkout where you already started a solr server manually. This is normally the case while developing.

You can tell the testserver to reuse the already running solr instead of starting an own server. To do this, set the ``SOLR_PORT`` to the running solr and the ``TESTSERVER_REUSE_RUNNING_SOLR`` to the same port.

This will tell the testserver to run a solr on port ``SOLR_PORT`` but reuse the existing server if there is already a running server.

Bildschirmfoto 2020-10-13 um 20 56 42

Bildschirmfoto 2020-10-13 um 20 57 07

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1046,6 +1046,19 @@ The ports used by the testserver can easily be changed through environment varia
 - ``ZSERVER_PORT`` - the port of the GEVER http server (default: ``55001``)
 - ``TESTSERVER_CTL_PORT`` - the port of the XMLRPC control server (default: ``55002``).
 - ``SOLR_PORT`` - the port of the Solr server which is controlled by the testserver (default: ``55003``).
+- ``TESTSERVER_REUSE_RUNNING_SOLR`` -  reuse the solr on the given port (default: ``None``).
+
+Special case for local development:
+
+A running solr is blocking all of its cores. Thus, running two solr servers on different ports with the same configuration is not possible. The first running server is blocking the core of the second one. This happens if you try to run the testserver with an `opengever.core` checkout where you already started a solr server manually. This is normally the case while developing.
+
+You can tell the testserver to reuse the already running solr instead of starting an own server. To do this, set the ``SOLR_PORT`` to the running solr and the ``TESTSERVER_REUSE_RUNNING_SOLR`` to the same port.
+
+This will tell the testserver to run a solr on port ``SOLR_PORT`` but reuse the existing server if there is already a running server.
+
+.. code::
+
+   SOLR_PORT=8983 TESTSERVER_REUSE_RUNNING_SOLR=8983 ./bin/testserver
 
 
 Custom fixtures
--- a/opengever/core/testserver.py
+++ b/opengever/core/testserver.py
@@ -15,6 +15,7 @@ from plone import api
 from plone.app.testing import applyProfile
 from plone.app.testing import FunctionalTesting
 from plone.testing import z2
+from requests.exceptions import ConnectionError
 from zope.configuration import xmlconfig
 from zope.globalrequest import setRequest
 import imp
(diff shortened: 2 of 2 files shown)


lukasgraf (changes requested):
Uhm, what? That would mean that *running tests* would possibly write to the Solr used for local development? Or possibly even a production (!) Solr that you have an SSH tunnel open to on `localhost:8983`.

lukasgraf:
@elioschmutz as discussed:
It's not quite as dangerous as I first feared, since the data is still separated into different solr cores inside the same instance (`development`, `testing` and `testserver`). But my uneasy feeling regarding `testserver` taking over control over a Solr instance it did not start itself remains.

The proper, long-term solution for this would IMHO be to separate the `testserver` Solr from the other ones at least by port, or even make it an entirely separate Solr instance. Only doing it by port doesn't seem to work (easily) because the Solr instance locks the cores on the filesystem, and prevents another instance accessing the same filesystem database from being started. And even though we only _use_ one specific core in the testserver, and an entirely different one for the development instance - Solr doesn't know that when booting, and needs to lock all the cores.

An intermediate solution could therefore be this:
Allow `bin/testserver` to reuse an already running Solr, but make it very explicit (instead of an automatism). Something like
`bin/testserver --reuse-running--solr=8983`. That could the testserver in a mode where it's given permission to _use_ that Solr, but shouldn't attempt to exercise any control over that process (don't attempt to start, and don't attempt to stop it when done).

njohner:
@elioschmutz what is the status here?

elioschmutz:
@njohner it's a draft PR again.
I had no time to update it. But I saw, that it's not easy to extend the bin/testserver script with additional arguments...

jone:
> Currently, the bin/testserver will fail on startup if a solr is already running. So it's not possible to run a local dev environment in addition with a running testserver.

@elioschmutz I think that is not true. The dev solr runs on `8983`, the testerver solr on `55003` by default.
You probably ran multiple testserver instances in parallel. You'd have to change the ports (e.g. through `SOLR_PORT`) when you do that.

elioschmutz:
> > Currently, the bin/testserver will fail on startup if a solr is already running. So it's not possible to run a local dev environment in addition with a running testserver.
> 
> @elioschmutz I think that is not true. The dev solr runs on `8983`, the testerver solr on `55003` by default.
> You probably ran multiple testserver instances in parallel. You'd have to change the ports (e.g. through `SOLR_PORT`) when you do that.

Unfortunately no:

Bildschirmfoto 2020-10-07 um 15 59 07

jone:
@elioschmutz you seem to have the `SOLR_PORT` environment variable configured in your shell. This has the effect that you override the port for both, dev and testserver, with the same value. That's why you have this problem.
See `export | grep SOLR_PORT`.
You either need to unset the variable in your shell or you need to prefix each command (e.g. `SOLR_PORT=1234 bin/testserver`).

elioschmutz:
No... 
Bildschirmfoto 2020-10-07 um 15 59 07

either something is completely wrong with my deployment or I missunderstand anything... but I can't get it run. The main solr is blocking all solr-cores on startup. So the testserver solr can't access the `testserver` core. See also the respons https://github.com/4teamwork/opengever.core/pull/6634#issuecomment-691090148

jone:
@elioschmutz `$SOLR_PORT` must not (!) be defined in your shell. Your screenshot shows that you have defined it, so your system is configured wrong 😉 

This is correct:

Bildschirmfoto 2020-10-12 um 09 26 00

elioschmutz:
No...

Bildschirmfoto 2020-10-12 um 10 06 29

and no...

Bildschirmfoto 2020-10-12 um 10 06 36

jone:
@elioschmutz you are right: I assumed that the problem is the shared port, but now I think that the problem is that the same data directory and lockfile is used.
I think that if you want to have two solrs running at the same time you either need two checkouts or two solr buildout parts (=two part directories, two war directories, two bin scripts)..

elioschmutz:
> @elioschmutz you are right: I assumed that the problem is the shared port, but now I think that the problem is that the same data directory and lockfile is used.
> I think that if you want to have two solrs running at the same time you either need two checkouts or two solr buildout parts (=two part directories, two war directories, two bin scripts)..

Yes, and thats exactly what I don't want 😉  I don't want to maintain two gever-checkouts all the time. I just wand to start my testserver beside of my gever if I need to. That was the reason why I opened this pr. It just reuses the already running solr.

In the end, I tried to extend the `bin/testserver` to accept some parameters as @lukasgraf suggested in his response. But I couldn't find a way how to pass additional arguments. Zope is in between and already consumes the arguments and will raise an error if there are not valid arguments.

jone:
> In the end, I tried to extend the `bin/testserver` to accept some parameters as @lukasgraf suggested in his response. But I couldn't find a way how to pass additional arguments. Zope is in between and already consumes the arguments and will raise an error if there are not valid arguments.

The argument parsing is part of the [robotframework server](https://github.com/plone/plone.app.robotframework/blob/master/src/plone/app/robotframework/server.py#L177).
I see two possible solutions:

1) copy the `server()` function into our code base and extend it. The challenge is then how to pass the information along.
2) use environment variables (e.g. `TESTSERVER_REUSE_RUNNING_SOLR=8983 bin/testserver`). You can easily read that in the solr testing layer code.

Personally, I'd go for 2). But I'm not in charge here 😜

elioschmutz:
Thanks for your input @jone . I like the second solution as well... @lukasgraf what do you think? Would it be an acceptable solution for you as well?

lukasgraf:
Option 2) sounds good to me as well 👍

lukasgraf (approved):
LGTM 👍 

I fear it doesn't make the logic for selecting the Solr port under all the possible conditions any easier to understand, especially for someone new to the code, but I don't currently have a clear suggestion how to do it any better.

PR #6636, opened 2020-09-11
Fix tus replace

buchi:
Customize `@tus-replace` endpoint to perform additional checks before replacing a file:
- The document must be checked out by the current user
- If locked, the lock token must be provided in the request
- If it's a proposal, the file must have a .docx extension

The main difference with the implementation in `quick_upload.py` is, that you must be able to upload a file if it's locked by you, because Office Connector always locks the file.

However the current Office Connector release (1.10.0) does not yet send the lock token when replacing a file. :-(
The check for the correct lock token would break uploading files with the current OC release.

As we need some time to ensure a wide rollout of an OC release with lock token support, the lock check is currently disabled. There's a test verifying this functionality which is skipped for one year. In one year we should be ready to enable the lock token check. If not, the test will fail.

Because of that

Jira: https://4teamwork.atlassian.net/browse/GEVER-975

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [x] api-change label added

### DIFF
--- a/opengever/api/tests/test_tus.py
+++ b/opengever/api/tests/test_tus.py
@@ -1,0 +1,131 @@
+from ftw.testbrowser import browsing
+from opengever.testing import IntegrationTestCase
+from six import BytesIO
+
+UPLOAD_DATA = b"abcdefgh"
+UPLOAD_LENGTH = len(UPLOAD_DATA)
+UPLOAD_METADATA = 'filename dGVzdC50eHQ=,content-type dGV4dC9wbGFpbg=='
+
+
+class TestTUSUpload(IntegrationTestCase):
+
+    def setUp(self):
+        super(TestTUSUpload, self).setUp()
+
+    def prepare_tus_replace(self, doc, browser, headers=None):
+        browser.open(
+            doc.absolute_url() + '/@tus-replace',
+            method='POST',
+            headers=dict({
+                "Accept": "application/json",
+                "Tus-Resumable": "1.0.0",
+                "Upload-Length": str(UPLOAD_LENGTH),
+                "Upload-Metadata": UPLOAD_METADATA,
+            }, **headers or {}),
+        )
+        self.assertEqual(browser.status_code, 201)
+        return browser.headers.get('Location')
+
+    def assert_tus_replace_succeeds(self, doc, browser, headers=None):
+        location = self.prepare_tus_replace(doc, browser, headers=headers)
+        browser.open(
+            location,
+            method='PATCH',
+            headers=dict({
+                "Accept": "application/json",
+                "Content-Type": "application/offset+octet-stream",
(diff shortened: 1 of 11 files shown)


opengever/api/tests/test_tus.py
```diff
@@ -0,0 +1,138 @@
...
+        datetime.now() < datetime(2021, 9, 11),
+        "Lock verification temporary disabled, because it's not yet supported "
+        "by Office Connector",
+    )
+    @browsing
+    def test_cannot_replace_document_if_locked_by_other(self, browser):
+        self.login(self.regular_user, browser)
+        self.checkout(self.document, browser)
+        self.lock(self.document, browser)
+
```
deiferni, line 116: i would expect you login with another user here? could you double check this test is correct?

opengever/api/tus.py
```diff
@@ -0,0 +1,55 @@
...
+
+        manager = getMultiAdapter((self.context, self.context.REQUEST),
+                                  ICheckinCheckoutManager)
+        if not manager.is_checked_out_by_current_user():
+            raise Forbidden("Document not checked out.")
+
+        # XXX: Currently not supported by the latest Office Connector 1.10.0
+        # Enable this after a grace period when OC > 1.10.0 has been rolled out
+        # to customers.
+        if manager.is_locked_by_other():
```
deiferni, line 35: you want to comment this out then, right 🤔 ?

deiferni (approved):
LGTM apart from comments and:
- what is the desired behavior with mails? we don't offer checkin/checkout for mails via ui, but should we somehow make sure `@tus-replace` blocks in such cases, too?

opengever/api/tests/test_tus.py
buchi: No, this tests the case when the document is locked but the correct lock token is not provided in the request.
This request can be made by the same user or by another user. But if it's made by another user, it will already fail earlier because the document is checked out by another user.

opengever/api/tus.py
buchi: Righ!

opengever/api/tests/test_tus.py
deiferni: ah ok maybe the test-name confused me. suggestion, rename to :`test_cannot_replace_document_if_lock_token_is_missing` or something similar?

buchi:
@deiferni Regarding mails, currently it's not possible to checkout mails. However mail messages can be replaced using the REST API with `PATCH` and  `@tus-replace` which is wrong, I guess. Not sure how we want to solve this.

opengever/api/tests/test_tus.py
buchi: amended

PR #6638, opened 2020-09-16
Add current_user to @config endpoint and remove userid, user_fullname and user_email.

tinagerber:
Instead of individual attributes of a user, the entire serialized user object is now returned in the `@config` endpoint.

This also requires some changes in the frontend: https://github.com/4teamwork/gever-ui/pull/1276

Jira: https://4teamwork.atlassian.net/browse/GEVER-698

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [x] api-change label added


lukasgraf:
@tinagerber I would like to see an assessment of potential performance impacts for this.

From the PR checklist:
> The author should advertise and sell the change in the PR body

Sell it to us - have considered any potential performance impacts (this endpoint gets called quite often, and is on the critical path for bootstrapping the app on the first load) and concluded that it should be ok?

I'm not asking for a detailed performance study, but maybe something like _"I've checked what serializing the entire user objects involves, and it does this and that, and therefore I think it should be ok"_.

tinagerber:
@lukasgraf Serializing the entire user objects involves the serialization of the user schema and the listing of all roles and principals through a catalog query.

lukasgraf (approved):
Thanks!
LGTM 👍 if it doesn't have a significant impact on performance.
As discussed on 📞 , the fact that the catalog now gets involved had me worried a bit, but might very well be that it still doesn't make a noticeable difference.

tinagerber:
I have now measured the performance and found no significant difference.

PR #6639, opened 2020-09-17
Do not manage security in deactivated workspaces

tinagerber:
With this PR we achieve that local roles can no longer be changed in deactivated workspaces. This also means that no new members can be invited.

Jira: https://4teamwork.atlassian.net/browse/GEVER-1017

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)


njohner (approved):
LGTM

PR #6640, opened 2020-09-21
Fix `status` API endpoint for mails.

deiferni:
Fix the `status` service endpoint when it is used for mails. Also port existing functional tests to separate integration test-cases and attempt to improve readability of test-cases and in case of failures.

Jira: https://4teamwork.atlassian.net/browse/GEVER-1048

## Checklist (Must have)

- [x] Changelog entry


lukasgraf (approved):
LGTM 👍

PR #6642, opened 2020-09-22
Don't resolve or deactivate a dossier if it is linked to an active workspace

tinagerber:
With this PR a dossier can no longer be resolved or deactivated if it is linked to an active workspace.
To have access to the WorkspaceClient you need the permission `opengever.workspaceclient: Use Workspace Client`. I have achieved this with the `elevated_privileges` context manager. Maybe there is a better way to do this?

Jira: https://4teamwork.atlassian.net/browse/CA-995

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- New translations

### DIFF
--- a/opengever/dossier/tests/test_deactivate.py
+++ b/opengever/dossier/tests/test_deactivate.py
@@ -6,14 +6,19 @@ from ftw.testbrowser import browsing
 from ftw.testbrowser.pages import editbar
 from ftw.testbrowser.pages import statusmessages
 from ftw.testbrowser.pages.statusmessages import error_messages
+from ftw.testbrowser.pages.statusmessages import info_messages
 from ftw.testing import freeze
 from opengever.base.role_assignments import RoleAssignmentManager
 from opengever.base.role_assignments import SharingRoleAssignment
 from opengever.dossier.behaviors.dossier import IDossier
 from opengever.dossier.behaviors.protect_dossier import IProtectDossier
 from opengever.testing import IntegrationTestCase
+from opengever.workspaceclient.interfaces import ILinkedWorkspaces
+from opengever.workspaceclient.tests import FunctionalWorkspaceClientTestCase
 from plone import api
+from plone.protect import createToken
 import json
+import transaction
 
 
 class TestDossierDeactivation(IntegrationTestCase):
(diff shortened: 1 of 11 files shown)


opengever/dossier/tests/test_deactivate.py
```diff
@@ -179,3 +184,81 @@ def test_deactivating_dossier_non_recursively_is_forbidden(self, browser):
                 u'type': u'Bad Request'}},
             browser.json)
         self.assert_workflow_state('dossier-state-active', self.empty_dossier)
+
+
+class TestDeactivateDossierWithWorkspaceClientFeatureEnabled(FunctionalWorkspaceClientTestCase):
```
njohner, line 189: TestDossierDeactivationWithWorkspaceClientFeatureEnabled? for consistency with the other test class here?

opengever/dossier/tests/test_resolve.py
```diff
@@ -1448,6 +1451,53 @@ class TestResolveConditionsRESTAPI(ResolveTestHelperRESTAPI, TestResolveConditio
     pass
 
 
+class TestResolveDossierWithWorkspaceClientFeatureEnabled(FunctionalWorkspaceClientTestCase,
```
njohner, line 1454: `TestResolveConditionsWithWorkspaceClientFeatureEnabled`? again for consistency

opengever/dossier/tests/test_resolve.py
```diff
@@ -1448,6 +1451,53 @@ class TestResolveConditionsRESTAPI(ResolveTestHelperRESTAPI, TestResolveConditio
...
+            browser.open(self.workspace,
+                         view='content_status_modify?workflow_action='
+                              'opengever_workspace--TRANSITION--deactivate--active_inactive',
+                         data={'_authenticator': createToken()})
+            self.resolve(self.dossier, browser)
+            self.assert_resolved(self.dossier)
+            self.assert_success(self.dossier, browser,
+                                ['The dossier has been succesfully resolved.'])
+
+
```
njohner, line 1500: I'd also test over restapi:
`TestResolveConditionsWithWorkspaceClientFeatureEnabledRESTAPI(ResolveTestHelperRESTAPI, TestResolveDossierWithWorkspaceClientFeatureEnabled)`

njohner (approved):
We could add ways to have access to the `Workspace` client without having the `opengever.workspaceclient: Use Workspace Client` permission, maybe by adding an `unrestricted_session` to the `WorkspaceClient`, but in the end, we would still need to use `elevated_privileges` to actually make the request to the other client, as we need to make sure that all workspaces get listed. So I think simply using `elevated_privileges` here is fine.

So apart from a few naming suggestions and also testing dossier resolution over the API, LGTM.

opengever/dossier/tests/test_resolve.py
tinagerber: 👍🏻 With the TraversalDriver it didn't work after all, so I now set an empty data list.

PR #6645, opened 2020-09-24
Provides the IVocabularyTokenized interface for elephant vocabularies.

elioschmutz:
ℹ️ TO THE REVIEWWER ℹ️ 
This PR is kind of changing the current api because it now handles elephantvocabularies properly. I'm not sure, if this change have to be mentioned as breaking api change or not. It was broken before, but it behaves differently now...

---------

[ElephantVocabularies](https://github.com/collective/collective.elephantvocabulary) are wrapper vocabularies around zope.schema vocabularies to restrict vocabulary entries.

The vocabulary factory for the schema field ([example schema field in gever](https://github.com/4teamwork/opengever.core/blob/485762035774a90c8a76ecda5c6be44901749092/opengever/document/behaviors/metadata.py#L117)) is created by the [`wrap_vocabulary`](https://github.com/collective/collective.elephantvocabulary/blob/master/collective/elephantvocabulary/__init__.py#L7) which creates a [`VocabularyFactory`](https://github.com/collective/collective.elephantvocabulary/blob/master/collective/elephantvocabulary/vocabulary.py#L16).
The factory then creates an instance of the `wrapper_class` which defaults to the [`WrapperBase`](https://github.com/collective/collective.elephantvocabulary/blob/master/collective/elephantvocabulary/base.py#L9) of `collective.elephantvocabulary`.All attribute calls on the wrapper-base instance are proxied to the underlying vocabulary.

This works all fine.

The problem is on the provided interfaces of the `WrapperBase` class. It does not respect the provided interfaces of the underlying vocabulary. Which means, serializing such a field will not serialize properly because it does not provide the right interfaces: [plone.restapi:serializer/dxfields](https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/serializer/dxfields.py#L51)

The serializer just returns the token of the vocabulary:

Bildschirmfoto 2020-09-24 um 21 10 57

This PR provides a new `wrap_vocabulary` method which changes the default `wrapper_class` to a new customized `WrapperBase` class which also provides the `IVocabularyTokenized` interface if the underlying vocabulary does. 

With this change, elephantvocabulary fields gets serialized properly:

Bildschirmfoto 2020-09-24 um 21 12 09

Jira: https://4teamwork.atlassian.net/browse/NE-39

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/api/tests/test_vocabularies.py
+++ b/opengever/api/tests/test_vocabularies.py
@@ -1,9 +1,12 @@
+from collective.elephantvocabulary.interfaces import IElephantVocabulary
 from ftw.testbrowser import browsing
+from opengever.api.schema.sources import get_field_by_name
 from opengever.base.behaviors.classification import IClassification
 from opengever.ogds.base.ou_selector import CURRENT_ORG_UNIT_KEY
 from opengever.testing import IntegrationTestCase
 from opengever.testing import SolrIntegrationTestCase
 from plone import api
+from plone.dexterity.utils import iterSchemata
 
 
 NON_SENSITIVE_VOCABUALRIES = [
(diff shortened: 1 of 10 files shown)


opengever/api/tests/test_vocabularies.py
```diff
@@ -510,3 +513,24 @@ def test_get_vocabulary_for_add(self, browser):
                            u'question', u'regulations', u'report', u'request']
         self.assertItemsEqual(expected_tokens,
                               [item['token'] for item in response.get('items')])
+
+
+class ElephantVocabularies(IntegrationTestCase):
```
njohner, line 518: `ElephantVocabularies` -> `TestElephantVocabularies`

njohner (approved):
Apart from comment, LGTM

PR #6647, opened 2020-09-28
bla

njohner:
_PR-Description: should contain all the information necessary for an outsider to understand the change without looking at the code or the issue!_

- _Why the change is necessary_
- _What the goal of the change is_
- _How change is achieved (e.g. design decisions)_
- _The author should advertise and sell the change in the PR body_

_Screenshot: whenever useful, but only as a visual aid._

## Checklist (Must have)

- [ ] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations


deiferni (changes requested):
blabla

opengever/api/tests/test_dossiertemplate.py
```diff
@@ -0,0 +1,147 @@
...
+        self.assertEqual(browser.status_code, 200)
+        self.assertEqual(browser.json[u'@id'],
+                         self.dossiertemplate.absolute_url())
+        self.assertEqual(browser.json[u'@type'],
+                         u'opengever.dossier.dossiertemplate')
+        self.assertEqual(browser.json[u'title'],
+                         self.dossiertemplate.title)
+        self.assertEqual(browser.json[u'title_help'],
+                         self.dossiertemplate.title_help)
+        self.assertEqual(
```
deiferni, line 25: `is_subdossier` is now missing here, i think that would be required for the ui to display different icons. we can also have subdossiers in templates.

deiferni (changes requested):
LGTM 👍  apart from comments. Some additional thoughts:
- have you double checked usage of `IDossier`? there may be cases where things "just worked" for now.
- maybe this is out of scope but i've been thinking that we could get rid of `AnnotationStorage`  for `IDossierTemplate`? IMO the only benefit the storage would have is for overlapping field names, which we don't have. what are your thoughs?

opengever/dossier/dossiertemplate/indexers.py
```diff
@@ -11,10 +16,40 @@ def DossierTemplateSubjectIndexer(obj):
...
     """Make dossier templates full text searchable."""
 
+    def __init__(self, context):
+        self.context = context
+
+    def __call__(self):
+        searchable = []
+        # append some other attributes to the searchableText index
+        # reference_number
+        refNumb = getAdapter(self.context, IReferenceNumber)
```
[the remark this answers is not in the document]
njohner: yeah they have just the repository root. It was indexed up to now, so I left it, but I'm fine with dropping that

opengever/api/tests/test_dossiertemplate.py
njohner: `is_subdossier` was already missing, that hasn't changed ;-).
Funny thing is that it is in the summary serialization (https://github.com/4teamwork/opengever.core/blob/master/opengever/api/serializer.py#L241-L255) and it's in there when the brain gets serialized (https://github.com/4teamwork/opengever.core/blob/master/opengever/api/serializer.py#L265-L268) so probably for navigation and such. I'll add it for the normal serializer as well.

njohner:
Ok this is ready for re-review. I've made the changes according to your feedback.
- I've tried to make sure that there was no problematic usage of `IDossier`, although it is not easy to be certain. But I've checked as best I could.
- As discussed I ended up leaving `IDossierTemplate` on the `AnnotationStorage`, as switching to attribute storage is not easily possible (because we have a specific marker interface, we need a factory).

I've also made one more change, to make sure to keep `is_subdossier` up to date for all types supporting it. I stumbled upon this while making the other changes, and while it was not problematic that the event handlers did a priori not update `is_subdossier` for `DossierTemplate`s, because they cannot be moved, I felt that this would most likely bite us at some point in the future, when, for example, we decide to not forbid moving `DossierTempalte`s in the new UI...

deiferni (approved):
LGTM 👍 

Nothing i insist on, only if it is not too much trouble:your last commit is a bit weird and could maybe be squashed with `Add supports_is_subdossier helper function.`?

njohner:
Ah sorry, this was meant to be squashed indeed (as said in the commit message 😅 )

PR #6648, opened 2020-09-29
Also activate group introspection for source_groups

elioschmutz:
This is a follow-up PR for #6631 

Adding and getting added group-members of a group through the @groups endpint is currently not possible because the `IGroupIntrospection` plugin is not activated for `source_groups` by default.

Bildschirmfoto 2020-09-29 um 10 21 42
Bildschirmfoto 2020-09-29 um 10 21 25

Adding users and getting users rely on the `getGroupMemberIds` method on a group which does not return any results if the `IGroupIntrospection` plugin is not activated. See https://github.com/4teamwork/opengever.core/blob/master/opengever/api/group.py#L212, won't return any results:

To be able to properly get all group members of a local group, it's required to activate the `IGroupIntrospection` plugin for the `source_groups`. This PR extends the existing upgradestep and updates the gever-setup.

Jira: https://4teamwork.atlassian.net/browse/CA-530

## Checklist (Must have)

- [x] Changelog entry ⚠️ changelog already exists from the previous PR: #6631 

## Checklist (if applicable)

- API change:
  - If breaking:
    - [x] api-change label added
- New functionality:
- DB-Schema migration
- New translations


opengever/core/upgrades/20200921135157_manage_groups_with_source_groups_plugin/upgrade.py
```diff
@@ -48,3 +49,10 @@ def __call__(self):
             plugins.activatePlugin(IGroupEnumerationPlugin, 'source_groups')
         elif 'source_groups' not in group_enumeration_plugins.get('active'):
             LOG.error('Source groups is not available for group enumeration')
+
+        # Make sure that source_groups is activated for  group introspection
```
njohner, line 53: here too

PR #6651, opened 2020-09-30
Also sync the group title with the ogds-group when adding a new group

elioschmutz:
This is a follow-up PR for #6631

This PR fixes an issue where the group title have not been synchronized with the ogds-group when creating a new group with the `@groups` endpoint.

Jira: https://4teamwork.atlassian.net/browse/CA-530

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/api/group.py
+++ b/opengever/api/group.py
@@ -116,7 +116,7 @@ class GeverGroupsPost(Service):
         # Add group to ogds
         session = create_session()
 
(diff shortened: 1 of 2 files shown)


njohner (approved):
Thanks! LGTM

PR #6652, opened 2020-09-30
Prevent setting invalid reference prefix number via API.

deiferni:
When creating repository folders via API so far the `reference_number_prefix` was not validated. Theoretically users could have chosen duplicate prefixes. This PR adds validation when creating repository folders via API.

I've decided to implement validation in the deserializer as it has all the information necessary.

- It knows whether we are adding or editing
- It has all the data, context and parent can be accessed easily

Unfortunately an invariant cannot be used as it does not know whether we are adding or editing. Also the add form does provide different data during add than the rest api. In the form context is unavailable while adding, while in the rest API the object to be validated is already created and passed as as context.

We also we leave validation via `WidgetValidatorDiscriminators` in place for now to have validation in the form.

I have not added an upgrade to fix potentially broken content as we don't expect such content to actually have been created.

_I could not find a way to consistently use the same validation path for forms and API. If anyone can provide insights here, please do enlighten me!_

Jira: https://4teamwork.atlassian.net/browse/CA-1080

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/api/repositoryfolder.py
+++ b/opengever/api/repositoryfolder.py
@@ -1,10 +1,49 @@
+from Acquisition import aq_inner
+from Acquisition import aq_parent
 from opengever.api.serializer import GeverSerializeFolderToJson
 from opengever.base.interfaces import IReferenceNumber
+from opengever.base.interfaces import IReferenceNumberPrefix as PrefixAdapter
 from opengever.repository.interfaces import IRepositoryFolder
+from plone.restapi.deserializer import json_body
+from plone.restapi.deserializer.dxcontent import DeserializeFromJson
+from plone.restapi.interfaces import IDeserializeFromJson
 from plone.restapi.interfaces import ISerializeToJson
+from zExceptions import BadRequest
 from zope.component import adapter
 from zope.interface import implementer
 from zope.interface import Interface
+
+
+@implementer(IDeserializeFromJson)
+@adapter(IRepositoryFolder, Interface)
+class DeserializeRepositoryFolderFromJson(DeserializeFromJson):
+
+    def __call__(self, validate_all=False, data=None, create=False):
+        if data is None:
+            data = json_body(self.request)
+
+        number = data.get('reference_number_prefix')
+        if number:
+            self.validate_reference_number(number, create=create)
+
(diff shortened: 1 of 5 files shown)


opengever/api/repositoryfolder.py
```diff
@@ -1,12 +1,51 @@
...
+            is_valid = PrefixAdapter(parent).is_valid_number(number)
+        else:
+            is_valid = PrefixAdapter(parent).is_valid_number(number,
+                                                             self.context)
+        if not is_valid:
+            msg = "The reference_number {} is already in use.".format(number)
+            error = {
+                "message": msg,
+                "field": "reference_number",
+                "error": "ValidationError"
```
njohner, line 44: Hum I think you just invented a 5th way of returning error through the API.
It seems that for schema validation we usually have a list in the message string and not a dictionary (see See https://4teamwork.atlassian.net/browse/CA-408.).

njohner (commented):
I'm not sure about the format of the exception. Apart from that LGTM.

opengever/api/repositoryfolder.py
deiferni: The thing is, this will be serialized into a string by the api, see tests. This is IMO a bit stupid, but at least consistent with what the api does.

deiferni: ah, no sorry, i see what you mean, it should be a list of errors. will amend.

njohner: yeah and in the list the errors are in a weird tuple format instead of a dictionary 😓

njohner: but it's stupid anyway to send that back a string, I can only agree there...

deiferni: the format is correct IMO, only the list is missing. See https://github.com/4teamwork/opengever.core/blob/174da57838c182b26d9df2a5842d7e0965397294/opengever/api/tests/test_document.py#L123-L126 for a "native" API example.

deiferni: Amended.

njohner (approved):
LGTM

PR #6658, opened 2020-10-05
Register OGDS group and user endpoints on the plone siteroot instead on a contactfolder

elioschmutz:
The endpionts `@ogds-users`, `@ogds-groups`, `@ogds-user-listing` and `@ogds-group-listing` are currently registered on a contactfolder. This means, we can only list groups and users or fetch additional information if we also have a contactfolder on the plonesite. Registering it only on a contact-folder seems to be wrong because ogds-users and ogds-groups have nothing to do with the contactfolder. It should be possible to fetch users and groups even if there is no contactfolder.

This PR registers the endpoint `@ogds-users`, `@ogds-groups`, `@ogds-user-listing` and `@ogds-group-listing` on the plone site root. The endpoints are no longer available on the contactfolder.

Jira: https://4teamwork.atlassian.net/browse/NE-83

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [x] api-change label added
- New functionality:
- DB-Schema migration
- New translations


deiferni (commented):
just for the record, we've registered it on the contact folder to able to restrict access via contact folder permissions, namely to `Reader` on the contact folder. I think we can loosen that restriction as IIRC we don't actually restrict contact folder `View` in any case.

With this change we bind permission to what is set on the plone site, which is probably fine. Please double-check the permissions and make sure that we don't expose the endpoint to `Anonymous` which would be bad.

deiferni (commented):
LGTM apart from thoughts in previous comment.

elioschmutz:
> just for the record, we've registered it on the contact folder to able to restrict access via contact folder permissions, namely to `Reader` on the contact folder. I think we can loosen that restriction as IIRC we don't actually restrict contact folder `View` in any case.
> 
> With this change we bind permission to what is set on the plone site, which is probably fine. Please double-check the permissions and make sure that we don't expose the endpoint to `Anonymous` which would be bad.

Every endpoint has the `zope2.View` permission. And gever or teamraum is never accessible by anonymous users. The `@users` endpoint was already accessible with the same premission on the plone root: https://github.com/4teamwork/opengever.core/pull/5779

So it should be fine 👍

elioschmutz:
> LGTM apart from thoughts in previous comment.

could you approve?

deiferni (approved):

PR #6660, opened 2020-10-05
Add dossiertemplates to @listing endpoint.

tinagerber:
With this PR it is now possible to list dossiertemplates with the `@listing` endpoint.
This has changed the order of the listing stats.

Jira: https://4teamwork.atlassian.net/browse/NE-93

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:


opengever/api/tests/test_listing.py
```diff
@@ -953,3 +953,34 @@ def test_task_templates_listing(self, browser):
...
+        self.login(self.regular_user, browser=browser)
+        query_string = '&'.join((
+            'name=dossiertemplates',
+            'columns=title',
+            'columns=description',
+            'columns=changed',
+            'sort_on=changed',
+            'sort_order=ascending',
+        ))
+        view = '?'.join(('@listing', query_string))
```
deiferni, line 968: using join for something which can always only be two params is a IMO bit weird. i would have expected you to use `format` here, for example.

opengever/api/tests/test_listing_stats.py
```diff
@@ -61,14 +61,15 @@ def test_listing_stats_full_response(self, browser):
                 u'@id': u'{}/@listing-stats'.format(self.dossier.absolute_url()),
                 u'facet_pivot': {
                     u'listing_name': [
-                        {u'count': 12, u'field': u'listing_name', u'value': u'documents'},
```
deiferni, line 64: we don't particularly care about order here, right? So maybe it'd be better to switch the assertion here to only assert on the items, but not on order. E.g.: `assertItemsEqual(expected, response.json['facet_pivot']['listing_name']`.

deiferni (approved):
LGTM 👍 , only two optional suggestions.

PR #6661, opened 2020-10-05
Manage groups in teamraum

elioschmutz:
Currently, it's only possible to let users participating on a teamraum. 

This PR extends the participation management in a teamraum to also manage groups.

Jira: https://4teamwork.atlassian.net/browse/NE-71

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,7 +4,7 @@ Changelog
 2020.12.0 (unreleased)
 ----------------------
 
-- Nothing changed yet.
+- Allow assigning groups as participants to a Teamraum [elioschmutz]
 
 
 2020.11.0 (2020-10-07)
--- a/opengever/api/participations.py
+++ b/opengever/api/participations.py
@@ -1,6 +1,9 @@
 from opengever.base.role_assignments import RoleAssignmentManager
 from opengever.base.role_assignments import SharingRoleAssignment
+from opengever.ogds.base.actor import Actor
+from opengever.ogds.base.actor import OGDSUserActor
 from opengever.workspace.activities import WorkspaceWatcherManager
+from opengever.workspace.interfaces import IWorkspaceFolder
 from opengever.workspace.participation import PARTICIPATION_ROLES
 from opengever.workspace.participation.browser.manage_participants import ManageParticipants
 from plone import api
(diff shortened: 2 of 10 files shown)


docs/HISTORY.txt
```diff
@@ -4,6 +4,7 @@ Changelog
 2020.11.0 (unreleased)
 ----------------------
 
+- Let groups participating on a teamraum. [elioschmutz]
```
njohner, line 7: -> Let groups participate in a Teamraum? OR Allow assigning groups as participants to a Teamraum?

opengever/api/participations.py
```diff
@@ -32,23 +32,30 @@ def publishTraverse(self, request, name):
         return self
 
     def prepare_response_item(self, participant):
-        userid = participant.get('token')
+        actor = Actor.lookup(participant.get('token'))
         role = PARTICIPATION_ROLES.get(participant.get('roles')[0])
-        member = api.user.get(userid=userid)
         managing_context = self.context.get_context_with_local_roles()
+        serialized_actor = getMultiAdapter(
+            (actor.represents(), self.request),
```
njohner, line 39: Ok I don't really like that. You serialize the object that the actor represents, which gives a mix of different data depending on what the actor is, and then you return a custom serialization of the actor, where you barely use any data from the `serialized_actor` anyway.
I'd much rather that you register a serializer for the `Actor` class, and this one should return the data you want. So:
- Add an `api_url` or something alike to the `Actor` classes (as described in https://4teamwork.atlassian.net/browse/CA-406)
- In the serializer of the actor `@id = self.context.api_url` and `@type = self.context.represents.portal_type` or whatever

njohner (changes requested):
Overall looks good. The serialization bothers me a bit though.

opengever/api/participations.py
elioschmutz: Discussed with @njohner over phone-call:

The `@participations` endpoint is optimized for the teamraum. We now use an `Actor` to be able to represent groups and users in the `participant` porperty of a serialized participation. A `participant` is not an Actor. We need special attributes like `is_local` or `active`. Using an `Actor` serializer here would not give us the right properties which we use for a participation. In addition, implementing an `Actor` serializer would need some more specifications because it will be used also for other parts of gever like the watchers (https://4teamwork.atlassian.net/browse/CA-406). So I would be blocked for something we don't really need right now and which should be solved in general.

The reason why I'm using the serialized actor representer and then extract the desired attributes is because I need the serialized attributes i.e. `@id` which will be generated by the user or group serializer. I would need to reimplement it just for this endpoint which does not makes sense. Not using the full serialized version is because I need a standardized json schema for each participant.

I'll add a comment and will link it in https://4teamwork.atlassian.net/browse/CA-406 to make sure we can take another look on this part as soon as we solved the actor serialization in general.

I'll overtake the full responsibility of this change.

opengever/base/role_assignments.py
```diff
@@ -314,6 +314,13 @@ def get_assignments_by_principal_id(self, principal_id):
         return [RoleAssignment.get(**data) for data
                 in self.storage.get_by_principal(principal_id)]
 
+    def get_roles_by_principal_id(self, principal_id):
+        roles = set()
+        for assignment_data in self.storage.get_by_principal(principal_id):
+                roles.update(assignment_data.get('roles', []))
+
+        return list(roles)
```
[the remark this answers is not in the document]
elioschmutz: no, not really... i'll remove it

elioschmutz (commented):
@njohner I updated the pr with your requested changes. Could you take another look?

opengever/api/participations.py
elioschmutz: @njohner see: https://github.com/4teamwork/opengever.core/pull/6661/files#diff-48baca4ba82bb2c2f0e5c0616608a3a0R38-R50

PR #6664, opened 2020-10-06
Don't make dossiers expandable with participations.

tinagerber:
We do not want dossiers to be expandable with participations, as this is bad for performance.
A changelog entry for the implementation of the @participations endpoint already exists.

Jira: https://4teamwork.atlassian.net/browse/NE-89

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:

### DIFF
--- a/docs/public/dev-manual/api/dossier_participations.rst
+++ b/docs/public/dev-manual/api/dossier_participations.rst
@@ -64,39 +64,6 @@ Ein Beteiligter kann in verschiedenen Rollen an einem Dossier beteiligt sein. Mi
       }
 
 
-Beteiligungen als erweiterbare Komponente
------------------------------------------
-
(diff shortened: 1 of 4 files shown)


deiferni (changes requested):
ehmm. have you considered profiling it and to get to the source of the performance hit? maybe it can be optimized. so currently you are just working around the issue by distributing the performance impact onto multiple requests, which is good for frontend repsonse time but not so good for the backend. also it will eliminate expansion from a place where it is IMO  "the right thing do to". we can have a look at this issue, if you want.

deiferni:
or, to clarify this a bit more, you don't *have to* use expansion, for expandable components. the idea is to provide some *potentially* expandable components, which can either be expanded in the initial request, or requested separately via the expansions `@id` which would be `https://example.org/ordnungssystem/fuehrung/dossier-1/@participations` in your case. It is up to the client what is expanded in the initial request and what is loaded separately.

tinagerber:
We decided to leave the expansion in, since you don't have to use it if you don't want to.

PR #6665, opened 2020-10-07
Makes the GET @groups endpoint available with the `opengever.api.ManageGroups` permission.

elioschmutz:
The `POST`, `PATCH` and `DELETE` verbs vor the `@groups` endpoint are already using the `opengever.api.ManageGroups` permission. We also need this permission for the `GET` verb to make sure, an administrator can manage groups properly.

The default permission assiged by the `plone.restapi` is `cmf.ManagePortal`

Issuer: https://4teamwork.atlassian.net/browse/NE-83

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.11.0 (unreleased)
 ----------------------
 
+- GET @groups endpoint is now available with the `opengever.api.ManageGroups` permission. [elioschmutz]
 - Bump docxcompose to 1.3.0 to support updating complex properties with no existing value. [deiferni]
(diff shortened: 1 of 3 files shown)


docs/HISTORY.txt
```diff
@@ -4,6 +4,7 @@ Changelog
 2020.11.0 (unreleased)
 ----------------------
 
+- GET @groups endpiont is now available with the `opengever.api.ManageGroups` permission. [elioschmutz]
```
njohner, line 7: endpiont -> endpoint
You have the same typo in the PR title and in the commit message

njohner (approved):
Apart from comment, LGTM

PR #6666, opened 2020-10-07
Add external_reference field to solr, reindex objects with values.

deiferni:
With this PR we add a field `external_reference` also to Solr. The field can have two sources, it can come from documents or dossiers.

The field `IDocumentMetadata.foreign_reference` is also indexed into the index external_reference:
https://github.com/4teamwork/opengever.core/blob/fe821de8281b88dd16276158d4d7ec7029c59a96/opengever/document/indexers.py#L119-L123
We include it here to reindex consistently and the same way as the field is indexed in the catalog. This will lead to somewhat inconsistent naming of `foreign_reference` v.s. `external_reference`.

I've quickly discussed it with @lukasgraf and we'e decided to leave it as is for now, with the following way out:
- should `foreign_reference` be used, we can reindex documents with a `foreign_reference` into an additional Solr field, also called `foreign_reference`
- we can then re-index affected documents via the `external_reference` catalog index
- we leave the catalog index in place, for potential future migrations
- after those migrations took place we can get rid of the `external_reference` catalog index and keep data in Solr in `external_reference` and `foreign_reference`.

Only objects with an actual value are reindexed, only a mininal amount of objects in production seem to have a value. Nevertheless, the upgrade is `deferrable` as it is not strictly required.

Documentation on searching 

Jira: https://4teamwork.atlassian.net/browse/CA-1062

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 ----------------------
 
 - Allow assigning groups as participants to a Teamraum [elioschmutz]
+- Add external_reference field to solr, reindex objects with values. [deiferni]
 
 
 2020.11.0 (2020-10-07)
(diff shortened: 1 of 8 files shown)


lukasgraf (approved):
LGTM 👍

PR #6667, opened 2020-10-08
Correct main-dossier expansion tests.

njohner:
The correct parameter name is "expand" and not "expansion" for requiring expansions over the REST-API. Tests were not failing because the main-dossier expansion gets expanded by default.

This is not for any jira issue, and I think does not deserve a changelog entry.

## Checklist (Must have)

- [ ] Changelog entry

### DIFF
--- a/opengever/api/tests/test_dossier.py
+++ b/opengever/api/tests/test_dossier.py
@@ -91,7 +91,7 @@ class TestMainDossierExpansion(IntegrationTestCase):
     def test_main_dossier_expansion_on_repository_root(self, browser):
         self.login(self.regular_user, browser=browser)
         browser.open(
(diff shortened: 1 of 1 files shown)


deiferni (approved):
LGTM 👍 

there is actually an issue as `main-dossier` expansion always expands, see https://4teamwork.atlassian.net/browse/CA-1010. that's why the tests passed 🙈.

njohner:
Ah I thought it was on purpose. But Yes it's defaulting to expanding, so it always expands. I could quickly correct it in that case, instead of just fixing the tests

deiferni:
you will also need to implement it properly with a separate service endpoint for the expandable component, good examples are available.

njohner:
Ok right. In that case we'd fix the implementation separately. I'll simply merge this one.

PR #6668, opened 2020-10-09
Also provide main_dossier for dossiertemplates

elioschmutz:
This PR extends the `get_main_dossier` utility function to also provide it for dossiertemplates.

Jira: https://4teamwork.atlassian.net/browse/NE-57

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.12.0 (unreleased)
 ----------------------
 
+- Also provide main_dossier for dossiertemplates [elioschmutz]
 - Allow assigning groups as participants to a Teamraum [elioschmutz]
 - Add external_reference field to solr, reindex objects with values. [deiferni]
 
--- a/opengever/api/tests/test_dossier.py
+++ b/opengever/api/tests/test_dossier.py
@@ -242,3 +242,66 @@ class TestMainDossierExpansion(IntegrationTestCase):
             },
             browser.json["@components"]['main-dossier'],
         )
+
+    @browsing
+    def test_main_dossier_expansion_on_dossiertemplate(self, browser):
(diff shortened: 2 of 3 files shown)


njohner:
See https://github.com/4teamwork/opengever.core/pull/6667

deiferni (commented):
The `main-dossier` expansion is currently not implemented correctly, as we touch it here i'd propose to fix it. See https://4teamwork.atlassian.net/browse/CA-1010.

elioschmutz:
@deiferni no, I think we should do this through `ca-1010`. I don't do anything with the expansion . I just extend the current implementation of the maindossier.

deiferni:
yes we could, of course. but we're touching the expansion and this would be a very good opportunity to fix it. for the record i'm not a fan of always deferring issues and i'm mentioning it because you might not have been aware of the other issue.

elioschmutz:
@deiferni Even if the expansion is broken, we hardcoded to always expand the `main-dossier` in the frontend. We would not need to fix anything in the frontend if fixing: `ca-1010`. But I prefer to remove the hardocded `main-dossier` expansion in the frontend and only expand it if necessary. This is what we want in the future anyway. So we need to do a little more in the frontend.

But again, I would not solve this through this PR because I don't do anything with expansions in the frontend nor in the backen. Fixing it now through this PR would just block my story.

But I can offer you to overtake `ca-1010` and fix it with a separate PR in both, the backend and in the frontend in this sprint.

deiferni (approved):
ok 👍 

nahh i'ts ok in that case we can leave it separate, i forgot you always expand in any case 😉. this might also remove some of the urgency i felt.

PR #6669, opened 2020-10-09
Add and index PHVS specific fields in solr.

njohner:
As listings have been switched to always using solr, we now need the PHVS specific fields to be indexed in solr in order to be able to display them in the listings. We therefore add the fields to the solr `managed-schema`. The upgrade step will only reindex object on the PHVS deployment as it queries for the interface supporting the PHVS specific fields.

I have tested this locally with the PHVS policy for the listings in the classic UI and made sure that indexing also works properly when modifying the fields (edit form in the classic and new UI).

For https://4teamwork.atlassian.net/browse/CA-950

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.12.0 (unreleased)
 ----------------------
 
+- Add and index PHVS specific fields in solr. [njohner]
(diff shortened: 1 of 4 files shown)


lukasgraf (approved):
LGTM 👍

PR #6673, opened 2020-10-13
Add basic support for graceful operation in readonly mode

lukasgraf:
This changeset adds basic support for graceful operation in readonly mode.

- Testing support
- A helper function to determine whether ReadOnly mode is enabled
- `is_readonly` flag in `@config` endpoint and the `@@gever_state` view
- Disabling GZip compression in `plone.app.caching` because it interferes with error message rendering and has been removed in future versions anyway.
- A basic viewlet at the top of the page that indicates ReadOnly mode if it's enabled
- A custom error page for `ReadOnlyError`s

Short of avoiding unexpected writes, these are the bare necessities to facilitate graceful operation of GEVER when an instance is in readonly mode.

### ReadOnly Viewlet

readonly_viewlet

_([Example with visible teaser](https://user-images.githubusercontent.com/405124/95836419-b1adf300-0d3f-11eb-81d5-6b02efea44ba.png))_

---

### ReadOnly Error Page

readonly_error

---

Jira: https://4teamwork.atlassian.net/browse/CA-3

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

    _(Trivial registry update)_
- New translations


opengever/readonly/utils.py
```diff
@@ -0,0 +1,10 @@
+from plone.memoize.view import memoize_contextless
+from zope.component.hooks import getSite
+
+
+def is_in_readonly_mode():
```
deiferni, line 5: this is one of the main "api" functions of `opengever.readonly`, right? i would therefore not put it in `utils` but instead in `__init__` so it can be imported like `from opengever.readonly import is_in_readonly_mode`. To me this would clarify that this is the one function to be used by other packages, whereas in `utils` i usually expect module-internal helpers and utilities. I'm aware that we don't do that consistently though, so up to you to disregard this comment 😉 .

opengever/api/tests/test_config.py
```diff
@@ -137,6 +138,19 @@ def test_config_contains_portal_url(self, browser):
         self.assertEqual(
             browser.json.get(u'portal_url'), 'http://nohost/portal')
 
+    @browsing
+    def test_config_contains_is_readonly_flag(self, browser):
+        self.login(self.regular_user, browser)
+
+        browser.open(self.config_url, headers=self.api_headers)
+        self.assertEqual(browser.status_code, 200)
+        self.assertEqual(browser.json.get(u'is_readonly'), False)
```
deiferni, line 147: `assertFalse`?

opengever/api/tests/test_config.py
```diff
@@ -137,6 +138,19 @@ def test_config_contains_portal_url(self, browser):
...
+
+        with ZODBStorageInReadonlyMode():
+            browser.open(self.config_url, headers=self.api_headers)
+            self.assertEqual(browser.status_code, 200)
+            self.assertEqual(browser.json.get(u'is_readonly'), True)
```
deiferni, line 152: `assertTrue`?

opengever/base/browser/errors.py
```diff
@@ -28,6 +28,7 @@ def __call__(self):
 
         exception = self.context
         self.error_type = type(exception).__name__
+        self.request.response.setHeader('Content-Type', 'text/html')
```
deiferni, line 31: tracebacks for managers are also rendered in an html response and not returned in "plain text", right?

opengever/base/browser/templates/error.pt
```diff
@@ -78,12 +78,38 @@
 
         <metal:othererror tal:condition="python:view.error_type!='NotFound'"
                           tal:define="is_manager view/is_manager;
+                                      is_readonly_error python:view.error_type=='ReadOnlyError';
```
deiferni, line 81: i'd prefer this logic to be moved to the view e.g. `view/is_readonly_error`.

opengever/base/locales/de/LC_MESSAGES/opengever.base.po
```diff
@@ -112,6 +112,11 @@ msgstr "Wenn Sie diesen Fehler dem Fachapplikationsverantwortlichen melden wolle
 msgid "description_no_privileges_for_page"
 msgstr "Sie haben unzureichende Berechtigungen um diese Seite anzuzeigen. Wenn Sie meinen diese Meldung fehlerhafterweise angezeigt zu bekommen, kontaktieren Sie bitte die fachapplikationsverantwortliche Person."
 
+#. Default: "Please notify your administrator about this error, and include a description the action you tried to perform as well as the error number below."
+#: ./opengever/base/browser/templates/error.pt
+msgid "description_please_notify_admin"
+msgstr "Bitte melden sie diesen Fehler dem Fachapplikationsverantwortlichen unter Angabe der Aktion welche Sie durchzuführen versucht haben, und der untenstehenden Fehlernummer."
```
deiferni, line 118: durchzuführen versucht haben -> durchgeführt haben? the first one doesn't sound right in my ears, even though it may be correct 😉.

deiferni (approved):
LGTM 🎉 , nice clean and reviewable commits!

Unfortunately that also means i had the focus for some nagging 😉. But really only very minor things and suggestions/comments, nothing that would block a merge if you decide against changing it.

opengever/testing/patch.py
```diff
@@ -74,3 +74,119 @@ def unindex(self, obj):
...
+
+    # With these changes DemoStorage is ready to answer truthfully in its
+    # isReadOnly() method. What follows now is actually suppressing writes
+    # by raising POSException.ReadOnlyError in the right places.
+    #
+    # This is done in a way that emulates the code in ZEO.ClientStorage as
+    # closely as possible (including the indirection through _check_trans()),
+    # in order for the two implementations to be easier to compare.
+
+    # IStorage.tpc_begin
```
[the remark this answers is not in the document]
lukasgraf: Yes, these are mainy there to visually structure the code, and separate the individual patches from each other. Also also declare with interface / implementation defines the patched method.

Prefixing it with `patch ...` seems like unnecessary noise, akin to `i += 1  # increment i by one`. Is it really that confusing?

What about `# === IStorage.new_oid`?

opengever/testing/tests/test_readonly_demostorage.py
```diff
@@ -0,0 +1,139 @@
...
+    def test_context_manager_resets_mode_on_exit(self):
+        with ZODBStorageInReadonlyMode():
+            pass
+
+        conn = self.portal._p_jar
+        storage = conn.db().storage
+
+        self.assertFalse(getattr(storage, '_is_read_only', False))
+        self.assertFalse(storage.isReadOnly())
+        self.assertFalse(storage.isReadOnly())
```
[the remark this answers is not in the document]
lukasgraf: Indeed. 👍

opengever/readonly/utils.py
lukasgraf: Yep, I was going back and forth on this. Don't like the utils either. I'll move it into `__init__.py` 👍

opengever/base/browser/errors.py
lukasgraf: Yey, they are. And the REST API serializes the `ReadOnlyError` just like any other exception, with the proper response Content-Type of `application/json` (doesn't involve this view). I'll actually add a test for that.

opengever/base/locales/de/LC_MESSAGES/opengever.base.po
lukasgraf: Err.. It sounds fine to me? 😉 
"Durchgeführt haben" sort of implies that the action succeeded. I prefer making it clear that it actually wasn't performed.

PR #6674, opened 2020-10-13
Add @dossier-from-template endpoint

tinagerber:
Changes in this PR:

- Add `folder_factories` to `portal_actions`. This way the folder_factories appear by default when the @actions endpoint is requested and we don't have to define, which categories are needed. With this information we can then find out where the dossiers, documents and tasks can be created from template.
- Implement `opengever.dossier.DossierTemplatesVocabulary`.
- Move some functions to a helper class to make them reusable. I'm not sure if it's pythonic to have a mixin class or if I copied that from frontend development. 🤔
- Add `@dossier-from-template` endpoint. The endpoint inherits from the class used to create content using POST requests to create the dossier. The remaining content is then created using the methods I moved to the mixin helper class.

Jira: https://4teamwork.atlassian.net/browse/NE-92

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:

### DIFF
--- a/docs/public/dev-manual/api/dossier_from_template.rst
+++ b/docs/public/dev-manual/api/dossier_from_template.rst
@@ -1,0 +1,33 @@
+.. _dossier_from_template:
+
+Geschäftsdossier ab Vorlage erstellen
+=====================================
+
+In einer Ordnungsposition kann über den Endpoint ``@dossier-from-template`` ein neues
+Geschäftsdossier ab Vorlage erstellt werden.
+
+Der Endpoint erwartet mindestens folgende Parameter:
+
+- ``template``: Das zu verwendende Template aus dem Vokabular ``opengever.dossier.DossierTemplatesVocabulary``
+- ``responsible``: Der Federführende des zu erstellenden Dossiers
+- ``title``: Der Titel des zu erstellenden Dossiers
+
+Zusätzlich können auch die anderen Felder des :ref:`label-dossier-schema`-Schemas gesetzt werden.
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+       POST /ordnungssystem/fuehrung/@dossier-from-template HTTP/1.1
+       Accept: application/json
+
+       {
+        "template": {"token": "1234567890"},
+        "description": "Dossier description"
+        "responsible": "rolf.ziegler",
+        "title": "Dossier title"
+       }
+
+
+Als Response wird die JSON-Repräsentation des neu erstellten Dossiers geliefert,
+siehe :ref:`Inhaltstypen <content-types>`.
--- a/opengever/api/templatefolder.py
+++ b/opengever/api/templatefolder.py
@@ -44,7 +44,7 @@ class DocumentFromTemplatePost(Service):
             raise BadRequest('Missing parameter template')
 
         vocabulary = getUtility(IVocabularyFactory,
-                             name='opengever.dossier.DocumentTemplatesVocabulary')
+                                name='opengever.dossier.DocumentTemplatesVocabulary')
         term = vocabulary(self.context).getTermByToken(token)
         template = term.value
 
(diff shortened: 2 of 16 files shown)


opengever/dossier/dossiertemplate/form.py
```diff
@@ -68,6 +70,15 @@ def get_dossier_templates(context):
     return SimpleVocabulary(terms)
 
 
+@implementer(IVocabularyFactory)
+class DossierTemplatesVocabulary(object):
+
+    def __call__(self, context):
+        if not IRestrictAddableDossierTemplates.providedBy(context):
+            return SimpleVocabulary([])
```
deiferni, line 78: I'm missing a test in this commit, you should test against possible permutations of `addable_dossier_templates` and potentially also `allow_add_businesscase_dossier`.

opengever/core/profiles/default/actions.xml
```diff
@@ -1,5 +1,7 @@
 <object xmlns:i18n="http://xml.zope.org/namespaces/i18n" name="portal_actions" meta_type="Plone Actions Tool">
 
+  <object name="folder_factories" meta_type="CMF Action Category" />
```
deiferni, line 3: doesn't this break any existing tests? can we test this?

docs/public/dev-manual/api/dossier_from_template.rst
```diff
@@ -0,0 +1,33 @@
+.. _dossier_from_template:
+
+Geschäftsdossier ab Vorlage erstellen
+=====================================
+
+In einem Ordnungsposition kann über den Endpoint ``@dossier-from-template`` ein neues
```
deiferni, line 6: einem -> einer

opengever/api/templatefolder.py
```diff
@@ -60,6 +65,45 @@ def reply(self):
...
+        self.title_ = data.get('title', None)
+
+        token = data.get('template')
+        if isinstance(token, dict):
+            token = token.get('token')
+        if not token:
+            raise BadRequest('Missing parameter template')
+        vocabulary = getUtility(IVocabularyFactory,
+                                name='opengever.dossier.DossierTemplatesVocabulary')
+        self.dossier_template = vocabulary(self.context).getTermByToken(token).value
```
deiferni, line 84: this can fail with a `500` internal server error for invalid tokens. please implement graceful validation that returns a `BadRequest`.

opengever/api/templatefolder.py
```diff
@@ -60,6 +65,45 @@ def reply(self):
...
+        return data
+
+    def reply(self):
+        if not self.context.restrictedTraverse('dossier_with_template/is_available')():
```
deiferni, line 88: IMO api services should not reuse views like that if it is avoidable. the view will go away eventually. please extract the visibility check so you can directly call it as a method here.

opengever/api/tests/test_templatefolder.py
```diff
@@ -153,6 +158,126 @@ def test_requires_add_permission(self, browser):
...
 
 
+class TestDossierFromTemplatePost(IntegrationTestCase):
+
+    features = ('dossiertemplate', )
+
+    @browsing
+    def test_create_dossier_from_template_within_repository_folder(self, browser):
+        self.login(self.regular_user, browser)
+        browser.open(
```
deiferni, line 168: i think instead of building the url youself you could also let testbrowser do it, by:
```
browser.open(self.leaf_repofolder, view='@vocabularies/opengever.dossier.DossierTemplatesVocabulary', …)
```

deiferni (changes requested):
LGTM apart from comments. I'd like to have another look when it is done, if you have questions don't hesitate to ☎️ !

opengever/core/profiles/default/actions.xml
tinagerber: Do you mean in gever-ui?

deiferni: no in the backed, items returned by `@actions` should differ, right?

tinagerber: Yes, but the endpoint is implemented in plone.restapi, so we have no tests for it.

tinagerber:
@deiferni I have implemented the desired changes, can you rereview?

deiferni (approved):
thanks 👍

PR #6675, opened 2020-10-13
Add @possible-participants endpoint.

tinagerber:
In this PR a new API endpoint `@possible-participants` is implemented. The endpoint is used to decide for which users/contacts/mailboxes participations can be added.

Jira: https://4teamwork.atlassian.net/browse/NE-90

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -10,6 +10,7 @@ Changelog
 - Add is_readonly flag for @config endpoint and @@gever_state view. [lgraf]
 - Add @dossier-from-template endpoint. [tinagerber]
 - Activate the groups plugin for source_groups. [elioschmutz]
+- Add @possible-participants endpoint. [tinagerber]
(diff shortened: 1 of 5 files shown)


njohner (changes requested):
Can you add tests for SQL participations as well?
Note that we return all possible participations, even the ones that will fail when added because there is already a participation for that participant... this is not very user-friendly, but it's the same as in the classic UI. The frontend will therefore need to deal with failing requests to add participations, or we need to update the vocabularies so that they only actually give back the allowed participants.

Also note that this will conflict with https://github.com/4teamwork/opengever.core/pull/6655, and I think also leads to three different places where we serialise participants in 3 different ways... depending on what happens with the refactoring PR, I think we would need to look at this endpoint again, and maybe use the handlers here, instead of adding another conditional

But LGTM

njohner (approved):
LGTM, as soon as tests are green ;-)

PR #6678, opened 2020-10-15
Update bootstrap.py in policy template.

tinagerber:
Since the bootstrap.py has been updated, it should also be updated in the policy template.
It was already updated in August, so there are no major changes.
Jira: https://4teamwork.atlassian.net/browse/NE-113

## Checklist (Must have)

- [ ] Changelog entry

### DIFF
--- a/opengever/policytemplates/policy_template/opengever.+package.name+/bootstrap.py
+++ b/opengever/policytemplates/policy_template/opengever.+package.name+/bootstrap.py
@@ -49,8 +49,8 @@ parser.add_option("--version",
 parser.add_option("-t", "--accept-buildout-test-releases",
                   dest='accept_buildout_test_releases',
                   action="store_true", default=False,
-                  help=("Normally, if you do not specify a --buildout-version, "
-                        "the bootstrap script and buildout gets the newest "
+                  help=("Normally, if you do not specify a --version, the "
+                        "bootstrap script and buildout gets the newest "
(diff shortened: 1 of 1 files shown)


opengever/policytemplates/policy_template/opengever.+package.name+/bootstrap.py
```diff
@@ -49,8 +49,8 @@
 parser.add_option("-t", "--accept-buildout-test-releases",
                   dest='accept_buildout_test_releases',
                   action="store_true", default=False,
-                  help=("Normally, if you do not specify a --buildout-version, "
-                        "the bootstrap script and buildout gets the newest "
+                  help=("Normally, if you do not specify a --version, the "
```
njohner, line 52: Hum that's strange, IMO we are updating a perfectly correct help message with one that is wrong??

njohner: I mean The --version flag prints the `bootstrap.py` version and then exits, where as the `--buildout-version` flag actually sets the buildout version.

tinagerber: 🤷‍♀️ Yes, this is weird, but this is the bootstrap.py we use everywhere now.

njohner: 😓

njohner (approved):

PR #6679, opened 2020-10-15
Use ParticipationHandler in possible participants endpoint.

njohner:
During the review of the refactoring of the participations (https://github.com/4teamwork/opengever.core/pull/6655) a new endpoint for possible participants was added (https://github.com/4teamwork/opengever.core/pull/6675). This is a follow up PR to also use the Participation handlers in that endpoint.

For https://4teamwork.atlassian.net/browse/NE-90 or maybe https://4teamwork.atlassian.net/browse/CA-17

## Checklist (Must have)

- [ ] Changelog entry: No changelog entry necessary IMO


lukasgraf (approved):
LGTM 👍

PR #6680, opened 2020-10-15
Extend the @listing endpoint to list `folder_contents`

elioschmutz:
This PR:

- adds a new listing to list all contents within a context
- adds a new listing-field `creator_fullname`

We need to list contents of a workspaces. Currently `contents` are workspace folders, documents and mails. A generic `folder_contents` listing lets the frontend decide, which types should appear in the `contents` tab.

Jira: https://4teamwork.atlassian.net/browse/NE-124

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -8,6 +8,7 @@ Changelog
 - Patch several login-related events to allow login during readonly mode. [lgraf]
 - Add optional support for WriteOnRead tracing in ReadOnlyError page. [lgraf]
 - Add videoconferencing URL to workspaces. [deiferni]
+- Add a new listing: `folder_contents` to the @listing endpoint. [elioschmutz]
 - Use custom error page for ReadOnlyErrors. [lgraf]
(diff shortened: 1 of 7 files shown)


njohner (commented):
Is that still a listing? when it's returning all types mashed together? shouldn't we simply use solr-search for that?

elioschmutz:
> Is that still a listing? when it's returning all types mashed together? shouldn't we simply use solr-search for that?

Why not? We use it like every other listing. It lists content like every other listing. So I think it's still a listing.

Using solr-search would of course be possible. But why don't we use solr-serach for all the other listings?

The frontend uses the `@listing` endpoint for listings. So if I would use the solr-search I would need to reimplement the listing  components for the `@solrsearch`. That's not what we want.

njohner (approved):
I'd change the last commit message "PEP8: fix linebreak". or simply "PEP8".

PR #6683, opened 2020-10-16
Allow login during readonly mode

lukasgraf:
This is a series of patches that facilitates successful login of users while read-only mode is active.

The aim is to allow login for **users that have already logged in once** to the system previously (in read-write mode). Users logging in for their **very first time** during read-only mode is out of scope, and is not guaranteed to work.

Patches the following methods / functions (_mostly with no-op replacements_):
- [`MembershipTool.setLoginTimes()`](https://github.com/plone/Products.PlonePAS/blob/5.1.0/src/Products/PlonePAS/tools/membership.py#L730-L748)
- [`plone.app.contentrules.handlers.user_logged_in()`](https://github.com/plone/plone.app.contentrules/blob/3.0.9/plone/app/contentrules/handlers.py#L271-L274)
- [`plone.protect.subscribers.onUserLogsIn()`](https://github.com/plone/plone.protect/blob/3.1.2/plone/protect/subscribers.py#L44-L67)
- [`MembershipTool.createMemberArea()`](https://github.com/plone/Products.PlonePAS/blob/5.1.0/src/Products/PlonePAS/tools/membership.py#L293-L371)

## Checklist (Must have)

- [x] Changelog entry


opengever/base/monkey/patches/readonly.py
```diff
@@ -0,0 +1,25 @@
...
+from opengever.readonly import is_in_readonly_mode
+from Products.PlonePAS.tools.membership import MembershipTool
+
+
+class PatchMembershipToolSetLoginTimes(MonkeyPatch):
+    """In read-only mode, allow login of existing users without updating
+    their last login times (which would cause a write).
+
+    The return value of False signals that this is not the user's very
+    first login.
```
njohner, line 18: Ok I was a bit confused here because we would also return False if it is the users first login but we are in read_only modus.

njohner (commented):
So far so good

opengever/base/monkey/patches/readonly.py
lukasgraf: Yes, I decided to signal "not initial login" here in order to avoid having to patch even more things that only happen on _initial_ login.

njohner (approved):
LGTM

PR #6685, opened 2020-10-19
Fix WOPI validator tests

buchi:
When running on Jenkins we don't have an interactive terminal.

## Checklist (Must have)

- [ ] Changelog entry


lukasgraf (approved):
LGTM 👍 , thanks for the quick fix.

PR #6686, opened 2020-10-19
Create and persist Bumblebee user salt on login.

lukasgraf:
Create and persist the user's bumblebee salt during login, if one doesn't exist yet.

This will make sure it's available when needed, and doesn't have to be calculated and persisted as part of an ad-hoc write-on-read. Since several writes already happen during login _(when the connection is in read-write mode)_, and the login request is whitelisted by `plone.protect` itself, this shouldn't have any adverse effects.

By doing this we can ensure that users that have already logged in at least once have their salt ready, and will be able to see Bumblebee previews in read-only mode.

_(On a login during read-only mode this needs to be suppressed - the whole idea is that we won't have to generate any salts while in read-only mode)_

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.12.0 (unreleased)
 ----------------------
 
+- Create Bumblebee user salt on login. [lgraf]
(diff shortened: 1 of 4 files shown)


njohner (approved):
LGTM

PR #6687, opened 2020-10-19
Docker services

buchi:
Adds support for using web-based services instead of locally installed executables for `msgconvert`, `sablon` and `pdflatex`.

This is a first step towards running `opengever.core` with Docker and turning into a service oriented architecture.

The services are provided as Docker images on [Docker Hub](https://hub.docker.com/search?q=4teamwork&type=image&sort=updated_at&order=desc):
- [msgconvert](https://hub.docker.com/r/4teamwork/msgconvert)
- [pdflatex](https://hub.docker.com/r/4teamwork/pdflatex)
- [sablon](https://hub.docker.com/r/4teamwork/sablon)

To use a specific service, its service URL has to be defined in an environment variable. It not defined, it falls back to use the executable. For local development the service URLs are defines as follows:

```
MSGCONVERT_URL=http://localhost:8090/
SABLON_URL=http://localhost:8091/
PDFLATEX_URL=http://localhost:8092/
```

This means that from now on, the services are used for local development by default. They can be started by executing `docker-compose up`. See also the [README](https://github.com/4teamwork/opengever.core/pull/6687/commits/af6e2a64c35b4a189eaafd04db28b8166211d790#diff-7b3ed02bc73dc06b7db906cf97aa91dec2b2eb21f2d92bc5caa761df5bbc168fR315)

Tests on CI currently default to use the executables as this is still what we use in production. However there are testing layers  for each service that provide fixtures for testing using the Docker services. Once we roll out the services in production, the default can be easily switched by adding these layers to the bases of the integration testing layer.

Jira: https://4teamwork.atlassian.net/browse/CA-1148
Architecture: https://4teamwork.atlassian.net/wiki/spaces/4TEAM/pages/295534663/GEVER+Zielarchitektur

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/development.cfg
+++ b/development.cfg
@@ -39,6 +39,9 @@ parts +=
     ${buildout:code-audit-parts}
     ${buildout:i18n-parts}
 
+parts -=
+    gems
+
 # example mysql configuration
 #ogds-db-name = opengever
 #ogds-db-user = opengever
@@ -56,6 +59,8 @@ environment-vars +=
     BUMBLEBEE_INTERNAL_PLONE_URL http://localhost:${instance:http-address}/fd
     BUMBLEBEE_PUBLIC_URL http://localhost:3000/
     TEAMRAUM_URL http://localhost:8080/fd
+    MSGCONVERT_URL http://localhost:8090/
+    SABLON_URL http://localhost:8091/
 
 zcml +=
   opengever.core
@@ -65,7 +70,8 @@ cores = ${solr:gever-cores}
 
 [test]
 initialization +=
-    os.environ['SABLON_BIN'] = '${buildout:sablon-executable}'
+    os.environ['MSGCONVERT_URL'] = 'http://localhost:8090/'
+    os.environ['SABLON_URL'] = 'http://localhost:8091/'
 
 [docxcompose]
 recipe = zc.recipe.egg:script
--- a/development.cfg
+++ b/development.cfg
@@ -61,6 +61,7 @@ environment-vars +=
     TEAMRAUM_URL http://localhost:8080/fd
     MSGCONVERT_URL http://localhost:8090/
     SABLON_URL http://localhost:8091/
+    PDFLATEX_URL http://localhost:8092/
 
 zcml +=
   opengever.core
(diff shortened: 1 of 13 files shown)


development.cfg
```diff
@@ -65,7 +71,9 @@ cores = ${solr:gever-cores}
 
 [test]
 initialization +=
-    os.environ['SABLON_BIN'] = '${buildout:sablon-executable}'
+    os.environ['MSGCONVER_URL'] = 'http://localhost:8090/'
```
jone, line 74: Typo

buchi: ✅

opengever/testing/docker.py
```diff
@@ -0,0 +1,87 @@
...
+        args = shlex.split(cmd)
+        proc = subprocess.Popen(
+            args, stdin=subprocess.PIPE, stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+        out, err = proc.communicate()
+        if proc.returncode != 0:
+            assert False, "Running '{}' failed. Command returned: {}".format(cmd, err)
+
+    def find_free_port(self):
```
lukasgraf, line 58: This is potentially subject to a Time-of-Check-to-Time-of-Use race condition.

The critical time window should be quite short, except for the first time a new docker image needs to be pulled. But since this should only come into play if no port picked by the Jenkins port allocator plugin is provided via `port_env`, and there usually should be tens of thousands of free ports, I don't see this being a realistic problem that would ever occur, so 👍

lukasgraf (approved):
Nice! 👍  🎉 

I tested it locally, and it does exactly what it says on the tin 😉 

Using it was straightforward and worked out of the box, as described in the README. I had a closer look at the following things:

- All services properly log the important things on the console, so it ends up in docker logs (HTTP requests as well as any output from the service)
- Fallback to direct invocation of executables works (if URLs not configured)
- Error messages during local development if services are not running (i.e. forgetting to `docker-compose up`) are pretty clear
- Error handling / logging works well:
  - Logging by services themselves is detailed and useful
  - Error logging on the side of the Plone instance is also very clear, and even contains a truncated copy of the service's error text with the root cause
  - In the case of the `pdflatex` service, the error handling in fact is substantially better than just the `PDFBuildFailed` from before 😉 
- Happy path for all three services worked without any problems

PR #6688, opened 2020-10-19
Implement sort group for the `@listing` endpoint

elioschmutz:
This PR implements a new query parameter for the `@listing` endpoint: `sort_first`.

This parameter can be used to order the listing results into two groups. We use this in combination with the `folder_contents` listing to order folderish contents first.

Bildschirmfoto 2020-10-19 um 16 12 42

### Performance
The function query executes in an average time of 200ms on a system with about one million contents. The tested query was:

`if(or(termfreq(portal_type, 'opengever.dossier.businesscasedossier'), termfreq(portal_type, 'opengever.document.document')),1, 0) desc` 

Jira: https://4teamwork.atlassian.net/browse/NE-124

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/docs/public/dev-manual/api/listings.rst
+++ b/docs/public/dev-manual/api/listings.rst
@@ -240,6 +240,7 @@ Optionale Parameter:
 - ``b_size``: Die maximale Anzahl der zurückzugebenden Elemente
 - ``sort_on``: Sortierung nach einem indexierten Feld
 - ``sort_order``: Sortierreihenfolge: ``ascending`` (aufsteigend) oder ``descending`` (absteigend)
+- ``sort_first``: Sortiert die Resultate in zwei Gruppen. Jede Gruppe wird anschließend gem. ``sort_on`` sortiert.
 - ``search``: Filterung nach einem beliebigen Suchbegriff
 - ``columns``: Liste der Felder, die zurückgegeben werden sollen.
 - ``filters``: Einschränkung nach einem bestimmten Wert eines Feldes
@@ -279,6 +280,31 @@ Optionale Parameter:
   .. sourcecode:: http
 
     GET /ordnungssystem/fuehrung/dossier-23/@listing?name=documents&facets:list=creator HTTP/1.1
+    Accept: application/json
+
+
+Bestimmte Inhalte zuerst sortieren:
+-----------------------------------
+Die Resultate können in zwei Gruppen aufgeteilt und anschließend sortiert werden. So können z.B. in einer Auflistung alle Ordner zuoberst angezeigt werden.
+
+Alle Inhalte welche zuerst angezeigt werden sollen bilden eine Gruppe, alle restlichen Inhalte bilden eine zweite Gruppe. Momentan werden nur folgende Felder als `sort_first` unterstütz:
+
+- ``portal_type``
+
+
+**Beispiel: Alle Dossiers zuoberst. Jede Gruppe wird nach Titel sortiert**
+
+  .. sourcecode:: http
+
+    GET /@listing?sort_first.portal_type:record:list=opengever.dossier.businesscasedossier&sort_on=sortable_title HTTP/1.1
+    Accept: application/json
+
+
+**Beispiel: Alle Dokumente und Mails zuoberst**
+
+  .. sourcecode:: http
+
+    GET /@listing?sort_first.portal_type:record:list=opengever.document.document&sort_first.portal_type:record:list=ftw.mail.mail& HTTP/1.1
     Accept: application/json
 
 
--- a/opengever/api/listing.py
+++ b/opengever/api/listing.py
@@ -50,6 +50,9 @@ OTHER_ALLOWED_FIELDS = set([
     'watchers',
 ])
 
+ALLOWED_ORDER_GROUP_FIELDS = set([
+    'portal_type'
+])
 
 FILTERS = {
     u'dossiers': [
@@ -200,7 +203,86 @@ class ListingGet(SolrQueryBaseService):
                 sort += ' desc'
             else:
                 sort += ' asc'
-        return sort
+
+        return self.extend_with_sort_first(sort, params)
+
+    def extend_with_sort_first(self, sort, params):
+        """The sort_first sorts docs in two steps. First, it will sort
+        all items having one of the provided values of a defined field first.
+        All other docs will be put into the second group. Each group will then
+        be sorted by the `sort_on`.
+
+        Having a result of the following docs ordered by sortable_title:
+
+        { 'portal_type': 'type1', title: 'A' }
+        { 'portal_type': 'type2', title: 'B' }
+        { 'portal_type': 'type2', title: 'C' }
+        { 'portal_type': 'type3', title: 'D' }
+        { 'portal_type': 'type1', title: 'E' }
+        { 'portal_type': 'type4', title: 'F' }
+
+        Can be grouped i.e. by type:
+
+        { 'portal_type': ['type1', type3'] }
+
+        which will result in the following sort order:
+
+        { 'portal_type': 'type1', title: 'A' } <= Group 1
+        { 'portal_type': 'type3', title: 'D' }
+        { 'portal_type': 'type1', title: 'E' }
+        { 'portal_type': 'type2', title: 'B' } <= Group 2
+        { 'portal_type': 'type2', title: 'C' }
+        { 'portal_type': 'type4', title: 'F' }
+        """
+        sort_first = params.get('sort_first', {})
+
+        if not sort_first:
+            return sort
+
+        if not isinstance(sort_first, record):
+            raise BadRequest('The sort_first parameter needs to be a record.')
+
+        if len(sort_first) > 1:
+            raise BadRequest('Exactly one sort_first field is required.')
+
+        # Currently, only one sort_first field is implemented. Thus, we just
+        # extract the first item and use this for further processing.
+        field_name, field_values = sort_first.items()[0]
+
+        if not field_values:
+            return sort
+
+        if isinstance(field_values, str):
+            field_values = [field_values]
+
+        if field_name not in ALLOWED_ORDER_GROUP_FIELDS:
+            raise BadRequest(
+                'Sort first field {} is not allowed. Allowed fields are: '
+                '{}.'.format(field_name,
+                             ','.join(ALLOWED_ORDER_GROUP_FIELDS)))
+
+        return ','.join([self._build_sort_first_func_string(field_name, field_values), sort])
+
+    @staticmethod
+    def _build_sort_first_func_string(field_name, field_values):
+        """Generates a solr function string to order by a group of values of
+        a specific field.
+
+        We achieve this by giving each solr-document a number of either 1 or 0,
+        depending on the field content and then sort by this number.
+
+        A conditional function query looks like this `if(test, value1, value2)`
+
+        The `termfreq`-function returns the number of times a term appears in
+        the field for each document. It allows to check whether any of
+        the field_values are included in the document's field value.
+        """
+        field_value_conditions = []
+        for field_value in field_values:
+            field_value_conditions.append(
+                'termfreq({}, {})'.format(field_name, field_value))
+
+        return 'if(or({}), 1, 0) desc'.format(','.join(field_value_conditions))
 
     def parse_requested_fields(self, params):
         return params.get('columns', None)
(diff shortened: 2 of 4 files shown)


docs/public/dev-manual/api/listings.rst
```diff
@@ -282,6 +283,31 @@ Optionale Parameter:
...
+Momentan werden nur folgende Felder für Gruppen unterstütz:
+
+- ``portal_type``
+
+
+**Beispiel: Alle Dossiers zuoberst. Jede Gruppe wird nach Titel sortiert **
+
+  .. sourcecode:: http
+
+    GET /@olisting?sort_group:record:list=opengever.dossier.businesscasedossier&sort_on=sortable_title HTTP/1.1
```
njohner, line 299: olisting -> listing

docs/public/dev-manual/api/listings.rst
```diff
@@ -282,6 +283,31 @@ Optionale Parameter:
...
+    Accept: application/json
+
+
+**Beispiel: Alle Dokumente und Mails zuoberst **
+
+  .. sourcecode:: http
+
+    GET /@olisting?sort_group:record:list=opengever.document.document&sort_group:record:list=ftw.mail.mail& HTTP/1.1
```
njohner, line 307: olisting -> listing

njohner (commented):
The code itself looks OK to me, though I did not review in details because I'm not sure we actually need this. Can't I simply sort on `portal_types` and then on `title` for example?

If we need to sort on folderish first, then we could also make a `copyField` of `object_provides` with an analyzer that would yield True for folderish and False for non-folderish objects? Or introduce an `is_folderish` field? Would be much simpler and much more generic. This seems to be a super specific group_by, that only works for 2 groups anyway, so with that single purpose in mind of sorting folderish objects first...

See https://lucene.apache.org/solr/guide/6_6/common-query-parameters.html#CommonQueryParameters-ThesortParameter

elioschmutz:
> The code itself looks OK to me, though I did not review in details because I'm not sure we actually need this. Can't I simply sort on `portal_types` and then on `title` for example?
> 
> If we need to sort on folderish first, then we could also make a `copyField` of `object_provides` with an analyzer that would yield True for folderish and False for non-folderish objects? Or introduce an `is_folderish` field? Would be much simpler and much more generic. This seems to be a super specific group_by, that only works for 2 groups anyway, so with that single purpose in mind of sorting folderish objects first...
> 
> See https://lucene.apache.org/solr/guide/6_6/common-query-parameters.html#CommonQueryParameters-ThesortParameter

Unfortunately, sorting by `portal_types` does not what we need. It would sort all portal-types and then all titles:

type1: A
type1: B
type2: A
type2: B
type3: A
type3: B

We want:

type1: A
type1: B
type2: A
type3: A
type2: B
type3: B

Solving it with a new field `is_folderish` was also one of my thoughts at the beginning. But I decided against it because it's not what we want if we take a closer look to it. `is_folderish` is technical. I.e. a task or a todo-list are both folderish but on the user point of view it's not. Of course, we could introduce a field which contains the value of non technical folderish content types. But through implementing it with `group_by`, we're much more flexible. It's not specific. It's made generic with a very small feature set. I just don't need more features now. But extend it to provide more than 2 groups or more fields is very easy.

With the current implementation the frontend can decide, what types should be grouped, depending on the context. We could even provide a group filter for the user where he can choose the groups by himself.

So I think it's worth to have this new feature.

@njohner I pre-discussed the solution with @buchi. So perhaps @buchi can also give his feedback to your suggestions or concerns.

njohner:
Something is wrong here, it contains the same commits as your PR for the `folder_contents` listing. You're using it as basebranch, but something doesn't work 😅 . I'll look only at the last commit

opengever/api/listing.py
```diff
@@ -200,7 +203,80 @@ def extract_sort(self, params, query):
...
                 sort += ' asc'
-        return sort
+
+        return self.extend_with_sort_group(sort, params)
+
+    def extend_with_sort_group(self, sort, params):
+        """A sort group sorts docs in two steps. First, it will sort
+        all items having one of the provided values of a defined field first.
+        All other docs will be put into the second group. Each group will then
+        be sorted by the `sort_order`.
```
njohner, line 213: `sort_on` not `sort_order`

opengever/api/listing.py
```diff
@@ -200,7 +203,80 @@ def extract_sort(self, params, query):
...
+            sort_group = {}
+
+        if not sort_group:
+            return sort
+
+        if len(sort_group) > 1:
+            raise BadRequest('Exactly one sort_group field is required.')
+
+        # Currently, only one sort_group field is implemented. Thus, we just
+        # extract the first item and use this for the further processing.
```
njohner, line 248: for the further processing. -> for further processing.

opengever/api/listing.py
```diff
@@ -200,7 +203,80 @@ def extract_sort(self, params, query):
...
+    def _build_sort_first_func_string(self, field_name, field_values):
+        """Generates a solr function string to order by a group of values of
+        a specific field.
+
+        We achieve this by giving each solr-document a number of either 1 or 0,
+        depending on the field content and then sort by this number.
+
+        A conditional function query looks like this `if(test, value1, value2)`
+
+        Using the `termfreq`-function allows us to check if a documents field value
```
njohner, line 271: That sentence is wrong. You do the opposite, check that one of the `field_values` is included in the document's field value.
I'd propose something like:
"Using the `termfreq` function allows to handle multiValued fields by checking whether any of the `field_values` are included in the document's field value. "

opengever/api/tests/test_listing.py
```diff
@@ -1328,3 +1330,169 @@ def test_dossier_participant_and_roles_facets(self, browser):
...
         browser.open(self.dossier, view=view, headers=self.api_headers)
         self.assertEqual(0, browser.json['items_total'])
+
+
+class TestListingSortFirst(SolrIntegrationTestCase):
+
+    def test_build_sort_first_func_string_builds_function_string_correctly(self):
+        # A rest-service does not inherit from BrowserView by default.
+        # https://github.com/plone/plone.rest/blob/master/src/plone/rest/zcml.py#L110
+        listing = type(ListingGet.__name__, (ListingGet, BrowserView), {})(self.portal, self.request)
```
njohner, line 1340: Or you could make `_build_sort_first_func_string` a static method, as it does not need `self`, and then you would not need that hack ;-)

opengever/api/listing.py
```diff
@@ -200,7 +203,80 @@ def extract_sort(self, params, query):
...
+        { 'portal_type': 'type1', title: 'A' } <= Group 1
+        { 'portal_type': 'type3', title: 'D' }
+        { 'portal_type': 'type1', title: 'E' }
+        { 'portal_type': 'type2', title: 'B' } <= Group 2
+        { 'portal_type': 'type2', title: 'C' }
+        { 'portal_type': 'type4', title: 'F' }
+        """
+        sort_first = params.get('sort_first', {})
+        if not isinstance(sort_first, record):
+            sort_first = {}
```
njohner, line 239: I'd rather raise a BadRequest here than silently dropping the user input

opengever/api/listing.py
```diff
@@ -200,7 +203,80 @@ def extract_sort(self, params, query):
...
+
+        if len(sort_first) > 1:
+            raise BadRequest('Exactly one sort_first field is required.')
+
+        # Currently, only one sort_first field is implemented. Thus, we just
+        # extract the first item and use this for the further processing.
+        field_name, field_values = sort_first.items()[0]
+
+        if isinstance(field_values, str):
+            field_values = [field_values]
```
njohner, line 252: I'd also check here that field_values is not an empty list. if it is, then return sort.

opengever/api/listing.py
```diff
@@ -200,7 +203,80 @@ def extract_sort(self, params, query):
...
+
+        Using the `termfreq`-function allows us to check if a documents field value
+        is included in the list of `field_values`.
+        """
+        field_value_conditions = []
+        for field_value in field_values:
+            field_value_conditions.append(
+                'termfreq({}, {})'.format(field_name, field_value))
+
+        return 'if(or({}), 1, 0) desc'.format(','.join(field_value_conditions) or 0)
```
njohner, line 279: If you disallow empty list for field_values, then you don't need the default to `0`, which is a bit confusing

opengever/api/tests/test_listing.py
```diff
@@ -1328,3 +1330,169 @@ def test_dossier_participant_and_roles_facets(self, browser):
...
+
+        self.assertEqual(
+            'if(or(0), 1, 0) desc',
```
njohner, line 1343: As I said, I'd drop that default to `0`... instead make sure to test that when `field_values` is an empty list, we simply sort normally on `sort_on`

opengever/api/tests/test_listing.py
```diff
@@ -1328,3 +1330,169 @@ def test_dossier_participant_and_roles_facets(self, browser):
...
+
+        with browser.expect_http_error(reason='Bad Request'):
+            browser.open(self.workspace, view=view, headers=self.api_headers)
+
+        self.assertEqual(
+            u'Group order field not_whitelisted_field is not allowed. Allowed fields are: portal_type.',
+            browser.json.get('message'))
+
+    @browsing
+    def test_sort_first_raises_error_if_using_with_multiple_field_names(self, browser):
```
njohner, line 1376: if_using -> if_used

opengever/api/tests/test_listing.py
```diff
@@ -1328,3 +1330,169 @@ def test_dossier_participant_and_roles_facets(self, browser):
...
+
+        with browser.expect_http_error(reason='Bad Request'):
+            browser.open(self.workspace, view=view, headers=self.api_headers)
+
+        self.assertEqual(
+            u'Exactly one sort_first field is required.',
+            browser.json.get('message'))
+
+    @browsing
+    def test_sort_first_raises_error_if_using_with_no_field_name(self, browser):
```
njohner, line 1394: if_using -> if_used

njohner:
Just a few comments...

opengever/api/listing.py
```diff
@@ -200,7 +203,85 @@ def extract_sort(self, params, query):
...
+
+        if not field_values:
+            return sort
+
+        if isinstance(field_values, str):
+            field_values = [field_values]
+
+        if field_name not in ALLOWED_ORDER_GROUP_FIELDS:
+            raise BadRequest(
+                'Group order field {} is not allowed. Allowed fields are: '
```
njohner, line 259: Group order field -> Sort first field

opengever/api/tests/test_listing.py
```diff
@@ -14,7 +16,7 @@
 from opengever.testing import IntegrationTestCase
 from opengever.testing.integration_test_case import SolrIntegrationTestCase
 from plone.uuid.interfaces import IUUID
-from zExceptions import BadRequest
+from Products.Five.browser import BrowserView
```
njohner, line 19: That should be an unused import

opengever/api/tests/test_listing.py
```diff
@@ -1328,3 +1330,178 @@ def test_dossier_participant_and_roles_facets(self, browser):
...
+        self.assertEqual(
+            'if(or(termfreq(portal_type, type1)), 1, 0) desc',
+            ListingGet._build_sort_first_func_string('portal_type', ['type1']))
+
+        self.assertEqual(
+            'if(or(termfreq(portal_type, type1),termfreq(portal_type, type2)), 1, 0) desc',
+            ListingGet._build_sort_first_func_string('portal_type', ['type1', 'type2']))
+
+    @browsing
+    def test_sort_first_raises_error_if_not_using_as_a_record(self, browser):
```
njohner, line 1347: if_not_using_as_a_record -> if_parameter_not_a_record ?

opengever/api/tests/test_listing.py
```diff
@@ -1328,3 +1330,178 @@ def test_dossier_participant_and_roles_facets(self, browser):
...
+
+        with browser.expect_http_error(reason='Bad Request'):
+            browser.open(self.workspace, view=view, headers=self.api_headers)
+
+        self.assertEqual(
+            u'The sort_first parameter needs to be a record.',
+            browser.json.get('message'))
+
+    @browsing
+    def test_sort_first_raises_error_if_used_not_whitelisted_field(self, browser):
```
njohner, line 1364: used_with_not_whitelisted

njohner (approved):
LGTM, just a few more small comments

PR #6690, opened 2020-10-20
Don't create journal entry when downloading file copy in readonly mode.

lukasgraf:
Don't create journal entry when downloading file copy in readonly mode (original documents or PDF previews from bumblebee).

This is to **prevent the write-on-read** that would otherwise happen during journaling of the file download in readonly mode.

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- New functionality:

### DIFF
--- a/opengever/journal/handlers.py
+++ b/opengever/journal/handlers.py
@@ -10,6 +10,7 @@ from opengever.base.oguid import Oguid
 from opengever.document.document import IDocumentSchema
 from opengever.dossier.browser.participants import role_list_helper
 from opengever.journal import _
+from opengever.readonly import is_in_readonly_mode
 from opengever.repository.repositoryroot import IRepositoryRoot
 from opengever.sharing.behaviors import IStandard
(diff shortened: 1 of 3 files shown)


opengever/journal/handlers.py
```diff
@@ -431,6 +432,8 @@ def document_file_reverted(context, event):
 
 
 def file_copy_downloaded(context, event):
+    if is_in_readonly_mode():
```
njohner, line 435: Ok so still add a journal entry when not in read-only mode. I thought we'd simply stop adding entries for these actions, but it seems that is how it's specified in the story, so 👍

njohner (approved):
LGTM

opengever/journal/handlers.py
lukasgraf: Yep, ditching the journal entries on download entirely was never the plan. Though I personally wouldn't mind, but that's a different topic 😉

PR #6691, opened 2020-10-21
Show culprit traceback on ReadOnlyError page to all users, not just managers

lukasgraf:
When debugging user-specific login related writes-on-read, it's desirable to have the **culprit traceback** shown for users other than managers - otherwise, these writes-on-read get quite tricky to hunt down.

Thefore I remove the `is_manager` restriction for write-on-read culprit tracebacks.

I don't really consider this a security risk in terms of information disclosure:
This functionality (_write-on-read tracing_) has to be enabled deliberately by developers, times out, and could only ever be seen by end users during a short period of time, under very specific circumstances (when RO mode is active).

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 ----------------------
 
 - Make is_in_readonly_mode() slightly more robust. [lgraf]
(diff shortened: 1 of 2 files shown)


njohner (approved):
LGTM

PR #6692, opened 2020-10-21
Make is_in_readonly_mode() more robust when getSite() hook returns None

lukasgraf:
This might for example happen when a request hits the Zope application root instead of the Plone site.

Returning `False` in this case is incorrect, strictly speaking, but the only safe choice to fall back to (in the sense that in writable production mode, the system must _never_ be considered readonly by accident).

But since 
- the result of this flag has no discernable effect on the Zope App Root
- the worst that could happen as a consequence of False-Negative answers to `is_in_readonly_mode()` is that actions that won't work are shown

this should be fine.

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.12.0 (unreleased)
 ----------------------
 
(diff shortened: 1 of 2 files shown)


njohner (approved):
LGTM

PR #6693, opened 2020-10-22
Fix mimetype for quickupload with custom mimetypes

buchi:
c.quickupload correctly determines the mimetype from Plone's mimetypes registry. However our quick upload adapter doesn't pass the mimetype to the `CreateDocumentCommand`. Because of this plone.namedfile determines the mimetype using Python's mimetypes module, which doesn't work for custom mimetypes.

JIRA: https://4teamwork.atlassian.net/browse/CA-1143

## Checklist (Must have)

- [x] Changelog entry


njohner (approved):
LGTM

PR #6694, opened 2020-10-22
Fix order of labels for participations field in the listing endpoint.

njohner:
It seems that I messed up the label order for the `participations` field in the API endpoint. We were returning `role label|participant label` instead of `participant label|role label`.

For https://4teamwork.atlassian.net/browse/CA-17
## Checklist (Must have)

- [x] Changelog entry

- API change: Documentation is still ok, and it's not a breaking change as such as it's just the returned labels that change
  - If breaking:
    - [ ] api-change label added


lukasgraf (approved):
LGTM 👍

PR #6697, opened 2020-10-23
Custom error page: Also log ReadOnlyError culprit traceback

lukasgraf:
Custom error page: Also log `ReadOnlyError` culprit traceback (if one is available) to error log in `var/log/instance?-log`.

This makes debugging easier in situations where the error page contents aren't easily viewable in the browser, like for AJAX requests.

## Checklist (Must have)

- [x] Changelog entry


njohner (approved):
LGTM

PR #6699, opened 2020-10-23
Avoid ftw.casauth write-on-read (last login times) during login

lukasgraf:
Avoid `ftw.casauth` write-on-read (last login times) during login:

Patch `CASAuthenticationPlugin.set_login_times()` to avoid the write-on-read during login when attempting to set login times.

This is the same issue that's already being patched for the `MembershipTool`, but `ftw.casauth` contains its own copy of `setLoginTimes()` and therefore needs to be patched separately.

## Checklist (Must have)

- [x] Changelog entry


njohner (approved):
LGTM

PR #6702, opened 2020-10-23
Add script to toggle read-only mode in zope.conf

buchi:
This script allows to enable or disable the read-only mode of all instances in a GEVER deployment without running buildout.
It adds an environment variable `GEVER_READ_ONLY_MODE=true` and sets the ZODB connection to read-only in the `zope.conf` file.

By default all instances except for `instance0` are toggled. If an instance is in read-write mode it's set to read-only and vice versa.

With the command line options `-r` and `-w` it's possible to enable or disable the read-only mode regardless of the current state (see also `bin/toggle-readonly --help`)

After toggling the affected instances needs to be restarted.

JIRA: https://4teamwork.atlassian.net/browse/CA-211

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -10,6 +10,7 @@ Changelog
 - Bump ftw.tabbedview to 4.2.1 to get fix for empty action lists. [lgraf]
 - Add workspacemeetings to @listing endpoint. [tinagerber]
 - Fix order of labels for participations field in the listing endpoint. [njohner]
+- Add script to toggle read-only mode in zope.conf. [buchi]
 
 
 2020.12.0 (2020-10-22)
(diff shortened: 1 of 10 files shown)


lukasgraf (changes requested):
LGTM! 👍 
Just a minor detail: I already created a subpackage [opengever.readonly](https://github.com/4teamwork/opengever.core/tree/master/opengever/readonly) where I tried to group all (or most) of the readonly related functionality together, unless other considerations weighed stronger (_like keeping all the monkey patches together_).

Would you mind moving the code accordingly?

buchi:
Didn't see the `readonly` subpackage. Moved the code there now.

njohner (approved):
LGTM

PR #6703, opened 2020-10-25
Disable write actions in readonly mode.

lukasgraf:
Disable write actions in readonly mode, so that the user only gets presented with actions that can actually work in RO.

This will prevent actions from being shown to the user (in the classic UI as well as the API) by filtering the user's roles and permissions with the use of three monkey patches to Plone:
- [`PloneUser.allowed`](https://github.com/plone/Products.PlonePAS/blob/5.x/src/Products/PlonePAS/plugins/ufactory.py#L154-L199) (roles)
- [`PloneUser.getRolesInContext`](https://github.com/plone/Products.PlonePAS/blob/5.x/src/Products/PlonePAS/plugins/ufactory.py#L146-L152) (roles)
- [`ZopeSecurityPolicy.checkPermission`](https://github.com/zopefoundation/AccessControl/blob/3.0.x/src/AccessControl/ImplPython.py#L453-L488) (permissions)

These patches filter out roles and permissions known to require write access, and therefore withold them from the active user. This will deactivate the respective functionality, and will work for both the classic UI as well as the REST API (short of a few oddities). The patches that are hooking into Plone's security system are applied conditionally for now (only when the `GEVER_READ_ONLY_MODE` environment variable is set during instance startup). During tests, they're always active.

Filtering both roles and permissions is necessary because roles are often not specific enough to be considered "only-write" (especially admin-like and Owner roles), so lots of actions need to be deactivated via permissions. Workflow transition on the other hand are often protected by guard roles, therefore we need to filter those too.

Some code that causes writes but isn't controled via fine-grained permissions has simply been made conditional on the is_in_readonly_mode() function.

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- New functionality:


opengever/readonly/tests/test_write_actions_disabled_in_readonly.py
```diff
@@ -65,6 +67,21 @@ def test_no_edit_for_dossier(self, browser):
         self.login(self.regular_user, browser)
         self.assert_no_edit_link(browser, self.dossier)
 
+    @browsing
+    def test_no_edit_for_task(self, browser):
+        self.login(self.regular_user, browser)
+
+        open_task = create(
```
njohner, line 88: Could we have used `self.seq_subtask_1` instead of creating a new one here?

opengever/readonly/tests/test_all_roles_and_permissions_mapped.py
```diff
@@ -0,0 +1,340 @@
...
+                If in doubt, add it to NON_WRITE_PERMISSIONS. The worst that
+                can happen is that the user gets a ReadOnlyError during
+                readonly mode because an action was offered to them that
+                can't work.
+
+                The other way around we would be depriving them from
+                functionality that only requires read access, and *could*
+                work in readonly mode.
+                """ % permission
+                self.fail(dedent(msg))
```
njohner, line 340: Nice thanks! I'd add tests that the sets are disjoint, just so that we don't end up with stuff in both...
So `self.assertEqual(set(), WRITE_PERMISSIONS.intersection(NON_WRITE_PERMISSIONS))`and self.assertEqual(set(), WRITER_ROLES.intersection(NON_WRITER_ROLES))

opengever/ogds/base/events.py
```diff
@@ -53,7 +53,7 @@ def setup_engine_options(event):
     """
     engine = event.engine
 
-    if is_in_readonly_mode():
+    if is_in_readonly_mode() and engine.name != 'sqlite':
```
njohner, line 56: So sqlite does not support read-only? When / where are we using sqlite?

njohner (approved):
🎉 LGTM!
Thanks for the very small and clear commits!
I had just some minor comments, that we can also take over from here.

Some typos in the commit messages:
first and second commit messages: actions possible during -> actions impossible during

Somewhere in the middle:
Do not mark notifications unread in readonly mode. -> Do not mark notifications read in readonly mode.

opengever/ogds/base/events.py
lukasgraf: SQLite is only used during tests.

Oracle also supports some sort of readonly mode from what it looks like, but its syntax is most likely different. By not limiting it to only Postgres here we should notice immediately when trying it on Oracle, and then can figure out the correct syntax once we cross that bridge.

njohner (approved):

PR #6704, opened 2020-10-26
Extend @config endpoint with application type.

tinagerber:
This PR adds the application type to the @config endpoint. This allows the frontend to decide which app should be loaded independently from the portal.

Jira: https://4teamwork.atlassian.net/browse/NE-97

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:


deiferni (approved):
LGTM 👍

PR #6705, opened 2020-10-26
@@task_report-view supports task lookup by the ressource-id through the `tasks` parameter.

elioschmutz:
Pre-discussed with @deiferni : https://4teamwork.slack.com/archives/C01BHEGBVU1/p1603707985016500

Jira: https://4teamwork.atlassian.net/browse/NE-56

----- 

The `@@task_report` view is a folder-action which should also work for the gever-ui.

Currently, the view wants a list of `task_ids` which should be included in the report. Unfortunately, we don't have `task_ids` in the `@listing` endpoint of the frontend. Including it would mean, that we have to index the `task_ids` in solr and expose it int the `@listing` endpoint. This was not desired. Thus, we need another way to lookup the tasks.

Because the tasks in a default listing are always within the current admin unit, we can do a task lookup by the resource-id (`@id`)

This PR extends the `@@task_report` view with a new parameter: `tasks`. It provides a list of task resource ids which should be included in the report.

The view now provides two ways to include tasks:

by `task_id`:
`@@task_report?task_ids:list=1&task_ids:list=2`

by the resource id:
`@@task_report?tasks:list=https://example.com/ordnungssystem/dossier-1/task-1&tasks:list=https://example.com/ordnungssystem/dossier-1/task-2`

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 ----------------------
 
 - Use oguid instead of intId as token in DossierTemplatesVocabulary. [tinagerber]
+- `@@task_report`-view supports task lookup by the ressource-id through the `tasks` parameter. [elioschmutz]
 - Extend @config endpoint with application type. [tinagerber]
 - Journalize creation of linked workspace and copying documents to and from it. [njohner]
 - Disable write actions during readonly mode. [lgraf]
--- a/opengever/globalindex/model/task.py
+++ b/opengever/globalindex/model/task.py
@@ -588,6 +588,13 @@ class TaskQuery(BaseQuery):
         return self._extend_with_physical_path(
             query, Task.physical_path, path).first()
 
+    def by_paths(self, paths, admin_unit_id=None):
+        """Returns all tasks whos physical_path is listed in `paths`.
+        """
+        admin_unit_id = admin_unit_id or get_current_admin_unit().id()
+        query = self.restrict().filter_by(admin_unit_id=admin_unit_id)
+        return query.filter(Task.physical_path.in_(paths)).all()
+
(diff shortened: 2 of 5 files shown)


njohner (approved):
LGTM, I'm just a bit unsure about using absolute URLs.

Well I just went through your discussion with @deiferni and it seems using `path` or `@id` was discussed already. I think I'd have argued for using `path`s, but as long as it's a conscious decision to introduce what I'd call inconsistencies, but it seems in that case because it's the direction we want to push towards in the future, well then fine with me.

elioschmutz:
@njohner I don't know the "right" way. But i.e. the copy-move endpoint can handle absolute urls, paths or UID (https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/services/copymove/copymove.py#L26). Perhaps we should also use something like the `get_object` method of `copymove` to provide this tree ways to lookup an object.

But for now, we just have to extend the `task_report` view to also provide a way to lookup tasks through a path or ID.

@deiferni could you summarize why you prefered to use `@id` instead of the `path`? Do you still agree with this decision?

deiferni:
With the current implementation honestly i do not particularly care 😅 . 
In the thread i was arguing what might be a good idea with a service endpoint. What we are actually doing here is pass a list of references, and there are several ways this can be done with the current `plone.restapi` implementation. But we will have to touch this again eventually when we remove the browser views and switch to a service endpoint. So if you two prefer `path` i don't have any objection!

elioschmutz:
Thanks @deiferni . I prefer to merge this PR 😂

PR #6708, opened 2020-10-27
Ensure remaining indices are dropped from catalog.

deiferni:
We seem to have different indexes still around for different deployments regarding the catalog indices `document_author`, `SearchableText` and `total_comments`. The following states exist:

- Neither `document_author` nor `SearchableText`:
  This happened when solr was activated with the `activate_solr.py` script
  from `opengever.maintenance` https://github.com/4teamwork/opengever.maintenance/blob/a05f10627dd8729db13128d7a767d98d7696b508/opengever/maintenance/scripts/activate_solr.py. It deleted the `document_author` index which was still present in `opengever.core`.
- Only `document_author`:
  This is the state for all new deployments where `activate_solr.py` has not been run as the `document_author` index was still present in `opengever.core`.
- Indices `document_author`, `SearchableText` and `total_comments`:      
  The upgrade step to delete indices was faulty and did not correctly remove `SearchableText` and `total_comments` due to a missing comma:https://github.com/4teamwork/opengever.core/blob/aa2e84c93894f5ed9876b760c18cd16edf2f4a10/opengever/core/upgrades/20200518084137_cleanup_catalog_indexes_and_metadata/upgrade.py#L17-L18

With this PR we make sure that the remaining two indexes are dropped consistently for all deployments. It also conditionally clears the lexicon and rebuilds the remaining ZCTextIndex indices to ensure minimal catalog size.

Note that the `document_author` metadata is left in place as it is still used by the `ILaTexListing` for documents.

I have locally tested the following permutations:
- ✅ &nbsp; A deployment with `SearchableText`, `document_author` and `total_comments`.
- ✅ &nbsp; A deployment with `document_author` only.
- ✅ &nbsp; A deployment with `document_author` and `searchable_filing_no` from the `opengever.dossier:filing` profile.
- ✅ &nbsp; A fresh deployment contains the correct indices.

Jira: https://4teamwork.atlassian.net/browse/CA-1141

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 
 - Use oguid instead of intId as token in DossierTemplatesVocabulary. [tinagerber]
 - `@@task_report`-view supports task lookup by the ressource-id through the `tasks` parameter. [elioschmutz]
+- Ensure `document_author` and `SearchableText` indices are dropped from catalog. [deiferni]
 - Extend @config endpoint with application type. [tinagerber]
 - Journalize creation of linked workspace and copying documents to and from it. [njohner]
 - Disable write actions during readonly mode. [lgraf]
--- a/opengever/core/upgrades/20201028100049_ensure_solr_only_indices_dropped_from_catalog/upgrade.py
+++ b/opengever/core/upgrades/20201028100049_ensure_solr_only_indices_dropped_from_catalog/upgrade.py
@@ -1,0 +1,92 @@
+from ftw.upgrade import UpgradeStep
+from ftw.upgrade.progresslogger import ProgressLogger
+import logging
+
+
+logger = logging.getLogger('ftw.upgrade')
+
+
+ZCTEXT_INDEXES_TO_REMOVE = [
+    'document_author',
+    'SearchableText',
+]
+
+INDEXES_TO_REMOVE = [
+    'total_comments',
+]
+
+
+class EnsureSolrOnlyIndicesDroppedFromCatalog(UpgradeStep):
+    """Ensure solr only indices dropped from catalog.
+
+    We seem to have different indexes still around for different deployments.
+    The following states exist:
+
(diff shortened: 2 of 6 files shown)


njohner (approved):
LGTM. I guess `reindex_zc_text_indices` can be costly? I wonder whether we wouldn't want to add `Description` in the list of indices to remove, just for safety. I mean it seems that we did not run the `activate_solr` script for all deployments, so there might be some where the `Description` index is still there?

deiferni:
@njohner i have found no such case (for SaaS at least) with https://github.com/4teamwork/opengever.maintenance/pull/229. The `Description` index should always have been dropped, just as all other indexes. I have thought about adding all indexes that should have been dropped with https://github.com/4teamwork/opengever.core/blob/019c74f11787dbab7a5d07d6dc681df8678abefe/opengever/core/hooks.py#L115-L127 or https://github.com/4teamwork/opengever.core/blob/019c74f11787dbab7a5d07d6dc681df8678abefe/opengever/core/upgrades/20200518084137_cleanup_catalog_indexes_and_metadata/upgrade.py#L5-L19 but then hesitated to do so. See https://4teamwork.atlassian.net/browse/CA-1141?focusedCommentId=18866 for a list of cases i found. Might be best to quickly discuss this approach tomorrow though.

The upgrade is surprisingly fast, my local testrun with about 25k objects took only some seconds. It does not load the actual objects but uses brains to reindex instead. This is identical to the solution applied in https://github.com/4teamwork/opengever.maintenance/blob/a05f10627dd8729db13128d7a767d98d7696b508/opengever/maintenance/scripts/activate_solr.py#L95-L99.

deiferni:
As discussed today, updated PR with recent findings. Pulling in @buchi for a 2nd 👀 , too.

buchi (approved):
👍 LGTM

PR #6709, opened 2020-10-27
Use oguid instead of intId as token in DossierTemplatesVocabulary

tinagerber:
When creating a dossier from a template we need access to the template in the frontend. Since the IntId cannot be resolved by an endpoint, the Oguid is now set as token in the DossierTemplatesVocabulary, because we can resolve the Oguid with the @resolve-oguid endpoint.

Discussed with @deiferni 

Since the vocabulary was only accessible via API since the last release, I do not consider it necessary to notify the channel and the Scrum Master. 

Jira: https://4teamwork.atlassian.net/browse/NE-43

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [x] api-change label added

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.13.0 (unreleased)
 ----------------------
 
+- Use oguid instead of intId as token in DossierTemplatesVocabulary. [tinagerber]
 - Extend @config endpoint with application type. [tinagerber]
 - Journalize creation of linked workspace and copying documents to and from it. [njohner]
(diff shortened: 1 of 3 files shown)


deiferni (approved):
I'm a bit worried that no tests had to be adapted. This could either mean functionality is not tested well or that test are written in such a way that they fetch the value from the vocabulary. Could you either confirm or double check tests exist? Apart from that 👍.

tinagerber:
The tests test the title of the templates, not the tokens

deiferni:
👌 :shipit:

PR #6710, opened 2020-10-28
Agenda-item attachments are now ordered based on the position in the relationField.

elioschmutz:
Jira: https://4teamwork.atlassian.net/browse/NE-108

A user should be able to order agena-item attachments in the new UI. This is already possible through the restapi with unsubmitted proposals. While submitting the proposal, the related documents gets sorted by the title. The old ui meeting view which displays submitted proposals will now show the attachments ordered by title as well.

By removing the ordering, we get the documents in the order as they where created. And this is the user defined order, exactly what we want to display. 

Bildschirmfoto 2020-10-28 um 17 23 23

This PR no longer orders the agenda-item attachments by name but by the position in the relationItem field instead:

Bildschirmfoto 2020-10-28 um 17 23 13

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/meeting/browser/meetings/agendaitem.py
+++ b/opengever/meeting/browser/meetings/agendaitem.py
@@ -189,8 +189,9 @@ class AgendaItemsView(BrowserView):
         if not item.has_proposal:
             return []
 
-        docs = IContentListing(item.proposal.resolve_submitted_documents())
-        return [doc.render_link() for doc in sorted(docs, key=lambda doc: doc.title_or_id().lower())]
+        submitted_proposal = item.proposal.resolve_submitted_proposal()
+        docs = IContentListing(submitted_proposal.get_documents())
+        return [doc.render_link() for doc in docs]
 
     def _serialize_submitted_excerpt(self, item):
         if not item.has_proposal:
--- a/opengever/meeting/proposal.py
+++ b/opengever/meeting/proposal.py
@@ -382,7 +382,7 @@ class SubmittedProposal(ModelContainer, ProposalBase):
         documents = catalog(
             portal_type=['opengever.document.document', 'ftw.mail.mail'],
             path=dict(query='/'.join(self.getPhysicalPath())),
-            sort_on='sortable_title'
+            sort_on='getObjPositionInParent'
             )
 
         ignored_documents = [self.get_excerpt()]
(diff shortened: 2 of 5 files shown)


opengever/meeting/browser/meetings/agendaitem.py
```diff
@@ -189,8 +189,11 @@ def _serialize_submitted_documents(self, item):
         if not item.has_proposal:
             return []
 
-        docs = IContentListing(item.proposal.resolve_submitted_documents())
-        return [doc.render_link() for doc in sorted(docs, key=lambda doc: doc.title_or_id().lower())]
+        proposal = item.proposal.resolve_proposal()
```
deiferni, line 192: this is for the meeting overview, right? using the proposal here is not correct. the proposal is in the dossier and in the repository. the user viewing meetings may not have permissions to view the repository and vice versa, you should use `resolve_submitted_proposal` instead.

deiferni (approved):
👍  LGTM except from wrong method called. Feel free to merge once it is fixed.

elioschmutz:
@deiferni that was a good point and it forced me to rethink my implementation.

A proposal contains the attachments in the `relatedItems` field which is an ordered list of documents. The attachments of submitted documents are objects within the container itself. And the sort order was forced to be by title on creation time.

So I have to remove the hardcoded ordering to get the user defined sort order in a later lookup of submitted documents.

Because I now changed the main-implementation, I'll re-request a review from you...

opengever/meeting/proposal.py
```diff
@@ -522,10 +522,7 @@ def can_comment(self):
         return api.user.has_permission('opengever.meeting: Add Proposal Comment', obj=self)
 
     def get_documents(self):
-        return sorted(
-            [relation.to_object for relation in self.relatedItems],
-            key=lambda document: document.title_or_id(),
-            )
+        return [relation.to_object for relation in self.relatedItems]
```
deiferni, line 525: So this was necessary to make sure the documents are created in the correct order as defined by the relation when submitting the proposal, right?
As the submitted documents are then just returned by https://github.com/4teamwork/opengever.core/blob/74ede62075a346754c5c85d11c7755431591442f/opengever/meeting/proposal.py#L380-L392 in existing order 🤔 ?

deiferni (approved):
LGTM 👍 , just one question for clarification.

opengever/meeting/proposal.py
elioschmutz: Yes, we have to create the documents in exactly this order. The catalog-query then orders by `objPositionInParent`.

Perhaps you didn't see it, but I also changed line: 385 from `sort_on='sortable_title'` to `sort_on='getObjPositionInParent'`.

deiferni: Ok thanks, :shipit:😄 !

PR #6711, opened 2020-10-30
Remove limit for Subject facet

bierik:
In https://4teamwork.atlassian.net/browse/NE-173, the user reported an issue, where not all facets are listed in the dossier listing.
This is becuase the default facet limit is 100 in solr 8.4: https://lucene.apache.org/solr/guide/8_4/faceting.html#field-value-faceting-parameters.
So, remove the limit of the Subject facets by including f.Subject.facet.limit=-1 in the solr query.

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 ----------------------
 
 - Agenda-item attachments are now ordered based on the position in the relationField. [elioschmutz]
+- Remove the limit for facets returned in the listing API endpoint. [Kevin Bieri]
 - Use oguid instead of intId as token in DossierTemplatesVocabulary. [tinagerber]
 - `@@task_report`-view supports task lookup by the ressource-id through the `tasks` parameter. [elioschmutz]
 - Ensure `document_author` and `SearchableText` indices are dropped from catalog. [deiferni]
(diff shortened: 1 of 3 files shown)


njohner (commented):
I'd add a regression test ;-)

And more generally, won't we have exactly the same issue with other fields? like `responsible` for a dossier or a task for example? I wonder if we don't want to set the limit to `-1` for all facets...
I guess this introduces the risk that the response will really get blown up by the facets, but on the other side having only part of the facets to choose from does not make sense either...

bierik:
I thought about setting the limit to `-1` for all facets but, as you said, the response may blow up. So I went for a trade off and just removed the limit for one field. Correct, this will just solve one particular issue but I am more afraid of letting the user wait for the solr response. In the ancient UI, this is not an issue, but I did not figure out how this is implemented (in ftw.tabbedview somehow 😄 ). Maybe you can give me a hint, so we can do it the same way.

elioschmutz:
> I thought about setting the limit to `-1` for all facets but, as you said, the response may blow up. So I went for a trade off and just removed the limit for one field. Correct, this will just solve one particular issue but I am more afraid of letting the user wait for the solr response. In the ancient UI, this is not an issue, but I did not figure out how this is implemented (in ftw.tabbedview somehow 😄 ). Maybe you can give me a hint, so we can do it the same way.

@bierik there is a general limit of `-1`: https://github.com/4teamwork/opengever.core/blob/master/opengever/tabbedview/filters.py#L187

bierik:
@njohner I added a regression test and set the `facet.limit` to `-1` as @elioschmutz recommended.

njohner (approved):
LGTM

PR #6712, opened 2020-10-30
Webactions integration in @actions endpoint

elioschmutz:
Currently, webactions are only manageable. To be able to display registered webactions in the UI, we have to expose registered webactions somewhere. We decided to integrate them into the `@actions` endpoint in a new category `webactions`.

This PR overrides the `@actions` endpoint for GEVER and extends the result with `webactions`. If there are no webactions available, it will return an empty list for the `webactions`-category.

Because a webaction consist of many attributes we don't want to expose to the end-user, we white-list the desired attributes for the serialized `webactions` in the `@actions` endpiont.

Bildschirmfoto 2020-10-30 um 13 44 06

Pre-Discussion: https://4teamwork.slack.com/archives/C01BHEGBVU1/p1603815306027700

Jira: https://4teamwork.atlassian.net/browse/NE-65

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 
 - Agenda-item attachments are now ordered based on the position in the relationField. [elioschmutz]
 - Remove the limit for facets returned in the listing API endpoint. [Kevin Bieri]
+- `@actions` endpoint also returns available webactions. [elioschmutz]
 - Use oguid instead of intId as token in DossierTemplatesVocabulary. [tinagerber]
 - `@@task_report`-view supports task lookup by the ressource-id through the `tasks` parameter. [elioschmutz]
 - Ensure `document_author` and `SearchableText` indices are dropped from catalog. [deiferni]
--- a/opengever/api/actions.py
+++ b/opengever/api/actions.py
@@ -1,0 +1,59 @@
+from opengever.base.interfaces import IOpengeverBaseLayer
+from opengever.webactions.interfaces import IWebActionsProvider
+from plone.restapi.interfaces import IExpandableElement
+from plone.restapi.serializer.converters import json_compatible
+from plone.restapi.services.actions.get import Actions
+from plone.restapi.services.actions.get import ActionsGet
+from zope.component import adapter
+from zope.component import queryMultiAdapter
+from zope.interface import implementer
+from zope.interface import Interface
+
+
+WEBACTION_ATTRIBUTE_WHITELIST = [
+    'action_id',
+    'title',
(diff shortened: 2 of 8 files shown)


docs/HISTORY.txt
```diff
@@ -4,6 +4,7 @@ Changelog
 2020.13.0 (unreleased)
 ----------------------
 
+- @actions` endpoint also returns available webactions. [elioschmutz]
```
njohner, line 7: misssing `

opengever/api/actions.py
```diff
@@ -0,0 +1,44 @@
...
+        actions['webactions'] = []
+
+        if not provider:
+            return
+
+        webactions = provider.get_webactions(flat=True)
+        for webaction in webactions:
+            for attr in webaction.keys():
+                if attr not in WEBACTION_ATTRIBUTE_WHITELIST:
+                    del webaction[attr]
```
njohner, line 42: This is fine because `get_webactions` actually returns a copy of the webaction, but I always find directly modifying mutable types like that a bit risky for the future. It's probably also less efficient to successively delete from the `webaction` dictionnary than create a new one. So I'd rather have
`webactions = [{key: action.get(key) for key in WEBACTION_ATTRIBUTE_WHITELIST} for action in provider.get_webactions(flat=True)]`

njohner (approved):
Apart from comment, LGTM.

opengever/api/actions.py
elioschmutz: 👍 updated

elioschmutz:
@njohner I udpated the PR according to your suggestions.

I realized, that I just extended the `GET @actions` endpoint but not the expansion `?expand=actions`. So I updated my PR to override the base `Actions`-adapter instead the `ActionsGet`-browserview.

Could you take another look? Thanks!

njohner (commented):
I'd say, in that case how about one test to make sure webactions appear in the expansion?

Otherwise LGTM

elioschmutz:
> I'd say, in that case how about one test to make sure webactions appear in the expansion?
> 
> Otherwise LGTM

🤔 there is one... https://github.com/4teamwork/opengever.core/pull/6712/files#diff-505e24b0e62316503d395fca49effe5c2e5c2fcfb14b5db268855ea9edc8e31bR807

njohner (approved):

njohner:
Sorry, didn't see 😅

PR #6713, opened 2020-10-30
Add `@actors` endpoint

njohner:
With this PR we make a first step towards consistently using actors in the REST-API, the advantage being that external applications should not have to know what identifier corresponds to what actor type and how to handle it. Actors are an abstraction that should be provided as such to external applications. Here we:
- Add a new `@actors` endpoint:
  - with `@actors/actor_id` one can get the data corresponding to a particular actor
  - A POST to `@actors` allows to retrieve a list of serialized actors
- We add a summary serialization of actors that we can use in different endpoints.
  - This summary serialization needs to be fast, and therefore only contains data that can be generated without having to retrieve the actor.
  - The summary serialization contains an `@id` allowing to retrieve the complete data for the given actor
- We now return `actor`s for different fields in different endpoints, notably:
  - the `@sharing` endpoint
  - the `@participations` endpoint
  - the `@watchers` endpoint
  - Dossier responsible (in dossier GET)
  - Issuer and responsible for tasks in the `@globalindex` endpoint

We avoid breaking changes in that we return the `actor`s in separate fields and add a note that these new fields supersede the existing ones which are deprecated.

I decided not to add too much data int he serialized actors, as I'd rather do this iteratively. The frontend should start using the actors and see what it really needs, so that we can decide what additional data belongs in the serialized actors. I have nevertheless added a new `get_portrait_url` method to actors and serialize this in the `@actors` endpoint, as this one field is necessary for the frontend and seemed generic enough to me.

For https://4teamwork.atlassian.net/browse/CA-406

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [x] api-change label added: changes are non-breaking for now but deserve to be announced


opengever/ogds/base/interfaces.py
```diff
@@ -133,3 +133,35 @@ class IInternalOpengeverRequestLayer(Interface):
 
 class ITeam(Interface):
     """Marker interface for team object wrappers."""
+
+
+class IActor(Interface):
```
lukasgraf, line 138: 💚

opengever/api/actors.py
```diff
@@ -63,3 +64,40 @@ def read_params(self):
...
+    a POST request containing a list of actor_ids in the request body.
+
+    POST /@actors HTTP/1.1
+    """
+
+    @property
+    def request_data(self):
+        return json_body(self.request)
+
+    def extract_data(self):
```
lukasgraf, line 93: The `extract_data` method is unused and can be deleted from what I can tell. (Or used and rewriten to require `actor_ids`).

opengever/api/actors.py
```diff
@@ -13,6 +13,18 @@
 from zope.publisher.interfaces import IPublishTraverse
 
 
+def serialize_actor_id_to_json_summary(actor_id):
+    """This method is used to return a standardized summary representation
+    of an actor. We should keep this very efficient as it will be used by
+    various endpoints returning actors or lists of actors for certein fields.
```
lukasgraf, line 19: `certain`

docs/public/dev-manual/api/watchers.rst
```diff
@@ -3,6 +3,8 @@ Beobachter
 
 Der ``@watchers`` Endpoint dient dazu, für Aufgaben und Weiterleitungen Beobachter zu registrieren.
 
+Diese Endpoints liefern das ``referenced_users``, welches überholt ist und durch das ``referenced_actors``
+Feld ersetzt wurde. ``referenced_users`` wird in der Zukunft nicht mehr unterstüzt.
```
lukasgraf, line 7: Diese Endpoints liefern zur Zeit sowohl ``referenced_users`` als auch ``referenced_actors`` zurück. ``referenced_users`` wird in in einer späteren Version jedoch nicht mehr unterstützt werden, und wird durch ``referenced_actors`` abgelöst.

docs/public/dev-manual/api/dossier_participations.rst
```diff
@@ -3,6 +3,7 @@ Beteiligungen
 
 Der ``@participations`` Endpoint behandelt Dossierbeteiligungen.
 
+Die ``participant_id`` und ``participation_role`` Felder sind überholt und wurden durch das ``participant_actor`` Feld ersetzt.
```
lukasgraf, line 6: ``participation_role`` -> ``participation_title`` (I think)

docs/public/dev-manual/api/actors.rst
```diff
@@ -0,0 +1,83 @@
+.. _actors:
+
+Aktoren
```
lukasgraf, line 3: I would simply call these by the English term `Actors`. In German it would be `Akteure`, but that sounds a bit stilted IMHO.

docs/public/dev-manual/api/actors.rst
```diff
@@ -0,0 +1,83 @@
...
+=======
+
+Ein Aktor ist eine generische Repräsentation von Benutzer, Gruppen, Teams, Kontakte, Eingangskörbe und Gremien. Aktoren werden in verschiedenen Endpoints und für gewisse Felder (zum Beispiel Dossierver Federführung) verwendet.
```
lukasgraf, line 6: Dossierver -> Dossier

docs/public/dev-manual/api/actors.rst
```diff
@@ -0,0 +1,83 @@
...
+        "portrait_url": "http://example.org/defaultUser.png"
+      }
+
+Via POST können die Daten von mehreren Aktoren mit einem Request abgefragt werden. Im Request-body wird die Liste von Aktor ID angegeben:
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+      POST /@actors/peter.mueller HTTP/1.1
```
lukasgraf, line 43: ✂️ `/peter.mueller`

lukasgraf (approved):
LGTM 👍!
Just a few minor points.

opengever/api/actors.py
njohner: Oh sorry, that looks like a big copy / paste fail

docs/public/dev-manual/api/dossier_participations.rst
njohner: participant_title

PR #6716, opened 2020-11-02
Use UID instead of oguid as token in DossierTemplatesVocabulary.

elioschmutz:
This PR is a follow-u pr of #6709 

We want to use the vocabulary in combination with a selection-widget which is based on solr.
Because we don't indexed the OGUID in solr, we're not able to intersect vocabulary items and solr items.

This PR now uses the `UID` instead which is already indexed in solr.

Jira: https://4teamwork.atlassian.net/browse/NE-43

Prediscussed with @njohner: https://4teamwork.atlassian.net/browse/NE-43?focusedCommentId=19007

> While we might want to try to uniformly return OGUIDs everywhere in the future, this is not a realistic option right now. Indexing the OGUID would also be very expensive, as it is not in the catalog, it would require an upgrade step that wakes all the objects from the catalog…

> So for now I advised that we go with the UID. If we ever decide to index the OGUID, and start using it everywhere in the API, we will anyway have to adapt the external applications.
## Checklist (Must have)

- [ ] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -7,6 +7,7 @@ Changelog
 - Agenda-item attachments are now ordered based on the position in the relationField. [elioschmutz]
 - Remove the limit for facets returned in the listing API endpoint. [Kevin Bieri]
 - Use oguid instead of intId as token in DossierTemplatesVocabulary. [tinagerber]
+- Use UID instead of intId as token in DossierTemplatesVocabulary. [tinagerber, elioschmutz]
 - `@@task_report`-view supports task lookup by the ressource-id through the `tasks` parameter. [elioschmutz]
 - Ensure `document_author` and `SearchableText` indices are dropped from catalog. [deiferni]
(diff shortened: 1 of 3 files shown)


docs/HISTORY.txt
```diff
@@ -5,7 +5,7 @@ Changelog
 ----------------------
 
 - Agenda-item attachments are now ordered based on the position in the relationField. [elioschmutz]
-- Use oguid instead of intId as token in DossierTemplatesVocabulary. [tinagerber]
+- Use UID instead of intId as token in DossierTemplatesVocabulary. [tinagerber]
```
njohner, line 8: tinagerber -> tinagerber, elioschmutz

njohner (approved):
apart from comment, LGTM

PR #6717, opened 2020-11-02
Add debug patches to hunt for out of sync modified (catalog v.s. objects).

deiferni:
With this PR we add patches that help to hunt down the very rarely occurring out of sync modified of catalog v.s. content objects. We suspect there is a missing reindex when the attribute is set. I have not managed to figure it out locally, so best strategy IMO would be to enable the patches on our dev system and wait.

The patches are inactive by default and should only be activated on a non-production system for the moment. They are designed in such a way that they can be activated for a deployment via buildout configuration in the `instance` section:

```
[instance]
environment-vars +=
    GEVER_DEBUG_MODIFIED_OUF_OF_SYNC True
```

We want to activate them on dev for now and as the issue occurs very rarely activation has to be persistent across instance restarts.

- Write on read tracing has already provided functionality for serializing stack-traces in a safe and unobtrusive way. I have extracted the methods into their own module and also and slightly refactored them to enable easier re-use.
- I have opted to put the patch into `opengever.debug` to clarify it's intention as a (potentially temporary) debug tool. 
- I had to change sentry settings as described in https://4teamwork.atlassian.net/browse/CA-190?focusedCommentId=19057.
- A sample sentry issue can be found at https://sentry.4teamwork.ch/organizations/sentry/issues/69656/?project=17.

As we now potentially patch the same method as write on read tracing i had to slightly change how write on read tracing retrieves and stores a reference to the patched methods to avoid side-effects between the two patches. Namely we need to prevent write on read tracing from erroneously unpatching the monkey patches provided in this PR in case they are activated. To achieve this we no store references to the patched method a patch time and not a module load
time:
- We now set the module-global reference to the original method in a deferred way at patch time.
- Some extra paranoia is added in form of additional assertions to make sure we don't set method references to
`None` by mistake.
- Also patching and unpatching is now protected by an additional lock to prevent simultaneous requests in separate threads from making a mess.
- The comparisons/assertions had to be rewritten slightly in order to work with the difference of `unbound method` and `function ` in `python2`, in case of `unbound method` we use `__func__` to compare against the wrapped function instead.
- To clarify entry points for clients the two separate patching and unpatching methods have been inlined into one method each.

I have tested:
- ✅  &nbsp;Reporting/logging modified to sentry works correctly and contains the serialized stacktrace (see above for link).
- ✅  &nbsp;Patching/unpatching write on read tracing still works.
- ✅  &nbsp;Patching/unpatching write on read tracing twice will hit the correct assertions.
- ✅  &nbsp;Patching/unpatching write on read tracing will leave the modified patch in place, should it be active.
- ✅  &nbsp;The write on read report is still correctly written and contains the serialized stacktrace.

_Nevertheless i'd be really glad for an extra careful review here, best for a local test-run to hunt for edge-cases i could not have spotted 😓_

Jira: https://4teamwork.atlassian.net/browse/CA-190

ToDo:
- [x] 🚧  &nbsp;cleanup https://sentry.4teamwork.ch/organizations/sentry/issues/69656/?project=17 once this PR is either merged or closed.
- [ ] 🚧  &nbsp;activate on dev once this is merged (separate PR).

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 ----------------------
 
 - Delete old upgrade steps up to and including 2018.5.7. [njohner]
+- Add monkey-patch to track out of sync modified. [deiferni]
 - Agenda-item attachments are now ordered based on the position in the relationField. [elioschmutz]
 - Remove the limit for facets returned in the listing API endpoint. [Kevin Bieri]
 - `@actions` endpoint also returns available webactions. [elioschmutz]
--- a/opengever/base/monkey/patches/__init__.py
+++ b/opengever/base/monkey/patches/__init__.py
@@ -34,6 +34,8 @@ from .tz_for_log import PatchZ2LogTimezone
 from .verify_object_paste import PatchCopyContainerVerifyObjectPaste
 from .webdav_lock_timeout import PatchWebDAVLockTimeout
 from .workflowtool import PatchWorkflowTool
+from opengever.debug import debug_modified_out_of_sync_env_var_is_set
+from opengever.debug.patches.modified_out_of_sync import PatchConnectionRegister
 from opengever.readonly import readonly_env_var_is_set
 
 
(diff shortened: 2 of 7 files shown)


deiferni:
@lukasgraf as discussed this morning/yesterday already i think it would make most sense if you could review this one. I have run into side-effects between your existing patches and my new ones, thus i had to touch the way the existing write on read patches store references to the patched methods. I'd be really glad for an extra careful review/opinion regarding feasibility of this change here!

lukasgraf (approved):
LGTM! 👍

Kudos for handling 🍌 [bananas](https://twitter.com/davisagli/status/1322185741959487495) 😉

PR #6719, opened 2020-11-03
Fix the Workspace `@participations` endpoint for NullActors

njohner:
`@participations` endpoint was failing for participations with a `participant_id` that could not be resolved to an actor. This is because the `represents` from the `NullActor` was broken. It now returns `None`, as it probably was intended to, which also has to be handled in the `@participations` endpoint.

With this PR we fix this issue, but we should probably remove the `represents` from the `Actor` and simply return an actor in this `@participations` endpoint. This endpoint was forgotten in https://github.com/4teamwork/opengever.core/pull/6713. I will either make a PR for this adaptation or open a new story.

For https://4teamwork.atlassian.net/browse/CA-1185

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.13.0 (unreleased)
 ----------------------
 
+- Fix the Workspace `@participations` endpoint for NullActors. [njohner]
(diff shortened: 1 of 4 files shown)


lukasgraf (approved):
LGTM! 👍

njohner:
Perfect, thanks for testing!

PR #6720, opened 2020-11-03
Fix resolving subdossiers when Teamraum Connect feature is enabled.

lukasgraf:
This fixes resolving subdossiers when the Teamraum Connect feature is enabled.

Subdossiers are not supposed to have linked workspaces (only main dossiers may). But because the `ILinkedWorkspace` adapter [deliberately raises a `ComponentLookupError`](https://github.com/4teamwork/opengever.core/blob/master/opengever/workspaceclient/linked_workspaces.py#L83-L84) when instantiated for a dossier, which would otherwise be adaptable just fine, that lead to a failure when checking dossier resolution preconditions.

This is a quick fix in order to still get it in before we cut the release. Overall, I would really question the approach of raising `ComponentLookupError` there instead of just returning empty lists etc.. But for now I fixed it in the most local way possible, in order to not destabilize the code base any more with an emergency fix.  

Jira: https://4teamwork.atlassian.net/browse/CA-1129

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 ----------------------
 
 - Convert bytestring values for IOpengeverBase.description field to unicode instead of raising an error. [elioschmutz]
+- Fix resolving subdossiers when Teamraum Connect feature is enabled. [lgraf]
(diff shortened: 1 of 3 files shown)


njohner (approved):
LGTM

PR #6721, opened 2020-11-03
Return participant_actor in workspace participations endpoint.

njohner:
The workspace `@participations` endpoint got forgotten in https://github.com/4teamwork/opengever.core/pull/6713. We now also return the `participant_actor` in that endpoint, so that in the long run we can remove the custom serialized `participant` from the response.

For  https://4teamwork.atlassian.net/browse/CA-406

## Checklist (Must have)

- [ ] Changelog entry: No CL needed IMO, this should have been included in the PR adding the `@actors` endpoint

## Checklist (if applicable)

- API change:
  - If breaking:
    - [x] api-change label added: as for the other PR part of that story, not breaking but should be announced.


deiferni (approved):
LGTM 👍

PR #6722, opened 2020-11-04
Interpolate target_url for webactions exposed in the @actions endpoint

elioschmutz:
This PR now uses the `WebActionsSafeDataGetter` instead directly the `IWebActionsProvider` adapter to lookup webactions in the `@actions` endpoint. This ensures, that the webaction target urls are interpolated properly and the fields are sanitized.

Currently, not flattened webactions from the `IWebActionsProvider` where filtered through: `_filter_action_data`. But not flattened. That's why we use `_filter_action_data` for flattened webactions now to get the same response for each type of representation. We have to whitelist the `display` attribute within the `IWebActionsProvider`. We need the `display` attribute in the serialized webactions for the `@actions` endpoint.

Jira: https://4teamwork.atlassian.net/browse/NE-65

## Checklist (Must have)

- [x] Changelog entry (already exists from #6712 

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/api/actions.py
+++ b/opengever/api/actions.py
@@ -1,5 +1,6 @@
 from opengever.base.interfaces import IOpengeverBaseLayer
 from opengever.webactions.interfaces import IWebActionsProvider
+from opengever.webactions.renderer import WebActionsSafeDataGetter
 from plone.restapi.interfaces import IExpandableElement
 from plone.restapi.serializer.converters import json_compatible
 from plone.restapi.services.actions.get import Actions
@@ -40,14 +41,10 @@ class GeverActions(Actions):
         environment (context, user, permissions, webaction-settings) in a flat
         structure.
         """
-        provider = queryMultiAdapter((self.context, self.request), IWebActionsProvider)
-
-        if not provider:
-            return
-
+        data_getter = WebActionsSafeDataGetter(self.context, self.request, None)
         webactions = [
             {key: webaction.get(key) for key in WEBACTION_ATTRIBUTE_WHITELIST}
-            for webaction in provider.get_webactions(flat=True)]
+            for webaction in data_getter.get_webactions_data(flat=True)]
 
         if webactions:
             actions['webactions'] = json_compatible(webactions)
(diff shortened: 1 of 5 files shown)


opengever/webactions/provider.py
```diff
@@ -133,7 +133,7 @@ def get_webactions(self, display=None, flat=False):
         webactions = self._sort(self._filter_for_user(webactions))
 
         if flat:
-            return webactions
+            return map(self._filter_action_data, webactions)
```
njohner, line 136: Fundamentally I agree here, the provider should be responsible of deciding what data is important

njohner: `display` will sort of be duplicate now in the case of `flat=False`, as it is in the keys and in the data. Doesn't matter too much though

opengever/api/actions.py
```diff
@@ -40,14 +41,10 @@ def extend_with_webactions(self, actions):
...
         structure.
         """
-        provider = queryMultiAdapter((self.context, self.request), IWebActionsProvider)
-
-        if not provider:
-            return
-
+        data_getter = WebActionsSafeDataGetter(self.context, self.request, None)
         webactions = [
             {key: webaction.get(key) for key in WEBACTION_ATTRIBUTE_WHITELIST}
```
njohner, line 46: I'm a bit bothered by the double whitelist, one in the provider and one here. Basically it should be the providers job to decide what data is important for a webaction. Of course we can decide in the endpoint here that some of the fields are not relevant (I think just the icons basically), but in that case a blacklist would be more appropriate. Now it feels very duplicated to me. And what are we actually blacklisting? The icon? Doesn't it make sense, generally speaking, to also return the icon for the webaction?

njohner (approved):
Basically LGTM, I'm not a big fan of the double whitelisting happening now in the provider and in the API, but I can live with it.

opengever/api/actions.py
elioschmutz: I agree! I'll remove the whitelist implementation since we filter the attributes already in the webactions-provider.

elioschmutz: 👍 updated in a new pr

PR #6723, opened 2020-11-04
Convert bytestring values for IOpengeverBase.description field to unicode instead of raising an error.

elioschmutz:
Jira: https://4teamwork.atlassian.net/browse/NE-247

This PR quick-fixes an issue where it is not possible to create content with an empty bytestring as the description. Currently, every plone content created through the `plone.restapi` without a description will have an empty bytestring instead an empty unicode string as the default value. 

Setting the description directly is possible. That's what the `plone.restapi` does. But setting it though the behavior will fail due to a guard in the `IBaseBehavior`. When creating a dossier form a dossier-template, the main dossier will be created through the `restapi` which works fine. But the subdossiers are created with the `createContentInContainer` method which will then raise an error.

The story https://4teamwork.atlassian.net/browse/CA-918 would fix the main issue by patching the `plone.restapi` and migrating the bad data-types of existing contents.

Because we need a quickfix, we decided to just remove our behavior field guard and transform bytestrings to unicode strings if necessary.

This quick-fix gets reverted in the scope of the story: https://4teamwork.atlassian.net/browse/CA-918

Full discussion: https://4teamwork.slack.com/archives/C060FEYKA/p1604433964032600

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.13.0 (unreleased)
 ----------------------
 
+- Convert bytestring values for IOpengeverBase.description field to unicode instead of raising an error. [elioschmutz]
 - Fix the Workspace `@participations` endpoint for NullActors. [njohner]
 - Delete old upgrade steps up to and including 2018.5.7. [njohner]
 - Add monkey-patch to track out of sync modified. [deiferni]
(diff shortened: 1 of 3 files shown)


deiferni (approved):
As already discussed LGTM 👍.

lukasgraf (approved):
LGTM 👍

PR #6724, opened 2020-11-04
Add flag to disregard retention period when creating a disposition.

deiferni:
With this PR we add a new new flag we allow to optionally disable the restriction to only add expired dossiers to a disposition. All closed dossiers will be allowed for selection disregarding any retention period setting.

- Add flag to disregard retention period, not enabled by default
- Expose flag in `@config`
- Partially related: follow naming conventions for REST-Api, `Config`->`ConfigGet`

Jira: https://4teamwork.atlassian.net/browse/CA-884

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/disposition/interfaces.py
+++ b/opengever/disposition/interfaces.py
@@ -54,6 +54,18 @@ class ISIPTransport(Interface):
         """
 
 
+class IDispositionSettings(Interface):
+
+    disregard_retention_period = schema.Bool(
+        title=u'Disregard retention period',
+        description=u'Whether it is only allowed to select expired dossiers '
+                    'for a disposition. If set to False only expired dossiers '
+                    'will be allowed, if set to True non-expired dossiers '
+                    'are allowed as well, disregarding their retention '
+                    'period.',
+        default=False)
+
+
 class IFilesystemTransportSettings(Interface):
 
     enabled = schema.Bool(
--- a/opengever/disposition/validators.py
+++ b/opengever/disposition/validators.py
@@ -1,7 +1,9 @@
 from opengever.disposition import _
 from opengever.disposition.disposition import IDispositionSchema
 from opengever.disposition.interfaces import IDisposition
+from opengever.disposition.interfaces import IDispositionSettings
 from plone import api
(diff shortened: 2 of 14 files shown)


opengever/disposition/interfaces.py
```diff
@@ -54,6 +54,18 @@ def is_enabled():
         """
 
 
+class IDispositionSettings(Interface):
+
+    disregard_retention_period = schema.Bool(
+        title=u'Disregard retention period',
+        description=u'Whether it is allowed to only select expired dossiers '
```
njohner, line 61: Whether it is allowed to only select -> Whether it is only allowed to select

opengever/disposition/validators.py
```diff
@@ -16,12 +17,16 @@ def validate(self, value):
...
+        disregard_retention_period = api.portal.get_registry_record(
+            name='disregard_retention_period', interface=IDispositionSettings)
+
         for dossier in value:
-            if not dossier.is_retention_period_expired():
-                raise Invalid(
-                    _(u'error_retention_period_not_expired',
-                      default=u'The retention period of the selected dossiers'
-                      ' is not expired.'))
+            if not disregard_retention_period:
```
njohner, line 24: if not disregard_retention_period and not dossier.is_retention_period_expired()?

njohner (commented):
What happens when a dossier is added to a disposition and then the dossier is reopened?

opengever/disposition/validators.py
deiferni: i could also apply de-morgan and make this `if not (disregard_retention_period or dossier.is_retention_period_expired())` 😏 🙃 

on a more serious note, i have wrapped all the negation into a separate method, hoping this will clarify what's going on and make things more readable.

deiferni:
@njohner amended as commented above. ready for re-review.

> What happens when a dossier is added to a disposition and then the dossier is reopened?
You can't. Once the dossier is part of a disposition its state will change to offered. As far as i've seen such dossiers cannot be reopened.

njohner (approved):
LGTM, thanks for the clarification about reopenning.

PR #6725, opened 2020-11-05
Use UID instead of intId as token in DocumentTemplatesVocabulary

elioschmutz:
Jira: https://4teamwork.atlassian.net/browse/NE-164

We want to use the vocabulary in combination with a selection-widget which is based on solr.
Because we don't indexed the OGUID in solr, we're not able to intersect vocabulary items and solr items.

This PR now uses the UID instead which is already indexed in solr.

Implemented according to https://github.com/4teamwork/opengever.core/pull/6716 where we already changed the vocabulary-token for the `DossierTemplateVocabulary`.

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 ----------------------
 
 - Add @notification-settings API endpoint. [tinagerber]
+- Use UID instead of intId as token in DocumentTemplatesVocabulary. [elioschmutz]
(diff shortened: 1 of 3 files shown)


njohner (approved):
LGTM

PR #6726, opened 2020-11-05
Add GEVER_COLORIZATION to the config endpoint

2e12:
The color should be displayed in new UI later, so it is necessary to display it in the endpoint.

https://4teamwork.atlassian.net/browse/CA-1147

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/docs/public/dev-manual/api/config.rst
+++ b/docs/public/dev-manual/api/config.rst
@@ -91,7 +91,8 @@ GEVER-Mandanten abgefragt werden.
               "workspace": false,
               "workspace_client": false
           },
-          "inbox_folder_url": "https://dev.onegovgever.ch/fd/eingangskorb/eingangskorb_afi"
+          "gever_colorization": "#37C35A",
+          "inbox_folder_url": "https://dev.onegovgever.ch/fd/eingangskorb/eingangskorb_afi",
           "is_admin_menu_visible": false,
           "is_emm_environment": false,
           "max_dossier_levels": 5,
@@ -240,6 +241,9 @@ features
 
     workspace_client
         Integration von GEVER mit einem Teamraum
+
+gever_colorization
+    Rahmen Farbe
 
 max_repositoryfolder_levels
     Maximale Verschachtelungstiefe von Ordnungspositionen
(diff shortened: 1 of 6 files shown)


opengever/base/viewlets/colorization.py
```diff
@@ -11,13 +11,17 @@
 ENVIRONMENT_KEY = 'GEVER_COLORIZATION'
 
 
+def get_color():
```
deiferni, line 14: One minor code-organization suggestion: I would not expose this helper function in `viewlets`, but instead put it into a new module `opengever.base.colorization` or something like that. The function is not only used in viewlets as one could think when it is exposed in viewlets but will be used in the api (and possibly other places) as well.

docs/public/dev-manual/api/config.rst
```diff
@@ -241,6 +242,9 @@ features
     workspace_client
         Integration von GEVER mit einem Teamraum
 
+gever_colorization
+    Rahmen Farbe für Test System
```
deiferni, line 246: ~~für Test System~~
it is more generic than only for test things, it is also used for training systems for example.

deiferni (approved):
LGTM 👍, just one suggestion!

deiferni (approved):
Actually one thing more, please add dedicated simple tests for the following permutations:
- no color set in env
- color set in env which is in lookup dict
- color set in env which is not in lookup dict

2e12:
@deiferni Thanks for the feedback. Would you mind taking another look at it? 😄

opengever/base/tests/test_gever_colorization.py
```diff
@@ -0,0 +1,20 @@
+from opengever.base.colorization import get_color
+from opengever.testing import IntegrationTestCase
+import os
+
+
+class TestGeverColorization(IntegrationTestCase):
```
deiferni, line 6: This can subclass `opengever.testing.test_case.TestCase`, the `IntegrationTestCase` builds a fixture with an actual plone and content. You actually need none of this here, as you are just unit-testing the function, this has the advantage that test-execution is much faster locally.

_Unfortunately now it is too late for you, but the next dev will be grateful 🙃._

deiferni (approved):
LGTM apart from suggestion. Feel free to 🚢  once you changed it and tests pass.

PR #6727, opened 2020-11-05
Move docker testing to prevent cyclic imports.

deiferni:
Having docker tests fixtures in opengever.testing caused cyclic imports in policy tests:

```
  File "/dev/opengever.ftw/opengever/ftw/tests/test_harvest_timer.py", line 5, in <module>
    from opengever.ftw.testing import OPENGEVER_FTW_FUNCTIONAL_TESTING
  File "/dev/opengever.ftw/opengever/ftw/testing.py", line 3, in <module>
    from opengever.core.testing import functional_session_factory
  File "/dev/opengever.ftw/src/opengever.core/opengever/core/testing.py", line 28, in <module>
    from opengever.testing.docker import MSGCONVERT_SERVICE_FIXTURE
  File "/dev/opengever.ftw/src/opengever.core/opengever/testing/__init__.py", line 10, in <module>
    from opengever.core.testing import MEMORY_DB_LAYER
ImportError: cannot import name MEMORY_DB_LAYER
```

See https://ci.4teamwork.ch/builds/379615/tasks/651857 for a failing test-run of a policy.

This was introduced very recently in https://github.com/4teamwork/opengever.core/pull/6687. I have refrained from making `opengever.core.testing` a package for this fix but IMO we should think about this in the future as the `foo_testing.py` modules start accumulate in `opengever.core`.

Jira: https://4teamwork.atlassian.net/browse/CA-1067

## Checklist (Must have)

- [ ] Changelog entry _IMO not necessary as it fixes an issue just introduced a few hours ago_


deiferni:
_I will not be around tomorrow, but please feel free to ☠️  and :shipit: this tomorrow, might be that some nightlies will fail because of this issue._

njohner (approved):
LGTM

njohner:
I'll ship, as the policies I'm supposed to update tonight are failing because of this.

PR #6730, opened 2020-11-09
Improve policy generator dynamic defaults.

deiferni:
With this PR we make another baby-step towards improving the policy generator a bit.

We add new defaults that can be guessed from previous input for the following questions:
- bumblebee app id
- ogds db name
- administrator group

We also switch question order, and ask for admin unit id first to generate defaults from the technical admin unit identifier instead from of from the user visible abbreviation. The questions for which defaults are generated are also asking for technical identifier, so that IMO makes more sense.

Jira: https://4teamwork.atlassian.net/browse/CA-1107

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/policytemplates/hooks.py
+++ b/opengever/policytemplates/hooks.py
@@ -144,12 +144,18 @@ def post_package_name(configurator, question, answer):
         answer,
         'onegovgever' if configurator.variables['is_gever'] else 'teamraum'
         )
+    bumblebee_app_id_default = u'{}_{}'.format(
+        'gever' if configurator.variables['is_gever'] else 'teamraum',
+        answer
+        )
     new_defaults = {
         'package.url': 'https://github.com/4teamwork/opengever.{}'.format(
             answer),
         'adminunit.abbreviation': answer,
         'adminunit.id': answer,
         'base.domain': domain_default,
+        'base.ogds_db_name': u'ogds_{}'.format(answer),
+        'base.bumblebee_app_id': bumblebee_app_id_default,
     }
     update_defaults(configurator, new_defaults)
     configurator.variables['package.name_capitalized'] = answer.capitalize()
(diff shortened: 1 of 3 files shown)


opengever/policytemplates/hooks.py
```diff
@@ -144,12 +144,19 @@ def post_package_name(configurator, question, answer):
...
+        )
     new_defaults = {
         'package.url': 'https://github.com/4teamwork/opengever.{}'.format(
             answer),
         'adminunit.abbreviation': answer,
         'adminunit.id': answer,
         'base.domain': domain_default,
+        'base.ogds_db_name': u'ogds_{}'.format(answer),
+        'base.bumblebee_app_id': bumblebee_app_id_default,
+        'deployment.ldap_ou': answer.capitalize(),
```
lukasgraf, line 159: I'm sceptical of this.

This will suggest incorrect OUs with a high rate, which then just will be accepted and can easily lead to a mess that then needs to be fixed, because it needs to correspond to what the OU is *actually* called in LDAP.

For example, this would happily suggest incorrect OUs like `Kwl`, `Kgs`, `Bgbaetterkinden`, `Bistumbasel`, ...

lukasgraf (changes requested):
To be honest, I'm not entirely convinced this is going in the right direction.

My initial approach (years ago) with the ordering of these questions and deriving identifiers from previous questions was to start with the highest fidelity one (the one that contains / retains the most information), and then derive further, "lossy" identifiers from it.

E.g.:
`Dürrenäsch` (highest fidelity, capitalisation and umlauts preserved) -> `duerrenaesch` (lowercased, transliteration applied. Non-reversible information loss).

opengever/policytemplates/hooks.py
deiferni: I went with it because IMO it would get it right 80% of the time, especially for SaaS where we are in direct control of the OU name and usually create it after or while the policy is generated. Also it is "just" a default which can be overridden by manual input. But i understand the potential fear of devs just turning the brain off and pressing enter 🙃 .

So omitting the OU default and leave it to the dev to fill it in would also work here if you're sceptical about this change. Would that work for you? Do you agree with the other two suggestions though?

lukasgraf (approved):

deiferni:
as discussed yesterday, amended. the OU default is out as we want to force devs to think about it.

PR #6731, opened 2020-11-10
Fix an encoding error on the local contacts tab.

deiferni:
We fix an encoding issue that arose due to inconsistent handling of unicode vs bytestrings.

Items passed to the helper methods can be anything from a brain to a solr-document and those do not gurantee unicode strings. Thus we consider these methods to be at the application boundary and make sure we work with unicode internally.

We also refactore the two nearly identical methods and merge them into one method with two shallow wrapper methods setting correct parameters.

Jira: https://4teamwork.atlassian.net/browse/CA-1208

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/contact/browser/contacts_tab.py
+++ b/opengever/contact/browser/contacts_tab.py
@@ -1,6 +1,7 @@
 from opengever.contact import _
 from opengever.tabbedview import BaseCatalogListingTab
 from opengever.tabbedview.helper import email_helper
+from Products.CMFPlone.utils import safe_unicode
 from zope.browserpage.viewpagetemplatefile import ViewPageTemplateFile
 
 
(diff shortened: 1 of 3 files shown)


opengever/contact/browser/contacts_tab.py
```diff
@@ -9,33 +10,39 @@ def authenticated_member(context):
...
 
 
-def linked_no_icon(item, value):
-    url_method = lambda: '#'
+def _render_link(item, value, include_css_class=False):
+    """Render a link to item with value as text.
+
+    Optionally include an additional css class provided by the item.
+
+    Item can be anything from brain to solr-document and those do not gurantee
```
njohner, line 25: gurantee ;-)

opengever/contact/tests/test_contact_tab_helpers.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def test_linked_with_bytestring_value(self):
+        self.assertEqual(
+            u'<span class="linkWrapper"><a href="#">f\xfc\xfc</a></span>',
+            linked(None, u'f\xfc\xfc'.encode('utf-8'))
+        )
+
+    def test_linked_with_bytestring_url(self):
+        item = MockItem()
+        item.getURL = Mock(return_value='http://example.com/foo')
```
njohner, line 27: Hum didn't you want to return something like `return_value=u'f\xfc\xfc'.encode('utf-8')`?

opengever/contact/tests/test_contact_tab_helpers.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def test_linked_no_icon_with_bytestring_value(self):
+        self.assertEqual(
+            u'<span class="linkWrapper"><a href="#">f\xfc\xfc</a></span>',
+            linked_no_icon(None, u'f\xfc\xfc'.encode('utf-8'))
+        )
+
+    def test_linked_no_icon_with_bytestring_url(self):
+        item = MockItem()
+        item.getURL = Mock(return_value='http://example.com/foo')
```
njohner, line 60: same here

njohner (commented):
I'm a bit confused by the tests, otherwise LGTM.

deiferni:
@njohner thanks, amended with your suggestions.

njohner (approved):
LGTM

PR #6735, opened 2020-11-12
A document within a resolved dossier is no longer movable

elioschmutz:
Jira: https://4teamwork.atlassian.net/browse/NE-241

Currently the `@actions` endpoint always returns the `move_items`, `move_proposal_items` and `move_item` actions if a dossier is closed.

This PR:
- updates the missing `available_expr` for the `move_items` and `move_proposal_items` folder buttons
- respects the dossier state in the documents `is_movable` method which will be used for the `available_expr` of documents

## Before
Bildschirmfoto 2020-11-12 um 12 58 04
Bildschirmfoto 2020-11-12 um 12 57 58

## After
Bildschirmfoto 2020-11-12 um 12 55 22
Bildschirmfoto 2020-11-12 um 12 55 12

The old UI handled this case directly in the [MoveItemsFormView](https://github.com/4teamwork/opengever.core/blob/master/opengever/dossier/move_items.py#L264). So the action was always visible, but it raised an error if it was not possible. With this PR, the action will also no longer be visible in the old ui. The guard still needs to be there because the view can still be accessed directly by modifying the url manually.

It's still possible to move documents out of closed dossiers through the restapis `@move` enpiont. This issue will be handled in a different story: https://4teamwork.atlassian.net/browse/NE-291?searchSessionId=45322838-a049-4f57-be46-737d26b20c9a&searchObjectId=15240&searchContainerId=10023&searchContentType=issue

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations


docs/HISTORY.txt
```diff
@@ -4,6 +4,7 @@ Changelog
 2020.14.0 (unreleased)
 ----------------------
 
+- A document within a closed dossier is no longer movable [elioschmutz]
```
njohner, line 8: I think:
- Disable action to move document within a closed dossier. 

would be closer to the truth, as the document as such is still movable (over the API notably, according to your description in the PR)

opengever/document/base.py
```diff
@@ -147,8 +147,18 @@ def is_inside_a_task(self):
         parent = aq_parent(aq_inner(self))
         return ITask.providedBy(parent)
 
+    def is_inside_a_closed_dossier(self):
+        dossier = get_main_dossier(self)
```
njohner, line 151: Hum that seems wrong. A document in a closed subdossier but in an open Main dossier should not be movable...

njohner: Please also add a test for that

opengever/core/upgrades/20201112123011_add_available_expr_for_move_items_folder_actions/upgrade.py
```diff
@@ -0,0 +1,9 @@
+from ftw.upgrade import UpgradeStep
+
+
+class AddAvailableExprForMoveItemsFolderActions(UpgradeStep):
```
njohner, line 4: FolderButtons?

njohner (commented):
I think your check for having the action or not is not correct, see my comment.

Apart from that LGTM.

opengever/document/base.py
elioschmutz: 👍 thats true 👀

I was not aware of the ability to close a subdossier

opengever/document/base.py
```diff
@@ -147,8 +147,18 @@ def is_inside_a_task(self):
         parent = aq_parent(aq_inner(self))
         return ITask.providedBy(parent)
 
+    def is_inside_a_closed_dossier(self):
+        dossier = self.get_parent_dossier()
```
njohner, line 151: I think for a document in an inbox, `self.get_parent_dossier()` will return `None`, so `IDossierMarker.providedBy(dossier)` will be `False` so `is_inside_a_closed_dossier` will return `False` so the document will be movable 😅 . So I think this is OK now, but I'm a bit scared. Can you add a test for a document inside an inbox?

njohner (approved):
I think it's ok. Can you just make sure that documents inside an inbox are still movable?

Apart from that, LGTM

opengever/document/base.py
elioschmutz: 🙈 oh man! I'll add such a test 👍

njohner: Merci!

PR #6736, opened 2020-11-12
Add simple support for meetings in a multi-admin-unit cluster.

deiferni:
With this PR we enable support for `opengever.meeting` in a cluster with multiple admin-units. No cross admin-unit communication is supported yet. But multiple separate deployments can now have `opengever.meeting` activated individually in the same cluster.

We make sure that:
- only local committees are selectable, where appropriate
- make sure that members are linked to an admin-unit, and only local members are displayed on each admin-unit

Notes/thoughts about the upgrade to add admin-unit to members:
-  We apply a graceful strategy in case we find a deployment in invalid state. Instead of failing hard and blocking a potential upgrade path we migrate using a placeholder and notify that manual cleanup is needed. We don't expect any deployments in such state as it is currently not supported.
- Sample sentry messages are provided in:
  - https://sentry.4teamwork.ch/organizations/sentry/issues/69805/?project=17
  - https://sentry.4teamwork.ch/organizations/sentry/issues/69804/?project=17
- There is no foreign-key constraint to the admin-unit, this is consistent with `Committee` and purposefully omitted with future servicification in mind. See: https://github.com/4teamwork/opengever.core/blob/73af257f462f1e3268cc7b9a745fd6f870e77222/opengever/meeting/model/committee.py#L55

Jira: https://4teamwork.atlassian.net/browse/CA-1116

- [ ] 🚧 &nbsp; once this PR is merged, remove the sample sentry issues

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- DB-Schema migration

### DIFF
--- a/opengever/core/upgrades/20201111125010_add_admin_unit_to_member/upgrade.py
+++ b/opengever/core/upgrades/20201111125010_add_admin_unit_to_member/upgrade.py
@@ -1,0 +1,102 @@
+from opengever.base.sentry import log_msg_to_sentry
+from opengever.core.upgrade import SchemaMigration
+from opengever.ogds.base.utils import get_current_admin_unit
+from plone import api
+from sqlalchemy import Column
+from sqlalchemy import String
+from sqlalchemy.sql import select
+from sqlalchemy.sql.expression import column
+from sqlalchemy.sql.expression import table
+
+
+UNIT_ID_LENGTH = 30
+
+
+members_table = table(
+    "members",
+    column("id"),
+    column("admin_unit_id"),
+)
+
+
+committees_table = table(
+    "committees",
+    column("id"),
+    column("admin_unit_id"),
+    column("int_id"),
+)
+
+
+class AddAdminUnitToMember(SchemaMigration):
+    """Add admin_unit column to Member.
+
+    We apply a graceful strategy in case we find a deployment in invalid state.
+    Instead of failing hard and blocking a potential upgrade path we migrate
+    using a placeholder and notify that manual cleanup is needed. We don't
+    expect any deployments in such state as it is currently not supported.
+    """
+
+    def migrate(self):
+        self.add_admin_unit_column()
+        self.migrate_data()
+        self.make_admin_unit_column_non_nullable()
+
+    def add_admin_unit_column(self):
+        self.op.add_column(
+            'members',
+            Column('admin_unit_id', String(UNIT_ID_LENGTH), nullable=True)
+        )
+
+    def has_committees_for_multiple_admin_units(self):
+        statement = select([committees_table.c.admin_unit_id]).distinct()
+        results = list(self.execute(statement))
+        return len(results) > 1
+
+    def migrate_data(self):
+        nof_members = self.execute(members_table.count()).scalar()
+        # there are no members, this means that either the meeting feature
+        # is not enabled, or not in use. no data migration is needed.
+        if nof_members <= 0:
+            return
+
+        current_admin_unit_id = get_current_admin_unit().id()
+
+        has_meeting_feature = api.portal.get_registry_record(
+            'opengever.meeting.interfaces.IMeetingSettings.is_feature_enabled')
+        # if we don't have the meeting feature active this means that another
+        # admin-unit in the cluster has the meeting feature, but we are not
+        # running schema migrations on that admin-unit.
+        # this should not be the case at the moment as it is not yet officially
+        # supported.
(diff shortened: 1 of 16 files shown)


deiferni:
_for the potential reviewer: the tests should pass again eventually we seem to have trouble with the new jenkins slave at the moment._

opengever/meeting/tests/test_committee_query.py
```diff
@@ -0,0 +1,34 @@
...
+            Builder('committee_model')
+            .having(int_id=5678, admin_unit_id=u'local', title=u'Local')
+        )
+        self.remote_admin_unit = create(
+            Builder('admin_unit')
+            .having(unit_id=u'remote')
+        )
+        self.remote_committee = create(
+            Builder('committee_model')
+            .having(int_id=5678, admin_unit_id=u'remote', title=u'Remote')
```
njohner, line 27: Are you using the same `int_id` as for `self.local_committee` on purpose?

opengever/core/upgrades/20201111125010_add_admin_unit_to_member/upgrade.py
```diff
@@ -0,0 +1,102 @@
...
+        has_meeting_feature = api.portal.get_registry_record(
+            'opengever.meeting.interfaces.IMeetingSettings.is_feature_enabled')
+        # if we don't have the meeting feature active this means that another
+        # admin-unit in the cluster has the meeting feature, but we are not
+        # running schema migrations on that admin-unit.
+        # this should not be the case at the moment as it is not yet officially
+        # supported.
+        # we gracefully continue with the upgrade and notify that manual
+        # cleanup is needed.
+        if not has_meeting_feature:
```
njohner, line 73: Why can't we simply determine the correct admin unit from the committees table and use that for the members?

opengever/meeting/model/member.py
```diff
@@ -3,20 +3,28 @@
...
+    return get_current_admin_unit().id()
+
+
 class Member(Base, SQLFormSupport):
 
     __tablename__ = 'members'
 
     member_id = Column("id", Integer, Sequence("member_id_seq"),
                        primary_key=True)
+    admin_unit_id = Column(String(UNIT_ID_LENGTH), nullable=False,
```
njohner, line 26: I'm a  bit confused right now as to why a member cannot be shared between several `admin_units`? Is it simply to keep the number of members lower for each admin_unit?

opengever/meeting/tests/test_members.py
```diff
@@ -109,6 +109,19 @@ def test_member_link(self):
         self.assertEqual(u'M\xfcller Peter', link.get('title'))
         self.assertEqual(u'M\xfcller Peter', link.text)
 
+    @browsing
+    def test_only_local_members_are_listed(self, browser):
+        create(Builder('member').having(admin_unit_id='foreign'))
```
njohner, line 114: Can you give that one a different name as `Peter Müller`, so that it's a bit clearer that it is not returned in the listing?

njohner (approved):
LGTM, apart from comments of course :-D!

And a typo in the last commit message:
Make sure exmaple content members have admin unit. -> Make sure example content members have an admin unit.

opengever/meeting/tests/test_committee_query.py
deiferni: nope, copy-paste. have changed it to clarify that this is a completely different committee.

opengever/core/upgrades/20201111125010_add_admin_unit_to_member/upgrade.py
deiferni: @njohner now i remember why i did not implement it that way:
a deployment could theoretically have members, but no committees. if we switch the implementation we would have to handle the case of no committees but with members. This would then somehow have to fall back to getting the current admin-unit in that case.

IMO this makes the upgrade needlessly complicated. I could add the option to guess the admin_unit from the committees, but this would only be additional to what we have at the moment, i could not simplify the code and always use the committees, as they could not be present at all. And i could also not get rid of the current safeguards that exist at the moment as we still need the fallback to the current admin_unit for the case that there are no committees.

Also this is something which is AFAIK not the case in production and/or staging anywhere as it is a new deployment option which will only be enabled after this PR has been merged.
Thus, if you agree, I'd propose to leave it as is and should we unexpectedly encounter this somewhere we just fix it manually.

njohner (approved):
LGTM

PR #6737, opened 2020-11-13
Don't save notification setting with default values as custom notification setting

tinagerber:
Until now a notification setting with the same values as NotificationDefault could be saved as NotificationSetting. Now the notification setting is deleted if it matches the NotificationDefault.

Jira: https://4teamwork.atlassian.net/browse/NE-99

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.14.0 (unreleased)
 ----------------------
 
+- Don't save notification setting with default values as custom notification setting. [tinagerber]
 - Add @notification-settings API endpoint. [tinagerber]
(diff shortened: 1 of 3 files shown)


lukasgraf:
What issue does this solve? Why is it a problem if an explixit, custom setting can be stored that happens to be the same value as the (current) default?

This basically introduces the exact same bug that we desperately tried to fix in Plone with all the patches in [`opengever.base.monkey.patches.default_values`](https://github.com/4teamwork/opengever.core/blob/master/opengever/base/monkey/patches/default_values.py).

Many of these patches disable incorrect logic like this:
```
if not value == default:
    field.save(value)
```

which would lead to values never getting persisted if they happen to be identical to the current default (which can change in the future).

PR #6739, opened 2020-11-16
Use a dedicated endpoint to upload document copy to workspace

lukasgraf:
This change introduces a dedicated `@upload-document-copy` endpoint for copying documents to Teamraum. 

In order to prepare for also storing metadata about linked documents, the existing use of TUS upload is replaced by a dedicated endpoint on workspaces that accepts everything needed to create a document copy in one single request.

The implementation using TUS was already using three spin-off requests to the remote system during a running transaction:
- Preparation of TUS upload
- TUS upload of the blob
- A PATCH to set metadata

This would have had to be extended by a fourth one to set link metadata for linking the document copy to the GEVER document. Therefore the TUS upload approach had to be abandoned by a more purpose-built endpoint.

Tests for this are problematic: Because some tests run on a functional ZServer layer, they are prone to conflicts: During the test run, a request is made to the GEVER side view that in turn issues another request to the remote side. This last request is issued from a running transaction, and starts a new transaction that may easily conflict (or deadlock) with the other transaction.

The entire testing around Teamraum Connect therefore needs a serious overhaul. For now I only could beat it into submission enough that it's reasonably stable (shouldn't be flaky on Jenkins), and documented.

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
        _(Not deemed necessary - internal endpoint not intended for public use)_
- New functionality:

### DIFF
--- a/opengever/api/workspace.py
+++ b/opengever/api/workspace.py
@@ -1,11 +1,16 @@
+from ftw.mail.mail import IMail
+from opengever.api.add import GeverFolderPost
 from opengever.api.serializer import GeverSerializeFolderToJson
+from opengever.document.document import IDocumentSchema
 from opengever.ogds.base.actor import Actor
 from opengever.workspace.interfaces import IWorkspace
 from opengever.workspace.participation import can_manage_member
 from plone.restapi.interfaces import ISerializeToJson
+from zExceptions import BadRequest
 from zope.component import adapter
 from zope.interface import implementer
 from zope.interface import Interface
+import json
 
 
 @implementer(ISerializeToJson)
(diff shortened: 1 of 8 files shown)


opengever/api/workspace.py
```diff
@@ -26,3 +42,94 @@ def __call__(self, *args, **kwargs):
...
+    - portal_type
+
+    The value in the form field `document_metadata` is expected to be a
+    JSON encoded string with the document's metadata fields in the
+    ISerializeToJson serialization format.
+
+    This implementation is loosely based on p.restapi.services.content.add.
+    """
+
+    def reply(self):
```
njohner, line 65: Would it be possible to reuse `FolderPost` here (https://github.com/4teamwork/opengever.core/blob/master/opengever/api/add.py), or for `UploadDocumentCopy` to inherit from it? We have our own version of `FolderPost` which should allow some more flexibility for customisation. There seems to be quite a bit of code duplication, which also makes me wonder whether all paths we have to create a document will lead to the same thing in the end (`CreateDocumentCommand`, `FolderPost`, `UploadDocumentCopy`, ...).

njohner (commented):
I have to say that I'm not particularly fan of introducing another separate code path to create documents. I feel like either `FolderPost` or `CreateDocumentCommand` should be at least partly reusable here. But maybe I'm missing something?
Otherwise LGTM

PR #6740, opened 2020-11-17
No longer exclude trashed documents in @listing-stats endpoint

tinagerber:
In order for the frontend to show how many documents are in the trash, it must be possible to query this. Therefore, the default filter "trashed:false" must no longer be used in @listing-stats endpoint.

Jira: https://4teamwork.atlassian.net/browse/NE-60

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [x] api-change label added

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 ----------------------
 
 - Use a dedicated endpoint to upload document copy to workspace. [lgraf]
+- No longer exclude trashed documents in @listing-stats endpoint. [tinagerber]
 - Add @notification-settings API endpoint. [tinagerber]
 - Use UID instead of intId as token in DocumentTemplatesVocabulary. [elioschmutz]
 - Add simple support for meetings in a multi-admin-unit cluster. [deiferni]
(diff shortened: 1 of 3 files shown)


docs/HISTORY.txt
```diff
@@ -5,6 +5,7 @@ Changelog
 ----------------------
 
 - Use a dedicated endpoint to upload document copy to workspace. [lgraf]
+- No longer count only documents that are not in the trash by default in @listing-stats endpoint. [tinagerber]
```
njohner, line 8: That's a lot of negations 😅 . How about "No longer exclude trashed documents in @listing-stats endpoint."

njohner (commented):
I think it bothers me that `trashed` documents are excluded by default in the `listing` endpoint but not in the `listing-stats` endpoint.
Wouldn't it be more consistent to exclude trashed documents by default except if `trashed:true` is passed as query? Or is that a problem? I'm not sure how the frontend uses that endpoint...

tinagerber:
@njohner I have also thought back and forth about what is better. But then there would be a different total number of documents, depending on whether "trashed:true" is queried or not. I would find this a bit confusing.

njohner:
Of course, I didn't quite remember how queries worked in that case... they only add information but do not filter results, so LGTM.

njohner (approved):

PR #6741, opened 2020-11-17
Link documents copied via Teamraum Connect

lukasgraf:
This links GEVER documents and their copies in Teamraum when they are copied using Teamraum Connect. 

_(This is motivated by later using these links to transfer back copied documents from Teamraum to GEVER as a new version of the original document)._

These links are internally stored in annotations on the documents on both sides, and are serialized in a document GET like so:

**On a GEVER document** (links to multiple TR documents possible):

```json
{
  "teamraum_connect_links": {
    "gever_document": null,
    "workspace_documents": [
      {
        "UID": "a8519078dc2a4b39bca2ab530b817376",
        "UID": "a7ea5e0fb1544a4bb513820210589d55",
      }
    ]
  },
  "...": "..."
}
```

**On a Teamraum document**:

```json
{
  "teamraum_connect_links": {
    "gever_document": {
      "UID": "cafde44103ae4c73b683321bed7bd054"
    },
    "workspace_documents": []
  },
  "...": "..."
}
```

Copied documents without a file are not linked, since there's no file that could ever be transferred back to GEVER.

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
- New functionality:

### DIFF
--- a/opengever/workspaceclient/linked_workspaces.py
+++ b/opengever/workspaceclient/linked_workspaces.py
@@ -5,6 +5,7 @@ from opengever.workspaceclient import _
 from opengever.workspaceclient.client import WorkspaceClient
 from opengever.workspaceclient.exceptions import WorkspaceNotFound
 from opengever.workspaceclient.exceptions import WorkspaceNotLinked
+from opengever.workspaceclient.interfaces import ILinkedDocuments
 from opengever.workspaceclient.interfaces import ILinkedWorkspaces
 from opengever.workspaceclient.storage import LinkedWorkspacesStorage
 from plone import api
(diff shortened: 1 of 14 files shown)


opengever/workspaceclient/linked_workspaces.py
```diff
@@ -148,12 +149,13 @@ def copy_document_to_workspace(self, document, workspace_uid):
         document_metadata = self._drop_non_metadata_fields(document_repr)
 
         if not file_:
```
deiferni, line 151: this is ouf of scope but just for the record i'm a bit puzzled why we allow them to be copied in the first place 🤷

deiferni (approved):
LGTM 👍 

i'd be glad if we can pick up and maybe discuss the testing issues that surfaced here eventually. IMO nothing that blocks :shipit: though.

PR #6743, opened 2020-11-17
Prevent checked out documents from being copied to or from a teamraum

lukasgraf:
This will prevent checked out documents from being copied _to_ a Teamraum, or _copied back_ from Teamraum to GEVER.

The individual requests for the respective documents will be answered with 400 Bad Request, and the document will therefore be flagged as "failed" in the document operation queue  (bottom right) by the frontend.

Jira: https://4teamwork.atlassian.net/browse/CA-1218

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- New functionality:

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.14.0 (unreleased)
 ----------------------
 
+- Prevent documents from being copied to workspace when checked out. [lgraf]
 - Link documents copied via Teamraum Connect. [lgraf]
 - Use a dedicated endpoint to upload document copy to workspace. [lgraf]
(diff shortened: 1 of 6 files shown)


docs/HISTORY.txt
```diff
@@ -4,6 +4,7 @@ Changelog
 2020.14.0 (unreleased)
 ----------------------
 
+- Prevent documents from being copied to workspace when checked out. [lgraf]
```
deiferni, line 7: to workspace -> to and from workspaces?

deiferni (approved):
LGTM 👍 

just one optional clarification in the CL, if you want to mention both sides.

PR #6744, opened 2020-11-19
Translate proposal review state and tranform issuer label

tinagerber:
For the search in the frontend there is a new tab for proposals. It should be possible to filter by review state and issuer. 
Therefore the following things have to be changed:
- For Issuer the actor label must be provided
- The review states of proposals must be translated

Jira: https://4teamwork.atlassian.net/browse/NE-211

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/api/solr_query_service.py
+++ b/opengever/api/solr_query_service.py
@@ -290,6 +290,7 @@ FIELDS_WITH_MAPPING = [
     ListingField('document_type', 'document_type', transform=translate_document_type),
     ListingField('filename', 'filename', filename),
     ListingField('filesize', 'filesize', filesize),
+    ListingField('issuer', 'issuer', transform=display_name),
     ListingField('issuer_fullname', 'issuer', 'issuer_fullname'),
     ListingField('keywords', 'Subject'),
     ListingField('mimetype', 'getContentType', sort_index='mimetype'),
--- a/opengever/base/locales/de/LC_MESSAGES/plone.po
+++ b/opengever/base/locales/de/LC_MESSAGES/plone.po
@@ -270,6 +270,21 @@ msgstr "Löschen"
 msgid "mail-transition-restore"
 msgstr "Wiederherstellen"
 
+msgid "proposal-state-active"
+msgstr "In Bearbeitung"
+
+msgid "proposal-state-cancelled"
+msgstr "Storniert"
+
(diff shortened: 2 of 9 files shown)


opengever/base/locales/de/LC_MESSAGES/plone.po
```diff
@@ -270,6 +270,21 @@ msgstr "Löschen"
 msgid "mail-transition-restore"
 msgstr "Wiederherstellen"
 
+msgid "proposal-state-active"
+msgstr "In Bearbeitung"
+
+msgid "proposal-state-cancelled"
+msgstr "Abgebrochen"
```
njohner, line 277: -> Storniert
The translations for the proposals already exist, they are just a bit hidden: https://github.com/4teamwork/opengever.core/blob/master/opengever/meeting/locales/de/LC_MESSAGES/opengever.meeting.po#L454-L458

opengever/api/solr_query_service.py
```diff
@@ -290,6 +290,7 @@ def hide_facet(self, facet):
     ListingField('document_type', 'document_type', transform=translate_document_type),
     ListingField('filename', 'filename', filename),
     ListingField('filesize', 'filesize', filesize),
+    ListingField('issuer', 'issuer', transform=display_name),
```
njohner, line 293: Wouldn't it be time to return the `actor-id` for new fields instead of the label?

njohner (approved):
LGTM

opengever/base/locales/de/LC_MESSAGES/plone.po
tinagerber: Ah, I didn't see that. Then they should probably be changed in `gever-ui` too

opengever/api/solr_query_service.py
tinagerber: We get something like that with the facets:
```
"facet_counts": {
    "niklaus.johner": {
      "administrator": {
        "count": 52,
        "label": "Johner Niklaus"
      },
      "tina.gerber": {
        "count": 42,
        "label": "Gerber Tina"
      },
      ....
}
```
So I would prefer the label.

njohner: I mean it's fine for now, but I think that we will change all those to `actorid`s in the future.

PR #6745, opened 2020-11-20
Fix persistence bug in linked documents storage

lukasgraf:
Before this fix, the following could happen:
- A property like `storage.linked_workspace_documents` was accessed on an uninitialized storage, as part of a display operation
- This lead to a regular, non-persisted dict being added to annotations because of the use of `setdefault`
- That dict then got committed as part of the transaction
- Subsequent write operations on the storage then picked up that non-persistent dict to add to it. Those changes then would never be persisted.

The fix therefore makes sure that `storage._link_storage(persistent=False)` never *stores* the empty, non-persistent dict, but instead just _returns_ it.

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.15.0 (unreleased)
 ----------------------
 
+- Fix persistence bug in linked documents storage. [lgraf]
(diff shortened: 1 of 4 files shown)


njohner (approved):
LGTM

PR #6750, opened 2020-11-26
Add a usersnap-api key question to the policy-generator

elioschmutz:
This PR extends the policy generator with adds a question to add a Usersnap API key to the policy.

Adding a key will automatically enable usersnap on the deployment.


njohner (commented):
I think we could simply ask whether we want to enable usersnap. The key itself has only two possible values, one for Teamraum and one for Gever, so I think we can set it automatically.

elioschmutz:
@njohner I thought to not make the api-keys public in such a clear way...

njohner (approved):

PR #6751, opened 2020-11-30
Support transferring documents from workspace back to GEVER as new version.

lukasgraf:
This adds support for transferring documents from teamraum back to GEVER **as a new version**.

The `@copy-document-from-workspace` endpoint accepts a new flag `as_new_version` (default: False), which will retrieve the document as a new version of the linked document, if possible. If not possible (e.g. if the document wasn't linked, doesn't have a file or is a mail), it will pick to create a _copy_ instead though). The endpoint will return the creation mode it picked as a property `teamraum_connect_retrieval_mode`(`copy` or `version`) in the serialized object in the response.

When retrieving a document as a new version, the original document will be **unlocked** (if locked). When retrieving as a copy, it **won't**.

Jira: https://4teamwork.atlassian.net/browse/CA-1222

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
- New functionality:

- New translations

### DIFF
--- a/opengever/api/tests/test_linked_workspaces.py
+++ b/opengever/api/tests/test_linked_workspaces.py
@@ -11,6 +11,8 @@ from opengever.testing.assets import load
 from opengever.workspaceclient.exceptions import WorkspaceNotLinked
 from opengever.workspaceclient.interfaces import ILinkedDocuments
 from opengever.workspaceclient.interfaces import ILinkedWorkspaces
+from opengever.workspaceclient.linked_workspaces import RETRIEVAL_MODE_COPY
+from opengever.workspaceclient.linked_workspaces import RETRIEVAL_MODE_VERSION
 from opengever.workspaceclient.storage import LinkedWorkspacesStorage
 from opengever.workspaceclient.tests import FunctionalWorkspaceClientTestCase
 from plone import api
@@ -886,6 +888,9 @@ class TestCopyDocumentFromWorkspacePost(FunctionalWorkspaceClientTestCase):
             self.assertEqual(len(children['added']), 1)
             document_copy = children['added'].pop()
             self.assertEqual(document_copy.absolute_url(), browser.json.get('@id'))
+            self.assertEqual(
+                RETRIEVAL_MODE_COPY,
+                browser.json.get('teamraum_connect_retrieval_mode'))
 
     @browsing
     def test_copy_document_with_file_from_a_workspace(self, browser):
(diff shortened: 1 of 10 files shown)


opengever/api/tests/test_linked_workspaces.py
```diff
@@ -858,7 +860,7 @@ def test_raises_when_document_is_not_within_linked_workspace(self, browser):
         self.assertEqual("Document not in linked workspace", str(cm.exception))
 
     @browsing
-    def test_copy_document_without_file_from_workspace(self, browser):
+    def test_copy_document_without_file_from_workspace_xxx(self, browser):
```
njohner, line 863: ✂️ `_xxx`

opengever/workspaceclient/tests/test_linked_workspaces.py
```diff
@@ -584,6 +584,36 @@ def test_copy_document_without_file_from_a_workspace(self):
...
+        """
+        workspace_doc = create(Builder('document')
+                               .within(self.workspace))
+
+        gever_doc = create(Builder('document')
+                           .within(self.dossier))
+
+        # Documents without a file are never linked by the current
+        # implementation. But even if they were, no attempt at creating a
+        # version should be made.
```
njohner, line 652: that would also hold true for when there was a file when the document was copied to the workspace, and then the file was deleted on the workspace document.

njohner (approved):
A few comments, but LGTM.

opengever/workspaceclient/tests/test_linked_workspaces.py
lukasgraf: I didn't even think of that 😉 , but it should, yes. I tried to state the condition in a positive way (just a regular document with a file, linked) to catch as many odd corner cases that we might not think of yet.

opengever/workspaceclient/linked_workspaces.py
```diff
@@ -244,10 +246,30 @@ def copy_document_from_workspace(self, workspace_uid, document_uid):
...
             title=title)
 
-        proxy_post = ProxyPost(document_repr)
-        proxy_post.context = self.context
-        proxy_post.request = getRequest()
-        return proxy_post.reply()
+        if as_new_version:
+            gever_doc_uid = document_repr['teamraum_connect_links']['gever_document']['UID']
+            catalog = api.portal.get_tool('portal_catalog')
+            gever_doc = catalog(UID=gever_doc_uid)[0].getObject()
```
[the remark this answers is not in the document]
lukasgraf: @njohner yes, it will fail. As it should, IMHO. In fact, you bring up a good point - I think we need some additional permission checks here, to make sure the user actualy has `Modify portal content` and the permission to create a new version, or the necessary add permission on the dossier in the `copy` case. I'll look into those.

gever_doc linked to workspace document: I considered that, but it's a bit unclear how we would even deal with that. It's not really necessary to perform the intended operation, so it would seem a bit harsh to then just reject the operation, when it could have worked. We could log, I guess, .

PR #6752, opened 2020-11-30
Add @link-to-workspace and teamraum-solrsearch endpoints

tinagerber:
Changes in this PR:
- To make linked dossiers searchable by a query, the dossier now provides an interface when it is linked to a workspace. Since workspaces and dossiers are on the same deployment in the tests, changing a dossier and a workspace at the same time would cause a database conflict error, so this cannot be tested.
- In order for a workspace to know whether it is linked to a dossier, the `Oguid` of the linked dossier is set as an external reference as soon as the workspace is linked.
- New endpoint `@link-to-workspace` added to link dossiers to existing workspaces
- Added new endpoint `@teamraum-solrsearch` to make solr searches from a GEVER installation to the connected teamraum installation. This is necessary in order to be able to suggest to the dossier for linking all workspaces that are not yet linked. Since the WorkspaceClient feature can only be tested through functional test cases, and Solr is required for this endpoint, this endpoint cannot be tested in a meaningful way.

Since GEVER and Teamraum are two separate installations, the modifications made here (external_reference, ILinkedToWorkspace interface) cannot be migrated automatically. Therefore two migration scripts were written for this purpose. https://github.com/4teamwork/opengever.maintenance/pull/233

Discussed with @deiferni and @njohner 

Jira: https://4teamwork.atlassian.net/browse/NE-100

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
- New translations


opengever/api/linked_workspaces.py
```diff
@@ -156,6 +156,23 @@ def reply(self):
...
+        # Disable CSRF protection
+        alsoProvides(self.request, IDisableCSRFProtection)
+        data = json_body(self.request)
+        workspace_uid = data.get('workspace_uid')
+        if not workspace_uid:
+            raise BadRequest("Property 'workspace_uid' is required")
+
+        ILinkedWorkspaces(self.context).link_to_workspace(workspace_uid)
+        self.request.response.setStatus(204)
+        return super(LinkToWorkspacePost, self).reply()
```
lukasgraf, line 173: You can simply use `return self.reply_no_content()` here (defined [in plone.restapi](https://github.com/plone/plone.restapi/blob/aed5e24abb9977935b4befebeb4c101b69dc88ea/src/plone/restapi/services/__init__.py#L38-L40))

This will return a no content marker which will be serialized as actually no content (empty string) instead of `None`, which will be turned into JSON `null`. You can then also omit the `setStatus(204)`, which it sets by default.

opengever/workspaceclient/locales/de/LC_MESSAGES/opengever.workspaceclient.po
```diff
@@ -28,3 +28,8 @@ msgstr "Dokument ${doc_title} in Teamraum ${workspace_title} kopiert."
 #: ./opengever/workspaceclient/linked_workspaces.py
 msgid "label_linked_workspace_created"
 msgstr "Teamraum ${workspace_title} zu diesem Dossier eröffnet."
+
+#. Default: "Linked to workspace ${workspace_title}."
+#: ./opengever/workspaceclient/linked_workspaces.py
+msgid "label_workspace_linked"
+msgstr "Mit Temraum ${workspace_title} verknüpft."
```
lukasgraf, line 35: Temraum -> Teamraum

lukasgraf (changes requested):
LGTM 👍 
Just a few minor points.

njohner (commented):
I didn't look at the code, but if I understood correctly we will need to run the maintenance scripts for that update. So don't forget to add it in the list of version specific things we need to do: https://4teamwork.atlassian.net/wiki/spaces/4TEAM/pages/123077675/GEVER+Releases+Versions-Spezifische+Punkte

lukasgraf (approved):

PR #5917, opened 2019-09-04
Integrate ToDos in GEVER-Notification-System

njohner:
We integrate `ToDo`s in the notification system. For that we:
* Add new activity settings for workspaces for `TODO_RESPONSIBLE_ROLE` and `WORKSPACE_MEMBER_ROLE`
* Add activities for assigning, closing, reopening and commenting a todo.
* Automatically update watchers on the `ToDo`s outside from the activities, in prevision of the view listing all objects that user is watching:
  * Add all workspace users as watchers to a newly created ToDo.
  * Add new workspace participant as watcher on all ToDos.

For https://github.com/4teamwork/opengever.core/issues/5796

## Checkliste
- [ ] Sind UpgradeSteps `deferrable`, oder können gewisse Schritte des Upgrades konditional ausgeführt werden? Upgrade step is not deferrable.
- [x] Gibts es eine DB-Schema Migration?
  - [x] Wurde alle Columns/Änderungen aus dem Modell in einer DB-Schema Migration nachgeführt.
  - [x] Sind Constraint-Namen maximal 30 Zeichen lang (`Oracle`)?
- [x] Gibt es neue Übersetzungen?
  - [x] Sind alle msg-Strings in Übersetzungen Unicode?
  - [x] Wird die richtige i18n-domain verwendet (Copy-Paste Fehler sind hier häufig)?
- [x] Changelog-Eintrag vorhanden/nötig?
- [x] Aktualisierung Dokumentation vorhanden/nötig?

PR #6753, opened 2020-11-30
fix todo watchers

njohner:
With https://github.com/4teamwork/opengever.core/pull/5917 we included ToDos in the Gever notification system so that workspace participants get notified about modifications on ToDos. There was a bug in the implementation, so that the new participant would get added to all ToDos instead (https://github.com/4teamwork/opengever.core/blob/d42390cc1380477d774fd1b32b2e75bc72beb808/opengever/workspace/activities.py#L44), also the ones he had no permission to see when he accepted the invitation over the browser view which uses `elevated_privileges` (https://github.com/4teamwork/opengever.core/blob/4a84462c2872b06db0595ed34ac7908cd1682e40/opengever/workspace/participation/browser/my_invitations.py#L222).

The good news is that there was a second error in the implementation so that the newly added participants were added as watchers with their workspace participation role (so one of  `WorkspaceGuest`, `WorkspaceMember`, `WorkspaceAdmin`) instead of the corresponding activity role `WORKSPACE_MEMBER_ROLE`. This makes the faulty `Subscription`s easy to find and delete.

Also as there is no activity setting for the `WorkspaceGuest`, `WorkspaceMember`, `WorkspaceAdmin` roles, such notifications will never get dispatched in the classic UI, but because the new frontend does not respect the user's activity settings and simply displays all notifications for a given user (the information is not even serialized: https://github.com/4teamwork/opengever.core/blob/master/opengever/activity/model/notification.py#L36-L45), these notifications would show up in the new UI.

So we did the following here:
- Only add a new participant as watcher on the `ToDo`s of the workspace he was added to
- Use correct role when adding a participant as watcher on the `ToDo`s
- Upgrade step to delete the incorrect subscriptions and re-add all current participants of the Workspace as watchers on the `ToDo`s

Follow-up stories:
- For the Problem in the Gever-UI: https://4teamwork.atlassian.net/browse/NE-378
- ActivityCenter should not manage subscriptions IMO: https://4teamwork.atlassian.net/browse/CA-1310

For https://4teamwork.atlassian.net/browse/CA-1297

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/core/upgrades/20201127160828_correct_to_do_watchers/upgrade.py
+++ b/opengever/core/upgrades/20201127160828_correct_to_do_watchers/upgrade.py
@@ -1,0 +1,57 @@
+from ftw.upgrade import UpgradeStep
+from opengever.activity import notification_center
+from opengever.activity.model import Subscription
+from opengever.activity.roles import WORKSPACE_MEMBER_ROLE
+from opengever.base.model import create_session
+from opengever.workspace.participation.browser.manage_participants import ManageParticipants
+from opengever.workspace.todo import ToDo
+from zope.globalrequest import getRequest
+import logging
+
+
+logger = logging.getLogger('ftw.upgrade')
+
+
+class CorrectToDoWatchers(UpgradeStep):
+    """Correct to do watchers.
+    """
+
+    deferrable = True
+
+    def __call__(self):
+        self.center = notification_center()
+        self.session = create_session()
+
+        # Because WorkspaceWatcherManager.new_participant_added not only added
+        # the new participant as watcher to all ToDos but also with the
+        # participants'role in the workspace instead of WORKSPACE_MEMBER_ROLE ,
+        # as role, we can identify the wrongly added subscriptions by their role.
+        roles = ["WorkspaceGuest", "WorkspaceMember", "WorkspaceAdmin"]
+        subscriptions = Subscription.query.filter(Subscription.role.in_(roles))
+
+        todos = set()
+        for subscription in subscriptions:
+            resource = subscription.resource
+            obj = resource.oguid.resolve_object()
+            if not isinstance(obj, ToDo):
+                logger.warning(
+                    'Subscription {} is not on a ToDo. '
+                    'Skipping.'.format(subscription))
+                continue
+            self.session.delete(subscription)
+            todos.add(obj)
+
+        # Now we make sure that all workspace participants are watchers from
+        # the todos in the workspace.
+        for todo in todos:
+            self._add_all_workspace_users_as_watchers(todo)
+
+    def _add_all_workspace_users_as_watchers(self, todo):
+        """Copy from WorkspaceWatcherManager
+        """
+        workspace = todo.get_containing_workspace()
+        manager = ManageParticipants(workspace, getRequest())
+
+        for member in manager.get_participants():
+            self.center.add_watcher_to_resource(
+                todo, member["userid"], WORKSPACE_MEMBER_ROLE)
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.15.0 (unreleased)
 ----------------------
 
+- Correct bug with watchers being wrongfully added to ToDos. [njohner]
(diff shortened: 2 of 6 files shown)


opengever/core/upgrades/20201127160828_correct_to_do_watchers/upgrade.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class CorrectToDoWatchers(UpgradeStep):
+    """Correct to do watchers.
+    """
+
+    deferrable = True
+
+    def __call__(self):
+        if not is_workspace_feature_enabled():
```
deiferni, line 24: do you need these two safeguards or would the upgrade just query empty result sets if they were omitted?
I'm thinking about deployments where this flag could have been changed, and omitting the checks would maybe be the better option? not sure though, so i'll leave it up to you.

opengever/core/upgrades/20201127160828_correct_to_do_watchers/upgrade.py
```diff
@@ -0,0 +1,66 @@
...
+
+        if not is_activity_feature_enabled():
+            logger.warning('Activity feature is disabled for this deployment.')
+            return
+
+        self.center = notification_center()
+        self.session = create_session()
+
+        # Because WorkspaceWatcherManager.new_participant_added not only added
+        # the new participant as watcher to all ToDos but also with as role the
```
deiferni, line 35: typo "with as role"?

opengever/core/upgrades/20201127160828_correct_to_do_watchers/upgrade.py
```diff
@@ -0,0 +1,66 @@
...
+        # we can identify the wrongly added subscriptions by their role.
+        roles = ["WorkspaceGuest", "WorkspaceMember", "WorkspaceAdmin"]
+        subscriptions = Subscription.query.filter(Subscription.role.in_(roles))
+
+        todos = set()
+        for subscription in subscriptions:
+            resource = subscription.resource
+            obj = resource.oguid.resolve_object()
+            if not isinstance(obj, ToDo):
+                logger.warning(
```
deiferni, line 46: couldn't this be many objects? or are we only adding subscriptions to todos and this is unexpected? if it is expected if would omit logging on warning level.

opengever/core/upgrades/20201127160828_correct_to_do_watchers/upgrade.py
```diff
@@ -18,6 +18,8 @@ class CorrectToDoWatchers(UpgradeStep):
     """Correct to do watchers.
     """
 
+    deferrable = True
```
deiferni, line 21: We'd prefer to have one commit per upgrade-step, so fixup/squash this one, if not too inconvenient.

deiferni (approved):
LGTM 👍 , 

some suggestions for consideration. nothing that blocks a merge or needs further discussion if you decide against it.

opengever/core/upgrades/20201127160828_correct_to_do_watchers/upgrade.py
njohner: You're right that the queries should be empty for deployments where the workspace feature is deactivated.

I've added the second check because I realised that we now have the activity center managing the subscriptions. This means that any system where the activity feature is deactivated will have subscriptions that are not up to date, and so it will not work properly if we ever switch the activity feature on.

But I'll simply make a story for that.

opengever/core/upgrades/20201127160828_correct_to_do_watchers/upgrade.py
njohner: no, but I turned the sentence differently, hope it will be clearer 😅

opengever/core/upgrades/20201127160828_correct_to_do_watchers/upgrade.py
njohner: No this should only be ToDos. When a new participant is added on a Workspace he gets added as watcher only on the ToDos. And this were the subscriptions that are wrong.

PR #6754, opened 2020-11-30
Consistently use name solr update chain.

deiferni:
We've decided to use an unclamimed identifier for this new thing which makes sure certain fields from a GEVER Solr are also available in a RIS Solr.

- Solr sync is already taken by the (command-line) tool to make sure GEVER and Solr are in sync.
- Replication is already taken for certain deployment styles where Solr/GEVER instances are distributed across multiple servers.

So, `update chain` it is.

## Checklist (Must have)

- [ ] ~~Changelog entry~~
- [ ] ~~Link to issue (Jira or GitHub) and backlink in issue (Jira)~~

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -235,10 +235,10 @@ Then run the `activate_solr` maintenance script:
     $ bin/instance run src/opengever.maintenance/opengever/maintenance/scripts/activate_solr.py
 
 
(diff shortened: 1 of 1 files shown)


lukasgraf (approved):
LGTM! 👍  💚

PR #6755, opened 2020-11-30
Allow downloading and sending a document checked out by another user

elioschmutz:
Jira: https://4teamwork.atlassian.net/browse/NE-49

Currently, if another user has checked out a document, the `download_copy` and `attach_to_mail` actions were not shown.

With this PR, the actions will be shown even if the document is checked out by another user. But it have to have at least one version.

If the user who checked out the document downloads the document, the current working copy will be downloaded.
If another user download the document, the latest version will be downloaded.

By default, a newly created document does not have any version. If a user checks out this document and working on it will directly modify the document. That's the reason why we don't provide the download (or attach to mail) action to another user, if there does not exist any version yet.

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.16.0 (unreleased)
 ----------------------
 
+- Allow downloading and sending a document checked out by another user. [elioschmutz]
 - Fix keyword filter for keywords that contain spaces. [tinagerber]
 - Add feature flag for todos. [tinagerber]
 - No longer zip-export empty tasks, prevent creation of empty folders in such cases. [deiferni]
--- a/opengever/document/fileactions.py
+++ b/opengever/document/fileactions.py
@@ -200,21 +200,28 @@ class DocumentFileActions(BaseDocumentFileActions):
 
     def is_download_copy_action_available(self):
         """Disable downloading copies when the document is checked out by
-        another user.
+        another user and document doesn't have any version.
         """
         manager = getMultiAdapter(
             (self.context, self.request), ICheckinCheckoutManager)
+
+        has_version = self.context.get_current_version_id(missing_as_zero=True) > 0
+
         return (
             super(DocumentFileActions, self).is_download_copy_action_available()
-            and not manager.is_checked_out_by_another_user())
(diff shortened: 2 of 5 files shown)


opengever/document/fileactions.py
```diff
@@ -200,21 +200,28 @@ def is_cancel_checkout_action_available(self):
 
     def is_download_copy_action_available(self):
         """Disable downloading copies when the document is checked out by
-        another user.
+        another user and there is not yet an existing version.
```
njohner, line 203: and there is not yet an existing version -> and document doesn't have any version?

opengever/document/fileactions.py
```diff
@@ -200,21 +200,28 @@ def is_cancel_checkout_action_available(self):
...
+        has_version = self.context.get_current_version_id(missing_as_zero=True) > 0
+
         return (
             super(DocumentFileActions, self).is_download_copy_action_available()
-            and not manager.is_checked_out_by_another_user())
+            and not (manager.is_checked_out_by_another_user() and not has_version))
 
     def is_attach_to_email_action_available(self):
+        """Disable attaching documents to email when the document is checked out by
+        another user and there is not yet an existing version.
```
njohner, line 216: same:  and there is not yet an existing version -> and document doesn't have any version?

docs/HISTORY.txt
```diff
@@ -4,6 +4,7 @@ Changelog
 2020.15.0 (unreleased)
 ----------------------
 
+- Download or send a document checked out by another user will use the lates version instead the current working copy. [elioschmutz]
```
njohner, line 7: lates -> latest
instead the -> instead of the

But I'd actually change that to `Allow downloading and sending a document checked out by another user`. I think that is the main thing done in this PR, that it uses the latest version is an implementation detail.

njohner (approved):
Apart from comments, LGTM.

Also there are 2 typos in the PR title and commit message (lates -> latest, instead the -> instead of the)

PR #6757, opened 2020-12-01
Only fix subscription on current admin unit.

njohner:
The upgrade step fixing the subscriptions on ToDos was not restricted to the current admin_unit. While it did work correctly because it would end up skipping subscriptions from other admin units, it would also log an unnecessary warning.

For https://4teamwork.atlassian.net/browse/CA-1297

## Checklist (Must have)

- [x] Changelog entry


deiferni (approved):
😓

PR #6760, opened 2020-12-02
Adding a subtask to a sequential task through the restapi respects the `position` parameter.

elioschmutz:
Jira: https://4teamwork.atlassian.net/browse/NE-338

This PR extends the task POST endpoint to also respect the `positions` parameter if adding a task to a sequential task.

This was already implemented in the current Add form: https://github.com/4teamwork/opengever.core/blob/master/opengever/task/browser/forms.py#L111, but was forgotten for the API.

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 ----------------------
 
 - Allow downloading and sending a document checked out by another user. [elioschmutz]
+- Adding a subtask to a sequential task through the restapi respects the `position` parameter [elioschmutz]
 - Fix keyword filter for keywords that contain spaces. [tinagerber]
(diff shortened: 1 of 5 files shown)


njohner (commented):
Just wondering about validation. What happens if I POST with something that is not an int as `position`?
Otherwise LGTM

elioschmutz:
> Just wondering about validation. What happens if I POST with something that is not an int as `position`?
> Otherwise LGTM

I updated the PR with an error handling for not parsable `position` attribute values

elioschmutz:
@njohner could you take a final look?

njohner (approved):
LGTM

PR #6761, opened 2020-12-03
Add feature flag for todos.

tinagerber:
This PR adds a feature flag for todos. This allows us to enable and disable ToDos and ToDo lists. By default, this feature is enabled so that no additional update steps are required in the policies.
With the policy generator the feature can be enabled and disabled.

Jira: https://4teamwork.atlassian.net/browse/NE-163

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:

### DIFF
--- a/opengever/base/configuration.py
+++ b/opengever/base/configuration.py
@@ -36,6 +36,7 @@ from opengever.repository.interfaces import IRepositoryFolderRecords
 from opengever.sharing.interfaces import ISharingConfiguration
 from opengever.task.interfaces import ITaskSettings
 from opengever.workspace.interfaces import IWorkspaceSettings
+from opengever.workspace.interfaces import IToDoSettings
 from opengever.workspaceclient.interfaces import IWorkspaceClientSettings
 from pkg_resources import get_distribution
 from plone import api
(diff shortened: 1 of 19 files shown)


opengever/workspace/__init__.py
```diff
@@ -8,3 +8,9 @@ def is_workspace_feature_enabled():
     from opengever.workspace.interfaces import IWorkspaceSettings
     return api.portal.get_registry_record(
         'is_feature_enabled', interface=IWorkspaceSettings)
+
+
+def is_todo_feature_enabled():
+    from opengever.workspace.interfaces import IToDoSettings
```
deiferni, line 14: are you running into cyclic imports here? if not please move it up to module level.

opengever/base/configuration.py
```diff
@@ -160,6 +161,7 @@ def get_features(self):
         features['resolver_name'] = api.portal.get_registry_record('resolver_name', interface=IDossierResolveProperties)
         features['sablon_date_format'] = api.portal.get_registry_record('sablon_date_format_string', interface=IMeetingSettings)  # noqa
         features['solr'] = api.portal.get_registry_record('use_solr', interface=ISearchSettings)
+        features['todo'] = api.portal.get_registry_record('is_feature_enabled', interface=IToDoSettings)
```
deiferni, line 164: please rename this to `workspace_todo`, we don't have good namespacing but todo is a bit too general IMO.

opengever/core/upgrades/20201203113604_add_to_do_feature_flag/registry.xml
```diff
@@ -0,0 +1,3 @@
+<registry>
+  <records interface="opengever.workspace.interfaces.IToDoSettings" />
```
deiferni, line 2: please restrict this upgrade step to the field you have added. if more fields are added in later upgrades and they are all installed at the same time this might conflict.

deiferni (changes requested):
I'd like to see a bit more testing here, e.g. also for api content creation. Apart from comments LGTM 👍.

tinagerber:
@deiferni I have made the desired changes. Can you rereview?

deiferni (approved):

PR #6764, opened 2020-12-04
Add sequence_type to task serializer

tinagerber:
In the frontend we want to show whether a "Standardablauf" is sequential or parallel. Therefore I added the field `sequence_type` to the task serializer.

Jira: https://4teamwork.atlassian.net/browse/NE-314

## Checklist (Must have)

- [x] Changelog entry


deiferni (approved):
LGTM 👍

PR #6765, opened 2020-12-04
Implement specific task zip representation, no longer export empty tasks.

deiferni:
We don't export tasks that don't contain any documents to prevent creation of empty folders for such tasks.

Aborting for empty tasks has been implemented as a catalog query for documents. This has proven to be the easiest way to hook into existing functionality of [ftw.zipexport](https://github.com/4teamwork/ftw.zipexport/ ) without breaking/changing the whole zipexport behavior.

Jira: https://4teamwork.atlassian.net/browse/CA-1167

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- New functionality:

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,7 +4,7 @@ Changelog
 2020.16.0 (unreleased)
 ----------------------
 
-- Nothing changed yet.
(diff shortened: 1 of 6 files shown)


njohner (approved):
LGTM. Thanks for all the testing 🥇 !

PR #6766, opened 2020-12-07
Use form action URL to re-render specific widget.

deiferni:
Instead of using browser location we use the form action to re-render specific widgets. The browser location does not always expose the add form name, for example when opened via POST from a tabbed view action.

The form action however always points to the correct form and is more reliable.

This fixes a bug with only rendering allowed proposal templates for the chosen committee on a proposal add form, when the form was opened with the "add proposal" action on a document tab in the classic ui.

Jira: https://4teamwork.atlassian.net/browse/CA-1234

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 
 - No longer zip-export empty tasks, prevent creation of empty folders in such cases. [deiferni]
 - Add sequence_type to task serializer. [tinagerber]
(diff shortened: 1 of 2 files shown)


njohner (approved):
Is this easily testable with the testbrowser?

Otherwise LGTM

deiferni:
@njohner No it is only testable manually as it is javascript

PR #6767, opened 2020-12-07
Add @assign-to-dossier endpoint for forwardings

elioschmutz:
This PR introduces a new endpoint `@assign-to-dossier` for forwardings.

## `@workflow` vs `@assign-to-dossier`
A forwarding can be assigned to a dossier through the workflow transition: `forwarding-transition-assign-to-dossier`.

Executing this transition will perform several steps:
- creating a new task
- linking the newly created task with the forwarding
- moving the forwarding into the years-folder of the inbox

Just calling the transition through the `@worflow` endpoint would be possible but comes with some disadvantages. The `@workflow` endpoint is a generic endpoint for performing workflow transitions on an object. The UI would not know anything about the newly created task nor about the moved forwarding.

The `@assign-to-dossier` endpoint takes care of it and will return the newly created task as the response.

## task-creation

The old UI was not able to directly create a task defined by the user. So the UI created the task with defaults taken from the forwarding (like the title) and showed up the edit-form of the already created task to the user.

The new ui should be able to collect all the user inputs and then assing the forwarding to a dossier with the user defined task attributes.

So this PR introduced a new parameter: `task` for the endpoint which allows us to directly pass a task-payload which will then be used for creating the task.

Jira: https://4teamwork.atlassian.net/browse/NE-73

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/docs/public/dev-manual/api/inbox.rst
+++ b/docs/public/dev-manual/api/inbox.rst
@@ -1,0 +1,86 @@
+.. _inbox:
+
+Eingangskorb
+============
+
+Weiterleitungen einem Dossier zuweisen
+--------------------------------------
+
+Eine Weiterleitung kann über eine Workflow-Transition einem Dossier zugewiesen werden.
+
+Für das Zuweisen kann sowohl der ``@workflow`` Endpoint oder aber auch der ``@assign-to-dossier`` Endpoint verwendet werden.
+
+Der Vorteil vom ``@assign-to-dossier`` gegenüber dem ``@worfklow`` Endpoint ist, dass dieser nach erfolgreichem Zuweisen den neu erstellten Task zurück gibt. Zudem werden Fehler im Payload besser behandelt als im generischen ``@worfklow`` Endpoint.
+
+Die Verwendung des ``@workflow``-Endpoints kann im Kapitel :ref:`Workflow <workflow>` nachgeschaut werden. Die zu verwendende Transition ist ``forwarding-transition-assign-to-dossier``.
+
+Beim Zuweisen einer Weiterleitung an ein Dossier wird eine neue Aufgabe im angegebenen Dossier erstellt und mit der Weiterleitung verknüpft. Die Weiterleitung wird erledigt und in den aktuellen Jahresordner der Inbox verschoben.
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+      POST /inbox/forwarding-1/@assign-to-dossier HTTP/1.1
+      Accept: application/json
+      Content-Type: application/json
+
+      {
+        "target_uid": "123",
+        "comment": "Bitte anschauen"
+      }
+
+
+**Beispiel-Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 201 Created
+      Content-Type: application/json
+
+      {
+        "@id": "http://example.org/ordnungssystem/dossier-1/task-1",
+        "...": "..."
+      }
+
+
+Aufgabeneigenschaften ändern
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Standardmässig wird beim Zuweisen einer Weiterleitung an ein Dossier eine Aufgabe erstellt, welche den Titel und weitere Attribute von der Weiterleitung erben.
+
+Die Aufgabe kann jedoch über den Endpoint komplett selber definiert werden. Dazu kann der Parameter ``task`` im Payload des Requests verwendet werden.
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+      POST /inbox/forwarding-1/@assign-to-dossier HTTP/1.1
+      Accept: application/json
+      Content-Type: application/json
+
+      {
+        "target_uid": "123",
+        "task": {
+          "title": "Wichtige Aufgabe aus einer Weiterleitung",
+          "task_type": "information",
+          "deadline": "2016-11-01",
+          "is_private": false,
+          "responsible": "robert.ziegler",
+          "responsible_client": "fa",
+          "revoke_permissions": true,
+          "issuer": "robert.ziegler"
+        }
+      }
+
+
+**Beispiel-Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 201 Created
+      Content-Type: application/json
+
+      {
+        "@id": "http://example.org/ordnungssystem/dossier-1/task-1",
+        "title": "Wichtige Aufgabe aus einer Weiterleitung",
+        "...": "..."
+      }
--- a/opengever/api/forwarding.py
+++ b/opengever/api/forwarding.py
@@ -1,0 +1,71 @@
+from opengever.dossier.behaviors.dossier import IDossierMarker
+from plone import api
+from plone.protect.interfaces import IDisableCSRFProtection
+from plone.restapi.deserializer import json_body
+from plone.restapi.interfaces import ISerializeToJson
+from plone.restapi.services import Service
+from plone.uuid.interfaces import IUUID
+from zExceptions import BadRequest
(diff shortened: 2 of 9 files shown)


opengever/base/transition.py
```diff
@@ -61,61 +61,68 @@ def _deserialize_values(self, transition_params, collect_errors=False):
         """Deserialize and validates the defined schema returns a values and
         a list of errors.
         """
-        schema_data = {}
         errors = []
         values = {}
 
         for schemata in self.schemas:
-            for name, field in getFields(schemata).items():
```
lukasgraf, line 69: Since you're already touching this part of the code: Could you please also rename `schemata` to `schema` in those two places where it still appears?  _(`schemata` is plural, for multiple schemas, and it's really confusing to read)_

opengever/inbox/transition.py
```diff
@@ -121,6 +129,19 @@ def create_successor_task(self, dossier):
 
         return task
 
+    def _deserialize_values(self, transition_params, collect_errors=False):
+        """Also deserializes the `task` data from the transition_params. This allows
+        us to directly create a specific task through the workflow transition.
+        """
+        values, errors = super(ForwardingAssignToDossierTransitionExtender, self)._deserialize_values(transition_params, collect_errors)
+        task = transition_params.get('task', None)
```
lukasgraf, line 137: `None` is always the default for `dict.get()` ✂️

docs/public/dev-manual/api/inbox.rst
```diff
@@ -0,0 +1,86 @@
+.. _inbox:
+
+Inbox
```
lukasgraf, line 3: Inbox -> Eingangskorb

docs/public/dev-manual/api/inbox.rst
```diff
@@ -0,0 +1,86 @@
...
+        "@id": "http://example.org/ordnungssystem/dossier-1/task-1",
+        "...": "..."
+      }
+
+
+Aufgabeneigenschaften ändern
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Standardmässig wird beim Zuweisen einer Weiterleitung an ein Dossier eine Aufgabe erstellt, welche den Titel und weitere Attribute von der Weiterleitung erben.
+
+Die Aufgabe kann jedoch über den Endpoint komplett selber definiert werden. Dazu kann der Parameter ``task`` im Payload des Reqeusts verwendet werden.
```
lukasgraf, line 50: Reqeusts -> Requests

docs/public/dev-manual/api/inbox.rst
```diff
@@ -0,0 +1,86 @@
...
+=====
+
+Weiterleitungen einem Dossier zuweisen
+--------------------------------------
+
+Eine Weiterleitung kann über eine Workflow-Transition einem Dossier zugewiesen werden.
+
+Für das Zuweisen kann sowohl der ``@workflow`` Endpoint oder aber auch der ``@assign-to-dossier`` Endpoint verwendet werden.
+
+Der Vorteil vom ``@assign-to-dossier`` gegenüber dem ``@worfklow`` Endpoint ist, dass dieser nach erfolgreichem Zuweisen den neu erstellten Task zurück gibt. Zudem werden Fehler im Payload besser behandlet als im generischen ``@worfklow`` Endpoint.
```
lukasgraf, line 13: behandlet -> behandelt

docs/public/dev-manual/api/inbox.rst
```diff
@@ -0,0 +1,86 @@
...
+
+Die Verwendung des ``@workflow``-Endpoints kann im Kapitel :ref:`Workflow <workflow>` nachgeschaut werden. Die zu verwendende Transition ist ``forwarding-transition-assign-to-dossier``.
+
+Beim Zuweisen einer Weiterleitung an ein Dossier wir eine neue Aufgabe im angegebenen Dossier erstellt und mit der Weiterleitung verknüpft. Die Weiterleitung wird erledigt und in den aktuellen Jahresordner der Inbox verschoben.
```
lukasgraf, line 17: wir -> wird

opengever/api/forwarding.py
```diff
@@ -0,0 +1,65 @@
...
+
+        self.request.response.setStatus(201)
+        self.request.response.setHeader("Location", task.absolute_url())
+
+        return self.serialize(task)
+
+    def extract_params(self):
+        json_data = json_body(self.request)
+
+        task_payload = json_data.get("task", None)
```
lukasgraf, line 43: ✂️ `None`

opengever/api/forwarding.py
```diff
@@ -0,0 +1,65 @@
...
+        target_uid = json_data.get("target_uid", None)
+        comment = json_data.get("comment", u'')
+
+        if not target_uid:
+            raise BadRequest('Required parameter "target_uid" is missing in body')
+
+        try:
+            target = api.content.get(UID=target_uid)
+        except:
```
lukasgraf, line 52: No bare `except:` please

opengever/api/forwarding.py
```diff
@@ -0,0 +1,65 @@
...
+            raise BadRequest('target_uid: "{}" does not exist')
```
lukasgraf, line 53: This string is not interpolated, `{}` never gets replaced with any value anywhere.

opengever/api/forwarding.py
```diff
@@ -0,0 +1,65 @@
...
+
+        if not IDossierMarker.providedBy(target):
+            raise BadRequest('target_uid: "{}" is not a dossier')
```
lukasgraf, line 56: Same here, missing string interpolation.

opengever/api/tests/test_assign_to_dossier.py
```diff
@@ -0,0 +1,187 @@
...
+from opengever.base.role_assignments import RoleAssignmentManager
+from opengever.base.role_assignments import SharingRoleAssignment
+from opengever.inbox.yearfolder import get_current_yearfolder
+from opengever.testing import IntegrationTestCase
+from opengever.testing import obj2brain
+from plone import api
+import json
+
+
+class TestAssignToDossier(IntegrationTestCase):
```
lukasgraf, line 12: There doesn't seem to be any tests for the better error handling in the `@assign-to-dossier` endpoint 😉

docs/public/dev-manual/api/inbox.rst
```diff
@@ -0,0 +1,86 @@
+.. _inbox:
+
+Inbox
+=====
+
+Weiterleitungen einem Dossier zuweisen
```
lukasgraf, line 6: I would generally rewrite this in a style where only the use of `@assign-to-dossier` is recommended, and the use of `@workflow` is discouraged.

lukasgraf (changes requested):

docs/public/dev-manual/api/inbox.rst
elioschmutz: as discussed, I let it how it is.

It's important to know that it is a workflow transition and that the transition can also be executed though the `@workflow` endpoint like every other transition.

opengever/api/forwarding.py
elioschmutz: `api.content.get` will return `None` if no content could be found. So the implementation is false anyway.

I updated it to an `if target`-block

opengever/api/tests/test_assign_to_dossier.py
elioschmutz: I also added some tests for the error-handling

lukasgraf (approved):
LGTM! 👍

PR #6768, opened 2020-12-07
Fix deletion of favorites when object is removed or trashed.

njohner:
When trashing or deleting an object, the corresponding favorites get deleted by an event handler. This did not work because the session did not get marked as changed and hence the changes were not persisted in the SQL DB. 

Moreover the favorites deletion failed for objects contained in the deleted object: the OGUID of the object removed could not be resolved anymore because its intid had already been removed from the catalog. We fix this by registering the handler earlier (on `ObjectWillBeRemovedEvent` instead of `ObjectRemovedEvent`).

We also add an upgrade step to fix potentially broken favorites for trashed and deleted objects.

For https://4teamwork.atlassian.net/browse/CA-1186

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/core/upgrades/20201207113732_clean_up_favorites_for_trashed_and_deleted_objects/upgrade.py
+++ b/opengever/core/upgrades/20201207113732_clean_up_favorites_for_trashed_and_deleted_objects/upgrade.py
@@ -1,0 +1,64 @@
+from opengever.core.upgrade import SQLUpgradeStep
+from plone import api
+from sqlalchemy.sql.expression import column
+from sqlalchemy.sql.expression import select
+from sqlalchemy.sql.expression import table
+from zope.app.intid.interfaces import IIntIds
+from zope.component import getUtility
+import logging
+
+
+LOG = logging.getLogger('ftw.upgrade')
+
+favorites_table = table(
+    'favorites',
+    column('id'),
+    column('plone_uid'),
+    column('admin_unit_id'),
+    column('int_id'),
+)
+
+
+class CleanUpFavoritesForTrashedAndDeletedObjects(SQLUpgradeStep):
+    """Clean up favorites for trashed and deleted objects.
+    """
+
+    deferrable = True
+
+    def migrate(self):
+
+        # We delete all favorites for trashed objects
+        current_admin_unit_id = api.portal.get_registry_record(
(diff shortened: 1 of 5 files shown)


opengever/core/upgrades/20201207113732_clean_up_favorites_for_trashed_and_deleted_objects/upgrade.py
```diff
@@ -0,0 +1,64 @@
...
+            .distinct()
+        ).fetchall()
+        favorite_uids = [row[0] for row in rows]
+
+        query = {"trashed": True,
+                 "UID": favorite_uids}
+        for brain in self.catalog_unrestricted_search(query=query):
+            self.execute(
+                favorites_table.delete()
+                .where(favorites_table.c.plone_uid == brain.UID)
```
deiferni, line 46: If you use an `IN` operator here with a pre-populated list, you can fire a single query instead of one query per record to delete.

opengever/core/upgrades/20201207113732_clean_up_favorites_for_trashed_and_deleted_objects/upgrade.py
```diff
@@ -0,0 +1,64 @@
...
+
+        rows = self.execute(
+            select([favorites_table.c.int_id])
+            .where(favorites_table.c.admin_unit_id == current_admin_unit_id)
+            .distinct()
+        ).fetchall()
+
+        favorite_to_delete = [row[0] for row in rows if row[0] not in intids]
+        for intid in favorite_to_delete:
+            self.execute(
```
deiferni, line 61: same here, i'd prefer one query, if possible.

deiferni (approved):
👍  LGTM apart from comments

I'm a bit confused by:
> When trashing or deleting an object, the corresponding favorites get deleted by an event handler. This did not work because the session did not get marked as changed and hence the changes were not persisted in the SQL DB.

Deletion of favorites runs in the same transaction as actual object deletion, so it should be changed, no? Or was the issue that by bypassing of the ORM with direct statement execution the session is never marked as changed?

njohner:
@deiferni Yes I think the problem was bypassing the ORM. The `Favorite.__table__` object is not bound to the current session, or something like that.

PR #6769, opened 2020-12-07
Backport filing number search fix (#6578)

lukasgraf:
This is a backport of #6578 to `2019.6-stable`.

I had to drop the new test from this fix when backporting, because the test requires the `SolrIntegrationTestCase`, which wasn't around yet in `2019.6`.

Tests against this branch are still failing though, because the boostrap invocation on Jenkins with `--setuptools-version` would require an updated `bootstrap.py` (and possibly newer versions of `setuptools` and `zc.buildout`), which we don't want to bump for this branch as part of a backport.

I tested the fix locally - I was able to reproduce the issue, the fix + upgrade step worked and fixed it 👍 

Runtime for the upgrade step (as tested locally) is around `8 reindexed dossiers / s`.

Jira: https://4teamwork.atlassian.net/browse/CA-1247

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,7 +4,8 @@ Changelog
 2019.6.5 (unreleased)
 ---------------------
 
-- Nothing changed yet.
+- Use filing_no field in advanced search form. [njohner]
+- Reindex SearchableText when filing number is set. [njohner]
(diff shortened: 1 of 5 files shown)


njohner (approved):
LGTM

PR #6770, opened 2020-12-07
Add public_trial field to listing endpoint.

tinagerber:
In order to display the `public_trial` in a document listing, it must be allowed in the listing endpoint.

Jira: https://4teamwork.atlassian.net/browse/NE-147

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:

### DIFF
--- a/opengever/api/solr_query_service.py
+++ b/opengever/api/solr_query_service.py
@@ -61,6 +61,10 @@ def translate_document_type(document_type):
         return term.title
 
 
+def translate_public_trial(public_trial):
+    return translate(public_trial, context=getRequest(), domain="opengever.base")
+
+
 def filesize(obj):
     try:
         filesize = obj.filesize
@@ -103,6 +107,13 @@ def translated_title(obj):
 
 def translated_task_type(obj):
     return task_type_helper(obj, obj.get("task_type"))
+
+
+def translated_public_trial(obj):
+    try:
+        return translate(obj.public_trial, context=getRequest(), domain="opengever.base")
+    except AttributeError:
+        return None
 
 
 def to_relative_path(value):
@@ -301,6 +312,8 @@ FIELDS_WITH_MAPPING = [
                  additional_required_fields=['bumblebee_checksum', 'path']),
     ListingField('preview_url', None, 'get_preview_frame_url', DEFAULT_SORT_INDEX,
                  additional_required_fields=['bumblebee_checksum', 'path']),
+    ListingField('public_trial', 'public_trial', accessor=translated_public_trial,
+                 transform=translate_public_trial),
     ListingField('reference_number', 'reference'),
     ListingField('relative_path', 'path', relative_path, transform=to_relative_path),
     ListingField('responsible', 'responsible', transform=display_name),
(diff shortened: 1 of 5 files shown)


opengever/api/solr_query_service.py
```diff
@@ -105,6 +105,13 @@ def translated_task_type(obj):
     return task_type_helper(obj, obj.get("task_type"))
 
 
+def translated_public_trial(brain):
+    try:
+        return translate(brain.public_trial, context=getRequest(), domain="opengever.base")
+    except AttributeError:
```
njohner, line 111: This seems unnecessary. All brains will have a `public_trial`, although it will be `None` for some
Seems to me:
```
return translate(brain.public_trial, context=getRequest(), domain="opengever.base") if brain.public_trial else None
```
would do no?

njohner (commented):
I'm pretty sure your facets will not be translated as is. You need to also define the `transform` for a translated field.

You can add a test similar to https://github.com/4teamwork/opengever.core/blob/master/opengever/api/tests/test_listing.py#L720-L739 ;-)

opengever/api/solr_query_service.py
tinagerber: Well it's not really a brain, it's a `OGSolrContentListingObject` and therefore it throws an AttributeError for workspaces for example. I renamed it to `obj`

tinagerber:
@njohner I don't need the label as a facet, I just need it as a translated field.

njohner:
Well, the API does allow to ask for the facets of that field and right now they are broken. And in a few months the frontend will want to add a filter for the `public_trial` and then that will be broken...

Also I've had a quick look, and I'd rather call the field `public_trial`. I don't see any good reason to diverge from the field's name here.

njohner:
Here is the current answer for the facets...
```
    "facets": {
        "public_trial_label": {
            "unchecked": {
                "count": 4,
                "label": "unchecked"
            }
        }
    },
```

tinagerber:
@njohner Now it's the same behavior as with `review_state_label`.
```
  "facets": {
    "review_state_label": {
      "document-state-draft": {
        "count": 54,
        "label": "document-state-draft"
      },
      "mail-state-active": {
        "count": 5,
        "label": "mail-state-active"
      }
    }
  },
  "items": [
    {
      "@id": "http://localhost:8081/fd/ordnungssystem/bevoelkerung-und-sicherheit/einwohnerkontrolle/dossier-1/document-5",
      "UID": "62411410e3f7462e8a7390e30f37f438",
      "document_type": null,
      "filename": "Memo - Bitte lesen.docx",
      "review_state_label": "Non défini",
      "title": "Memo - Bitte lesen"
    },
    ...
```

 I have named it public_trial_label so that there is still the possibility to get public_trial untranslated.

tinagerber:
@njohner Do you think it's wrong for `review_state_label`?

njohner:
As discussed, I think it's wrong for `public_state_label`, and I think it's wrong to even have `public_state_label`.
I've prepared a story to have a closer look at that (https://4teamwork.atlassian.net/browse/CA-1334).

njohner (approved):
LGTM

PR #6771, opened 2020-12-08
Only expose translated title fields for active languages in schema and serialization via API.

deiferni:
With this PR we change the API behavior with translated title fields to appear dynamically according to deployments currently active languages. The following changes are made:
- The fields `title_de` and `title_fr` will only appear in the `@schema` and `@types` endpoints if their corresponding lanugage has been set to active on the plone site.
- The fields `title_de` and `title_fr` will also only be serialized if their corresponding language is active.
- Deserialization (`POST`ing or `PATCH`ing via API) will be a bit more permissive and also accept inactive language fields.

I have attempted to implement the changes in a generic and, if possible, non-intrusive way.

Unfortunately schema serialization in `plone.restapi` [has been implemented in a function](https://github.com/plone/plone.restapi/blob/1fb8709402d0f33ac530dd8969f817bbb9aacb4c/src/plone/restapi/types/utils.py#L250-L256), so a monkey patch was necessary. I have implemented it in such a way that it performs inspection and checks for the `ITranslatedTitle` behavior on the FTI. That way it should automatically work for all translated types and is also active for our `@schema` endpoint and for the `@types` endpoint of `plone.restapi`.

Being permissive with deserialization has been guided by the following thoughts:
- The field/storage is always available in code, it is just hidden when serializing content and schema
- No harm is done when the field is set even if not displayed, on the contrary:
  - it might help to prepare a potential migration
  - if the UI (or another client) decides to send the data (for whatever reason) we will still be able to accept it and store it correctly
- It was less intrusive to implement than [overriding the whole `get_schema_data` method](https://github.com/plone/plone.restapi/blob/1fb8709402d0f33ac530dd8969f817bbb9aacb4c/src/plone/restapi/deserializer/dxcontent.py#L79-L154) in `plone.restapi`

Potential discussion points for the review:
- I've introduced a new field type `TranslatedTextLine` as a helper class, should this technical detail be hidden from the docs (let it appear as `TextLine` only)?
- Currently there will be no way to retrieve language fields for inactive languages via API. As an alternative we could also refrain from omitting it from serialization, but we might need to see how the new UI form implementation looks like to be certain here.

Jira: https://4teamwork.atlassian.net/browse/CA-1226

_This needs to be shipped together with corresponding UI changes in https://github.com/4teamwork/gever-ui/pull/1448._

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [x] api-change label added

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 ----------------------
 
 - Add feature flag for todos. [tinagerber]
+- Only expose translated title fields for active languages in schema and serialization via API. [deiferni]
 - No longer zip-export empty tasks, prevent creation of empty folders in such cases. [deiferni]
 - Add sequence_type to task serializer. [tinagerber]
 - Fix only rendering allowed proposal templates when proposal add form is opened from documents tab. [deiferni]
--- a/docs/public/dev-manual/api/content_types.rst
+++ b/docs/public/dev-manual/api/content_types.rst
@@ -8,6 +8,18 @@ Inhaltstypen
    :backlinks: none
 
 
+Übersetzte Titel
+----------------
+
+Die Felder für übersetzte Titel der Inhalte sind der Vollständigkeit halber
+in der Dokumentation alle aufgeführt. Je nach konfigurierter Sprache auf den
+Systemen stehen sie aber nicht alle zur Verfügung. Betroffen davon sind die
+Felder ``title_de`` und ``title_fr``. Eine Abfrage auf den ``@schema`` Endpoint
+liefert die aktuell gültigen Schemas eines Deployments.
+Das Erstellen und Modifizieren von Inhalten mittels ``POST`` oder ``PATCH``
+erlaubt immer alle übersetzten Felder.
+
+
 Schemas
 -------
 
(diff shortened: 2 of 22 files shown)


lukasgraf (approved):
LGTM! 👍

lukasgraf:
> Being permissive with deserialization has been guided by the following thoughts [...]
👍 

> Potential discussion points for the review:
> - I've introduced a new field type `TranslatedTextLine` as a helper class, should this technical detail be hidden from the docs (let it appear as `TextLine` only?

I think it's ideal the way it is. The detail is there should anyone care about it, but the Zope types are usually not relevant for API consumers anyway (ASCIILine vs. TextLine), the types that apply to them are on a JSON level (=string), which are expressed in the JSON schema.

> - Currently there will be no way to retrieve language fields for inactive languages via API. As an alternative we could also refrain from omitting it from serialization, but we might need to see how the new UI form implementation looks like to be certain here.

I would leave it like this. If the need arises to serialize content for inactive languages we can still revisit this, but for now I like the consistent behavior.

deiferni:
Good news, we've found out the UI already supports optional fields, see https://github.com/4teamwork/gever-ui/pull/1448. This PR can be un-drafted and :shipit:.

deiferni:
thanks for the feedback @lukasgraf, i'll move forward here.

PR #6772, opened 2020-12-08
Fix keyword filter for keywords that contain spaces

tinagerber:
Until now, the keyword filter did not find dossiers when filtering for keywords containing spaces.
To filter for a keyword that consists of several words, the keyword must be enclosed in quotation marks.

Jira: https://4teamwork.atlassian.net/browse/NE-365

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.16.0 (unreleased)
 ----------------------
 
+- Fix keyword filter for keywords that contain spaces. [tinagerber]
 - Add feature flag for todos. [tinagerber]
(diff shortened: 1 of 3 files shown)


lukasgraf (approved):
LGTM 👍

njohner (approved):
Can it happen that we filter by subject but do not pass it as a list? If so you should also quote on line 172 (`filters.append(u'{}:{}'.format(key, escape(value)))`).

Apart from that LGTM

njohner:
😅

tinagerber:
No, it's always a dict: https://github.com/4teamwork/opengever.core/blob/532b0a2b0ef1c8e56a1102e18ad590cdff135ba4/opengever/tabbedview/filters.py#L134-L137

PR #6773, opened 2020-12-08
Make bumblebee internal plone url overwriteable.

deiferni:
By extracting it into a configuration variable when can override the variable in anything that extends from development.cfg. This has become necessary when one wants to run the dockerized bumblebee locally.

## Checklist (Must have)

- ~~[ ] Changelog entry~~
- ~~[ ] Link to issue (Jira or GitHub) and backlink in issue (Jira)~~


njohner (approved):
LGTM

PR #6774, opened 2020-12-09
Implement OGDS sync for local groups

buchi:
OGDS sync for local groups is needed for teamraum to reflect changes that can be done in various places (e.g. ZMI or import of groups) in OGDS.

By default sync of local groups is only enabled for installations with enabled workspace feature.
It can be executed on any site using the control panel.

JIRA: https://4teamwork.atlassian.net/browse/CA-1332

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/ogds/base/sync/ogds_updater.py
+++ b/opengever/ogds/base/sync/ogds_updater.py
@@ -13,6 +13,7 @@ from opengever.ogds.base.sync.import_stamp import set_remote_import_stamp
 from opengever.ogds.models.group import Group
 from opengever.ogds.models.user import User
 from plone import api
+from Products.CMFCore.utils import getToolByName
 from Products.CMFPlone.interfaces import IPloneSiteRoot
 from Products.CMFPlone.utils import safe_unicode
 from Products.LDAPMultiPlugins.interfaces import ILDAPMultiPlugin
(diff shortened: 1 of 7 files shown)


opengever/ogds/base/sync/ogds_updater.py
```diff
@@ -423,6 +501,26 @@ def ldap_groups_and_members(self):
...
+
+        groups = {}
+        members = {}
+        for group in groups_plugin.getGroups():
+            groupid = group.getId()
+            if groupid in IGNORED_LOCAL_GROUPS:
+                continue
+            groups[groupid] = {
+                'groupid': groupid.decode('utf8'),
+                'title': groups_plugin.getGroupInfo(groupid)['title'].decode('utf8'),
```
deiferni, line 517: is it guaranteed we always have a `title` (e.g. fallback to groupid)? if not, this statement will need to be a bit more careful with missing titles.

deiferni (approved):
could you add a test that documents how conflicts are handled when the group exists in LDAP and in plone?

apart from that LGTM 👍

opengever/ogds/base/sync/ogds_updater.py
buchi: The `title` should always exist, however it could be `None` if the group was added programmatically.
https://github.com/zopefoundation/Products.PluggableAuthService/blob/master/Products/PluggableAuthService/plugins/ZODBGroupManager.py#L180

I've implemented it a little bit more defensively now.

buchi:
Added tests for conflicting groups. Basically the first one wins.

PR #6775, opened 2020-12-09
@assign-to-dossier endpoint deserializes combined values for the responsible

elioschmutz:
Creating a task requires a `responsible` and a `responsible_client`. From the vocabularies, we receive combined values for the responsible field like `fd:username`. Instead of separating this values in the frontend and sending it in separate attributes to the server, we deserialize combined values in the backend and separate it properly: https://github.com/4teamwork/opengever.core/blob/master/opengever/api/task.py#L105

The newly introduced endpoint `@assing-to-dossier` in https://github.com/4teamwork/opengever.core/pull/6767  does not implement this special deserialization of the responsible field.

This PR extends the endpoint to also provide combined values for task responsibles.

Jira: https://4teamwork.atlassian.net/browse/NE-73

## Checklist (Must have)

- [x] Changelog entry (⚠️ not required because its a follow-up pr)

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/api/forwarding.py
+++ b/opengever/api/forwarding.py
@@ -1,3 +1,4 @@
+from opengever.api.task import deserialize_responsible
 from opengever.dossier.behaviors.dossier import IDossierMarker
 from plone import api
 from plone.protect.interfaces import IDisableCSRFProtection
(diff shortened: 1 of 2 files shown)


njohner (approved):
LGTM

PR #6776, opened 2020-12-10
Download checked out document also respect the init version 0

elioschmutz:
Jira: https://4teamwork.atlassian.net/browse/NE-49
This is a follow-up PR of https://github.com/4teamwork/opengever.core/pull/6755

I assumed, that the first version will be created as soon as the user is checking in a document the first time.

But that's not right. There is a file-setter on the document model which is creating an initial version 0 if the file-field is updated the first time. Means, if a user creates a new document, there will be no version at all. If the user checks out the document and saves it with some changes without checking it in, an initial version 0 will be created by the file-setter: https://github.com/4teamwork/opengever.core/blob/master/opengever/document/document.py#L263

I handled the version 0 like there is no version.

This PR updates the implementation of the base-PR to also respect the version 0 as a version. 

## Checklist (Must have)

- [x] Changelog entry (ℹ️  not necessary because it's a follow-up PR within the same version)

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/api/tests/test_actions.py
+++ b/opengever/api/tests/test_actions.py
@@ -7,6 +7,7 @@ from opengever.trash.trash import ITrashable
 from opengever.workspaceclient.interfaces import ILinkedWorkspaces
 from opengever.workspaceclient.tests import FunctionalWorkspaceClientTestCase
 from plone import api
+from plone.namedfile.file import NamedBlobFile
 from plone.protect import createToken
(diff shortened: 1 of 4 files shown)


njohner (approved):
LGTM

PR #6778, opened 2020-12-10
Fix type of file contentType on eCH0147 import

buchi:
contentType was of type `pyxb.binding.datatypes.token` before, which is not JSON serializable.

https://sentry.4teamwork.ch/organizations/sentry/issues/65198/?project=17

I've not added an upgrade step to fix existing documents because it requires to check all documents which is costly and eCH0147 import is rarely used.

I've fixed the affected system with the following script:
```python
from pyxb.binding.datatypes import token
import transaction

site = app.digs
docs = site.portal_catalog.unrestrictedSearchResults(portal_type='opengever.document.document')

for doc in docs:
    obj = doc.getObject()
    if obj.file:
        if type(obj.file.contentType) == token:
            obj.file.contentType = str(obj.file.contentType)
            print "Fixed contenType for %s" % doc.getPath()

transaction.commit()
```

JIRA: https://4teamwork.atlassian.net/browse/CA-1338

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -16,6 +16,7 @@ Changelog
 - Add sequence_type to task serializer. [tinagerber]
 - Fix only rendering allowed proposal templates when proposal add form is opened from documents tab. [deiferni]
 - Add OGDS sync for local groups. [buchi]
+- Fix type of file contentType on eCH0147 import. [buchi]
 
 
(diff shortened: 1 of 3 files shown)


lukasgraf (approved):
LGTM 👍

PR #6780, opened 2020-12-10
Improve document and template folder actions

elioschmutz:
Jira: https://4teamwork.atlassian.net/browse/NE-255

This PR adds missing tests for the document and template folder actions and makes some corrections to the exposed template actions:
- `delete`-action will be displayed for documents within dossier templates
- `attach_to_email` will no longer be displayed for templates
- `trash_document` will no longer be displayed for templates
- `send_as_mail` will no longer be displayed within the templates section
- `trash` and `untrash` will no longer be displayed within the templates section

See also: https://4teamwork.atlassian.net/browse/NE-431

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations


opengever/api/tests/test_actions.py
```diff
@@ -505,6 +505,82 @@ def test_untrashing_available_for_trashed_document(self, browser):
...
+    def test_available_file_actions_for_subdossier_template_document(self, browser):
+        self.login(self.dossier_responsible, browser)
+
+        template = create(Builder('document')
+                          .within(self.subdossiertemplate)
+                          .titled(u'Werkst\xe4tte')
+                          .with_dummy_content())
+
+        expected_file_actions = [
+            {u'icon': u'', u'id': u'oc_direct_checkout', u'title': u'Checkout and edit'},
```
njohner, line 550: hum why is that different from `dossier_template_document`??

opengever/document/fileactions.py
```diff
@@ -99,7 +99,10 @@ def is_revert_to_version_action_available(self):
 
     def is_trash_document_available(self):
         trasher = ITrashable(self.context)
-        return trasher.verify_may_trash(raise_on_violations=False)
+        return (
+            trasher.verify_may_trash(raise_on_violations=False)
```
njohner, line 103: Shouldn't we rather add that check in the `trasher.verify_may_trash` method?

njohner (commented):
LGTM, just some minor comments

opengever/api/tests/test_actions.py
elioschmutz: 👍 Updated. It was because I used a different user

opengever/document/fileactions.py
elioschmutz: I'm not sure. It's possible to trash. But it's not available for this object. So I think, it belongs to this file...

elioschmutz:
@njohner I updated the PR and I also added some more commits to also update the template folder actions. Could you take another look?

njohner (approved):
LGTM

elioschmutz:
@njohner FYI, I added a new commit to also remove `attach_documents` in the templates-area.

njohner (approved):
LGTM

PR #6781, opened 2020-12-11
Add invitation_group_dn to teamraum policy template.

njohner:
The `invitation_group_dn` was not part of the policy templates for teamraum deployments. It might be unnecessary for certain setups where the LDAP itself adds newly created users to the correct group (as in SG), but IMO, I'd rather have that in the template and delete it from the created policy when necessary. In general this needs to be set for invitations to work correctly.

For https://4teamwork.atlassian.net/browse/CA-12

## Checklist (Must have)
- [x] Changelog entry

### DIFF
--- a/opengever/policytemplates/policy_template/.mrbob.ini
+++ b/opengever/policytemplates/policy_template/.mrbob.ini
@@ -208,6 +208,9 @@ deployment.workspace_users_group.question = Workspace users group
 deployment.workspace_users_group.default = tr_users
 deployment.workspace_users_group.required = False
 
+setup.invitation_group_dn.question = Invitation Group DN
+setup.invitation_group_dn.required = True
+
 setup.geverui.question = Enable new GEVER UI
 setup.geverui.required = True
 setup.geverui.default = true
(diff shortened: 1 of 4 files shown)


opengever/policytemplates/policy_template/.mrbob.ini
```diff
@@ -208,6 +208,10 @@ deployment.workspace_users_group.question = Workspace users group
 deployment.workspace_users_group.default = tr_users
 deployment.workspace_users_group.required = False
 
+setup.invitation_group_dn.question = Invitation Group DN
+setup.invitation_group_dn.default = cn=tr_users,ou=Groups,ou=...,dc=ch
```
lukasgraf, line 212: I'm not sure whether it's a good idea to default this with a DN that is technically invalid, and make it optional. Shouldn't this just be required, and essentially not have a default (because it will be different for every single deployment)?

njohner: yes works for me. I'll quickly test if I can have a required value that is only required for Teamraum templates.

lukasgraf (approved):
LGTM 👍

PR #6782, opened 2020-12-14
Enable nightly jobs in policy templates.

njohner:
This is necessary for all deployments with the gever-ui activated, because of the `NightlyRoleAssignmentReports`. so basically all new deployments...

For https://4teamwork.atlassian.net/browse/CA-12

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.16.0 (unreleased)
 ----------------------
 
+- Enable nightly jobs in teamraum policy templates. [njohner]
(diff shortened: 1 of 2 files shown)


njohner:
As discussed, this is too risky for now. We have no way of currently knowing whether the cronjobs are setup correctly for all our deployments and whether the nightly jobs get executed correctly or not. In case they don't, enabling nightly jobs will lead to dossier resolution jobs accumulating...
Instead I have opened a Jira issue that will first tackle the problem of making sure cronjobs get executed correctly on all deployments (https://4teamwork.atlassian.net/browse/CA-1346).

PR #6784, opened 2020-12-15
Add lock expansion, override unlock endpoint and add unlock file action

tinagerber:
Since it was only possible to remove a `STEALABLE LOCK` as creator with the @unlock endpoint until now, the endpoint had to be overwritten. Now a `lock_type` can be specified to determine which lock should be removed. Furthermore all users can remove the `copied_to_workspace` lock, not only the creator.

Furthermore, a lock expansion was added, so that it can be determined in the frontend with which `lock_type` an object is locked, in order to be able to remove this lock afterwards.

Additionally, an `unlock` file action was added to decide when a user can unlock a document.

Refined with @njohner 

Jira: https://4teamwork.atlassian.net/browse/NE-340

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:

### DIFF
--- a/opengever/api/lock.py
+++ b/opengever/api/lock.py
@@ -45,3 +45,23 @@ class Unlock(Service):
             alsoProvides(self.request, IDisableCSRFProtection)
 
         return lock_info(self.context)
+
+def can_safely_unlock(self, lock_type=STEALABLE_LOCK):
+    if not lock_type.user_unlockable:
+        return False
+
+    info = self.lock_info()
+    # There is no lock, so return True
+    if len(info) == 0:
+        return True
+
+    userid = getSecurityManager().getUser().getId() or None
+    for l in info:
+        # There is another lock of a different type
+        if not hasattr(l['type'], '__name__') or \
+           l['type'].__name__ != lock_type.__name__:
+            return False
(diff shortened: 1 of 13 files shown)


opengever/api/lock.py
```diff
@@ -64,4 +87,7 @@ def can_safely_unlock(self, lock_type=STEALABLE_LOCK):
         # The lock is in fact held by the current user
         if l['creator'] == userid:
             return True
+        # workspace lock can also be unlocked by users who have not created the lock
+        if lock_type.__name__ == LOCK_TYPE_COPIED_TO_WORKSPACE_LOCK:
```
lukasgraf, line 91: AND if they do have the `Modify portal content` permission on the object.

Currently, that's what the `@unlock` endpoint is protected with. But I'd like a test that makes sure the user actually needs `Modify portal content` to remove the workspace lock of another user.

lukasgraf (changes requested):

tinagerber:
@lukasgraf I have now added the "Change portal content" test. However, I did not write a test because it was not possible to implement it in efficient time. Can you rereview?

lukasgraf (approved):
LGTM 👍

PR #6788, opened 2020-12-16
Add feature flag for workspace meetings.

tinagerber:
This PR adds a feature flag for workspace meetings. This allows us to enable and disable WorkspaceMeetings. By default, this feature is enabled so that no additional update steps are required in the policies.
With the policy generator the feature can be enabled and disabled.

Jira: https://4teamwork.atlassian.net/browse/NE-290

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.16.0 (unreleased)
 ----------------------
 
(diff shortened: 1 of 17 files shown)


njohner (approved):
LGTM, although I'm not sure how happy I am that we feature flag every single content type in Teamraum...

PR #6789, opened 2020-12-17
Protect dossier participations endpoints with modify permission.

tinagerber:
Participations should not be able to be modified if a dossier cannot be modified.

Jira: https://4teamwork.atlassian.net/browse/NE-219

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/api/tests/test_dossier_participations.py
+++ b/opengever/api/tests/test_dossier_participations.py
@@ -254,6 +254,32 @@ class TestParticipationsPost(IntegrationTestCase):
              "type": "BadRequest"},
             browser.json)
 
+    @browsing
+    def test_post_participation_for_resolved_dossier_raises_unauthorized(self, browser):
+        self.login(self.regular_user, browser=browser)
+
+        with browser.expect_http_error(401):
+            browser.open(self.expired_dossier, view='/@participations',
+                         method='POST',
+                         headers=self.api_headers,
+                         data=json.dumps({'participant_id': self.valid_participant_id,
+                                          'roles': ['regard']}))
+        self.assertEqual({"message": "You are not authorized to access this resource.",
+                          "type": "Unauthorized"}, browser.json)
+
+    @browsing
+    def test_post_participation_for_inactive_dossier_raises_unauthorized(self, browser):
+        self.login(self.regular_user, browser=browser)
+
+        with browser.expect_http_error(401):
+            browser.open(self.inactive_dossier, view='/@participations',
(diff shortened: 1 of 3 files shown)


opengever/api/tests/test_dossier_participations.py
```diff
@@ -254,6 +254,19 @@ def test_post_participation_with_existing_participant_raises_bad_request(self, b
              "type": "BadRequest"},
             browser.json)
 
+    @browsing
+    def test_post_participation_for_expired_dossier_raises_unauthorized(self, browser):
```
njohner, line 258: Whether it's expired or not makes no difference here. What you want is test for a resolved dossier. the `expired_dossier` is the only resolved dossier in the fixture though. But for clarity I would change the test name. Then you can either leave the rest as is, or use the `resolvable_dossier` and resolve it in the test.

opengever/api/tests/test_dossier_participations.py
```diff
@@ -425,6 +451,18 @@ def test_delete_participations_with_invalid_participant_id_raises_bad_request(se
         self.assertEqual({"message": "chaosqueen is not a valid id",
                           "type": "BadRequest"}, browser.json)
 
+    @browsing
+    def test_delete_participation_for_offered_dossier_raises_unauthorized(self, browser):
```
njohner, line 455: Again whether the dossier is offered or not does not make a difference here...

njohner (commented):
does this work correctly in the classic UI already? Is there a test?

Otherwise LGTM.

tinagerber:
@njohner Why should I only test resolved dossiers? It's more about testing that you can't modify participations if the dossier can't be modified. This is also true for other states. 

It already works in the old UI.

njohner:
Expired is not a workflow state. An expired dossier is in the resolved state. So for that test I'd really rename it. IMO it's clearer that you are testing the "Modify" permission when you look at a resolved dossier. Otherwise I just wonder why the dossier should be expired in order for the POST to raise unauthorized.

If you want to test `offered`, sure why not. From a workflow point of view, a dossier can be edited until it is resolved or inactivated. Offering and then archiving it does not change the state of the dossier much and is not much used by any clients yet.

tinagerber:
Ah I didn't see that the ones in the fixtures are almost all in the "resolved" state, as there are definitely other states in the workflow. Then I will just test inactive and resolved dossiers.

njohner (approved):
LGTM

PR #6790, opened 2020-12-18
Prevent attempts to edit locked documents in Office Online.

tinagerber:
Until now, a document was editable with Office Online even if it was locked. The office_online_edit action was also available. This is no longer the case with this pr.
Jira: https://4teamwork.atlassian.net/browse/NE-446

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.16.0 (unreleased)
 ----------------------
 
+- Prevent attempts to edit locked documents in Office Online. [tinagerber]
 - Add feature flag for workspace meetings. [tinagerber]
 - Do not allow to modify the participations of a dossier via @participations endpoint if dossier cannot be modified. [tinagerber]
 - Fix unicode error in meeting overview. [njohner]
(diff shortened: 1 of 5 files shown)


lukasgraf:
@tinagerber did you test this with OfficeOnline?

tinagerber:
@lukasgraf No, since you can't easily set it up locally, I thought I could test it on the dev system once it's merged.

lukasgraf:
@tinagerber It's fine that you weren't able to test it locally. But it would be nice to mention that in the PR then, so we know 😉 

The reason I'm asking is that I suspect the following could be a problem:
- Editing with OfficeOnline checks out the document in a special collaborative mode (`collaborative=True`), where even other users are allowed to check out the document again (to join the editing session in OO)
- It's possible though that during the first checkout, the document is actually also getting locked (not sure, I will have to look at the code / test that)
-  That then could mean that with your change, those subsequent edit attempts would fail

I will have to look into this in detail on monday.

lukasgraf (approved):
As discussed and tested: LGTM! 👍 
(I missed the `if get_lock_token(self.context):` above your code that already handles the case of WOPI locking)

PR #6793, opened 2020-12-23
No longer allow to trash document templates.

tinagerber:
Changes in this PR:
- Change templatefolder workflow so that content can no longer be trashed
- Untrash all documents in the template folder that are trashed
Jira: https://4teamwork.atlassian.net/browse/NE-411

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)


opengever/core/upgrades/20201223134632_do_not_allow_to_trash_content_in_templatefolder/upgrade.py
```diff
@@ -0,0 +1,27 @@
...
+                                      reindex_security=False)
+
+    def delete_trashed_documents_in_template_folder(self):
+        catalog = api.portal.get_tool('portal_catalog')
+        portal_path = '/'.join(api.portal.get().getPhysicalPath())
+        templatefolder_query = {'path': {'query': portal_path, 'depth': 1},
+                                'portal_type': 'opengever.dossier.templatefolder'}
+        templatefolder_paths = [brain.getPath() for brain in catalog(templatefolder_query)]
+        for templatefolder_path in templatefolder_paths:
+            document_query = {'path': {'query': portal_path},
```
njohner, line 22: euh you are deleting all trashed documents, not just the ones in the templatefolders... I guess you wanted to use `templatefolder_path`

opengever/core/upgrades/20201223134632_do_not_allow_to_trash_content_in_templatefolder/upgrade.py
```diff
@@ -0,0 +1,27 @@
...
+                              'object_provides': 'opengever.document.behaviors.IBaseDocument',
+                              'trashed': True}
+            for obj in self.objects(document_query, 'Delete trashed documents in templatefolder'):
+                obj.aq_parent.manage_delObjects([obj.getId()])
+                print('Deleted obj')
```
njohner, line 27: Looks like a leftover

opengever/core/upgrades/20201223134632_do_not_allow_to_trash_content_in_templatefolder/upgrade.py
```diff
@@ -0,0 +1,27 @@
...
+        portal_path = '/'.join(api.portal.get().getPhysicalPath())
+        templatefolder_query = {'path': {'query': portal_path, 'depth': 1},
+                                'portal_type': 'opengever.dossier.templatefolder'}
+        templatefolder_paths = [brain.getPath() for brain in catalog(templatefolder_query)]
+        for templatefolder_path in templatefolder_paths:
+            document_query = {'path': {'query': portal_path},
+                              'object_provides': 'opengever.document.behaviors.IBaseDocument',
+                              'trashed': True}
+            for obj in self.objects(document_query, 'Delete trashed documents in templatefolder'):
+                obj.aq_parent.manage_delObjects([obj.getId()])
```
njohner, line 26: you could use `api.content.delete`...

njohner (commented):
I'm really not sure we want an upgrade step deleting documents.
Can you have a look at how many such trashed templates there are, or rather how many customers actually have trashed document templates? If the number is small enough, I'd rather opt for a maintenance script for the deletion.

njohner:
I also quickly discussed with @deiferni . On suggestion is to untrash the trashed template documents instead of deleting them, and then inform the client to check whether he wants to delete them or not.
But basically we need some sort of assessment of the current situation so that we can decide how to proceed...

tinagerber:
@njohner I checked the SaaS deployments on the ipet server, there are 4 deployments that have trashed documents in the template folder, so it probably affects a couple of deployments.

njohner:
I've checked Koeniz (0), Muri (1) and HBA (0), HEPVS(1), AI (140), Nidau (35), fmh (0)

njohner:
@tinagerber We had a quick chat with @deiferni and we don't want to delete the documents in an upgrade step. Instead untrash them. And please add a note in the version specific points that we should check before the update whether there are such documents and if there are a lot, discuss with the client in advance to clean that up (maybe delete them before the update).

tinagerber:
@njohner I untrashed the documents now in the upgrade step and added a note in https://4teamwork.atlassian.net/wiki/spaces/4TEAM/pages/123077675/GEVER+Releases+Versions-Spezifische+Punkte

njohner (approved):
LGTM

PR #6794, opened 2020-12-30
Implement solr indexer getObjPositionInParent for tasktemplates, todos and todolists

elioschmutz:
Jira: https://4teamwork.atlassian.net/browse/NE-283

Currently, we don't provide a solr-index for the position of an object within its parent for manually sorted content. This leads to workarounds i.e. for ordering tasktemplates (https://github.com/4teamwork/opengever.core/blob/master/opengever/tasktemplates/browser/tasktemplates.py#L72) or performance issues i.e. for displaying todos and todolists (https://github.com/4teamwork/gever-ui/blob/master/src/components/workspace/ToDoTab.vue#L144).

This PR implements the index `getObjPositionInParent` for solr. The indexer itself will only index the position for some whitelisted types which are `Todos`, `TodoLists` and `Tasktemplates`.

If the `ReorderedEvent` is triggered on an object, all its children will be reindexed. Even here, we do this only for a whitelisted set of containers which are `TodoLists`, `Workspaces` and `Tasktemplatefolders`. This will not unnecessary try to index children of objects we already know, that there will be no indexable children.

### Upgrade-Time
The Upgrade is defferable.

The most expensive task is to activate the objects.
Calculating the position within the parent takes about 3 Seconds for 10'000 objects. But we expect to have < 1000 objects per deployment. So the upgradestep should be blazing fast and there is no need to deffer the upgradestep.

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/base/handlers.py
+++ b/opengever/base/handlers.py
@@ -14,6 +14,9 @@ from opengever.dossier.utils import supports_is_subdossier
 from opengever.globalindex.handlers.task import sync_task
 from opengever.repository.interfaces import IRepositoryFolder
 from opengever.task.task import ITask
+from opengever.tasktemplates.content.templatefoldersschema import ITaskTemplateFolderSchema
+from opengever.workspace.interfaces import IToDoList
+from opengever.workspace.interfaces import IWorkspace
 from plone.app.workflow.interfaces import ILocalrolesModifiedEvent
 from Products.CMFPlone.interfaces import IPloneSiteRoot
 from sqlalchemy import and_
@@ -25,6 +28,10 @@ from zope.lifecycleevent.interfaces import IObjectMovedEvent
 from zope.sqlalchemy.datamanager import mark_changed
 
 reindex_after_copy = ['created']
+
+CONTAINERS_SUPPORTING_OBJ_POSITION_IN_PARENT = (IWorkspace,
+                                                IToDoList,
+                                                ITaskTemplateFolderSchema)
 
 
 def object_copied(context, event):
@@ -135,6 +142,15 @@ def object_modified(context, event):
 
     if should_track_touches(context):
         notify(ObjectTouchedEvent(context))
+
+
+def contents_reordered(context, event):
+    """Reindexes the 'getObjPositionInParent' index for each child object of the
+    context. We only want to reindex for containers which are orderable by the users.
+    """
+    if any(iface.providedBy(context) for iface in CONTAINERS_SUPPORTING_OBJ_POSITION_IN_PARENT):
+        for obj in context.listFolderContents():
+            obj.reindexObject(idxs=["getObjPositionInParent"])
 
 
 def update_favorited_repositoryfolder(context, event):
--- a/opengever/base/indexes.py
+++ b/opengever/base/indexes.py
@@ -6,15 +6,24 @@ from opengever.base.behaviors.translated_title import ITranslatedTitleSupport
 from opengever.base.interfaces import IReferenceNumber
 from opengever.base.model import SORTABLE_TITLE_LENGTH
 from opengever.bundle.sections.constructor import BUNDLE_GUID_KEY
+from opengever.tasktemplates.content.tasktemplate import ITaskTemplate
+from opengever.workspace.interfaces import IToDo
+from opengever.workspace.interfaces import IToDoList
 from plone.dexterity.interfaces import IDexterityContent
 from plone.i18n.normalizer.base import mapUnicode
 from plone.indexer import indexer
 from Products.CMFCore.interfaces import IFolderish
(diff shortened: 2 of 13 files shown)


opengever/base/handlers.py
```diff
@@ -137,6 +139,22 @@ def object_modified(context, event):
...
 
+def contents_reordered(context, event):
+    """Reindexes the 'getObjPositionInParent' index for each child object of the
+    context. We only want to reindex for containers which are orderable by the users.
+    """
+    whitelisted_context_interfaces = (IWorkspace, IToDoList)
+
+    # There is no specific interface for the tasktemplatefolder. So we have
+    # to check for the portaltype.
+    is_tasktemplatefolder = context.portal_type == 'opengever.tasktemplates.tasktemplatefolder'
```
deiferni, line 150: you should be able to use `ITaskTemplateFolderSchema`, no?

opengever/base/handlers.py
```diff
@@ -137,6 +139,22 @@ def object_modified(context, event):
         notify(ObjectTouchedEvent(context))
 
 
+def contents_reordered(context, event):
+    """Reindexes the 'getObjPositionInParent' index for each child object of the
+    context. We only want to reindex for containers which are orderable by the users.
+    """
+    whitelisted_context_interfaces = (IWorkspace, IToDoList)
```
deiferni, line 146: I'd prefer to have these in a constant somewhere in module scope. it is not something local that can change, but global "configuration". also a constant would be consistent to how it is done in other places and you could declare the intent by choosing a good constant name.

opengever/base/indexes.py
```diff
@@ -106,3 +110,19 @@ def watchers(obj):
...
+def getObjPositionInParent(obj):
+    """This indexer is only used by solr and only for a small set of whitelisted
+    interfaces. We only whitelist intefaces where we really need to sort by
+    position in parent.
+
+    The plone_catalog indexer uses a special index type called 'GopipIndex' which
+    is a fake index. So this function will not be used by the
+    plone catalogs 'getObjPositionInParent' index.
+    """
+    to_index = (ITaskTemplate, IToDo, IToDoList)
```
deiferni, line 125: same here, a constant holding this configuration would IMO be preferable.

opengever/base/handlers.py
```diff
@@ -137,6 +139,22 @@ def object_modified(context, event):
...
+
+    # There is no specific interface for the tasktemplatefolder. So we have
+    # to check for the portaltype.
+    is_tasktemplatefolder = context.portal_type == 'opengever.tasktemplates.tasktemplatefolder'
+
+    if any(iface.providedBy(context) for iface in whitelisted_context_interfaces) \
+       or is_tasktemplatefolder:
+        map(lambda obj: obj.reindexObject(idxs=["getObjPositionInParent"]),
```
deiferni, line 154: Ok … so this may be personal preference, but 
```python
for obj in context.listFolderContents():
    obj.reindexObject(idxs=["getObjPositionInParent"])
```
seems so much more readable to me. IMO `map` is fine if you can do something purely functional like `map(math.sin, xrange(3))` and if you're interested in the output data structure. Here we are just calling a method on an object in an object oriented manner.
I won't insist on this if you strongly disagree, but please consider changing it.

deiferni (approved):
LGTM 👍 , only some nagging.

opengever/base/handlers.py
elioschmutz: 😄 I agree... I think I coded too much JS 😉

opengever/base/handlers.py
elioschmutz: What's about: `INDEX_OBJ_POSITION_IN_PARENT_FOR`

opengever/base/indexes.py
elioschmutz: What's about: INDEX_OBJ_POSITION_IN_PARENT_FOR

opengever/base/handlers.py
elioschmutz: What's about: `USER_ORDERED_CONTAINERS`?

deiferni: or `CONTAINERS_SUPPORTING_OBJ_POSITION_IN_PARENT` 🤔

opengever/base/handlers.py
elioschmutz: Ah great! I wasn't able to find it 👍

elioschmutz:
@deiferni 👍 I updated the PR. Would you like to take a final look even if you have already approved it?

deiferni (approved):

PR #6796, opened 2021-01-06
Do not update touched date of children when moving an object.

njohner:
We do not want to update the touched date of the children of an object that is moved. This is both wrong and leads to errors in the indexing, probably because `IObjectMovedEvent` gets dispatched to all children and in the touched handler we propagate the update of touched to all parents...

For https://4teamwork.atlassian.net/browse/CA-1395

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2020.16.0 (unreleased)
 ----------------------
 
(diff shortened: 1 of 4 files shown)


deiferni (approved):
LGTM 👍

lukasgraf (approved):
LGTM 👍

PR #6798, opened 2021-01-06
Fix workflow transitions for tasktemplatefolders and tasktemplates over the restapi.

elioschmutz:
This PR fixes an issue where it was not possible to run workflow transitions for tasktemplatefoders and tasktemplates over the restapi.

The plone.restapi expects to have a configured workflow `reivew_history` to perform a transition: https://github.com/4teamwork/opengever.core/blob/master/opengever/api/transition.py#L97

If there is no such `review_history`, it will raise an error:

Bildschirmfoto 2021-01-06 um 16 32 22

We could have fixed this issue by making the code more robust against the missing `review_history`. But as discussed in the slack-channel with @deiferni https://4teamwork.slack.com/archives/C01BHEGBVU1/p1609946899012300, it's enough to just configure the workflow-definition properly for the two types.

ℹ️ I did not add a test for this case because I just changed the workflow configuration and we do not test missconfigurations.

Jira: https://4teamwork.atlassian.net/browse/NE-85

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations


opengever/api/tests/test_tasktemplates.py
```diff
@@ -0,0 +1,125 @@
+from ftw.testbrowser import browsing
+from opengever.testing import IntegrationTestCase
+from opengever.trash.trash import ITrashable
+import json
+
+
+class TestWorkflowPost(IntegrationTestCase):
+
+    @browsing
+    def test_transition_reassigning_task_and_revoking_users_permissions_can_be_executed(self, browser):
```
njohner, line 10: these tests are dupplicate, seem to be the same as in api/tests/test_workflow? Is that a copy/paste fail?

opengever/api/tests/test_tasktemplates.py
```diff
@@ -0,0 +1,125 @@
...
+        self.assertDictContainsSubset(
+            {u'title': u'task-state-in-progress',
+             u'actor': u'kathi.barfuss',
+             u'action': u'task-transition-reassign',
+             u'review_state': u'task-state-in-progress'},
+            browser.json
+        )
+
+    @browsing
+    def test_transition_close_task_and_revoking_users_permissions_can_be_executed(self, browser):
```
njohner, line 33: same here and below....

opengever/core/profiles/default/workflows/opengever_templatefolder_workflow/definition.xml
```diff
@@ -267,6 +267,6 @@
       <expression>state_change/getDateTime</expression>
     </default>
     <guard>
-  </guard>
+   </guard>
```
njohner, line 270: Still not aligned with `<guard>` tag

njohner (commented):
There is something fishy in the tests. Otherwise LGTM

opengever/api/tests/test_tasktemplates.py
elioschmutz: 😱 oh.... I somehow checked in this file 🙈 I'll remove the whole file.

As I wrote already in the PR description, I did not tend to add any test for this change because it's a configuration...

njohner (approved):
Still marked as conflicting but otherwise LGTM

PR #6802, opened 2021-01-08
Update changelog after hotfix release 2020.14.6.

njohner:
For https://4teamwork.atlassian.net/browse/CA-1395

## Checklist (Must have)
- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -10,7 +10,6 @@ Changelog
 2021.1.0 (2021-01-06)
 ---------------------
 
-- Do not update touched date of children when moving an object. [njohner]
 - Introduces a new solr-index 'getObjPositionInParent' for tasktemplates, todolists and todos. [elioschmutz]
 - Prevent attempts to edit locked documents in Office Online. [tinagerber]
 - Add feature flag for workspace meetings. [tinagerber]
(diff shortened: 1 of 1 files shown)


docs/HISTORY.txt
```diff
@@ -41,14 +40,14 @@ Changelog
 - Setup placeful workflow for workspace root in default content. [buchi]
 
 
-2020.15.1 (2020-12-03)
-----------------------
+2020.15.1 (2020-12-03) does not include 2020.14.6
```
lukasgraf, line 43: I wouldn't mess with the section headers in the changelog, TBH. This could possibly confuse `zest.releaser` or other automated tools.

Add it as the first "change" for this two sections? Like
```
- Does not include changes from hotfix release 2020.14.6
```

lukasgraf (changes requested):

docs/HISTORY.txt
njohner: We have modified the headers sometimes for such hotfixes, e.g. https://github.com/4teamwork/opengever.core/blob/master/docs/HISTORY.txt#L908. But I do like your proposal though.

lukasgraf: I see. Disregard my comment then ;)

lukasgraf (approved):

PR #6803, opened 2021-01-08
Update all translation files using bin/i18n-build.

lukasgraf:
Updates all translation files (`.po` / `.pot`) using `bin/i18n-build --all`.

For once, I also checked in the ones that contain no changes other than the updated timestamp, because they're a pain to filter out, and because I want clean slate before starting work on the English translations, even when it's just whitespace changes.

## Checklist (Must have)

- [ ] Changelog entry
    _(Deemed unnecessary)_


njohner (approved):
LGTM

PR #6804, opened 2021-01-08
Normalize how reponsible is stored in tasktemplates created over restapi and form.

elioschmutz:
Jira: https://4teamwork.atlassian.net/browse/NE-85

Currently, if you create a tasktemplate over the old UI with an interactive user for responsible, i.e. "Sachbearbeiter", it will save the value `current_user` for the `responsible` field and `interactive_user` for the responsible_client. For the field `issuer` it will just store: `current_user`.

Bildschirmfoto 2021-01-08 um 16 10 30

Bildschirmfoto 2021-01-08 um 16 11 06

but fetching it over the restapi will return:

Bildschirmfoto 2021-01-08 um 16 28 04

If we add the same tasktample over the restapi, it behaves completely different. Form the query-source, we get `interactive_users:current_user` as the value for the field:
Bildschirmfoto 2021-01-08 um 16 18 37

and if we add it, it will just save the whole string in the model:

Bildschirmfoto 2021-01-08 um 16 18 37

If we now look in the old UI, the it can no longer resolve the interactive user:

Bildschirmfoto 2021-01-08 um 16 23 28

because it stored the whole value and no client:

Bildschirmfoto 2021-01-08 um 16 30 17

In short, it's a mess...

This PR tries to fix this issue by introducing interactive actors.

An interactive actor can be either `interactive_actor:responsible` or `interactive_actor:current_user`. We now always store the whole actor-identifier for the `responsible` or the `issuer`. This allows us to remove the workaround of storing `interactive_users` in the `responsible_client` field to detect interactive users.

Bildschirmfoto 2021-02-01 um 14 22 27

By always using the actor-identifier for the `responsible`and the `issuer`, we're able to use the `Actor` class do directly check if a given user is an interactive actor or not.

ℹ️  prediscussed with @njohner 

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [x] api-change label added
- New functionality:
- DB-Schema migration
- New translations


opengever/ogds/base/actor.py
```diff
@@ -45,6 +46,19 @@
...
+}
+
+INTERACTIVE_ACTOR_CURRENT_USER = {
+    'id': 'interactive_actor:current_user',
+    'label': _(u'interactive_actor_current_user', default=u'Current user')
+}
+
+INTERACTIVE_ACTORS = [INTERACTIVE_ACTOR_RESPONSIBLE, INTERACTIVE_ACTOR_CURRENT_USER]
+INTERACTIVE_ACTORS_BY_ID = {actor.get('id'): actor for actor in INTERACTIVE_ACTORS}
+
```
njohner, line 61: Basically the only thing you need from all of this is a mapping between `id` and `label`, but I agree that it's cleaner to have the label mapping separate from the list of available IDs for interactive users. Still I think that `INTERACTIVE_ACTOR_RESPONSIBLE ` and `INTERACTIVE_ACTOR_CURRENT_USER ` are really superfluous and rather confusing as they are not `Actor`s. I feel naming could be improved and having a clear list of IDs and mapping to the labels is all you need and would be clearer IMHO:
```
INTERACTIVE_ACTOR_LABELS = {'interactive_actor:responsible': _(u'interactive_actor_responsible', default=u'Responsible'), 'interactive_actor:current_user': _(u'interactive_actor_current_user', default=u'Current user')} 
INTERACTIVE_ACTOR_IDS = INTERACTIVE_ACTOR_LABELS.keys()

File: opengever/ogds/base/actor.py [hunk 2]
```diff
@@ -451,6 +465,43 @@ def get_portrait_url(self):
...
 
 
+@implementer(IActor)
+class InteractiveActor(Actor):
+
+    css_class = 'actor-interactive-actor'
+    actor_type = 'interactive_actor'
+
+    def __init__(self, identifier):
+        actor = INTERACTIVE_ACTORS_BY_ID.get(identifier)
```
njohner, line 475: I find that a bit confusing not I have an `Actor` with an `actor` attribute. As proposed above here I would simply do:
```
        if not identifier in INTERACTIVE_ACTOR_IDS:
            raise ValueError('Interactive actor must be one of {}'.format(
                             ', '.join(INTERACTIVE_ACTOR_IDS)))
            self.identifier = identifier
```

opengever/ogds/base/actor.py
```diff
@@ -451,6 +465,43 @@ def get_portrait_url(self):
...
+        self.actor = actor
+
+    def corresponds_to(self, user):
+        return False
+
+    def get_profile_url(self):
+        return None
+
+    def get_label(self, with_principal=True):
+        return translate(self.actor.get('label'), context=getRequest())
```
njohner, line 490: Now here:
```
return translate(INTERACTIVE_ACTOR_LABELS[self.identifier], context=getRequest())
```

opengever/ogds/base/actor.py
```diff
@@ -471,6 +522,9 @@ def create_inbox_actor(self, org_unit=None):
     def is_team(self):
         return self.identifier.startswith('team:')
 
+    def is_interactive_actor(self):
+        return self.identifier in INTERACTIVE_ACTORS_BY_ID
```
njohner, line 526: `return self.identifier in INTERACTIVE_ACTOR_IDS`

opengever/ogds/base/locales/fr/LC_MESSAGES/opengever.ogds.base.po
```diff
@@ -36,6 +36,16 @@ msgstr "Oui"
 msgid "inbox_label"
 msgstr "Boîte de réception"
 
+#. Default: "Current user"
+#: ./opengever/ogds/base/actor.py
+msgid "interactive_actor_current_user"
+msgstr "Collaborateur"
```
njohner, line 42: "Utilisateur actif"

opengever/tasktemplates/sources.py
```diff
@@ -46,9 +45,10 @@ def search(self, query_string):
         return self.terms
 
     def _extend_terms_with_interactive_users(self, query_string):
-        for value, title in interactive_users().items():
-            if query_string.lower() in title.lower():
-                self.terms.insert(0, self.getTerm(value))
+        for actor_data in INTERACTIVE_ACTORS:
```
njohner, line 33: `for actor_id in INTERACTIVE_ACTOR_IDS:` ...

opengever/tasktemplates/sources.py
```diff
@@ -87,9 +76,10 @@ def search(self, query_string):
...
```
njohner, line 64: `for actor_id in INTERACTIVE_ACTOR_IDS`...

opengever/tasktemplates/browser/form.py
```diff
@@ -1,355 +0,0 @@
-from AccessControl import Unauthorized
```
njohner, line 1: Wow so that whole form was just used for interactive users mess?

opengever/core/upgrades/20210129153702_migrate_task_template_interactive_users/upgrade.py
```diff
@@ -0,0 +1,23 @@
...
+        query = {'object_provides': [
+            ITaskTemplate.__identifier__,
+        ]}
+
+        for obj in self.objects(query, 'Migrate tasktemplate interactive users'):
+            if obj.issuer in ['responsible', 'current_user']:
+                obj.issuer = 'interactive_actor:{}'.format(obj.issuer)
+                obj.reindexObject(idxs=['issuer'])
+
+            if obj.responsible_client == 'interactive_users':
```
njohner, line 20: I thought those created over the REST-API would have `interactive_users:...` in the `responsible` field but `None` in the `responsible_client` field no?

njohner (approved):
Overall LGTM 🎉 ! It's nice how much stuff you could remove from the code... 
just a few small things, mainly some namings in the Actors that I did find confusing.

And here some more suggestions:
- PR title: Equalize the way how we store the reponsible value for tasktemplates between the restapi and form creation. -> Normalize how reponsible is stored in tasktemplates created over restapi and form.
- One of the commit messages: Do no loner separate interactive users into respponsible_client and responsible -> No longer split up interactive users into respponsible_client and responsible

opengever/core/upgrades/20210129153702_migrate_task_template_interactive_users/upgrade.py
elioschmutz: yes, that's right. This is the upgradestep to migrate tasktemplates created over the old UI...

opengever/ogds/base/actor.py
elioschmutz: Hmm... not exactly...I basically needed the constants `INTERACTIVE_ACTOR_RESPONSIBLE` and `INTERACTIVE_ACTOR_CURRENT_USER` to address one specific interactive actor, i.e. here https://github.com/4teamwork/opengever.core/pull/6804/files#diff-c567bc658bc1ac26faa51b531bd1620244f208cfa552dfc6ae7b36a19f48cf8b

So if I would go with your suggestion, I would have to use magic strings everywhere to check what kind of interactive user it is to handle special cases with one specific interactive actor or if I want one specific ID of an interactive user.

One option to avoid the constants could be to add two methods to the `InteractiveActor` class: `is_responsible` and `is_current_user`. But I'm not sure if this is better...

what do you think?

elioschmutz (commented):
@njohner thanks for the review.

I'm not sure about your suggestion. See my comment. What do you think? Should I still go with your suggestion? And do you have a better idea for how to solve the resulting magic-strings if I remove the constants with the ID?

opengever/tasktemplates/browser/form.py
elioschmutz: is not registered nor used anywhere... I was a little confused, but i'ts very old. So I think we've just forgotten to remove it when we implemented a new form.

opengever/ogds/base/actor.py
elioschmutz: Discussed with @njohner over slack:

I use constants for the identifieres and then I go with the suggestion of @njohner

elioschmutz:
@njohner I pushed a new commit with your suggested change for the actor constants 👍

opengever/tasktemplates/browser/form.py
njohner: ah ok 👍

njohner (approved):
LGTM

PR #6805, opened 2021-01-08
Initialize English translations

lukasgraf:
This PR initializes the English translations based on information from the existing German `.po` files.

These new message catalogs for `en` were created by using the corresponding `.po` file from the`de` locale as a base, and modifying each block according to this strategy:
- Add a comment above the block with the value of the German translation.
  This is to provide context directly in the `.po` file for later, when we'll be filling in or checking English translations.

- If a `Default` is present, use that as the new translation (`msgstr`).
  Defaults in og.core are always natural language English sentences or words.

- If no default is present, use `msgid` as the translation.
  This might also be an English sentence, or it could be a `label_*` style identifier. In either case, the German translation is there as a comment to then later replace a label with a proper translation, or double-check a (probably questionable) existing translation.

As a whole, this operation should not result in any change as to how messages are translated into English (if English were to be enabled for the current codebase). It just materializes the translated messages into `.po` files.

However, there were a couple edge cases that needed fixing, where the existing default lookup / extraction didn't work properly. Details in commit messages.

Since a review based on "files changed" is impossible, here's a gist that's a compilation of all the diffs for the German `.po` files and their corresponding new English `.po` file: https://gist.github.com/lukasgraf/d4a3f06ed3110a6c6df6daca13915fa4

## Checklist (Must have)

- [x] Changelog entry


njohner (approved):
Nice, LGTM

PR #6807, opened 2021-01-10
Clean up English og.advancedsearch translations.

lukasgraf:
Clean up English og.advancedsearch translations.

### DIFF
--- a/opengever/advancedsearch/locales/en/LC_MESSAGES/opengever.advancedsearch.po
+++ b/opengever/advancedsearch/locales/en/LC_MESSAGES/opengever.advancedsearch.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
-"POT-Creation-Date: 2021-01-08 11:40+0000\n"
+"POT-Creation-Date: 2021-01-10 18:02+0000\n"
 "PO-Revision-Date: 2012-12-06 12:40+0100\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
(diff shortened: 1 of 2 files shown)


opengever/advancedsearch/locales/en/LC_MESSAGES/opengever.advancedsearch.po
```diff
@@ -176,11 +177,13 @@ msgstr "To"
...
 msgid "label_trashed"
-msgstr "Also search in the recycle bin"
+msgstr "Also search trashed content"
 
 #. German translation: lokal gespeicherte Aufgaben
+#. MARKER: admin-unit
 #: ./opengever/advancedsearch/advanced_search.py
 msgid "task"
-msgstr "task"
+msgstr "Tasks on this admin unit"
```
njohner, line 189: Makes sense to me, but then Maybe the German translation is bad?

njohner (approved):
LGTM

opengever/advancedsearch/locales/en/LC_MESSAGES/opengever.advancedsearch.po
lukasgraf: Yes, I'm not happy with that one either. It brings up the horror of having to find an English translation for `Mandant`, which will probably appear in many more translations down the line, where it can't easily be avoided (like it could here, by just calling them "local tasks").

_Tennant_ could possibly work, but _client_, as used in so many places, is definitely just plain wrong. So for this one I went with the technical description (admin unit), which is, though certainly confusing to the user without further explanation, not incorrect.

Once we come up with a canonical translation for "Mandant", it should be easy to search & replace across `.po` files. So in other words, I'm deferring that decision for now.

PR #6808, opened 2021-01-11
Change key for agenda item list document to "documents" in zip export.

njohner:
Grimlock wants the agenda item list document under the `documents` key and in a list and not as it is now under the `agenda_item_list` key (see @fguyer 's comment: https://4teamwork.atlassian.net/browse/CA-149?focusedCommentId=14380). This key is already in use in Grimlock for zips generated by RIS, but it's a new key in Gever, so displaying it should work in Grimlock without needing any changes.
Note that Grimlock expects items in the `documents` section to have a `title`, so we also need to add the title of the agenda item list in the Zip file.

Note that I tested the export and import in Grimlock with the new format.

For https://4teamwork.atlassian.net/browse/CA-1401

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 ---------------------
 
 - Bump docxcompose to version 1.3.1 to add support for dateformats. [njohner]
+- Change key for agenda item list document to "documents" in zip export. [njohner]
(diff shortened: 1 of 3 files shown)


lukasgraf (approved):
LGTM 👍

PR #6809, opened 2021-01-11
Fix indexing in Solr when setting document_type to None

njohner:
When `document_type` is set to `None`, the change does not get indexed in solr. This is fixed in https://github.com/4teamwork/ftw.solr/pull/177.

I will change the second commit to a version bump when `ftw.solr` is released.

For https://4teamwork.atlassian.net/browse/CA-1336

## Checklist (Must have)

- [x] Changelog entry


lukasgraf (approved):
LGTM! 👍  Nice and clean!

PR #6810, opened 2021-01-12
Add relatedItems field to todo

tinagerber:
ToDo now has a `relatedItems` field that can be used to refer to existing documents in the workspace.

Jira: https://4teamwork.atlassian.net/browse/NE-61

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- New translations

### DIFF
--- a/opengever/base/source.py
+++ b/opengever/base/source.py
@@ -219,3 +219,42 @@ class DossierPathSourceBinder(SolrObjPathSourceBinder):
         del source.navigation_tree_query['portal_type']
 
         return source
+
+
+class WorkspacePathSourceBinder(SolrObjPathSourceBinder):
+    """A Special PathSourceBinder wich only search in the workspace
+    of the actual context
+    """
+
+    def __init__(self, navigation_tree_query=None, filter_class=CustomFilter, **kw):
+        self.selectable_filter = filter_class(**kw)
+        self.navigation_tree_query = navigation_tree_query
+
+    def __call__(self, context):
+        """ gets workspace path and uses it in the navigation_tree_query """
+        workspace_path = ''
+        parent = context
+        while not IPloneSiteRoot.providedBy(parent) and \
+                parent.portal_type != 'opengever.workspace.root':
+            workspace_path = '/'.join(parent.getPhysicalPath())
+            parent = aq_parent(aq_inner(parent))
+        if not self.navigation_tree_query:
+            self.navigation_tree_query = {}
+
+        self.navigation_tree_query['path'] = {'query': workspace_path}
+
+        # Extend path in selectable_filter, to make sure only objects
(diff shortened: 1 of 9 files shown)


opengever/base/source.py
```diff
@@ -219,3 +219,43 @@ def __call__(self, context):
...
+    """A Special PathSourceBinder wich only search in the workspace
+    of the actual context
+    """
+
+    def __init__(self, navigation_tree_query=None, filter_class=CustomFilter, **kw):
+        self.selectable_filter = filter_class(**kw)
+        self.navigation_tree_query = navigation_tree_query
+
+    def __call__(self, context):
+        """ gets workspace path and put it to the navigation_tree_query """
```
njohner, line 234: gets workspace path and uses is in the navigation_tree_query

opengever/base/source.py
```diff
@@ -219,3 +219,43 @@ def __call__(self, context):
...
+
+        source = self.path_source(
+            context,
+            selectable_filter=self.selectable_filter,
+            navigation_tree_query=self.navigation_tree_query)
+
+        # The path source bases on the navtree strategy, which adds a
+        # portal_type query option, which disables all types not-to-list
+        # in the navigation. This is not a navigation - so remove this
+        # limitation.
```
njohner, line 258: Well, that comment is super unclear. I see it's copied from somewhere else, but I feel like we could improve that a bit. I've browsed the code a bit and I think what is meant is:
```
The PathSource builds its query using the navtree.QueryBuilder, which adds a portal_type filter to only show certain portal_types. This is not desired here, so we remove that part of the query
```

opengever/workspace/tests/test_todo.py
```diff
@@ -170,6 +170,22 @@ def test_read(self, browser):
         self.assertEqual(u'opengever_workspace_todo--STATUS--active',
                          browser.json['review_state'])
 
+    @browsing
+    def test_document_related_with_todo(self, browser):
+        self.login(self.workspace_member, browser)
+        self.set_related_items(self.todo, [self.workspace_document])
+        browser.open(self.todo, method='GET', headers=self.api_headers)
+        self.assertEqual(200, browser.status_code)
+
```
njohner, line 179: Add a check that there is only one related item:
`self.assertEqual(1, browser.json['relatedItems'])`

njohner (commented):
I guess this is also supposed to work over the API? Does it get validated? Or can I set a relation to any document over the API? Can you add a test for that?

Otherwise LGTM!

tinagerber:
> I guess this is also supposed to work over the API? Does it get validated? Or can I set a relation to any document over the API? Can you add a test for that?
> 

@njohner I added a test

njohner (approved):
LGTM

PR #6819, opened 2021-01-13
Clean up English og.dossier translations.

lukasgraf:
Clean up English `og.dossier` translations.

There's a few messages in `opengever.dossier.po` around resolving and (re)activating dossiers that are quite clunky, but are used in **REST API responses** (error messages for workflow transitions). Found that out the hard way (already changed them, got weird test failures 😭 ).

I aborted those changes and left them untouched, and instead added a cautionary marker. We certainly don't want to push breaking API changes when it's "just" translations that are supposed to be updated.

The second commit is purely casing changes. From "magic eight ball casing" to "sentence casing". Should make it a bit easier to review having those separately.

### DIFF
--- a/opengever/dossier/locales/en/LC_MESSAGES/opengever.dossier.po
+++ b/opengever/dossier/locales/en/LC_MESSAGES/opengever.dossier.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
-"POT-Creation-Date: 2021-01-08 11:40+0000\n"
+"POT-Creation-Date: 2021-01-13 22:26+0000\n"
 "PO-Revision-Date: 2017-05-23 04:53+0000\n"
 "Last-Translator: Jacqueline Sposato <jacqueline.sposato@gmail.com>\n"
 "Language-Team: German <https://translations.onegovgever.ch/projects/onegov-gever/opengever-dossier/de/>\n"
@@ -19,7 +19,7 @@ msgstr ""
 #. German translation:  (nächtliche Abschlussarbeiten ausstehend)
 #: ./opengever/dossier/viewlets/byline.py
 msgid " (after resolve jobs pending)"
-msgstr " (after resolve jobs pending)"
+msgstr " (nightly post-resolve jobs pending)"
 
 #. German translation:  (wird gerade abgeschlossen)
 #: ./opengever/dossier/viewlets/byline.py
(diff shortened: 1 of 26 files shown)


opengever/dossier/locales/en/LC_MESSAGES/opengever.dossier.po
```diff
@@ -559,51 +568,53 @@ msgstr "Move Items"
...
 #: ./opengever/dossier/move_items.py
 msgid "help_destination"
-msgstr "Live Search: search the Plone Site"
+msgstr "Live search: Search this site"
 
 #. German translation: Geben Sie die vollständige Ablagenummer an.
 #: ./opengever/dossier/filing/advanced_search.py
 msgid "help_filing_number"
-msgstr "help_filing_number"
+msgstr "Exact filling number"
```
njohner, line 592: Complete filing number?

opengever/dossier/locales/en/LC_MESSAGES/opengever.dossier.po
```diff
@@ -559,51 +568,53 @@ msgstr "Move Items"
...
 msgid "help_number_of_containers"
-msgstr "help_number_of_containers"
+msgstr "Number of physical containers (for large files)"
 
 #. German translation: Neuen Kommentar erfassen.
 #. Default: "Create a new note"
 #: ./opengever/dossier/viewlets/templates/note.pt
 msgid "help_text_add_note"
-msgstr "Create a new note"
+msgstr "Add a new note"
```
njohner, line 611: I wonder whether we should be using "comment" instead of "note", just to be closer to the German

njohner (approved):
LGTM. 
Concerning "note" vs "comment", I leave it up to you whether we want to stick more closely to the German or not.

opengever/dossier/locales/en/LC_MESSAGES/opengever.dossier.po
```diff
@@ -187,102 +187,109 @@ msgstr "Subdossier"
...
 #: ./opengever/dossier/archive.py
 msgid "The Dossier has been resolved"
-msgstr "The Dossier has been resolved"
+msgstr "Dossier has been resolved"
 
 #. German translation: Das Dossier ${dossier} hat ein ungültiges Enddatum.
+#. MARKER: used-in-api-response
 #: ./opengever/dossier/resolve.py
 msgid "The dossier ${dossier} has a invalid end_date"
 msgstr "The dossier ${dossier} has a invalid end_date"
```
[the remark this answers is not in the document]
lukasgraf: Yeah, I know. That's one of those that I had to revert because they're used in API responses. Fixing it would cause a breaking API change.

opengever/dossier/locales/en/LC_MESSAGES/opengever.dossier.po
lukasgraf: TBH, the German messages are the ones that feel a bit wonky here. (And they probably should be using "Notizen" more consistently)

njohner: Works for me, maybe leave a note behind to look at the German in that case?

PR #6832, opened 2021-01-17
Clean up English og.core translations.

lukasgraf:
Clean up English `og.core` translations.

### DIFF
--- a/opengever/core/locales/en/LC_MESSAGES/opengever.core.po
+++ b/opengever/core/locales/en/LC_MESSAGES/opengever.core.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
-"POT-Creation-Date: 2021-01-08 11:40+0000\n"
+"POT-Creation-Date: 2021-01-17 16:16+0000\n"
 "PO-Revision-Date: 2017-09-18 15:21+0200\n"
 "Last-Translator: Philippe Gross <p.gross@4teamwork.ch>\n"
 "Language-Team: 4teamwork <info@4teamwork.ch>\n"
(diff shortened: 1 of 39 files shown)


opengever/core/locales/en/LC_MESSAGES/plone.po
```diff
@@ -24,19 +24,19 @@ msgstr "Active"
 
 #. German translation: Verwalten
 msgid "DossierManager"
-msgstr "DossierManager"
+msgstr "Manage dossiers"
```
lukasgraf, line 28: I've been quite defensive about touching too many (translations of) roles names here in the `plone` domain. They _should_ only affect display to the user, but since we're also using those for setup via Excel and Lawgiver (which I think is not even visible for users for the most part), I tried to touch only those that really needed it.

opengever/core/locales/en/LC_MESSAGES/opengever.core.po
```diff
@@ -357,12 +359,12 @@ msgstr "Task"
...
 #: ./opengever/core/profiles/default/types/opengever.tasktemplates.tasktemplate.xml
 msgid "TaskTemplate"
-msgstr "TaskTemplate"
+msgstr "Task Template"
 
 #. German translation: Standardablauf
 #: ./opengever/core/profiles/default/types/opengever.tasktemplates.tasktemplatefolder.xml
 msgid "TaskTemplateFolder"
-msgstr "TaskTemplateFolder"
+msgstr "Task Template Folder"
```
njohner, line 367: I've always hated that translation. Can you add a marker to have another look at this one? I think it should be something more in the line of `Process template`

njohner, line 367: I've always hated that translation. Can you add a marker to have another look at this one? I think it should be something more in the line of `Process template`

njohner (approved):
LGTM

opengever/core/locales/en/LC_MESSAGES/opengever.core.po
lukasgraf: That makes two of us. 😉  Will do.

PR #6835, opened 2021-01-18
Clean up English og.trash translations.

lukasgraf:
Clean up English `og.trash` translations.

### DIFF
--- a/opengever/trash/locales/en/LC_MESSAGES/opengever.trash.po
+++ b/opengever/trash/locales/en/LC_MESSAGES/opengever.trash.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: opengever.trash 1.0\n"
-"POT-Creation-Date: 2021-01-08 11:40+0000\n"
+"POT-Creation-Date: 2021-01-18 09:44+0000\n"
 "PO-Revision-Date: 2017-05-03 09:38+0200\n"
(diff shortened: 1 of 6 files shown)


opengever/trash/locales/en/LC_MESSAGES/opengever.trash.po
```diff
@@ -120,20 +120,20 @@ msgstr "The document is still checked out."
...
 #: ./opengever/trash/remover.py
 msgid "msg_is_removed"
-msgstr "The document is already removed."
+msgstr "The document has already been removed."
 
 #. German translation: Löschen
 #: ./opengever/trash/upgrades/profiles/4100/actions.xml
 msgid "remove"
-msgstr "remove"
+msgstr "Remove"
```
njohner, line 134: in `og.core` I think you replaced `Remove` with `Delete`. Well on the other hand it's a translation that only appears in an upgrade step xml, so it does not matter ...

njohner (approved):
LGTM

opengever/trash/locales/en/LC_MESSAGES/opengever.trash.po
lukasgraf: Yes, it's possible that this is not entirely consistent across all contexts yet (neither is it in German, BTW 😉 ).
But I think that's something that much easier (time-efficient) to iron out once it's shipped and we can just switch the language to `en-us` for local development. At least I plan to do that, so we'll certainly catch some oddities that way.

PR #6842, opened 2021-01-18
Fix loading next batch in gallery view

buchi:
We need to configure the current page in the table source as Solr only fetches the requested batch. Before with portal catalog we were able to access all results.

JIRA: https://4teamwork.atlassian.net/browse/CA-1160

## Checklist (Must have)

- [x] Changelog entry


njohner (approved):
LGTM

PR #6849, opened 2021-01-19
Add is_task_addable_in_main_task and is_task_addable_before attributes to @tasktree endpoint, sort on getObjPositionInParent

tinagerber:
In the new frontend, it should be possible to create new subtasks in a sequential process between the subtasks. Therefore, each subtask must know whether new subtasks can be added to the main task.

- To decide whether a subtask can be added at the end, the field `is_task_addable_in_main_task` is required.
- To decide whether a subtask can be added before another subtask, the field `is_task_addable_before` is required.
- Until now, the @tasktree endpoint returned the task sorted by creation date. Since this means that those created later do not appear in the correct position for sequential tasks, they are now sorted according to getObjPositionInParent for sequential tasks. This first had to be indexed.

Jira: https://4teamwork.atlassian.net/browse/NE-315

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [x] api-change label added

### DIFF
--- a/docs/public/dev-manual/api/tasks.rst
+++ b/docs/public/dev-manual/api/tasks.rst
@@ -484,6 +484,8 @@ Angaben zum übergeordneten Dossier einer Aufgabe ist in der GET Repräsentation
       }
 
 
+.. _tasktree:
+
 Aufgabenhierarchie
 -------------------
 Zu einer Aufgabe kann die Aufgabenhierarchie bestehend aus Hauptaufgabe und allen Unteraufgaben abgefragt werden.
--- a/docs/public/dev-manual/api/tasks.rst
+++ b/docs/public/dev-manual/api/tasks.rst
@@ -487,7 +487,7 @@ Angaben zum übergeordneten Dossier einer Aufgabe ist in der GET Repräsentation
 Aufgabenhierarchie
 -------------------
 Zu einer Aufgabe kann die Aufgabenhierarchie bestehend aus Hauptaufgabe und allen Unteraufgaben abgefragt werden.
-Dazu steht ein spezifischer Endpoint `@tasktree` zur Verfügung.
+Dazu steht ein spezifischer Endpoint `@tasktree` zur Verfügung. Die Aufgaben werden nach Erstelldatum sortiert zurückgeliefert. Bei sequenziellen Aufgabenabläufen werden die Aufgaben nach Aufgabenfolge sortiert.
 
 **Beispiel-Request**:
 
--- a/docs/public/dev-manual/api/tasks.rst
+++ b/docs/public/dev-manual/api/tasks.rst
@@ -521,7 +521,8 @@ Dazu steht ein spezifischer Endpoint `@tasktree` zur Verfügung.
             "review_state": "task-state-in-progress",
             "title": "Eine Aufgabe"
           }
-        ]
+        ],
+        "is_task_addable_in_main_task": true
       }
 
 Die Aufgabenhierarchie kann auch direkt über den GET-Request eines Tasks mittels Expansion angefordert werden.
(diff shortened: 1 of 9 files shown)


njohner (commented):
I don't think you will have the information you want with your change. Adding new subtasks in a sequential task depends on where you want to add it. See https://github.com/4teamwork/opengever.core/blob/master/opengever/task/browser/templates/overview.pt#L105-L110

tinagerber:
@njohner We also want to be able to create new subtasks while the context is a subtask of the main task, not only when the context is the main task. We already have the rest of the information needed to determine if a subtask is addable in the new frontend.

njohner:
Yes I understood you want to be able to add new subtasks from within a subtask, that is fine and will be possible if you add the information in the `task_tree`. But subtasks cannot be added anywhere in a sequential task, it depends on the review states of the subtasks. I don't want the frontend to decide that on its own.
If the only thing you want to know is whether basically tasks are addable in a main task, then add that to the task serialization, not the `task_tree`.
We can have a 📞 if you want

tinagerber:
@njohner I have changed the PR as we discussed on the phone.
I also indexed `getObjPositionInParent` for sequential tasks to sort on in the @tasktree endpoint. Do you want to do a review?

docs/public/dev-manual/api/tasks.rst
```diff
@@ -532,6 +532,8 @@ Die Aufgabenhierarchie kann auch direkt über den GET-Request eines Tasks mittel
      GET http://example.org/ordnungssystem/fuehrung/dossier-1/task-1?expand=tasktree HTTP/1.1
      Accept: application/json
 
+Für sequenzielle Aufgaben steht zusätzlich das Feld ``is_task_addable_before`` zur Verfügung.
```
njohner, line 537: sequenzielle Aufgaben -> sequenzielle Standardabläufe?

opengever/api/tasktree.py
```diff
@@ -40,6 +42,9 @@ def is_task_addable_in_main_task(self):
                 return True
         return False
 
+    def is_task_addable_before(self, obj):
+        return (obj.review_state() == TASK_STATE_PLANNED) and self.is_task_addable_in_main_task()
```
njohner, line 46: Are you sure that condition is correct?
I was just testing a bit with sequential tasks and its a huge mess. We can get into situations with mutliple open tasks and what not. See for example https://dev.onegovgever.ch/fd/ordnungssystem/fuehrung/gemeinderecht/erlasse-der-gemeinde/gemeindeordnung/dossier-1282/task-858
But it's not your job to fix that. I don't like that we use two different conditions between frontend and backend, but backend seems broken anyway. So I'd say go ahead with that for now, and I'll open a story in the backend...

njohner (approved):
I'm half happy, but it's not your fault. So LGTM.

docs/public/dev-manual/api/tasks.rst
tinagerber: I am never sure if "Standardablauf" only refers to the template or also to the tasks.

opengever/api/tasktree.py
tinagerber: I also tested in the old frontend and it is definitely wrong there. With these conditions, as far as I know, it should no longer be possible to have multiple open tasks.

njohner: I created a story: https://4teamwork.atlassian.net/browse/CA-1429

docs/public/dev-manual/api/tasks.rst
tinagerber: I have now called it "Aufgabenabläufe".

PR #6850, opened 2021-01-19
Rename field label 'Public access level' to 'Disclosure status'.

lukasgraf:
Rename field label 'Public access level' to 'Disclosure status'.

According to feedback during review, see
https://www.termdat.bk.admin.ch/Entry/EntryDetail/473650?strLanguages=2-6-7-8-3-0


njohner (approved):
LGTM

PR #6851, opened 2021-01-19
List informed principals in TaskAddedActivity description.

njohner:
We add the list of informed principals to the description of the `TaskAddedActivity`.
I also decided to align the labels to the top of the cells, which makes it much more readable IMHO. 
The screenshot below shows the E-mail for a created Task with a long title, description and several informed principals:

Screenshot 2021-01-19 at 16 36 33

For https://4teamwork.atlassian.net/browse/CA-1227

## Checklist (Must have)
- [x] Changelog entry

### DIFF
--- a/opengever/task/activities.py
+++ b/opengever/task/activities.py
@@ -103,6 +103,8 @@ class TaskAddedActivity(BaseTaskActivity):
     def collect_description_data(self, language):
         """Returns a list with [label, value] pairs.
         """
+        informed_principals = [Actor.lookup(actorid).get_label() for actorid in
+                               self.context.informed_principals]
         data = [
             [_('label_task_title', u'Task title'), self.context.title],
             [_('label_deadline', u'Deadline'),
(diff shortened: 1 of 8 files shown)


opengever/task/activities.py
```diff
@@ -103,6 +103,8 @@ def render_description_markup(self, data, language):
     def collect_description_data(self, language):
         """Returns a list with [label, value] pairs.
         """
+        informed_principals = [Actor.lookup(actorid).get_label() for actorid in
```
deiferni, line 106: will this behave well for failed lookups?

deiferni (approved):
Don't we have tests that test the actual mail rendering? If we have them i would suggest you also amend a test there.

Apart from that, LGTM 👍 .

opengever/task/activities.py
njohner: Yes that should behave well. `lookup` will return a `NullActor`, which returns the identifier or empty string as label (https://github.com/4teamwork/opengever.core/blob/master/opengever/ogds/base/actor.py#L159-L160)

njohner:
You're right we had test for the mail content. I added a test to make sure it contains the `info at`

PR #6852, opened 2021-01-20
Respect active languages languages in WorkspaceRoot and PrivateRoot forms.

njohner:
Some of the tests for the translated titles were not testing anything. Indeed they were asserting equality between empty lists, probably since we started using combined language codes like 'de-ch'.

We also improve the tests by really checking that translated title fields are only shown for the active languages both in the Add and the Edit forms. This led to discovery that both `WorkspaceRoot` and `PrivateRoot` were not handling the translated fields correctly. I therefore also fix this with this PR.

For https://4teamwork.atlassian.net/browse/CA-1040

## Checklist (Must have)
- [x] Changelog entry

### DIFF
--- a/opengever/base/tests/test_translated_title.py
+++ b/opengever/base/tests/test_translated_title.py
@@ -100,6 +100,36 @@ class TestTranslatedTitleFieldsInEditForms(IntegrationTestCase):
             browser, ['de', 'fr'])
 
     @browsing
+    def test_modifying_language_fields_in_edit_form(self, browser):
+        self.login(self.manager, browser=browser)
+        lang_tool = api.portal.get_tool('portal_languages')
+        lang_tool.supported_langs = ['de-ch', 'fr-ch']
+
+        browser.open(self.repository_root, view='edit')
+
+        browser.fill({'Title (German)': u'Ordnungssystem',
+                      u'Title (French)': u"syst\xe8me d'ordre"})
+        browser.find('Save').click()
+
+    @browsing
+    def test_label_is_renamed_to_title_for_sites_with_only_one_active_language(self, browser):
+        self.login(self.manager, browser=browser)
+        lang_tool = api.portal.get_tool('portal_languages')
+        lang_tool.supported_langs = ['de-ch', 'fr-ch']
+
+        browser.open(self.repository_root, view='edit')
+        self.assertEquals(
+            'Title (French)',
+            browser.css('label[for=form-widgets-ITranslatedTitle-title_fr]').first.text)
+
+        lang_tool.supported_langs = ['fr-ch']
+
+        browser.open(self.repository_root, view='edit')
+        self.assertEquals(
+            'Title',
+            browser.css('label[for=form-widgets-ITranslatedTitle-title_fr]').first.text)
+
+    @browsing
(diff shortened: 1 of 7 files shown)


opengever/base/tests/test_translated_title.py
```diff
@@ -59,18 +59,24 @@ def test_has_translation_behavior(self):
...
             if lang in TranslatedTitle.SUPPORTED_LANGUAGES
         ]
 
         first_fields = [
             _input.name
             for _input in browser.css("#form input[type!='hidden']")[:len(translated_fields)]
         ]
-        self.assertEquals(translated_fields, first_fields)
+        self.assertEqual(num_fields, len(first_fields))
+        self.assertItemsEqual(translated_fields, first_fields)
```
deiferni, line 79: So this will no longer assert on the order of items, but i guess we don't really care what oder the languages appear in 🤔 ?

opengever/base/tests/test_translated_title.py
```diff
@@ -84,7 +90,7 @@ def test_translated_title_is_on_top_when_editing_the_inbox_container(
         browser.open(inbox_container, view='edit')
         statusmessages.assert_no_error_messages()
 
-        self.assert_translated_title_fields_in_browser(browser)
+        self.assert_translated_title_fields_in_browser(browser, 1)
```
deiferni, line 93: can this ever be anything other than 1? If i understand correctly we only assert on the title field and we make sure there is always one field, right? If that is correct i would not pass the parameter but instead hard-code the assertion in the custom assertion method.

The literal `1` in the method call is also not self explanatory when i just read the method call line, i'll have to go and look at the method definition to understand what's going on. If you want to keep the parameter i'd suggest to make sure one can understand what's going on by either making it a keyword argument or by slightly renaming the assertion method.

deiferni (approved):
LGTM, just some nagging about readability. Nothing that blocks a merge.

opengever/base/tests/test_translated_title.py
njohner: Yeah, I don't think we want to assert the order. But I guess I could if you want. The order should be fix, but I don't think it matters.

opengever/base/tests/test_translated_title.py
njohner: No it does not have to be 1. If more than one language is enabled, then the parameter will be larger than one.
As for the parameter naming, I agree with you, I'll improve

opengever/base/tests/test_translated_title.py
deiferni: Just a double check that it is a deliberate change. If it does not matter in your opinion that's fine 👍 .

njohner:
@deiferni That PR has fundamentally changed from only being a fix of the tests to also including fixing the behavior of `PrivateRoot` and `WorkspaceRoot`.  I had plan to fix those in the PR activating English, but as I improved the tests further here, they started failing for these two content types, so I decided to include this in this PR. So please have another look at the whole thing. Sorry about that!

deiferni (approved):
LGTM 👍 

_IIRC we did not implement support for these content types as they are only editable by a `Manager`, and thus never actually edited by users and only setup programmatically._

PR #6856, opened 2021-01-20
Fix deactivating committees with canceled meeting.

deiferni:
With this PR we fix deactivating committees with canceled meetings inside.

The query used in the guard to detect if the committee can be deactivated did not consider canceled meetings.

Jira: https://4teamwork.atlassian.net/browse/CA-1397

## Checklist (Must have)

- [x] Changelog entry


njohner (approved):
LGTM

PR #6857, opened 2021-01-21
Implement POST @notifications endpoint to mark notifications as read.

tinagerber:
We want to be able to mark all notifications as read in the new frontend. For this purpose, the POST method for the @notifications endpoint is implemented. Specified by @deiferni 

Jira: https://4teamwork.atlassian.net/browse/NE-185

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:

### DIFF
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -18,6 +18,7 @@ Other Changes
 
 - The field ``custom_properties`` is now included in the ``@schema`` endpoint for Documents and Mails (see :ref:`content-types`).
 - ``@tasktree``: Attributes ``is_task_addable_in_main_task`` and ``is_task_addable_before`` added (see :ref:`docs <tasktree>`).
+- ``@notifications``: request method POST is added to mark all notifications as read (see :ref:`docs <mark-notifications-as-read>`).
 
 
 2021.2.0 (2021-01-20)
--- a/opengever/api/notifications.py
+++ b/opengever/api/notifications.py
@@ -1,9 +1,12 @@
 from opengever.activity import notification_center
+from opengever.activity.model.notification import Notification
 from opengever.readonly import is_in_readonly_mode
 from plone import api
 from plone.restapi.batching import HypermediaBatch
 from plone.restapi.deserializer import json_body
 from plone.restapi.services import Service
+from sqlalchemy.sql.expression import false
+from sqlalchemy.sql.expression import true
 from zExceptions import BadRequest
 from zExceptions import NotFound
 from zExceptions import Unauthorized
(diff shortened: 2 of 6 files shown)


docs/public/dev-manual/api/api_changelog.rst
```diff
@@ -18,6 +18,7 @@ Other Changes
 
 - The field ``custom_properties`` is now included in the ``@schema`` endpoint for Documents and Mails (see :ref:`content-types`).
 - ``@tasktree``: Attributes ``is_task_addable_in_main_task`` and ``is_task_addable_before`` added (see :ref:`docs <tasktree>`).
+- ``@notifications``: request method POST is added to mark notifications as read (see :ref:`docs <mark-notifications-as-read>`).
```
njohner, line 21: to mark notifications -> to mark all notifications

Marking one notification as read was already possible with the PATCH...

opengever/api/notifications.py
```diff
@@ -137,3 +140,53 @@ def read_params(self):
                              "as path parameter.")
 
         return self.params
+
+
+class NotificationsPost(Service):
+    """API endpoint to mark notifications as read.
```
njohner, line 146: to mark notifications -> to mark all notifications

opengever/api/notifications.py
```diff
@@ -137,3 +140,53 @@ def read_params(self):
...
+        self.params = []
+
+    def publishTraverse(self, request, name):
+        # Consume any path segments after /@notifications as parameters
+        self.params.append(name)
+        return self
+
+    def reply(self):
+        userid, = self.read_params()
+        if not json_body(self.request).get('mark_all_notifications_as_read'):
```
njohner, line 168: That is a bit mean. I should be allowed to pass `mark_all_notifications_as_read: False` in the request body, although it does not make much sense no? Or maybe simply adapt the error message to be more specific and say that `mark_all_notifications_as_read=True` is required for the endpoint.

opengever/api/notifications.py
```diff
@@ -137,3 +140,53 @@ def read_params(self):
...
+            raise BadRequest("Property 'mark_all_notifications_as_read' is required.")
+
+        latest_notification_id = json_body(self.request).get('latest_client_side_notification')
```
njohner, line 171: I'd add a check that the latest_notification_id corresponds to a vaild notification for the current user.

opengever/api/tests/test_notifications.py
```diff
@@ -269,3 +269,85 @@ def test_raises_unauthorized_when_accessing_notification_of_other_user(self, bro
...
+        self.login(self.regular_user, browser=browser)
+        url = '{}/@notifications/{}'.format(self.portal.absolute_url(), self.regular_user.getId())
+        with browser.expect_http_error(400):
+            browser.open(url,
+                         method='POST',
+                         headers=self.api_headers,
+                         data=json.dumps({'latest_client_side_notification': 0}))
+        self.assertEqual({"message": "Property 'mark_all_notifications_as_read' is required.",
+                          "type": "BadRequest"}, browser.json)
+
```
njohner, line 338: I'd add a test with `'mark_all_notifications_as_read=False` just so that we have that behavior specified

njohner (commented):
A few small comments, but otherwise LGTM

opengever/api/notifications.py
tinagerber: I don't think `mark_all_notifications_as_read: False` makes much sense, because otherwise we have an endpoint that does nothing. This will first make sense when you can create new notifications with POST @notifications or something. I will adjust the error message.

opengever/api/tests/test_notifications.py
tinagerber: I updated the error message in the test.

tinagerber:
@njohner I updated the PR. Can you take another look?

njohner (approved):
LGTM

PR #6859, opened 2021-01-21
Add English support for translated titles.

njohner:
With this PR we add `title_en` to the `ITranslatedTitle` behavior. 

The main issue encountered was with tests, as currently tests were using `en` as language (browser preferred language), but translated titles would default to German because the English title did not exist. The end result is that the path are based on the German titles but normalized with the English normaliser. So leaving the English titles in the test fixture was not an option because it made the objects invalid (could not be edited), adding them would lead to the English title being used for generating the path. Testing in German was also not an option, because in that case the German normalizer would be used, also affecting the path. In the end, we simply set all the English titles equal to the German ones, as it was the only way to have valid objects while preserving their paths.

Some of the tests are touched twice, as the decision for adding German values in the `title_en` in the fixture came late in the resolution of the issue. Cleaning that up would have been too much effort, sorry about that.

For https://4teamwork.atlassian.net/browse/CA-1040

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [ ] api-change label added
- New translations


opengever/bundle/tests/test_section_constructor.py
```diff
@@ -82,6 +82,25 @@ def test_creates_itranslated_title_content(self):
...
+            u"title_fr": u"Reporoot",
+            u"title_en": u"Reporoot",
+        }
+        section = self.setup_section(previous=[item])
+        list(section)
+
+        portal = api.portal.get()
+        content = portal.get('reporoot')
+        self.assertIsNotNone(content)
+        self.assertEqual(u'Reporoot', content.title_de)
```
lukasgraf, line 101: Using distinct values here (instead of `Reporoot` for all three) would help assert that they end up in the right attribute :wink:

opengever/repository/locales/en/LC_MESSAGES/opengever.repository.po
```diff
@@ -245,6 +245,11 @@ msgstr "Repositoryfolder description"
 msgid "label_repositoryfolder_title_de"
 msgstr "Repositoryfolder title (German)"
 
+#. Default: "Repositoryfolder title (English)"
+#: ./opengever/repository/browser/excel_export.py
+msgid "label_repositoryfolder_title_en"
+msgstr "Repositoryfolder title"
```
lukasgraf, line 251: This should be `Repositoryfolder title (English)`, no? Or did we translate these in a way where the active language is never suffixed?

docs/public/dev-manual/api/api_changelog.rst
```diff
@@ -16,6 +16,7 @@ Breaking Changes
 Other Changes
 ^^^^^^^^^^^^^
 
+- Add support for english: new field ``title_en`` is returned wherever appropriate (``@schema``, ``@types`` and simple GET for diverse content types) when english is enbaled for the deployment.
```
lukasgraf, line 19: enbaled -> enabled, english -> English

lukasgraf (changes requested):
Nice, LGTM! 👍 

There's only one issue of note I can see - the `navigation.json` view fails entirely with a JSON encoding error if just one of the brains contains `Missing.Value`. More on that in a separate comment below.

lukasgraf:
I was able to construct an issue that should be addressed I think:
When upgrading an existing site, enabling English as a supported language and then switching to it, the `navigation.json` view will fail to render with

```
Traceback (innermost last):
  Module opengever.base.browser.navigation, line 57, in __call__
  Module opengever.base.browser.navigation, line 77, in json
  Module json, line 244, in dumps
  Module json.encoder, line 207, in encode
  Module json.encoder, line 270, in iterencode
  Module json.encoder, line 184, in default
TypeError: Missing.Value is not JSON serializable
```

This is because the `title_en` catalog metadata column is added (via upgrade step), but not initialized with any values (which is correct, I think). But the `navigation.json` view then needs to take into account that `brain.Title` (which delegates to `brain.title_<current_language>`) might return `Missing.Value`, which is not JSON serializable.

Something along these lines:
(`opengever/base/browser/navigation.py:93`)

```python
    def _brain_to_node(self, brain):
        translated_title = brain.Title
        if translated_title is Missing.Value:
            translated_title = None

        return {'text': translated_title,
                'description': brain.Description,
                'url': brain.getURL(),
                'uid': brain.UID,
                'active': brain.review_state != REPOSITORY_FOLDER_STATE_INACTIVE, }
```

opengever/repository/locales/en/LC_MESSAGES/opengever.repository.po
njohner: Exactly active language is not suffixed.

docs/public/dev-manual/api/api_changelog.rst
njohner: 👍 😓

njohner:
Thanks for the thorough testing! I've updated the PR according to your input.

lukasgraf (approved):
🎉 🚢

njohner:
ouch forgot to merge my commits that need merging. Will do 😅

PR #6862, opened 2021-01-22
Allow workspace members to delete meetings

elioschmutz:
Meetings should be deletable by members.


opengever/core/lawgiver.zcml
```diff
@@ -383,8 +383,11 @@
         />
 
     <lawgiver:map_permissions
-        action_group="delete todos"
-        permissions="opengever.workspace: Delete Todos"
+        action_group="delete workspacecontents"
+        permissions="
+                     Delete objects,
```
buchi, line 388: This makes all workspace items deletable, but for documents we only want to allow trashing.

elioschmutz: As discussed, that's still the case.

PR #6867, opened 2021-01-26
Return related_documents in journal endpoint.

tinagerber:
When creating manual journal entries, document references can be set. But until now, these were not returned with a GET request to the @journal endpoint. 

Jira: https://4teamwork.atlassian.net/browse/NE-310

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:


opengever/api/journal.py
```diff
@@ -112,6 +114,16 @@ def reply(self):
 
         return result
 
+    def get_related_documents(self, documents):
+        related_documents = []
+        for document in documents:
+            obj = Oguid.parse(document['id']).resolve_object()
```
njohner, line 120: I guess title and resolve link are not enough?

opengever/api/journal.py
```diff
@@ -112,6 +114,16 @@ def reply(self):
...
+            serialized_document = queryMultiAdapter((obj, self.request), ISerializeToJsonSummary)()
+            serialized_document['checked_out'] = obj.checked_out_by()
+            serialized_document['file_extension'] = obj.get_file_extension()
```
njohner, line 123: I'm not a big fan of adding more information into that serialization here. Either the summary serialization is enough for external applications to show a summary of an object or we need to adapt the summary serialization. But customizing the summary serialization with additional data is not the way to go IMO. That also means that you do not have a standardised way of representing documents in the frontend...

njohner (commented):
Basically yes, but see my comments...

opengever/api/journal.py
tinagerber: We want to display an icon for the document in the frontend, as we do (almost) everywhere. For this we need the information @type, file_extension and checked_out. We also need a title and a link to the document in some form. So, what do you recommend?

njohner: Ok I feel like the frontend almost always wants that information when displaying a document. So I would overwrite the summary serialiser. Otherwise we end up adding those fields everywhere in the API.

njohner: If you see that it's a pain, leave it as is and make a story. I'm not sure whether the summary serializer is being used elsewhere...

tinagerber:
@njohner I've added the two fields to the summary serializer. Can you rereview?

docs/public/dev-manual/api/api_changelog.rst
```diff
@@ -17,6 +17,7 @@ Breaking Changes
 Other Changes
 ^^^^^^^^^^^^^
 
+- The fields ``checked_out`` and ``file_extension`` are now included in the summary serializatuin of documents and mails.
```
njohner, line 21: serializatuin -> serialization

njohner (approved):
Thanks, LGTM

PR #6868, opened 2021-01-26
Return only badge notifications in @notifications endpoint.

tinagerber:
In the new frontend, all notifications returned by the @notifications endpoint are displayed as badge notifications. This meant that notifications that were not badge notifications were also displayed.
With the changes in this PR, the @notifications endpoint only returns badge notifications.

It would also have been possible to introduce an is_badge field and filter by it in the frontend, but I think the chosen method is better, as it gives the backend control over which notifications are displayed in the frontend.

Jira: https://4teamwork.atlassian.net/browse/NE-378

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [x] api-change label added

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2021.3.0 (unreleased)
 ---------------------
 
+- Return only badge notifications in @notifications endpoint. [tinagerber]
 - Enable Usersnap by default in SaaS policy template. [lgraf]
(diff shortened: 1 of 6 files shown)


njohner (approved):
Yes I tend to agree that there is no reason to expose over the API notifications that have already been sent by E-mail to a user...

LGTM

PR #6869, opened 2021-01-26
Only show create_proposal action on dossiers.

tinagerber:
Until now, the `create_proposal` action has appeared on various contents. But the action should only appear on dossiers.

Jira: https://4teamwork.atlassian.net/browse/NE-313

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/repository/tests/test_repositoryfolder_tabs.py
+++ b/opengever/repository/tests/test_repositoryfolder_tabs.py
@@ -89,12 +89,12 @@ class TestRepositoryFolderDocumentsTab(SolrIntegrationTestCase):
         self.assertEquals([], tabbedview.major_actions().text)
 
     @browsing
-    def test_create_proposal_visible_when_meeting_feature_enabled(self, browser):
+    def test_create_proposal_action_not_visible(self, browser):
         self.activate_feature('meeting')
         self.login(self.manager, browser)
         browser.open(self.branch_repofolder)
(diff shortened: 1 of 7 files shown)


njohner (commented):
Basically ok, but
- tests are failing
- How was that solved in the classic UI until now? It should be possible to use the same conditions for both, otherwise we still run the risk that the action does not appear at the correct locations in the new UI

tinagerber:
This is the same action that is used in the classic UI. Without these changes, the action in the classic UI is also available at RepositoryFolder level (in the documents tab). If you try to create a proposal there, an exception is thrown.

opengever/repository/tests/test_repositoryfolder_tabs.py
```diff
@@ -88,14 +88,6 @@ def test_visible_actions(self, browser):
 
         self.assertEquals([], tabbedview.major_actions().text)
 
-    @browsing
-    def test_create_proposal_visible_when_meeting_feature_enabled(self, browser):
```
njohner, line 92: At least it was documented that it was wrong 🥇 

Instead of removing that test, can you make it a regression test, i.e. check that the button does not appear?

njohner:
Ok I had not realised it also changed the classic-UI.

njohner (approved):
LGTM

njohner (approved):
LGTM

PR #6872, opened 2021-01-28
Add a new solr-index 'is_folderish'.

elioschmutz:
This PR adds the plone `is_folderish` index to the solr.

Jira: https://4teamwork.atlassian.net/browse/NE-281

We reuse the already existing indexer from plone which checks the `isPrincipiaFolderish`, as well as the `INonStructuralFolder` to decide if a content is folderish or not.

See https://github.com/plone/Products.CMFPlone/blob/4.3.18/Products/CMFPlone/CatalogTool.py#L235 for more information about the plone implementation.

⌛ Estimated Time to reindex:
- 80'000: 2 min
- 1'000'000: 25 min

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2021.3.0 (unreleased)
 ---------------------
 
+- Add a new solr-index 'is_folderish'. [elioschmutz]
 - Return only badge notifications in @notifications endpoint. [tinagerber]
 - Only show create_proposal action on dossiers. [tinagerber]
 - Enable Usersnap by default in SaaS policy template. [lgraf]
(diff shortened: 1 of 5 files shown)


njohner (commented):
LGTM, but I'm a bit unsure about the upgrade step. I know we've had a few upgrades indexing stuff in Solr that did not work properly on larger deployments and ended up not indexing anything. Can't really remember why either. @buchi is this one OK?

njohner (approved):

njohner:
Ok never mind, it seems this comes from an upgrade step from @buchi and was discussed with him and @lukasgraf .

PR #6874, opened 2021-01-28
Add blocked_local_roles field in serialization of dossiers and repofolders, index field in solr

tinagerber:
Changes in this PR:
- Index the field `blocked_local_roles` for dossiers and repofolders in Solr. Indexers for the types already exist, as the index is also in the catalog
- Allow `blocked_local_roles` field in `@listing` endpoint
- Add `blocked_local_roles` to dossier and repofolder serialization
- Allow to filter for a boolean term with `filters.<name>:record:boolean`, until now, such a query threw an exception

Jira: https://4teamwork.atlassian.net/browse/NE-275

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:

### DIFF
--- a/opengever/api/tests/test_dossier.py
+++ b/opengever/api/tests/test_dossier.py
@@ -23,6 +23,12 @@ class TestDossierSerializer(IntegrationTestCase):
         )
 
     @browsing
+    def test_dossier_serializer_contains_blocked_local_roles(self, browser):
+        self.login(self.regular_user, browser=browser)
+        browser.open(self.dossier, method="GET", headers=self.api_headers)
+        self.assertIn("blocked_local_roles", browser.json)
+
+    @browsing
     def test_undeterminable_subdossier_within_items(self, browser):
         self.login(self.regular_user, browser=browser)
         browser.open(self.dossier, method="GET", headers=self.api_headers)
--- a/opengever/api/tests/test_serializer.py
+++ b/opengever/api/tests/test_serializer.py
@@ -64,6 +64,13 @@ class TestRepositoryFolderSerializer(IntegrationTestCase):
         self.assertEqual(u'Client1 1.1', browser.json.get(u'reference_number'))
 
     @browsing
+    def test_repofolder_serialization_contains_blocked_local_roles(self, browser):
+        self.login(self.regular_user, browser)
+        browser.open(self.leaf_repofolder, headers=self.api_headers)
+        self.assertEqual(200, browser.status_code)
+        self.assertIn('blocked_local_roles', browser.json)
+
+    @browsing
     def test_repofolder_serialization_contains_is_leafnode(self, browser):
         self.login(self.regular_user, browser)
         browser.open(self.leaf_repofolder, headers=self.api_headers)
--- a/opengever/core/upgrades/20210128083955_index_blocked_local_roles_in_solr_for_repository_folders_and_dossiers/upgrade.py
+++ b/opengever/core/upgrades/20210128083955_index_blocked_local_roles_in_solr_for_repository_folders_and_dossiers/upgrade.py
@@ -1,0 +1,19 @@
+from ftw.upgrade import UpgradeStep
+from opengever.dossier.behaviors.dossier import IDossierMarker
+from opengever.repository.interfaces import IRepositoryFolder
+
+
+class IndexBlockedLocalRolesInSolrForRepositoryFoldersAndDossiers(UpgradeStep):
+    """Index blocked_local_roles in solr for repository folders and dossiers.
+    """
+
+    deferrable = True
+
+    def __call__(self):
+        query = {'object_provides': [
+            IDossierMarker.__identifier__,
+            IRepositoryFolder.__identifier__,
+        ]}
+
+        for obj in self.objects(query, 'Index blocked_local_roles field in solr.'):
+            obj.reindexObject(idxs=['blocked_local_roles'])
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 ---------------------
 
 - Add a new solr-index 'is_folderish'. [elioschmutz]
+- Include blocked_local_roles in serialization of dossiers and repofolders. [tinagerber]
 - Index blocked_local_roles in solr and allow field in @listing endpoint. [tinagerber]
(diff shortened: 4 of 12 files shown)


opengever/api/tests/test_dossier.py
```diff
@@ -22,6 +22,12 @@ def test_dossiers_within_items_are_subdossiers(self, browser):
             'All dossiers within "self.dossier" must be subdossiers.'
         )
 
+    @browsing
+    def test_dossier_serializer_contains_blocked_local_roles(self, browser):
+        self.login(self.regular_user, browser=browser)
+        browser.open(self.dossier, method="GET", headers=self.api_headers)
+        self.assertFalse(browser.json["blocked_local_roles"])
```
njohner, line 29: self.assertIn("blocked_local_roles", browser.json)

opengever/api/tests/test_serializer.py
```diff
@@ -63,6 +63,13 @@ def test_repofolder_serialization_contains_reference_number(self, browser):
         self.assertEqual(200, browser.status_code)
         self.assertEqual(u'Client1 1.1', browser.json.get(u'reference_number'))
 
+    @browsing
+    def test_repofolder_serialization_contains_blocked_local_roles(self, browser):
+        self.login(self.regular_user, browser)
+        browser.open(self.leaf_repofolder, headers=self.api_headers)
+        self.assertEqual(200, browser.status_code)
+        self.assertFalse(browser.json.get(u'blocked_local_roles'))
```
njohner, line 71: That will most likely not fail if 'blocked_local_roles' is not in browser.json

opengever/core/upgrades/20210128083955_index_blocked_local_roles_in_solr_for_repository_folders_and_dossiers/upgrade.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+
+    deferrable = True
+
+    def __call__(self):
+        query = {'object_provides': [
+            IDossierMarker.__identifier__,
+            IRepositoryFolder.__identifier__,
+        ]}
+
```
njohner, line 17: This is just a thought, but because it's a boolean index that is in the catalog, we might avoid waking up all the objects:
```
    def __call__(self):
        manager = getUtility(ISolrConnectionManager)
        query = {
            'object_provides': [
                IDossierMarker.__identifier__,
                IRepositoryFolder.__identifier__],
            'blocked_local_roles': False
            }
        
        contents = self.catalog_unrestricted_search(query)
        for content in ProgressLogger('Index blocked_local_roles=False in Solr', contents):
            contents.blocked_local_roles = False
            handler = getMultiAdapter((content, manager), ISolrIndexHandler)
            handler.add(['blocked_local_roles'])

        query['blocked_local_roles'] = True
        contents = self.catalog_unrestricted_search(query)
        for content in ProgressLogger('Index blocked_local_roles=True in Solr', contents):
            contents.blocked_local_roles = True
            handler = getMultiAdapter((content, manager), ISolrIndexHandler)
            handler.add(['blocked_local_roles'])

        manager.connection.commit(soft_commit=False, extract_after_commit=False)
```
I haven't tried, but could be worth a shot. Would make it way faster...

njohner (approved):
Just one proposal for the upgrade. Otherwise LGTM

opengever/core/upgrades/20210128083955_index_blocked_local_roles_in_solr_for_repository_folders_and_dossiers/upgrade.py
njohner: @lukasgraf what do you think?

lukasgraf: @njohner hmm, your suggested code doesn't look quite right to me 😉 
The two assignments `contents.blocked_local_roles = ` are meant to assign to `content` instead of `contents`, I assume. But even then, that would mean you'd assign to a `Brain`, which is either immutable (not sure), but at the very least transient in nature.

With `getMultiAdapter((content, manager), ISolrIndexHandler)` the _brain_ then be used for adaption to look up the `ISolrIndexHandler`. I have my doubts on whether that's going to work as intended.

njohner: Yeah I know it's transient on the brain, that is intended. I'm not quite sure it would work, but the `ISolrIndexHandler` then tries to look for the attribute on the `context`, i.e. on the brain (Ok, again, it might first try to wrap it https://github.com/4teamwork/ftw.solr/blob/master/ftw/solr/handlers.py#L99-L105).
I'm indeed not sure it would work, and it's a bit of a hack. On the other hand, it would be much faster than getting all the dossiers...

lukasgraf: You're right, using the brain for adaptation apparently does work indeed. And now I'm getting your idea - because `blocked_local_roles` is a catalog index, but not a metadata column, you'd need to slap it onto the brain for it to be picked up. I wasn't quite getting that part before. I think it's worth a try - but should be tested well.

njohner: yeah sorry, my explanation was not very precise. But that was the idea yes.

tinagerber: This does not work, an AttributeError is thrown during `content.blocked_local_roles = False`

tinagerber: @njohner @lukasgraf Do you have any other suggestions? Otherwise I'll merge.

njohner: ok too bad. Was worth a try!
No other suggestion from my side...

PR #6875, opened 2021-01-28
Insert a full stop at the end of a sentence for some translations.

tinagerber:
Since several translations have already been adapted in this sprint, I have dispensed with a CHANGELOG entry.

Jira: https://4teamwork.atlassian.net/browse/NE-512

## Checklist (Must have)

- [ ] Changelog entry

### DIFF
--- a/opengever/api/transition.py
+++ b/opengever/api/transition.py
@@ -184,7 +184,7 @@ class GEVERDossierWorkflowTransition(GEVERWorkflowTransition):
             # another violated precondition.
             self.request.response.setStatus(400)
             msg = self.translate(str(e))
-            errors = ['The dossier %s has a invalid end_date' % title
+            errors = ['The dossier %s has an invalid end date.' % title
                       for title in e.invalid_dossier_titles]
             return dict(error=dict(
                 type='PreconditionsViolated',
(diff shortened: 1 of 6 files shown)


njohner (commented):
I guess those affect the API. This means they are breaking changes...

tinagerber:
So, do you think a change to an error message should be listed as a breaking change in the API changelog?

lukasgraf:
@tinagerber yes, because are currently the only way a client could recognize why one of those workflow transitions has failed. That's why I marked them with `MARKER: used-in-api-response` in the `.po` files, and didn't touch them during my English translations.

In my opinion it simply wasn't worth the risk of breaking external applications. I would holf off on this issue for now, and double check with PO first whether we really want to take that risk for some typos. And if we do change it, it definitely needs a breaking API changelog entry, unfortunately.

lukasgraf:
Or in other words: If we do change these API responses, we should change them _once_ in a way where we output a machine readable identifier that never changes (like `no_all_documents_checked_in`) in addition to human readable error messages, so that the translated messages become irrelevant.

phgross:
I just talked to our two biggest "API clients" (RIS, UBit-Laufwerk), they're not using the transitions endpoint at all or at least they're not making a specific error handling. 

IMHO i don't think its a big risk to change the messages. A corresponding error handling would have been based on a translated error message and would have been implemented in a very "shaky" way. So my proposal would be, to add a API Changelog Entry to the actual changes. 

@lukasgraf @njohner what you think?

njohner:
I'd agree that if we change the messages, we might as well go for something machine readable.

lukasgraf:
Exactly, I agree with @njohner.
👍 for a breaking API change (thanks @phgross for checking with stakeholders), but I would use this chance to return, for the errors, both a machine readable identifier (e.g. `error_invalid_end_date`) as well as the (fixed) human readable message.

elioschmutz (approved):

elioschmutz:
Sorry, ignore my approvement! Was not the right PR 🙈

PR #6877, opened 2021-01-29
Update to Plone 4.3.20

buchi:
Let's update at least to the latest and last ever release of Plone 4.3 which btw. was first released in April 2013. This gets us some security fixes and various other bug fixes. Even some new functionality is included, e.g. events after group creation or modification.

I've checked the change logs carefully and didn't find anything that sounds like a breaking change. There is one change that leads to a slightly different behavior: Indexing of objects acquired attributes from its parents, if the attribute was missing on the object itself. This means that e.g. a document was indexed with the `start` date of its parent dossier. With the latest release, acquired attributes are no longer indexed. Some test assertions had to be fixed because of this.

**Most notable changes for 4.3.20:**

- Integrated PloneHotfix20200121 for increased security.
- Moved the security check if a url is in the portal to a small separate package: Products.isurlinportal. You can immediately use this on Plone 4.3 and higher. Keep an eye on updates for this package: newer versions will increase the security. Often the impact of fixes is too small to warrant a real security hotfix package, but we want to do more regular fixes here.
- Use Products.isurlinportal 1.1.0 with security hardening against whitespace: https://github.com/plone/Products.isurlinportal/issues/1 1
- Removed broken X-XSS-Protection header from classic theme and unstyled theme.
- Products.PluggableAuthService: Added new events to be able to notify when a principal is added to or removed from a group. Notify these events when principals are added or removed to a group in ZODBGroupManager. See https://github.com/zopefoundation/Products.PluggableAuthService/issues/17
- z3c.autoinclude: When environment variable Z3C_AUTOINCLUDE_DEBUG is set, log which packages are being automatically included.

Full changelog is available here: https://community.plone.org/t/plone-4-3-20-soft-released/12750

**Most notable changes for 4.3.19:**

- Add X-Content-Type-Options and X-XSS-Protection security headers.
- Add new event to be able to notify group creation.
- Fixes slow lookup of documentToKeyMap in GopipIndex. About 66x speedup. This may add up to seconds less on large navtree renderings.

Full changelog is available here: https://community.plone.org/t/plone-4-3-19-released/10219

JIRA: https://4teamwork.atlassian.net/browse/ROAD-1549

## Checklist (Must have)

- [x] Changelog entry


lukasgraf (approved):
LGTM 👍 
I would however suggest to wait until after the upcoming release with merging this one, so that we can merge it at the beginning of a sprint instead of at the end. So it spends more time on DEV to catch any subtle side effects.

PR #6879, opened 2021-02-01
Fix wrapping of keywords in keywordwidget.

njohner:
Long keywords were not being wrapped, so that the "x" to remove them was not visible. We now wrap them, see screenshot below:

Screenshot 2021-02-01 at 10 58 40

## Checklist (Must have)

- [x] Changelog entry


lukasgraf (approved):
LGTM 👍

PR #6889, opened 2021-02-05
Docugate integration

buchi:
Documents can be created from Docugate templates using a dedicated add form. The add form has the same fields as the regular document add form, except for the file field. Documents are created in shadow state.

There's also a REST API endpoint `@document_from_docugate` for creating documents from Docugate with similar functionality to the add form used by the new frontend.

After creating the shadow document, Office Connector gets called with an `oc:...` URL. It then fetches the needed metadata from the REST API endpoint `oc_docugate`.

Then OC fetches the XML document required by Docugate containing all Doc Properties from the `docugate-xml` view.

It then calls Docugate with the XML document. After getting the new document from Docugate, OC opens the document in the Office desktop app and enters the regular edit cycle.

When editing is finished, OC uploads the document.

Documents get a new action "Retry with Docugate" to be able to continue/restart the Office Connector part if it wasn't finished for some reason.

The OC part is not yet implemented.
Frontend part: https://github.com/4teamwork/gever-ui/pull/1595

JIRA: https://4teamwork.atlassian.net/browse/CA-887

## Checklist (Must have)

- [x] Changelog entry


njohner (approved):
Nice, thanks for the small and clear commits. LGTM.

PR #6890, opened 2021-02-06
Bump cryptography to version 2.8

buchi:
The currently used version can now longer be compiled on newer macOS versions and is pretty old.
2.8 is the last version that support OpenSSL 1.0.1 which we still have on some CentOS 6 installations and can be compiled on macOS 11 as well.

## Checklist (Must have)

- [ ] Changelog entry


lukasgraf (approved):
LGTM 👍 
Though I believe we should also bump `asn1crypto` (to at least 1.0.0). The issue described here https://github.com/wbond/asn1crypto/issues/158 has been encountered by several people on Homebrew OpenSSL + Catalina setups. But that can be looked at separately IMHO.

buchi:
@lukasgraf bumped `asn1crypto` to latest version.

buchi:
@lukasgraf I've noticed that cryptography no longer depends on asn1crypto.
https://github.com/pyca/cryptography/pull/4941

Thus I'm removing the pinning as we don't use it anywhere else.

lukasgraf:
@buchi great! Thanks for checking 👍

PR #6891, opened 2021-02-08
Set SameSite=Lax flag for session authentication cookie

buchi:
This is best practice in terms of security.

With `SameSite=Lax` the cookie is only sent if the user is navigating to the site (following a link or submitting a form with GET method). The cookie is not sent when loading images, AJAX requests, Iframes and forms with POST method.

See also:
https://4teamwork.atlassian.net/wiki/spaces/4TEAM/pages/501907630/Sicherheitsrichtlinien+f+r+Webanwendungen#4.-Restriktive-Cookies

## Checklist (Must have)

- [x] Changelog entry


lukasgraf (approved):
LGTM 👍

PR #6892, opened 2021-02-08
Remove LDAPUserFolder encoding patch

buchi:
Our custom version 2.28.post3 already has the encoding set to utf-8.

## Checklist (Must have)

- [ ] Changelog entry


lukasgraf (approved):
LGTM 👍

PR #6893, opened 2021-02-08
New API endpoint @save-document-as-pdf

tinagerber:
Changes in this PR:

- Only allow to save a document as pdf if document isn't checked out.  Until now, checked-out documents could also be saved as PDFs.This resulted in the fact that when saving a document, the last version of the document was saved as pdf, but when checking whether the document could be converted, the working copy was used. This could lead to errors.
- New endpoint `@save-document-as-pdf`. In order to be able to offer this function in the new frontend, a new endpoint is needed. For this, some functionalities had to be moved to the global scope in order to reuse them in the endpoint.

Jira: https://4teamwork.atlassian.net/browse/NE-352

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - New functionality:

### DIFF
--- a/docs/public/dev-manual/api/documents.rst
+++ b/docs/public/dev-manual/api/documents.rst
@@ -319,3 +319,25 @@ Versionen auflisten:
             "version": 0
         }
     ]
+
+.. _save-document-as-pdf:
+
+Dokument als PDF speichern
+--------------------------
+
+Mit dem ``@save-document-as-pdf`` kann ein Dokument oder eine Version eines Dokuments als PDF gespeichert werden. Als ``document_uid`` wird die UID des Dokuments erwartet. Wenn keine ``version_id`` mitgegeben wird, wird die aktuelle Version verwendet. Der Endpoint steht auf Stufe Dossier, Teamraum und Teamraum-Ordner zur Verfügung.
+
+  .. sourcecode:: http
+
+    POST /ordnungssystem/dossier-23/@save-document-as-pdf HTTP/1.1
+    Accept: application/json
+
+    {
+      "document_uid": "f923b2321f174b408c3bd483db9bfa66",
+      "version_id": 2
+    }
+
+  .. sourcecode:: http
+
+    HTTP/1.1 201 Created
+    Location: /ordnungssystem/dossier-4/document-1
--- a/opengever/api/save_document_as_pdf.py
+++ b/opengever/api/save_document_as_pdf.py
@@ -1,0 +1,49 @@
+from opengever.document.behaviors import IBaseDocument
+from opengever.document.browser.save_pdf_document_under import trigger_conversion
+from opengever.document.forms import create_destination_document
+from opengever.document.forms import is_version_convertable
+from opengever.document.versioner import Versioner
+from plone import api
+from plone.protect.interfaces import IDisableCSRFProtection
+from plone.restapi.deserializer import json_body
+from plone.restapi.services import Service
+from zExceptions import BadRequest
+from zope.interface import alsoProvides
+
+
+class SaveDocumentAsPdfPost(Service):
+
+    def get_document(self, data):
+        uid = data.get('document_uid', None)
+        if not uid:
+            return
+        catalog = api.portal.get_tool('portal_catalog')
+        brain = catalog(UID=uid, object_provides=[IBaseDocument.__identifier__])
+        if brain:
+            return brain[0].getObject()
+
+    def reply(self):
+        # Disable CSRF protection
+        alsoProvides(self.request, IDisableCSRFProtection)
+
+        data = json_body(self.request)
+        document = self.get_document(data)
+        if not document:
+            raise BadRequest(
+                "Property 'document_uid' is required and should be a UID of a document")
+
+        version_id = data.get('version_id', None)
+        if version_id and not isinstance(version_id, int):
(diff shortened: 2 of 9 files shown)


docs/public/dev-manual/api/documents.rst
```diff
@@ -319,3 +319,25 @@ Versionen auflisten:
...
+
+Mit dem ``@save-as-pdf`` kann ein Dokument oder eine Version eines Dokuments als PDF in ein Dossier gespeichert werden. Als ``destination_folder`` wird die UID des Zieldossiers erwartet. Wenn keine ``version_id`` mitgegeben wird, wird die aktuelle Version verwendet.
+
+  .. sourcecode:: http
+
+    POST /ordnungssystem/dossier-23/document-123/@save-as-pdf HTTP/1.1
+    Accept: application/json
+
+    {
+      "destination_folder": "f923b2321f174b408c3bd483db9bfa66",
```
njohner, line 336: I'd use something more generic than `destination_folder`. This action is close to a copy IMO, I would therefore go with `target` instead. 

And actually both `copy` and `move` POST to the destination folder with `source` as parameter instead, this does make sense because we are creating a content in the `target` so it should be a post to that object. So I'd propose to also do it like this here. Call the endpoint on the destination folder and add `source` as parameter.
Maybe @lukasgraf has some thoughts here?

opengever/api/tests/test_save_as_pdf.py
```diff
@@ -0,0 +1,171 @@
...
+    @browsing
+    def test_save_as_pdf_post_without_add_permission_raises_unauthorized(self, browser):
+        self.login(self.regular_user, browser)
+
+        with elevated_privileges():
+            self.empty_dossier.__ac_local_roles_block__ = True
+            self.empty_dossier.manage_setLocalRoles(self.regular_user.id,
+                                                    roles=['Authenticated', 'Reader'],
+                                                    verified=True)
+        with browser.expect_unauthorized():
```
njohner, line 109: I guess unauthorized is raised at some point when the document is being created?
If we were doing a `POST` on the destination folder we could handle that cleanly directly in the Endpoint registration

njohner (commented):
See my comments. I feel like this should be a POST to the target container and not onto the document

tinagerber:
@njohner I have changed the endpoint so that the POST is on the target container. However, I decided to call the property `document_uid`, so it is immediately clear what is required.

opengever/api/save_document_as_pdf.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class SaveDocumentAsPdfPost(Service):
+
+    def get_document(self, data):
+        uid = data.get('document_uid', None)
+        if not uid:
+            return
+        catalog = api.portal.get_tool('portal_catalog')
+        brain = catalog(UID=uid, object_provides=[IDocumentSchema.__identifier__])
```
njohner, line 21: You want `IBaseDocument` here, because the endpoint should also work for E-mails

opengever/api/tests/test_save_document_as_pdf.py
```diff
@@ -0,0 +1,182 @@
...
+          "type": "BadRequest"}, browser.json)
+
+    @browsing
+    def test_save_as_pdf_post_with_invalid_document_uid_raises_bad_request(self, browser):
+        self.login(self.regular_user, browser=browser)
+
+        with browser.expect_http_error(400):
+            browser.open(self.empty_dossier, view='@save-document-as-pdf', method='POST',
+                         headers=self.api_headers,
+                         data=json.dumps({"document_uid": self.mail_eml.UID()}))
```
njohner, line 67: As I said, this should work for E-mails, see https://github.com/4teamwork/opengever.core/blob/master/opengever/document/browser/configure.zcml#L55-L60

opengever/api/tests/test_save_document_as_pdf.py
```diff
@@ -0,0 +1,182 @@
...
+
+        manager = getMultiAdapter((self.document, self.request), ICheckinCheckoutManager)
+        manager.checkout()
+
+        with browser.expect_http_error(400):
+            browser.open(self.empty_dossier, view='@save-document-as-pdf', method='POST',
+                         headers=self.api_headers,
+                         data=json.dumps({"document_uid": self.document.UID()}))
+
+        self.assertEqual({"message": "This document is not convertable.",
```
njohner, line 151: That's pretty misleading as error message in that case no?

njohner (commented):
See my comments, I think the endpoint has to work for mails. This is normally in the checklist and should not be deleted when something is done for documents.

opengever/api/tests/test_save_document_as_pdf.py
tinagerber: I don't see a problem here. The document is not convertible when it is checked out. The action does not appear at all, so it should not happen that the endpoint is called for a checked-out document.

njohner: Well I guess that if I get that error message I would think the document is not convertable, but it's actually just in checkout. I mean I could get the latest version and convert that. This suggests that it is generally not convertable, maybe because of the file type or something.
But ok, we can't handle every error path specifically.

njohner (approved):
LGTM

PR #6894, opened 2021-02-08
Set security related HTTP headers

buchi:
Setting these headers is best practice in terms of security.

### No content type sniffing
Only load scripts and stylesheets with the correct content type:
```
X-Content-Type-Options: nosniff
```

### XSS browser protection
Enable browser protection for reflected XSS attacks:
```
X-XSS-Protection: 1; mode=block
```

### No referrer for external links
Only send referrer on requests to the same origin
```
Referrer-Policy: same-origin
```

### Content-Security-Policy
For HTML resources we only allow resources from the same origin, but we have to allow eval and inline js/css. Images are allowed from the same origin and from data urls. We disallow plugins and framing.
```
Content-Security-Policy: default-src 'self' 'unsafe-eval' 'unsafe-inline'; img-src 'self' data:; object-src 'none'; frame-ancestors 'none'
```

For other resources we disallow the loading of any resources and disable framing.
```
Content-Security-Policy: default-src 'none'; frame-ancestors 'none'
```

### Disallow framing (Clickjacking protection)
`X-Frame-Options` header is already set by plone.protect

See also: https://4teamwork.atlassian.net/wiki/spaces/4TEAM/pages/501907630/Sicherheitsrichtlinien+f+r+Webanwendungen

The content security policy has the potential to break things and needs careful testing.

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/base/subscribers.py
+++ b/opengever/base/subscribers.py
@@ -142,21 +142,23 @@ def set_security_headers(event):
         # Enable browser protection for reflected XSS attacks
         response.setHeader('X-XSS-Protection', '1; mode=block')
 
-        content_type = response.headers.get('content-type', '')
-        if content_type.startswith('text/html'):
-            # Allow resources from current origin, eval, inline js and css
-            # Allow images from current origin and data urls
-            # Disallow plugins and framing
-            response.setHeader(
-                'Content-Security-Policy',
-                "default-src 'self' 'unsafe-eval' 'unsafe-inline'; "
-                "img-src 'self' data:; "
-                "object-src 'none'; frame-ancestors 'none'")
-        else:
-            # Disallow the loading of any resources and disable framing
-            response.setHeader(
-                'Content-Security-Policy',
-                "default-src 'none'; frame-ancestors 'none'")
+        # Keep existing CSP header to allow browser views to set a custom CSP
+        if response.getHeader('Content-Security-Policy') is None:
+            content_type = response.headers.get('content-type', '')
+            if content_type.startswith('text/html'):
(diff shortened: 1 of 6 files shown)


opengever/base/subscribers.py
```diff
@@ -130,3 +130,33 @@ def scrub_server_version(event):
...
+
+        # Enable browser protection for reflected XSS attacks
+        response.setHeader('X-XSS-Protection', '1; mode=block')
+
+        content_type = response.headers.get('content-type', '')
+        if content_type.startswith('text/html'):
+            # Allow resources from current origin, eval, inline js and css
+            # Allow images from current origin and data urls
+            # Disallow plugins and framing
+            response.setHeader(
```
lukasgraf, line 150: I believe this has the potential to break HubSpot (possibly UserSnap as well), since it loads images and scripts from its own domains.

lukasgraf (approved):
LGTM 👍 

As discussed, probably a good idea to test this first on a few select sites before merging to master. Or actually use `Content-Security-Policy-Report-Only` plus a view t receive the reports if it turns out to become a pandora's box.

opengever/base/subscribers.py
buchi: HubSpot and Usersnap are included by the new frontend only and not affected by these headers (headers for the frontend are set by Apache/Nginx).

But Office Online was broken because of this. I've added the possibility that browser views can set their own CSP header and included appropriate an appropriate CSP for the WOPI edit page.

lukasgraf (approved):
LGTM 👍

PR #6895, opened 2021-02-08
Add two property sheets to fixture.

deiferni:
This helps E2E tests as they will contain test-cases where custom properties are required.

Jira: https://4teamwork.atlassian.net/browse/NE-367

## Checklist (Must have)

- [ ] ~~Changelog entry~~

### DIFF
--- a/opengever/document/tests/test_document.py
+++ b/opengever/document/tests/test_document.py
@@ -109,7 +109,10 @@ class TestDocument(IntegrationTestCase):
         self.login(self.regular_user, browser)
         browser.open(self.document, view='@@view')
         self.assertEqual(u'Vertr\xe4gsentwurf', browser.css('.documentFirstHeading').first.text)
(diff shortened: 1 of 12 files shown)


njohner (approved):
LGTM

all4manu (commented):
Seems OK as far as I can tell. The only thing I'm not sure about is the required boolean field. This means the backend rejects `null` as a value. Currently the user would have to click twice on the checkbox in order to send `false`.

If we can expect required boolean fields, then the UI has to handle this case and either always send false or at least if it is a required field.

deiferni:
@all4manu yeah i was unintentionally extra mean and made them all required. It is theoretically possible to have them a required though, with what zope/plone offers. I would recommend to handle this in the UI, so the backend won't raise any errors.
Another option would be to always send a value, i.e. always send `false` when the checkbox is disabled, regardless of the required.

PR #6896, opened 2021-02-10
Fix custom property choice field (de-)serialization.

deiferni:
This PR fixes some issues that surfaced due to `Choice` fields not being initialized correctly and thus skipping validation.

- Choice fields have a special condition in their validate method, skipping validation during field initialization https://github.com/zopefoundation/zope.schema/blob/bdc8935158e287549e671ee14a6f17174b9c6aa7/src/zope/schema/_field.py#L486-L488
- Our version of `plone.supermodel` has a bug which never unsets the value https://github.com/plone/plone.supermodel/blob/20812f8162ac549bad6898047c53f8dc8e3d1780/plone/supermodel/exportimport.py#L164, this is only fixed in https://github.com/plone/plone.supermodel/pull/12

We also change how `Choice` field values are represented in the API. Choice fields should be serialized and deserialized consistently with vocabulary based choice fields. They should accept a nested dict with token and title, and also be serialized in that format.
_I'm tracking this as a breaking change, because in theory it is. In practice we don't use this functionality anywhere outside the UI and the change will have no impact as the UI already works against the branch to be merged with this PR._

We also fix serializing/deserializing choice values with special characters while defining the property sheet and storing/reading it from XML. We amend integration tests to use non-ascii characters.

Jira: https://4teamwork.atlassian.net/browse/NE-367

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:

### DIFF
--- a/opengever/propertysheets/definition.py
+++ b/opengever/propertysheets/definition.py
@@ -13,6 +13,7 @@ from plone.supermodel import loadString
 from plone.supermodel import model
 from plone.supermodel import serializeSchema
 from zope.component import getUtility
+from zope.schema import Choice
 from zope.schema import getFieldNamesInOrder
 from zope.schema import getFieldsInOrder
 from zope.schema import ValidationError
@@ -63,6 +64,28 @@ class PropertySheetSchemaDefinition(object):
         if assignments is None:
             assignments = tuple()
         self.assignments = assignments
+
+        for field in self.get_fields():
+            self._init_field(field[1])
+
+    def _init_field(self, field):
+        """Make sure field initialization is completed.
+
+        Choice fields are constructed by `ChoiceHandler`, a choice-field
+        specific `IFieldExportImportHandler` implementation. It does not seem
+        to construct the fields via their constructor and thus never sets
+        the `_init_field` instance variable to `False` to signal that
+        initialization is complete. This will cause the field to always skip
+        validation.
+
+        To work around this issue we set the attribute manually after the
+        schema class and its fields have been loaded or a field is added.
+
+        This is fixed with https://github.com/plone/plone.supermodel/pull/12
+        and available when we make the move to plone 5.
+        """
+        if isinstance(field, Choice):
+            field._init_field = False
 
     def __eq__(self, other):
         if isinstance(other, PropertySheetSchemaDefinition):
@@ -137,6 +160,7 @@ class PropertySheetSchemaDefinition(object):
         field = factory(**properties)
         schema = IEditableSchema(self.schema_class)
         schema.addField(field)
+        self._init_field(field)
 
     def get_fields(self):
         """Return a list of (name, field) tuples in native schema order."""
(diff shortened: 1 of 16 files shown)


opengever/propertysheets/definition.py
```diff
@@ -64,6 +65,25 @@ def __init__(self, name, schema_class, assignments=None):
             assignments = tuple()
         self.assignments = assignments
 
+        for field in self.get_fields():
+            self._init_field(field[1])
+
+    def _init_field(self, field):
+        """Make sure field initialization is completed.
+
+        Fields are constructed by `ChoiceHandler`, a choice-field specific
```
njohner, line 69: I guess only choice fields are constructed by the `ChoiceHandler`? So sentence should probably read "Choice fields are constructed by `ChoiceHandler`..."

opengever/propertysheets/serializer.py
```diff
@@ -0,0 +1,65 @@
...
+    def get_value(self, default=None):
+        value = super(PropertySheetFieldSerializer, self).get_value(
+            default=default
+        )
+        return self.serialize_custom_properties(value)
+
+    def serialize_custom_properties(self, value):
+        """Serialize nested custom property field data structure to JSON.
+
+        Handles invalid data grafecully and returns it, whenever possible.
```
njohner, line 34: I wonder whether we really want to be that permissive. This will basically prevent us from noticing if data is stored incorrectly no?
I get that we need to be somewhat permissive, so that if a property sheet is deleted for example, the object is still valid.

njohner (approved):
LGTM. My only concern is that we might be so lenient with serialization and deserialization that we will not realise when bad data is being stored because of a bug or another...

opengever/propertysheets/definition.py
deiferni: Yes indeed, thanks. Amended and also added pointer to https://github.com/plone/plone.supermodel/pull/12 which provides a fix.

opengever/propertysheets/serializer.py
deiferni: I was going back and forth about this as well. I wanted to prevent having issues with invalid data when content is displayed. I wanted to prevent a worst-case scenario when e.g. a user modifies the property sheet, we have a bug and suddenly  all documents don't render any more. That's why i'm currently permissive during serialization but strict
in deserialization/validation when setting data via field, see https://github.com/4teamwork/opengever.core/blob/6ddfb3caa0a50c46b76b3b39b4a1108447cc70ea/opengever/propertysheets/field.py#L73-L85

Also i would like to mention, deserialization may appear permissive in this PR, but there will always be validation after deserialization, which will reject invalid data, see https://github.com/plone/plone.restapi/blob/056f968d65aabbe6d0f34b24e6b8dcdb42bc6124/src/plone/restapi/deserializer/dxfields.py#L52-L54

njohner: As discussed, it makes sense in that case to be more liberal in serialization.

PR #6897, opened 2021-02-12
Fix inbox document overview for managers 

lukasgraf:
The document overview for documents that are inside an inbox got broken, because the [code for the `show_edit_archival_file_link()`](https://github.com/4teamwork/opengever.core/blob/ad3f7ac1452cc21ae266a5c08100055b149c3f83/opengever/document/browser/overview.py#L352-L361) doesn't deal with the fact that `get_parent_dossier()` might return `None`.

This case is never an issue for regular users because they lack the necessary permissions and the previous condition already terminates the method, but for managers it broke the overview for documents in inboxes.

Fixes https://sentry.4teamwork.ch/organizations/sentry/issues/71907/

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2021.4.0 (unreleased)
 ---------------------
 
+- Fix inbox document overview for managers. [lgraf]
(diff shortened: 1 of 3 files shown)


njohner (approved):
LGTM

PR #6411, opened 2020-05-08
Add "is_admin_menu_visible" to the @config API endpoint.

mbaechtold:
This new information is used by the the GEVER UI in order to determine if it should display the menu item "Verwaltung" in the navigation drawer. 

The chosen solution has been discussed with @all4manu and @deiferni prior to its implementation and was deemed as a "good-is-good-enough" solution. All other options we discussed (dedicated endpoint @navigationdrawer, enhancement of @navigation or @actions endpoints) were deemed bloated, "unnatural" or violated the YAGNI principle.

This pull requests belongs to [GEVER-297](https://4teamwork.atlassian.net/browse/GEVER-297), a sub task of the Jira issue [GEVER-36](https://4teamwork.atlassian.net/browse/GEVER-36).

## Checklist (Must have)

- [x] Changelog entry
- [x] Documentation updated (notably for API and deployment)
- [x] Link to issue (Jira or GitHub) and backlink in issue (Jira)

PR #6898, opened 2021-02-15
Rename the attribute 'is_admin_menu_visible' from the @config endpoint to 'is_admin'.

elioschmutz:
The `@config`-Endpint exposes an attribute called `is_admin_menu_visible` which was introduced for the GEVER-UI to indicate, if the management-area in the frontend should be visible or not. Since the introduction of this attribute, the frontend has changed and is now providing a management-area for every gever user, not only for admins. The frontend decides depending on attributes value, which cards should be shown in the management area. In addition, it protects special management routes with the value of this attribute.

In general, the frontend just needs to know if the current user is an admin or not. So we rename this attribute in the `@config` endpoint to make it semantically correct.

Slack-Discussion: https://4teamwork.slack.com/archives/C01BHEGBVU1/p1613382646011400?thread_ts=1613380520.009300&cid=C01BHEGBVU1

Jira: https://4teamwork.atlassian.net/browse/NE-366

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [x] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2021.4.0 (unreleased)
 ---------------------
 
+- Rename the attribute 'is_admin_menu_visible' from the @config endpoint to 'is_admin'. [elioschmutz]
 - Fix custom property choice field (de-)serialization. [deiferni]
 - Bump ftw.casauth to 1.3.1. [lgraf]
 - Add @save-document-as-pdf API endpoint. [tinagerber]
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -11,6 +11,7 @@ API Changelog
 Breaking Changes
 ^^^^^^^^^^^^^^^^
 
+- Rename the attribute ``is_admin_menu_visible`` from the ``@config`` endpoint to ``is_admin``.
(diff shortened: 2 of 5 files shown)


njohner (approved):
Hum I thought I remembered there was a discussion back then to decide whether we should expose an `is_admin` or rather `is_admin_menu_visible` and obviously decided for the latter. I can't remember why though, and there is nothing in the PR / issues (https://github.com/4teamwork/opengever.core/pull/6411). So for me this change makes sense

elioschmutz:
@njohner I also extended the PR description with the Slack-Discussion where I discussed it with @lukasgraf

lukasgraf:
@njohner correct, because properties like `is_admin` and `is_inbox_user` are prone to getting abused to do actual permission checks in the frontend. In this case however they're only being used for functional purposes - whether do display a card or not. Actual permission checks still happen in the backend.

And the other aspect is that Admin (and Inbox group members) are a bit special in what their roles imply - this is a case where we also do **Role-Checks** in the backend in some cases, even though that's an anti-pattern (access control should usually be done via permission checks). But doing that would result in a ton of fine-grained permissions that then just are all assigned to the `Administrator` role anyway.

And the situation in the frontend is similar, therefore I opted for making an exception in this case with my recommendation.

PR #6899, opened 2021-02-15
Add 'is_inbox_user' attribute to the @config endpoint

elioschmutz:
This PR introduces a new `@config` attribute: `is_inbox_user` which is required for the frontend to decide if inbox specific dashboard cards should be visible for the current user or not.

The condition for the `is_inbox_user` is taken from the old implementation: https://github.com/4teamwork/opengever.core/blob/ffe96663f855bcf81f7359b8817827f4bf9918a1/opengever/tabbedview/browser/personal_overview.py#L140

I think it makes no sense to refactor the old implementation to a reusable function. So I just reimplemented the required part directly in the `@config` endpoint. The decision who can see the tabs/cards is made in the frontend based on the newly `is_inbox_user` and `is_admin`(https://github.com/4teamwork/opengever.core/pull/6898)

Slack-Discussion: https://4teamwork.slack.com/archives/C01BHEGBVU1/p1613380520009300
Jira: https://4teamwork.atlassian.net/browse/NE-366

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/api/config.py
+++ b/opengever/api/config.py
@@ -4,6 +4,8 @@ from opengever.base.colorization import get_color
 from opengever.base.interfaces import IGeverSettings
 from opengever.inbox.utils import get_current_inbox
 from opengever.officeconnector.helpers import is_client_ip_in_office_connector_disallowed_ip_ranges
+from opengever.ogds.base.utils import get_current_org_unit
+from opengever.ogds.models.service import ogds_service
 from opengever.private import get_private_folder_url
 from plone.restapi.services import Service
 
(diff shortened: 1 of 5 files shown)


opengever/api/config.py
```diff
@@ -38,3 +40,7 @@ def add_additional_infos(self, config):
 
         inbox = get_current_inbox(self.context)
         config['inbox_folder_url'] = inbox.absolute_url() if inbox else ''
+
+        inbox = get_current_org_unit().inbox()
```
njohner, line 44: I think the inbox is already defined two lines above ;-)

njohner (approved):
apart from comment, LGTM

opengever/api/config.py
elioschmutz: it's not the same. The first one is the plone-inbox, this here is the inbox from the sql... But probably it's better to also name it like this 😉 

I'll update it to:

`plone_inbox` and `ogds_inbox`

PR #6900, opened 2021-02-16
Update policy templates

njohner:
For some reason we were setting the APPS_ENDPOINT_URL only for teamraum deployments. I think all new deployments will have the Gever-UI and need the APPS_ENDPOINT_URL to be set.

Moreover we now setup new deployments (on new servers) with several tools installed as services, notably sablon, msgconvert and pdflatex. I've now added a question that will set the URLs for these services correctly in sich cases.

## Checklist (Must have)

- [x] Changelog entry

For https://4teamwork.atlassian.net/browse/CA-984

### DIFF
--- a/opengever/policytemplates/policy_template/opengever.+package.name+/deployment-+base.server_name+-+base.deployment_number+-+base.domain+.cfg.bob
+++ b/opengever/policytemplates/policy_template/opengever.+package.name+/deployment-+base.server_name+-+base.deployment_number+-+base.domain+.cfg.bob
@@ -34,7 +34,7 @@ environment-vars +=
     BUMBLEBEE_INTERNAL_PLONE_URL https://{{{base.domain}}}/
     BUMBLEBEE_PUBLIC_URL https://{{{base.domain}}}/
     BUMBLEBEE_SECRET {{{base.bumblebee_secret}}}
-{{% if is_teamraum %}}
     APPS_ENDPOINT_URL {{{base.apps_endpoint_url}}}
+{{% if is_teamraum %}}
     WORKSPACE_SECRET {{{base.workspace_secret}}}
 {{% endif %}}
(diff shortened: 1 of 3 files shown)


opengever/policytemplates/policy_template/opengever.+package.name+/deployment-+base.server_name+-+base.deployment_number+-+base.domain+.cfg.bob
```diff
@@ -26,15 +26,23 @@ supervisor-memmon-instance-size = 1200MB
...
 [instance0]
 environment-vars +=
     BUMBLEBEE_APP_ID {{{base.bumblebee_app_id}}}
     BUMBLEBEE_INTERNAL_PLONE_URL https://{{{base.domain}}}/
     BUMBLEBEE_PUBLIC_URL https://{{{base.domain}}}/
     BUMBLEBEE_SECRET {{{base.bumblebee_secret}}}
-{{% if is_teamraum %}}
     APPS_ENDPOINT_URL {{{base.apps_endpoint_url}}}
+{{% if base.setup_with_services %}}
+    MSGCONVERT_URL http://localhost:8090/
```
lukasgraf, line 41: Are these ports also correct for production, not just local development? Also, this implies one of each these services for all of the deployments on a server, correct?

lukasgraf (approved):
LGTM 👍 
Just a question.

opengever/policytemplates/policy_template/opengever.+package.name+/deployment-+base.server_name+-+base.deployment_number+-+base.domain+.cfg.bob
njohner: yes true. At least for the two first productive deployments we have with the services, that is how it's configured.
@buchi can you confirm that the plan is one of each of these services per server?

buchi: yep, that's the plan.

PR #6902, opened 2021-02-17
Remove CORS policy

buchi:
It seems this was added accidentally. In general we don't want to allow any cross origin requests.

## Checklist (Must have)

- [ ] Changelog entry

### DIFF
--- a/opengever/base/configure.zcml
+++ b/opengever/base/configure.zcml
@@ -28,15 +28,6 @@
   <include file="schema.zcml" />
   <include file="widgets.zcml" />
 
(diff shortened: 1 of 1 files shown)


lukasgraf (approved):
LGTM 👍 

I added that CORS policy back in [this commit]https://github.com/4teamwork/opengever.core/pull/6001/commits/aeb3b0689496faeef0dd6d68812c11d0b2af34c2#diff-070bca8c99699f66b6b946749506db75c36c3e782126497f751c70e0114c3b98R31), and I'm reasonably certain that was an accidental checkin.

PR #6903, opened 2021-02-17
Adapt policy templates for ianus portal.

njohner:
As we now always use ianus for new deployments, there is no reason to use the `casauth` policy of the old portal in the policy templates. Also Authentication and resetting credentials should not happen over the ldap plugin. Authentication anyway gets deactivated during setup (https://github.com/4teamwork/opengever.core/blob/master/opengever/setup/deploy.py#L138-L141)

For https://4teamwork.atlassian.net/browse/CA-984

## Checklist (Must have)
- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2021.4.0 (unreleased)
 ---------------------
 
+- Adapt policy templates for ianus portal. [njohner]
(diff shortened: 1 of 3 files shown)


lukasgraf:
Why drop `ICredentialsResetPlugin`? (FYI, this basically means "logout" or "invalidate session". Not password reset).

lukasgraf (approved):
LGTM 👍!

PR #6904, opened 2021-02-19
Extend solrsearch endpoint, with breadcrumbs information option.

phgross:
Needed by the gever-ui, for displaying the breadcrumbs for each searchresult in the search results page. 

Because we do not have the information currently in SOLR and an indexing would lead to bad updating problems (changing the title of the repositoryfolder), we decide to use the real object for the generation, but introduce a flag and do not add the breadcrumbs by default. So it's only used in the detailed search results page, similar as we do in the classic ui. Discussed with buchi.

i was not sure how to implement the flag, that were my thoughts:
* Why not using the `expansion` functionality: The results is added to each result item not as a main key to the response.
* Why not handling the breadcrumbs as a requested field: the generation leads to longer requests, this would not be visible for the api user, if we would fake it as "solr-index"

JIRA: https://4teamwork.atlassian.net/browse/HG-994

## Checklist (Must have)

- [X] Changelog entry

## Checklist (if applicable)
- API change:
  - [X] Documentation is updated
  - [X] API Changelog entry (see [guide]- New functionality:

### DIFF
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -8,7 +8,10 @@ API Changelog
 2021.5.0 (unreleased)
 ---------------------
 
-- Nothing changed yet.
+Other Changes
+^^^^^^^^^^^^^
+
+- Add ``breadcrumbs`` option to the ``@solrsearch`` endpoint.
 
 
 2021.4.1 (2021-02-25)
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -11,7 +11,7 @@ API Changelog
 Other Changes
 ^^^^^^^^^^^^^
 
-- Add ``breadcrumbs`` option to the ``@solrsearch`` endpoint.
+- Add ``breadcrumbs`` option to the ``@solrsearch`` endpoint, only available for small batch sizes (max. 50 items).
 
 
 2021.4.1 (2021-02-25)
--- a/docs/public/dev-manual/api/searching.rst
+++ b/docs/public/dev-manual/api/searching.rst
@@ -229,6 +229,10 @@ Batching
 ~~~~~~~~
 Der Endpoint stellt die Standard-Paginierung gem :ref:`Kapitel Paginierung <batching>` zur Verfügung.
 
+Breadcrumbs
+~~~~~~~~~~~
+Wird eine ``@solrsearch`` Anfrage mit ``breadcrumbs=1`` Parameter ergänzt, so werden die einzelnen Suchtreffer unter dem Key ``breadcrumbs`` mit den Breadcrumb Informationen ergänzt.
+
 
 Teamraum Solr Suche
 -------------------
--- a/docs/public/dev-manual/api/searching.rst
+++ b/docs/public/dev-manual/api/searching.rst
@@ -231,7 +231,7 @@ Der Endpoint stellt die Standard-Paginierung gem :ref:`Kapitel Paginierung <batc
 
 Breadcrumbs
 ~~~~~~~~~~~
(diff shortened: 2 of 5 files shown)


opengever/api/solrsearch.py
```diff
@@ -80,6 +90,14 @@ def parse_requested_fields(self, params):
                 # Requesting additional fields is required in order to determine if
                 # the repository folder is a leaf node.
                 requested_fields.extend(['@type', 'has_sametype_children'])
+
+        # Add path to the requested_fields, used for breadcrumb generation
+        if self.show_breadcrumbs:
+            if requested_fields:
+                requested_fields.append('path')
+            else:
+                requested_fields = ['path']
```
njohner, line 99: Normally if the user does not request any fields, he will get the default fields. Here you change that behavior so that if a user does not request any fields but sets the `show_breadcrumbs` flag, he will only get the "path" field back.

njohner (commented):
I think I would not add the `path` field in the `parse_requested_fields` method, because that method is supposed to parse the requested fields 😅 , and you have a problem when the user does not request any. Maybe overwrite `extract_field_list` instead?

elioschmutz:
The frontend provides a batchsizes up to 200 for the search view. But with this batch-size it would takes about 3 seconds for one request and the frontend will continuously request new results while typing. This will break the instances...

So I think we should decrease the batch-size in the frontend and also introduce a max row limit of about 100 or even less, 50 if using the `breadcrumbs` option...

phgross:
@njohner thx for the review, updated according to your and elios feedback.

njohner:
@elioschmutz do we have some latency before sending a request in the frontend in the live search? maybe that is also an issue? If I'm typing fast I don't expect to see an update of the results. Also if I understand correctly we will not return breadcrumbs in the live search right?

docs/public/dev-manual/api/searching.rst
```diff
@@ -231,7 +231,7 @@ Der Endpoint stellt die Standard-Paginierung gem :ref:`Kapitel Paginierung <batc
 
 Breadcrumbs
 ~~~~~~~~~~~
-Wird eine ``@solrsearch`` Anfrage mit ``breadcrumbs=1`` Parameter ergänzt, so werden die einzelnen Suchtreffer unter dem Key ``breadcrumbs`` mit den Breadcrumb Informationen ergänzt.
+Wird eine ``@solrsearch`` Anfrage mit ``breadcrumbs=1`` Parameter ergänzt, so werden die einzelnen Suchtreffer unter dem Key ``breadcrumbs`` mit den Breadcrumb Informationen ergänzt. Diese Ergänzung ist nur bei kleinen Batchsizes (maximal 50) erlaubt.
```
njohner, line 234: @elioschmutz so I guess in the frontend the user is not allowed to choose the number of results per page anymore?

I'm not quite sure this is really necessary and it might lead to some confusion. But OK for me nonetheless, especially since we raise a specific and pretty explicit error in that case.

docs/public/dev-manual/api/api_changelog.rst
```diff
@@ -8,7 +8,10 @@ API Changelog
 2021.5.0 (unreleased)
 ---------------------
 
-- Nothing changed yet.
+Other Changes
+^^^^^^^^^^^^^
+
+- Add ``breadcrumbs`` option to the ``@solrsearch`` endpoint.
```
njohner, line 14: I think we should specify the batch size limit here. It needs to be visible for users of the API.

njohner (approved):
LGTM. Just maybe add batch size limit in the API changelog.

docs/public/dev-manual/api/searching.rst
phgross: Yes, the choose option is disabled by https://github.com/4teamwork/gever-ui/pull/1619.

PR #6907, opened 2021-02-22
Fix rejecting submitted proposal containing mail with extracted trashed attachment.

njohner:
- When refusing a submitted proposal, it gets deleted (https://github.com/4teamwork/opengever.core/blob/master/opengever/meeting/proposal.py#L411-L412)
- Because an E-mail is deleted, we end up in the `mail_deleted` handler which tries to remove the `IExtractedFromMail` interface from the extracted document.
- This fails because the extracted document is trashed and hence `uuidToObject` returns None because the document is trashed

We simply fix this by using an unrestricted version of `uuidToObject`. 

I did not add tests, because I think the whole scenario is very convoluted and should actually not happen. I actually think it's wrong that attachments from E-mails in a submitted proposal get extracted into the meeting dossier. I've created a follow-up story for that (https://4teamwork.atlassian.net/browse/CA-1822).

For https://4teamwork.atlassian.net/browse/CA-1739

## Checklist (Must have)
- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2021.5.0 (unreleased)
 ---------------------
 
+- Fix rejecting submitted proposal containing mail with extracted trashed attachment. [njohner]
 - Also set title_en and title_fr for meetings in policy templates. [njohner]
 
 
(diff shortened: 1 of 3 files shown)


lukasgraf (approved):
LGTM 👍

PR #6909, opened 2021-02-23
No longer show warning about failed notification deliveries if recipient user doesn't have an email address

lukasgraf:
A user might have some notifications enabled for the "E-Mail" channel, but not have an email address tied to their account / OGDS user.

Before this change, this resulted in the actor that triggered the notification getting an unhelpful warning message saying that one or more notifications could not be dispatched (because an attempt was made to have the Mailhost send an email with a recipient of `None`).

This PR changes this behavior so that this case is silently ignored (except for a warning that gets logged).

In addition, exceptions that happen during notification dispatching will now be logged to Sentry.

Follow-Up Story: [CA-1829](https://4teamwork.atlassian.net/browse/CA-1829) _(Benachrichtigungseinstellungen: Channels ausgrauen falls keine E-Mail Adresse hinterlegt)_

Jira: https://4teamwork.atlassian.net/browse/CA-1661

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/activity/tests/test_plone_center.py
+++ b/opengever/activity/tests/test_plone_center.py
@@ -122,7 +122,7 @@ class TestNotifactionCenterErrorHandling(FunctionalTestCase):
         self.assertEquals(['Item created'], info_messages())
 
     @browsing
-    def test_shows_message_if_dispatchers_raise_an_exception(self, member):
+    def test_missing_email_address_for_notification_recipient_doesnt_produce_warning(self, member):
         create(Builder('ogds_user')
                .having(userid='hugo.boss', email=None)
                .in_group(self.org_unit.users_group))
@@ -137,8 +137,5 @@ class TestNotifactionCenterErrorHandling(FunctionalTestCase):
 
         browser.css('#form-buttons-save').first.click()
 
-        self.assertEquals(
-            ['A problem has occurred when trying to dispatch a notification. '
-             'The notification therefore was not dispatched (or only partially).'],
-            warning_messages())
+        self.assertEquals([], warning_messages())
         self.assertEquals(['Item created'], info_messages())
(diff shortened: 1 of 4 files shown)


opengever/activity/tests/test_plone_center.py
```diff
@@ -122,7 +122,7 @@ def test_successfully_add_activity(self, member):
         self.assertEquals(['Item created'], info_messages())
 
     @browsing
-    def test_shows_message_if_dispatchers_raise_an_exception(self, member):
+    def test_missing_email_address_for_notification_recipient_doesnt_produce_warning(self, member):
```
njohner, line 125: We don't seem to have any test anymore that the message if shown if something goes wrong during dispatch. Well I guess we don't really want that message to appear in any reasonable scenario anyway...

njohner (approved):
LGTM. So now we log a warning for missing E-mail and log an error to sentry for other failed dispatches, but not for missing E-mail.

opengever/activity/tests/test_plone_center.py
lukasgraf: @njohner exactly. I spent some time trying to construct one that still produces the message for other scenarios, but it turned out to be nearly impossible. The dispatching (and error handling thereof) happens during form submit. But any of the obvious violations that could lead to errors during dispatching would already lead to form validations failing, therefore preventing the form from even being submitted.

And yes, in the end, after looking at the code that's involved, I would say the scenarios that lead to this message being displayed are now few and far in between. The main one is actions performed by or involving zopemaster.

lukasgraf:
> LGTM. So now we log a warning for missing E-mail and log an error to sentry for other failed dispatches, but not for missing E-mail.

Exactly.

PR #6910, opened 2021-02-24
Bump plone.formwidget.namedfile to version 2.0.6.

njohner:
This is to get a fix for a bug where the selected file gets discared upon validation error of the form. This has only become an issue with the introduction of the custom properties, which can lead to validation errors in the document add form.

Note that this is the newest version of the widget we can bump to without needing to bump other dependencies.

Also note that I had to fix one test. The widget for the message field of mails always gets displayed in the NO_DOWNLOAD_DISPLAY_MODE, which leads to an error when there is no message. As a Mail without a message is anyway not valid, this is a problem only in tests, for mails constructed with the builders. I checked that this is not a problem for documents that don't have a file for example.

For https://4teamwork.atlassian.net/browse/CA-1507

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2021.5.0 (unreleased)
 ---------------------
 
+- Bump plone.formwidget.namedfile to version 2.0.6 to fix bug with discared file upon validation error. [njohner]
 - Fix rejecting submitted proposal containing mail with extracted trashed attachment. [njohner]
 - Also set title_en and title_fr for meetings in policy templates. [njohner]
 
--- a/opengever/base/locales/plone.pot
+++ b/opengever/base/locales/plone.pot
@@ -431,3 +431,6 @@ msgstr ""
 
 msgid "label_name"
 msgstr ""
+
+msgid "file_already_uploaded"
+msgstr ""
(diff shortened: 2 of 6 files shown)


lukasgraf (commented):
Did you go over the relevant changelog section for the bump we would be doing? So `1.0.15` to  `2.0.6`?

njohner:
Yeah I went over the changelog and didn't see anything that concerned me. It's actually a rather small jump all things considered.

lukasgraf (approved):
Changes as such look good to me 👍 
It's just the thing we discussed: Since this affects rather critical functionality in GEVER (uploading, downloading and displaying files), I would have a closer look at a few things:

It does *seem* to work, but I don't quite understand why. I would expect our `no_download_input.pt` and `no_download_display.pt` templates needing to be adapted, but somehow in seems to work in the cases I tested.

One important case needs to be checked though: If a document is currently being edited by someone else, I must not be able to gain access to the working copy in progress via the widget's download link (that's the rationale behind our `NO_DOWNLOAD_...` modes).

njohner:
This solution is deemed too risky, so we decided not to go down this path. Superseded by https://github.com/4teamwork/opengever.core/pull/6919

PR #6911, opened 2021-02-24
Cleanup conditionals protecting for changed date not set yet.

njohner:
The changed date has now been indexed on all deployments and the conditionals protecting against states where it had been introduced but not set yet are no longer needed.

Note that I had to delete two tests, that were actually testing some of the conditionals that were only introduced temporarily as a safeguard for missing changed dates.

For https://4teamwork.atlassian.net/browse/CA-1667

## Checklist (Must have)
- [x] Changelog entry

### DIFF
--- a/opengever/tabbedview/catalog_source.py
+++ b/opengever/tabbedview/catalog_source.py
@@ -178,11 +178,8 @@ class GeverCatalogTableSource(FilteredTableSourceMixin, CatalogTableSource):
             else:
                 filters.append(u'{}:{}'.format(key, escape(value)))
 
-        # Todo: modified be removed once the changed metadata is filled on
-        # all deployments.
-        # https://github.com/4teamwork/opengever.core/issues/4988
         fl = ['UID', 'getIcon', 'portal_type', 'path', 'id',
-              'bumblebee_checksum', 'modified']
+              'bumblebee_checksum']
         fl = fl + [c['column'] for c in self.config.columns if c['column']]
         params = {
             'fl': fl,
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2021.5.0 (unreleased)
 ---------------------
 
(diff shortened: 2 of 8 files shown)


opengever/tabbedview/catalog_source.py
```diff
@@ -178,11 +178,8 @@ def solr_results(self, query):
             else:
                 filters.append(u'{}:{}'.format(key, escape(value)))
 
-        # Todo: modified be removed once the changed metadata is filled on
-        # all deployments.
-        # https://github.com/4teamwork/opengever.core/issues/4988
         fl = ['UID', 'getIcon', 'portal_type', 'path', 'id',
-              'bumblebee_checksum', 'modified']
+              'bumblebee_checksum']
```
lukasgraf, line 182: Hmm. Are we sure about this one? Shouldn't `changed` then be added to the field list? And could this possibly result in a breaking API change in one of the endpoints?

lukasgraf (commented):
LGTM, apart from one question.

opengever/tabbedview/catalog_source.py
njohner: Yes this should be fine. We needed `modified` here because:
- In the documents tab, we started displaying [changed instead of modified](https://github.com/4teamwork/opengever.core/blob/master/opengever/tabbedview/browser/tabs.py#L156-L161)
- This means we are requesting the `changed` column from solr [because it's one of the columns](https://github.com/4teamwork/opengever.core/blob/master/opengever/tabbedview/catalog_source.py#L186)
- We had a fallback in the transform to get [modified instead if changed was empty](https://github.com/4teamwork/opengever.core/blob/master/opengever/tabbedview/helper.py#L257-L263)
- For this fallback to work we needed to request `modified` as well. Hence the above. We do not need this anymore.

As for your second question, this is the catalog source, this is not exposed over the API.

lukasgraf (approved):
LGTM 👍

PR #6913, opened 2021-02-25
Include mails in dispositions and use changed instead of modified for date range calculation

njohner:
With this PR we:
- Now also include mails in `Disposition`s.
- Use `changed` instead of `modified` in the determination of dossier and document date range calculations.

For https://4teamwork.atlassian.net/browse/CA-1467

## Checklist (Must have)
- [x] Changelog entry


opengever/disposition/tests/test_model.py
```diff
@@ -187,16 +187,16 @@ def test_entstehungszeitraum_includes_mails_in_calculation(self):
 
         create(Builder('document')
                .within(self.empty_dossier)
-               .with_modification_date(DateTime(2016, 3, 1))
-               .with_creation_date(DateTime(2015, 1, 1)))
+               .with_modification_date(DateTime(2016, 3, 1, 12))
```
njohner, line 191: Time is added here because of timezone issues. The Zope `DateTime` is GMT+1 and if no hour is specified it is set to midnight. Because `changed` is localized to UTC in the builders, this would when change the date to the day prior.

lukasgraf (approved):
LGTM! 👍

PR #6917, opened 2021-02-26
Dispatch notification for documents added to tasks.

lukasgraf:
With this change we now dispatch notifications for documents added to tasks (controlled via the `task commented` notification settings group).

This activity was previously unmapped in the configuration, and therefore produced errors during notification settings lookup, as well as not being dispatched.

The notification settings group "task commented" is not a perfect fit, but the least bad of the available ones, and we don't want to undo the simplification by introducing yet another group.

In addition, errors during notification lookup are now handled more specifically, by raising an intermediate `UnmappedActivityKind` exception that ends up in Sentry together with a sensible traceback (instead of [producing an `AttributeError`](https://sentry.4teamwork.ch/organizations/sentry/issues/71995/?project=17&query=is%3Aunresolved+cluster%3Agever.fmh.ch&statsPeriod=14d) somewhere down the line). The catch-all error handling will still apply though (doesn't prevent the actor's operation), and produce the warning message about some notifications not getting dispatched.

Jira: https://4teamwork.atlassian.net/browse/CA-1661

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/activity/tests/test_notification_center.py
+++ b/opengever/activity/tests/test_notification_center.py
@@ -9,6 +9,7 @@ from opengever.activity.model import Notification
 from opengever.activity.model import Resource
 from opengever.activity.model import Subscription
 from opengever.activity.model import Watcher
+from opengever.activity.notification_settings import UnmappedActivityKind
 from opengever.activity.roles import TASK_ISSUER_ROLE
 from opengever.activity.roles import TASK_RESPONSIBLE_ROLE
 from opengever.activity.roles import WATCHER_ROLE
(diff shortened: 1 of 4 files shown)


opengever/activity/tests/test_notification_center.py
```diff
@@ -659,6 +660,27 @@ def test_digest_dispatcher_sets_digest_flag_depending_on_the_setting(self):
         self.assertFalse(hugos_note.is_digest)
         self.assertTrue(peters_note.is_digest)
 
+    def test_unmapped_activity_kind_raises_specific_exception(self):
+        """If an activity kind is not mapped at all to a notification
+        settings group at all, something went wrong and we'll want to know.
```
njohner, line 665: twice "at all"

opengever/activity/tests/test_notification_center.py
```diff
@@ -659,6 +660,27 @@ def test_digest_dispatcher_sets_digest_flag_depending_on_the_setting(self):
...
+
+        A specific exception should therefore be raised for this. When using
+        the higher level add_activity() method on the PloneNotificationCenter
+        though, it will be caught though and won't prevent the attempted
```
njohner, line 669: twice though

njohner (approved):
Apart from comment, LGTM

PR #6918, opened 2021-02-26
The @sharing endpoint returns a batched result set if using the search param.

elioschmutz:
The default local-roles serializer returns all current local roles by default. If we pass a search-param, it will return all available entries which could possibly be added to the local roles. This result set can contain a lot of entries:

Bildschirmfoto 2021-02-26 um 09 57 41

Currently, such a response will break the UI.

So this PR customizes the local-roles serializer to return a batched result-set if searching for entries but returns an unbatched result-set if just requesting the current local roles.

Bildschirmfoto 2021-02-26 um 11 47 27

ℹ️ The default batch-size is at 25 which should solve the issue. We do not implement batching in the frontend-sharing-tab, but just display the available entries.

Jira: https://4teamwork.atlassian.net/browse/CA-1831

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [x] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2021.5.0 (unreleased)
 ---------------------
 
+- The @sharing endpoint now returns a batched result set if using the search param.  [elioschmutz]
 - Cleanup conditionals protecting for changed date not set yet. [njohner]
 - Use changed instead of modified in date range calculation for SIP packages. [njohner]
 - Include mails in SIP package. [njohner]
--- a/opengever/api/local_roles.py
+++ b/opengever/api/local_roles.py
@@ -1,0 +1,41 @@
+from AccessControl.interfaces import IRoleManager
+from opengever.base.interfaces import IOpengeverBaseLayer
+from plone.restapi.batching import HypermediaBatch
+from plone.restapi.interfaces import ISerializeToJson
+from plone.restapi.serializer.local_roles import SerializeLocalRolesToJson
+from zope.component import adapter
+from zope.interface import implementer
+
+
+@adapter(IRoleManager, IOpengeverBaseLayer)
+@implementer(ISerializeToJson)
+class GeverSerializeLocalRolesToJson(SerializeLocalRolesToJson):
+    """The default local-roles serializer returns all current local roles
+    by default. If we pass a search-param, it will return all available
+    entries which could possibly be added to the local roles. This result
+    set can contain a lot of entries.
+
+    The GEVER-customization of this serializer returns a batched result-set
+    if searching for entries but returns an unbatched result-set
+    if just requesting the current local roles.
+    """
+    def __init__(self, context, request):
+        self.context = context
+        self.request = request
+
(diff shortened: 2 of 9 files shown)


docs/HISTORY.txt
```diff
@@ -4,6 +4,7 @@ Changelog
 2021.5.0 (unreleased)
 ---------------------
 
+- The @sharing endpoint returns a batched result set if using the search param.  [elioschmutz]
```
lukasgraf, line 7: I would phrase this as "... endpoint **now** returns ..." to make it clear what was the behavior before and after. (Same for API changelog).

_Change_ logs should describe a _change_, not a state. 😉

opengever/api/local_roles.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def __call__(self, search=None):
+        results = super(GeverSerializeLocalRolesToJson, self).__call__(search)
+
+        entries = results.get('entries', [])
+        if not search:
+            results['entries_total'] = len(entries)
+        else:
+            batch = HypermediaBatch(self.request, entries)
+            results['entries_total'] = batch.items_total
```
lukasgraf, line 34: No - please do not invent a different batching parameter style!
It's `items`, and `items_total`, etc..., as documented in [our docs](https://docs.onegovgever.ch/dev-manual/api/batching/) and [upstream in plone.restapi](https://plonerestapi.readthedocs.io/en/latest/batching.html).

lukasgraf (changes requested):

lukasgraf:
Also:
>  [x] Documentation is updated

Really? (Just a quick description and an internal link to our [batching docs](https://docs.onegovgever.ch/dev-manual/api/batching/) is enough)

opengever/api/local_roles.py
elioschmutz: The `local_roles` serializer uses the key `entries`. So I use `entries_total` which is the logical pendent for `entries`.

lukasgraf: Yes, but while that may be a stylistic consideration, the batching protocol being the same has actual practical implications: In order for people to be able to write reusable pagination implementations in their clients, we must use the same protocol everywhere. How else could a batching iterator like [`batchingService.js`](https://github.com/4teamwork/gever-ui/blob/master/src/batchingService.js) in gever-ui ever work? Or similar implementations in RIS / VM? 

If anything, the `entries` key is the odd one out here and should probably be renamed to `items`. If we're already making breaking changes to the `@sharing` endpoint, this would probably be as good a time as any to do it.

elioschmutz: as discussed, I renamed `entries` to `items` and `entries_total` to `items_total` only if the `search` parameter is available.

elioschmutz:
@lukasgraf can you take a final look pls?

lukasgraf (approved):
LGTM, thanks! 👍

PR #6919, opened 2021-03-01
Reset value of NamedFileWidget in DocumentAddForm when validation fails.

njohner:
When validation fails when a new file is uploaded over the `NamedFileWidget`, it will then behave as if it already had a value set, but this will not work correctly upon resubmission of the form because the file was not stored server side.
This problem is addressed in newer version of the widget, but it comes with its own set of problems, so [we gave up that path](https://github.com/4teamwork/opengever.core/pull/6910).

Here we propose another fix for the document add form. We reset the value of `NamedFileWidget` in the `DocumentAddForm` when validation fails. This is not ideal, but at least the widget does correctly show that currently no file will be uploaded if the form is submitted again. This covers the most likely scenario when using custom
property sheets in the classic UI.

This does not fix the issue we could have in the edit form, but this is much less likely, as it requires changing the file of the edit metadata form and at the same time changing the document type to one that requires new custom properties. This is an edge case we can live with.

For https://4teamwork.atlassian.net/browse/CA-1507

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/opengever/document/forms.py
+++ b/opengever/document/forms.py
@@ -55,6 +55,19 @@ class DocumentAddForm(add.DefaultAddForm):
         super(DocumentAddForm, self).updateFields()
         self.groups = omit_custom_properties_group(self.groups)
 
+    @button.buttonAndHandler(_('Save'), name='save')
+    def handleAdd(self, action):
+        """ Reset value of NamedFileWidget when validation fails, to avoid
+        displaying an already uploaded file with "Keep existing file" radio
+        button, which would actually not get set upon form resubmission.
+        """
+        super(DocumentAddForm, self).handleAdd(self, action)
+        if self.status != self.formErrorsMessage:
+            return
+        for group in self.groups:
+            if 'file' in group.widgets:
+                group.widgets.get('file').value = None
+
 
 class DocumentAddView(add.DefaultAddView):
     """Provide a registerable view for the Document file upload form."""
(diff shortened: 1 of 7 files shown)


opengever/document/forms.py
```diff
@@ -55,6 +55,15 @@ def updateFields(self):
         super(DocumentAddForm, self).updateFields()
         self.groups = omit_custom_properties_group(self.groups)
 
+    @button.buttonAndHandler(_('Save'), name='save')
+    def handleAdd(self, action):
+        super(DocumentAddForm, self).handleAdd(self, action)
+        if self.status != self.formErrorsMessage:
```
lukasgraf, line 61: Could you add a short comment here that describes why we do this?
Just looking at this code as-is it's quite hard to judge what it aims to accomplish.

Something like
```
# Avoid having file widget say "Keep existing file" (which it wouldn't) 
# by resetting it upon on form validation failure.
```
?

lukasgraf (approved):
LGTM 👍 , apart from the request for a clarifying comment 😉

opengever/document/forms.py
njohner: Sorry, that was planned, and somehow I ended up only commenting in the test 😅 .

njohner:
I'll wait with this one until the release is created. I don't see any reason to rush it, as namedfile widget is quite central and we don't want it having issues in SG for example...

PR #6920, opened 2021-03-02
Don't resolve or deactivate a dossier if it has linked workspaces without view permission.

elioschmutz:
This PR is a follow-up PR of: https://github.com/4teamwork/opengever.core/pull/6642

It should not be possible to resolve/deactivate a dossier if there are active linked workspaces. It's possible, that the current user does not see some linked workspaces because he does not have the `View` permission on it. So the check in https://github.com/4teamwork/opengever.core/pull/6642 will not detect such workspaces.

The `elevated_privileges` is only for the GEVER-instance and has no effect on the teamraum instance. So even if we run it with elevated privileges, the request to the teamraum is made with the currently logged in user and the permissions on the source instance, so the teamraum. It just makes sure, that the current user can access the `WorkspaceClient`.

We could refactor this, but that's not part of this bugfix!

To fix the issue, this PR introduces a new method `has_linked_workspaces_without_view_permission` which returns a boolean indicating if there are linked workspaces where the current user does not have the view permission. This allows us to introduce a new check for the resolve/deactivate action with a specific error message.

Bildschirmfoto 2021-03-04 um 10 13 39

Fixes: https://4teamwork.atlassian.net/browse/CA-1851

## Checklist (Must have)

- [x] Changelog entry

## Checklist (if applicable)

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/dossier/base.py
+++ b/opengever/dossier/base.py
@@ -300,6 +300,9 @@ class DossierContainer(Container):
         return bool(active_proposals)
 
     def is_linked_to_active_workspaces(self):
+        """Returns all linked active workspaces accessible by the current
+        user.
+        """
         if not is_workspace_client_feature_enabled():
             return False
 
@@ -311,6 +314,22 @@ class DossierContainer(Container):
             params = {'review_state': 'opengever_workspace--STATUS--active'}
             active_workspaces = linked_workspaces_adapter.list_non_cached(**params)['items_total']
             return bool(active_workspaces)
+
+    def has_linked_workspaces_without_view_permission(self):
+        """It's possible that the currently logged in user has no view permission
+        to all linked workspaces. This function returns a boolean to indicate
+        if there are such workspaces linked with the current dossier.
+        """
+        if not is_workspace_client_feature_enabled():
+            return False
+
+        with elevated_privileges():
+            linked_workspaces_adapter = queryAdapter(self, ILinkedWorkspaces)
+            if not linked_workspaces_adapter:
+                return False
+
+            linked_workspaces = linked_workspaces_adapter.list_non_cached()['items_total']
+        return bool(len(linked_workspaces_adapter.storage.list()) - linked_workspaces)
 
     def is_all_checked_in(self):
         """Check if all documents in this path are checked in."""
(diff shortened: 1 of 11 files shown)


opengever/dossier/base.py
```diff
@@ -312,6 +315,22 @@ def is_linked_to_active_workspaces(self):
...
+        """
+        if not is_workspace_client_feature_enabled():
+            return False
+
+        with elevated_privileges():
+            linked_workspaces_adapter = queryAdapter(self, ILinkedWorkspaces)
+            if not linked_workspaces_adapter:
+                return False
+
+            linked_workspaces = linked_workspaces_adapter.list_non_cached()['items_total']
```
njohner, line 331: hum, but that means that a user who doesn't have view permission on a linked workspace that is not active anymore, will not be able to close the dossier? That does not seem correct to me.

njohner (commented):
I guess this is an improvement but it's really barely good enough. Non-active workspaces to which the user has no access should not prevent closing the dossier.
If I look at https://4teamwork.atlassian.net/browse/CA-1180, the second variant of the proposed implementation would actually allow to do things right, both for dossier resolution and for representation of the workspaces. So if we go with this for now, I think we should clarify in https://4teamwork.atlassian.net/browse/CA-1180, that we will implement variant 1 and come back to this here and fix it.

opengever/dossier/base.py
elioschmutz: @njohner Discussed in https://4teamwork.atlassian.net/browse/CA-1851 with @phgross and Erich. Decision is here: https://4teamwork.atlassian.net/browse/CA-1851?focusedCommentId=22434

The behavior is correct. Because a user who cannot see all workspaces, even if they're closed, can't decide if there are still relevant documents within the workspaces or not. A user who wants to close a dossier needs to have access to every linked workspace.

elioschmutz:
@njohner I force-pushed and updated the error-message according to the input of the PO

njohner (approved):
LGTM

PR #6921, opened 2021-03-03
Filter out trash and untrash folder buttons in @actions endpoint on repofolders.

njohner:
In the new frontend, the `trash` action is visible in the documents tab on repository folder level. Because the documents listing shows all documents, even the the ones in resolved / inactive dossiers, the action can fail for certain documents. This failure is not handled by the frontend.
Here we implement the same approach as in the classic UI, i.e. we simply [do not show the action at the repository level](https://github.com/4teamwork/opengever.core/blob/master/opengever/repository/browser/tabs.py#L8-L12). I think this makes sense, I don't really see a use case for trashing documents from different main dossiers at the same time.

For https://4teamwork.atlassian.net/browse/CA-1828

## Checklist (Must have)

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -7,6 +7,7 @@ Changelog
 - Fixed automatic start of a next task inside a sequential task process. [phgross]
 - Only show "add task to process" link, if next task is not yet started. [phgross]
 - Fix adding sequential task process on first position. [phgross]
+- Filter out trash and untrash folder buttons in @actions on repository root and folders. [njohner]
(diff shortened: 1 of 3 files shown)


elioschmutz (commented):
What's with the RepoRoot? Should't the actions be removed from the root as well?

njohner:
There is no document listing on the repo root ;-)

njohner:
Oh never mind, that is not True for the frontend!

njohner:
I also filter out the actions on the repository root now. Thanks for the catch!

elioschmutz (approved):

PR #6922, opened 2021-03-03
Update PR template with creation of follow-up stories

njohner:
As discussed in the sprint review, when a PR leaves a state behind that will require further improvements, for example because it leads to some technical debt, does not cover some edge cases, or is simply an incomplete implementation of a feature that we know will lead to support requests, complaints, confusion further down the line, we want to make sure to create follow-up stories to sort these out.
This is added as a reminder in the PR template.

Moreover I propose to simplify the checklist a bit, make it a single checklist with a mandatory and optional part.

## Checklist (Must have)

- [ ] Changelog entry -> no changelog here
- [ ] Link to issue (Jira or GitHub) and backlink in issue (Jira) -> no jira issue, came from sprint review

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -11,15 +11,12 @@ _Screenshot: whenever useful, but only as a visual aid._
 Definition of Done: https://4teamwork.atlassian.net/wiki/spaces/CHX4TW/pages/917562/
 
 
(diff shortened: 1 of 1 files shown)


elioschmutz (approved):

njohner:
I'll merge that one before the release...

PR #6923, opened 2021-03-03
Sequential task fixes

phgross:
This PR provides some fixes for the sequential task process feature:
- Fixed automatic start of a next task inside a sequential task process.
- Only show "add task to process" link, if next task is not yet started.
- Fix adding sequential task process on first position.

This ensures that only one task is open in a process.

For https://4teamwork.atlassian.net/browse/CA-1429

## Checklist (Must have)
- [x] Changelog entry
- [x] Works also for the new UI

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -13,6 +13,7 @@ Changelog
 
 - Fix support in TransitionExtender for RelationChoice fields. [phgross]
 - Allow any authenticated users to use the REST API. [phgross]
+- Fix adding sequential task process on first position. [phgross]
 - The @sharing endpoint now returns a batched result set if using the search param.  [elioschmutz]
 - Cleanup conditionals protecting for changed date not set yet. [njohner]
 - Use changed instead of modified in date range calculation for SIP packages. [njohner]
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,9 @@ Changelog
 2021.6.0 (unreleased)
 ---------------------
 
+- Fixed automatic start of a next task inside a sequential task process. [phgross]
+- Only show "add task to process" link, if next task is not yet started. [phgross]
+- Fix adding sequential task process on first position. [phgross]
 - Don't resolve or deactivate a dossier if it has linked workspaces without view permission. [elioschmutz]
 - Reset value of NamedFileWidget in DocumentAddForm when validation fails. [njohner]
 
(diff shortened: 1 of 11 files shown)


njohner (commented):
Fixes look good, but if possible using the same code for the new and classic UI would be nice.
Also I tested the branch locally and two of the problems I'd found seem to be still present:
- skip a task and then reopen it -> on the main task in the sequential task representation, it is still represented as skipped
- Skip a task, resolve the previous task in the sequence. It opens the next one. Then reopen the skipped task. Two open tasks. But this one is probably unimportant.

phgross:
Hi @niklaus I've fixed the issue with `skipped-open` transition. There was no TransitionExtender registered for this transition. This leads to two "open" tasks inside a sequential process, but to no other problems or issues.

I also tried to use the same code for the api (new UI) and the classic UI, but currently the complete task overview is based on the SQL Task and not the real object. So I don't think its worth the risk to refactor the TaskOverview or do unneccessary SQL queries in the API Endpoint.

docs/HISTORY.txt
```diff
@@ -13,6 +13,7 @@ Changelog
 
 - Fix support in TransitionExtender for RelationChoice fields. [phgross]
 - Allow any authenticated users to use the REST API. [phgross]
+- Fix adding sequential task process on first position. [phgross]
```
njohner: Move it to 2021.6.0

docs/HISTORY.txt
```diff
@@ -13,6 +13,7 @@ Changelog
 
 - Fix support in TransitionExtender for RelationChoice fields. [phgross]
 - Allow any authenticated users to use the REST API. [phgross]
+- Only show "add task to process" link, if next task is not yet started. [phgross]
```
njohner: This one too

docs/HISTORY.txt
```diff
@@ -13,6 +13,7 @@ Changelog
 
 - Fix support in TransitionExtender for RelationChoice fields. [phgross]
 - Allow any authenticated users to use the REST API. [phgross]
+- Fixed automatic start of a next task inside a sequential task process. [phgross]
```
njohner: and this one

docs/HISTORY.txt
```diff
@@ -4,6 +4,9 @@ Changelog
 2021.6.0 (unreleased)
 ---------------------
 
+- Fixed automatic start of a next task inside a sequential task process. [phgross]
+- Only show "add task to process" link, if next task is not yet started. [phgross]
+- Fix adding sequential task process on first position. [phgross]
```
njohner, line 9: ah ok 👍

njohner (approved):
LGTM!

PR #6924, opened 2021-03-03
Filter out folder_delete folder button in @actions on repofolders.

njohner:
`folder_delete` action is available in new frontend for administrators on the repository folder level, but it isn't actually allowed to delete any objects. In the classic UI this action is not enabled on the documents tab of the repository folder. We do the same here and filter it out. 

This was not a requirement of any story, but stumbled upon it while looking at https://4teamwork.atlassian.net/browse/CA-1828

## Checklist (Must have)
- [x] Changelog entry


phgross (approved):
👍 lgtm - besides the failing tests 😉

njohner:
Hum tests were failing because I managed to screw up the `actions.xml`. Now all good 😅

PR #6927, opened 2021-03-04
Fix support in TransitionExtender for RelationChoice fields.

phgross:
The RelationChoice validator from the plone.restapi, make changes on the passed in data, so its necessary to pass in a deepcopy. Otherwise we pass in a list of objects instead of paths to the `doActionFor` call, which leads to an error in restapi's deserializer.

https://4teamwork.atlassian.net/browse/CA-1693

## Checklist
- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2021.5.0 (unreleased)
 ---------------------
 
+- Fix support in TransitionExtender for RelationChoice fields. [phgross]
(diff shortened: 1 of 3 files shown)


njohner (approved):
LGTM. 
I think this might be the corresponding sentry issue?
https://sentry.4teamwork.ch/organizations/sentry/issues/71540

phgross:
@njohner thanks for the review! 

No this case is something different, I think there is no corresponding sentry issue, because the new UI had not provided the possibility until now to add relations when changing a transition.

PR #6930, opened 2021-03-09
Add edit_items folder button action.

tinagerber:
This action is needed in the new frontend to edit the metadata of multiple dossiers or documents at the same time.
The action should be available for all dossier and document listings.

Jira: https://4teamwork.atlassian.net/browse/NE-682

## Checklist

- [x] Changelog entry


njohner (commented):
Ok I'm not sure I understand what we are doing here. You are adding an action that has an URL expression that points to a view that does not exist. It's also an action that is displayed everywhere, because it only looks at the view permission?

If it's just to know whether the action should be displayed in the new UI, then it should be a Folder Action, not a Folder button:
https://github.com/4teamwork/opengever.core/blob/master/opengever/core/profiles/default/actions.xml#L566
and it should not define an `url_expr`

njohner:
I really don't think this should be a `folder_button`. And yes I know you can't do a better permission check. My point was more, what is the use in that case to have the action come from the backend? I see two options:
- Make it a `folder_actions`. This category is there explicitly to tell the frontend which actions to show. So you'll have also have to check `folder_actions` in category when getting the folder actions in the frontend (https://github.com/4teamwork/gever-ui/blob/master/src/store/modules/actions.js#L73)
- Make it a pure frontend action. As this is a pure convenience action in the frontend, and that the backend does anyway not add any useful information as to where the action should get displayed, you could just as well make it a pure frontend action. Basically add it by hand in the list of available folder actions (https://github.com/4teamwork/gever-ui/blob/master/src/store/modules/actions.js#L73).

We can of course quickly discuss this if you want

tinagerber:
@njohner I moved the action to the `folder_actions` category. Can you rereview?

PR #6932, opened 2021-03-09
Include team tasks in globalindex

buchi:
When filtering by responsible using the `@globalindex` endpoint, also return tasks assigned to a team the responsible belongs to. Ensures the same behavior we have in the classic UI.

JIRA: https://4teamwork.atlassian.net/browse/CA-1454

## Checklist

- [x] Changelog entry


njohner (approved):
LGTM

PR #6934, opened 2021-03-10
Properly override inflator import step

buchi:
Redeclaring it in `configure.zcml` may result in a `ConfigurationConflictError` depending on the order the packages are
loaded.

I'm not sure why, but sometimes I get a different order of the package paths included in the instance script. Seems to happen particularly after running buildout for the first time. Running buildout again usually "fixes" the order.
We also had the issue with deployments on a server, so it seems not to depend on the local dev environment.

E.g. if the grokcore packages are added first to the path then it probably doesn't work:

```
sys.path[0:0] = [
  '/apps/eggs/grokcore.annotation-1.3-py2.7.egg',
  '/apps/eggs/grokcore.site-1.6.1-py2.7.egg',
  '/apps/eggs/grokcore.viewlet-1.11-py2.7.egg',
  '/apps/eggs/grokcore.component-2.5.1-py2.7.egg',
  '/apps/eggs/grokcore.security-1.6.3-py2.7.egg',
```

## Checklist

- [ ] Changelog entry

### DIFF
--- a/opengever/setup/creation/configure.zcml
+++ b/opengever/setup/creation/configure.zcml
@@ -15,23 +15,6 @@
       handler="opengever.setup.creation.setuphandler.unit_creation">
     </genericsetup:importStep>
 
-  <include package="ftw.inflator" />
-  <configure package="ftw.inflator.creation">
-    <genericsetup:importStep
-        name="ftw.inflator.content_creation"
-        title="ftw.inflator: content creation"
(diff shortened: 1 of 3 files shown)


njohner:
ha, I've done the same thing today: https://github.com/4teamwork/opengever.core/pull/6945

njohner (approved):
LGTM

PR #6935, opened 2021-03-11
Create initial version upon document checkin.

njohner:
In case the document was checked-out and in without having ever been modified, the initial version will be missing and needs to be created, as it will not be created by the file setter. In all other scenarios the initial version should already exist.

I don't see any way we could identify concerned documents and recreate the initial version though. It's also not that important, as the missing version has the same file as the first version. It also seems that we have wrong creators[ for the initial version of older documents](https://github.com/4teamwork/opengever.core/pull/4632), so we overwrite the `@history` endpoint to return the correct creator for the initial versions. This also ensures that the same creators are displayed in the new frontend as in the classic UI [without having to fix it in the frontend](https://github.com/4teamwork/gever-ui/pull/1654).

I did not document this change in the API changelog, because it's simply a fix of what is returned by the API, not a real change in the API.

For https://4teamwork.atlassian.net/browse/CA-1842

## Checklist
- [x] Changelog entry

- API change:
  - [ ] API Changelog entry (see [guide])
  - If breaking:
    - [ ] api-change label added

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -5,6 +5,7 @@ Changelog
 ---------------------
 
 - Implement a new content-type: opengever.workspace.meetingagendaitem. [elioschmutz]
+- Create initial version upon checkin. [njohner]
 - Add edit_items folder action. [tinagerber]
 - Update .gitignore of policytemplates for deployment on CentOS 8. [njohner]
 - Change p7m extension to eml (or extension configured in the registry) in mail download. [njohner]
--- a/opengever/api/configure.zcml
+++ b/opengever/api/configure.zcml
@@ -152,6 +152,13 @@
       permission="opengever.trash.UntrashContent"
       />
 
+  <plone:service
+      method="GET"
+      name="@history"
+      for="opengever.document.document.IDocumentSchema"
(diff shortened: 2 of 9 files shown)


phgross (approved):
👍 looks good

phgross:
@njohner is the JIRA Link the correct issue? 🤔

njohner:
Now it is sorry

njohner:
I've updated this PR to also handle the case where the creator of the initial version is wrong in the `@history` endpoint. Can you have another look?

phgross (commented):
@njohner changes are looking good 👍.  I just wonder if this might be a problem for the migrated documents? Not sure if the Creator is correctly set on migrated documents or only on the initial version. Maybe doublecheck.

njohner:
It matches what we are doing in the classic UI https://github.com/4teamwork/opengever.core/blob/master/opengever/document/browser/versions_tab.py#L104-L107

njohner:
I think the creator always gets set when creating an object (actually first time the object is indexed): https://github.com/plone/plone.dexterity/blob/master/plone/dexterity/content.py#L433-L454
So for migrated documents, it should be zopemaster for all of them

PR #6936, opened 2021-03-11
Allow to create tasks from document within task

tinagerber:
The action `new_task_from_document` should now also be available for documents within tasks.
To determine in the frontend where the new task should be created, the field `containing_subdossier_url` must also be serialised for documents.

Jira: https://4teamwork.atlassian.net/browse/NE-677

## Checklist

- [x] Changelog entry

- API change:
- New functionality:

### DIFF
--- a/opengever/document/fileactions.py
+++ b/opengever/document/fileactions.py
@@ -1,5 +1,3 @@
-from Acquisition import aq_inner
-from Acquisition import aq_parent
 from ftw.bumblebee.mimetypes import is_mimetype_supported
 from opengever.api.lock import can_unlock_obj
 from opengever.bumblebee import is_bumblebee_feature_enabled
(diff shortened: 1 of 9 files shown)


opengever/document/tests/test_base.py
```diff
@@ -39,11 +39,23 @@ def test_containing_subdossier_title_returns_subdossier_title(self):
             u'2016',
             self.subdocument.containing_subdossier_title())
 
+    def test_containing_subdossier_title_returns_subdossier_url(self):
```
njohner, line 42: subdossier_title -> subdossier_url

njohner (commented):
LGTM

opengever/document/fileactions.py
```diff
@@ -114,10 +112,10 @@ def is_untrash_document_available(self):
         return trasher.verify_may_untrash(raise_on_violations=False)
 
     def is_new_task_from_document_available(self):
-        parent = aq_parent(aq_inner(self.context))
-        is_inside_dossier = IDossierMarker.providedBy(parent)
-        may_add_task = api.user.has_permission('opengever.task: Add task', obj=parent)
-        return is_inside_dossier and may_add_task
+        dossier = self.context.get_parent_dossier()
```
njohner, line 115: Hum, this will now probably be available at too many places. For example documents inside a proposal? Documents in the template folder?

njohner: Ok as discussed, this should be all good:
- The template area does not provide IDossierMarker, so the action will not be available
- The whole meeting area is not implemented in the frontend so for now there is no risk for submitted proposals. I guess we will have to see that when we implement the meeting in the new frontend...

njohner (approved):

PR #6925, opened 2021-03-04
Add workspace meeting agenda item

elioschmutz:
This PR implements a new portal type: `opengever.workspace.meetingagendaitem`.

The type is addable in a workspace.meeting and does not provide a workflow.

It is deletable and orderable by any teamraum member.

The type is only used in the new UI.

Jira: https://4teamwork.atlassian.net/browse/CA-1734

## Checklist (Must have)

_Everything has to be done/checked. Checked but not present means the author deemed it unnecessary._

- [x] Changelog entry
- [x] Link to issue (Jira or GitHub) and backlink in issue (Jira)

## Checklist (if applicable)

_Only applicable should be left and checked._

- API change:
  - [x] Documentation is updated
  - [ ] API Changelog entry (see [guide](https://4teamwork.atlassian.net/wiki/spaces/4TEAM/pages/451248812/API+Changelog+Guidelines))
  - If breaking:
    - [ ] api-change label added
    - [ ] #delivery channel notified about breaking change
    - [ ] Scrum master is informed
- New functionality:
  - [ ] for `document` also works for `mail`
  - [ ] for `task` also works for `forwarding`
- Upgrade steps (changes in profile):
  - [ ] Make it deferrable if possible
  - [ ] Execute as much as possible conditionally
- DB-Schema migration
  - [ ] All changes on a model (columns, etc) are included in a DB-schema migration.
  - [ ] Constraint names are shorter than 30 characters (`Oracle`)
- [ ] Change could impact client installations, client policies need to be adapted
- New translations
  - [ ] All msg-strings are unicode
- Change in schema definition:
  - [ ] If `missing_value` is specified, then `default` has to be set to the same value

PR #6938, opened 2021-03-15
Change wrong i18n domain for workspace agenda item schema

elioschmutz:
Jira: https://4teamwork.atlassian.net/browse/CA-1734

This PR fixes a wrong i18n domain for the workspace agenda item schema. There is no need to update the translations because all the translation strings are already translated. The `opengever.document` translations have not been update in the meantime, thus, there is also no need to rebuild them.

A changelog is not required because it's a fix of a feature introduced in this PR: #6925 

## Before
Bildschirmfoto 2021-03-15 um 14 11 10

## After
Bildschirmfoto 2021-03-15 um 14 02 20

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/docs/schema-dumps/opengever.workspace.meetingagendaitem.schema.json
+++ b/docs/schema-dumps/opengever.workspace.meetingagendaitem.schema.json
@@ -18,13 +18,13 @@
         },
         "decision": {
             "type": "object",
-            "title": "Decision",
+            "title": "Beschluss",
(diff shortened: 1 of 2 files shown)


njohner (approved):
LGTM

PR #6939, opened 2021-03-15
Index agenda items in the workspace meeting searchable text.

njohner:
As we do not want to display `WorkspaceMeetingAgendaItem`s as separate objects (no detail view), but we want their content to be searchable, we need to add their content to the `SearchableText` of the `WorkspaceMeeting`. For this:
- We define a `SearchableTextExtender`
- We add event handlers to reindex the `WorkspaceMeeting` when an `WorkspaceMeetingAgendaItem` is added or modified.

For https://4teamwork.atlassian.net/browse/CA-1734

## Checklist
- [x] Changelog entry

### DIFF
--- a/opengever/workspace/indexers.py
+++ b/opengever/workspace/indexers.py
@@ -21,6 +21,8 @@ INDEXED_IN_MEETING_SEARCHABLE_TEXT = ['title', 'text', 'decision']
 
 def to_safe_utf8(obj, fieldname):
     value = getattr(obj, fieldname)
+    if value is None:
+        return
     if isinstance(value, RichTextValue):
         value = api.portal.get_tool(name='portal_transforms').convertTo(
             'text/plain', value.output, mimetype='text/html').getData()
@@ -57,4 +59,4 @@ class WorkspaceMeetingSearchableTextExtender(object):
                 [to_safe_utf8(agendaitem, fieldname) for fieldname
                  in INDEXED_IN_MEETING_SEARCHABLE_TEXT])
 
-        return ' '.join(searchable)
+        return ' '.join(filter(None, searchable))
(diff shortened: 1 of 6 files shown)


opengever/workspace/indexers.py
```diff
@@ -1,9 +1,59 @@
...
+
+
+@implementer(dexteritytextindexer.IDynamicTextIndexExtender)
+@adapter(IWorkspaceMeeting)
+class WorkspaceMeetingSearchableTextExtender(object):
+    """This extender includes the WorkspaceMeetingAgendaItems in the
+    WorkspaceMeeting SearchableText.
+
+    The searchable text gets updated when WorkspaceMeetingAgendaItems are added
+    or odified via event handles (see workspace_meeting_agendaitem_added, and
```
elioschmutz, line 36:
```suggestion
    or modified via event handles (see workspace_meeting_agendaitem_added, and
```

opengever/workspace/indexers.py
```diff
@@ -1,9 +1,59 @@
...
     return ''
+
+
+INDEXED_IN_MEETING_SEARCHABLE_TEXT = ['title', 'text', 'decision']
+
+
+def to_safe_utf8(obj, fieldname):
+    value = getattr(obj, fieldname)
+    if isinstance(value, RichTextValue):
+        value = value.output
```
elioschmutz, line 25: should we really index the plain html? or would it be better to extract the text? I think the html to text transform can already do that...

njohner: Good point thanks!

elioschmutz (approved):
👍

njohner:
@elioschmutz I've update according to your suggestions and fixed the issue we discussed with indexing the string "None" into the `SearchableText` when `text` or `decision` is None.

PR #6940, opened 2021-03-15
Remove workspace meeting agendaitem from search

njohner:
`WorkspaceMeetingAgendaItem` are only meant to be represented on the the `WorkspaceMeeting` and should therefore not be findable in searches. Instead their content gets indexed in the `SearchableText` of the `WorkspaceMeeting`. So here we make sure that they get filtered out of the different search views and endpoints.

The `@solrsearch` endpoint is implemented in such a way that by default it filters out the blacklisted portal_types (notably `WorkspaceMeetingAgendaItem`), but if a `portal_type` filter is specified, it will be respected, so that it is possible to search for `WorkspaceMeetingAgendaItem`.

As `WorkspaceMeetingAgendaItem` get introduced in this version, this is not an API change and I did not adapt the documentation. I don't think that we need to describe in the documentation that the blacklisted `portal_types` are retrievable.

For https://4teamwork.atlassian.net/browse/CA-1734

## Checklist

- [x] Changelog entry
- API change: 
  - If breaking:
    - [ ] api-change label added

### DIFF
--- a/opengever/base/tests/test_search.py
+++ b/opengever/base/tests/test_search.py
@@ -1,14 +1,11 @@
 from ftw.testbrowser import browsing
-from opengever.base.interfaces import ISearchSettings
 from opengever.testing import IntegrationTestCase
 from opengever.testing import obj2brain
 from opengever.testing import SolrIntegrationTestCase
 from plone import api
 from plone.app.contentlisting.interfaces import IContentListing
-from plone.registry.interfaces import IRegistry
 from plone.uuid.interfaces import IUUID
 from zope.component import getMultiAdapter
-from zope.component import getUtility
 
 
 class TestOpengeverSearch(IntegrationTestCase):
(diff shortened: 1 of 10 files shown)


opengever/workspace/tests/test_workspace_meeting_agenda_item.py
```diff
@@ -270,3 +272,19 @@ def test_workspace_meeting_gets_reindexed_when_agendaitem_is_added(self, browser
...
+
+    @browsing
+    def test_workspace_meeting_agendaitems_are_excluded_from_search(self, browser):
+        self.login(self.workspace_member, browser)
+        # builder of workspace_meeting_agendaitem does not reindex
+        # the workspace meeting
+        self.workspace_meeting.reindexObject()
+        self.commit_solr()
+
+        url = u'{}/@solrsearch?q=Lagebericht'.format(self.portal.absolute_url())
```
elioschmutz, line 284: same here...

opengever/base/tests/test_search.py
```diff
@@ -354,6 +383,16 @@ def test_solr_results_with_filter(self):
                 'hl.snippets': 3,
             })
 
+    def test_search_filters_portal_types_by_default(self):
+        self.assertEqual(
+            self.search.solr_filters(), self.default_portal_types_filter())
+
+    def test_search_respects_portal_types_filters_if_provided(self):
+        self.request.form.update({'portal_type': ['opengever.document.document']})
```
elioschmutz, line 391: shouldn't we test with an excluded type?

elioschmutz (approved):
👍 

I also tested it with the related frontend PR: https://github.com/4teamwork/gever-ui/pull/1671
it works as expected

opengever/base/tests/test_search.py
njohner: Ha, I was confused here too for a while, but this actually tests the `@@search` view, which always excludes the `opengever.workspace.agendaitem`. I think that is the correct behavior here as it's what also happens in the `@@search` if solr is disabled.

njohner:
@elioschmutz I've updated according to your comments. And thanks for the testing!

PR #6942, opened 2021-03-17
Remove Disqus from documentation

njohner:
Because Disqus has data protection issues, so we want to get rid of it. Moreover we don't really need it anymore now that we have the HubSpot chat in the documentation.

For https://4teamwork.atlassian.net/browse/CA-1506

## Checklist

- [x] Changelog entry


lukasgraf (approved):
LGTM 👍

PR #6943, opened 2021-03-18
Implement group_by_type for @solrsearch endpoint

tinagerber:
This PR implements a new parameter `group_by_type` for the `@solrsearch` endpoint. With this parameter it is possible to sort search results by type. I have chosen to sort by the listing names of `@listing` endpoint, as this gives equal score to the different dossier types and also does not differentiate between documents and mails.

Discussed in https://4teamwork.slack.com/archives/C01BHEGBVU1/p1616142371002900

Jira: [NE-681]

## Checklist

- [x] Changelog entry

- API change:
- New functionality:

[NE-681]: https://4teamwork.atlassian.net/browse/NE-681


opengever/api/solrsearch.py
```diff
@@ -103,7 +104,34 @@ def extract_sort(self, params, query):
...
+        max_score = len(group_by_type)
+        query_string = u''
+
+        for index, name in enumerate(group_by_type):
+            if name not in FILTERS:
+                raise BadRequest("group_by_type type '{}' is not allowed. "
+                                 "Allowed types are: {}.".format(name, ', '.join(FILTERS)))
+
+            interface = FILTERS[name][0].replace('object_provides:', '')
+            query_string += u'if(termfreq(object_provides,{}),{},'.format(interface,
```
njohner, line 132: So score in solr are normally between 0 and 1? So that this term will always trump over the rest of the score?

njohner (approved):
LGTM!

opengever/api/solrsearch.py
tinagerber: No, it will return a value between 0 and number of specified types for each content and sort by that. The sort parameter is used.

njohner: Ah yes I seen now. This gets returned as the sort parameter in the end. 👍

PR #6946, opened 2021-03-19
Meeting minutes pdf

buchi:
Provides the functionality to create a PDF with the meeting minutes from a workspace meeting.

Uses WeasyPrint for PDF generation from HTML, which is integrated as dockerized a web service.
The reason for not using the already available PDFLaTeX, is its poor support for handling HTML tables.

JIRA: https://4teamwork.atlassian.net/browse/CA-1743

## Checklist

- [x] Changelog entry

- New translations

### DIFF
--- a/opengever/core/profiles/default/actions.xml
+++ b/opengever/core/profiles/default/actions.xml
@@ -936,6 +936,18 @@
       <property name="visible">True</property>
     </object>
 
+    <object name="meeting_minutes_pdf" meta_type="CMF Action" i18n:domain="opengever.core">
+      <property name="title" i18n:translate="">Download meeting minutes PDF</property>
+      <property name="description" i18n:translate="" />
+      <property name="url_expr">string:@@meeting_minutes_pdf</property>
+      <property name="icon_expr" />
+      <property name="available_expr">python:here.restrictedTraverse('@@plone_interface_info').provides('opengever.workspace.interfaces.IWorkspaceMeeting')</property>
+      <property name="permissions">
+        <element value="View" />
+      </property>
+      <property name="visible">True</property>
+    </object>
+
   </object>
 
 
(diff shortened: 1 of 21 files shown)


opengever/workspace/browser/templates/meeting_minutes.pt
```diff
@@ -0,0 +1,111 @@
...
+    </h2>
+    <div tal:condition="nocall: agenda_item/text" tal:content="structure agenda_item/text"></div>
+    <div tal:condition="nocall: agenda_item/decision">
+      <h3><span i18n:translate="label_decision">Decision</span>:</h3>
+      <p tal:replace="structure agenda_item/decision"></p>
+    </div>
+  </tal:repeat>
+
+</body>
+</html>
```
njohner, line 111: Missing new line

opengever/core/profiles/default/actions.xml
```diff
@@ -936,6 +936,18 @@
       <property name="visible">True</property>
     </object>
 
+    <object name="meeting_minutes_pdf" meta_type="CMF Action" i18n:domain="opengever.core">
```
njohner, line 939: So is that action visible in the classic UI? Otherwise I'd move that to the `FOLDER ACTIONS` category, which we use for actions that are only to tell the Gever-UI whether to show it or not.

njohner (approved):
Looks good overall.
Could you add the `WEASYPRINT_URL` in the policy template for Teamraum deployments (I guess conditional on the meeting feature being active)?

opengever/workspace/browser/templates/meeting_minutes.pt
buchi: amended

opengever/core/profiles/default/actions.xml
buchi: Yes, it is and it also works in the classic UI. No need to hide it in the classic UI IMO.

buchi:
Added `WEASYPRINT_URL` to the policy template.

PR #6948, opened 2021-03-22
Deployment no longer fails if ldap is not in authentication plugins.

njohner:
With [this PR](https://github.com/4teamwork/opengever.core/pull/6903), we removed LDAP from the Authentication plugins in the policy templates, as it anyway gets deactivated during setup. Problem is that the deployment script fails while trying to deactivate the plugin.

For https://4teamwork.atlassian.net/browse/CA-1801

## Checklist

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2021.6.1 (unreleased)
 ---------------------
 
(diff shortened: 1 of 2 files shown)


lukasgraf (approved):
LGTM 👍

PR #6949, opened 2021-03-22
Meeting participants

elioschmutz:
This PR provides a new field `attendees` for a workspace meeting.

Jira: https://4teamwork.atlassian.net/browse/CA-1740

Bildschirmfoto 2021-03-26 um 11 29 49

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/workspace/utils.py
+++ b/opengever/workspace/utils.py
@@ -27,29 +27,40 @@ def is_within_workspace(context):
     return bool(filter(IWorkspace.providedBy, aq_chain(context)))
 
 
+def get_containing_workspace(context):
+    """ Returns the containing workspace of the current context
+    """
+    # Avoid circular imports
+    from opengever.workspace.interfaces import IWorkspace
+    workspace = filter(IWorkspace.providedBy, aq_chain(context))
+    return workspace[0] if workspace else None
+
+
 def get_workspace_user_ids(context, disregard_block=False):
     """ Walks up the Acquisition chain and collects all userids assigned
     to a role with the View permission.
     """
-    if not is_within_workspace(context):
+    containing_workspace = get_containing_workspace(context)
+    if not containing_workspace:
         return []
 
     users = set([])
-    allowed_roles_to_view = roles_of_permission(context, 'View')
+    allowed_roles_to_view = roles_of_permission(containing_workspace, 'View')
     portal = api.portal.get()
 
     def is_valid_userid(*args):
         user, roles, role_type, name = args
         return role_type == u'user' and set(roles) & set(allowed_roles_to_view)
 
-    for content in aq_chain(context):
+    for content in aq_chain(containing_workspace):
         if aq_base(content) == aq_base(portal):
             break
         userroles = portal.acl_users._getLocalRolesForDisplay(content)
         users = users.union(set(
             map(itemgetter(0),
                 filter(lambda args: is_valid_userid(*args), userroles))))
-        if getattr(aq_base(context), '__ac_local_roles_block__', None) and not disregard_block:
+        if getattr(aq_base(containing_workspace), '__ac_local_roles_block__', None) \
+                and not disregard_block:
             break
 
     return list(users)
--- a/opengever/workspace/vocabularies.py
+++ b/opengever/workspace/vocabularies.py
@@ -1,7 +1,11 @@
+from opengever.ogds.base.sources import ActualWorkspaceMembersSource
+from opengever.ogds.models.user import User
 from opengever.workspace.interfaces import IWorkspaceFolder
 from opengever.workspace.participation import PARTICIPATION_ROLES
 from opengever.workspace.participation.browser.manage_participants import ManageParticipants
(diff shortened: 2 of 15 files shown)


opengever/workspace/workspace_meeting.py
```diff
@@ -57,6 +57,17 @@ class IWorkspaceMeetingSchema(model.Schema):
         title=_(u'label_video_call_link', default=u'Video Call Link'),
         required=False)
 
+    form.widget('participants', KeywordFieldWidget, async=True)
+    participants = schema.List(
+        title=_(u"label_participants", default=u"Participants"),
+        value_type=schema.Choice(
+            source=ActualWorkspaceMembersSourceBinder(),
```
njohner: We will have the same issue here that users from groups assigned to a given workspace cannot be added as participants, see https://4teamwork.atlassian.net/browse/CA-1880. Can you add a comment in the corresponding Jira story?

opengever/workspace/utils.py
```diff
@@ -27,29 +27,40 @@ def is_within_workspace(context):
     return bool(filter(IWorkspace.providedBy, aq_chain(context)))
 
 
+def get_containing_workspace(context):
+    """ Returns the responsible workspace of the current context
```
njohner, line 31: hum, responsible seems out of place here. I'd say "containing workspace" is clearer.

opengever/workspace/utils.py
```diff
@@ -27,29 +27,40 @@ def is_within_workspace(context):
...
     to a role with the View permission.
     """
-    if not is_within_workspace(context):
+    containing_workspace = get_containing_workspace(context)
+    if not containing_workspace:
         return []
 
     users = set([])
-    allowed_roles_to_view = roles_of_permission(context, 'View')
+    allowed_roles_to_view = roles_of_permission(containing_workspace, 'View')
```
njohner, line 48: You can't do that, you are breaking the vocabulary here. This is really meant to walk up the acquisition chain from the current context.
- It is used in the `WorkspaceSharingView`. So imagine a workspace containing a folder and a subfolder, both with local roles blocked. An admin gives pemissions to some dude on the folder. A normal user should be allowed to give that dude permissions on the subfolder
- Note that the vocabulary name is confusing, it says `ActualWorkspaceMembers`, but that is not what the vocabulary actually is meant to return. I think for now it would be ok, but it will break in the future. For example, as soon as we allow to add ToDos or Meetings in a workspace folder. I want to assign a responsible to that meeting, I should only be allowed to choose him among users that have permissions in the folder, not users that have permissions on the workspace.

opengever/workspace/vocabularies.py
```diff
@@ -56,3 +60,22 @@ def __call__(self, context):
                                         participant.get('name'))))
 
         return SimpleVocabulary(terms)
+
+
+@implementer(IVocabularyFactory)
+class ActualWorkspaceMembersVocabulary(object):
```
njohner, line 66: Hum, so we make an enumerable vocabulary here. Don't forget we allow adding permissions to whole groups on a  Teamraum. I don't see it as impossible to have a teamraum accessible by all users of a given organisation.
Can't we implement what we need with the query source?

njohner (commented):
I'm a bit unsure overall here. Maybe someone else wants to give input?

njohner (commented):
Sorry forgot another comment I had. Please call the field `attendees` or something alike. We have enough participants and participations in the code basis...

opengever/workspace/workspace_meeting.py
elioschmutz: 👍 I updated the Jira-Story to use the newly introduced vocabulary in this PR

opengever/workspace/utils.py
elioschmutz: Summary of the phonecall with @njohner:

- The sharing view will only be used for the old ui. It should not be possible to add users which are not participating to the current workspace. https://github.com/4teamwork/opengever.core/issues/6041#issuecomment-560485965. The workspace folders are using a different vocabulary, the `PossibleWorkspaceFolderParticipantsVocabulary`. That's not affected by my changes and still works as expected. The changed function here: `get_workspace_user_ids` is now doing what it should do, it returns all userids which are participating on the current workspace.
- Currently it's not planned to add meetings or todos on subfolders. So this should be ok

opengever/workspace/vocabularies.py
elioschmutz: Just found out, that the vocabulary is already batched 😄  

Bildschirmfoto 2021-03-24 um 17 20 23

So we won't get any performance issues here. But we need a vocabulary to be able to display initial items.

njohner (approved):
LGTM

elioschmutz:
@buchi (@njohner ) I updated the PR to also include the new field in the PDF. Could you take a look at the last commit pls?

lukasgraf (approved):
LGTM 👍

PR #6950, opened 2021-03-23
Fix oc_checkout endpoint

buchi:
Make it work with shadow documents that don't have a content type stored in annotations. This is the case for Docugate as we don't know yet the content type.

JIRA: https://4teamwork.atlassian.net/browse/CA-1411

## Checklist

- [x] Changelog entry


lukasgraf (approved):
LGTM 👍

PR #6951, opened 2021-03-24
Introduce a new field dossier_type and customproperty slots for dossiers.

phgross:
This PR introduces a customproperty field for dossier, including a new field `dossier_type`, similar to the `document_type` field for documents and mails. I made the following decisions:
- For the dossier_type values I introduced new vdexvocabulary, which can be customized by a GEVER policy. By default there is only one value, which is also hidden by the registry entry.                                                                                                                                                                 
- If there are no values for the field available, the field is hidden in the add and edit forms.
- The customproperty field provides a default slot and slots for each dossier_type.

I also had to fix the vocabulary_url for unnamed vocabularies in the @schema representation, for example dossier_type or document_type.

https://4teamwork.atlassian.net/browse/CA-1493

## Checklist

- [x] Changelog entry
- API change:
- Upgrade steps (changes in profile)

### DIFF
--- a/docs/public/dev-manual/api/propertysheets.rst
+++ b/docs/public/dev-manual/api/propertysheets.rst
@@ -83,8 +83,9 @@ Einzelne Felder werden in folgendem Format erwartet:
 
 Die für das Assigmnet verwendeten Assignment-Slots müssen aus dem Vokabular
 ``opengever.propertysheets.PropertySheetAssignmentsVocabulary`` stammen. Ein
-Spezifalfall ist dabei der Default-Slot ``IDocument.default``, welcher
-unabhängig vom Dokumenttyp Feld immer dargestellt wird.
+Spezialfall ist dabei der Default-Slot ``IDocument.default`` bzw.
+``IDossier.default``, welcher unabhängig vom Dokument- oder Dossiertyp Feld
+immer dargestellt wird.
 
 Zudem müssen Assignments
 eindeutig sein, mehrere Property Sheets dem gleichen Assignment-Slot zuzuweisen
@@ -147,10 +148,11 @@ ist im Moment nicht unterstützt.
 Serialisierung/Deserialisierung von Custom Properties
 -----------------------------------------------------
 
-Im Moment sind Custom Properties auf Dokumenten und Mails unterstützt. Die
-Auswahl des zu validerenden Property Sheets basiert auf dem Wert des Feldes
-`document_type`. Ausnahme ist dabei der Default-Slot ``IDocument.default``
-welcher unabhängig des Feldwertes von ``document_type`` immer dargestellt wird.
+Im Moment sind Custom Properties auf Dokumenten, Mails und Dossiers unterstützt.
+Die Auswahl des zu validierenden Property Sheets basiert auf dem Wert des Feldes
+`document_type` bzw. `dossier_type`. Ausnahme ist dabei der Default-Slot
+``IDocument.default`` bzw. ``IDossier.default`` welcher unabhängig des Typen
+Feldwertes immer dargestellt wird.
 Ist für den Assignment-Slot
 ``IDocumentMetadata.document_type.<document_type_value>`` ein Property Sheet
 registriert, so werden Feldwerte dieses Property Sheets validiert. Hat das
--- a/opengever/dossier/behaviors/customproperties.py
+++ b/opengever/dossier/behaviors/customproperties.py
@@ -1,0 +1,29 @@
+from opengever.dossier import _
+from opengever.propertysheets.assignment import get_dossier_assignment_slots
+from opengever.propertysheets.assignment import DOSSIER_DEFAULT_ASSIGNMENT_SLOT
+from opengever.propertysheets.field import PropertySheetField
+from plone.autoform.interfaces import IFormFieldProvider
+from plone.supermodel import model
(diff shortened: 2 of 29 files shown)


opengever/dossier/browser/forms.py
```diff
@@ -28,6 +29,14 @@
 import base64
 
 
+def has_active_dossier_types(context):
+    factory = wrap_vocabulary(
+        'opengever.dossier.dossier_types',
+        hidden_terms_from_registry='opengever.dossier.interfaces.IDossierType.dossier_types')
+    terms = [aa for aa in factory(context)]
```
njohner, line 36: aa -> term?

opengever/dossier/interfaces.py
```diff
@@ -250,3 +250,16 @@ class IDossierTasksPdfMarker(Interface):
 
 class IDossierJournalPdfMarker(Interface):
     """Deprecated marker Interface for dossier journal document."""
+
+
+class IDossierType(Interface):
+    """plone.app.registry schema for the dossier types setting."""
+
+    dossier_types = schema.List(
```
njohner, line 258: So this is not actually a list of dossier types, but of hidden terms right? Can we improve the naming here, maybe simply `hidden_terms` or `hidden_dossier_types`?

opengever/dossier/browser/forms.py
```diff
@@ -28,6 +29,14 @@
...
+    return len(terms) > 0
```
njohner, line 37: I guess `len(factory(context))` does not work? or `len(list(factory(context)))`

opengever/dossier/behaviors/customproperties.py
```diff
@@ -0,0 +1,29 @@
...
+from plone.autoform.interfaces import IFormFieldProvider
+from plone.supermodel import model
+from zope.interface import alsoProvides
+
+
+class IDossierCustomProperties(model.Schema):
+
+    custom_properties = PropertySheetField(
+        request_key='form.widgets.IDossier.dossier_type',
+        attribute_name='dossier_type',
```
njohner, line 14: So we needed `dossier_type` because we do not support `None` here?

opengever/propertysheets/tests/test_dossier_custom_properties.py
```diff
@@ -0,0 +1,215 @@
...
+        }
+        browser.open(
+            self.dossier,
+            method="PATCH",
+            data=json.dumps(good_data),
+            headers=self.api_headers,
+        )
+        self.assertEqual(
+            good_data["custom_properties"],
+            IDossierCustomProperties(self.dossier).custom_properties,
```
njohner, line 113: hum does that actually test that required field is validated? shouldn't you try to PATCH with no data for the mandatory field, which would raise, and then maybe fill the field correctly and show it passes?

docs/public/dev-manual/api/propertysheets.rst
```diff
@@ -83,8 +83,9 @@ Einzelne Felder werden in folgendem Format erwartet:
 
 Die für das Assigmnet verwendeten Assignment-Slots müssen aus dem Vokabular
 ``opengever.propertysheets.PropertySheetAssignmentsVocabulary`` stammen. Ein
-Spezifalfall ist dabei der Default-Slot ``IDocument.default``, welcher
-unabhängig vom Dokumenttyp Feld immer dargestellt wird.
+Spezifalfall ist dabei der Default-Slot ``IDocument.default`` bzw.
```
njohner, line 86: Spezifalfall 🤣

docs/public/dev-manual/api/propertysheets.rst
```diff
@@ -147,10 +148,11 @@ ist im Moment nicht unterstützt.
 Serialisierung/Deserialisierung von Custom Properties
 -----------------------------------------------------
 
-Im Moment sind Custom Properties auf Dokumenten und Mails unterstützt. Die
-Auswahl des zu validerenden Property Sheets basiert auf dem Wert des Feldes
-`document_type`. Ausnahme ist dabei der Default-Slot ``IDocument.default``
-welcher unabhängig des Feldwertes von ``document_type`` immer dargestellt wird.
+Im Moment sind Custom Properties auf Dokumenten, Mails und Dossiers unterstützt.
+Die Auswahl des zu validerenden Property Sheets basiert auf dem Wert des Feldes
```
njohner, line 152: validerenden -> validierenden

njohner (approved):
last commit message: wron -> wrong

Otherwise LGTM!

opengever/dossier/browser/forms.py
phgross: Yes, the elephantvocabulary does not consider the hidden terms when querying for the length, i added a comment in the code.

opengever/dossier/behaviors/customproperties.py
phgross: Yes, so we can use the existing fields without any adjustments. Story CA-1493 does not require a dossier type field, but it will be added in later stories (see https://4teamwork.atlassian.net/browse/ROAD-1420).

Therefore it makes no sense to adjust the propertysheetfield so that it can be used without a "type field", if the dossier_type field will be added two-three sprints later anyway.

lukasgraf (approved):
LGTM 👍 - just a small copy&paste fail.

PR #6952, opened 2021-03-24
Fix catalog inconsistency when creating a dossier from a dossiertemplate. 

njohner:
When creating a dossier from a dossier template, the children of the main dossier get created with `DeactivatedCatalogIndexing`. This does not work as expected as we Monkey patch the same indexing methods than `collective.indexing`. Because our patch gets applied first, the patch from `collective.indexing` will still be adding indexing operations to the indexing queue, and our patched methods will only be used when the queue is processed. This is why we need to make sure to process the queue when entering this context manager so that items already in the queue get processed correctly, and before exiting the context manager so that indexing of queued items really gets skipped.

Moreover we need to make sure to reindex the main dossier after creation of the child objects (because adding children modifies the container).

This issue was discovered while looking at https://4teamwork.atlassian.net/browse/CA-1361

## Checklist

- [x] Changelog entry

### DIFF
--- a/opengever/base/monkey/patches/cmf_catalog_aware.py
+++ b/opengever/base/monkey/patches/cmf_catalog_aware.py
@@ -4,47 +4,38 @@ from opengever.base.role_assignments import RoleAssignmentManager
 from plone import api
 from Products.CMFCore.utils import _getAuthenticatedUser
 from zope.globalrequest import getRequest
-from zope.interface import alsoProvides
-from zope.interface import Interface
-from zope.interface import noLongerProvides
-from zope.i18n import translate
-
-
-class IDisableCatalogIndexing(Interface):
-    """Marker-interface to disable the catalog
-    indexing functions.
-
-    If this interface is provided by the request, all the
-    catalog-index-methods will be disabled.
-
-    Use the DeactivatedCatalogIndexing contextmanager to
-    get in use of this functionality.
-    """
 
 
 class DeactivatedCatalogIndexing(object):
     """Contextmanager: Deactivates catalog-indexing
     """
     def __enter__(self):
-        alsoProvides(getRequest(), IDisableCatalogIndexing)
+        PatchCMFCatalogAware()()
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        noLongerProvides(getRequest(), IDisableCatalogIndexing)
+        PatchCMFCatalogAware().unpatch()
+
+
+class CatalogAlreadyPatched(Exception):
+    """Will be raised if we try to patch the catalog indexing methods
+    more than once.
+    """
 
 
 class PatchCMFCatalogAware(MonkeyPatch):
     """Patch the Products.CMFCore.CMFCatalogAware indexObject, reindexObject
     and unindexObject methods.
 
-    This patch is deactivated by default and can be activated through
-    the DeactivatedCatalogIndexing context manager:
+    This patch is deactivated is not applied by default and can be activated
+    through the DeactivatedCatalogIndexing context manager:
 
     >>> with DeactivatedCatalogIndexing():
     ...     object.reindexObject  # Does nothing
     ...     object.unindexObject  # Does nothing
     ...     object.indexObject  # Does nothing
 
-    If the patch is activated, it skips the catalog index-methods.
+    If the patch is activated, it skips the catalog index-methods. The patch
+    gets removed when exiting the context manager.
 
     What's the motivation behind this patch?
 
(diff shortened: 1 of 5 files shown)


lukasgraf (commented):
How does this affect bundle import? The [bundle import pipeline also uses this context manager](https://github.com/4teamwork/opengever.core/blob/ffe96663f855bcf81f7359b8817827f4bf9918a1/opengever/bundle/sections/disabled_indexing.py#L8-L24).

To be honest, I'm really not a fan of triggering the `processQueue()` upon entering / leaving the context manager. That is an operation that could, at least potentially, take several hours (e.g. during an upgrade) that is just triggered when using a benign sounding context manager, probably wrapped around a very specific piece of code.

Did you explore the option of changing the monkey patch order at all? Adding a ZCML dependency to `collective.indexing` in the right place (before our patch) should result in the patch order being reversed, and as far as I understand your comment, that should also fix the issue. But again, would need to be investigaged with bundle import because that also [might unpatch the `collective.indexing` patches](https://github.com/4teamwork/opengever.core/blob/930fe22b933c8527a711d76cc20c9bedcd08329d/opengever/bundle/console.py#L82-L84).

njohner:
I tried changing the Monkey patch order, by importing `collective.indexing` in our patch. This did change the order, but somehow probably led to an infinite loop when creating a dossier from a template (I did not check what was really happening, just that Gever was hanging for ever when creating a dossier from a template). That is why I chose that approach. We can of course try to find a better solution.

As for the bundle import, I saw it was also using the context manager. As for whether this was a problem there, I am not sure. I did not want to invest too much time, as this was only half part of the actual issue I was supposed to fix.

A safer option could be to patch the Queue itself (https://github.com/plone/collective.indexing/blob/master/src/collective/indexing/queue.py#L95-L105). We can discuss that when I get back to work.

njohner:
Ok I've given it some more thoughts and I think the current approach with adding an interface in the context manager cannot work:
1. If our patch is applied first, the collective.indexing patch second, then indexing an object will still add to the queue, and our patch's effect will be applied when the queue is processed, so that cannot work
2. If collective.indexing patch is applied first, then ours: in that case things will work as expected. If `IDisableCatalogIndexing` is provided by the request, indexing operations will do nothing (objects will not land in the queue), but if the queue is processed, the original indexing methods will be applied so the objects in the queue will get correctly indexed.

Problem with solution 2. is that the call to `unpatch_collective_indexing` will remove both patches, so that our `DeactivatedCatalogIndexing` context manager will not work anymore. So I see two solutions:
1. Modify the context manager so that it patches and unpatches the indexing methods instead of adding the `IDisableCatalogIndexing` to the request.
2. write our own `unpatch` method to correctly remove the `collective.indexing` patches, or reapply our own patch after the `unpatch_collective_indexing` call.

I'd find solution 1 cleaner. It also has the advantage that we do not have to check for disabled indexing all the time.

lukasgraf:
@njohner I agree. After some thought, I also think solution 1 is better (patching the indexing methods instead of using the request layer).

opengever/base/monkey/patches/cmf_catalog_aware.py
```diff
@@ -58,38 +49,44 @@ class PatchCMFCatalogAware(MonkeyPatch):
     and do it manually at the end of your tasks.
     """
 
+    original_indexing_methods = {}
```
deiferni, line 52: I'm having a very bad feeling about this class attribute. It will be shared between threads. Please redesign this a bit, so that we use the same instances of `PatchCMFCatalogAware` for patching/unpatching withing one request/thread.

1) move creation of the mapping into `__init__` or create lazily in `__call__`

opengever/base/monkey/patches/cmf_catalog_aware.py
```diff
@@ -4,47 +4,38 @@
...
-    get in use of this functionality.
-    """
 
 
 class DeactivatedCatalogIndexing(object):
     """Contextmanager: Deactivates catalog-indexing
     """
     def __enter__(self):
-        alsoProvides(getRequest(), IDisableCatalogIndexing)
+        PatchCMFCatalogAware()()
```
deiferni, line 13: don't create multiple instances of the monkey patch, instead do 

2. `self.catalogpatch = PatchCMFCatalogAware()` (and call it separately)

opengever/base/monkey/patches/cmf_catalog_aware.py
```diff
@@ -4,47 +4,38 @@
...
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        noLongerProvides(getRequest(), IDisableCatalogIndexing)
+        PatchCMFCatalogAware().unpatch()
```
deiferni, line 16: then use the same instance with the thread-safe mapping to unpatch

3. `self.catalogpatch.unpatch()`

deiferni (changes requested):
The concept looks good, i think there is a serious flaw regarding thread safety in the implementation though. Don't hesitate to ☎️  if you want to discuss this!

Apart from that LGTM 👍 .

elioschmutz:
Besprochen mit @lukasgraf : PR ist veraltet.

PR #6953, opened 2021-03-25
Reindex documents with missing searchable text.

njohner:
Indexing of the SearchableText of documents in solr would sometimes fail because the blob had been updated by the doc properties updater in the meantime. This leads to missing SearchableText. Here we find and reindex such documents.

This upgrade step requires a bump of ftw.solr (https://github.com/4teamwork/ftw.solr/pull/179)

For https://4teamwork.atlassian.net/browse/CA-1361

## Checklist

- [x] Changelog entry


lukasgraf (commented):
Edit: _Sorry, commented on the wrong PR._

opengever/core/upgrades/20210325095837_reindex_missing_searchable_text/upgrade.py
```diff
@@ -0,0 +1,33 @@
...
+            fl=["UID", "SearchableText"],
+            rows=1000000)
+        for doc in resp.docs:
+            if "SearchableText" in doc:
+                continue
+
+            obj = unrestrictedUuidToObject(doc['UID'])
+            if not obj:
+                continue
+            print "reindexing", obj
```
lukasgraf, line 32: This should be a log message instead of `print`.

lukasgraf (approved):
LGTM 👍 , just a stray `print`

opengever/core/upgrades/20210325095837_reindex_missing_searchable_text/upgrade.py
njohner: Right thanks. I added some correct logging instead

PR #6957, opened 2021-03-29
When delegating tasks via API, informed_principals can be set.

tinagerber:
It should now be possible to notify additional principals when delegating a task, just like when creating a task.

Jira: https://4teamwork.atlassian.net/browse/NE-671

## Checklist

- [x] Changelog entry

- API change:
- New functionality:

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2021.7.0 (unreleased)
 ---------------------
 
+- When delegating tasks via API, informed_principals can be set. [tinagerber]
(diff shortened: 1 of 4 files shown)


elioschmutz (approved):
👍

elioschmutz (approved):
👍 

Is there a reason why the field is written in the past? Sounds like we use this field for some other reasons, but can't find anything....

tinagerber:
The field is used only to inform additional users when a task is created. Maybe it is written in the past, because on the field you can see which users were informed when the task was created. 🤷‍♀️

PR #6959, opened 2021-03-30
Index custom properties in Solr as dynamic fields

buchi:
This allows filtering and sorting by custom properties in Solr. However there's no UI integration so far.

Properties are indexed using the corresponding Solr type (string, int, boolean) in dynamic fields.
Text (multiline) properties are not indexed as it doesn't make sense to filter/sort on multiline text.
Only properties of the active and the default assignment are indexed.

The name patterns for dynamic fields are:
- `*_custom_field_string`
- `*_custom_field_int`
- `*_custom_field_boolean`

E.g. a custom property with id `supplier` and type `TextLine` would be indexed in the dynamic field `supplier_custom_field_string`. The assignment is not part of the Solr field name, which could potentially lead to conflicts in case of having the same id in the default and the active assignment. However this can be neglected as it doesn't make sense to do this and it would be difficult to represent that in the UI anyway.

JIRA: https://4teamwork.atlassian.net/browse/CA-1285

## Checklist

- [x] Changelog entry

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -6,6 +6,7 @@ Changelog
 
 - Add dossier_type field for dossiertemplates. [phgross]
 - Index custom properties in searchable text. [buchi]
+- Index custom properties in Solr dynamic fields. [buchi]
 
 
 2021.7.0 (2021-04-01)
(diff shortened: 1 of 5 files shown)


phgross (approved):
Nice 🎉 

_doublecheck the changelog slot before merge_

PR #6963, opened 2021-04-06
Properly reset the responsible watcher if a user accepts a task assigned to a team.

elioschmutz:
This PR properly handles watchers for team tasks accepted by a user.

With this change, if a user accepts a task assigned to a group, the group will be removed as a watcher of the resource and the user will be added. 

Jira: https://4teamwork.atlassian.net/browse/CA-1953

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/task/tests/test_activities.py
+++ b/opengever/task/tests/test_activities.py
@@ -484,6 +484,41 @@ class TestTaskActivites(FunctionalTestCase):
         self.assertEquals([], warning_messages())
 
 
+class TestTaskActivityIntegration(IntegrationTestCase):
+
+    features = ('activity', )
+
+    @browsing
+    def test_removes_old_group_responsibles_from_watchers_list_if_accepting_team_task(self, browser):
+        self.login(self.manager)
+        team_task = create(Builder('task')
+                           .having(responsible_client='fa',
+                                   responsible='team:1',
+                                   issuer=self.meeting_user.id,
+                                   task_type='correction')
+                           .within(self.dossier))
+
+        resource = notification_center().fetch_resource(team_task)
+        create_session().refresh(resource)
+        subscriptions = resource.subscriptions
+
+        self.assertItemsEqual(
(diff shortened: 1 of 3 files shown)


opengever/task/tests/test_activities.py
```diff
@@ -484,6 +485,43 @@ def test_delegate_activity(self, browser):
         self.assertEquals([], warning_messages())
 
 
+class TestTaskActivityIntegration(IntegrationTestCase):
+
+    features = ('activity', )
+
+    @browsing
+    def test_removes_old_group_responsibles_from_watchers_list_if_accepting_team_task(self, browser):
+        ogds_group = Group.query.get('projekt_a')
```
phgross, line 494: You query the group by its id to only get the id of the group --> ✂️

opengever/task/transition.py
```diff
@@ -181,15 +181,22 @@ def after_transition_hook(self, transition, disable_sync, transition_params):
...
-    def reassign_team_task(self, response):
+    def reassign_reassign_task(self, response):
         old_responsible = ITask(self.context).responsible
-        ITask(self.context).responsible = api.user.get_current().getId()
+        current_user_id = api.user.get_current().getId()
+        if (current_user_id == old_responsible):
+            return
+
+        center = base_notification_center()
+        import pdb; pdb.set_trace()
```
phgross, line 195: ✂️  😉

opengever/task/transition.py
```diff
@@ -181,15 +181,22 @@ def after_transition_hook(self, transition, disable_sync, transition_params):
         response = add_simple_response(
             self.context, transition=transition, text=transition_params.get('text'))
 
-        if self.context.is_team_task:
```
phgross, line 184: The reassign_team_task is no longer conditional, but I think it should be.

phgross (changes requested):

elioschmutz:
@phgross could you take another look?

phgross (approved):
👍

PR #6965, opened 2021-04-07
Fix searching for group descriptions with umlauts in search terms.

phgross:
Currently when searching for a group, via the @sharing endpoint and the search_terms contains umlaut, the search for the additional group_title attribute (`description`) failed silently with a UnicodeWarning. Because combining the attribute and the search_term raised an UnicodeDecodeError.

Fixes https://4teamwork.atlassian.net/browse/CA-2012

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-2012.bugfix
+++ b/changes/CA-2012.bugfix
@@ -1,0 +1,1 @@
+Fix searching for group descriptions with umlauts in search terms.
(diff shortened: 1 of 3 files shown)


njohner (approved):
In commit title "umaluts" should be "umlauts".
Are there corresponding sentry issues?

Apart from that, LGTM

PR #6966, opened 2021-04-08
Add attendees solr index for workspace meetings.

tinagerber:
With this PR, attendees for workspace meetings are indexed in Solr. This is necessary to display a meetings card on the teamraum dashboard.
Based on: https://github.com/4teamwork/opengever.core/pull/6858

Jira: [NE-678]

## Checklist

- [x] Changelog entry

- API change:
 
  [NE-678]: https://4teamwork.atlassian.net/browse/NE-678

### DIFF
--- a/solr-conf/managed-schema
+++ b/solr-conf/managed-schema
@@ -128,6 +128,7 @@
     <field name="is_folderish" type="boolean" indexed="true" stored="true"/>
 
     <!-- GEVER specific fields -->
+    <field name="attendees" type="string" indexed="true" stored="false" multiValued="true"/>
     <field name="blocked_local_roles" type="boolean" indexed="true" stored="false" />
     <field name="bumblebee_checksum" type="string" indexed="false" stored="false" />
     <field name="changed" type="pdate" indexed="true" stored="false" />
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -4,6 +4,7 @@ Changelog
 2021.7.1 (unreleased)
 ---------------------
 
+- Add attendees solr index for workspace meetings. [tinagerber]
 - Fix broken task template responsibles [elioschmutz]
 - Provide dossier_reference_number mergefield value also for ad-hoc proposals. [phgross]
 - Fix plone site deletion by skipping certain event handlers. [njohner]
--- a/opengever/core/upgrades/20210409160907_index_attendees_in_solr/upgrade.py
+++ b/opengever/core/upgrades/20210409160907_index_attendees_in_solr/upgrade.py
@@ -1,0 +1,13 @@
+from ftw.upgrade import UpgradeStep
+from opengever.workspace.interfaces import IWorkspaceMeeting
+
+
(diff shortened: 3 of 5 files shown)


opengever/core/upgrades/20210407174002_reindex_meeting_participations/upgrade.py
```diff
@@ -0,0 +1,13 @@
...
+
+class ReindexMeetingParticipations(UpgradeStep):
+    """Reindex meeting participations.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
+        query = {'object_provides': IWorkspaceMeeting.__identifier__}
+        for meeting in self.objects(query, 'Reindex meeting participations'):
+            meeting.reindexObject(idxs=['participations'])
```
njohner, line 13: `participations` is a solr index only, so you are actually reindexing the full meetings here. Add `UID` or another cheap index in that list to avoid that.

njohner (commented):
I'm not a big fan of this, but maybe it's just me. We are reusing the already existing `participations` index for meetings, but it will basically use a different format for meetings than for dossiers. `participations` also has a special treatment in the filters of the listing endpoint (https://github.com/4teamwork/opengever.core/blob/master/opengever/api/listing.py#L178-L199) because it's basically a combined index (participants and roles), whereas for meetings it's really a list of participants. Also if you look at the documentation of the API, `participations` is `Beteiligungen`, which is not the same as a "Teilnehmer" for meetings. I think I'd rather have a separate index for that. Also I think that solr only stores the set of indexes that actually have a value for a given document, so reusing already existing indexes does not provide any advantage, but maybe @buchi can contradict me here?

Maybe someone else can give an opinion?

buchi:
I would not reuse the existing `participations` field as there's no advantage to this. So I would go with a new field `attendees` to distinguish it clearly.

It would only make sense if we wanted to query participations for dossiers and meetings together, but as the indexed values are different, this is not possible.

Regarding performance or storage, a separate field index is rather better. (When you query for attendees, you query a smaller index.)

tinagerber:
@njohner I have now defined a new solr field `attendees`. Can you rereview?

solr-conf/managed-schema
```diff
@@ -128,6 +128,7 @@
     <field name="is_folderish" type="boolean" indexed="true" stored="true"/>
 
     <!-- GEVER specific fields -->
+    <field name="attendees" type="string" indexed="true" stored="false" multiValued="true"/>
```
njohner, line 131: So it is not plan to list the attendees in the card right? The field will only be queriable but not retrievable here (which I think is fine, just making sure)...

njohner (approved):
LGTM. Maybe quickly ask Erich that we do not plan to display the attendees in the listing of meetings.

solr-conf/managed-schema
tinagerber: No, it must only be possible to query the meetings of the current user.

PR #6973, opened 2021-04-15
API: Reject years before 1900 for date and datetime fields.

lukasgraf:
Because the Python `strftime` methods can't represent dates earlier than 1900, we reject them in the API by using custom field serializers.

I didn't override the `PersistentDatetimeFieldDeserializer` that we also have (in the same file), because that only gets used for system-generated timestamps, like entries in task or proposal histories. These should never get set to anything earlier than 1900 ;-) 

Jira: https://4teamwork.atlassian.net/browse/CA-1882

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [x] api-change label added

### DIFF
--- a/opengever/api/tests/test_deserializer.py
+++ b/opengever/api/tests/test_deserializer.py
@@ -1,8 +1,9 @@
+from base64 import b64encode
 from ftw.testbrowser import browsing
 from opengever.testing import IntegrationTestCase
-import os
-from base64 import b64encode
 from opengever.testing.assets import path_to_asset
+import json
+import os
 
 
 def _base64_str(s):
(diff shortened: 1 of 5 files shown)


opengever/api/tests/test_deserializer.py
```diff
@@ -223,3 +224,60 @@ def test_tus_can_upload_p7m_mail(self, browser):
...
+        self.assertEqual('BadRequest', response.json['type'])
+        self.assertIn("'field': 'start'", response.json['message'])
+        self.assertIn("'year=1700 is invalid. Year must be >= 1900.'",
+                      response.json['message'])
+
+        response = browser.open(
+            self.workspace_meeting,
+            method='PATCH',
+            headers=self.api_headers,
+            data=json.dumps({'document_date': '1900-01-01 13:45:00'})
```
njohner, line 280: 'document_date' -> 'start'

njohner (approved):
I wonder whether we want an upgrade step? Probably rather complicated to query all fields for invalid dates though, and not really necessary...

Otherwise, apart from comment, LGTM.

opengever/api/tests/test_deserializer.py
lukasgraf: Good catch, thanks! Fixed.

lukasgraf:
Even harder than finding and querying all those fields would be the question: How do we fix it? What date would we actually set the value to? I honestly don't think this can be addressed by an automated upgrade step, but instead would need to be looked at and fixed on a case-by-case basis, when we see it in Sentry.

njohner:
True 😅 . Of course we can't fix it...

PR #6975, opened 2021-04-16
Open Office Online in new tab

buchi:
Microsoft has removed the close action from Word and has announced to do the same for Excel and PowerPoint in the future. The user now needs to close the tab to finish editing. Thus we open Office Online in a new tab.

JIRA: https://4teamwork.atlassian.net/browse/CA-1940

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/6975.other
+++ b/changes/6975.other
@@ -1,0 +1,1 @@
+Open Office Online in new tab. [buchi]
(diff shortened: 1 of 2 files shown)


njohner (approved):
LGTM

PR #6980, opened 2021-04-20
No longer export meeting.json in meeting zip export of original files.

deiferni:
Grimlock does no longer support uploading documents which are not a PDF, but Users nevertheless sometimes upload the ZIP with original files to Grimlock.

Instead of adding validation to Grimlock which may be expensive, we no longer add the `meetings.json` metadata file to the zip of originals. This will prevent such ZIPs being processed by Grimlock. This JSON file is only used for Grimlock import, but as the import for originals is no longer possible it does not make sense to include the metadata in the zip.

I had to move the tests around a bit:
- An exhaustive functional test which asserts the whole structure of the `meetings.json` file in the downloaded zip, i moved it to the PDF ZIP oriented tests
- I dropped some of the tests which only test  a subset of aforementioned test
- Also I extracted tests that only assert the json file structure without a browser download

Jira: https://4teamwork.atlassian.net/browse/CA-1732

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-1732.other
+++ b/changes/CA-1732.other
@@ -1,0 +1,1 @@
+No longer include `meetings.json` metadata file in ZIP download of original files. [deiferni]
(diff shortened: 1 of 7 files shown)


njohner (approved):
LGTM

PR #6981, opened 2021-04-21
Fix `@history` endpoint when no history exists.

deiferni:
With this PR we prevent an exception in the `@history` endpoint for users who only have the `Reader` role on context they access.

We need to workaround the following problem:
- plone.restapi relies on history viewlet https://github.com/plone/plone.restapi/blob/d78a8411d1166ee785ee71292a540f39fc2cf44c/src/plone/restapi/services/history/get.py#L42 and iterates over its result https://github.com/plone/plone.restapi/blob/d78a8411d1166ee785ee71292a540f39fc2cf44c/src/plone/restapi/services/history/get.py#L54 
- however full history returns `None` when the history can't be retrieved or is empty https://github.com/plone/plone.app.layout/blob/0ba390844d7c4f7dc88fe1688e80876eea58be66/plone/app/layout/viewlets/content.py#L481 
- which is the case when there are no versions and user can’t access initial review state, for which user needs `Review … ` permissions https://github.com/plone/plone.app.layout/blob/0ba390844d7c4f7dc88fe1688e80876eea58be66/plone/app/layout/viewlets/content.py#L344-L347 

The easiest to implement workaround was to patch the re-used viewlet and change its behavior to return an empty list instead of `None`.

Other solutions would have involved to either overwrite the full `@history` endpoint in `plone.restapi` or call the potentially expensive method `fullHistory` or its delegates beforehand.

Jira: https://4teamwork.atlassian.net/browse/CA-2110
Source-Story in RIS: https://4teamwork.atlassian.net/browse/HG-1132

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-2110.bugfix
+++ b/changes/CA-2110.bugfix
@@ -1,0 +1,1 @@
+Fix `@history` endpoint when no history exists. [deiferni]
--- a/opengever/api/tests/test_history.py
+++ b/opengever/api/tests/test_history.py
@@ -1,9 +1,11 @@
 from ftw.testbrowser import browsing
+from opengever.base.role_assignments import RoleAssignmentManager
+from opengever.base.role_assignments import SharingRoleAssignment
+from opengever.document.versioner import Versioner
 from opengever.testing import IntegrationTestCase
 from opengever.testing.helpers import create_document_version
 from plone import api
 import json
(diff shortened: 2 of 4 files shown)


njohner (approved):
LGTM

PR #6982, opened 2021-04-23
Fix policy generator for GEVER policies

phgross:
* Apps endpoint URL question should not be ignored for GEVER deployments
* check enable_workspace_meeting_feature should only be used for teamraum policies.

- [x] Changelog entry


njohner (approved):
LGTM

PR #6984, opened 2021-04-23
Planned tasks can now be opened manually when necessary. 

njohner:
It seems that the option to not directly open the first subtask when triggering a sequential tasktemplate has never worked correctly. Indeed this leads to all subtasks being in the planned state and the planned-open transition was only allowed as an internal transition. It was therefore not possible for the user to open the first task in such cases and hence to complete the task.

As Icon for the transition I've chosen the simple start button.
I've also updated the message displayed in the journal. The other one, was anyway never displayed, as when the transition is done automatically, no response is added.

Screenshot 2021-04-23 at 14 16 56

For https://4teamwork.atlassian.net/browse/CA-2052

## Checklist

- [x] Changelog entry

- New functionality:


phgross (approved):
👍 lgtm

PR #6985, opened 2021-04-23
Cleanup changelog after hotfix release 2021.5.2.

phgross:


HISTORY.rst
```diff
@@ -63,6 +63,13 @@ Changelog
 - When filtering by responsible in globalindex also return tasks assigned to a team the responsible belongs to. [buchi]
 
 
+2021.5.2 (2021-04-20)
+---------------------
+
+- Fix in-progress to close transition (API), for multi adminunit tasks. [phgross]
```
deiferni, line 69: i am a bit confused as to why these are not in any of the newer versions yet 🤔 . should we maybe remove the towncrier file to prevent them from appearing twice?

deiferni (approved):
👍  LGTM apart form the confusion.

HISTORY.rst
phgross: @deiferni https://4teamwork.slack.com/archives/C01BHEGBVU1/p1619169838032500

PR #6986, opened 2021-04-26
Switch GEVER-UI setting to an overall admin_unit setting.

phgross:
Remove path limitation in the geverui cookie, because the decision which UI the user wants to use, should be for the all admin units. Accepting multi admin unit tasks with different UI selected, leads to not found errors (see https://4teamwork.atlassian.net/browse/CA-2063).

Not sure if there is a possibility to migrate or update existing settings.

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-2063.other
+++ b/changes/CA-2063.other
@@ -1,0 +1,1 @@
+Switch GEVER-UI setting to a overall admin_unit setting.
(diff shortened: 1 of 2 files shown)


buchi (changes requested):
I think we should set the path explicitly to `/`. If not the browser will restrict the cookie to the current directory according to RFC 6265 section 4.1.2.4:

"If the server omits the Path attribute, the user agent will use the "directory" of the request-uri's path component as the default value."

buchi (approved):
never mind. it seems that js-cookie sets the path to `/` by default.

PR #6988, opened 2021-04-27
Bump docxcompose to version 1.3.2

buchi:
Now handles DocProperties in a case-insensitive manner, which is the correct behavior and avoids document corruption.

JIRA: https://4teamwork.atlassian.net/browse/CA-2142

## Checklist

- [x] Changelog entry


deiferni (approved):
LGTM 👍

PR #6990, opened 2021-04-27
Improve archival file conversion job management when resolving dossiers.

njohner:
To avoid swamping bumblebee with PDF conversion jobs when executing the `NightlyAfterResolveJobs`, we add a limit to the number of conversion jobs triggered by a deployment in one night.
Note that the `AfterResolveJobs.trigger_pdf_conversion` method was not very smart as it submitted conversion jobs for all documents in the dossier being resolved, but as dossier resolution is recursive, this happens for every subdossier too. We had already introduced a mechanism to at least [avoid sending several jobs for the same document in a single request](https://github.com/4teamwork/opengever.core/pull/5600), but this is not sufficient. Indeed it would still queue conversion jobs for documents from a previously resolved subdossier (which was not resolved in the same request). Also this will not work properly if the `AfterResolveJobs` for a large dossier and its subdossiers are handled over several nights by the nightly jobs. We therefore change `AfterResolveJobs.trigger_pdf_conversion` to only submit documents of the current dossier which are not contained in any of its subdossiers (so directly contained documents or contained in tasks or proposal that are in that dossier). This is much cleaner IMHO and allows us to set a limit to the number of conversion jobs queued to bumblebee in one night (not that we always send all documents for a given dossier / subdossier together, so that the limit is not a hard limit in the sense that it is only checked between jobs)

For https://4teamwork.atlassian.net/browse/CA-1871

## Checklist

- [x] Changelog entry

### DIFF
--- a/opengever/dossier/base.py
+++ b/opengever/dossier/base.py
@@ -233,19 +233,25 @@ class DossierContainer(Container):
 
         return subdossiers
 
-    def get_contained_documents(self):
+    def get_contained_documents(self, unrestricted=False):
         """Returns all documents that are not contained in any subdossier,
         i.e. documents whose parent dossier is this object, this includes
         documents in tasks directly contained in this dossier, etc.
         """
-        path = '/'.join(self.getPhysicalPath())
-        documents = self.portal_catalog.unrestrictedSearchResults(
-            path=path,
-            object_provides=IBaseDocument.__identifier__)
+        query = {
+            'path': '/'.join(self.getPhysicalPath()),
+            'object_provides': IBaseDocument.__identifier__}
+
+        if unrestricted:
+            search_function = self.portal_catalog.unrestrictedSearchResults
+        else:
+            search_function = self.portal_catalog
+
+        documents = search_function(query)
 
         docs_in_subdossiers = set()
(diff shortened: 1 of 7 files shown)


opengever/dossier/base.py
```diff
@@ -233,6 +233,24 @@ def get_subdossiers(
 
         return subdossiers
 
+    def get_contained_documents(self):
+        """Returns all documents that are not contained in any subdossier,
+        i.e. documents whose parent dossier is this object, this includes
+        documents in tasks directly contained in this dossier, etc.
+        """
+        path = '/'.join(self.getPhysicalPath())
+        documents = self.portal_catalog.unrestrictedSearchResults(
```
phgross, line 242: I would add a keyword argument `unrestricted` which is false by default, otherwise the method `get_contained_documents` gets reused and gets documents unrestricted by mistake.

phgross (changes requested):
Looks good, besides the thing mentioned ...

njohner:
@phgross I've added one more commit to include the changes you requested. Can you have another look?

phgross (approved):
👍

PR #6992, opened 2021-04-28
Fix permanently delete workspace documents

elioschmutz:

In the PR https://github.com/4teamwork/opengever.core/pull/6567 we introduced the possibility to remove trashed workspace documents. It was a soft-delete, so the delete action was just using another workflow transition which ensures, that nobody can see the deleted document instead managers.

In the PR https://github.com/4teamwork/opengever.core/pull/6620 we introduced a new placeful workflow for workspaces and a new workflow for workspace documents. This new workspace have no longer provided the transitions to soft-delete the documents. 

On the frontend-side, the delete action was still present, but deleting it was no longer possible:

https://user-images.githubusercontent.com/557005/116458007-6636de80-a864-11eb-9563-f8a0289f4ddc.mov

This PR fixes the broken delete action for workspace documents by introducing a new dedicated action for workspace documents and a new permission to be able to permanently delete workspace documents.

We don't want a soft-delete for workspace documents as it was before. So we don't add transitions to the workflow but added a new DELETE endpoint instead.

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations


opengever/api/tests/test_actions.py
```diff
@@ -1338,6 +1351,75 @@ def test_actions_for_document_in_inactive_dossier(self, browser):
...
+        )
+
+    @browsing
+    def test_actions_for_trashed_document_in_workspace_folder(self, browser):
+        self.login(self.workspace_member, browser)
+        browser.open(self.workspace_folder_document.absolute_url() + '/@trash',
+                     method='POST', headers={'Accept': 'application/json'})
+
+        expected_object_buttons = [
+            {u'icon': u'', u'id': u'copy_item', u'title': u'Copy item'},
```
njohner, line 1414: huh, I don't think copy and move make sense for trashed documents? These actions are not available on the Gever side for trashed documents

opengever/core/profiles/default/actions.xml
```diff
@@ -605,6 +605,20 @@
...
+    <object name="delete_workspace_document" meta_type="CMF Action" i18n:domain="plone">
+      <property name="title" i18n:translate="">Delete</property>
+      <property name="description" i18n:translate="" />
+      <property name="url_expr" />
+      <property name="link_target" />
+      <property name="icon_expr" />
+      <property name="available_expr" />
+      <property name="permissions">
+        <element value="opengever.workspace: Delete Documents" />
+        <element value="Remove GEVER content" />
```
njohner, line 617: Hum, I don't think that makes sense? A User with `Remove GEVER content` permission will not be able to use that action...

opengever/core/upgrades/20210428195416_add_workspace_document_delete_actions/actions.xml
```diff
@@ -0,0 +1,21 @@
...
+    <object name="delete_workspace_document" meta_type="CMF Action" i18n:domain="plone">
+      <property name="title" i18n:translate="">Delete</property>
+      <property name="description" i18n:translate="" />
+      <property name="url_expr" />
+      <property name="link_target" />
+      <property name="icon_expr" />
+      <property name="available_expr" />
+      <property name="permissions">
+        <element value="opengever.workspace: Delete Documents" />
+        <element value="Remove GEVER content" />
```
njohner, line 14: ✂️

opengever/core/profiles/default/actions.xml
```diff
@@ -986,6 +986,19 @@
...
+    <object name="delete_workspace_document" meta_type="CMF Action" i18n:domain="plone">
+      <property name="title" i18n:translate="">Delete</property>
+      <property name="description" i18n:translate="" />
+      <property name="url_expr" />
+      <property name="link_target" />
+      <property name="icon_expr" />
+      <property name="available_expr">python:here.restrictedTraverse('@@plone_interface_info').provides('opengever.trash.trash.ITrashed')</property>
+      <property name="permissions">
+        <element value="opengever.workspace: Delete Documents" />
+        <element value="Remove GEVER content" />
```
njohner, line 998: Same here

opengever/core/upgrades/20210428195416_add_workspace_document_delete_actions/actions.xml
```diff
@@ -18,4 +18,22 @@
...
+    <object name="delete_workspace_document" meta_type="CMF Action" i18n:domain="plone">
+      <property name="title" i18n:translate="">Delete</property>
+      <property name="description" i18n:translate="" />
+      <property name="url_expr" />
+      <property name="link_target" />
+      <property name="icon_expr" />
+      <property name="available_expr">python:here.restrictedTraverse('@@plone_interface_info').provides('opengever.trash.trash.ITrashed')</property>
+      <property name="permissions">
+        <element value="opengever.workspace: Delete Documents" />
+        <element value="Remove GEVER content" />
```
njohner, line 32: ✂️

njohner (commented):
@deiferni can you also have a quick look? 
This makes me nervous, because we introduce a permanent delete for documents, which has the potential to allow deleting documents also in Gever.
I'd definitely not allow that action for "Remove Gever Content" permission, restrict it to a permission that only exists on the Teamraum side. Apart from that it looks OK to me.

opengever/api/tests/test_actions.py
njohner: But that can be treated in a separate issue...

opengever/core/profiles/default/actions.xml
elioschmutz: Right. I though, we need both, but it really makes no sense

elioschmutz:
@njohner hum... the permission is only managed in the teamraum, so it should be safe... but probably your're right that it makes you nervous...

I could improve the `DeleteWorkspaceContent` class ,which is responsible for deleting workspace content, to only delete config within the `IWorkspaceRoot`. That would be really safe then... what do you think?

elioschmutz:
@njohner @deiferni so what do you think? is it enough to extend the `DeleteWorkspaceContent` class with another check? Or do you have another idea?

deiferni (commented):
@elioschmutz we have quickly discussed it and the check in `DeleteWorkspaceContent` would be nice.

Also while implementing please have a look at `ContentDelete` in `plone.restapi`. Maybe you can reuse it instead of copying. Something i just stumbled upon.

elioschmutz:
@njohner @deiferni 👍 updated

njohner (approved):
LGTM

PR #6994, opened 2021-04-30
Move workspace links, when moving a dossier with linked workspace in to a dossier.

phgross:
Workspace links are only allowed on main dossiers, but with moving a dossier into a subdossier it was possible to make a main dossier with linked workspaces to a subdossier. This PR fixes this issue, by automatically move the workspace link to the main_dossier. For https://4teamwork.atlassian.net/browse/CA-1276

There is a small possibility, where a move ist not possible, because the user is not allowed to update the linked workspaces. In that case it raises an Exception. Which is currently not handled with a specific error message in the frontend (I added a not to https://4teamwork.atlassian.net/browse/CA-396, because we need to define the right format for such messages first).

## Checklist

- [x] Changelog entry

- API change:

### DIFF
--- a/changes/CA-1276.bugfix
+++ b/changes/CA-1276.bugfix
@@ -1,0 +1,1 @@
+Transfer workspace link to parent dossier when moving dossier into another dossier. [phgross]
--- a/opengever/dossier/handlers.py
+++ b/opengever/dossier/handlers.py
@@ -11,6 +11,8 @@ from opengever.globalindex.handlers.task import sync_task
 from opengever.globalindex.handlers.task import TaskSqlSyncer
 from opengever.meeting.handlers import ProposalSqlSyncer
 from opengever.task.task import ITask
+from opengever.workspaceclient.interfaces import ILinkedToWorkspace
+from opengever.workspaceclient.interfaces import ILinkedWorkspaces
 from plone import api
 from plone.app.workflow.interfaces import ILocalrolesModifiedEvent
 from Products.CMFPlone.interfaces import IPloneSiteRoot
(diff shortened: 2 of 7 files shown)


changes/CA-1276.bugfix
```diff
@@ -0,0 +1 @@
+Move workspace links, when moving a dossier with linked workspace in to a dossier. [phgross]
```
njohner, line 1: How about "Transfer workspace link to parent dossier when moving dossier into another dossier."

opengever/dossier/handlers.py
```diff
@@ -169,3 +171,18 @@ def update_dossier_touched_date_for_move_event(obj, event):
     """
     if obj == event.object:
         update_dossier_touched_date(obj, event)
+
+
+def move_connected_teamraum_to_main_dossier(obj, event):
+    """If a dossier with linked workspaces gets moved in to a dossier,
```
njohner, line 177: in to -> into

opengever/workspaceclient/linked_workspaces.py
```diff
@@ -369,6 +370,35 @@ def has_linked_workspaces(self):
...
+        main_dossier_oguid = Oguid.for_object(main_dossier).id
+        main_dossier_linked_workspace = ILinkedWorkspaces(main_dossier)
+
+        uids = self.storage.list()
+        for workspace_uid in uids:
+            try:
+                self.client.update_dossier_uid(workspace_uid, main_dossier_oguid)
+            except Exception:
+                raise Exception(
+                    u'Linked workspaces could not been updated, '
```
njohner, line 390: been -> be

opengever/workspaceclient/linked_workspaces.py
```diff
@@ -369,6 +370,35 @@ def has_linked_workspaces(self):
...
+                    'moving this dossier into a dossier not possible.' )
+
+            self.storage.remove(workspace_uid)
+            noLongerProvides(self.context, ILinkedToWorkspace)
```
njohner, line 394: This should be done outside the for loop

opengever/workspaceclient/linked_workspaces.py
```diff
@@ -369,6 +370,35 @@ def has_linked_workspaces(self):
...
+            self.context.reindexObject(idxs=['object_provides'])
```
njohner, line 395: this too

opengever/workspaceclient/linked_workspaces.py
```diff
@@ -369,6 +370,35 @@ def has_linked_workspaces(self):
...
+
+            main_dossier_linked_workspace.storage.add(workspace_uid)
+            if not ILinkedToWorkspace.providedBy(self.context):
```
njohner, line 398: That block should also also be outside the for loop

njohner (commented):
apart from comments, LGTM

opengever/workspaceclient/linked_workspaces.py
phgross: 🙈 thx for the catch!

PR #6996, opened 2021-04-30
Disable `remove` action for workspaces

elioschmutz:
Follow-Up PR for #6992 

This PR disables the `remove` action for workspaces and workspace folders.

The `remove` action is a special action which was implemented for an `opengever.document.document`. The intention of this action is to not permanently delete the content. It just changes the workflow state to a protected state. So we don't want this action for workspaces.

## Before:
Bildschirmfoto 2021-04-30 um 14 38 10

## After:
Bildschirmfoto 2021-04-30 um 14 38 42

## Checklist

- [ ] Changelog entry ℹ️ we don't need a history entry because the previous PR comes into the same release as this one.

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations


opengever/api/tests/test_actions.py
```diff
@@ -1645,6 +1645,89 @@ def test_available_folder_button_actions_for_subdossier_template_folder(self, br
...
+            {u'icon': u'', u'id': u'untrashed', u'title': u'Restore from trash'},
+            {u'icon': u'', u'id': u'remove', u'title': u'Delete'},
+            {u'icon': u'', u'id': u'create_forwarding', u'title': u'Forward'},
+            {u'icon': u'',
+             u'id': u'pdf_dossierlisting',
+             u'title': u'Print selection (PDF)'},
+            {u'icon': u'',
+             u'id': u'pdf_taskslisting',
+             u'title': u'Print selection (PDF)'}
+        ]
```
njohner, line 1683: So the `delete_workspace_document` action is in another category right?

njohner (approved):
And no CL because it's in the same release as the newly introduced `delete_workspace_document`action?

LGTM

opengever/api/tests/test_actions.py
elioschmutz: Yes, and is tested here: https://github.com/4teamwork/opengever.core/pull/6992/commits/c520ea8a2c34fcd0b45e434a3ecac1dace8eb2a0#diff-505e24b0e62316503d395fca49effe5c2e5c2fcfb14b5db268855ea9edc8e31b

PR #6997, opened 2021-04-30
Bundle import: Skip configuration import if not exists.

phgross:
Since #6968 its no longer be possible to import a bundle without a configuration file.

## Checklist
- [x] Changelog entry


njohner (approved):
LGTM

PR #6998, opened 2021-05-03
Remove value restriction for the archival_value field.

phgross:
According to https://4teamwork.atlassian.net/browse/CA-2129 the value restriction on the `archival_value` field, is "fachlich" incorrect and no longer desired.

The PR replaces the RestrictedVocabulary with a SimpleVocabulary, with this changes also the value_propagation for no longer valid values is redundant.

## Checklist

- [x] Changelog entry

### DIFF
--- a/opengever/base/behaviors/lifecycle.py
+++ b/opengever/base/behaviors/lifecycle.py
@@ -14,9 +14,13 @@ from zope import schema
 from zope.component import getUtility
 from zope.container.interfaces import IContainerModifiedEvent
 from zope.interface import alsoProvides
+from zope.interface import implementer
 from zope.interface import Interface
 from zope.interface import provider
 from zope.schema.interfaces import IContextAwareDefaultFactory
+from zope.schema.interfaces import IVocabularyFactory
+from zope.schema.vocabulary import SimpleTerm
+from zope.schema.vocabulary import SimpleVocabulary
 
 
 class ILifeCycleMarker(Interface):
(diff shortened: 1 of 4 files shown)


opengever/base/behaviors/lifecycle.py
```diff
@@ -189,20 +193,26 @@ def custody_period_default(context):
...
+
+
+@implementer(IVocabularyFactory)
+class ArchivalValueVocabulary(object):
+
+    def __call__(self, context):
+        terms = []
+        for archival_value in ARCHIVAL_VALUE_CHOICES:
+            terms.append(
+                SimpleTerm(value=archival_value, token=archival_value,
```
deiferni, line 212: This will use tokens with whitespaces in them. While probably fine, you could easily normalize them with a normalizer as we do in e.g.: https://github.com/4teamwork/opengever.core/blob/1408d8883a5331d66163e1e9af5a3a669a926342/opengever/meeting/committee.py#L38-L50

deiferni (approved):
LGTM 👍

opengever/base/behaviors/lifecycle.py
phgross: Yes, that is the case, but it was also the case with the earlier implementation and did not cause any problems. I don't want to make any changes to the vocabulary if possible, except that all entries are available to prevent possible side effects.

deiferni: Changing the tokens won't change something persistent 😉, so this would be safe. But i'm not insisting, it is fine with whitespaces, just a small imperfection and potential annoyance when debugging and copy pasting stuff. Nothing that blocks a merge.

PR #7000, opened 2021-05-05
Allow querying by custom properties in `@listing` endpoint.

deiferni:
We add support to query the `@listing` endpoint with custom properties/dynamic fields. Available custom columns by listing are provided globally in a `@listing-custom-fields` endpoint.

The custom fields are provided in a global endpoint on the plone site, so that the UI can request these mostly static information once, and then retrieve it from from global application state rather than querying the available fields every time before a listing is requested from the backend.

The `@listing-custom-fields` endpoint returns all available fields based on currently registered property sheets. It takes care of associating property sheets to available listing names and merges fields from different property sheets with the same name and type while building the list.

The `@listing` endpoint now allows all those additional fields to be retrieved via the `columns` argument.

I have moved the responsibility/knowledge about what fields can be indexed in solr to the property sheet schema definition class and provided a distinct field class for solr fields with `SolrDynamicField`.

I have discovered some assumptions built into https://github.com/4teamwork/opengever.core/pull/6959 which are not yet completely validated upon input. The changes here also silently rely on certain states not being introduced into the system. If you do (which you currenlty can) behavior is undefined
- make sure default slot and type-based slot property names don't overlap
- only allow string type values in choices

_I will amend validation in a separate PR._

## Checklist

- [x] Changelog entry

- API change:

### DIFF
--- a/docs/public/dev-manual/api/listings.rst
+++ b/docs/public/dev-manual/api/listings.rst
@@ -126,6 +126,8 @@ werden. Folgende Felder stehen zur Verfügung:
 - ``UID``: UID des Objektes
 - ``watchers``: Liste von Beobachtern des Objekts (Benutzernamen)
 
+Weitere dynamische Felder sind gemäss :ref:`benutzerdefinierte Felder <listing-property_sheets>` verfügbar.
+
 Je nach Auflistungstyp und Inhalt sind bestimmte Felder nicht verfügbar. In diesem
 Fall wird der Wert ``none`` zurückgegeben. So haben Dossiers bspw. keinen Dateinamen,
 siehe Tabelle:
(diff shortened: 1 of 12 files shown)


opengever/api/listing_custom_fields.py
```diff
@@ -0,0 +1,50 @@
...
+}
+
+
+class ListingCustomFieldsGet(Service):
+    """API Endpoint which returns custom fields available for listings.
+
+    It returns a nested data structure with custom fields for each supported
+    listing, if available.
+    Custom fields are provided as follows:
+    - Custom field source are propety sheets registerd for a type associated
```
njohner, line 19: propety -> property

docs/public/dev-manual/api/listings.rst
```diff
@@ -243,6 +245,54 @@ siehe Tabelle:
...
     +--------------------------+----------+---------+--------------+--------------------+---------+---------+---------+----------+-----------------+------------------+-----------------+----------+------------------+----------------+
 
+.. _listing-property_sheets:
+
+Benutzerdefinierte Felder:
+--------------------------
+
+Falls :ref:`benutzerdefinierte Felder <propertysheets>` definiert sind, stehen
+mit dem Endpoint ``@listing-custom-fields`` weitere, dynamische Felder zur
+Verfügung. Der Endpoint kann z.B. dafür benutzt werden für ein Filtermenü für
```
njohner, line 255: für ein Filtermenü für -> für ein Filtermenü oder für

docs/public/dev-manual/api/listings.rst
```diff
@@ -243,6 +245,54 @@ siehe Tabelle:
...
+sichtbare Spalten darzustellen. Er leifert `Titel`, `Typ` und `Name` zurück.
```
njohner, line 256: `Titel`, `Typ` und `Name` zurück. -> `titel`, `type` und `name` zurück.

njohner: and shouldn't you use double ticks here?

njohner (approved):
LGTM 🎉

docs/public/dev-manual/api/listings.rst
deiferni: I was attempting to do it descriptive, without the identifiers. But it seems confusing, i have changed it as you have suggested.

PR #7001, opened 2021-05-06
Properly handle non string choices in property sheet definition.

deiferni:
With this PR we explicitly prevent that users add non string choice values to a choice field in a property sheet definition.

This would not have been prevented before indirectly due to a call to `encode` on all values. Nevertheless we add explicit handling to disallow non string values in choice fields and raise an exception with a more explicit message.

Amends https://github.com/4teamwork/opengever.core/pull/7000, i have not added a separate CL entry.

## Checklist

- ~~[ ] Changelog entry~~


njohner (approved):
LGTM

PR #7002, opened 2021-05-06
Prevent creation of property sheets with conflicting fields.

deiferni:
We now enforce unique fields names within certain subsets of slots.

This has become necessary as our Solr dynamic field implementation of https://github.com/4teamwork/opengever.core/pull/6959 potentially shares field names between different property sheets. To prevent undefined behavior when a sheet for the default slot and a sheet for a type specific slot define the same field, we don't allow this to happen in the first place.

We also don't allow the same property sheet to be assigned to the default slot and a type specific slot at the same time.

The implementation here is a bit more strict that would be necessary as same name but different type would currently result in different Solr dynamic field names. I think it is a reasonable and future proof restrictions, and users can just choose a new name like `title_1` to avoid conflicts.

_I have not written an api changelog as the API is currently only available for `Manager` and not customer facing._

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-1494.other
+++ b/changes/CA-1494.other
@@ -1,0 +1,1 @@
+Prevent adding property sheets with conflicting field names. [deiferni]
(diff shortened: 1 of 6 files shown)


njohner (approved):
LGTM!

PR #7003, opened 2021-05-07
Add close-remote-task endpoint 

phgross:
Adds an additional API endpoint `close-remote-task` which allows to close a remote task and copy documents (related or contained) to a dossier of the current admin_unit.

https://4teamwork.atlassian.net/browse/CA-1641

## Checklist

- [x] Changelog entry

- API change:

### DIFF
--- a/docs/public/dev-manual/api/close_remote_task.rst
+++ b/docs/public/dev-manual/api/close_remote_task.rst
@@ -1,0 +1,25 @@
+.. _close_remote_task:
+
+Remote-Task "Zur Kentnissnahme" abschliessen
+============================================
+
+Der ``@close-remote-task`` Endpoint ermöglicht den Abschluss einer mandantenübergreifende Aufgabe vom Type "zur Kentnissnahme" abzuschliessen und dabei gewisse an die Aufgabe angehängte Dokumente in den eigenen Mandanten zu kopieren.
+
+Im untenstehenden Beispiel ist ``fd`` der Remote-Mandant, und der Benutzer, hugo.boss, will die Aufgabe abschliessen und dabei zwei Dokumente auf seinem lokalen Mandanten ``rk`` kopieren.
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+       POST /rk/@accept-remote-task HTTP/1.1
+       Content-Type: application/json
+       Accept: application/json
+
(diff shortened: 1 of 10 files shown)


opengever/task/browser/close.py
```diff
@@ -203,9 +228,14 @@ def handle_save(self, action):
             dmkey = 'close:%s' % oguid
             task = Task.query.by_oguid(oguid)
 
-            self.copy_documents(
+            closer = CloseTaskHelper()
+            intids_mapping = closer.copy_documents(
                 task, data['dossier'], dm.get(dmkey, 'documents'))
-            self.close_task(task, dm.get(dmkey, 'text'))
```
njohner, line 208: Do you still need the `ChooseDossierStepForm.close_task` and `ChooseDossierStepForm.copy_documents` methods? I think they are not used anymore no?

docs/public/dev-manual/api/close_remote_task.rst
```diff
@@ -0,0 +1,25 @@
+.. _close_remote_task:
+
+Remote-Task "Zur Kentnissnahme" abschliessen
+============================================
+
+Der ``@close-remote-task`` Endpoint ermöglicht den Abschluss einer mandantenübergreifende Aufgbe vom Type "zur Kentnissnahme" abzuschliessen und dabei gewisse an die Aufgabe angehängte Dokumente in den eigenen Mandanten zu kopieren.
```
njohner, line 6: Aufgbe -> Aufgabe

njohner (commented):
Nice! Just two comments:
- Are we making sure somewhere that we are only copying documents that are actually referenced in the task?
- Isn't the OGUID still missing in the serialization of the related items, or will that just work when we require it over the API?

opengever/task/browser/close.py
phgross: you're right 👍

phgross:
@njohner No there is no such check, but i think it's not really necessary. Because all the document copy/transport stuff its done by the real permissions. So nothing else is possible, than copying a document of a foreign client, where I have access as a user, into a dossier. But this is also possible down- and upload.

The oguid can be fetched with the `metadata_fields` attribute.

njohner (approved):
Ok works for me

PR #7004, opened 2021-05-10
Add @upload-structure endpoint

njohner:
We add an endpoint @upload-structure to help prepare a dossier upload into Gever. The endpoint will check whether the proposed upload is possible and reply with information about the structure that would need to be created in Gever, including 
which containers should be created and the `portal_type`s that should be used.

The endpoint checks whether the user is allowed to create content in the given context and whether the objects (container, documents and mails) that would need to be created for that upload are allowed in the current context, including whether the maximal subdossier depth would be exceeded.

For https://4teamwork.atlassian.net/browse/CA-2212

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [ ] api-change label added

### DIFF
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -11,7 +11,7 @@ Other Changes
 
 - The ``@participations`` endpoint now prevents removing the last ``WorkspaceAdmin`` from a workspace.
 - Added ``@listing-custom-fields`` endpoint and allow retrieving custom properties in ``@listing``. (see :ref:`docs <listing-property_sheets>`)
-
+- Added ``@upload-structure`` endpoint. (see :ref:`docs <upload-structure>`)
 
 2021.9.0 (2021-04-29)
 ---------------------
--- a/docs/public/dev-manual/api/upload_structure.rst
+++ b/docs/public/dev-manual/api/upload_structure.rst
@@ -1,0 +1,76 @@
+.. _upload-structure:
+
+Dossier und Dokumente hochladen
+===============================
+
+Der ``@upload-structure`` Endpoint kann verwendet werden um zu überprüfen ob der Upload von Datein möglich ist. Eine Liste von relativen Pfade muss im ``files`` Parameter mitgegeben werden. Der Endpoint berechnet die Struktur von Dossiers und Dokumente welche erstellt werden müsste um die ``files`` Liste in GEVER zu erstellen.
+
+
+  .. sourcecode:: http
+
+    POST /ordnungssystem/dossier-23/@upload-structure HTTP/1.1
+    Accept: application/json
+
+    {
+        "files": ["folder/file.txt", "folder/subfolder/mail.msg"]
+    }
+
+
+  .. sourcecode:: http
+
+    HTTP/1.1 200 OK
+    Content-Type: application/json
+
+    {
+        "items": {
+            "folder": {
+                "@type": "opengever.dossier.businesscasedossier",
+                "is_container": true,
+                "items": {
+                    "file.txt": {
+                        "@type": "opengever.document.document",
+                        "is_container": false,
+                        "relative_path": "folder/file.txt"
+                    },
+                    "subfolder": {
+                        "@type": "opengever.dossier.businesscasedossier",
+                        "is_container": true,
+                        "items": {
+                            "mail.msg": {
+                                "@type": "ftw.mail.mail",
+                                "is_container": false,
+                                "relative_path": "folder/subfolder/mail.msg"
+                            }
+                        },
+                        "relative_path": "folder/subfolder"
+                    }
+                },
+                "relative_path": "folder"
+            }
+        },
+        "items_total": 4,
+        "max_container_depth": 2
+    }
+
+Wenn die Erstellung dieser Struktur nich möglich wäre, zum Beispiel wegen Berechtigungen, Subdossier Tiefe oder weil im Gewünschtem Kontext keine Dossiers / Dokumente erstellt werden können, dann wird mit ``400 Bad Request`` geantwortet.
+
+  .. sourcecode:: http
+
+    POST /ordnungssystem/@upload-structure HTTP/1.1
+    Accept: application/json
+
+    {
+        "files": ["file.txt"]
+    }
+
+
+  .. sourcecode:: http
+
+    HTTP/1.1 400 Bad Request
+    Content-Type: application/json
+
+    {
+        "message": "Some of the objects cannot be added here",
+        "type": "BadRequest"
+    }
+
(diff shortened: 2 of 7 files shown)


opengever/api/upload_structure.py
```diff
@@ -0,0 +1,73 @@
...
+        return files
+
+    def extract_structure(self, files):
+        structure = {'items': {}}
+        max_depth = 0
+        items_total = 0
+        for path in sorted(files):
+            curr_depth = 0
+            segments = path.split("/")
+            substructure = structure
```
deiferni, line 73: could you use parent/child as names here? You are building a tree after all.

opengever/api/upload_structure.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def extract_data(self):
+        data = json_body(self.request)
+        files = data.get("files", None)
+        if not files:
+            raise BadRequest("Property 'files' is required")
+        return files
+
+    def extract_structure(self, files):
+        structure = {'items': {}}
```
deiferni, line 67: and this would be root

changes/CA-1224.bugfix
```diff
@@ -1 +1 @@
-Bundle import: Skip configuration import if not exists. [phgross]
+Add @upload-structure endpoint. [njohner]
```
deiferni, line 1: you have overwritten another changelog 🙈

docs/public/dev-manual/api/api_changelog.rst
```diff
@@ -11,7 +11,7 @@ Other Changes
 
 - The ``@participations`` endpoint now prevents removing the last ``WorkspaceAdmin`` from a workspace.
 - Added ``@listing-custom-fields`` endpoint and allow retrieving custom properties in ``@listing``. (see :ref:`docs <listing-property_sheets>`)
-
+- Added ``@upload-structure`` endpoint.
```
deiferni, line 14: Link to relevant docs section please 😉 !

docs/public/dev-manual/api/upload_structure.rst
```diff
@@ -0,0 +1,74 @@
+Dossier und Dokumente hochladen
+===============================
+
+Der ``@upload-structure`` Endpoint kann verwendet werden um zu überprüfen ob der Upload von Datein möglich ist. Eine Liste von relativen Pfade muss im ``files`` Parameter mitgegeben werden. Der Endpoint berechnet die Struktur von Dossiers und Dokumente welche erstellt werden müsste um die ``files`` Liste im Gever zu erstellen.
```
deiferni, line 4: im Gever -> in GEVER

opengever/api/tests/test_upload_structure.py
```diff
@@ -0,0 +1,458 @@
...
+
+        self.assert_upload_structure_raises_bad_request(
+            browser, self.dossier, [None],
+            u'Empty filename not supported')
+
+    @browsing
+    def test_upload_structure_raises_if_user_cannot_add_content_in_context(self, browser):
+        self.login(self.regular_user, browser)
+
+        self.assert_upload_structure_raises_bad_request(
```
deiferni, line 126: IMO `400` is not correct. Instead it should be `403` Forbidden (The user is not authorized and and re-authenticating will make no difference).

opengever/api/upload_structure.py
```diff
@@ -0,0 +1,130 @@
...
+        "files": ["folder1/file1.txt", "folder1/folder2/file2.eml"]
+    }
+    """
+
+    mail_type = 'ftw.mail.mail'
+    document_type = 'opengever.document.document'
+
+    def __init__(self, context, request):
+        super(UploadStructurePost, self).__init__(context, request)
+        if is_within_workspace(self.context):
```
deiferni, line 30: Ok so i would have expected a bit more object orientation here. E.g. with an adapter. You could adapt context and delegate all the questions you ask below to the adapter. I scribbled together a non working example hopefully illustrating the idea/principle. ☎️  if you want?

```python
@adapts(Interface)
class DefaultUploadStructure(object):
  container_type = None

  def check_permission(self):
    return api.user.has_permission('Add portal content', obj=self.context)
  
  def check_dossier_depth(self)
    return False

@adapts(IDossier)
class DossierUploadStructure(object):
  container_type = "opengever.dossier.businesscasedossier"

  def check_dossier_depth(self)
    return self.depth < MAX_ALLOWED_DEPTH

@adapts(IWorkspace)
class WorkspaceUploadStructure(object):
  container_type = "opengever.workspace.workspacefolder"

  def check_dossier_depth(self)
    return True
```

opengever/api/upload_structure.py
```diff
@@ -0,0 +1,130 @@
...
+            self.container_type = 'opengever.workspace.folder'
+        elif IPrivateContainer.providedBy(context):
+            self.container_type = 'opengever.private.dossier'
+        else:
+            self.container_type = 'opengever.dossier.businesscasedossier'
```
deiferni, line 35: Couldn't we handle dossier template/template folder? Or did we cut that off on purpose?

opengever/api/upload_structure.py
```diff
@@ -0,0 +1,130 @@
...
+
+        if self._is_dossier():
+            current_depth = self.context._get_dossier_depth()
+        else:
+            current_depth = 0
+
+        max_depth = api.portal.get_registry_record(
+            name='maximum_dossier_depth',
+            interface=IDossierContainerTypes,
+            default=100,
```
deiferni, line 109: this should not be necessary. Isn't there a default already? And you should skip past retrieving the configuration where it is not relevant.

deiferni (approved):
Looking almost ready IMO and really nicely tested! 

We can discuss the proposed changes if you want. Nothing that blocks a merge IMO.

opengever/api/upload_structure.py
njohner: I don't really need "child", I could do with only "root" and "parent". But maybe it helps readability, so I changed it anyway

opengever/api/upload_structure.py
njohner: yes for the default it is not necessary.
For the second point, we are skipping in case of workspaces (in a later commit)

opengever/api/upload_structure.py
njohner: We could, but for now it was decided not to allow uploading of dossiers on the template level

njohner:
Ok I think I've included all your comments. Can you have another look?

opengever/api/upload_structure.py
```diff
@@ -0,0 +1,205 @@
...
+    container_type = None
+
+    def __init__(self, context):
+        self.context = context
+
+    def __call__(self, files):
+        self.check_preconditions()
+        self.structure = self.extract_structure(files)
+        self.check_structure()
+        return
```
deiferni, line 51: either just return `self.structure` or drop the unnecessary return

deiferni (approved):
LGTM 👍  apart from:
- nagging
- you completely dropped the CL entry now i think. Please add a towncrier file.

opengever/api/upload_structure.py
njohner: yep sorry, was undecided whether to actually assign the structure to self in the first place, and the return got left behind...

PR #7005, opened 2021-05-10
Fix document template deletion

elioschmutz:
Follow-Up PR for: #6992

This PR fixes an issue introduced in #6992 where it was no longer possible to delete template documents.

Unfortunately teamraum documents and gever documents are the same. The previous implementation registered a new endpoint to delete documents but only handled workspace documents. This prevented the user to remove gever documents.

With this change, we use the default plone restapi delete endpoint if the user tries to delete a gever document.

Since we no longer need the DeleteWorkspaceContent class to delete gever content, we can revert the commit which introduced the check which validated that we're within the workspace root.

## Checklist

- [ ] Changelog entry ⚠️ not needed because it's the same release

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/api/delete.py
+++ b/opengever/api/delete.py
@@ -1,0 +1,33 @@
+from opengever.api.workspace import delete_workspace_content
+from opengever.workspace.utils import is_within_workspace_root
+from plone import api
+from plone.restapi.services.content.delete import ContentDelete
+from zExceptions import Forbidden
+
+
+class DeleteDocument(ContentDelete):
+    """Deletes a document content
+
+    We want to use the default rest-api delete implementation with the default
+    permission for gever documents but a special implementation for workspace
+    documents.
+    """
+    def reply(self):
+        if is_within_workspace_root(self.context):
+            return self.handle_teamraum()
+        else:
+            return self.handle_gever()
+
+    def handle_teamraum(self):
+        if not api.user.has_permission(
+                'opengever.workspace: Delete Documents', obj=self.context):
+            raise Forbidden()
+
+        delete_workspace_content(self.context)
+        return self.reply_no_content()
+
+    def handle_gever(self):
+        if not api.user.has_permission('Delete objects', obj=self.context):
+            raise Forbidden
+
+        return super(DeleteDocument, self).reply()
--- a/opengever/api/tests/test_document.py
+++ b/opengever/api/tests/test_document.py
@@ -205,6 +205,45 @@ class TestDocumentPost(IntegrationTestCase):
             browser.json['message'])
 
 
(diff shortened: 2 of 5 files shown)


opengever/api/delete.py
```diff
@@ -0,0 +1,33 @@
...
+    def reply(self):
+        if is_within_workspace_root(self.context):
+            return self.handle_teamraum()
+        else:
+            return self.handle_gever()
+
+    def handle_teamraum(self):
+        if not api.user.has_permission(
+                'opengever.workspace: Delete Documents', obj=self.context):
+            raise Unauthorized()
```
deiferni, line 24: I think what you want here is `Forbidden` (403) as 401 might trigger authentication on the client-side.

opengever/api/delete.py
```diff
@@ -0,0 +1,33 @@
...
+
+        delete_workspace_content(self.context)
+        return self.reply_no_content()
+
+    def handle_gever(self):
+        if not api.user.has_permission('Delete objects', obj=self.context):
+            raise Unauthorized
```
deiferni, line 31: same here

opengever/api/tests/test_document.py
```diff
@@ -205,6 +205,45 @@ def test_does_not_allow_mails_as_file(self, browser):
...
+
+    @browsing
+    def test_manager_can_permanently_delete_document(self, browser):
+        self.login(self.manager, browser)
+        document_id = self.document.id
+        self.assertIn(document_id, self.dossier.objectIds())
+        browser.open(self.document, method='DELETE', headers=self.api_headers)
+        self.assertEqual(204, browser.status_code)
+        self.assertNotIn(document_id, self.dossier.objectIds())
+
```
deiferni, line 245: i would have expected tests for a workspace here. Or are they someplace else?

deiferni (approved):
LGTM 👍  apart from comments. As deletion is a truly dangerous area i'd suggest to pull @njohner in with a 2nd review.

opengever/api/tests/test_document.py
elioschmutz: tests for teamraum-documents have already been added in the previous PR. See https://github.com/4teamwork/opengever.core/blob/master/opengever/api/tests/test_document.py#L154

njohner (approved):
LGTM

elioschmutz:
@deiferni @njohner updated the PR commits. Had to revert one commit. We don't need this commit anymore since we do not use the `DeleteWorkspaceContent` class to delete documents. All  the other types only exists within a workspace.

PR #7006, opened 2021-05-10
Translate the error message if trying to remove the last remaining workspace admin.

elioschmutz:
Jira: https://4teamwork.atlassian.net/browse/CA-1840

This PR translates the error message if trying to remove the last remaining workspace admin.

This allows us to directly use it in the gever-ui.

## Checklist

- [ ] Changelog entry

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/workspace/locales/de/LC_MESSAGES/opengever.workspace.po
+++ b/opengever/workspace/locales/de/LC_MESSAGES/opengever.workspace.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
-"POT-Creation-Date: 2021-03-26 10:33+0000\n"
+"POT-Creation-Date: 2021-05-10 20:48+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI +ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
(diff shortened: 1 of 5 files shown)


elioschmutz:
@deiferni after several reteries, I endet up in just translating the error message.

My first attempt was to introduce my own exception which then generates it's own "type" in the response:

```json
{
  "type": "MyCustomException",
  "messsage": "Lorem ipsum"
}

but this has several disatvantages. First of all, I need to construct the response object by myself within the endpoints using the participation-manager. Further I have to set the http state manually. Last and the worst, I had to manually abort the transaction in the rest-endpoint. If we throw a `BadRequest` it will all be done automatically.

In the frontend (https://github.com/4teamwork/gever-ui/pull/1747) is now displaying the `message` of an error if available.

Probably we should discuss it further

njohner (changes requested):
I tend to be against this. Is it really the role of the backend to choose how an error should be displayed in the frontend? If not, then we want explicit and easy to parse error messages (as this is the only thing the frontend can parse to make its own representation).

If you look at https://4teamwork.atlassian.net/browse/CA-410, we decided not to translate error messages in the backend.

I'm fine with revisiting the whole concept we've written there, but that is out of scope here, and I don't want to introduce one more way we are returning errors in the API without thinking carefully about it.

elioschmutz:
@njohner ah sure, I read the issue but have not seen that part, that we don't want to translate.

So I 👍 for not solving it here. I already spend too much time for this and we should do it in the correct way this time

njohner:
@elioschmutz Yes the current implementation in the backend is a pain, because the view serializing the exceptions is not flexible enough. Notably it always returns status 500 except for a few exceptions, but it does not work for custom exception types using inheritance. I think we should overwrite that view so that if a custom exception inherits for example from `BadRequest`, well then it sets the status code to 400, and it should serialize additional data when present on the exception.

PR #7009, opened 2021-05-12
Attachment extraction now also works for mails in a workspace.

tinagerber:
Until now, extracting attachments from mail in a workspace did not work because no parent was found where the attachment could be saved.

Jira: https://4teamwork.atlassian.net/browse/CA-2222

## Checklist

- [x] Changelog entry

- API change:
  - [x] API Changelog entry (see [guide](https://4teamwork.atlassian.net/wiki/spaces/4TEAM/pages/451248812/API+Changelog+Guideline

### DIFF
--- a/changes/CA-2222.bugfix
+++ b/changes/CA-2222.bugfix
@@ -1,0 +1,1 @@
+Attachment extraction now also works for mails in a workspace. [tinagerber]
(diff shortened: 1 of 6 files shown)


deiferni (approved):
LGTM 👍

PR #7010, opened 2021-05-17
Add workspaces_without_view_permission attribute to @linked-workspaces endpoint

tinagerber:
In the new frontend, a warning should be displayed on the dossier overview if the dossier is linked to workspaces to which the user does not have access. Therefore, it must be known whether such workspaces exist.

Jira: https://4teamwork.atlassian.net/browse/CA-1180

## Checklist

- [x] Changelog entry

- API change:

### DIFF
--- a/changes/CA-1180.other
+++ b/changes/CA-1180.other
@@ -1,0 +1,1 @@
+Add `workspaces_without_view_permission` attribute to @linked-workspaces endpoint. [tinagerber]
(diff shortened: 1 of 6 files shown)


njohner (commented):
We already have a `has_linked_workspaces_without_view_permission` on the dossier. We could also serialize that on the dossier level? Also `has_linked_workspaces_without_view_permission` is less efficient because it is not cached. On the other side it is always correct (because it is not cached).
Also your solution will break if we can delete workspaces, which is fine for now. Not sure if that is a planed feature or not though.
Finally a bit unrelated, but isn't `has_linked_workspaces_without_view_permission` broken?

tinagerber:
I did not use `has_linked_workspaces_without_view_permission` because it is not cached. Since the linked workspaces list is cached, it makes no sense for me to use the uncached list for this attribute, since it is only used to present a warning.
If at some point workspaces can be deleted, I guess we will have to thoroughly rework Teamraum connect again, otherwise it will not be possible to close a dossier either.

tinagerber:
And why is `has_linked_workspaces_without_view_permission` broken? I don't see it.

njohner (approved):
As discussed, LGTM.

PR #7011, opened 2021-05-18
Include backreference list to dossier and documents api serialization.

phgross:
This PR adds an additional key to the dossier and documents api serialization which lists summaries of all backreferences. It uses plone.restapis default summary serializer for the backreferences.

This is used for display backreferences in the new UI, see https://4teamwork.atlassian.net/browse/CA-1970

## Checklist
- [x] Changelog entry

- API change:

### DIFF
--- a/changes/CA-1970.feature
+++ b/changes/CA-1970.feature
@@ -1,1 +1,1 @@
-Include backreference list to dossier api serialization. [phgross]
+Include backreference list in dossier and documents api serialization. [phgross]
--- a/changes/CA-1970.feature
+++ b/changes/CA-1970.feature
@@ -1,0 +1,1 @@
+Include backreference list to dossier api serialization. [phgross]
--- a/opengever/api/serializer.py
+++ b/opengever/api/serializer.py
@@ -1,3 +1,4 @@
+from Acquisition import aq_inner
 from ftw.bumblebee.interfaces import IBumblebeeable
 from ftw.bumblebee.interfaces import IBumblebeeDocument
 from Missing import Value as MissingValue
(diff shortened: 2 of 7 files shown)


opengever/api/serializer.py
```diff
@@ -90,6 +94,28 @@ def extend_with_groupurl(result, context, request):
...
+    intids = getUtility(IIntIds)
+    catalog = getUtility(ICatalog)
+    relations = catalog.findRelations(
+        {'to_id': intids.getId(aq_inner(context)),
+         'from_attribute': reference_attribute_name})
+    ids = [relation.from_id for relation in relations if relation.from_object]
+
+    for intid in ids:
+        summaries.append(
+            getMultiAdapter((intids.queryObject(intid), request),
```
njohner, line 112: So we are actually fetching the objects here and then summary serializing them. I guess this is a small set of objects, so it's not really an issue, and we anyway do not have any other choice. 

As a sidenote, is the step over the `ids` actually necessary? couldn't we instead simply do
`summaries = [getMultiAdapter((relation.from_object, request),ISerializeToJsonSummary)() for relation in relations]`?

changes/CA-1970.feature
```diff
@@ -1 +1 @@
-Include backreference list to dossier api serialization. [phgross]
+Include backreference list to dossier and documents api serialization. [phgross]
```
njohner, line 1: to -> in

njohner (approved):
Mainly the one comment, not sure why we need to go over the IntIds...
Otherwise LGTM

opengever/api/serializer.py
phgross: 🙈 you’re totally right! thx

PR #7012, opened 2021-05-18
Automatically assign placeful workflow policies when workspace root, private root and inbox are created.

deiferni:
With this PR we make sure that placeful workflow policies are assigned automatically for the following content types:
- workspace root
- inbox
- private root

Until now placeful workflow policy assignment relied on having a correct initial content json file. This covers most of the ways that content is created in production but would not have been future-proof when:
- creating the content manually as `Manager`
- importing content via bundle (Policyless or normal bundle)

We change the approach and make sure the application is always in control of correct content type configuration. To achieve so we move placeful workflow policy assignment into an event handler.

We also drop the `placefulworkflowupdater` section from the import pipeline. So content imported with that pipeline from a `gever_content` folder in a generic setup profile will no longer set any placeful workflows by reading the `_placefulworkflow` attribute.  We do this for two reasons:
- prevent policies from overwriting the placeful workflow now controlled by application code
- not having to migrate all policies and drop the attribute

**Changes to workflow policy assignment**
We slightly change how placeful workflow policies for inboxes are assigned in case of nested inboxes. Previously the placeful workflow policy would be assigned to the inbox container only. With the changes in this PR we always assign the placeful workflow policy to the inbox, and not to the inbox container.

**Out of scope:**
A migration for existing inboxes will be implemented with https://4teamwork.atlassian.net/browse/CA-2262.

Jira: https://4teamwork.atlassian.net/browse/CA-1358

## Checklist

- [x] Changelog entry

### DIFF
--- a/opengever/setup/cfgs/gever_content.cfg
+++ b/opengever/setup/cfgs/gever_content.cfg
@@ -16,7 +16,6 @@ pipeline =
     schemaupdater
     dx_schemaupdater
     workflowupdater
-    placefulworkflowupdater
     propertiesupdater
     modelconstructor
     local_roles
--- a/changes/CA-1358.bugfix
+++ b/changes/CA-1358.bugfix
@@ -1,0 +1,1 @@
+Automatically assign placeful workflow policies when workspace root, private root and inbox are created. [deiferni]
--- a/opengever/api/tests/test_linked_workspaces.py
+++ b/opengever/api/tests/test_linked_workspaces.py
@@ -856,7 +856,7 @@ class TestListDocumentsInLinkedWorkspaceGet(FunctionalWorkspaceClientTestCase):
                   u'file_extension': u'',
                   u'filename': u'',
                   u'is_leafnode': None,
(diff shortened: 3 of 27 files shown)


opengever/setup/cfgs/gever_content.cfg
```diff
@@ -16,7 +16,6 @@ pipeline =
     schemaupdater
     dx_schemaupdater
     workflowupdater
-    placefulworkflowupdater
```
phgross, line 19: I wonder if we remove the placefulworkflowupdater, will the specification in the existing policies just be ignored or will errors occur?

phgross (approved):
LGTM 👍

opengever/setup/cfgs/gever_content.cfg
deiferni: @phgross yes, i have double checked that:
- there is no exception while processing policies with the `_placefulworkflow`
- the attribute is discarded and not set on the object

PR #7013, opened 2021-05-19
Add primary_repository URL to the @config endpoint.

phgross:
For GEVER deployments with multiple repositories, the gever-ui needs to know which is the primary one. Therefore I added the `primary_repository` to the `@config` endpoint.  For https://4teamwork.atlassian.net/browse/CA-1119

In the refinement part of the story it was described that the navigation endpoint should be extended. This would be much more difficult for the UI and above all it would mean that a catalog query for repositoryroots would have to be performed for every route change. Therefore I decided to extend the config endpoint instead.

## Checklist

- [x] Changelog entry

- API change:

### DIFF
--- a/opengever/api/config.py
+++ b/opengever/api/config.py
@@ -7,6 +7,7 @@ from opengever.officeconnector.helpers import is_client_ip_in_office_connector_d
 from opengever.ogds.base.utils import get_current_org_unit
 from opengever.ogds.models.service import ogds_service
 from opengever.private import get_private_folder_url
+from opengever.repository.browser.primary_repository_root import PrimaryRepositoryRoot
 from plone.restapi.services import Service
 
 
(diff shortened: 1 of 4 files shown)


opengever/api/config.py
```diff
@@ -44,3 +45,7 @@ def add_additional_infos(self, config):
         ogds_inbox = get_current_org_unit().inbox()
         current_user = ogds_service().fetch_current_user()
         config['is_inbox_user'] = current_user in ogds_inbox.assigned_users()
+
+        primary_root = PrimaryRepositoryRoot(
+            self.context, self.request).get_primary_repository_root()
+        config['primary_repository'] = primary_root.absolute_url()
```
deiferni, line 51: This is easily testable, please add some simple tests. 
_Actually i am a bit puzzled as to why no existing tests fail._

deiferni (approved):
👍  LGTM but please add simple testing.

opengever/api/config.py
phgross: i didn't add it to the branch 🙈.

PR #7014, opened 2021-05-19
Fix filtering on values containing spaces in @listing endpoint.

tinagerber:
In order to filter correctly in solr, terms containing spaces must be surrounded by quotation marks.

Jira: https://4teamwork.atlassian.net/browse/CA-2176

## Checklist

- [x] Changelog entry

- API change:

### DIFF
--- a/opengever/api/solr_query_service.py
+++ b/opengever/api/solr_query_service.py
@@ -1,10 +1,9 @@
-from copy import copy
 from copy import deepcopy
 from DateTime import DateTime
 from DateTime.interfaces import DateTimeError
 from ftw.solr.converters import to_iso8601
 from ftw.solr.interfaces import ISolrSearch
-from ftw.solr.query import SPECIAL_CHARS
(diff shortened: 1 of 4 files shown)


opengever/api/solr_query_service.py
```diff
@@ -30,19 +29,14 @@
 import Missing
 
 
-TO_ESCAPE = copy(SPECIAL_CHARS)
-TO_ESCAPE.append(' ')
-
-
 def filter_escape(term):
```
deiferni, line 32: Could you add unit-tests for this function? It looks like it does not use any plone/zope dependencies and you can just use it in a unit test. Advantages of doing so:
- Document and test extensively how it behaves for edge cases (leading/trailing whitespace …)
- Have fast feedback during development (next time 😉 )

deiferni (approved):
LGTM 👍  but please add tests for edge cases so we gain more certainty this works correctly and as expected.

PR #7015, opened 2021-05-21
Fix @config endpoint for sites without a repository root.

phgross:
Handle sites without a repository root in the config endpoint. This fixes the currently failing UI tests.

_Changelog entry - not necessary_


njohner (approved):
LGTM.

PR #7016, opened 2021-05-21
Fix a problem in relation deserializer when not all path elements are accessible.

phgross:
plone.restapi's RelationChoiceFieldDeserializer uses restrictedTraverse to fetch relations given by URL or path. But this only works if every single path element is accesssible by the current user. Therefore I customized the Relation deseralizer and fetch the element in the same manner the ZCatalog does - fetch the parent unrestricted and the object itself restricted.

See https://4teamwork.atlassian.net/browse/CA-2220

## Checklist

- [x] Changelog entry

### DIFF
--- a/opengever/api/relationfield.py
+++ b/opengever/api/relationfield.py
@@ -1,0 +1,79 @@
+from opengever.base.interfaces import IOpengeverBaseLayer
+from plone.dexterity.interfaces import IDexterityContent
+from plone.restapi.deserializer.relationfield import RelationChoiceFieldDeserializer
+from plone.restapi.interfaces import IFieldDeserializer
+from Products.CMFCore.utils import getToolByName
+from z3c.relationfield.interfaces import IRelationChoice
+from zope.component import adapter
+from zope.component import getMultiAdapter
+from zope.component import queryUtility
+from zope.interface import implementer
+from zope.intid.interfaces import IIntIds
+import six
+
+
+@implementer(IFieldDeserializer)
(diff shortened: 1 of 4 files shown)


opengever/api/tests/test_relationfield.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @browsing
+    def test_relation_by_path_is_possible(self, browser):
+        self.login(self.regular_user, browser=browser)
+
+        browser.open(
+            self.dossier.absolute_url(),
+            method='PATCH',
+            data=json.dumps(
+                {'relatedDossier': [u'/ordnungssystem/fuhrung/vertrage-und-vereinbarungen/dossier-7']}
```
njohner, line 37: you could use `self.empty_dossier.absolute_url_path()`

opengever/api/relationfield.py
```diff
@@ -0,0 +1,79 @@
...
+            )
+
+        self.field.validate(obj)
+        return obj
+
+    def get_obj_by_path(self, portal, path):
+        """restrictedTraverse checks all of the objects along the path are
+        validated with the security machinery. But we only need to know if the
+        user is able to access the given object."""
+        path = path.split('/')
```
njohner, line 74: Is there any scenario where the path can end with a slash? In that case `path = path.rstrip("/").split("/")` would be safer.

njohner (approved):
LGTM. Some comments but it's up to you.

PR #7020, opened 2021-05-26
Fix SQLHypermediaBatch for undefined sort orders.

phgross:
When sorting according to a non-unique column, response of SQLHypermediaBatches could contain duplicate items and other items could be missing. This is because the sort order of successive queries is not guaranteed to be identical (except when specifically sorting on a key with a unique value for each item).

This issue has been fixed with #6624 for ogds listings but not for all cases we use the SQLHypermediaBatches, therefore i moved the additional sorting on unique key to the SQLHyperMediaBatch itself.

Closes https://4teamwork.atlassian.net/browse/CA-2060

## Checklist
- [x] Changelog entry

### DIFF
--- a/changes/CA-2060.bugfix
+++ b/changes/CA-2060.bugfix
@@ -1,0 +1,1 @@
+Fix SQLHypermediaBatch for undefined sort orders.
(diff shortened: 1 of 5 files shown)


njohner (approved):
LGTM

PR #7023, opened 2021-05-27
Properly resolves path_parent field queries for the @solrsearch endpoint if they are relative to the virtual root

elioschmutz:
The `path_parent` field query of the `@solrsearch` endpoint properly resolves paths relative to the virtual host url and joins multiple `path_parent` field queries with an OR operator.

Jira: https://4teamwork.atlassian.net/browse/CA-1588

This change is required because a frontend does not know anything about the physical path of an object which is required to do solr-path queries.

the path indexes of an object is always the physical path:

Bildschirmfoto 2021-05-27 um 13 16 22

But a frontend does not know anything about it. I.e.

the url http://www.example.com/examplecontent can point to the same physical path like the url http://localhost:8080/fd/examplecontent =>

http://www.example.com/examplecontent => `/plone/examplecontent`
http://localhost:8080/plone/examplecontent => `/plone/examplecontent`

But we can only extract the path relative to the current site domain:

http://www.example.com/examplecontent => `/examplecontent`
http://localhost:8080/plone/examplecontent => `/plone/examplecontent`

In the backend, we need to replace this path relative to the virtual host with the physical path of the content itself to be able to do a solr query.

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/api/solrsearch.py
+++ b/opengever/api/solrsearch.py
@@ -1,3 +1,4 @@
+from ftw.solr.query import make_filters
 from ftw.solr.query import make_path_filter
 from ftw.solr.query import make_query
 from opengever.api.breadcrumbs import Breadcrumbs
@@ -60,10 +61,46 @@ class SolrSearchGet(SolrQueryBaseService):
         else:
             filters = []
 
+        self.add_path_parent_filters(filters)
         self.add_path_filters(filters, params)
         self.add_portal_types_filter(filters)
 
         return filters
+
+    def add_path_parent_filters(self, filters):
+        """A frontend usualy only knows the virtual path, not the real physical path
+        of an object. But the solr index value contains the physical path of
+        the object. Thus, we need to replace the paths relative to the virtual root
+        with the physical path of an object.
+
+        In addition, it joins multiple path_parent filters with an OR operator which
+        makes it possible to query for multiple path_parents
+        """
+        requested_parent_paths = []
+        portal = api.portal.get()
+        for query in list(filters):
+            if not query.startswith("path_parent:"):
+                continue
+
+            # extract the path from the query and unescape
+            path = query.split(":", 1)[1].replace('\\', '')
+
+            # A frontend does not know anything about the physical path of an object.
+            # We have to take care of this. So we have to removes the virtual
+            # portal path to get the relative path from the plone site root
+            relativePath = path.lstrip('/').lstrip(portal.virtual_url_path()).lstrip('/')
+
+            # And then we can extend the physical portal path with the requested
+            # relative path
+            physicalPath = portal.getPhysicalPath()
+            if relativePath:
+                physicalPath += (relativePath, )
+
+            requested_parent_paths.append('/'.join(physicalPath))
+            filters.remove(query)
+
+        if (requested_parent_paths):
+            filters.extend(make_filters(path_parent=requested_parent_paths))
 
     def extract_path_filter_value(self, filters):
         """If path is not specified we search in the current context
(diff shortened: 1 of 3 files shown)


opengever/api/solrsearch.py
```diff
@@ -60,11 +61,49 @@ def extract_filters(self, params):
...
+        requested_parent_paths = []
+        portal = api.portal.get()
+        for query in list(filters):
+            if not query.startswith("path_parent:"):
+                continue
+
+            # extract the path from the query and unescape
+            path = query.split(":", 1)[1].replace('\\', '')
+
+            # A frontend does not now anything about the physical path of an object.
```
njohner, line 88: now -> know

opengever/api/solrsearch.py
```diff
@@ -60,11 +61,49 @@ def extract_filters(self, params):
...
+            # portal path to get the relative path from the plone site root
+            relativePath = path.lstrip('/').lstrip(portal.virtual_url_path()).lstrip('/')
+
+            # And then we can extend the physical portal path with the requested
+            # relative path
+            physicalPath = portal.getPhysicalPath()
+            if relativePath:
+                physicalPath += (relativePath, )
+
+            content = portal.restrictedTraverse('/'.join(physicalPath), None)
```
njohner, line 99: I wonder whether this should not be unrestricted? I mean I should be allowed to search inside an object on which I have no view permissions no?

opengever/api/solrsearch.py
```diff
@@ -60,11 +61,49 @@ def extract_filters(self, params):
...
+            # A frontend does not now anything about the physical path of an object.
+            # We have to take care of this. So we have to removes the virtual
+            # portal path to get the relative path from the plone site root
+            relativePath = path.lstrip('/').lstrip(portal.virtual_url_path()).lstrip('/')
+
+            # And then we can extend the physical portal path with the requested
+            # relative path
+            physicalPath = portal.getPhysicalPath()
+            if relativePath:
+                physicalPath += (relativePath, )
```
njohner, line 97: I'm surprised there is no function doing this already. But I searched and did not find anything.

opengever/api/solrsearch.py
```diff
@@ -60,11 +61,49 @@ def extract_filters(self, params):
...
+
+            content = portal.restrictedTraverse('/'.join(physicalPath), None)
+            resolved_path = '/'.join(content.getPhysicalPath()) if content else path
```
njohner, line 100: do we need this? do you see scenarios where it does not resolve to any object but we still want to use the path as filter?

opengever/api/solrsearch.py
njohner: I'm a bit confused why this is not enough. Why then resolve the object and then get the physical path from it. This is already the physical path no?

njohner (approved):
I'm not sure whether the part transforming the path needs to be that complicated, but otherwise LGTM

opengever/api/solrsearch.py
elioschmutz: you're completely right 👍 I don't need to resolve the object

opengever/api/solrsearch.py
elioschmutz: removed because there is no need to resolve the object itself.

opengever/api/solrsearch.py
elioschmutz: removed because it's not needed anymore

PR #7026, opened 2021-05-28
Add reference_number_addendum field to repository root and use it in reference number

tinagerber:
This change is necessary in order to have the possibility of having unique reference numbers even when using two repository roots. 
If the field `reference_number_addendum` is changed, `reference` and `sortable_reference` must be reindexed in Solr.

Jira: https://4teamwork.atlassian.net/browse/CA-2088

## Checklist

- [x] Changelog entry

- New translations

### DIFF
--- a/docs/schema-dumps/opengever.repository.repositoryroot.schema.json
+++ b/docs/schema-dumps/opengever.repository.repositoryroot.schema.json
@@ -24,6 +24,12 @@
             "description": "",
             "_zope_schema_type": "TextLine"
         },
+        "reference_number_addendum": {
+            "type": "string",
+            "title": "Aktenzeichen Zusatz",
+            "description": "Achtung: \u00c4nderung erfordert Neuindexierung von \"reference\" und \"sortable_reference\".",
+            "_zope_schema_type": "TextLine"
+        },
         "title_de": {
             "type": "string",
             "title": "Titel (deutsch)",
@@ -68,6 +74,7 @@
         "valid_from",
         "valid_until",
         "version",
+        "reference_number_addendum",
         "title_de",
         "title_fr",
         "title_en"
--- a/opengever/repository/locales/de/LC_MESSAGES/opengever.repository.po
+++ b/opengever/repository/locales/de/LC_MESSAGES/opengever.repository.po
@@ -3,7 +3,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: 1.0\n"
-"POT-Creation-Date: 2021-01-25 09:56+0000\n"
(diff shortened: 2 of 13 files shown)


docs/schema-dumps/opengever.repository.repositoryroot.schema.json
```diff
@@ -24,6 +24,12 @@
             "description": "",
             "_zope_schema_type": "TextLine"
         },
+        "reference_number_addendum": {
```
njohner, line 27: How about using "reference_number_suffix" instead? We use `reference_number_prefix` for other content types

opengever/repository/locales/de/LC_MESSAGES/opengever.repository.po
```diff
@@ -43,6 +43,11 @@ msgstr "Hinzufügen von Geschäftsdossiers erlauben"
 msgid "description_allow_add_businesscase_dossier"
 msgstr "Wählen Sie, ob es in dieser Ordnungsposition erlaubt ist, Geschäftsdossiers hinzuzufügen. Ist diese Option deaktiviert, kann der Benutzer nur Dossiers aus einer Vorlage oder Spezialdossiers erstellen."
 
+#. Default: "Attention: Change requires reindexing of \"reference\" and \"sortable_reference\" in Solr."
+#: ./opengever/repository/repositoryroot.py
+msgid "description_reference_number_addendum"
+msgstr "Achtung: Änderung erfordert Neuindexierung von \"reference\" und \"sortable_reference\" im Solr."
```
njohner, line 49: Not only in solr. It's also in the catalog...

njohner (approved):
LGTM. 
I think I'm fine with addendum too, suffix  is just a suggestion.

docs/schema-dumps/opengever.repository.repositoryroot.schema.json
tinagerber: Since I asked yesterday in the Slack channel and people were in favor of `addendum`, I will keep `addendum`.

opengever/repository/locales/de/LC_MESSAGES/opengever.repository.po
tinagerber: Ah, I thought I read somewhere that the index had been removed from the catalog, but that was probably imagination. 🙈

docs/schema-dumps/opengever.repository.repositoryroot.schema.json
njohner: Ah hadn't seen the slack discussion, was on my day off

PR #7028, opened 2021-05-31
Unify inbox placeful workflow assignment.

deiferni:
With this upgrade step we make sure that placeful workflows for inboxes
are assigned in a consistent way across all deployments.

We ensure that:
- every inbox has document and mail placeful workflows assigned
- inbox containers no longer have placeful workflows

Assigning placeful workflows to inboxes is now also done automatically
by the system when an inbox is created as of https://github.com/4teamwork/opengever.core/pull/7012. This PR should ensure consistent state for the future.

Jira: https://4teamwork.atlassian.net/browse/CA-2262

## Checklist

- [x] Changelog entry

### DIFF
--- a/opengever/core/upgrades/20210519102848_fix_inbox_workflow_policy_assignment/upgrade.py
+++ b/opengever/core/upgrades/20210519102848_fix_inbox_workflow_policy_assignment/upgrade.py
@@ -1,0 +1,152 @@
+from ftw.upgrade import ProgressLogger
+from ftw.upgrade import UpgradeStep
+from ftw.upgrade.placefulworkflow import PlacefulWorkflowPolicyActivator
+from ftw.upgrade.utils import SizedGenerator
+from ftw.upgrade.workflow import WorkflowSecurityUpdater
+from opengever.inbox.container import IInboxContainer
+from opengever.inbox.inbox import IInbox
+from plone import api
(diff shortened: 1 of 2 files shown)


opengever/core/upgrades/20210519102848_fix_inbox_workflow_policy_assignment/upgrade.py
```diff
@@ -0,0 +1,146 @@
...
+                    inbox, has_container_workflow=has_container_workflow
+                )
+
+    def fix_global_inbox(self):
+        query = {
+            'object_provides': [IInbox.__identifier__]
+        }
+        inboxes = self.catalog_unrestricted_search(
+            query, full_objects=True
+        )
```
njohner, line 103: Can we log a warning here if `len(inboxes) > 1`, as this is a configuration we do not support anymore?

njohner (approved):
LGTM

opengever/core/upgrades/20210519102848_fix_inbox_workflow_policy_assignment/upgrade.py
deiferni: have amended as suggested. I am quite certain this case does not exist anywhere. But even if it does, it will be migrated correctly.

PR #7029, opened 2021-06-01
Register a WillBeMoved handler which initializes the reference_number

phgross:
Register a WillBeMoved handler which initalizes reference_number. Because we can't control event handler orders, we need to make sure that the reference_number is already initalized for the main dossier. 
Otherwise it can happen that the reference number of subdossiers or documents are reindexed, but the dossier reference number is not yet correctly updated (see https://4teamwork.atlassian.net/browse/CA-2290).

I also cleaned up other event handlers.

PR contains an upgradestep which reindexes the `reference` and `sortable_reference` in solr.

## Checklist
- [x] Changelog entry


opengever/dossier/handlers.py
```diff
@@ -73,8 +73,6 @@ def set_former_reference_after_moving(obj, event):
     prefix_adapter = IReferenceNumberPrefix(parent)
     prefix_adapter.set_number(obj)
 
-    obj.reindexObject(idxs=['reference', 'sortable_reference'])
```
njohner, line 58: A bit unsure here. I'm actually unsure why we even need to set the reference number here. It anyway happens in `save_reference_number_prefix`

opengever/core/upgrades/20210205084521_add_sortable_reference_number_index/upgrade.py
```diff
@@ -16,9 +16,14 @@ def __call__(self):
...
 
     def index_sortable_reference_number(self):
-        manager = getUtility(ISolrConnectionManager)
-        query = {'object_provides': IDexterityContent.__identifier__}
-        for obj in self.objects(query, 'Index sortable_reference in Solr'):
-            handler = getMultiAdapter((obj, manager), ISolrIndexHandler)
-            handler.add(['sortable_reference'])
-        manager.connection.commit(soft_commit=False, extract_after_commit=False)
+
+        # Disable this upgradestep because its done by the later upgradestep
```
njohner, line 20: its -> it's

opengever/core/upgrades/20210601120606_update_reindex_reference_and_sortable_reference_index/upgrade.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def __call__(self):
+        self.index_sortable_reference_number()
+
+    def index_sortable_reference_number(self):
+        manager = getUtility(ISolrConnectionManager)
+        query = {'object_provides': IDexterityContent.__identifier__}
+        for obj in self.objects(query, 'Reindex reference and sortable_reference in Solr'):
+            handler = getMultiAdapter((obj, manager), ISolrIndexHandler)
+            handler.add(['reference', 'sortable_reference'])
```
njohner, line 24: Shouldn't we be reindexing in the catalog as well? This would be problematic because it's both an index and a metadata, so in that case maybe check whether it's correct in the catalog before reindexing the complete metadata?

njohner (commented):
LGTM, just unsure about the upgrade step?

njohner:
Forgot the following comments:
In “Remove no longer needed additional task and proposal sync” commit description:
  - ensure -> ensures
  - initalized -> initialized
  - triggerd -> triggered

phgross:
@njohner updated the PR and extended the upgradestep, with a catalog check and update. Could you take another look?

njohner (approved):
LGTM

PR #7032, opened 2021-06-04
Allow cookie based authentication when using @login endpoint

buchi:
Allows us to authenticate with a cookie instead of a JWT.
We use the @login endpoint only for local development and testing. In production we use ftw.casauth for login.

JIRA: https://4teamwork.atlassian.net/browse/CA-2075

## Checklist

- [x] Changelog entry


phgross (approved):
👍 LGTM

PR #7034, opened 2021-06-07
Fix inbound mail.

njohner:
Of course in https://github.com/4teamwork/opengever.core/pull/7019 I broke the only feature I couldn't really test locally 😓 ...

For https://4teamwork.atlassian.net/browse/CA-1504

## Checklist

- [ ] Changelog entry

### DIFF
--- a/opengever/mail/browser/inbound.py
+++ b/opengever/mail/browser/inbound.py
@@ -20,3 +20,4 @@ class GeverMailInbound(MailInbound):
                 'message', filename, filelike, self.request)
         except Invalid as e:
             raise MessageContainsVirus(e.message)
+        return msg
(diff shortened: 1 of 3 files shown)


deiferni (approved):
🐽 🙈

deiferni:
Hm actually, don't you have any test for the inbound view at all? I would prefer at least a happy-path test which is inpired by the existing tests in `ftw.mail` https://github.com/4teamwork/ftw.mail/blob/4eae23c37fb614b83c5d271426a79d896caad708/ftw/mail/tests/test_inbound.py.

njohner:
Yeah I wrote a test for virus detection in the inbound view, but didn't manage to write an integration test case for the happy path (somehow finding the user by E-mail did not work). 
I'll give it another shot.

njohner:
Voila, also testing the happy path now!

deiferni (approved):
LGTM 👍  thanks!

_the way i understand it this would have caught your previous error_

njohner:
Yes, this would have caught the previous error. This really creates the E-mail over the inbound view...

PR #7036, opened 2021-06-07
Fixed moving dossier with a tasktemplate process.

phgross:
When moving a dossier with a task_process it is not guaranteed that the objects are moved in the tasktemplate order, so it can happen, that the predecessor does not yet exists. Because the intid stays the same after move, its safe to just don't update the predecessor.

Fixes https://4teamwork.atlassian.net/browse/CA-2343

## Checklist
- [x] Changelog entry

### DIFF
--- a/opengever/globalindex/model/task.py
+++ b/opengever/globalindex/model/task.py
@@ -1,5 +1,6 @@
 from datetime import date
 from DateTime import DateTime as ZopeDateTime
+from opengever.base.exceptions import InvalidOguidIntIdPart
 from opengever.base.model import Base
 from opengever.base.model import create_session
 from opengever.base.model import is_oracle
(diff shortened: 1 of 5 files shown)


opengever/globalindex/model/task.py
```diff
@@ -236,7 +237,18 @@ def sync_with(self, plone_task):
...
-        predecessor = plone_task.get_tasktemplate_predecessor()
+        try:
+            predecessor = plone_task.get_tasktemplate_predecessor()
+        except InvalidOguidIntIdPart as e:
+            # When moving a dossier with a task_process it is not guaranteed
+            # that the objects are moved in the tasktemplate order, so it can
+            # happen, that the predecessor does not yet exists. Because the
+            # intid stays the same after move, its safe to not updated the predecessor.
+            if not graceful:
+                raise e
```
deiferni, line 248: please use a naked `raise` to preserve the exception traceback.

deiferni (approved):
👍  LGTM apart from comment.

PR #7039, opened 2021-06-08
Allow to move dossiers from an old repository root to a new one.

elioschmutz:
Jira: https://4teamwork.atlassian.net/browse/CA-2089

This PR extends the `RepositorySourcePathBinder` to allow moving contents between an old and a new repository-root if the user is within an old repository root.

The implementation differs to the main refinement in the following case:
I have not implemented the registry entry. It behaves like in the frontend: if the user tries to move an object within an old repository root, he is allowed to move it to the new one but not vice versa.

## Moving within an old repository root:
Bildschirmfoto 2021-06-08 um 15 42 37

# Moving within a new repository root:
Bildschirmfoto 2021-06-08 um 15 40 46

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/base/tests/test_source_binder.py
+++ b/opengever/base/tests/test_source_binder.py
@@ -1,6 +1,9 @@
+from ftw.builder import Builder
+from ftw.builder import create
 from ftw.testbrowser import browsing
 from opengever.base.source import DossierPathSourceBinder
 from opengever.base.source import RepositoryPathSourceBinder
+from opengever.repository.browser.primary_repository_root import PrimaryRepositoryRoot
 from opengever.testing import IntegrationTestCase
 from opengever.testing import SolrIntegrationTestCase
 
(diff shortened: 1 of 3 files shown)


opengever/base/tests/test_source_binder.py
```diff
@@ -43,6 +46,84 @@ def test_query_is_limited_to_current_repository(self):
...
+
+        self.assertItemsEqual(
+            [
+                'source old repository', 'source new repository', 'source test document',
+                '1. source branch new', 'source test inbox document',
+                'source test private document', 'source test proposal document'
+            ],
+            [term.title for term in source.search("source")])
+
+    def test_query_returns_only_primary_root_if_context_is_within_old_reporoot(self):
```
phgross, line 90: is_within_new_reporoot

phgross (approved):
👍 LGTM besides the thing mentioned.

PR #7041, opened 2021-06-09
Fix persisting default values for objects created over API

njohner:
With this PR we fix an issue where default values do not get persisted on objects created over the API. The problem arose only when setting as value the `default_value` for the field, as in that case our `set_default_values` method would skip setting the default for that field as the user has explicitly passed a value, but that value would then be ignored in the deserializer as it is equal to the default value (https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/deserializer/dxcontent.py#L137-L139).
Because we cannot easily check whether the passed value matches the default value in `set_default_values` (because we have the raw value, not yet deserialized), there were two choices:
- Monkey patch the `DeserializeFromJson.get_schema_data` 
- modify `set_default_values` to persist default values also for  fields for which the user is explicitly setting a value. 

I opted for the latter approach, as I deem it safer (hard to be sure that there is not something similar happening in an other path for creation of objects), and because it avoids overwriting `DeserializeFromJson`.

I have also analysed the situation on dev.onegovgever.ch with a maintenance script (https://github.com/4teamwork/opengever.maintenance/pull/265). There were around 1000 objects with non-persisted default values. The fields concerned are:
```
Missing (by field):
  opengever.base.behaviors.classification.IClassification.classification           271
  opengever.base.behaviors.classification.IClassification.privacy_layer            865
  opengever.base.behaviors.classification.IClassification.public_trial             681
  opengever.document.behaviors.metadata.IDocumentMetadata.document_date            42
  opengever.document.behaviors.metadata.IDocumentMetadata.preserved_as_paper       71
  opengever.task.task.ITask.deadline                                               209

Summary (by object):
Missing: 1086
OK: 12814
```
Of those `document_date` and `deadline` have dynamic default values and hence are problematic. I've therefore added upgrade steps to fix those. Note that for tasks, it might be that the task has gotten reindexed (happens for example when commenting on a task, or opening the edit form). In that case the `deadline` stored on the SQL object and in the catalog, will be the default for the day the task got reindexed. Still this is the value for the deadline users see in the classic UI, and I think is fair to persist that value.

Persisting the defaults for the other fields requires going over all the objects. I propose to do this separately in a nightly job. I will prepare a story for that.

For https://4teamwork.atlassian.net/browse/CA-2125

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-2125.bugfix
+++ b/changes/CA-2125.bugfix
@@ -1,0 +1,1 @@
+Fix persisting default values when creating objects over the API. [njohner]
--- a/opengever/api/tests/test_task.py
+++ b/opengever/api/tests/test_task.py
@@ -1,8 +1,10 @@
+from datetime import date
 from datetime import datetime
 from ftw.builder import Builder
 from ftw.builder import create
 from ftw.testbrowser import browsing
 from ftw.testing import freeze
+from opengever.base.default_values import get_persisted_values_for_obj
 from opengever.base.response import IResponseContainer
 from opengever.core.testing import OPENGEVER_FUNCTIONAL_ACTIVITY_LAYER
 from opengever.ogds.base.Extensions.plugins import activate_request_layer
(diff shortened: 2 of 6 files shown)


deiferni (approved):
LGTM 👍

PR #7043, opened 2021-06-11
Cleanup mail workflow_history for mail created before 2016.

phgross:
In earlier years of GEVER the opengever_mail_workflow had used `state` instead of `review_state` as state_variable. This leads to an KeyError in plone.restapis WorklowInfo service (see https://sentry.4teamwork.ch/organizations/sentry/issues/69441). This PR adds an upgradestep which fixe the workflow_history for those mails.

https://4teamwork.atlassian.net/browse/CA-2095

The bug has been introduced with 0820123893dd8a4abf85aa3ac3cea78793035be9 and fixed with 3cf26ce1edf9bb665d0df6719d95b9aad32d3d53 so we add additional two years to make sure that the fix has been installed in production.

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-2095.bugfix
+++ b/changes/CA-2095.bugfix
@@ -1,0 +1,1 @@
+Cleanup mail workflow_history for mails created before 2016. [phgross]
--- a/opengever/core/upgrades/20210611084449_fix_mail_review_history/upgrade.py
+++ b/opengever/core/upgrades/20210611084449_fix_mail_review_history/upgrade.py
@@ -1,0 +1,49 @@
+from ftw.upgrade import UpgradeStep
+from plone import api
+from Products.CMFCore.WorkflowCore import WorkflowException
+from datetime import datetime
+from DateTime import DateTime
+
+
+
+class FixMailReviewHistory(UpgradeStep):
+    """Fix mail review_history.
+    """
+
+    deferrable = True
+
+    def __call__(self):
+        wftool = api.portal.get_tool('portal_workflow')
+        catalog = api.portal.get_tool('portal_catalog')
+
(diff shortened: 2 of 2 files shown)


changes/CA-2095.bugfix
```diff
@@ -0,0 +1 @@
+Cleanup mail workflow_history for mail created before 2016. [phgross]
```
njohner, line 1: for mail created -> for mailS created

opengever/core/upgrades/20210611084449_fix_mail_review_history/upgrade.py
```diff
@@ -0,0 +1,48 @@
...
+from Products.CMFCore.WorkflowCore import WorkflowException
+from datetime import datetime
+from DateTime import DateTime
+
+
+
+class FixMailReviewHistory(UpgradeStep):
+    """Fix mail review_history.
+    """
+
```
njohner, line 12: We can probably make this one deferrable?

opengever/core/upgrades/20210611084449_fix_mail_review_history/upgrade.py
```diff
@@ -0,0 +1,48 @@
...
+    def __call__(self):
+        self.install_upgrade_profile()
```
njohner, line 14: ✂️

njohner (approved):
In commit message: "which fixe the workflow" -> "which fixeS the workflow"
Otherwise LGTM

opengever/core/upgrades/20210611084449_fix_mail_review_history/upgrade.py
phgross: 👍

PR #7044, opened 2021-06-11
Disable unwrapping of inbound mails.

njohner:
Unwrapping of inbound mails is rather confusing, notably when there are several attachments to an inbound mail, any of which being an E-mail, it will save that attachment into Gever instead of the whole E-mail. As this behavior is not necessary anymore we simply deactivate it for all deployments.

For https://4teamwork.atlassian.net/browse/CA-2375

## Checklist
- [x] Changelog entry

### DIFF
--- a/opengever/policytemplates/policy_template/opengever.+package.name+/opengever/+package.name+/+adminunit.id+/profiles/default/registry.xml.bob
+++ b/opengever/policytemplates/policy_template/opengever.+package.name+/opengever/+package.name+/+adminunit.id+/profiles/default/registry.xml.bob
@@ -9,6 +9,10 @@
     <value key="is_auto_refresh_enabled">{{{setup.bumblebee_auto_refresh}}}</value>
   </records>
 
+  <records interface="ftw.mail.interfaces.IMailSettings">
+    <value key="unwrap_mail">False</value>
+  </records>
+
   <records interface="opengever.base.interfaces.ISearchSettings">
     <value key="use_solr">True</value>
   </records>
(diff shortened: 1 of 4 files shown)


phgross (changes requested):
Changes are looking good, but we should disable this setting also in the gever and teamraum policytemplates, because the setting is True by default.

njohner:
thanks for the feedback. Indeed I forgot the templates!

opengever/policytemplates/policy_template/opengever.+package.name+/opengever/+package.name+/+adminunit.id+/profiles/default/registry.xml.bob
```diff
@@ -9,6 +9,10 @@
     <value key="is_auto_refresh_enabled">{{{setup.bumblebee_auto_refresh}}}</value>
   </records>
 
+  <records interface="ftw.mail.interfaces.IMailSettings">
```
phgross, line 12: Just wonder, is this file also included in teamraum policies?

phgross (approved):
LGTM - just wonder if it works also for new teamraum policies

opengever/policytemplates/policy_template/opengever.+package.name+/opengever/+package.name+/+adminunit.id+/profiles/default/registry.xml.bob
njohner: yes.

PR #7045, opened 2021-06-11
Allow trashing Workspace folders

njohner:
With this pull request we add the possibility to trash workspace folders. I've completely refactored trashing because the whole `Trashable` behavior did not make much sense:
- The `trashed` view was registered on `ITrashableMarker`, but it actually trashed objects recieved through the `paths` parameter.
- We therefore had the `ITrashable` behavior on all sorts of portal types that could not be trashed (basically all containers that could contain documents).
- So `ITrashable` served as two purposes, marking objects that can be trashed and also containers on which the action to trash content should appear (but of course the view would be happy to try to trash any object passed in `paths`, not just the children of that container)
- A behavior is something that should work the same for all content types on which it is registered, whereas `ITrashable` was specifically designed for documents and would simply raise errors for other content types (even for content types on which it was registered)

I therefore decided to reimplement this with adapters instead, allowing to easily customize the trashing for different content types. I also took the chance to somewhat improve the naming used throughout.

Note that I will make a separate PR for deleting trashed workspace folders.

For https://4teamwork.atlassian.net/browse/CA-981

## Checklist

- [x] Changelog entry

- API change:
  - [x] API Changelog entry (see [guide]- New functionality:


opengever/trash/trash.py
```diff
@@ -165,7 +166,8 @@ def is_removed(self):
 
     def is_parent_trashed(self):
         container = aq_parent(aq_inner(self.context))
-        return ITrasher(container).is_trashed()
+        trasher = queryAdapter(container, ITrasher)
+        return trasher and trasher.is_trashed()
```
elioschmutz, line 170: shouldn't the adapter be registered on every context? maybe not on the plone siteroot, but then we can register it there as well, no?

and this change seems to be in the false commit

opengever/document/fileactions.py
```diff
@@ -268,3 +270,21 @@ def _can_edit_with_office_online(self):
         if self.context.is_locked():
             return False
         return True
+
+
+@implementer(IFileActions)
+@adapter(IWorkspaceFolder, Interface)
+class WorkspaceFolderFileActions(object):
```
elioschmutz, line 277: I think we should use `object_buttons` and `folder_buttons` (or `folder_actions`), not `file_actions`. Or what is the purpose of registering `file_actions` for workspace folders?

opengever/base/browser/folder_buttons_availability.py
```diff
@@ -126,7 +128,11 @@ def is_trash_available(self):
         """Trash action should only be shown on dossier level, as we otherwise
         don't know whether the documents can be trashed (they could be in an
         inactive or resolved dossier).
+        The action also does not need to be available on trashed contents.
         """
+        trasher = queryAdapter(self.context, ITrasher)
+        if trasher and trasher.is_trashed():
```
elioschmutz, line 134: same here, why do we need to handle this case?

opengever/base/browser/folder_buttons_availability.py
```diff
@@ -135,7 +141,11 @@ def is_untrash_available(self):
         """Untrash action should only be shown on dossier level, as we otherwise
         don't know whether the documents can be untrashed (they could be in an
         inactive or resolved dossier).
+        Also untrashing content in a trashed container is not allowed.
         """
+        trasher = queryAdapter(self.context, ITrasher)
+        if trasher and trasher.is_trashed():
```
elioschmutz, line 147: and here

elioschmutz (approved):
😍 very clear and easy to review, thanks 👍 !

And I like your refactoring decision 👍

opengever/core/profiles/default/actions.xml
```diff
@@ -456,10 +456,10 @@
       <property name="visible">True</property>
     </object>
 
-    <object name="trashed" meta_type="CMF Action" i18n:domain="opengever.core">
-      <property name="title" i18n:translate="">trashed</property>
+    <object name="trash_content" meta_type="CMF Action" i18n:domain="opengever.core">
```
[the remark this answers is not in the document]
njohner: @all4manu does the frontend part and he knows the actions have changed ;-)

opengever/trash/tests/test_trash.py
```diff
@@ -533,33 +534,47 @@ def test_dossier_cannot_be_trashed(self):
...
+        with self.assertRaises(TrashError) as exc:
             trasher.trash()
+        self.assertEqual('Not trashable', str(exc.exception))
         self.assertFalse(ITrashed.providedBy(obj))
 
     def test_repofolder_cannot_be_trashed(self):
         self.login(self.manager)
         obj = self.empty_repofolder
-        self.assertFalse(ITrashableMarker.providedBy(obj))
+        trasher = ITrasher(obj)
```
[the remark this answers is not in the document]
njohner: Does it make a difference? I've used ITrasher(obj) everywhere

opengever/document/fileactions.py
njohner: The way I see it (but I might be lost in all that mess), is that `object_buttons` is used for actions that get executed on the context, but is also used in the classic UI. `file_actions` is its counterpart for gever-ui only actions. Similarly `folder_buttons` is used for actions that are executed on contained objects (so for listing actions) and `folder_buttons` is its counterpart for gever-ui only actions.
So I'd say we cannot add it to `object_buttons` because it's a frontend only action (although that is debatable I guess). So `file_actions` seems to me is the right place. 
Now to be honest I would like to rename `file_actions` to `context_actions`. I find its implementation with the view delegating to adapters and defaulting to not displaying the actions very practical and it would let itself generalise nicely to all content-types. I did not want to tackle this here though, so I just made a baby step and planned on creating a story to have another look at all our action categories and trying to clean that up.

opengever/trash/trash.py
njohner: The adapter is registered on every content, but not on the plone site root.
This change is in this commit, because the problem arises when getting the actions on workspaces (because then it tries to get the adapter on the plone site root), so it actually fixes an issue that arises with the actions added here.
I'm actually not sure whether registering the adapter for the plone site root would be better?

opengever/base/browser/folder_buttons_availability.py
njohner: again because of the plone site root

opengever/base/browser/folder_buttons_availability.py
njohner: and same here

opengever/trash/tests/test_trash.py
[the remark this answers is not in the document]
elioschmutz: no, not really. so just never mind. It's just the default way to lookup adapters. But here in testing it's ok to use it. I think there are a lot of such tests.

opengever/trash/trash.py
elioschmutz: There is no downside of registering it for the root but it makes it easier to use it because we don't have to check if we there is an adapter or not... so I'd register it on the root. You decide

opengever/document/fileactions.py
elioschmutz: or more explicit to the rest-api: `api_context_actions`? that we know in the future for what it is for? but it's ok if you open a new issue for this 👍

njohner: https://4teamwork.atlassian.net/browse/CA-2438

opengever/trash/trash.py
njohner: Ok I did register it.

PR #7048, opened 2021-06-15
Fix solr indexing of customproperties assigned to a specific dossier_type

phgross:
The customproperty field tried to access the dossier_type field directly on the dossier object. But this is not possible because it is part of the IDossier behavior, wich uses AnnotationStorage. So I introduced an additional parameter attribute_interface for the customproperty field.

Fixes https://4teamwork.atlassian.net/browse/CA-2406

## Checklist

- [x] Changelog entry

- Upgrade steps:

### DIFF
--- a/opengever/propertysheets/field.py
+++ b/opengever/propertysheets/field.py
@@ -49,10 +49,12 @@ class PropertySheetField(Field):
         assignemnt_prefix,
         valid_assignment_slots_factory,
         default_slot,
+        schema_interface=None,
         **kwargs
     ):
         self.request_key = request_key
         self.attribute_name = attribute_name
+        self.schema_interface = schema_interface
         self.assignemnt_prefix = assignemnt_prefix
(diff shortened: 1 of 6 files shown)


opengever/propertysheets/field.py
```diff
@@ -49,10 +49,12 @@ def __init__(
         assignemnt_prefix,
         valid_assignment_slots_factory,
         default_slot,
+        attribute_interface=None,
         **kwargs
     ):
         self.request_key = request_key
         self.attribute_name = attribute_name
+        self.attribute_interface = attribute_interface
```
deiferni, line 57: i'm not a fan of the naming here. What do you think about `schema_interface`? IIRC this is also the term that plone uses for its types.

deiferni (approved):
LGTM 👍  apart from nagging about names

opengever/propertysheets/field.py
phgross: 👍 i agree, thx!

PR #7049, opened 2021-06-15
Support solr stats component in `@solrsearch`.

deiferni:
Add support to return simple statistics for numeric, string and date fields as specified in the solr documentation at
https://solr.apache.org/guide/8_8/the-stats-component.html.

As with the facet endpoint the query is passed more or less directly to solr. To receive the stats component in the response to solrsearch both enabling it with `stats=true` and passing all the the desired fields by repeating the `stats.field<fieldname>` parameter is necessary.

We take care to not return any statistics for the blacklisted fields.

Jira: https://4teamwork.atlassian.net/browse/CA-1695

## Checklist

- [x] Changelog entry

- API change:

### DIFF
--- a/changes/CA-1695.feature
+++ b/changes/CA-1695.feature
@@ -1,0 +1,1 @@
+Support returning results for the solr stats component in the `@solrsearch` endpoint. [deiferni]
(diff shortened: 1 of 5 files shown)


phgross (approved):
👍 LGTM

PR #7051, opened 2021-06-16
Prevent changing the is_private field of existing tasks via API.

phgross:
By registering a specific PATCH service for tasks, which raises if the is_private changes.

Fixes https://4teamwork.atlassian.net/browse/CA-438

## Checklist
- [x] Changelog entry
- [x] API Changelog entry

### DIFF
--- a/opengever/api/task.py
+++ b/opengever/api/task.py
@@ -24,6 +24,8 @@ from plone.restapi.interfaces import IExpandableElement
 from plone.restapi.interfaces import ISerializeToJson
 from plone.restapi.interfaces import ISerializeToJsonSummary
 from plone.restapi.services import Service
+from plone.restapi.services.content.update import ContentPatch
+from zExceptions import BadRequest
 from zExceptions import BadRequest
 from zExceptions import Unauthorized
 from zope.component import adapter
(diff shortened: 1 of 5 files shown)


opengever/api/task.py
```diff
@@ -302,3 +304,17 @@ def add_object_to_context(self):
...
+
+class TaskPatch(ContentPatch):
+    """Specific Patch service for tasks, to prevent changing the
+    is_private field."""
+
+    def reply(self):
+        current_is_private_value = self.context.is_private
+        data = super(TaskPatch, self).reply()
+
+        if self.context.is_private != current_is_private_value:
```
deiferni, line 317: how does the classic UI prevent this? I don't like that with this change we move knowledge about changing objects attributes "far away" from the object into the class responsible for request handling. Do you see a better way to do this?

opengever/api/tests/test_task.py
```diff
@@ -497,6 +497,24 @@ def test_edit_responsible_with_responsible_client(self, browser):
...
         self.assertEqual('james.bond', self.task.responsible)
 
+    @browsing
+    def test_changing_is_private_raise_bad_request(self, browser):
+        self.login(self.regular_user, browser=browser)
+
+        # no change
+        browser.open(self.task, json.dumps({"is_private": False}),
+                     method="PATCH", headers=self.api_headers)
+
```
deiferni, line 507: looks like two newlines. the linter should complain.

deiferni (approved):
As discussed validator or invariants might work, but probably won't. So 👍 .

opengever/api/task.py
phgross: I have checked again briefly the use of a validator or invariant instead, but didn't worked out, we do not have the necessary informations (context, former value, add or edit).

PR #7052, opened 2021-06-16
Fix sending additional files to pdflatex service.

deiferni:
Fix two bugs with the `pdflatex` service when sending additional files:
- Instead of the file content we would just send the path to the file
- We would not send the filename, leading to every file be prefixed with `.file` upon save. This broke references to the file.

I have opted to read the file myself before passing it to requests to be in control of closing the file once it is read. Requests does not take care of this and reading the files in a context manager guarantees that we are in control of closing. I am expecting the files to be relatively small (logos for headers and footers, for example).

Jira: https://4teamwork.atlassian.net/browse/CA-2417

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-2417.bugfix
+++ b/changes/CA-2417.bugfix
@@ -1,0 +1,1 @@
+Fix sending additional files to pdflatex service (e.g. header logos). [deiferni]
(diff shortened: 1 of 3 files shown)


phgross (approved):
👍

PR #7054, opened 2021-06-16
Allow deleting Workspace folders

njohner:
With this pull request we enable deleting workspace folders. 
- To allow clean recursive precondition checks, we refactor deletion to use adapters, so that different object types can easily implement different preconditions for deletion.
- We rename the existing delete actions. `delete_workspace_document` was used both as a `folder_button` and as an `object_button` and as a `folder_action`. It does not make sense to introduce a new `folder_action`, as they are anyway called on selected children of a container, of mixed types. Instead the action was renamed to `delete_workspace_content` and applies now for both documents and workspace folders. Similarly the `delete_workspace_document` folder button was renamed to `delete_workspace_context` as is also available for both workspace documents and folders.
- We also fix deletion of E-mails which was not possible until now.
- We now only allow deleting trashed objects

For https://4teamwork.atlassian.net/browse/CA-981

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:


opengever/workspace/delete.py
```diff
@@ -68,3 +71,17 @@ class WorkspaceMeetingAgendaItemDeleter(BaseWorkspaceContenteDeleter):
...
+
+    permission = 'opengever.workspace: Delete Workspace Folders'
+
+    def verify_may_delete(self, main=True):
+        super(WorkspaceFolderDeleter, self).verify_may_delete(main=main)
+        for obj in self.context.contentValues():
+            deleter = queryAdapter(obj, IDeleter)
+            if deleter is None:
+                raise Forbidden()
+            deleter.verify_may_delete(main=False)
```
elioschmutz, line 96: Instead of introducing the `main` property, couldn't you just move this implementation to the base class and always iterate over the content objects with `contentObjects()` instead `contextValues()`? It should return no objects for non-folderish types... Or as an alternative check for the `IDexterityContainer` interface?

changes/CA-981_3.feature
```diff
@@ -0,0 +1 @@
+Only allow deleting Workspace contents when trashed. [njohner]
```
elioschmutz, line 1: can't you merge this two entries?

opengever/api/tests/test_delete.py
```diff
@@ -1,7 +1,86 @@
...
+        obj = self.disposition
+
+        # Admins have the Delete objects permission on dispositions...
+        obj.manage_permission("Delete objects", roles=[])
+        self.assert_cannot_delete(obj, browser)
+
+        obj.manage_permission("Delete objects", roles=["Administrator"])
+        self.assert_can_delete(obj, browser)
+
+
```
elioschmutz, line 105: are you aware of that there are already similar tests for teamraum subtypes in `opengever/workspace/tests/`?

I.e.:

- https://github.com/4teamwork/opengever.core/blob/master/opengever/workspace/tests/test_todo.py#L249
- https://github.com/4teamwork/opengever.core/blob/master/opengever/workspace/tests/test_workspace_meeting_agenda_item.py#L97
- ...

elioschmutz (approved):
👍

opengever/api/tests/test_delete.py
njohner: Yes I'm aware. My thoughts were as follows:
- tests in `opengever.workspace` test for roles, i.e. what can a `workspace_member` or a `workspace_guest` do. This is important as it is what matters for the users in the end, but it is pretty generic.
- As deletion of content is sensitive, I wanted to have specific tests that the permissions we want are what grants users the right to delete certain content types. Especially important because of the refactoring.
So IMO the tests are different in their philosophy and it's justified to have both sets of tests.

opengever/workspace/delete.py
njohner: Ouh, thanks, I think `objectValues()` is what I wanted to use.
Now concerning the `main` argument. I could implement recursion directly in the base class, but there are certain checks I only want to make on the first level object (in this case the `check_within_workspace`, which does not have to be repeated for every child). So I would still either need the `main` argument, or I would need to introduce a second method, as I've done in the trasher: https://github.com/4teamwork/opengever.core/blob/master/opengever/trash/trash.py#L129-L140
That is why I chose to simply do it only for folders. Also note that for `ToDoList`s I actually don't really want the recursion, because it is forbidden to delete a non-empty ToDoList anyway (happens through an event handler: https://github.com/4teamwork/opengever.core/blob/master/opengever/workspace/subscribers.py#L34-L45)

changes/CA-981_3.feature
njohner: I think it's clearer in two separate entries.
- We allow deleting two new content types that could not be deleted before
- We restricted deletion of certain content types (including documents that could already be deleted before) to trashed objects.
Merging those will only make it more confusing IMO, would give something like `Allow deleting trashed WorkspaceFolders and Mails and restrict deletion of workspace documents to trashed documents.`

njohner: I will change the second one though, to make it more specific (because some workspace contents can be deleted when not trashed, e.g. ToDos)

PR #7055, opened 2021-06-17
Improve error handling when quota limit gets exceeded on API requests.

phgross:
Fixes https://4teamwork.atlassian.net/browse/CA-2247, different from what is described in the JIRA issue, I also corrected the TUS upload, because that is the one that is used in the gever-ui.

## Checklist

- [x] Changelog entry

- API change:
  - [x] API Changelog entry

### DIFF
--- a/opengever/api/tests/test_document.py
+++ b/opengever/api/tests/test_document.py
@@ -1,8 +1,10 @@
 from ftw.testbrowser import browsing
 from opengever.document.interfaces import ICheckinCheckoutManager
 from opengever.document.versioner import Versioner
+from opengever.private.interfaces import IPrivateFolderQuotaSettings
 from opengever.testing import IntegrationTestCase
 from opengever.workspaceclient.interfaces import ILinkedDocuments
+from plone import api
 from plone.namedfile.file import NamedBlobFile
 from plone.uuid.interfaces import IUUID
 from zope.component import getMultiAdapter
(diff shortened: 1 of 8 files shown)


opengever/api/tus.py
```diff
@@ -1,17 +1,32 @@
...
 from zExceptions import Forbidden
 from zope.component import getMultiAdapter
 from zope.i18n import translate
-from zope.publisher.interfaces import IPublishTraverse
 from zope.interface import implementer
+from zope.publisher.interfaces import IPublishTraverse
+
+
+@implementer(IPublishTraverse)
+class GeverUploadPatch(UploadPatch):
```
njohner, line 15: So basically this one is for mails and UploadPatch below is used for documents?

opengever/quota/browser/error.py
```diff
@@ -8,6 +9,9 @@ class ForbiddenByQuotaView(BrowserView):
 
     def __call__(self):
         # this is an error, we must not commit
```
njohner, line 11: comment should be moved down or adapted

opengever/api/tests/test_tus.py
```diff
@@ -136,3 +169,13 @@ def test_cannot_replace_proposal_if_not_docx(self, browser):
...
+
+    @browsing
+    def test_raises_if_quota_is_exceeded(self, browser):
+        self.login(self.regular_user, browser)
+
+        api.portal.set_registry_record(interface=IPrivateFolderQuotaSettings,
+                               name='size_hard_limit', value=1)
+
+        self.assert_tus_upload_fails(
+            self.private_dossier, browser, code=507, reason='Insufficient Storage')
```
njohner, line 181: Can you test that no child was added in `private_dossier`?

opengever/api/tests/test_document.py
```diff
@@ -223,6 +225,19 @@ def test_does_not_allow_mails_as_file(self, browser):
...
             browser.json['message'])
 
+    @browsing
+    def test_raises_if_quota_is_exceeded(self, browser):
+        self.login(self.regular_user, browser)
+
+        api.portal.set_registry_record(interface=IPrivateFolderQuotaSettings,
+                                       name='size_hard_limit', value=1)
+
+        with browser.expect_http_error(code=507, reason='Insufficient Storage'):
```
njohner, line 235: Here too, check that nothing was added to the `private_dossier`

njohner (approved):
LGTM, just would like to be sure that we are not committing anything, as we are actually catching the exception.

opengever/api/tus.py
phgross: Yes actually it does. But GeverUploadPatch applies to all TUS uploads, even if there would be other file types in the future. The document upload TUS inherits from GeverUploadPatch, so it works for Documents as well.

phgross:
@njohner Thx for the review, I added additional `transaction.aboort()` in both exception handlers, to make sure nothing is comitted.

PR #7060, opened 2021-06-21
Support dossier resolving incl. assignment of the filing number via API.

phgross:
With #5558 support for resolving a dossier has been introduced, but support when the filing_number feature is enabled, was not implemented. This PR add this support, used for https://4teamwork.atlassian.net/browse/CA-2090.

My thoughts and decisions:
* The API should use the same schema (including fields, default_values and values) as the z3c.form does. Therefore the Dossier TransitionExtender returns now this schema, when the filing_number feature is enabled.
* So that the new frontend does not have to implement the form logic itself, it must be possible to query the schema, similar to what is already the case in the types endpoint. For this reason, a "worklow-schema" endpoint is now available.

IMHO the ArchiveForm and Archiver should be renamed, because actually it has nothing to do with archiving but with the assignment of filing number. But I'll do that in an separate PR.

## Checklist

- [x] Changelog entry

- API change:


opengever/dossier/archive.py
```diff
@@ -109,14 +108,33 @@ def cancel(self, action):
...
+    def run(self, **kwargs):
+
+        action = kwargs['filing_action']
+        filing_year = kwargs['filing_year']
+        filing_prefix = kwargs['filing_prefix']
+
+        if action == METHOD_RESOLVING_AND_FILING:
+            self.archive(filing_prefix, filing_year)
+
+        if action == METHOD_RESOLVING_EXISTING_FILING:
```
njohner, line 116: if -> elif

opengever/dossier/archive.py
```diff
@@ -109,14 +108,33 @@ def cancel(self, action):
...
+            # archive all with the existing filing number
+            filing_no = IFilingNumber(self.context).filing_no
+            filing_prefix = IDossier(self.context).filing_prefix
+            self.archive(filing_prefix, filing_year, number=filing_no)
+
+        if action == METHOD_RESOLVING:
```
njohner, line 122: if -> elif

opengever/dossier/resolve.py
```diff
@@ -152,76 +161,18 @@ def resolve(self):
...
+            IDossierArchiver(self.context).run(**transition_params)
+
+        self.resolver.resolve(
+            end_date=transition_params.get('dossier_enddate'), **transition_params)
 
     def is_archive_form_needed(self):
         return self.resolver.is_archive_form_needed()
 
 
-class DossierResolutionStatusmessageMixin(object):
```
njohner, line 162: That's probably in the wrong commit

opengever/dossier/tests/test_archiver.py
```diff
@@ -440,3 +442,109 @@ def test_precondition_violation_raises_error_on_archive_form(self, browser):
...
+        self.assertEquals(former_filing_number,
+                          IFilingNumber(self.empty_dossier).filing_no)
+
+    @browsing
+    def test_resolving_and_set_new_filing_number(self, browser):
+        self.login(self.secretariat_user, browser)
+
+        former_filing_number = u'Hauptmandant-Administration-2013-1'
+        IFilingNumber(self.empty_dossier).filing_no = former_filing_number
+        browser.open(self.empty_dossier, view='transition-archive')
```
njohner, line 500: I don't think that line is necessary?

docs/public/dev-manual/api/workflow.rst
```diff
@@ -74,3 +74,9 @@ Endpoint verwendet:
 
    .. literalinclude:: examples/workflow_post.json
       :language: http
+
+
+Workflow-Schema
+~~~~~~~~~~~~~~~
+
+Gewisse Workflow Transitions, wie bspw. der Dossier-Abschluss, erwarten je nach Konfiguration weiter Angaben. Diese Schemas können mit einem Request auf den zusätzlichen API Endpoint ``@workflow-schema`` abgefragt werden. Mit dem Rückgabewert kann ein entsprechendes Formular generiert werden.
```
njohner, line 82: I guess we could use that for some of the task transitions that also require additional data?

njohner (approved):
LGTM 🎉

docs/public/dev-manual/api/workflow.rst
phgross: Yes, thats the point. That's the reason why I implemented in a generic way. I'll open a follow up story, to update the implementation also for tasks.

PR #7063, opened 2021-06-23
Return a placeholder pdf instead of an image if a pdf is not digitally available

tinagerber:
Until now we returned a placeholder image if a pdf was not digitally available.
Therefore, a URL to an svg image was specified as `pdf_url` in the document serialization. Since the PDFViewer in the new frontend can only display PDFs, this has led to an error.

Jira: https://4teamwork.atlassian.net/browse/CA-2380

## Checklist

- [x] Changelog entry


deiferni (approved):
LGTM 👍

this will have no side-effect for the classical ui, rigth?

PR #7066, opened 2021-06-25
Restrict document move via API

tinagerber:
Until now, documents could be moved via API even though they were in a task, for example. This is now restricted.

Jira: https://4teamwork.atlassian.net/browse/CA-2424 and https://4teamwork.atlassian.net/browse/CA-2450

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [x] api-change label added
- New functionality:

### DIFF
--- a/opengever/api/tests/test_move.py
+++ b/opengever/api/tests/test_move.py
@@ -79,3 +79,78 @@ class TestMove(IntegrationTestCase):
             {u'type': u'Forbidden', u'message':
              u'Documents within the repository cannot be moved to the private repository.'},
             browser.json)
+
+    @browsing
+    def test_document_inside_a_task_cannot_be_moved(self, browser):
+        self.login(self.regular_user, browser)
+
+        with browser.expect_http_error(code=403):
+            browser.open(self.dossier, view='/@move',
+                         data=json.dumps({"source": self.taskdocument.absolute_url()}),
+                         method='POST', headers=self.api_headers)
+
+        self.assertEqual({u'type': u'Forbidden', u'message':
+                          u'Documents inside a task cannot be moved.'}, browser.json)
+
+    @browsing
+    def test_mail_inside_a_task_cannot_be_moved(self, browser):
+        self.login(self.regular_user, browser)
+        mail = create(Builder('mail').titled('Good news').within(self.task))
+
+        with browser.expect_http_error(code=403):
+            browser.open(self.dossier, view='/@move',
+                         data=json.dumps({"source": mail.absolute_url()}),
+                         method='POST', headers=self.api_headers)
+
+        self.assertEqual({u'type': u'Forbidden', u'message':
+                          u'Documents inside a task cannot be moved.'}, browser.json)
+
+    @browsing
+    def test_document_inside_a_proposal_cannot_be_moved(self, browser):
+        self.login(self.regular_user, browser)
+
+        with browser.expect_http_error(code=403):
+            browser.open(self.dossier, view='/@move',
+                         data=json.dumps({"source": self.proposaldocument.absolute_url()}),
+                         method='POST', headers=self.api_headers)
+
+        self.assertEqual({u'type': u'Forbidden', u'message':
+                          u'Documents inside a proposal cannot be moved.'}, browser.json)
+
+    @browsing
+    def test_mail_inside_a_proposal_cannot_be_moved(self, browser):
+        self.login(self.regular_user, browser)
+        mail = create(Builder('mail').titled('Good news').within(self.proposal))
+
+        with browser.expect_http_error(code=403):
+            browser.open(self.dossier, view='/@move',
+                         data=json.dumps({"source": mail.absolute_url()}),
+                         method='POST', headers=self.api_headers)
+
+        self.assertEqual({u'type': u'Forbidden', u'message':
+                          u'Documents inside a proposal cannot be moved.'}, browser.json)
+
+    @browsing
+    def test_document_inside_a_closed_dossier_cannot_be_moved(self, browser):
+        self.login(self.regular_user, browser)
+
+        with browser.expect_http_error(code=403):
+            browser.open(self.dossier, view='/@move',
+                         data=json.dumps({"source": self.expired_document.absolute_url()}),
+                         method='POST', headers=self.api_headers)
+
+        self.assertEqual({u'type': u'Forbidden', u'message':
+                          u'Documents inside a closed dossier cannot be moved.'}, browser.json)
+
+    @browsing
+    def test_mail_inside_a_closed_dossier_cannot_be_moved(self, browser):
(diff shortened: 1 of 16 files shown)


opengever/document/move.py
```diff
@@ -0,0 +1,17 @@
...
+from opengever.dossier.dossiertemplate.behaviors import IDossierTemplateMarker
+from opengever.dossier.templatefolder.interfaces import ITemplateFolder
+from zope.component import adapter
+
+
+@adapter(IBaseDocument)
+class DocumentMovabiliyChecker(DefaultMovabilityChecker):
+
+    def validate_movement(self, target):
+        if IDossierTemplateMarker.providedBy(target) or \
```
njohner, line 20: I feel like an equivalent of "is_within_workspace" would be more robust here. So maybe a `is_within_templates` and `is_within_repository`?

opengever/base/interfaces.py
```diff
@@ -454,6 +454,13 @@ def delete(report_uid):
         """
 
 
+class IMovabilityChecker(Interface):
+
+    def validate_movement(target):
+        """Returns an error message if object cannot be moved
```
njohner, line 460: I'm not a big fan of that. returning None when everything is fine and a string when not. This also leads to the fact that if `validate_movement()` is truthy, the movement is invalid and if it is falsy, the movement is valid.
I'd much rather have that `validate_movement` raises an exception when you can't move, maybe a custom `MoveError` or simply a `Forbidden`. This is also easier with inheritance...

opengever/api/tests/test_move.py
```diff
@@ -46,3 +48,78 @@ def test_document_within_repository_cannot_be_moved_to_templates(self, browser):
         self.assertEqual({u'type': u'Forbidden', u'message':
                           u'Documents inside the repository cannot be moved to the templates.'},
                          browser.json)
+
+    @browsing
+    def test_document_insida_a_task_cannot_be_moved(self, browser):
```
njohner, line 53: insida -> inside_a or inside

opengever/api/tests/test_move.py
```diff
@@ -46,3 +48,78 @@ def test_document_within_repository_cannot_be_moved_to_templates(self, browser):
...
+        with browser.expect_http_error(code=403):
+            browser.open(self.dossier, view='/@move',
+                         data=json.dumps({"source": self.taskdocument.absolute_url()}),
+                         method='POST', headers=self.api_headers)
+
+        self.assertEqual({u'type': u'Forbidden', u'message':
+                          u'Documents inside a task cannot be moved.'}, browser.json)
+
+    @browsing
+    def test_mail_insida_a_task_cannot_be_moved(self, browser):
```
njohner, line 65: same here

opengever/api/tests/test_move.py
```diff
@@ -46,3 +48,78 @@ def test_document_within_repository_cannot_be_moved_to_templates(self, browser):
...
+        with browser.expect_http_error(code=403):
+            browser.open(self.dossier, view='/@move',
+                         data=json.dumps({"source": mail.absolute_url()}),
+                         method='POST', headers=self.api_headers)
+
+        self.assertEqual({u'type': u'Forbidden', u'message':
+                          u'Documents inside a task cannot be moved.'}, browser.json)
+
+    @browsing
+    def test_document_insida_a_proposal_cannot_be_moved(self, browser):
```
njohner, line 78: and here

opengever/api/tests/test_move.py
```diff
@@ -46,3 +48,78 @@ def test_document_within_repository_cannot_be_moved_to_templates(self, browser):
...
+        with browser.expect_http_error(code=403):
+            browser.open(self.dossier, view='/@move',
+                         data=json.dumps({"source": self.proposaldocument.absolute_url()}),
+                         method='POST', headers=self.api_headers)
+
+        self.assertEqual({u'type': u'Forbidden', u'message':
+                          u'Documents inside a proposal cannot be moved.'}, browser.json)
+
+    @browsing
+    def test_mail_insida_a_proposal_cannot_be_moved(self, browser):
```
njohner, line 90: and here

opengever/api/tests/test_move.py
```diff
@@ -46,3 +48,78 @@ def test_document_within_repository_cannot_be_moved_to_templates(self, browser):
...
+        with browser.expect_http_error(code=403):
+            browser.open(self.dossier, view='/@move',
+                         data=json.dumps({"source": mail.absolute_url()}),
+                         method='POST', headers=self.api_headers)
+
+        self.assertEqual({u'type': u'Forbidden', u'message':
+                          u'Documents inside a proposal cannot be moved.'}, browser.json)
+
+    @browsing
+    def test_document_insida_a_closed_dossier_cannot_be_moved(self, browser):
```
njohner, line 103: 😄

opengever/api/tests/test_move.py
```diff
@@ -46,3 +48,78 @@ def test_document_within_repository_cannot_be_moved_to_templates(self, browser):
...
+        with browser.expect_http_error(code=403):
+            browser.open(self.dossier, view='/@move',
+                         data=json.dumps({"source": self.expired_document.absolute_url()}),
+                         method='POST', headers=self.api_headers)
+
+        self.assertEqual({u'type': u'Forbidden', u'message':
+                          u'Documents inside a closed dossier cannot be moved.'}, browser.json)
+
+    @browsing
+    def test_mail_insida_a_closed_dossier_cannot_be_moved(self, browser):
```
njohner, line 115: 🦊

njohner (commented):
Hum, it seems there is an `is_movable` method on the document, which is what is used for the classic UI. This also checks whether the document is locked.
Now we have two code paths that do exactly the same. Why not use your adapter also for the classic UI?
On more question, how do we handle not moving documents into the private folder?

tinagerber:
@njohner I have now also implemented that documents cannot be moved to the private root.

It is already taken care that locked documents cannot be moved. See https://github.com/4teamwork/opengever.core/blob/6300999fcedc94be1006b67c0f83f24e01240e07/opengever/locking/tests/test_locks.py#L220

I don't see a useful way the adapter could be used in the classic UI, since we can't use the same error messages if they are to be translated in the new UI.

tinagerber:
@njohner I also added the checks for documents in the inbox.

njohner:
As discussed, I'd still rather have the API and the classic UI sue the same code to test whether something is moveable. Makes it easier to maintain and avoids breaking it in the future. We have several examples where we throw errors in the classic UI but return a message in the API, for example:
- https://github.com/4teamwork/opengever.core/blob/master/opengever/dossier/statusmessage_mixin.py

tinagerber:
@njohner I've adapted it as discussed.

njohner (approved):
🎉 Thanks LGTM! 🎉

PR #7068, opened 2021-06-29
Fix resolving a subdossier when filing_number feature is enabled.

phgross:
When resolving a subdossier via API on a GEVER with enabled filing_number feature, the filing_number information was wrongly enforced. However, this should only be the case when closing a main dossier.

Fixes https://sentry.4teamwork.ch/organizations/sentry/issues/75815/?project=-1&query=is%3Aunresolved+error_log_id%3A1624881505.230.688737328325&statsPeriod=14d

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-2090.bugfix
+++ b/changes/CA-2090.bugfix
@@ -1,0 +1,1 @@
+Fix resolving a subdossier when filing_number feature is enabled. [phgross]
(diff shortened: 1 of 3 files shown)


njohner (approved):
LGTM

PR #7071, opened 2021-07-02
ContactsSource fallbacks to `ogds_user` type if no type is explicitly given.

elioschmutz:
Jira: https://4teamwork.atlassian.net/browse/CA-2492

The `ContactSource` object is used by the `@participation` endpoint if the `contact` feature is enabled. Currently, the source requires an id to lookup in a kind of an actor id with `type:id`. But it's not an actor because there are completely custom types used for the source.

In the frontend, 1. we can't use actors since the api does not return actor-ids consistently and 2. using actors would not work, because the contact source does only need a kind of an actor, not an actor 🙈 .

So to go straight, this PR just assumes that we want to lookup an `ogds_user` if there is no given actor prefix.

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/contact/tests/test_sources.py
+++ b/opengever/contact/tests/test_sources.py
@@ -84,3 +84,15 @@ class TestContactsSource(FunctionalTestCase):
 
         self.assertTerms(
             [(self.meier_ag, u'Meier AG [2222]')], self.source.search('2222'))
+
+    def test_getTermByToken_falls_back_to_ogds_user_if_not_contact_type_is_given(self):
+
+        self.assertEqual(
+            'ogds_user:test_user_1_',
+            self.source.getTermByToken('ogds_user:test_user_1_').token
+        )
+
+        self.assertEqual(
+            'ogds_user:test_user_1_',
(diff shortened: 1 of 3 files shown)


opengever/contact/tests/test_sources.py
```diff
@@ -84,3 +84,15 @@ def test_supports_fuzzy_search(self):
 
         self.assertTerms(
             [(self.meier_ag, u'Meier AG [2222]')], self.source.search('2222'))
+
+    def test_getTermByToken_fallbacks_to_ogds_user_if_not_contact_type_is_given(self):
```
njohner, line 88: The noun "fallback" is indeed one word, but the verb would be "fall back", therefore in that case it would be "falls back" and not "fallbacks"

njohner: The same holds for the commit message

njohner (approved):
LGTM

PR #7072, opened 2021-07-02
Minor optimization of mail attachment code.

njohner:
While working on fixing eml attached to E-mails I found some inefficiencies in the code on the Gever side, that I fix here.

For https://4teamwork.atlassian.net/browse/CA-2375

## Checklist
- [x] Changelog entry

### DIFF
--- a/opengever/mail/browser/mail.py
+++ b/opengever/mail/browser/mail.py
@@ -1,5 +1,4 @@
 from Acquisition import aq_inner
-from ftw.mail import utils
 from ftw.mail.mail import View
 from opengever.base import _ as ogbmf
 from opengever.document import _ as ogdmf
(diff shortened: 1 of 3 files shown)


opengever/mail/browser/mail.py
```diff
@@ -46,7 +45,7 @@ def get_attachment_css_class(self, lookup):
     @instance.memoize
     def attachments(self):
         context = aq_inner(self.context)
-        attachments = utils.get_attachments(context.msg)
+        attachments = context.attachment_infos
```
phgross, line 48: The method is memoized anyway, so the win will not be that big, but still makes sense

phgross (approved):
👍 LGTM

opengever/mail/browser/mail.py
njohner: Yeah I know

PR #7074, opened 2021-07-05
Include information about the current admin_unit in the config endpoint.

phgross:
Used to display current app information in the gever-ui (see https://4teamwork.atlassian.net/browse/CA-2338).

Currently org_units are not really accessible directly via REST, so the `@id` in the endpoint ist not really an URL - but I think thats not a problem for now.

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-2338.feature
+++ b/changes/CA-2338.feature
@@ -1,0 +1,1 @@
+Include information about the curren admin_unit in the config endpoint. [phgross]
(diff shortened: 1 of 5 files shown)


deiferni (approved):
LGTM 👍

phgross:
@tinagerber you're right 😅  ... https://github.com/4teamwork/opengever.core/pull/7079

PR #5580, opened 2019-04-18
Allow members to access plone vocabularies through restapi.

elioschmutz:
Closes #5449

PR #7075, opened 2021-07-05
Allow authenticated users to access vocabularies via API.

phgross:
With PR #5580 the vocabulary endpoint was made accessible for all Members. But for requests on a foreign admin_unit, the user only has the `Authenticated` Role but still need to access vocabularies, for example the `SupportedContentLanguages` vocabulary. 

Opening the endpoint for AuthenticatedUsers is not problematic, because all critical information (users, groups etc.) are visible for the whole client group anyway or their contents are only visible for users with access rights anyway (see #5449 for more information). Furthermore a test ensures that no new vocabularies with sensitive data are made accessible.

For https://4teamwork.atlassian.net/browse/CA-2338

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-2338-2.feature
+++ b/changes/CA-2338-2.feature
@@ -1,0 +1,1 @@
+Allow authenticated users to access vocabularies via API. [phgross]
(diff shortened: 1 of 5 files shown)


njohner (approved):
LGTM. Of course this potentially gives access to some new information to users, notably the list of Keywords that are used on another client where they do not have permissions.
I personally don't see that as problematic though

PR #7076, opened 2021-07-06
Add review_state and include_context parameter to @navigation endpoint.

tinagerber:
In the new UI it should be possible to display only the active dossiers in the dossier structure. However, the current dossier should also be displayed in the dossier structure if it has an inactive review state but only the active dossiers are displayed. This is ensured with the include_context parameter.

Jira: https://4teamwork.atlassian.net/browse/CA-2364

## Checklist

- [x] Changelog entry

- API change:

### DIFF
--- a/opengever/api/navigation.py
+++ b/opengever/api/navigation.py
@@ -70,16 +70,35 @@ class Navigation(object):
                 raise BadRequest("No root found for interface: {}".format(
                     root_interface.__identifier__))
 
-        items = api.content.find(
-            object_provides=content_interfaces,
-            path='/'.join(root.getPhysicalPath()),
-            sort_on='sortable_title',
-        )
+        query = {'object_provides': content_interfaces,
+                 'path': '/'.join(root.getPhysicalPath()),
+                 'sort_on': 'sortable_title'}
+
+        if self.request.form.get('review_state'):
+            query['review_state'] = self.request.form.get('review_state')
+
+        items = api.content.find(**query)
+
+        if self.request.form.get('include_context'):
+            items = self.include_context_branch(items, root.UID(), content_interfaces)
 
         nodes = map(self.brain_to_node, items)
         result['navigation']['tree'] = make_tree_by_url(nodes)
 
         return result
+
+    def include_context_branch(self, items, root_uid, content_interfaces):
+        all_uids = {brain.UID for brain in items}
+        if self.context.UID() in all_uids:
+            return items
+        for item in self.context.aq_chain:
+            item_uid = item.UID()
+            if item_uid == root_uid:
+                break
+            all_uids.add(item_uid)
+        return api.content.find(
+            UID=list(all_uids),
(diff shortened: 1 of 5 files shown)


opengever/api/navigation.py
```diff
@@ -70,17 +70,37 @@ def __call__(self, expand=False):
...
+        if self.context.UID() in all_uids:
+            return items
+        for item in self.context.aq_chain:
+            item_uid = item.UID()
+            if item_uid == root_uid:
+                break
+            all_uids.add(item_uid)
+        return api.content.find(
+            UID=list(all_uids),
+            object_provides=content_interfaces,
```
njohner, line 101: I think you should drop the `object_provides` here, but I guess it's debatable whether the `include_context` flag should take precedence over the `content_interfaces` or not.

opengever/api/navigation.py
```diff
@@ -70,17 +70,37 @@ def __call__(self, expand=False):
...
+        all_uids = {brain.UID for brain in items}
+        if self.context.UID() in all_uids:
+            return items
+        for item in self.context.aq_chain:
+            item_uid = item.UID()
+            if item_uid == root_uid:
+                break
+            all_uids.add(item_uid)
+        return api.content.find(
+            UID=list(all_uids),
```
njohner, line 100: does it have to be cast into a list?

Also concerning the overall approach: I guess that can be a very large list of UIDs here, any idea what the performance of this will be? The endpoint gets called a lot, so it should perform. Another solution would be to only search for the new UIDs and add them into the set of already found items, the only thing you would loose is the sorting on `sortable_title`

njohner (commented):
Overall LGTM.
I do like that your implementation is that simple, but I'm a bit worried about performance. I know we've had issues with querying a list of UIDs, but I think it was from the frontend, so the issue was just that the URL was becoming very long. Still could you make a quick analysis of how long such a query takes compared to the initial query?
Also I think you should allow for more than one `review_state`, otherwise we will be revisiting this in a few months when we realise we cannot show open tasks for example. Or that it will not work with two `content_interfaces`, because each has different review states...

opengever/api/navigation.py
tinagerber: I first implemented it this way, but then discarded it due to the sorting issue, as it looks strange when the order changes when you change the filter.

tinagerber: Ah and I have to cast it into a list, it doesn't work with a set

tinagerber:
@njohner I compared the requests with a dossier structure containing 120 active dossiers and could not find any significant performance differences. In any case, we only use the code block when visiting an inactive dossier and the filter is active.

About the review state: It is already possible to filter  by a list of review states, but I could verify this with a test.

tinagerber:
@njohner I've added another test,  to filter by multiple review states. Can you rereview?

njohner (approved):
LGTM

PR #7077, opened 2021-07-06
Add new @submit-additional-documents endpoint.

njohner:
We add a new `@submit-additional-documents` endpoint.
The endpoint takes as input parameter a list of document UIDs. This avoids the whole problem with dealing with absolute urls and should not be a problem for the frontend, which also has the UIDs from the content selector.
I was a bit more indecisive regarding the response. Because depending on whether a given document was already submitted or not, the document will either be copied to the submitted proposal, updated with a new version or nothing will be done, I deemed it necessary to include the performed action for each document in the response. The response is there fore a list of dictionaries. For each document it returns its URL and the performed action:
```
[{"source" : url, "action": "copied"}]
``` 
The URL is under the "source" key, because we might want to include the "target" (i.e. information about the document on the submitted proposal side) in the future? I wonder whether it would be cleaner and more flexible to return a dictionary under "source" instead, something like:
```
[{"source" : {`@id`:url, `UID`: uid}, "action": "copied"}]
```
Input welcome.

For https://4teamwork.atlassian.net/browse/CA-2036

## Checklist

- [x] Changelog entry
- API change:
- New functionality:

### DIFF
--- a/changes/CA-2036.feature
+++ b/changes/CA-2036.feature
@@ -1,0 +1,1 @@
+Add @submit-additional-documents endpoint. [njohner]
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -18,6 +18,7 @@ Other Changes
 - ``@config`` endpoint extended with current admin_unit information.
 - ``@trigger-task-template``: Support overriding the deadline for each task (see :ref:`trigger_task_template` for updated examples).
 - ``@navigation``: Add ``review_state`` and ``include_context`` parameters (see :ref:`docs <navigation>`)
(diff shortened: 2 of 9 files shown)


phgross (approved):
👍 LGTM _(very clean and easy to read PR 😍 )_ 

I would keep the return value as simple as possible for now, so as you implemented it. Of course, a list of dict in the source key could possibly be a bit more flexible, but I don't really expect any new requests here. Especially since we don't know how it will go on with the RISSPV.

njohner:
@phgross Thanks for the review. I've also talked with @elioschmutz who will be doing the frontend part. He wished for the summary-serialized objects in the response.
I'll leave it open for now, and see with him when he really starts working on it

elioschmutz:
We use the Background-Queue in the frontend with single requests. So your response works fine for what we need it.

PR #7080, opened 2021-07-08
Fix selecting all items for Solr based listings.

deiferni:
The provided fix just fixes the issue for tabbedview/table based listings in the classical UI.

It works around the fact that the Solr based source does not always have all contents loaded to apply batching but uses Solr side batching. This breaks an assumption in the catalog based "select all" functionality which seems to expect an unbatched list of all brains being available.

As a workaround we tell the source that we want all results. The query will be optimized and will only query for the `path` field. I have tested it locally on a dossier with 20k documents. Returning the selected paths to the frontend seems performant enough so that i can bump the hard-coded maximum to 100k.

Jira: https://4teamwork.atlassian.net/browse/CA-2553

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-2553.bugfix
+++ b/changes/CA-2553.bugfix
@@ -1,0 +1,1 @@
+Fix selecting all items for solr based listings in the classical UI. [deiferni]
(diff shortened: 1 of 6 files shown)


njohner (approved):
LGTM

PR #7082, opened 2021-07-09
Provide field unspecific mail download view

phgross:
The view decides independently whether original message or eml file is returned and does no longer require a fieldname as additional url paramenter. This allows the frontend to just call a view and not handle the businesslogic by them self.

For https://4teamwork.atlassian.net/browse/CA-2528 

## Checklist

- [x] Changelog entry


deiferni (approved):
LGTM
LGTM 👍

PR #7067, opened 2021-06-29
Remove document watchers feature flag.

tinagerber:
Document watcher feature flag is no longer needed.

Jira: https://4teamwork.atlassian.net/browse/CA-2462

_Everything has to be done/checked. Checked but not present means the author deemed it unnecessary._

- [x] Changelog entry
- [x] Link to issue (Jira or GitHub) and backlink in issue (Jira)

_Only applicable should be left and checked._

- API change:
  - [x] Documentation is updated
  - [ ] API Changelog entry (see [guide](https://4teamwork.atlassian.net/wiki/spaces/4TEAM/pages/451248812/API+Changelog+Guidelines))

PR #7084, opened 2021-07-12
Fix upgrade step.

tinagerber:
Follow up for https://github.com/4teamwork/opengever.core/pull/7067

Since the feature flag has been removed again, this upgrade step no longer works because the field no longer exists on the IDocumentSettings interface.


deiferni (approved):
LGTM 👍

PR #7085, opened 2021-07-13
Improve role assignment reports creation and display

phgross:
- Return a fix and sorted list of referenced_roles in the @role-assignments endpoint. This allows the frontend to improve the display of a role assignment report.
- Provide a possibility to query a source for all users and groups. Therefore I add a new endpoint @globalsources for field unspecific sources.

For [CA-2361]

## Checklist

- [x] Changelog entry

- API change:

[CA-2361]: https://4teamwork.atlassian.net/browse/CA-2361


docs/public/dev-manual/api/api_changelog.rst
```diff
@@ -11,6 +11,7 @@ Breaking Changes
 
 - ``@move``: Restrict moving of documents via API according to the same rules as in the classic UI.
 - ``@listing``: Add ``sequence_type`` as allowed field (see :ref:`docs <listings>`).
+- ``@role-assignments``: returns now a fix list of roles under the key ``referenced_roles``.
```
deiferni, line 14: Return a fix**ed** list of roles ~~under~~ **at** the key

changes/CA-2361.bugfix
```diff
@@ -0,0 +1 @@
+Return a fix and sorted list of referenced_roles in the @role-assignments endpoint. [phgross]
```
deiferni, line 1: fix**ed**?

opengever/api/role_assignment_reports.py
```diff
@@ -1,5 +1,6 @@
 from opengever.base.interfaces import IRoleAssignmentReportsStorage
 from opengever.ogds.base.actor import Actor
+from opengever.sharing.browser.sharing import ROLE_MAPPING
```
deiferni, line 3: should we move it out of sharing if it is used by multiple packages? our inter package dependencies are messy enough already 😓 .

deiferni (approved):
LGTM 👍  apart from comments.

opengever/api/role_assignment_reports.py
phgross: IMHO the sharing is the right place to be. Actually the role_assignment_report should be in the `sharing` as well, but we do not really have a concept what belongs in the api package and what does not. what do you think?

deiferni: I have started to create api packages in the sub-packages e.g. https://github.com/4teamwork/opengever.core/tree/master/opengever/propertysheets/api. Maybe that's an option here, too?
But we can of course also :shipit: as is if you can't split it now.

phgross: Yes, i think this is a much more reasonable solution, because otherwise we will only have one package api in the long term. 😅 But we should discuss this in general and decide so that we all do it the same way in the future.

I will merge therefore the PR without moving it into an api subdirectory, and we can talk about and decide next Tuesday.

PR #7088, opened 2021-07-15
Remove ftw.tika dependency from the policytemplate.

phgross:
I'm currently removing all the tika dependencies from our saas policies, so the policytemplate should not create new ones 😉 

Part of CA-1213

- [x] Changelog entry

_Removing and uninstalling tika will follow in a separate PR_


deiferni (approved):
LGTM 👍

PR #7090, opened 2021-07-20
Enable cross-client transitions via API for forwardings.

tinagerber:
Changes in this PR:
- Add `@accept-remote-forwarding` endpoint to accept cross client forwardings
- Move refuse transition functionalities to a transition extender for `forwarding-transition-refuse` transition, this way we can refuse forwardings via API
- Add transition_response to POST `@workflow` endpoint, so that we can return the url of the forwarding that will be created when a forwarding is refused. This allows redirecting to the new forwarding after the transition in the new front end.

For [CA-2019]

## Checklist

- [x] Changelog entry

- API change:

[CA-2019]: https://4teamwork.atlassian.net/browse/CA-2019

### DIFF
--- a/docs/public/dev-manual/api/accept_remote_forwarding.rst
+++ b/docs/public/dev-manual/api/accept_remote_forwarding.rst
@@ -1,0 +1,59 @@
+.. _accept-remote-forwarding:
+
+Remote-Weiterleitung akzeptieren
+================================
+
+Der ``@accept-remote-forwarding`` Endpoint auf Dossiers erlaubt es, eine mandantenübergreifende Weiterleitung im Eingangskorb abzulegen. Optional kann die Weiterleitung einem Dossier zugewiesen werden (es wird eine Aufgabenkopie im angegebenen Dossier erstellt).
+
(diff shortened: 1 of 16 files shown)


opengever/api/tests/test_accept_remote_forwarding.py
```diff
@@ -0,0 +1,341 @@
...
+    @browsing
+    def test_invalid_dossier_oguid_raises_bad_reqest(self, browser):
+        self.login(self.secretariat_user, browser)
+        url = '{}/@accept-remote-forwarding'.format(self.inbox.absolute_url())
+
+        with browser.expect_http_error(code=400):
+            browser.open(
+                url, method='POST',
+                data=json.dumps({'forwarding_oguid': Oguid.for_object(self.inbox_forwarding).id,
+                                 'dossier_oguid': 'fd:101010'}),
```
njohner, line 145: So these tests only work because of the order in which the errors are treated in the endpoint. Because here for example the dossier_oguid is invalid, but actually the forwarding_oguid is not remote. But I guess we can't really avoid that...

docs/public/dev-manual/api/accept_remote_forwarding.rst
```diff
@@ -0,0 +1,59 @@
...
+        ],
+        "responsible": {
+          "title": "Ratskanzlei: Boss Hugo (hugo.boss)",
+          "token": "rk:hugo.boss"
+        },
+        "responsible_client": {
+          "title": "Ratskanzlei",
+          "token": "rk"
+        },
+        "review_state": "task-state-in-progress"
```
njohner, line 58: I this now a task because it was accepted in a dossier?

njohner (approved):
LGTM, last commit was a bit hard to judge though...

PR #7091, opened 2021-07-21
Fix storing text when accepting a forwarding.

deiferni:
The transition extender was not registered for that transition so user input text was discarded.

I have also added a quick test for all task and forwarding transitions to make sure we have extenders registered correctly for all other transitions. Luckily this seems to be the case (or not required for one task transition).

For https://4teamwork.atlassian.net/browse/CA-2627

The text is now persisted ans shows up as expected:
Screenshot 2021-07-21 at 14 37 01

## Checklist

- [x] Changelog entry

- New functionality:

### DIFF
--- a/opengever/inbox/tests/test_accept.py
+++ b/opengever/inbox/tests/test_accept.py
@@ -1,5 +1,6 @@
 from ftw.builder import Builder
 from ftw.builder import create
+from opengever.base.response import IResponseContainer
 from opengever.task.browser.accept.utils import accept_forwarding_with_successor
 from opengever.testing import FunctionalTestCase
 from plone.app.testing import TEST_USER_ID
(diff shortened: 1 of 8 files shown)


opengever/inbox/tests/test_accept.py
```diff
@@ -16,6 +17,10 @@ def test_successor_task_is_created_with_information_task_type(self):
...
                             .within(inbox))
         task = accept_forwarding_with_successor(
-            self.portal, forwarding.oguid.id, 'OK, thx.', dossier=dossier)
+            self.portal, forwarding.oguid.id, u'OK, thx.', dossier=dossier)
 
         self.assertEquals('information', task.task_type)
+
+        responses = IResponseContainer(forwarding)
+        self.assertEqual(1, len(responses))
+        self.assertEqual(u'OK, thx.', responses.list()[0].text)
```
tinagerber, line 26: Can you make sure that the answer is also available for the task?

tinagerber (approved):
LGTM

deiferni:
@tinagerber as discussed this morning i am suggesting to ship the behaviour with the changed order. I have not managed to get switching to `forwarding-transition-assign-to-dossier` instead of `forwarding-transition-accept` when we have a dossier selection. It does not work due to an event ordering issue and IMO the change is out of scope for the fix that we are proposing with this PR. I will create a separate issue and we can tacke it when the switch to the new UI is complete. I have created an issue: https://4teamwork.atlassian.net/browse/CA-2642.

opengever/inbox/tests/test_accept.py
deiferni: 👍

tinagerber (approved):

PR #7093, opened 2021-07-23
Always return error message in @trash endpoint if content is not trashable

tinagerber:
Since the error message `The document has been returned as excerpt` was not caught, a `204 No content` was returned in this case.

Now an error message is always returned in case of an error.

For [CA-2472]

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [x] api-change label added

[CA-2472]: https://4teamwork.atlassian.net/browse/CA-2472


phgross (approved):
👍

PR #7094, opened 2021-07-23
Improve policy templates for Teamraum deployments.

njohner:
Use correct defaults for groups assignments for Teamraum deployments and make `invitaiton_group_dn` non-mandatory, is this typically should not be set for on-premise deployments where the `invitaiton_group_dn` is different for TEST and PROD deployments. It will anyway correctly default to the users group, i.e. `tr_users` in Gever (see https://github.com/4teamwork/opengever.core/blob/master/opengever/workspace/participation/browser/my_invitations.py#L93-L117).

For [CA-1885]

## Checklist

- [x] Changelog entry

[CA-1885]: https://4teamwork.atlassian.net/browse/CA-1885

### DIFF
--- a/changes/CA-CA-1885.other
+++ b/changes/CA-CA-1885.other
@@ -1,0 +1,1 @@
+Improve policy templates for Teamraum deployments. [njohner]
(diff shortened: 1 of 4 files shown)


deiferni (approved):
LGTM 👍

PR #7096, opened 2021-07-26
Add excel role assignment-report generation and download view.

phgross:
For [CA-2362](https://4teamwork.atlassian.net/browse/CA-2362)

Uses the already existing XLSReporter. Because the role-assignment report ist not a real object, I did not register a portal_action, instead the action is defined in the frontend.

## Checklist
- [x] Changelog entry

### DIFF
--- a/opengever/base/browser/role_assignment_excel_report.py
+++ b/opengever/base/browser/role_assignment_excel_report.py
@@ -1,0 +1,69 @@
+from opengever.base import _
+from opengever.base.browser.reporting_view import BaseReporterView
+from opengever.base.interfaces import IRoleAssignmentReportsStorage
+from opengever.base.reporter import XLSReporter
+from opengever.sharing.browser.sharing import ROLE_MAPPING
+from plone.app.uuid.utils import uuidToObject
+from zExceptions import BadRequest
+from zExceptions import NotFound
+from zope.interface import implements
+from zope.publisher.interfaces import IPublishTraverse
+
+
+class RoleAssignmentReportExcelDownload(BaseReporterView):
+
+    implements(IPublishTraverse)
+
+    def __init__(self, context, request):
+        super(RoleAssignmentReportExcelDownload, self).__init__(context, request)
+        self.params = []
+
+    def publishTraverse(self, request, name):
+        self.params.append(name)
+        return self
+
+    def __call__(self):
+        storage = IRoleAssignmentReportsStorage(self.context)
+
+        # Expects report_id as path parameter
+        if not len(self.params) == 1:
+            raise NotFound()
+
+        self.report_id = self.params[0]
+        try:
+            report = storage.get(self.report_id)
+        except KeyError:
+            raise BadRequest(u"Invalid report_id '{}'".format(self.report_id))
+
+        items = list(self.prepare_report_for_export(report))
+        reporter = XLSReporter(self.request, self.columns(), items)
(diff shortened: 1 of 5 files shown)


opengever/base/browser/role_assignment_excel_report.py
```diff
@@ -0,0 +1,82 @@
...
+from zope.interface import implements
+from zope.publisher.interfaces import IPublishTraverse
+
+
+def get_title(value):
+    return value.Title()
+
+
+def readable_boolean(value):
+    return
```
deiferni, line 23: I think/hope this is unused. _A better name would be `as_none` 😂 ._

opengever/base/browser/role_assignment_excel_report.py
```diff
@@ -0,0 +1,82 @@
...
+from plone.app.uuid.utils import uuidToObject
+from Products.Five.browser import BrowserView
+from zExceptions import BadRequest
+from zExceptions import NotFound
+from zope.interface import implementer
+from zope.interface import implements
+from zope.publisher.interfaces import IPublishTraverse
+
+
+def get_title(value):
```
deiferni, line 18: I don't really see the purpose of this function as it is just a method call without any additional logic/conversion. It seems unused though.

opengever/base/browser/role_assignment_excel_report.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def publishTraverse(self, request, name):
+        self.params.append(name)
+        return self
+
+    def __call__(self):
+        storage = IRoleAssignmentReportsStorage(self.context)
+
+        # Expects report_id as path parameter
+        if not len(self.params) == 1:
```
deiferni, line 42: simplify please to: `len(params) != 1`

opengever/base/browser/role_assignment_excel_report.py
```diff
@@ -0,0 +1,82 @@
...
+            raise NotFound()
+
+        self.report_id = self.params[0]
+        try:
+            report = storage.get(self.report_id)
+        except KeyError:
+            raise BadRequest("Invalid report_id '{}'".format(self.report_id))
```
deiferni, line 49: Use a unicode string to be guarded against unicode input. Otherwise you can run into a 500 here.

opengever/base/tests/test_role_assignment_excel_report.py
```diff
@@ -0,0 +1,58 @@
...
+
+        self.assertEquals(
+            'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
+            browser.headers['content-type'])
+        self.assertEquals(
+            'attachment; filename="report_0.xlsx"',
+            browser.headers['content-disposition'])
+
+        data = browser.contents
+        with NamedTemporaryFile(delete=False, suffix='.xlsx') as tmpfile:
```
deiferni, line 44: You're not cleaning up the tempfile. Are you sure you need to keep the file?
I think `load_workbook` should load all necessary data into memory and then we don't need access to the file later.
I may be wrong though and in that case please add manual cleanup to remove the file after the test-case. `AddCleanup` should do the trick: https://docs.python.org/3/library/unittest.html#unittest.TestCase.addCleanup.

deiferni (approved):
LGTM 👍 . Some suggestions and minor improvements in comments.

PR #7097, opened 2021-07-26
Fix retrieving lazy initial version for documents.

deiferni:
The initial version is created lazily so documents that have not been touched in any way after adding do not have an initial version. We handle this case by returning the document itself as the initial version.

For https://4teamwork.atlassian.net/browse/CA-2623

## Checklist

- [x] Changelog entry

- API change:


phgross (approved):
👍

PR #7098, opened 2021-07-26
Fix download lazy initial version for documents.

deiferni:
The initial version is created lazily so documents that have not been touched in any way after adding do not have an initial version. We handle this case by retrieving the file version from the document itself.

I've seen that input validation in the view has not been done so nicely and could result in 500 range server side errors. I have amended to turn them into 400 client side errors where possible.

This is a follow-up for https://github.com/4teamwork/opengever.core/pull/7097 which only fixes one part of the issue.

For https://4teamwork.atlassian.net/browse/CA-2623

## Checklist

- [x] Changelog entry

### DIFF
--- a/opengever/base/viewlets/download.py
+++ b/opengever/base/viewlets/download.py
@@ -2,8 +2,10 @@ from opengever.base import _
 from opengever.base.behaviors.utils import set_attachment_content_disposition
 from opengever.document.versioner import Versioner
 from plone.dexterity.primary import PrimaryFieldInfo
+from Products.CMFEditions.interfaces.IArchivist import ArchivistRetrieveError
 from Products.Five import BrowserView
 from Products.statusmessages.interfaces import IStatusMessage
+from zExceptions import BadRequest
 
 
 class DownloadFileVersion(BrowserView):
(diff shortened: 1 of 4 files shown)


opengever/base/viewlets/download.py
```diff
@@ -2,18 +2,36 @@
...
+            raise BadRequest(u'Missing parameter "version_id".')
+        try:
+            version_id = int(version_id)
+        except ValueError:
+            raise BadRequest(u'Invalid version id "{}".'.format(version_id))
 
-        old_obj = Versioner(self.context).retrieve(version_id)
-        primary_field_info = PrimaryFieldInfo(old_obj)
+        versioner = Versioner(self.context)
+        if not versioner.has_initial_version() and version_id == 0:
```
njohner, line 24: I feel like we should handle this directly in the versioner. Would avoid having to deal with this issue in N different places.

njohner (approved):
LGTM, but I wonder whether we could not centralise getting the lazy initial version in the `Versioner`

opengever/base/viewlets/download.py
deiferni: I do agree. But as discussed now is not the time 😅!

PR #6490, opened 2020-06-08
Catch up with special handling for signed/multipart from ftw.mail.

deiferni:
With this PR we add support for special handling of signed/multipart message subparts. It is added to `ftw.mail` in https://github.com/4teamwork/ftw.mail/pull/58.

With this change attachments inside the signed subpart are listed and discovered correctly:

Screenshot 2020-06-08 at 16 56 39

Before one would only see an attachment for the signed subpart, but pythons walk() would not recurse into the subtree:

Screenshot 2020-06-08 at 16 55 23

https://4teamwork.atlassian.net/browse/GEVER-467
https://4teamwork.atlassian.net/browse/GEVER-501

Things still open here:
- [x] Cut a release of https://github.com/4teamwork/ftw.mail/pull/58 and bump version in here once it is approved
- [x] Drop TMP commit

## Checklist (Must have)

_Everything has to be done/checked. Checked but not present means the author deemed it unnecessary._

- [x] Changelog entry
- [ ] Documentation updated (notably for API and deployment)
- [x] Link to issue (Jira or GitHub) and backlink in issue (Jira)

PR #7099, opened 2021-07-27
Update ftw.mail to fix issues with message/rfc822 attachments.

njohner:
We have introduced a few hacks when extracting attachment data from mails in Gever, notably unwrapping `message/rfc822` attachments. It makes sense to have these fixes in `ftw.mail`, so they can be used for the attachment download view, which did currently not handle such attachments correctly.
I also noticed that, because we have overwritten the whole `AttachmentView`, all changes needed to be made twice, both in `ftw.mail` and in `opengever.core` (e.g. https://github.com/4teamwork/ftw.mail/pull/58/commits/13a1c4a3c8becad9e0b2165943b93fadd11487a4 and https://github.com/4teamwork/opengever.core/pull/6490). I decided to extract the necessary functionality in `ftw.mail` to separate methods, so that we could overwrite those specifically in `opengever.core`, to avoid overwriting the whole View.

With these changes (and the bump of `ftw.mail`) we fix several issues:
- Handle mail attached to mail with missing filenames.
- Fix download of mail attached to mail.

For [CA-2375]

## Checklist

- [x] Changelog entry

[CA-2375]: https://4teamwork.atlassian.net/browse/CA-2375


changes/CA-2375.bugfix
```diff
@@ -0,0 +1 @@
+Update ftw.mail to fix issues with message/rfc822 attachments.
```
deiferni, line 1: add your name 😄 ?

opengever/mail/browser/attachment.py
```diff
@@ -1,71 +1,17 @@
...
     """Returns the attachment at the position specified in the request.
     """
 
-    def render(self):
-        """ Copied from ftw.mail.attachment:
-        Just add the contenttype-check (see l.51)
-        """
+    def set_content_type(self, content_type, filename):
+        if content_type == 'application/octet-stream':
+            mtr = getToolByName(self.context, 'mimetypes_registry')
```
deiferni, line 12: use `plone.api` to get tools?

sources.cfg
```diff
@@ -8,6 +8,7 @@ extensions += mr.developer
 development-packages =
   opengever.maintenance
   collective.js.timeago
+  ftw.mail
```
deiferni, line 11: ✂️

sources.cfg
```diff
@@ -18,3 +19,4 @@ auto-checkout = ${buildout:development-packages}
 [branches]
 Products.LDAPUserFolder = ftw
 Products.LDAPMultiPlugins = ftw
+ftw.mail = nj/CA-2375/fix_attached_multipart_eml
```
deiferni, line 22: ✂️

deiferni (approved):
⚠️ you need to remove ftw.mail form sources/auto checkout before merging!

Apart from that and some nagging LGTM 👍.

opengever/mail/browser/attachment.py
njohner: Sure why not...

PR #7101, opened 2021-07-29
Bump ftw.solr to version 2.9.4 to improve listing performance.

njohner:
In the `@listing` endpoint, we were fetching the objects each time an item had a missing value for a field, because those are not returned by solr so that the attribute was missing on the SolrDocument, leading to trying to get the attribute from the Object instead. This is highly inefficient when displaying columns that are mostly empty. See https://github.com/4teamwork/ftw.solr/pull/182 for a detailed explanation.

I've created a follow-up issue to fix the `completed` field: https://4teamwork.atlassian.net/browse/CA-2675

For [CA-2654]

## Checklist
- [x] Changelog entry

[CA-2654]: https://4teamwork.atlassian.net/browse/CA-2654


phgross (approved):
LGTM 👍

PR #7102, opened 2021-08-02
Add API endpoint to unlink a workspace from a dossier.

phgross:
I made the following decisions:
- Becuase the "gever-workspace-link" ist not really traversable, I decided against a `DELETE` endpoint and implemented a `@unlink-workspace` POST endpoint.
- We don't remove the linked document, so if a workspcae gets unlinked and linked again, the linked documents can be copied as new version to the GEVER dossier. But existing locks will always be unlocked, when unlinking a workspace.
- Because a GEVER dossier can have multiple linked workspaces, we need to know which documents belongs to which workspace, to make sure only those document-locks get unlocked. Therefore I added a new catalog index and metadata `gever_doc_uid` on the workspace side, which stores the UID of the linked gever document. This allows GEVER with a request on the workspace an a query on the own catalog, to fetch all gever-documents linked with a specific workspace.

For [CA-2392](https://4teamwork.atlassian.net/browse/CA-2392)

## Checklist

- [x] Changelog entry

- API change:
  - [x] API Changelog entry (see [guide]- Upgrade steps (changes in profile):
 - Upgradestep
- New translations

### DIFF
--- a/opengever/api/linked_workspaces.py
+++ b/opengever/api/linked_workspaces.py
@@ -321,3 +321,16 @@ class CopyDocumentFromWorkspacePost(LinkedWorkspacesService):
             raise BadRequest("Property 'document_uid' is required")
         as_new_version = bool(data.get('as_new_version', False))
         return workspace_uid, document_uid, as_new_version
+
+
+class ListLinkedDocumentUIDsFromWorkspace(Service):
+
+    def reply(self):
+        catalog = api.portal.get_tool('portal_catalog')
+        brains = catalog(path='/'.join(self.context.getPhysicalPath()),
+                         object_provides=IBaseDocument.__identifier__)
+
+        uids = [brain.gever_doc_uid for brain in brains
+                if brain.gever_doc_uid]
+
+        return {'gever_doc_uids': uids}
--- a/opengever/api/linked_workspaces.py
+++ b/opengever/api/linked_workspaces.py
@@ -185,6 +185,27 @@ class LinkToWorkspacePost(LinkedWorkspacesService):
         return self.reply_no_content()
 
 
+class UnlinkWorkspacePost(LinkedWorkspacesService):
+    """API Endpoint to unlink a dossier from an existing workspace.
+    """
+
+    def render(self):
+        if not self.context.is_open():
+            raise Unauthorized
+        return super(UnlinkWorkspacePost, self).render()
+
+    @request_error_handler
+    def reply(self):
+        alsoProvides(self.request, IDisableCSRFProtection)
+        data = json_body(self.request)
+        workspace_uid = data.get('workspace_uid')
+        if not workspace_uid:
+            raise BadRequest("Property 'workspace_uid' is required")
+
+        ILinkedWorkspaces(self.context).unlink_workspace(workspace_uid)
+        return self.reply_no_content()
+
+
 class CopyDocumentToWorkspacePost(LinkedWorkspacesService):
     """API Endpoint to copy a document to a linked workspace.
     """
--- a/opengever/workspaceclient/linked_workspaces.py
+++ b/opengever/workspaceclient/linked_workspaces.py
@@ -22,10 +22,10 @@ from plone.locking.interfaces import ILockable
 from plone.memoize import ram
 from plone.restapi.interfaces import ISerializeToJson
 from time import time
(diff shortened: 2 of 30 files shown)


opengever/api/linked_workspaces.py
```diff
@@ -185,6 +185,27 @@ def reply(self):
         return self.reply_no_content()
 
 
+class UnlinkWorkspacePost(LinkedWorkspacesService):
+    """API Endpoint to link a dossier to an existing workspace.
```
njohner, line 189: unlink a dossier from an existing workspace

opengever/api/linked_workspaces.py
```diff
@@ -321,3 +321,16 @@ def validate_data(self, data):
             raise BadRequest("Property 'document_uid' is required")
         as_new_version = bool(data.get('as_new_version', False))
         return workspace_uid, document_uid, as_new_version
+
+
+class ListLinkedDocumentUIDsFromWorkspace(Service):
```
njohner, line 326: Looking at this I think this endpoint is not necessary, but maybe I misunderstand something. See below

opengever/workspaceclient/linked_workspaces.py
```diff
@@ -409,6 +414,20 @@ def list_documents_in_linked_workspace(self, workspace_uid, **kwargs):
             metadata_fields=["UID", "filename", "checked_out"],
             **kwargs)
 
+    def get_documents_linked_with_workspace(self, workspace_url):
+        """Returns a list of gever documents wich are linked with
+        the given workspace.
+        """
+
```
njohner, line 421: couldn't we simply use `self.list_documents_in_linked_workspace`, adding `gever_doc_uids` in the `metadata_fields`?

opengever/workspaceclient/linked_workspaces.py
```diff
@@ -409,6 +414,20 @@ def list_documents_in_linked_workspace(self, workspace_uid, **kwargs):
...
+        result = self.client.request.get(
+            '{}/@list-linked-gever-documents-uids'.format(workspace_url))
+        uids = result.json().get('gever_doc_uids', [])
+
+        catalog = api.portal.get_tool('portal_catalog')
+        brains = catalog(path='/'.join(self.context.getPhysicalPath()),
+                         UID=uids, full_objects=True)
+        return [brain.getObject() for brain in brains]
```
njohner, line 429: How come you need to `getObject` here if you requested `fill_objects=True` above?

opengever/workspaceclient/linked_workspaces.py
njohner: or simply `self.client.search`

opengever/workspaceclient/tests/test_linked_workspaces.py
```diff
@@ -954,3 +954,21 @@ def test_unlink_is_journalized(self):
 
             self.assert_journal_entry(self.dossier, 'Unlinked workspace',
                                       u'Unlink workspace Ein Teamraum.')
+
+    def test_unlink_workspace_unlocks_linked_documents(self):
```
njohner, line 958: Can you test with two linked workspaces, to make sure it only unlocks the documents from the workspace being unlocked?

njohner (commented):
LGTM, just some doubts about the necessity of the `list-linked-gever-documents-uids` endpoint. Maybe we can have a quick call...

opengever/workspaceclient/linked_workspaces.py
njohner: As discussed, this is OK. This avoids getting all the documents from teamraum and then filtering on the gever side. Instead we return only the ones linked. Also requesting only the ones linked to gever documents from the Gever dossier (by passing all the UIDs) would not work, because the `@search` endpoint only supports `GET`

njohner (approved):

opengever/workspaceclient/linked_workspaces.py
phgross: I also doubled checked plone.restapi's `@querystring-search` endpoint, this would though allow a query by a POST request and passing all the UIDs, but brings some other issues, the metadata_fields is not support, limiting by the workspace path would be a bit inconvenient and we would need to handle batching on the GEVER side.

PR #7103, opened 2021-08-02
Update sharing permissions

njohner:
While preparing the group migration for HBA, I saw that the sharing permission were, IMO, not correct for several workflows. We therefore:
- Fix sharing permissions on private and inbox documents. In the shadow state, the owner had the permission to delegate roles. This should not be a problem, as documents are really only transiently
in the shadow state.
- Fix sharing permissions on `tasktemplatefolder`s. Sharing permissions were not defined on the `tasktemplatefolder` so
that they would be inherited from the folder. This allows role managers to manage permissions on `tasktemplatefolder`s and `tasktemplate`s, which is not correct.
- Fix sharing permissions on `Inbox`es. No reason that administrators should be allowed to delegate roles
on the inboxes.

For [CA-2342]

## Checklist
- [x] Changelog entry

[CA-2342]: https://4teamwork.atlassian.net/browse/CA-2342


phgross (approved):
👍 LGTM

PR #7105, opened 2021-08-02
Follow-up: Update sharing permissions

njohner:
Sorry in https://github.com/4teamwork/opengever.core/pull/7103 I forgot to fix the disposition and proposal workflows. I don't think it is on purpose that "Role managers" can manage permissions on Dispositions and Proposals.
@phgross what do you think?

For [CA-2342]

## Checklist

- [ ] Changelog entry -> no changelog, because it's a follow-up

[CA-2342]: https://4teamwork.atlassian.net/browse/CA-2342


phgross (commented):
I think you need to touch the upgradestep, because the existing one is already installed on our dev system.

Besides that, it looks good.

njohner:
Right thanks!

njohner:
All done

phgross (approved):

PR #7106, opened 2021-08-03
Fix returning translated title in solrsearch and listing endpoints.

njohner:
We fix returning the translated title in the `listing` and `solrsearch` endpoints. The check that we were making to decide whether to return a translated title was never passing, as `OGSolrContentListingObject` does not support `providedBy`.

For [CA-2649]

## Checklist
- [x] Changelog entry

[CA-2649]: https://4teamwork.atlassian.net/browse/CA-2649

### DIFF
--- a/opengever/api/solr_query_service.py
+++ b/opengever/api/solr_query_service.py
@@ -5,6 +5,7 @@ from ftw.solr.converters import to_iso8601
 from ftw.solr.interfaces import ISolrSearch
 from ftw.solr.query import escape
 from opengever.base.behaviors.translated_title import ITranslatedTitleSupport
+from opengever.base.behaviors.translated_title import TRANSLATED_TITLE_PORTAL_TYPES
 from opengever.base.helpers import display_name
 from opengever.base.solr import OGSolrContentListing
 from opengever.base.utils import get_preferred_language_code
@@ -102,9 +103,8 @@ def filename(obj):
 
 
 def translated_title(obj):
-    if ITranslatedTitleSupport.providedBy(obj):
-        attr = 'title_{}'.format(get_preferred_language_code())
-        return getattr(obj, attr, obj.Title())
+    if obj.portal_type in TRANSLATED_TITLE_PORTAL_TYPES:
+        return obj.getObject().Title(language=get_preferred_language_code())
     else:
         return obj.Title()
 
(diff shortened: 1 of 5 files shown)


opengever/api/solr_query_service.py
```diff
@@ -102,9 +102,8 @@ def filename(obj):
 
 
 def translated_title(obj):
-    if ITranslatedTitleSupport.providedBy(obj):
-        attr = 'title_{}'.format(get_preferred_language_code())
-        return getattr(obj, attr, obj.Title())
+    if ITranslatedTitleSupport.__identifier__ in obj.object_provides:
+        return obj.getObject().Title(language=get_preferred_language_code())
```
tinagerber, line 106: Why do we now get the object when the title is translatable? That not good for performance

opengever/api/tests/test_listing.py
```diff
@@ -185,6 +185,26 @@ class TestListingWithRealSolr(SolrIntegrationTestCase):
 
     maxDiff = None
 
+    @browsing
+    def test_repository_folders_translatable_title(self, browser):
+        self.enable_languages()
+
+        self.login(self.regular_user, browser=browser)
+        view = '@listing?name=repository_folders&columns=title&columns=title'
```
tinagerber, line 193:
```suggestion
        view = '@listing?name=repository_folders&columns:list=title'

File: opengever/api/solr_query_service.py (hunk 1 above)
Comment (lukasgraf) [reply opengever/api/solr_query_service.py:106]: Good catch!

`title_de` and `title_fr` are not currently indexed in Solr, so they need to be fetched from the actual object. In the future we should index them, but for repository folders we'll need to take care to not index their raw values (`Umwelt`), but the composed ones prefixed with reference number (`7.3 Umwelt`).

Because objects with `ITranslatedTitle` are always structural, they will already be in the cache in most cases. Also, there's rather few of them compared to other objects, so I think the fallback to the object should be ok in this case.

File: opengever/api/solr_query_service.py [hunk 3]
```diff
@@ -336,7 +335,8 @@ def hide_facet(self, facet):
     ListingField('task_type', 'task_type', accessor=translated_task_type, transform=translate_task_type),
     ListingField('thumbnail_url', None, 'preview_image_url', DEFAULT_SORT_INDEX,
                  additional_required_fields=['bumblebee_checksum', 'path']),
-    ListingField('title', 'Title', translated_title, 'sortable_title'),
+    ListingField('title', 'Title', translated_title, 'sortable_title',
+                 additional_required_fields=['object_provides']),
```
lukasgraf, line 339: As discussed, I wouldn't always request `object_provides`. The list of provided interfaces is *huge* for most objects, and can easily make up 70-80% of a solr document if otherwise few fields were requests.

Instead, we should be able to do with check with a static list of portal types, and write a test that makes sure the list is up-to-date by looking up all FTIs and their behaviors.

lukasgraf (changes requested):

opengever/api/solr_query_service.py
njohner: And to complement, we were already fetching the object before (there is no other way to get to the translated titles). It was just not obvious that we were fetching the object, because there is a fallback to getting attributes from the underlying object in `OpengeverCatalogContentListingObject`. I probably broke this with https://github.com/4teamwork/ftw.solr/pull/182, which tries to avoid that fallback whenever possible.

njohner:
Ok I modified the implementation to use a static list of portal types that have the translated_title behavior and added a test to make sure the list is up-to-date.
Can someone give it another look?

lukasgraf (approved):
LGTM! 👍

PR #7108, opened 2021-08-04
Allow deleting repository folders over the REST-API.

njohner:
We enable deletion of empty repository folder over the REST-API.

For [CA-2108]

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [ ] api-change label added

[CA-2108]: https://4teamwork.atlassian.net/browse/CA-2108

### DIFF
--- a/opengever/api/tests/test_delete.py
+++ b/opengever/api/tests/test_delete.py
@@ -85,6 +85,33 @@ class DeleteGeverObjects(IntegrationTestCase, APITestDeleteMixin):
         obj.manage_permission("Delete objects", roles=["Records Manager"])
         self.assert_can_delete(obj, browser)
 
+    @browsing
+    def test_deleting_repository_root_is_not_allowed(self, browser):
+        self.login(self.manager, browser)
+        self.assert_cannot_delete(self.repository_root, browser, code=403)
+        self.assertEqual(
+            u'Deleting repository folders and roots is not allowed',
+            browser.json['message'])
+
+    @browsing
+    def test_deleting_repository_folder_requires_delete_objects_permission(self, browser):
+        self.login(self.records_manager, browser)
+        obj = self.empty_repofolder
+
+        self.assert_cannot_delete(obj, browser, code=403)
+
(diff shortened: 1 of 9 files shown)


opengever/api/tests/test_delete.py
```diff
@@ -85,6 +85,33 @@ def test_deleting_disposition_requires_delete_objects_permission(self, browser):
...
+            browser.json['message'])
+
+    @browsing
+    def test_deleting_repository_folder_requires_delete_objects_permission(self, browser):
+        self.login(self.records_manager, browser)
+        obj = self.empty_repofolder
+
+        self.assert_cannot_delete(obj, browser, code=403)
+
+        obj.manage_permission("Delete objects", roles=["Records Manager"])
```
phgross, line 103: I would prefer a `self.login(self.administrator, browser)` here.

phgross (approved):
👍 LGTM 

I just wonder if we should raise a Unauthorized instead of an BadRequest, but I'm not sure. what were your thoughts?

opengever/api/tests/test_delete.py
njohner: hum, all the tests for deletion are like this (yeah well I wrote them all :-D). My thinking is that I really want to test that it's the "Delete objects" permissions that is required. Logging in with another user will give my many more / different permissions, so doesn't quite test what I wanted to test.

njohner:
As for `Unauthorized` instead of BadRequest, I did think about that. 
It would need to be `Forbidden` and not `Unauthorized` though. `Unauthorized` means that you are not authenticated and that you might be allowed to do that action if you were, so you should initiate authentication. `Forbidden` means that the server refuses to do the action, re-authentication would make no difference.

Apart from that I wasn't sure what is better. The action will not be available for repository folders that are not empty, so normally the client should not try to delete such an object. That's why I thought it's basically a client error to try that request, hence `BadRequest`. Thinking about it now, I guess `Forbidden` is better.

PR #7109, opened 2021-08-05
Fix permission issue when unlinking a workspace.

phgross:
Follow up for PR #7102  Make API endpoint `@list-linked-gever-documents-uids` accessible for all with view permission. GEVER Users does not have the role `WorkspaceClientUser` on the teamraum side in a regular GEVER <-> temraum setup, but only in the development setup. Also the endpoint queries the UIDs now unrestricted to make sure every linked documents gets unlocked.

If the gever user does not have permission on all linked gever documents, the method raises an `BadRequest`, so unlinking is not possible for this user.

For [CA-2392]

- Changelog already added in the previous commit.

[CA-2392]: https://4teamwork.atlassian.net/browse/CA-2392

### DIFF
--- a/opengever/workspaceclient/linked_workspaces.py
+++ b/opengever/workspaceclient/linked_workspaces.py
@@ -1,3 +1,4 @@
+from AccessControl.unauthorized import Unauthorized
 from opengever.api.add import GeverFolderPost
 from opengever.base.interfaces import IDuringContentCreation
 from opengever.base.oguid import Oguid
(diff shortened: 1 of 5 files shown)


njohner (commented):
Yes the fix LGTM. But I think we need to give a bit more thought to the whole permission issue here. For example I think we need to use `unrestrictedSearchResults` on the teamraum side (https://github.com/4teamwork/opengever.core/blob/master/opengever/api/linked_workspaces.py#L328-L336), security does not matter, as the user does not access the objects there, but only retrieves UIDs of the documents in the GeverSide.
Then on the Gever side, we probably need to check that there are no documents that are linked which the user cannot see? Otherwise those would remain locked. In that case, I think the action should fail. (so here: https://github.com/4teamwork/opengever.core/blob/master/opengever/workspaceclient/linked_workspaces.py#L427-L429 compare the results for restricted and unrestricted search?)

phgross:
@njohner I updated my PR could you take another look?

opengever/workspaceclient/linked_workspaces.py
```diff
@@ -179,7 +180,13 @@ def unlink_workspace(self, workspace_uid):
         workspace = self.client.unlink_workspace(workspace_uid)
 
         # cleanup document locks
-        docs = self.get_documents_linked_with_workspace(workspace['@id'])
+        try:
+            docs = self.get_documents_linked_with_workspace(workspace['@id'])
+        except Unauthorized:
+            raise BadRequest(
+                'Your not allowed to access and unlock all linked '
```
njohner, line 187: Your -> you're / you are

njohner (approved):
Apart from comment, LGTM

PR #7110, opened 2021-08-05
Add document approvals storage and API endpoints.

phgross:
This PR introduces document approvals. The PR adds storage, ApprovalList adapter to manage approvals and API Endpoint additions: The @version endpoints includes now also approvals and the GET Request provides an `approval` expansion.

The feature is currently not used anywhere, but only in the tests. The integration will follow in later stories.
For [CA-2266]

## Checklist

- [x] Changelog entry

- API change:
  - [ ] API Changelog entry - _I did not add a changelog entry yet, so we can modify the endpoint in later PRs if necessary._

[CA-2266]: https://4teamwork.atlassian.net/browse/CA-2266


opengever/document/approvals.py
```diff
@@ -62,6 +62,22 @@ def _initialize_storage(self, create_if_missing=False):
 
         self._storage = ann.get(self.ANNOTATIONS_KEY)
 
+    def remove_all_except(self, except_version):
+        """Removes all approval from the document except the ones for
+        the given version."""
+
+        if not self._storage:
+            return []
```
njohner, line 70: no need to return an empty list... just `return`

opengever/document/approvals.py
```diff
@@ -62,6 +62,22 @@ def _initialize_storage(self, create_if_missing=False):
...
+
+        except_approvals = PersistentList([
+            approval for approval in self._storage
+            if approval.get('version_id') == except_version])
+
+        IAnnotations(self.context)[self.ANNOTATIONS_KEY] = except_approvals
+
+    def remove_all(self):
+        IAnnotations(self.context)[self.ANNOTATIONS_KEY] = PersistentList()
```
njohner, line 79: Why not `IAnnotations(self.context).pop(self.ANNOTATIONS_KEY)`
And dont' you have to check whether the storage exists?

opengever/document/subscribers.py
```diff
@@ -32,3 +35,17 @@ def set_copyname(doc, event):
...
         translate(_('copy_of', default="copy of"), context=request),
         doc.title)
+
+
+def cleanup_document_approvals(doc, event):
+    """When a document gets copied, only the current version approvals
+    should be copied."""
+
+    try:
+        current_version = Versioner(event.original).get_current_version_id()
```
njohner, line 45: What happens for a document with no initial version. Can that document have been approved? If so I guess you should use `missing_as_zero=True` here and probably be careful at other places

opengever/document/tests/test_approvals.py
```diff
@@ -58,3 +60,32 @@ def test_current_datetime_is_default_approval_time(self):
...
+        self.assertEqual(4, len(approvals.get()))
+
+        copy = api.content.copy(self.document, target=self.subdossier)
+
+        self.assertEqual(2, len(IApprovalList(copy).get()))
+        self.assertEqual(
+            [{'task_uid': IUUID(self.task),
+              'approver': self.regular_user.id,
+              'approved': datetime(2021, 7, 2),
+              'version_id': 2},
```
njohner, line 86: Do you need to change the `version_id` after copying? I guess the copied document has version_id 0 no?

njohner (commented):
LGTM just a few comments. Main point is what happens with documents without initial versions?

opengever/document/subscribers.py
phgross: You're right, thx.

opengever/document/tests/test_approvals.py
phgross: Indeed, thx.

phgross:
@njohner thx for your review! I updated my PR.

For documents without an initial version, the approvals are still handled for version 0. i think this is correct. Because a document without a version is in fact in version 0 and it is actually only a technical detail that this does not really exist yet. Do you see other problems/issues? 🤔

opengever/api/tests/test_document.py
```diff
@@ -187,15 +187,28 @@ def test_approvals_expansion(self, browser):
...
-                           u'version_id': 1}],
-                         browser.json['@components']['approvals'])
+        self.assertEqual(
+            [{u'approved': u'2021-07-02T00:00:00',
+              u'approver': u'kathi.barfuss',
+              u'task': {
+                  u'@id': u'http://nohost/plone/ordnungssystem/fuhrung/vertrage-und-vereinbarungen/dossier-1/task-1/task-2',
+                  u'@type': u'opengever.task.task',
+                  u'description': u'',
+                  u'is_leafnode': None,
```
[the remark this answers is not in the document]
phgross: Will fix that in a separate PR.

opengever/document/approvals.py
```diff
@@ -0,0 +1,131 @@
...
+            data[version_id].append(approval)
+
+        return data
+
+    def cleanup_copied_approvals(self, current_version):
+        """Remove approvals from previous versions and reset the version_id
+        for the current version.
+        """
+        if not current_version:
+            self.storage.remove_all()
```
njohner, line 128: return

opengever/document/approvals.py
```diff
@@ -0,0 +1,131 @@
...
+        if not approver:
+            approver = api.user.get_current().id
+
+        if not approved:
+            approved = datetime.now()
+
+        self.storage.add(approved, approver, IUUID(task), version_id)
+
+    def get(self):
+        return [Approval(**item) for item in self.storage.list()]
```
njohner, line 109: `return [approval for approval in self]` ? Maybe even `return list(self)` would work😅

njohner (approved):
LGTM thanks! There is one more concern with permissions IMO.

For documents without initial version I cannot judge whether everything works as expected, but I would expect:
- Approving a document without initial version should store `version 0` in the approval list
- When copying a document without initial version but an approval, that approval should be copied with
- Serialization of a document without initial version but an approval should return that approval as `version 0`

opengever/api/serializer.py
```diff
@@ -178,6 +180,20 @@ def oguid_converter(value):
     return value.id
 
 
+@adapter(Approval)
+@implementer(IJsonCompatible)
+def approval_converter(approval):
+    task = getMultiAdapter((approval.get_task_brain(), getRequest()),
```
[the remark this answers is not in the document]
phgross: I updated my PR, the serializer returns None on the `task` key, if the task is not accessible, see last commit.

phgross:
@njohner I updated my PR. Added a test for documents without a version and also handle the "permission-problem" if the task is not accessible in the serializer.

njohner (approved):
LGTM thanks!

njohner:
I'm merging this one, as philippe is on holidays...

PR #7111, opened 2021-08-06
Clarify changelog conventions for backports.

njohner:
Every time I do a backport I'm annoyed with our lack of convention concerning backports. I think there is only one way of doing this correctly, which is having dupplicate entries and consider every PATCH version to not be included in the following MAJOR and MINOR versions. This is the only way to reflect the reality of how we work, i.e. with diverging branches for PATCH versions...

For [CA-2690]

## Checklist

- [ ] Changelog entry -> nope

[CA-2690]: https://4teamwork.atlassian.net/browse/CA-2690

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,6 +1,8 @@
 Changelog
 =========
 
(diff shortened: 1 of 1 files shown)


phgross (approved):
👍 i agree

lukasgraf (approved):
I agree as well 👍 - I think this is the only sane way.

deiferni (approved):
Yay 👍 .

PR #7112, opened 2021-08-06
Fix update of favorites when trashing / deleting an object.

njohner:
When an object is deleted or trashed, the associated favorites get deleted. This led to the position of the favorites for the concerned users to not be a gapless sequence from 0 to n, which in turn breaks reordering of the favorites.
This is fixed here by ensuring that favorite positions get updated when an object is trashed or deleted. We also add an upgrade step to fix broken `Favorite.position` sequences.
Note that I made the upgrade step a `SchemaMigration` so that it only happens once.

For [CA-2468]

## Checklist
- [x] Changelog entry

[CA-2468]: https://4teamwork.atlassian.net/browse/CA-2468

### DIFF
--- a/changes/CA-2468.bugfix
+++ b/changes/CA-2468.bugfix
@@ -1,0 +1,1 @@
+Favorite positions get updated correctly when trashing / deleting an object. [njohner]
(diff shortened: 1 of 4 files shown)


tinagerber (approved):
LGTM

PR #7113, opened 2021-08-09
Fix setting reminder over accept form.

njohner:
Setting a reminder over the accept form was not working correctly, because the reminder was set after the changes to the task were synced (which happens on the `IObjectModifiedEvent`: https://github.com/4teamwork/opengever.core/blob/master/opengever/globalindex/handlers/configure.zcml#L9-L13). This could lead to the notification not being generated if the task wasn't modified, and hence synced again, before the reminder day...
Because this might be an issue in other situations as well, we sync the reminders synchronously every time a reminder is set.

For [CA-2683]

## Checklist

- [x] Changelog entry

[CA-2683]: https://4teamwork.atlassian.net/browse/CA-2683

### DIFF
--- a/changes/CA-2683.bugfix
+++ b/changes/CA-2683.bugfix
@@ -1,0 +1,1 @@
+Fix setting reminder over accept form. [njohner]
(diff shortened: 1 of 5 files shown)


lukasgraf (approved):
LGTM 👍

PR #7116, opened 2021-08-11
Include SPV metadata in document and proposal serialization

tinagerber:
In the metadata overview of proposals and documents in the new frontend, some metadata concerning the SPV is still missing. In order to display them, some additional fields are needed.

For [CA-2543]

## Checklist

- [x] Changelog entry

- API change:

[CA-2543]: https://4teamwork.atlassian.net/browse/CA-2543

### DIFF
--- a/opengever/api/proposal.py
+++ b/opengever/api/proposal.py
@@ -63,12 +63,16 @@ class SerializeSubmittedProposalToJson(GeverSerializeFolderToJson):
         else:
             result[u'excerpt'] = None
 
-        meeting = self.context.load_model().get_meeting()
+        proposal_model = self.context.load_model()
+        result['committee'] = {'title': proposal_model.committee.title,
+                               '@id': proposal_model.committee.get_url()}
+
+        meeting = proposal_model.get_meeting()
         if meeting:
             result[u'meeting'] = {
                 'title': meeting.title,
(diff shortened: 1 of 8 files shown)


opengever/api/proposal.py
```diff
@@ -63,7 +63,10 @@ def __call__(self, *args, **kwargs):
         else:
             result[u'excerpt'] = None
 
-        meeting = self.context.load_model().get_meeting()
+        proposal_model = self.context.load_model()
+        result['committee_url'] = proposal_model.committee.get_url()
```
deiferni, line 67: Is there a reason to not build a nested structure here? I would have expected to have a `committee` attribute with a nested `@id` and `title`.

deiferni (approved):
LGTM 👍  apart from suggestion.

opengever/api/proposal.py
tinagerber: I was not sure which is the better way. I decided against it because the title is already serialized with the committee_oguid and so we would have the title twice. But if you prefer a nested structure, I can change it.

deiferni: I personally would lean towards consistency but this is a detail and opinions may vary. Your call 😉 .

PR #7117, opened 2021-08-11
Redirect to notification resource while resolving.

deiferni:
Change the current behavior to always redirect to notification resource in `@@resolve_oguid` if user has permission to view the resource, even for foreign notifications. The  previous behavior was to return an error if the notification was not visible for the current user. The change adds more convenience in case the link to the notification is
shared between users unintentionally (e.g by forwarding a mail).

For https://4teamwork.atlassian.net/browse/CA-2235

## Checklist

- [x] Changelog entry

### DIFF
--- a/opengever/activity/browser/resolve.py
+++ b/opengever/activity/browser/resolve.py
@@ -2,7 +2,6 @@ from opengever.activity import notification_center
 from opengever.base.browser.resolveoguid import ResolveOGUIDView
 from opengever.base.exceptions import InvalidOguidIntIdPart
 from opengever.ogds.models.service import ogds_service
-from opengever.readonly import is_in_readonly_mode
 from plone import api
 from zExceptions import NotFound
 from zExceptions import Unauthorized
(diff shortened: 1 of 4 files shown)


opengever/activity/browser/resolve.py
```diff
@@ -21,10 +21,9 @@ def __call__(self):
             raise NotFound('Invalid notification_id ({}) is given'.format(
                 self.request.get('notification')))
 
-        if not self.check_permission():
-            raise Unauthorized()
+        if self.check_permission():
+            self.mark_as_read()
```
njohner, line 25: I find the method name `check_permission` somewhat confusing. How about `belongs_to_current_user` or `current_user_owns_notification` or something of that sort?

njohner (approved):
LGTM.

opengever/activity/browser/resolve.py
deiferni: I have not introduced this so i initially decided against touching it. But you also mentioning it convinces me to do something. IMO that functionality would be better placed on the model and not in a view as it contains logic. I have moved it accordingly.

PR #7118, opened 2021-08-11
Fix workspace daily digest notification defaults.

deiferni:
We don't want to send daily digests for workspace members when todos are added by default. No migration for existing deployments is provided, this will purposefully reduce the amount of daily digests that we send and this is desired.

I've had trouble with some tests suddenly failing. IMO they are not related to the change and were flaky and one of them not correct due to a weird setup. I have reimplemented them as integration test case and now they seem to work correctly.

## Checklist

- [x] Changelog entry


opengever/inbox/tests/test_activities.py
```diff
@@ -245,6 +222,75 @@ def test_assign_to_dossier_open_successor_task(self, browser):
...
+        )
+        self.assertItemsEqual(
+            [(u'jurgen.konig', u'task_responsible'),
+             (u'inbox:fa', u'task_issuer')],
+            [(subscription.watcher.actorid, subscription.role)
+             for subscription in successor_resource.subscriptions]
+        )
+
+    @browsing
+    def test_accepting_and_assign_forwarding_with_successor_and_updated_responsibles(self, browser):
```
njohner, line 262: Here too, I find the test name confusing... `test_accepting_and_assigning_forwarding_with_successor_updates_responsibles`?

njohner (approved):
LGTM. Just some confusion for me on the test namings...

opengever/inbox/tests/test_activities.py
```diff
@@ -245,6 +222,75 @@ def test_assign_to_dossier_open_successor_task(self, browser):
             [(subscription.watcher.actorid, subscription.role)
              for subscription in resource.subscriptions])
 
+    @browsing
+    def test_accepting_forwarding_with_successor_updated_responsibles(self, browser):
```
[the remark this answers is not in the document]
deiferni: Even after porting the tests i am also not 100% sure. They test various side-effects regarding the notification center of this monstrosity https://github.com/4teamwork/opengever.core/blob/757027767071afb31d7b886c2f79999f9841b5df/opengever/task/browser/accept/utils.py#L65-L181.

I have renamed as you have suggested this may clarify a bit more. For the record this does not concern functionality i have touched in this PR. I would prefer to introduce as few unrelated changes as possible. Changing handling of forwardings is huge and out of scope here.

opengever/inbox/tests/test_activities.py
deiferni: Same here. Renamed the test 👍 .

PR #7119, opened 2021-08-12
Remove `@@` prefix from links to personal preferences browser view.

deiferni:
With this PR we drop the `@@` prefix from links to the personal preferences browser view in two places:
- From the link in notification mails
- From the action url displayed in the personal navigation menu

Dropping the prefix will prevent the URL being handled by rewrite rules redirecting requests to the backend regardless of the cookie enabling/disabling the new UI. It is intended to work in combination with https://github.com/4teamwork/gever-ui/pull/1887 and fix/improve URL handling to notification settings in the following cases:
- Links from the mail should work on both the classical and new UI (for GEVER) and for the new UI (for temaraum)
- Handling works when changing the UI from classical to new (We've decided against special handling for the inverted case as this switch is not expected to happen often)

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-2708.bugfix
+++ b/changes/CA-2708.bugfix
@@ -1,0 +1,1 @@
+Remove `@@` prefix from links to personal preferences browser view. [deiferni]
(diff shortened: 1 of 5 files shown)


njohner (approved):
LGTM

PR #7121, opened 2021-08-13
Add new @reference-numbers endpoint.

njohner:
Add new @reference-numbers `GET` and `DELETE` endpoints for administrators. The `GET` returns a list of all reserved positions on a given context, while the `DELETE` endpoint allows to liberate an inactive position (position that was previously in use but is not used any longer).

For [CA-2096]

## Checklist
- [x] Changelog entry
- API change:
  - If breaking:
    - [ ] api-change label added

[CA-2096]: https://4teamwork.atlassian.net/browse/CA-2096

### DIFF
--- a/docs/public/dev-manual/api/reference_numbers.rst
+++ b/docs/public/dev-manual/api/reference_numbers.rst
@@ -1,0 +1,48 @@
+.. _reference-numbers:
+
+Ordnungspositionsnummer
+=======================
+
+Ordnungspositionsnummer können von Administratoren auf Stufe Ordnungssystem und Ordnungspositionen mit dem ``@reference-numbers`` Endpoint verwaltet werden.
+
+Eine GET-Request liefert eine Liste von den Ordnungspositionsnummer, welche auf dem Kontext verwendet werden.
+
+  .. sourcecode:: http
+
+    GET /@reference-numbers HTTP/1.1
+    Accept: application/json
+
+  .. sourcecode:: http
+
+    HTTP/1.1 200 OK
+    Content-Type: application/json
+
+    [
+        {
+            "active": true,
+            "prefix": "1",
+            "title": "Führung"
+        },
+        {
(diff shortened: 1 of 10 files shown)


opengever/api/tests/test_reference_numbers.py
```diff
@@ -0,0 +1,41 @@
...
+        self.assertEqual(
+            [{u'active': True,
+              u'prefix': u'1',
+              u'title': u'F\xfchrung'},
+             {u'active': True,
+              u'prefix': u'2',
+              u'title': u'Rechnungspr\xfcfungskommission'},
+             {u'active': True,
+              u'prefix': u'3',
+              u'title': u'Spinn\xe4nnetzregistrar'}],
```
phgross, line 41: Does serialization/title for an no longer existing object works? maybe add a test.

docs/public/dev-manual/api/reference_numbers.rst
```diff
@@ -0,0 +1,48 @@
+.. _reference-numbers:
+
+Aktenzeichen
+============
+
+Aktenzeichen können von Administratoren auf Stufe Ordnungssystem und Ordnungspositionen mit dem ``@reference-numbers`` Endpoint verwaltet werden.
```
phgross, line 6: Maybe we could add here a note, that its not really the complete reference-number your able to see/delete, but only the local part of the reference number.

phgross (approved):
LGTM 👍, just two details.

opengever/api/tests/test_reference_numbers.py
njohner: Yeah, but it was returning a label, which is not ideal. I added a commit to handle that case a bit better and return `None` for the title instead.

njohner:
@phgross can you have another quick look?

phgross (approved):
🚢🚢🚢

PR #7122, opened 2021-08-13
Fix updating document title in favorites with API.

deiferni:
Fix updating document title in favorites when document title is changed via API request. The api seems to choose a different interface here (IMO the correct one) and our changed detection heuristic did not pick it up correctly.

I am also adding a separate test for dossier favorite titles which can also change. This should cover the two most important use-cases. Task titles can't really change and repo favorites are handled differently anyway.

For https://4teamwork.atlassian.net/browse/CA-2688

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-2688.bugfix
+++ b/changes/CA-2688.bugfix
@@ -1,0 +1,1 @@
+Fix updating document title in favorites when document title is changed via API. [deiferni]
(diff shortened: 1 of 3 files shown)


njohner (approved):
LGTM

PR #7123, opened 2021-08-16
Revert merge of upgrade reindexing reference and sortable_reference.

njohner:
Merging the older upgrade step reindexing sortable_reference with a newer one is counterproductive. Indeed this leads to indexing all sortable_reference numbers in solr a second time for deployments that are on a version between 2021.4 and 2021.12. Indeed the upgrade adding the `sortable_reference` index was included in 2021.4 ( https://github.com/4teamwork/opengever.core/pull/6732) and the upgrade steps were merged in 2021.12 (https://github.com/4teamwork/opengever.core/pull/7029). This is notably the case for SG...

For [CA-2350]

## Checklist

- [x] Changelog entry

[CA-2350]: https://4teamwork.atlassian.net/browse/CA-2350

### DIFF
--- a/opengever/core/upgrades/20210205084521_add_sortable_reference_number_index/upgrade.py
+++ b/opengever/core/upgrades/20210205084521_add_sortable_reference_number_index/upgrade.py
@@ -16,14 +16,9 @@ class AddSortableReferenceNumberIndex(UpgradeStep):
         self.index_sortable_reference_number()
 
     def index_sortable_reference_number(self):
-
-        # Disable this upgradestep because it's done by the later upgradestep
-        # 20210601120606_update_reindex_reference_and_sortable_reference_index
-        return
-
-        # manager = getUtility(ISolrConnectionManager)
-        # query = {'object_provides': IDexterityContent.__identifier__}
-        # for obj in self.objects(query, 'Index sortable_reference in Solr'):
-        #     handler = getMultiAdapter((obj, manager), ISolrIndexHandler)
-        #     handler.add(['sortable_reference'])
-        # manager.connection.commit(soft_commit=False, extract_after_commit=False)
+        manager = getUtility(ISolrConnectionManager)
+        query = {'object_provides': IDexterityContent.__identifier__}
+        for obj in self.objects(query, 'Index sortable_reference in Solr'):
+            handler = getMultiAdapter((obj, manager), ISolrIndexHandler)
+            handler.add(['sortable_reference'])
+        manager.connection.commit(soft_commit=False, extract_after_commit=False)
--- a/opengever/core/upgrades/20210205084521_add_sortable_reference_number_index/upgrade.py
+++ b/opengever/core/upgrades/20210205084521_add_sortable_reference_number_index/upgrade.py
@@ -17,8 +17,13 @@ class AddSortableReferenceNumberIndex(UpgradeStep):
 
     def index_sortable_reference_number(self):
         manager = getUtility(ISolrConnectionManager)
+        solr_connection = manager.connection
+
         query = {'object_provides': IDexterityContent.__identifier__}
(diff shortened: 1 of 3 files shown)


opengever/core/upgrades/20210601180416_update_reindex_reference_and_sortable_reference_index/upgrade.py
```diff
@@ -19,19 +15,12 @@ def __call__(self):
...
             obj = self.catalog_unrestricted_get_object(brain)
             if IReferenceNumber(obj).get_number() != brain.reference:
-                obj.reindexObject(idxs=['reference'])
-
-            # update solr
-            handler = getMultiAdapter((obj, manager), ISolrIndexHandler)
-            handler.add(['reference', 'sortable_reference'])
-
-        manager.connection.commit(soft_commit=False, extract_after_commit=False)
+                obj.reindexObject(idxs=['reference', 'sortable_reference'])
```
lukasgraf, line 26: Hmm. Don't we still update `sortable_reference` twice, unnecessarily, like this? (For this version that will get merged to `master`).

Once in the older upgrade step above (for deployments that haven't run that one yet), and again here?

opengever/core/upgrades/20210205084521_add_sortable_reference_number_index/upgrade.py
```diff
@@ -16,14 +16,9 @@ def __call__(self):
...
-        # for obj in self.objects(query, 'Index sortable_reference in Solr'):
-        #     handler = getMultiAdapter((obj, manager), ISolrIndexHandler)
-        #     handler.add(['sortable_reference'])
-        # manager.connection.commit(soft_commit=False, extract_after_commit=False)
+        manager = getUtility(ISolrConnectionManager)
+        query = {'object_provides': IDexterityContent.__identifier__}
+        for obj in self.objects(query, 'Index sortable_reference in Solr'):
+            handler = getMultiAdapter((obj, manager), ISolrIndexHandler)
+            handler.add(['sortable_reference'])
+        manager.connection.commit(soft_commit=False, extract_after_commit=False)
```
deiferni, line 24: I wonder if we should commit to Solr every so often here? This upgrade could grow very large otherwise. But then again the most critical deployments have run it already. So i'll leave it up to you.

deiferni (approved):
Just one thing you'll have to maybe think about a bit. Apart from that exactly what we had in mind this morning. Thanks! 👍

opengever/core/upgrades/20210601180416_update_reindex_reference_and_sortable_reference_index/upgrade.py
lukasgraf: Nevermind, I just realized the second upgrade step is conditional, and only reindexes `sortable_reference` for a few. So it doesn't do everything the first one does.

lukasgraf (approved):

opengever/core/upgrades/20210205084521_add_sortable_reference_number_index/upgrade.py
njohner: The only deployments that did not already run that upgrade are:
- stabs (160'000 objects in the catalog)
- nidau (100'000 objects)
- Schulung köniz
They are not terribly large. But I'll introduce intermediate commits. It's only solr, so transactionality is anyway not guaranteed

opengever/core/upgrades/20210601180416_update_reindex_reference_and_sortable_reference_index/upgrade.py
njohner: Yeah this should happen very seldom, IMO. It's only for objects that were moved, maybe copied...

PR #7124, opened 2021-08-16
Add Solr field approval_state plus corresponding indexer and logic 

lukasgraf:
This adds
- a method `ApprovalList.get_approval_state()` to determine the current `approval_state` of a document based on the contents of the approval lists in annotations
- plus the corresponding **Solr index** `approval_state`
- and the **indexer** on the Plone side.

### Caveat: Revert to version

Currently I only made sure the `approval_state` index is _updated_ in Solr when a document gets **reverted**. But we maybe should also fix the approval lists when a document gets reverted to a version that has previously been approved, similar to what is being done when a document is copied. 

⚠️ So right now, reverting to a document version that has previously been approved will not retain that approval state, and the document will therefore be "approved in an older version".  

For [CA-2267](https://4teamwork.atlassian.net/browse/CA-2267)

## Checklist

- [x] Changelog entry
- API change:
- New functionality:

### DIFF
--- a/changes/CA-2267.feature
+++ b/changes/CA-2267.feature
@@ -1,0 +1,1 @@
+Add approval_state Solr field and corresponding Plone indexer. [lgraf]
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -12,6 +12,7 @@ Breaking Changes
 Other Changes
 ^^^^^^^^^^^^^
 
+- ``@listing``: Add ``approval_state`` column
 - Include ``committee`` in proposal serialization.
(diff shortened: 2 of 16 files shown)


phgross (approved):
LGTM 👍 

I'm not really sure about the "revert-caveat", should discuss this with Erich tomorrow, but we could merge this PR as it is and change the revert behavior in a separate PR if necessary at all.

lukasgraf:
Yeah, I've briefly looked into it, and copying around approvals could be quite messy. Also it could get weird when displaying detailed approvals, because dates vs. versions are out of order, and technically, that version with that version ID never got approved.

So requiring the reverted version to be approved again might not be the worst thing. -> 👍 let's discuss it tomorrow.

PR #7127, opened 2021-08-18
Allow Reader to revive bumblebee previews.

deiferni:
Allow users with the `Reader` role to revive bumblebee previews. Previously this was limited to admins and managers but there is not really a reason to do so.

Tests are done with the regular user though as reader must be set up for each test case. Testing with the regular user covers the common use case better and is thus IMO preferrable.

For [CA-2132](https://4teamwork.atlassian.net/browse/CA-2132)

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-2132.other
+++ b/changes/CA-2132.other
@@ -1,0 +1,1 @@
+Allow Reader to revive bumblebee previews. [deiferni]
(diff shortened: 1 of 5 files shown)


phgross (approved):
👍 LGTM.

There is a `self.reader_user` in our testing fixtures, but not sure if that would fit for your test.

deiferni:
@phgross i have seen that one but it needs manual setup of permissions. That is why i have switched to the user we currently have which is IMO a better fit.

PR #7131, opened 2021-08-19
Add support for groups in @share-content endpoint.

tinagerber:
Changes in this PR:
- The `ActualWorkspaceMembersSource` is extended so that the users of groups are also included
- With the parameter `include_groups` groups can now be included in the `@actual-workspace-members` endpoint
-  The `@share-content` endpoint can now handle groups

For [CA-1923]

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [x] api-change label added

[CA-1923]: https://4teamwork.atlassian.net/browse/CA-1923

### DIFF
--- a/opengever/api/share_content.py
+++ b/opengever/api/share_content.py
@@ -11,20 +11,19 @@ from zope.interface import alsoProvides
 
 class ShareContentPost(Service):
 
-    def get_email_adresses(self, users):
+    def get_email_adresses(self, actors):
         service = ogds_service()
-        emails = []
-        for user in users:
-            emails.append(service.fetch_user(user['token']).email)
+        for actor in actors:
+            emails.append(service.fetch_user(actor['token']).email)
         return ', '.join(emails)
 
     def extract_data(self):
         data = json_body(self.request)
         self.comment = data.get('comment', u'')
-        self.users_to = data.get('users_to', [])
-        if not self.users_to:
-            raise BadRequest("Property 'users_to' is required")
-        self.users_cc = data.get('users_cc', [])
+        self.actors_to = data.get('actors_to', [])
+        if not self.actors_to:
+            raise BadRequest("Property 'actors_to' is required")
+        self.actors_cc = data.get('actors_cc', [])
 
     def reply(self):
         if not is_within_workspace(self.context):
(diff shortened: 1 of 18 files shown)


opengever/sharing/tests/test_sharing.py
```diff
@@ -391,8 +391,7 @@ def test_search_result_is_batched(self, browser):
                      method='Get', headers={'Accept': 'application/json'})
 
         result = browser.json
-
-        self.assertEqual(20, result['items_total'])
+        self.assertEqual(19, result['items_total'])
```
njohner, line 394: what happened here?

opengever/workspace/utils.py
```diff
@@ -45,30 +45,35 @@ def get_containing_workspace(context):
...
-    users = set([])
+    actor_ids = set([])
     allowed_roles_to_view = roles_of_permission(containing_workspace, 'View')
     portal = api.portal.get()
 
-    def is_valid_userid(*args):
-        user, roles, role_type, name = args
-        return role_type == u'user' and set(roles) & set(allowed_roles_to_view)
+    def is_valid_actorid(valid_role_type, *args):
+        actor, roles, role_type, name = args
```
njohner, line 68: 😮 Can you change the signature of that method please? `*args` when we know exactly what 4 positional arguments we want does not make any sense.

opengever/ogds/base/sources.py
```diff
@@ -899,6 +900,31 @@ def search_query(self):
...
+        if groupids:
+            query = query.filter(Group.groupid.in_(groupids))
+        else:
+            # Avoid filter for an empty list.
+            query = query.filter(sql.false())
+        return query
+
+    def getTerm(self, value):
+        try:
+            group = self.base_query.filter(Group.groupid == value).one()
```
njohner, line 921: Why are we using `base_query` instead of `search_query`? won't that return terms for values that are not in the vocabulary?

opengever/api/share_content.py
```diff
@@ -13,8 +13,15 @@ class ShareContentPost(Service):
...
         service = ogds_service()
+        emails = set()
         for actor in actors:
-            emails.append(service.fetch_user(actor['token']).email)
+            group = service.fetch_group(actor['token'])
+            user = service.fetch_user(actor['token'])
+            if group:
+                emails.update([user.email for user in group.users])
+            elif user:
+                emails.add(user.email)
```
njohner, line 23: I'd use actors here instead. Then you don't have to do special treatments for users and groups, that's why we have actors ;-). Something like:
```
for actor_id in actors:
    actor = ActorLookup.lookup(actor_id)
    for representative in actor.representatives:
        emails.add(representative.email)
```

njohner (approved):
LGTM, just a few suggestions and comments

opengever/sharing/tests/test_sharing.py
tinagerber: Since Kathi's email is now foo@example.com and no longer kathi.barfuss@gever.local, Kathi is no longer found when searching for "g".

opengever/ogds/base/sources.py
tinagerber: Yes, you're right, I changed it.

PR #7132, opened 2021-08-19
Fix handling of relations when an object is deleted.

njohner:
We fix the handling of relations when deleting an object. There were 2 main issues:
- The `z3c.relationfield` only handles fields stored on the object, while we store our relations in the annotations
- When the `to_object` of a relation is deleted, the relation remains in the catalog, but is marked as broken. We did not handle that case anywhere (API, overviews, etc).

I fixed the event handlers, so that relations get deleted from the catalog when the `from_object` is deleted. 
I also fixed the serializer, so that broken relations are not serialized to the frontend, so that we don't have to handle that case everywhere in the frontend (before they got serialized as `None`, so for a relation list `[None]`). 
I also added handling of broken relations for documents (and all derived types), as well as in the dossier overview, to cover the most likely cases. There are probably other places where we might still have failures, but this seems to never have been a problem. I did not want to invest more time, as this will only lead to errors in the classic ui, and as deletion is still something vera rare in Gever.

I did not add an API changelog entry for this.

For [CA-371] and [CA-2693]

Sentry: 
- https://sentry.4teamwork.ch/organizations/sentry/issues/76984
- https://sentry.4teamwork.ch/organizations/sentry/issues/77009

## Checklist

- [x] Changelog entry

- API change:
  - [ ] API Changelog entry
  - If breaking:
    - [ ] api-change label added

[CA-371]: https://4teamwork.atlassian.net/browse/CA-371
[CA-2693]: https://4teamwork.atlassian.net/browse/CA-2693

### DIFF
--- a/changes/CA-371.bugfix
+++ b/changes/CA-371.bugfix
@@ -1,0 +1,1 @@
+Fix handling of relations when an object is deleted. [njohner]
--- a/opengever/api/configure.zcml
+++ b/opengever/api/configure.zcml
@@ -33,6 +33,7 @@
   <adapter factory=".serializer.SerializeBrainToJsonSummary" />
   <adapter factory=".serializer.SerializeAdminUnitModelToJsonSummary" />
   <adapter factory=".serializer.SerializeOrgUnitModelToJsonSummary" />
+  <adapter factory=".serializer.GeverRelationListFieldSerializer" />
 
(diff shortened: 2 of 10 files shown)


lukasgraf (approved):
LGTM! 👍 🎉

PR #7133, opened 2021-08-19
Fix dossier protection over the rest-api

elioschmutz:
This PR fixes an issue where the dossier have not been protected when using the rest-api.

The information about the dossier protection users are stored in a behavior and are provided in the add/edit form of a dossier in the old UI. If the user fills out the `dossier_manager` and chooses a user for the `reading` or `reading_and_writing` field, the dossier gets protected after saving the form based on this fields. 

The plone rest-api will set the fields properly if provided but will not effectively protect the dossier. It just stores the field values.

We discussed in the team different approaches. The favorite approach would have been to provide a dedicated endpoint for this use case. To implement it without hacks, we would have been reimplementing the whole protection feature with an annotation storage. Because we don't want to touch the old ui, we don't want this refactoring yet.

So we decided to just extend the `POST` and `PATCH` endpoints for plone contents and handle it as we do it in the old ui.

Permission: the permission checks are done on the field layer. The z3c form fields are protected with a specific permission. So there is no need to re-check the permission in the endpoints.

For [CA-2401]

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations


opengever/api/add.py
```diff
@@ -116,7 +118,12 @@ def reply(self):
             self.request.response.setStatus(400)
             return dict(error=dict(type="DeserializationError", message=str(e)))
 
-        return self.add_object()
+        result = self.add_object()
```
njohner, line 121: I'd rather that you do not modify this class. As it says in the docstring, this is a copy `plone.restapi.services.content.add.FolderPost`, but with the code split-up so that it can be overwritten. This allows to keep a better overview of what we have customized for Gever.
You could instead customize `add_object` of the `GeverFolderPost` class?

opengever/api/configure.zcml
```diff
@@ -877,6 +877,22 @@
       permission="cmf.AddPortalContent"
       />
 
+  <plone:service
+      method="POST"
+      for="Products.CMFCore.interfaces.IFolderish"
+      factory=".add.FolderPost"
```
njohner, line 883: I'm not sure why we need this. Dossier protection is only for businesscasedossiers, which provide the `IDossierMarker`, so will use the `GeverFolderPost` service

opengever/api/configure.zcml
```diff
@@ -877,6 +877,22 @@
...
+      layer="opengever.base.interfaces.IOpengeverBaseLayer"
+      permission="cmf.AddPortalContent"
+      />
+
+  <plone:service
+      method="PATCH"
+      for="Products.CMFCore.interfaces.IContentish"
```
njohner, line 890: Why not register that only for `IProtectDossierMarker`?

opengever/api/update.py
```diff
@@ -0,0 +1,24 @@
...
+    """Gever specific implementation to handle dossier protection
+    """
+    def reply(self):
+        result = super(GeverContentPatch, self).reply()
+
+        if IProtectDossierMarker.providedBy(self.context):
+            protect_fields = getFieldNames(IProtectDossier)
+            updated_fields = json_body(self.request).keys()
+            protect_fields_changed = len(
+                set(protect_fields) - set(updated_fields)) != len(protect_fields)
```
njohner, line 19: That's confusing. wouldn't `protect_fields_changed = set(protect_fields).intersection(set(updated_fields))` also work? Seems more straightforward to me.

opengever/api/tests/test_protect_dossier.py
```diff
@@ -0,0 +1,119 @@
...
+
+class TestProtectDossier(IntegrationTestCase):
+
+    @browsing
+    def test_do_not_protect_dossier_by_default(self, browser):
+        self.login(self.dossier_manager, browser=browser)
+
+        data = {
+            '@type': 'opengever.dossier.businesscasedossier',
+            'title': 'Protected dossier',
```
njohner, line 15: 'Protected dossier' -> 'Unprotected dossier'

opengever/api/tests/test_protect_dossier.py
```diff
@@ -0,0 +1,119 @@
...
+        self.assertEqual(201, browser.status_code)
+        self.assertEqual(1, len(children['added']))
+        dossier = children['added'].pop()
+
+        self.assertTrue(IProtectDossier(dossier).is_dossier_protected())
+
+    @browsing
+    def test_can_protect_dossier_on_patch(self, browser):
+        self.login(self.dossier_manager, browser=browser)
+
```
njohner, line 60: self.assertFalse(IProtectDossier(self.dossier).is_dossier_protected())

opengever/api/tests/test_protect_dossier.py
```diff
@@ -0,0 +1,119 @@
...
+        data = {
+            '@type': 'opengever.dossier.businesscasedossier',
+            'title': 'Protected dossier',
+            'responsible': self.dossier_manager.getId(),
+            'reading': [self.regular_user.getId()],
+            'dossier_manager': self.dossier_manager.getId()
+        }
+
+        browser.open(self.leaf_repofolder, data=json.dumps(data),
```
njohner, line 83: Shouldn't that throw a validation error?

opengever/api/tests/test_protect_dossier.py
```diff
@@ -0,0 +1,119 @@
...
+    def test_only_dossier_manager_can_protect_dossier_on_patch(self, browser):
+        self.login(self.regular_user, browser=browser)
+
+        data = {
+            'reading': [self.regular_user.getId()],
+            'dossier_manager': self.dossier_manager.getId()
+        }
+
+        browser.open(self.dossier, data=json.dumps(data),
+                     method='PATCH', headers=self.api_headers)
```
njohner, line 98: Same here?

njohner (commented):
LGTM overall. A few comments and some doubts about validation

opengever/api/add.py
elioschmutz: 👍

elioschmutz: `GeverFolderPost` is registered for `IDossierMarker`. Means, if I `POST` to a dossier. But I need to handle dossier creation, so a `POST` to any folderish context.

The naming is a little bit confusing. If I read `GeverFolderPost` I'd not assume that it's only for dossiers, but for any folderish gever object.

So I can't use `GeverFolderPost` and registering it for any folderish content object seems to be wrong. So what I could do is to rename `GeverFolderPost` to `DossierPost` and let it inherit from a new class `GeverFolderPost` which contains my changes:

Inheritance hierarchy:
- `DossierPost` (current `GeverFolderPost`)
- `GeverFolderPost` (new class with my changes)
- `FolderPost`

opengever/api/configure.zcml
elioschmutz: Because it makes it harder to provide multiple behavior functionalities if we register services for each behavior

I'd agree if we register a dedicated endpoint for the dossier protection. But here we just use the `PATCH` endpoint to update a content type.

opengever/api/update.py
elioschmutz: 👍 much better

opengever/api/tests/test_protect_dossier.py
elioschmutz: no, not through the plone restapi. Protected fields will just be ignored: https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/deserializer/dxcontent.py#L98

opengever/api/tests/test_protect_dossier.py
elioschmutz: see my previous answer

elioschmutz:
@njohner I answered your comments. Could you re-comment my answers? Just call me if needed

opengever/api/add.py
njohner: What you propose works for me. Registering `GeverFolderPost` for all folderish contents would also work for me. It was only registered for Dossiers, because there was no point in registering it for other content types until now.

opengever/api/configure.zcml
njohner: Sorry you're right. We POST to create a dossier in another folderish object...

opengever/api/tests/test_protect_dossier.py
njohner: I don't like all these things getting silently dropped, but we're not changing the RESTAPI... so ok

njohner (approved):
LGTM

PR #7130, opened 2021-08-18
Revert "Fix workspace daily digest notification defaults."

deiferni:
Merging https://github.com/4teamwork/opengever.core/pull/7118 seems to cause tests on `master` to fail. We have a flaky or order-dependent test. I'll need to have a look again. Sorry about the inconvenience.

For [CA-2708](https://4teamwork.atlassian.net/browse/CA-2708)

## Checklist

_Everything has to be done/checked. Checked but not present means the author deemed it unnecessary._

- [ ] ~~Changelog entry~~
- [x] Link to issue (Jira or GitHub) and backlink in issue (Jira)

PR #7136, opened 2021-08-23
Fix workspace daily digest notification defaults and flaky tests. 

deiferni:
Wit this PR we bring https://github.com/4teamwork/opengever.core/pull/7118 back from the graveyard to the battlefield. We had to temporarily revert the proposed changed due to unfixed flakiness with https://github.com/4teamwork/opengever.core/pull/7130. I've chosen to put changes and flakiness fixes together in one PR but i can also ship them separately if desired.

**Changes**

We don't want to send daily digests for workspace members when todos are added by default. No migration for existing deployments is provided, this will purposefully reduce the amount of daily digests that we send and this is desired.

I've had trouble with some tests suddenly failing. IMO they are not related to the change and were flaky and one of them not correct due to a weird setup. I have reimplemented them as integration test case and now they seem to work correctly.

**Fixing flaky tests**

To fix flaky tests we aggressively expire all objects in the session before querying the objects we assert against. I am just fighting the symptoms here though as a proper solution would involve to fight with this monstrosity https://github.com/4teamwork/opengever.core/blob/757027767071afb31d7b886c2f79999f9841b5df/opengever/task/browser/accept/utils.py#L65-L181 which IMO is way out of scope for this issue.

I am relatively convinced that we got rid of the flaky tests. Before the changes in this PR approximately every 2nd run would be flaky and fail. With the changes i was no longer able to reproduce any failures. I have also successfully re-run the CI jobs without failures. If the reviewer chooses to validate this claim (Please do!) it can be done as follows:

_Remove the test cache first as executing tests against a fresh cache and a loaded cache behaved different:_
```
echo "y" | bin/remove-test-cache
for i in {1..5}; do bin/test-cached -t TestForwardingActivitesIntegration; done
```

_And just repeat the flaky test without a cached fixture to run them the same way as CI:_
```
for i in {1..5}; do bin/test -t TestForwardingActivitesIntegration; done
```

## Checklist

- [x] Changelog entry


lukasgraf (approved):
LGTM 👍 
Also, I did run the tests locally as described, and was not able to provoke any failures (during 2x5 runs, cached/uncached). 🎉

PR #7147, opened 2021-09-01
Allow editing of a document with Office Online even if the document is not locked

tinagerber:
Since the edit_office_online action used the lock token to determine if a document was checked out collaboratively, the action was no longer available when the lock was removed.

For [CA-2370]

## Checklist

- [x] Changelog entry

[CA-2370]: https://4teamwork.atlassian.net/browse/CA-2370


njohner (approved):
LGTM

PR #7148, opened 2021-09-01
Update local roles when dossier protection is revoked.

tinagerber:
Since it is not possible to authorize users through the `reading` and `reading_and_writing` fields if no dossier manager is set, it is not necessary to check if a dossier manager is set when making a change, as this prevents dossier protection from being removed if all fields are cleared at the same time.

For [CA-2774]

## Checklist

- [x] Changelog entry

[CA-2774]: https://4teamwork.atlassian.net/browse/CA-2774

### DIFF
--- a/opengever/dossier/behaviors/protect_dossier.py
+++ b/opengever/dossier/behaviors/protect_dossier.py
@@ -182,7 +182,7 @@ class DossierProtection(AnnotationsFactoryImpl):
         to the current dossier also if it not defined in the schema. This
         is to prevent locking out yourself.
         """
-        if not self.need_update(force_update):
+        if not self.is_update_allowed(force_update):
             return
 
         self.update_role_inheritance()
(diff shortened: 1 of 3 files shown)


opengever/dossier/behaviors/protect_dossier.py
```diff
@@ -199,10 +199,7 @@ def need_update(self, force_update=False):
         """
         if force_update:
             return True
-
-        return api.user.has_permission(
-            'opengever.dossier: Protect dossier', obj=self.context) and \
-            self.dossier_manager
+        return api.user.has_permission('opengever.dossier: Protect dossier', obj=self.context)
```
njohner, line 202: Where is the protection avoiding having no `dossier_manager` now? That might have side effects.
I'd suggest something like:
- rename that method to `is_update_allowed` (current naming is plain wrong)
- modify the check to:
`api.user.has_permission('opengever.dossier: Protect dossier', obj=self.context) and (self.dossier_manager or self.is_dossier_protected())`

I think that is what we want to check. If the user protects the dossier (as opposed to removing the protection), then a dossier manager is required.

tinagerber: Or rather `api.user.has_permission('opengever.dossier: Protect dossier', obj=self.context) and (self.dossier_manager or not self.is_dossier_protected())`

njohner: euh yes sorry 😅

njohner (approved):
LGTM

PR #7150, opened 2021-09-06
Add comments to validator indicating usage.

deiferni:
I've gone through the validators double-checking if things are validated properly. I am leaving some comments for future readers so that we don't have to double-check again.

## Checklist

- [ ] ~~Changelog entry~~


phgross (approved):
👍

PR #7152, opened 2021-09-06
Prevent non-docx and empty files in a PATCH request to a proposal document.

deiferni:
Bring API validation in line with classic ui validation for proposal documents. We fix the following two potential issues:
- Prevent users from replacing a proposal document with a non-docx document.
- Prevent users from removing the file from a proposal document.

_This still leaves a potential loophole as we don't inspect file data. This is consistent with behavior in the classical UI and also i'm not sure if want to do so and can do so reliably._

## Checklist

- [x] Changelog entry
- API change:

### DIFF
--- a/changes/CA-2156-1.bugfix
+++ b/changes/CA-2156-1.bugfix
@@ -1,0 +1,1 @@
+Prevent non-docx and empty files in a PATCH request to a proposal document. [deiferni]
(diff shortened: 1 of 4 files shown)


njohner (approved):
Is the API change missing on purpose? I'd say it's fine though, as this should not make a difference for the user...

deiferni:
@njohner you are right. I have added an entry.

PR #7153, opened 2021-09-06
Allow checkin for collaborative checkout if it's not locked

buchi:
Include additional infos in document serialization and status endpoint:
- file modification time
- checkout collaborators
- lock time and timeout

This info will be displayed in the document status bar.

Allow checkin of collaborative checkouts for all collaborators if the document is not locked (lock has expired).
Usually Office Online renews the lock regularly. It happens that Office Online does not unlock the document correctly, e.g. if the user closes the Office Online tab while not being connected to the internet. In this case the lock expires and the collaborators should be able to checkin the document.

For [CA-2339](https://4teamwork.atlassian.net/browse/CA-2339)

## Checklist

- [x] Changelog entry

- API change:
- New functionality:

### DIFF
--- a/opengever/api/tests/test_mail.py
+++ b/opengever/api/tests/test_mail.py
@@ -30,6 +30,7 @@ class TestGetMail(IntegrationTestCase):
         self.assertFalse(browser.json['is_shadow_document'])
         self.assertFalse(0, browser.json['current_version_id'])
         self.assertEqual([], browser.json['attachments'])
+        self.assertEqual(self.mail_eml.message._p_mtime, browser.json['file_mtime'])
 
     @browsing
     def test_contains_also_original_message(self, browser):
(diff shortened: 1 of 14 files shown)


opengever/api/tests/test_mail.py
```diff
@@ -28,7 +28,7 @@ def test_additional_metadata_for_mails(self, browser):
         self.assertIsNone(browser.json['containing_subdossier_url'])
         self.assertFalse(browser.json['trashed'])
         self.assertFalse(browser.json['is_shadow_document'])
-        self.assertFalse(0, browser.json['current_version_id'])
+        self.assertEqual(0, browser.json['current_version_id'])
```
njohner, line 31: 😅

opengever/document/tests/test_checkout.py
```diff
@@ -112,12 +113,24 @@ def test_collaborative_checkin_only_allowed_for_other_collaborators(self):
         with self.assertRaises(Unauthorized):
             manager.checkin(collaborative=True)
 
+    def test_collaborative_checkin_allowed_if_not_locked(self):
```
njohner, line 116: You are not testing collaborative checkin but checking for a collaboratively checked out document, so `test_checking_allowed_if_collaboratively_checked_out_and_not_locked` or something along that line would be more correct?

njohner (approved):
LGTM

opengever/document/tests/test_checkout.py
buchi: renamed it as proposed

PR #7154, opened 2021-09-06
Reindex responsible when accepting team task.

njohner:
Accepting a team-task changes the responsible but did not reindex it. This was not visible in the classic UI because the listings use the global index. In the new frontend, the listings show the wrong responsible as the task listing comes from solr.

requires https://github.com/4teamwork/ftw.upgrade/pull/220

For [CA-2785]

## Checklist
- [x] Changelog entry

[CA-2785]: https://4teamwork.atlassian.net/browse/CA-2785


versions.cfg
```diff
@@ -102,7 +102,7 @@ ftw.testbrowser = 1.30.1
 ftw.testing = 1.20.2
 ftw.tika = 2.9.0
 ftw.tokenauth = 1.1.0
-ftw.upgrade = 3.1.0
+ftw.upgrade = 3.1.1
```
deiferni, line 105: This is a bugfix level release. Nothing to be fixed now but please use a minor next time.

deiferni (approved):
LGTM 👍

versions.cfg
njohner: True I guess that is more on the MINOR level...

PR #7157, opened 2021-09-07
No longer allow adding a repository folder to a repository folder when the repository folder is deactivated

tinagerber:
Since the repository workflow allows adding repository folders when repository folders are deactivated, the repository folder incorrectly appears in the add menu.

For [CA-2175]

## Checklist

- [x] Changelog entry

- API change:

[CA-2175]: https://4teamwork.atlassian.net/browse/CA-2175


opengever/repository/tests/test_repositoryfolder.py
```diff
@@ -126,6 +127,16 @@ def test_only_repofolder_addable_when_already_contains_repositories(self, browse
             ['Repository Folder'],
             factoriesmenu.addable_types())
 
+    @browsing
+    def test_repofolder_not_addable_when_repofolder_is_deactivated(self, browser):
+        self.login(self.administrator, browser)
+        browser.open(self.branch_repofolder)
+        self.assertIn('Repository Folder', factoriesmenu.addable_types())
+        editbar.menu_option('Actions', 'Deactivate').click()
+        with self.assertRaises(ValueError) as cm:
```
deiferni, line 136: i think you cold do `assertFalse(factoriesmenu.visible())` to have a nicer test here.

opengever/core/profiles/default/workflows/opengever_repository_workflow/definition.xml
```diff
@@ -236,8 +236,6 @@
     <permission-map
         name="opengever.repository: Add repositoryfolder"
         acquired="False">
-      <permission-role>Administrator</permission-role>
-      <permission-role>Manager</permission-role>
```
deiferni, line 240: I would prefer to leave manager in. IMO we should always be privileged to bypass these restrictions for maintenance.

deiferni (commented):
Two small suggestions. Apart from that 👍 .

opengever/core/profiles/default/workflows/opengever_repository_workflow/definition.xml
tinagerber: You think it is necessary that a manager can create new repository folders in deactivated repository folders? This is not the case with other content types, e.g. deactivated dossiers.  I think a manager knows that he has to reactivate the repository folder to create new repository folders.

deiferni: ah yes, you're right. In that case it is good the way it is for consistency reasons. :shipit:!

deiferni (approved):

PR #7158, opened 2021-09-07
Sanitize extracted document_author from mails.

deiferni:
When extracting `document_author` from mails and setting it on D&Duploaded mails we did bypass input validation of the form fields. We change how the from header is handled by adding manual sanitation of
extracted values. The field is relatively permissive and only disallows newlines and carriage returns. We also strip the value before attempting to lookup an actor so that we could also handle leading/trailing whitespace should it make its way into the header somehow.

For [CA-2152](https://4teamwork.atlassian.net/browse/CA-2152)

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-2152.bugfix
+++ b/changes/CA-2152.bugfix
@@ -1,0 +1,1 @@
+Sanitize document_author after extracting from mail header. [deiferni]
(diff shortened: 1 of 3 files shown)


njohner (approved):
LGTM

PR #7159, opened 2021-09-07
Initialize custom properties with default values on non-interactive document creations

lukasgraf:
This initializes custom properties with default values for

- Documents and mails **uploaded with TUS**
- Documents and mails **uploaded via QuickUpload**
- Mails received via **inbound mail**
- Documents **extracted from mail attachments**

## Checklist

- [ ] Changelog entry (*not necessary - fix for behavior that was introduced in the same release*)

- New functionality:

### DIFF
--- a/opengever/propertysheets/creation_defaults.py
+++ b/opengever/propertysheets/creation_defaults.py
@@ -1,4 +1,17 @@
 from opengever.propertysheets.storage import PropertySheetSchemaStorage
+from zope.component import queryAdapter
+
+
+def initialize_customproperties_defaults(obj, behavior_iface):
+    behavior = queryAdapter(obj, behavior_iface)
+    if behavior:
+        custom_prop_defaults = get_customproperties_defaults(
+            behavior_iface['custom_properties'])
+
+        # Only attempt to set defaults if no actual custom properties exist yet
+        if not behavior.custom_properties:
+            behavior.custom_properties = custom_prop_defaults
+            obj.reindexObject()
 
 
 def get_customproperties_defaults(propsheet_field):
--- a/opengever/propertysheets/creation_defaults.py
+++ b/opengever/propertysheets/creation_defaults.py
@@ -2,7 +2,7 @@ from opengever.propertysheets.storage import PropertySheetSchemaStorage
 from zope.component import queryAdapter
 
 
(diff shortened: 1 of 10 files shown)


opengever/propertysheets/tests/test_creation_defaults.py
```diff
@@ -0,0 +1,163 @@
...
+            )
+            .with_field(
+                'textline', u'notrequired', u'Optional field with default', u'',
+                required=False,
+                default=u'Not required, still has default'
+            )
+            .with_field(
+                'multiple_choice', u'languages', u'Languages', u'',
+                required=True, values=[u'de', u'fr', u'en'],
+                default={u'de', u'en'},
```
deiferni, line 58: This set instance here could be potentially referenced when setting it on an object leading to shared state/defaults. Is there some functionality somewhere guarding against this?

opengever/propertysheets/creation_defaults.py
```diff
@@ -0,0 +1,32 @@
...
+
+    storage = PropertySheetSchemaStorage()
+    for slot_name in propsheet_field.valid_assignment_slots_factory():
+        slot_values = {}
+        definition = storage.query(slot_name)
+
+        if definition is not None:
+            for (field_name, field) in definition.get_fields():
+                if field.default is not None:
+                    slot_values[field.__name__] = field.default
```
deiferni, line 41: Ah here it is … you might have to guard against mutable defaults and create a copy of sets/lists/…. See previous comment. Unless of course the field already takes care of such things while accessing the default.

deiferni (approved):
LGTM 👍 

I've discovered one potential issue where the default (mutable) might be shared between objects when setting it from the fields default. Not sure what happens under the hood there when retrieving the default but best have a close look before we ship.

opengever/propertysheets/creation_defaults.py
lukasgraf: Great catch! 🏹 
It was indeed an issue. 💥  Added a failing test and a fix.

lukasgraf:
@deiferni thanks for the review! I addressed the mutable kwarg issue, and also decided to make the `Location` header parsing in TUS upload more robust (last 3 commits). Would you mind having another quick look?

PR #7160, opened 2021-09-08
Fixed FixSharingPermissions upgradestep.

phgross:
Avoid workflow update for all documents.

During the SG testrun I realized that the upgrade step `FixSharingPermissions`, which intends to update only the inbox and private documents, updates the workflow for all documents. Which would make the upgrade step unnecessarily longer. 

The reason is how the ftw.upgrades `WorkflowSecurityUpdater` works. It fetches the types for the given workflows (https://github.com/4teamwork/ftw.upgrade/blob/master/ftw/upgrade/workflow.py#L179) and queries all those types. Because we use a placeful workflow for inbox_documents and private_documents, the type `opengever.document.document` is queried and we update all documents.

I have therefore rewritten the upgrade step so that we update the inbox and private documents manually.

For [CA-2352]

## Checklist

- [x] Changelog entry

[CA-2352]: https://4teamwork.atlassian.net/browse/CA-2352

### DIFF
--- a/changes/CA-2352.bugfix
+++ b/changes/CA-2352.bugfix
@@ -1,0 +1,1 @@
+Avoid workflow update for all documents, in the FixSharingPermissions upgradestep. [phgross]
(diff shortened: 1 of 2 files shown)


njohner (approved):
LGTM

deiferni (approved):
LGTM 👍

PR #7161, opened 2021-09-08
Add support for unicode userids in ogds-users and ogds-user-listing endpoints.

njohner:
It seems IMAG has one user with a unicode userid. We fix the `ogds-users` and `ogds-user-listing` endpoints so that they support non-ascii userids.

For [CA-2751]

## Checklist

- [x] Changelog entry

- API change:

[CA-2751]: https://4teamwork.atlassian.net/browse/CA-2751

### DIFF
--- a/opengever/ogds/models/user.py
+++ b/opengever/ogds/models/user.py
@@ -76,7 +76,7 @@ class User(Base):
         super(User, self).__init__(**kwargs)
 
     def __repr__(self):
-        return '<User %s>' % self.userid
+        return '<User %r>' % self.userid
 
     def __eq__(self, other):
         if isinstance(other, User):
@@ -93,7 +93,7 @@ class User(Base):
         if not with_principal:
             return self.fullname()
 
-        return "%s (%s)" % (self.fullname(), self.userid)
+        return u"%s (%s)" % (self.fullname(), self.userid)
 
     def fullname(self):
         """Return a visual representation of the UserPersona as String.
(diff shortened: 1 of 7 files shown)


opengever/ogds/models/user.py
```diff
@@ -75,8 +75,11 @@ def __init__(self, userid, **kwargs):
         self.userid = userid
         super(User, self).__init__(**kwargs)
 
+    def __unicode__(self):
+        return u'<User %s>' % self.userid
+
     def __repr__(self):
-        return '<User %s>' % self.userid
+        return unicode(self).encode('utf-8')
```
lukasgraf, line 82: The `__repr__` of an object should always be ASCII only. *)

So the dynamic part of its repr should also be repr'd. E.g.
```
return '<User %r>' % self.userid' 
#              ^
# (Or, the more cumbersome .format() syntax if you prefer ;-)) 
```
Which should result in something like `"<User u'stefan.h\\xe4nni'>"`.

So I would actually ditch the `def __unicode__`.

*) Having the `__repr__` stricly ASCII only avoids encoding issues when it gets dumped on an interactive console (with unpredictable encoding settings), or logged as part of a warning for example (logging non-ASCII reliably to 1) stdout on console and 2) stdout that gets piped, e.g. to `grep`, requires some trickery).

lukasgraf (changes requested):

njohner:
Thanks, I included your changes

lukasgraf (approved):
LGTM! 👍

PR #7162, opened 2021-09-08
Add upgrade to fix mail authors with newlines.

deiferni:
Follow up for https://github.com/4teamwork/opengever.core/pull/7158 to add an upgrade that also fixes affected content in production. In contrary to my initial assumptions this seems to have happened occasionally and we prefer to have correct data in production 😉 .

- The upgrade applies the same replacements as is done by input sanitation in https://github.com/4teamwork/opengever.core/blob/73a17fb6fd9838a758f97622d96d8665269fbd38/opengever/mail/mail.py#L420
- Would also convert any non-unicode `document_authors` should they exist

For [CA-2152](https://4teamwork.atlassian.net/browse/CA-2152)

## Checklist

- [ ] ~~Changelog entry~~

### DIFF
--- a/opengever/core/upgrades/20210908103657_remove_newlines_from_mail_document_authors/upgrade.py
+++ b/opengever/core/upgrades/20210908103657_remove_newlines_from_mail_document_authors/upgrade.py
@@ -1,0 +1,27 @@
+from ftw.upgrade import UpgradeStep
+from Products.CMFPlone.utils import safe_unicode
+
+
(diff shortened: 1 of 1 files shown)


njohner (approved):
I might be blind, but where do you "convert any non-unicode document_authors should they exist"?
Apart from that, LGTM

deiferni:
@njohner you are not blind, i did not push my recent changes. Thx for paying attention 😓 . Amended now.

PR #7164, opened 2021-09-08
Prevent transferring working copies in remote task.

deiferni:
Prevent transferring documents checked out by different users when completing successor tasks in `@complete-successor-task`. This makes the behavior consistent with validation in the classic UI where such documents are not allowed.

Screenshot 2021-09-08 at 14 56 33

Error handling is unfortunately very basic for the new UI but we have other issues that block having something better right now.

## Checklist

- [x] Changelog entry

- API change:

### DIFF
--- a/changes/CA-2156-2.bugfix
+++ b/changes/CA-2156-2.bugfix
@@ -1,0 +1,1 @@
+Prevent transferring checked out documents when completing successor tasks. [deiferni]
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -22,6 +22,7 @@ Other Changes
 - Prevent setting ``file`` to ``null`` for ``opengever.document.document`` if it is inside an ``opengever.meeting.proposal``.
 - Include checkout collaborators and file modification time in document serialization.
 - Include checkout collaborators, file modification time, lock time and lock timeout in document status.
+- ``@complete-successor-task``: Prevent transferring checked out documents when completing successor tasks.
 
 
 2021.17.0 (2021-08-30)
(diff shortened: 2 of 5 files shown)


changes/CA-2156-2.bugfix
```diff
@@ -0,0 +1 @@
+Prevent transferring documents checked out by different users when completing successor tasks. [deiferni]
```
njohner, line 1: From what I see it does not matter if I checked it out myself or some other user. So I'd rephrase that to:
"Prevent transferring checked out documents when completing successor tasks. [deiferni]"

docs/public/dev-manual/api/api_changelog.rst
```diff
@@ -22,6 +22,7 @@ Other Changes
 - Prevent setting ``file`` to ``null`` for ``opengever.document.document`` if it is inside an ``opengever.meeting.proposal``.
 - Include checkout collaborators and file modification time in document serialization.
 - Include checkout collaborators, file modification time, lock time and lock timeout in document status.
+- ``@complete-successor-task``: Prevent transferring documents checked out by different users.
```
njohner, line 25: same here

opengever/api/complete_successor_task.py
```diff
@@ -71,6 +72,10 @@ def reply(self):
         # Map any supported document reference type to an IntId
         docs_to_deliver = map(self._resolve_doc_ref_to_intid, docs_to_deliver)
 
+        invalid = get_checked_out_documents(docs_to_deliver)
+        if invalid:
+            raise BadRequest('Some documents are checked out by a different user.')
```
njohner, line 77: and here

njohner (approved):
See comment, otherwise LGTM

changes/CA-2156-2.bugfix
deiferni: Thanks! This is what i first implemented but then i saw that the classic ui always blocks, even if the current user has the checkout. amended!

PR #7165, opened 2021-09-08
Drop validation requiring a file or a flag.

deiferni:
Drop validation for documents/mails either requiring a file or the `preserved_as_paper` flag to be `True`.

We don't really see the value of this validation. Also it did not properly validate in API requests so far, so we will have content where the validation was violated.

See https://4teamwork.slack.com/archives/C01BHEGBVU1/p1631113754129300 for confirmation that this is desired.

## Checklist

- [x] Changelog entry

- New functionality:


phgross (approved):
👍

PR #7166, opened 2021-09-09
Respect maximum dossier depth when moving a dossier.

njohner:
Moving a dossier over the API did not validate the dossier depth. Also note that in the classic UI we only checked whether we can add a dossier in the target, we did not check whether the dossier we are adding itself contains subdossiers and if that would respect the dossier depth. This is now also fixed.

For [CA-1207]

## Checklist
- [x] Changelog entry
- API change:
  - If breaking:
    - [ ] api-change label added

[CA-1207]: https://4teamwork.atlassian.net/browse/CA-1207

### DIFF
--- a/opengever/dossier/move_items.py
+++ b/opengever/dossier/move_items.py
@@ -2,6 +2,7 @@ from Acquisition import aq_chain
 from Acquisition import aq_inner
 from Acquisition import aq_parent
 from OFS.CopySupport import CopyError, ResourceLockedError
+from opengever.base.adapters import DefaultMovabilityChecker
 from opengever.base.interfaces import IMovabilityChecker
 from opengever.base.source import RepositoryPathSourceBinder
 from opengever.base.source import SolrObjPathSourceBinder
@@ -20,6 +21,7 @@ from z3c.form.interfaces import HIDDEN_MODE
 from z3c.relationfield.schema import RelationChoice
 from zExceptions import Forbidden
 from zope import schema
+from zope.component import adapter
 from zope.interface import Interface, Invalid
 import z3c.form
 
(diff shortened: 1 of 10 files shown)


opengever/dossier/move_items.py
```diff
@@ -358,3 +366,28 @@ def validate(self, value):
...
+        if not target.is_dossier_structure_addable(substructure_depth):
+            raise Forbidden(u'This would exceed maximally allowed dossier depth.')
+
+    def dossier_structure_depth(self):
+        return self._dossier_structure_depth(self.context, 1)
+
+    def _dossier_structure_depth(self, dossier, current_depth):
+        if not dossier.has_subdossiers():
+            return current_depth
+        return max([self._dossier_structure_depth(brain.getObject(), current_depth + 1)
```
phgross, line 392: Is this only executed when actually moving or also otherwise otherwise it might be performance technically a problem and something like that might be more efficient: 
```
brains = catalog(path=dossier.getPhysicalPath(), object_provides=IDossierMarker.__identifier__)
max_level = max([len(brain.getPath().split('/')) for brain in brains]
return max_level - len(dossier.getPhysicalPath())
```

phgross (commented):
LGTM - not sure about the Performance.

How is the 403 handled in the new UI?

opengever/dossier/move_items.py
njohner: This is only executed for moves, but still, your approach is more efficient and more readable. 👍

njohner:
Error is handled the same as other move errors (like trying to move a proposal document), i.e. it's not handled 😉 . It looks like that:
Screenshot 2021-09-17 at 08 12 26

phgross (approved):
👍

PR #7168, opened 2021-09-13
Continue temporary disable lock check in tus-replace

elioschmutz:
Continue temporary disable lock check in tus-replace because office-connector still does not proivde this feature.

See https://github.com/4teamwork/opengever.core/pull/6636 for more information.


njohner (approved):
LGTM, but can you make sure we have that on the radar? Like is there an issue somewhere, maybe in the technical backlog?

elioschmutz:
@njohner no, we don't have such a story. I'll create one 👍 (/cc @buchi is that correct?)

elioschmutz:
Ok, perfect! Thanks @buchi. So I'll just merge this PR without creating a tracking story.

PR #7170, opened 2021-09-14
Fix ++widget++ traversal when form contains custom properties

lukasgraf:
The `++widget++` traversal for autocomplete searches in the `responsible` field cause the entire form to be updated. When the form contained custom properties widgets, these got updated as well. This caused issues when the custom property widgets did not contain valid values yet at this point.

We therefore skip updating custom properties widgets on `++widget++` traversal requests to avoid this.

I tried to write a test for this, but I was unable to create a test that actually *failed* first.

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-2482.bugfix
+++ b/changes/CA-2482.bugfix
@@ -1,0 +1,1 @@
+Fix ++widget++ traversal when form contains custom properties. [lgraf]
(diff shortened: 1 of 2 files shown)


phgross (approved):
👍

PR #7171, opened 2021-09-14
Make performance improvements for API endpoints using SQLHypermediaBatch.

phgross:
The current implementation of SQLHyperMediaBatch passed in the SQLAlchemy query in to the Batch mechanism, this lead to several unnecessary SQL queries and performance Issue on deployments with a large amount of tasks (`@globalindex` endpoint).

I switched the implementation of the `SQLHyperMedia`, so that the query is only called when initializing the HyperMediaBatch and make sure that the query.count() is also only called once.

On local measurements with 50K tasks, the @globalindex endpoint could be reduced from approx. 5s to approx. 500ms.

For [CA-2852]

## Checklist

- [x] Changelog entry

[CA-2852]: https://4teamwork.atlassian.net/browse/CA-2852

### DIFF
--- a/changes/CA-2852.bugfix
+++ b/changes/CA-2852.bugfix
@@ -1,0 +1,1 @@
+Improve performance for SQL API endpoints, which uses the SQLHypermediaBatch. [phgross]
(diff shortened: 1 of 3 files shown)


buchi (approved):
👍🏻 LGTM

phgross:
@buchi I had to make some adjustments here. But it is now the stand which is also in use for koeniz. Could you take another look at it?

PR #7172, opened 2021-09-14
Transform default *value* to *token* in custom property schemas.

lukasgraf:
Custom properties don't have support for controlling the vocabularies on `Choice` fields yet. Their vocabularies therefore get autogenerated. During that autogeneration, the `token` is always made 7-bit ASCII when it's derived from `values`.

For values that contain non-ASCII characters, this is achieved using `unicode_escape(value)`. That therefore leads to `token`s that are different from `values` (for these terms).

Because the REST API never should expose vocabulary _values_, the client is expected to pass the `token` in its requests. But that means any Field `default`s also need to be translated to the respective `token` before serializing the schema.

## Checklist

- [x] Changelog entry

- API change:
  - [ ] Documentation is updated (nothing to update)
  - [ ] API Changelog entry (deemed unnecessary - bugfix, not a breaking change)

### DIFF
--- a/opengever/propertysheets/schema.py
+++ b/opengever/propertysheets/schema.py
@@ -22,6 +22,15 @@ def get_property_sheet_schema(schema_class):
     required = []
     for field in iter_fields(fieldsets):
         name = field.field.getName()
+
+        # Transform default *value* to *token*
+        default = getattr(field.field, 'default', None)
+        if default is not None:
+            vocab = getattr(field.field, 'vocabulary', None)
+            if vocab:
+                term = vocab.getTerm(field.field.default)
+                properties[name]['default'] = term.token
+
         # Determine required fields
         if field.field.required:
             required.append(name)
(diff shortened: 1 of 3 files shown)


opengever/propertysheets/schema.py
```diff
@@ -22,6 +22,15 @@ def get_property_sheet_schema(schema_class):
     required = []
     for field in iter_fields(fieldsets):
         name = field.field.getName()
+
+        # Transform default *value* to *token*
+        default = getattr(field.field, 'default', None)
+        if default is not None:
+            vocab = default = getattr(field.field, 'vocabulary', None)
```
phgross, line 29: Why is the `default =` necessary?

phgross (approved):
👍 besides the thing mentioned

opengever/propertysheets/schema.py
lukasgraf: Argh. It's not, of course 😉 🙈
Thanks, I'll fix it.

PR #7173, opened 2021-09-14
Propertysheets: Add allow_unmapped parameter to default_from_member options.

lukasgraf:
This adds a new parameter `allow_unmapped` (default: `False`) to the `default_from_member` options for propertysheet definitions.

This will therefore change the default behavior of `default_from_member`: Unless `allow_unmapped` is set to `True`, unmapped values are not returned as defaults, but the `fallback` is used instead.

With `allow_unmapped == True`, unmapped values are returned 1:1.

The reason for this is to add an additional level of robustness, where unexpected (i.e. unmapped) values don't result in invalid defaults being produced.

For [CA-2288](https://4teamwork.atlassian.net/browse/CA-2288)

## Checklist

- [x] Changelog entry

- API change:
  - [x] API Changelog entry
  - If breaking:
    - [ ] api-change label added _(deemed unnecessary, this API is currently internal and hasn't been used yet)_

### DIFF
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -3,7 +3,7 @@
 API Changelog
 =============
 
-2021.19.0 (unreleased)
+2021.18.1 (unreleased)
 ----------------------
 
 Breaking Changes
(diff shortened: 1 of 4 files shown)


phgross (approved):
👍

PR #7174, opened 2021-09-15
Make import of repository excel files more robust.

njohner:
I'm fed up with trying to transform numbers to Text format in Excel, which is simply broken. And I don't want to replace "ja" with "Ja" or the opposite depending on the column. So here it is...

For [CA-1270]

## Checklist

- [x] Changelog entry

[CA-1270]: https://4teamwork.atlassian.net/browse/CA-1270

### DIFF
--- a/changes/CA-1269.other
+++ b/changes/CA-1269.other
@@ -1,0 +1,1 @@
+Make import of repository excel files more robust. [njohner]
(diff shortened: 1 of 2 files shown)


phgross (approved):
🚢 🚢 🚢 

Thx!

PR #7175, opened 2021-09-15
Enable nightly jobs by default.

njohner:
New deployments should all have nightly jobs enabled.

For [CA-1269]

## Checklist

- [x] Changelog entry

[CA-1269]: https://4teamwork.atlassian.net/browse/CA-1269

### DIFF
--- a/changes/CA-1269_2.other
+++ b/changes/CA-1269_2.other
@@ -1,0 +1,1 @@
+Enable nightly jobs in policy templates. [njohner]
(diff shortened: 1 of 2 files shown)


lukasgraf (approved):
LGTM 👍

phgross (approved):
LGTM 👍 

I just wonder if we should switch the flag itself as well? So that this setting also applies to local development

njohner:
@phgross I was first switching the flag default value, but that might be risky and enable the nightlies for deployments where no value is specified in the registry. We discussed that with @lukasgraf  and decided against it.

PR #7176, opened 2021-09-15
Track nightly jobs and OGDS sync runs with timestamps 

lukasgraf:
- Track nightly job runs with a timestamp on the Plone site
- Such a timestamp already exists for OGDS sync
- Add helper functions that determine, based on those timestamps, whether these jobs have been executed in the last 24h:
  - `nightly_run_within_24h()`
  - `ogds_sync_within_24h()`
 
- Add helper functions provide the underlying information:
  - `get_ogds_sync_stamp()`
  - `get_nightly_run_timestamp()`
  - `get_job_counts()`

The nightly jobs timestamp will not so much track *success*, but any *attempt* to run the jobs (that could be successful). So it gets updated whenever the runner is invoked by the cron job and:
- we're not outside the time window
- nightly jobs are not disabled via registry
- system load at the start doesn't lead to an immediate abort

For [CA-1346](https://4teamwork.atlassian.net/browse/CA-1346)

## Checklist

- [x] Changelog entry

### DIFF
--- a/opengever/nightlyjobs/runner.py
+++ b/opengever/nightlyjobs/runner.py
@@ -8,6 +8,7 @@ from opengever.nightlyjobs.interfaces import INightlyJobProvider
 from opengever.nightlyjobs.interfaces import INightlyJobsSettings
 from plone import api
 from plone.subrequest import subrequest
+from zope.annotation import IAnnotations
 from zope.component import getAdapters
 from zope.component import provideUtility
 from zope.globalrequest import getRequest
@@ -21,6 +22,16 @@ import transaction
 def nightly_jobs_feature_enabled():
     return api.portal.get_registry_record(
         'is_feature_enabled', interface=INightlyJobsSettings)
+
+
+def nightly_run_within_24h():
+    portal = api.portal.get()
+    last_run = IAnnotations(portal).get('last_nightly_run')
+
+    if not last_run:
+        return False
+
+    return last_run + timedelta(hours=24) > datetime.now()
 
 
 class TimeWindowExceeded(Exception):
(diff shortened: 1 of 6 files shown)


opengever/nightlyjobs/runner.py
```diff
@@ -97,12 +108,26 @@ def get_job_providers(self):
...
                                INightlyJobProvider)}
 
+    def update_last_run_timestamp(self):
+        """Update timestamp that tracks when nightly jobs were last executed.
+
+        This tracks *attempts* to run them, not success. The timestamp will
+        be updated whenever the runner starts executing jobs, and
+        - we're not outside the time window
+        - nightly jobs are not disabled via registry
+        - system load at the start doesn't lead to an immediate abort
```
njohner, line 118: If I see correctly, it will basically not get updated if we don't finish at least one job (or if there was nothing to run). So if there is a very big job that doesn't commit because it doesn't finish, then the timestamp will not get updated.
I think that is actually beneficial.

opengever/nightlyjobs/runner.py
```diff
@@ -23,6 +24,16 @@ def nightly_jobs_feature_enabled():
...
 
 
+def nightly_run_within_24h():
+    portal = api.portal.get()
+    last_run = IAnnotations(portal).get('last_nightly_run')
+
+    if not last_run:
+        return False
+
+    return last_run + timedelta(hours=24) > datetime.now()
```
njohner, line 34: Because the timestamp might not get updated right away when starting the nightly jobs (because it only gets committed at the end of the first job), I think we could take some safety margin here? Worse case we only notice after two days that the nightlies don't run, but at least we don't have too many false negatives. So maybe use 28 hours? I think 4 hours is our usual time window...

njohner (approved):
LGTM

opengever/nightlyjobs/runner.py
lukasgraf: Exactly. If there's nothing to run, it will get updated, because everything is fine. If a big job runs that never commits however, it's likely that when it re-runs the next night, it will also not finish, and block the queue. So in that case (when no progress is being made), we'd also get a warning.

opengever/nightlyjobs/runner.py
lukasgraf: Good point, thanks! I first looked at an approach where the timestamp gets updated and committed immediately, but that was somewhat risky (could commit other changes in a dirty transaction) and also sent the `TransactionInterceptor` for IntegrationTestCases in a tailspin.

So I'll add 4h + 1h safety margin. As you say, it really doesn't make a difference when we only notice after 48h instead of 24h.

lukasgraf:
@njohner could you have another look please?

njohner (approved):
LGTM

PR #7177, opened 2021-09-15
Automatically close main task when all subtasks in a sequential or parallel task are in a final state.

tinagerber:
This saves the additional effort of manually closing the main task when all subtasks are already closed.

For [CA-2372]

## Checklist

- [x] Changelog entry

[CA-2372]: https://4teamwork.atlassian.net/browse/CA-2372


opengever/task/tests/test_task_from_template.py
```diff
@@ -471,3 +471,103 @@ def test_subsubtask_is_not_part_of_sequential_process(self, browser):
...
+
+        activities = Resource.query.get_by_oguid(
+            Oguid.for_object(self.sequential_task)).activities
+
+        activity = activities[-1]
+        self.assertEquals(u'task-transition-in-progress-tested-and-closed', activity.kind)
+        self.assertEquals(u'Task closed', activity.label)
+
+        last_journal_entry_type = get_journal_entry(self.dossier, -1)['action']['type']
+        self.assertEqual(last_journal_entry_type, 'Task modified')
```
njohner, line 573: The journal entry is in the name of the user who closed the last subtask?

opengever/task/task.py
```diff
@@ -674,6 +674,14 @@ def open_next_task(self):
 
         next_task.sync()
 
+    def close_main_task(self):
+        parent = aq_parent(aq_inner(self))
+        if not parent:
+            return
```
njohner, line 680: Should we check that the parent is a task?

opengever/task/handlers.py
```diff
@@ -92,6 +93,15 @@ def cancel_subtasks(task, event):
 
 
 def start_next_task(task, event):
```
njohner, line 95: Maybe rename that handler?

opengever/task/handlers.py
```diff
@@ -92,6 +93,15 @@ def cancel_subtasks(task, event):
...
+    if not task.is_from_tasktemplate or not task.get_is_subtask():
+        return
+
+    if event.action in FINAL_TRANSITIONS:
```
njohner, line 99: I'm not a 100% sure we want to do this also for parallel tasktemplates? @phgross ? Or maybe check with the PO...

njohner (commented):
LGTM, but I'd like to make sure we want this for parallel taskstemplates as well...

opengever/task/handlers.py
tinagerber: I discussed this with the PO's, we want this for parallel tasks as well.

opengever/task/tests/test_task_from_template.py
tinagerber: Yes. I have now also tested this.

opengever/task/task.py
tinagerber: Since I already tested in the event handler if the task is a subtask, I didn't test this. Since the method is only used in the handler, I would do without it for now.

opengever/task/handlers.py
tinagerber: Hmm, no good name comes to mind. Maybe `review_state_changed` ??

tinagerber:
@njohner I noticed an error in the TaskChecker, which I also corrected. Can you rereview?

njohner (approved):
LGTM

PR #7179, opened 2021-09-16
Redirect to workspace if the invitation link refers to an already accepted invitation

tinagerber:
Until now, an error was thrown when an invitation link was opened that referred to an already accepted invitation.
Now you will be redirected directly to the workspace the invitation was for or, if the user is not logged in, to the login form.

For [CA-1130]

## Checklist

- [x] Changelog entry

[CA-1130]: https://4teamwork.atlassian.net/browse/CA-1130

### DIFF
--- a/opengever/workspace/participation/browser/my_invitations.py
+++ b/opengever/workspace/participation/browser/my_invitations.py
@@ -77,10 +77,8 @@ class MyWorkspaceInvitations(BrowserView):
         invitation = self._get_invitation(iid)
         if not invitation:
             raise BadRequest('Wrong invitation')
-        if invitation['status'] != STATE_PENDING:
-            raise BadRequest('Invitation not pending')
 
-        return self._get_invitation(iid), payload
+        return invitation, payload
 
     def _get_invitation(self, iid):
         try:
@@ -130,6 +128,15 @@ class MyWorkspaceInvitations(BrowserView):
         invitation, payload = self.get_invitation_and_validate_payload()
         with elevated_privileges():
             target_workspace = uuidToObject(invitation['target_uuid'])
+
+        if invitation['status'] != STATE_PENDING:
+            if not api.user.is_anonymous():
+                return self.request.RESPONSE.redirect(target_workspace.absolute_url())
+            elif not payload.get("no_redirect"):
+                params = {'next': target_workspace.absolute_url()}
+                redirect_url = "{}/login?{}".format(
+                    get_gever_portal_url(), urlencode(params))
+                return self.request.RESPONSE.redirect(redirect_url)
 
         if api.user.is_anonymous():
             if payload.get("no_redirect", None):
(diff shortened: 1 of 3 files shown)


opengever/workspace/participation/browser/my_invitations.py
```diff
@@ -77,8 +79,8 @@ def get_invitation_and_validate_payload(self):
         invitation = self._get_invitation(iid)
         if not invitation:
             raise BadRequest('Wrong invitation')
-        if invitation['status'] != STATE_PENDING:
-            raise BadRequest('Invitation not pending')
+        if invitation['status'] == STATE_DECLINED:
+            raise BadRequest('Invitation has already been declined.')
 
         return self._get_invitation(iid), payload
```
njohner, line 85: `return invitation, payload`

opengever/workspace/participation/browser/my_invitations.py
```diff
@@ -77,8 +79,8 @@ def get_invitation_and_validate_payload(self):
         invitation = self._get_invitation(iid)
         if not invitation:
             raise BadRequest('Wrong invitation')
-        if invitation['status'] != STATE_PENDING:
-            raise BadRequest('Invitation not pending')
+        if invitation['status'] == STATE_DECLINED:
+            raise BadRequest('Invitation has already been declined.')
```
njohner, line 83: I'd say if a user has declined an invitation but actually has permissions on a workspace, he could still be redirected no? So I would remove that condition here

opengever/workspace/participation/browser/my_invitations.py
```diff
@@ -208,6 +211,8 @@ def _maybe_create_ogds_entry(self):
     def _accept(self, invitation):
         with elevated_privileges():
             target = uuidToObject(invitation['target_uuid'])
+            if invitation['status'] == STATE_ACCEPTED:
```
njohner, line 214: hum I don't see why we handle this here, see above

opengever/workspace/participation/browser/my_invitations.py
```diff
@@ -143,7 +145,8 @@ def accept(self):
             accept_params = {'iid': invitation['iid'],
                              'no_redirect': 1}
 
-            if not self.storage()._find_user_id_for_email(invitation['recipient_email']):
+            if not self.storage()._find_user_id_for_email(invitation['recipient_email']) and \
```
njohner, line 148: I'd do this differently here:
- `if not invitation['status'] == STATE_PENDING:` redirect directly to the workspace (or to login with the workspace as next url, but not the `accept` url)
-  Then the rest would be as it was before, i.e. either the user already exists, in which case he goes to login, or he does not exist, in which case he goes to registration

njohner (commented):
I'd propose to do the flow a bit differently. See my comments. If unclear or if you disagree, we can have a quick 📞

tinagerber:
@njohner I rewrote it. Can you have another look?

njohner (approved):
LGTM

PR #7180, opened 2021-09-17
Add checkbox to purge solr when deploying Gever.

njohner:
This is not for any story in particular, this just annoyed me once too many times this morning...

This will require a bump of ftw.solr though to include https://github.com/4teamwork/ftw.solr/pull/184, because we do not have access to the registry when setting up Gever, so we cannot do a check whether solr is enabled...

I also take the opportunity to not set `use_solr` to False in examplecontent. This was also annoying because the development deployment was basically broken after setup...

## Checklist

- [x] Changelog entry


lukasgraf (approved):
LGTM! 👍

PR #7182, opened 2021-09-17
Implement sqlite isolation for e2e tests.

elioschmutz:
Isolation for the sqlite in e2e tests:

Related frontend PR: https://github.com/4teamwork/gever-ui/pull/1948

## before

https://user-images.githubusercontent.com/557005/133816226-d5adde95-084f-4b82-a0a5-4eb182cfa447.mov

## after
https://user-images.githubusercontent.com/557005/133815502-3e2cb4bd-2182-4d49-93ea-8f71bc178965.mov

### DIFF
--- a/opengever/core/testserver.py
+++ b/opengever/core/testserver.py
@@ -6,6 +6,7 @@ from ftw.testing import staticuid
 from ftw.testing.layer import COMPONENT_REGISTRY_ISOLATION
 from opengever.base.interfaces import ISearchSettings
 from opengever.base.model import create_session
+from opengever.core import sqlite_testing
 from opengever.core.solr_testing import SolrReplicationAPIClient
 from opengever.core.solr_testing import SolrServer
 from opengever.core.testing import activate_bumblebee_feature
@@ -27,6 +28,35 @@ import transaction
 SOLR_PORT = os.environ.get('SOLR_PORT', '55003')
 SOLR_CORE = os.environ.get('SOLR_CORE', 'testserver')
 REUSE_RUNNING_SOLR = os.environ.get('TESTSERVER_REUSE_RUNNING_SOLR', None)
+
+
+class SQLiteBackup(object):
+    backup_data = ''
+
+    def isolate(self):
+        if self.backup_data:
+            self.restore()
+        else:
+            self.backup()
+
+    def backup(self):
+        for line in create_session().bind.raw_connection().connection.iterdump():
+            # We only want to store sql statements of the
+            # DML (Data Manipulation Language). Thus, we skip all other statements.
+            #
+            # When we restore the data in the restore-function, we'll empty all
+            # tables and feed it with the backedup data.
+            if 'CREATE TABLE' in line or \
+               'opengever_upgrade_version' in line or \
+               'CREATE INDEX' in line:
+                continue
+            self.backup_data += line
+
+    def restore(self):
+        sqlite_testing.truncate_tables()
+        create_session().bind.raw_connection().connection.executescript(self.backup_data)
+
+sqlite_backup = SQLiteBackup()
 
 
 class TestserverLayer(OpengeverFixture):
@@ -152,6 +182,8 @@ class TestServerFunctionalTesting(FunctionalTesting):
         super(TestServerFunctionalTesting, self).testSetUp()
         self.context_manager = self.isolation()
         self.context_manager.__enter__()
+
+        sqlite_backup.isolate()
         transaction.commit()
 
     def testTearDown(self):
--- a/opengever/core/testserver_selftest.py
+++ b/opengever/core/testserver_selftest.py
@@ -85,6 +85,20 @@ class TestserverSelftest(object):
                  u'email': u'99001@example.org'},
                 browser.json)
 
+            # Make sure the dossier is not yet in the ist of favorites
+            favorites_url = self.plone_url + '/@favorites/' + SITE_OWNER_NAME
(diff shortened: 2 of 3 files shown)


opengever/core/testserver.py
```diff
@@ -1,3 +1,5 @@
+# from opengever.base import model
+# from z3c.saconfig import named_scoped_session
```
jone, line 2: ✂️

opengever/core/testserver.py
```diff
@@ -15,20 +18,56 @@
...
 from requests.exceptions import ConnectionError
 from zope.configuration import xmlconfig
 from zope.globalrequest import setRequest
+from zope.site.hooks import siteinfo
 import imp
 import os
 import pytz
 import transaction
-
+from ZODB.POSException import ConflictError
```
jone, line 30: Müssen die Imports nicht sortiert werden?

opengever/core/testserver.py
```diff
@@ -15,20 +18,56 @@
...
+            if 'CREATE TABLE' in line or \
+               'opengever_upgrade_version' in line or \
+               'CREATE INDEX' in line:
+                continue
+            self.backup_data += line
+
+        self.siteinfo_sm = siteinfo.sm
+
+    def restore(self, retry=0, max_retry=5):
+            try:
```
jone, line 52: Einrückung falsch

opengever/core/testserver.py
```diff
@@ -15,20 +18,56 @@
...
+                sqlite_testing.truncate_tables()
+                create_session().bind.raw_connection().connection.executescript(self.backup_data)
+            except Exception as e:
```
jone, line 55: Was für eine Exception wird hier gecatcht?

Ich verstehe das Problem hier noch zu wenig, was genau passiert. Wie kann ich das reproduzieren?
Dies schaut für mich nach einem Workaround aus, resp. wir verstehen es eben zu wenig. Wenn wir es besser verstehen können wir das vielleicht ohne catch-all verhindern.

Eine These ist dass wir ein Timing-Problem haben wie die Requests vom frontend her kommen, oder ein db-state-problem zwischen commit vom setup und requests auf der regulären DB connection. Am besten schauen wir das zusammen an.

jone (changes requested):

jone:
@elioschmutz ich habe meinen PR #7187 rebased und bereinigt.

opengever/core/testserver_selftest.py
```diff
@@ -110,6 +124,12 @@ def selftest(self):
                   'UID': u'createtemplates00000000000000002'}],
                 browser.json['items'])
 
+            # Make sure that the SQL isolation works by checking whether the favorite we added
+            # beafore teardown/setup is properly removed.
+            browser.open(favorites_url)
+            self.assertNotIn(dossier_oguid, [favorite['oguid'] for favorite in browser.json],
+                             'SQL issolation does not work properly.')
```
jone, line 131:
```suggestion
                             'SQL isolation does not work properly.')
```

opengever/core/testserver_selftest.py
```diff
@@ -110,6 +124,12 @@ def selftest(self):
                   'UID': u'createtemplates00000000000000002'}],
                 browser.json['items'])
 
+            # Make sure that the SQL isolation works by checking whether the favorite we added
+            # beafore teardown/setup is properly removed.
```
jone, line 128:
```suggestion
            # beafore teardown/setup is properly removed.
```
```suggestion
            # before teardown/setup is properly removed.
```

jone (approved):
Found some typos 🙈 , but other than that 👍 

For the record:
This isolation method is probably not very fast (we didn't measure it), but it is simple and works.

If we later need to make it faster, we can look into ideas such as:
- using the ftw.testing transaction interceptor here too; but we'd have to let the HTTP server be single threaded and reuse the same database connection without ever committing (this will be tricky)
- using pool of clean and prepared in-memory databases that we can switch and cleanup asynchronously (this will be tricky too)

Using PostgreSQL here is an option (we have a ancient branch somewhere), but it will not make the restore faster or easier.

elioschmutz:
I measured the speed of the sql-`isolate` method:

- The average speed of rolling back with this method based on 100 rollbacks is: `11ms`
- Currently, we have ~500 e2e tests. Means, the total time for rolling back the sql database when running all tests is: `5.5sec`.
- Running all tests in total takes in average: 35min on mclaren. The rollback of the sql-db would make 0.26% of the total running time.

I think this is fast enough and there is no need to improve the rollback-speed of the sql.

jone:
> * The average speed of rolling back with this method based on 100 rollbacks is: `11ms`

Cool! That's a lot faster than I expected! (I've mesured it years ago with a different approach, taking much longer).

opengever/core/testserver.py
```diff
@@ -29,6 +30,30 @@
...
+
+    def isolate(self):
+        if self.backup_data:
+            self.restore()
+        else:
+            self.backup()
+
+    def backup(self):
+        for line in create_session().bind.raw_connection().connection.iterdump():
+            if 'CREATE TABLE' in line or \
```
lukasgraf, line 44: Could you add a short comment here as to why these statements are skipped? Is the intent here to just skip DDL, and only back up DML?

lukasgraf (approved):
LGTM 👍, just a small request for clarification.

Welcome back from the dungeon, @elioschmutz and @jone 😜

PR #7184, opened 2021-09-20
Propertysheets: Avoid 'RequiredMissing' for empty multiple_choice fie…

lukasgraf:
If an API client specifies `None` as the field value for a `multiple_choice` field, the `CollectionFieldDeserializer` will turn it into `[None]`, which will cause a `RequiredMissing` error.

We therefore replace `None` with an empty list in the propertysheets deserializer.

This for example happens on `dev` for the `fachbereich` field: Since 4teamwork/gever-ui#1926, the frontend now sends

```
  "custom_properties": {
    "IDossier.default": {
      "fachbereich": null,
      "id_fachapplikation": null
    }
  }
```

instead of just `"custom_properties": {}` _(and it can't easily distinguish between multi-valued and single-valued fields, and "no default" vs. "default of `None`")._

For [CA-2482](https://4teamwork.atlassian.net/browse/CA-2482) (already closed, small follow-up bugfix)

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-2482-3.bugfix
+++ b/changes/CA-2482-3.bugfix
@@ -1,0 +1,1 @@
+Propertysheets: Avoid 'RequiredMissing' for empty multiple_choice fields [lgraf]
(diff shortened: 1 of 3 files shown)


njohner (approved):
LGTM

PR #7186, opened 2021-09-22
Filter non-documentish types for IRelatedDocuments.relatedItems backrefs

lukasgraf:
This change makes sure that **on documents**, non-documentish types are also filtered from the `back_references_...` property returned by the REST API, similarly to what already is [being done in the classic UI with `related_items(documents_only=True)`](https://github.com/4teamwork/opengever.core/blob/f7d69a99f17fcf38f94de2cd7f3a8c6c972142f8/opengever/document/browser/overview.py#L269-L282).

For [CA-2936](https://4teamwork.atlassian.net/browse/CA-2936)

## Checklist

- [x] Changelog entry

- New functionality:

### DIFF
--- a/opengever/api/serializer.py
+++ b/opengever/api/serializer.py
@@ -107,7 +107,8 @@ def extend_with_groupurl(result, context, request):
         api.portal.get().absolute_url(), context.groupid)
 
 
-def extend_with_backreferences(result, context, request, reference_attribute_name):
+def extend_with_backreferences(result, context, request, reference_attribute_name,
+                               documents_only=False):
     """Extend the given result dict with an additional key
     `back_references_{reference_attribute_name}` for example
     `back_references_relatedDossiers` and a list of backreferences.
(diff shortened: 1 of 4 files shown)


opengever/api/serializer.py
```diff
@@ -118,6 +118,13 @@ def extend_with_backreferences(result, context, request, reference_attribute_nam
     relations = catalog.findRelations(
         {'to_id': intids.getId(aq_inner(context)),
          'from_attribute': reference_attribute_name})
+
+    # Filter non-documentish types for IRelatedDocuments.relatedItems backrefs
+    if IBaseDocument.providedBy(context):
```
njohner, line 123: I'd find it a bit more explicit if you added an additional parameter `documents_only` to the `extend_with_backreferences` method instead and set it to `True` in the document serializer. But up to you.

njohner (approved):
LGTM

PR #7189, opened 2021-09-26
Introduce testserver `isolate` xmlrpc listener

elioschmutz:
In e2e tests, we have a lot of failing tests because the frontend sends requests to the backend while the backend is in between tearing down and setting up. The backend will raise because the database is not yet ready. E2E tests will the fail and the Z2Server is broken for all upcoming tests. (credits goes to @jone). 

The error is:

`xmlrpclib.Fault: <Fault 1: "<type 'exceptions.KeyError'>:'test-stack-4'">` (example test-run with this error: https://ci.4teamwork.ch/builds/453082/tasks/812702)

Currently, the frontend needs to send two separate requests for the isolation: one for tearing down and a second for setting up the database. We fix this error by introducing a new listener for the xmlrpc-server called `isolate`. This listener is responsible to properly teardown and setup the zodb. We can now use only one single request to make isolation in the zodb possible.

This prevents the frontend to make requests in between.

### DIFF
--- a/base-testserver.cfg
+++ b/base-testserver.cfg
@@ -11,7 +11,6 @@ recipe = zc.recipe.egg
 eggs =
     ${test:eggs}
     plone.app.robotframework
-scripts = robot-server=testserver
 initialization =
     import os
     # Enable conditional readonly patches during tests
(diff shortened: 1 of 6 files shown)


elioschmutz:
closed by: https://github.com/4teamwork/opengever.core/pull/7192

PR #7192, opened 2021-09-27
Make testserver more sturdy

jone:
**The goal is to make the testserver more sturdy, so that it does not break when doing unexpected things such as making HTTP requests while doing isolation (teardown/setup) between tests.**

- _⚠️ currently merging into #7182; rebase needed ⚠️_
- superseding #7189 
- part of: https://4teamwork.atlassian.net/browse/CA-749

 
## Why?

Until now, the testserver was built expecting that the client is well-behaved, does not make any HTTP requests while doing isolation and waits until the testserver is ready again. But fact is that our JavaScript-based frontend is not that well-behaved. It does not wait (long enough) until the setup is finished or may send requests after tear down is initialized.

## Problems

Making HTTP requests when the testserver is not ready can cause problems such as:
- `KeyError: 'test-stack-4'` (access to the component registry when we shouldn't
- Handling exceptions, causing the logging to access `time()` while we are recording a mocker and thus it does not return an `int`
- trying to access the database from a no longer valid database connection

These problems cause the testserver to break in an unrecoverable manner.

## What?

1. **`bin/testerverctl isolate`**
In addition to `bin/testerverctl zodb_setup` and `bin/testerverctl zodb_teardown` we now also have a `bin/testerverctl isolate` method, so that from a client perspective everything necessary is done at once and one transaction. This makes it a lot easier to control from the frontend.
(Thanks @elioschmutz for implementing that in #7189).

2. **Implement a "isolation readyness" concept**
The `IsolationReadyness` component knows whether the testserver has finished setting up and is ready. The component blocks incoming HTTP requests when the testserver is not ready for up to 5 seconds. The sole purpose is to protect the consistency of the testserver and make it more sturdy; the client is responsible for timing the setup and the requests correctly and may have to cope with strange results when requests are executed with the database state of another test case.
When a request gets a isolation timeout (5 seconds), the response will not be answered at all and will keep being pending (I couldn't get it working with a proper error response; but I think this should be the case very often).

3. **Logging**
Since we now have to copy/patch some parts of robotframework for customizations, the testserver now logs to `stderr` whenever a `-v` flag is used. Using more (e.g. `-vv`) increases the log level.

4. **Tests**
I've integrated a case in the selftest, which used to be enough for letting the testserver crash. This case works now, proving that the testserver got sturdier 😉

### DIFF
--- a/opengever/core/testserver_zope2server.py
+++ b/opengever/core/testserver_zope2server.py
@@ -2,6 +2,7 @@ from plone.app.robotframework import server as robotframework_server
 from six.moves.xmlrpc_server import SimpleXMLRPCServer
 import argparse
 import logging
+import sys
 
 
 """To make e2e tests more robust during teardown/setup of the zodb, we have
@@ -115,6 +116,7 @@ def server():
         global HAS_VERBOSE_CONSOLE
         HAS_VERBOSE_CONSOLE = True
         loglevel = logging.ERROR - (args.verbose - 1) * 10
+        logging.getLogger().addHandler(logging.StreamHandler(sys.stderr))
     else:
         loglevel = logging.ERROR
     logging.basicConfig(level=loglevel)
--- a/opengever/core/testserver_zope2server.py
+++ b/opengever/core/testserver_zope2server.py
@@ -6,6 +6,7 @@ import argparse
 import logging
 import os
 import sys
+import threading
 import time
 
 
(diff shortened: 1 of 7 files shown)


elioschmutz (approved):
🎉  works perfectly!

opengever/core/testserver_zope2server.py
```diff
@@ -0,0 +1,234 @@
...
+from ZPublisher import Publish
+import argparse
+import logging
+import os
+import sys
+import threading
+import time
+
+
+"""To make e2e tests more robus when teardown/setup the zodb, we have
```
buchi, line 13:
```suggestion
"""To make e2e tests more robust during teardown/setup of the zodb, we have
```

lukasgraf (approved):
LGTM 👍 🎉

PR #7194, opened 2021-09-29
Add new endpoint @accessible-workspaces

tinagerber:
This endpoint is needed so that an administrator can check which workspaces a user can view.

For [CA-1685]

## Checklist

- [x] Changelog entry

- API change:

[CA-1685]: https://4teamwork.atlassian.net/browse/CA-1685


opengever/core/upgrades/20210928145722_add_permission_to_manage_workspaces/rolemap.xml
```diff
@@ -0,0 +1,8 @@
+<rolemap>
+  <permissions>
+    <permission name="opengever.workspace: Manage Workspaces" acquire="True">
+      <role name="Manager" />
+      <role name="Administrator" />
```
njohner, line 5: I think this might / will lead to problems in some deployments. Although it might not be more problematic than it already is, I can't tell. But when setting up a Teamraum we have the `workspace_administrators_group`, which get the `WorkspaceAdmin` global role. Now we request the `Administrator` role to see admin functionality...

njohner: I think we need to discuss whether we need the `workspace_adminsitrators_group` in the setup or not. Basically I think we simply need an `administrators` group, just as in normal gever. This corresponds to the `tr_administrators` group, or maybe the `tr_admin` group? I'd keep only one of those.

opengever/api/accessible_workspaces.py
```diff
@@ -0,0 +1,61 @@
...
+    def publishTraverse(self, request, name):
+        # Consume any path segments after service name as parameters
+        self.params.append(name)
+        return self
+
+    def reply(self):
+        userid = self.read_params().decode('utf-8')
+        user = api.user.get(userid)
+        if not user:
+            raise BadRequest('Invalid userid "{}".'.format(userid))
```
njohner, line 31: Careful, `userid` is unicode, so raising this exception might fail as is

opengever/api/accessible_workspaces.py
```diff
@@ -0,0 +1,61 @@
...
+
+        allowed_roles_and_users = [u'user:{}'.format(userid)]
+        allowed_roles_and_users.extend(['user:{}'.format(group) for group in user.getGroups()])
```
njohner, line 34: Can groups ids contain non-ASCII chartacters?

opengever/api/tests/test_accessible_workspaces.py
```diff
@@ -0,0 +1,90 @@
...
+            workspace, 'fa_users',
+            ['WorkspaceMember'])
+        self.commit_solr()
+
+        url = '{}/@accessible-workspaces/{}'.format(self.portal.absolute_url(),
+                                                    self.workspace_member.getId())
+        browser.open(url, headers=self.api_headers)
+        self.assertEqual(
+            {u'@id': u'http://nohost/plone/@accessible-workspaces/beatrice.schrodinger',
+             u'items': [{u'@id': u'http://nohost/plone/workspaces/workspace-2',
```
njohner, line 46: u'http://nohost/plone/workspaces/workspace-2' -> `workspace.absolute_url()`

njohner (commented):
A few comments, and I think we really need to discuss the whole Admin role situation in Teamraum

opengever/core/upgrades/20210928145722_add_permission_to_manage_workspaces/rolemap.xml
njohner: As discussed, this is ok, I've created an issue to address the setup issue: https://4teamwork.atlassian.net/browse/CA-2989

opengever/api/accessible_workspaces.py
tinagerber: I added an u

tinagerber:
@njohner Can you have another look?

njohner (approved):
LGTM

PR #7195, opened 2021-09-30
Find search term also within a word in @listing endpoint.

tinagerber:
Until now, only the beginning of a word has been searched.

For [CA-1580]

## Checklist

- [x] Changelog entry

- API change:


njohner (commented):
How about the solr search endpoint (see `catalog_source.py`)?
There we also switch between simple and complex search automatically. Should we rather do something similar for the listing endpoint? In the listing endpoint we are actually filtering not searching. Maybe we can quickly discuss what the expected behavior should be. I'm a bit unwilling to let a simple comment from a user decide for us here (although I would agree that this is an improvement).

njohner:
As discussed, this is a clear improvement, but we should make the same improvement for the solr search endpoint.

PR #7196, opened 2021-10-04
Include document version in Office Connector metadata

buchi:
In Office Connector we want to notify the user if his checked out local copy of a document is older than the current version on the server. For this we need to get the current version.

This situation can occur if the document is not checked-in by Office Connector because of a problem and then manually checked-in by the user.  If meanwhile another user creates a new version and the first user wants to continue working on the current version, OC still uses his local copy.

For [CA-2971]

## Checklist

- [x] Changelog entry

[CA-2971]: https://4teamwork.atlassian.net/browse/CA-2971

### DIFF
--- a/changes/CA-2971.other
+++ b/changes/CA-2971.other
@@ -1,0 +1,1 @@
+Include document version in Office Connector metadata. [buchi]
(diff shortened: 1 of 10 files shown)


njohner (approved):
LGTM

PR #7197, opened 2021-10-04
Bump Products.LDAPMultiPlugins to version 1.15.post4

buchi:
Now groups are filtered case-insensitively by the provided base DN. This makes the configuration less error prone.

## Checklist

- [x] Changelog entry


deiferni (approved):
LGTM 👍

PR #7198, opened 2021-10-04
Truncate filenames when updating favorites.

deiferni:
As the filename column has a length limit we must enforce it by truncating too long data. This adds another layer of sanitation to the already existing id normalizer. The normalizer seems to be a bit lenient when it comes to file extensions potentially allowing very long file extensions. We add strict sanitation for such cases to prevent errors while writing to the database.

For [CA-960](https://4teamwork.atlassian.net/browse/CA-960)

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-960.bugfix
+++ b/changes/CA-960.bugfix
@@ -1,0 +1,1 @@
+Truncate too long filenames when updating favorites. [deiferni]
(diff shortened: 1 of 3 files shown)


lukasgraf (approved):
LGTM 👍

PR #7204, opened 2021-10-08
Fix soft-delete for inbox documents.

elioschmutz:
This PR fixes an issue where the soft-delete of inbox documents does not work.

It's because we filter by workflow actions with a category of `workflow`: https://github.com/4teamwork/opengever.core/blob/master/opengever/api/transition.py#L43

We somehow forgot to set this category for the inbox-workflows.

For [CA-3021]

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

[CA-3021]: https://4teamwork.atlassian.net/browse/CA-3021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/api/tests/test_workflow.py
```diff
@@ -73,6 +73,30 @@ def test_transition_remove_document_and_revoking_users_permission_can_be_execute
...
+    def test_transition_remove_document_can_be_executed(self, browser):
+        self.login(self.manager)
+        # allow `secretariat_user` to remove gever content
+        self.portal.manage_permission(
+            'Remove GEVER content', roles=['Editor'], acquire=True)
+
+        self.login(self.secretariat_user, browser=browser)
+        ITrasher(self.inbox_document).trash()
+
+        # this soft-deletes the document so that `secretariat_user` no longer has
```
njohner, line 86: I think this is a copy-paste. You can ✂️ this comment IMO

PR #7205, opened 2021-10-11
Update reference number and searchable text when moving documents and dossiers.

njohner:
The reference number did not get reindexed when moving a document. I think this was more or less conscious as it wasn't displayed anywhere back then, and that reindexing the SearchableText would have been too expensive. Now that the searchable text is only in solr and split up in text from the file and `metadata` it's efficient to update.

Moreover I also found that the SearchableText was not updated when moving a dossier. I've corrected that as well and added some more tests for update of indexes, metadata and solr data when moving dossiers and documents.

For [CA-3018]

## Checklist

- [x] Changelog entry

[CA-3018]: https://4teamwork.atlassian.net/browse/CA-3018?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/dossier/handlers.py
+++ b/opengever/dossier/handlers.py
@@ -81,7 +81,7 @@ def save_reference_number_prefix(obj, event):
     if not prefix_adapter.get_number(obj):
         prefix_adapter.set_number(obj)
 
-    obj.reindexObject(idxs=['reference', 'sortable_reference'])
+    obj.reindexObject(idxs=['reference', 'sortable_reference', 'SearchableText'])
 
 
 def reindex_containing_subdossier_for_contained_objects(dossier, event):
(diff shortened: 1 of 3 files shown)


opengever/dossier/handlers.py
```diff
@@ -81,7 +81,7 @@ def save_reference_number_prefix(obj, event):
     if not prefix_adapter.get_number(obj):
         prefix_adapter.set_number(obj)
 
-    obj.reindexObject(idxs=['reference', 'sortable_reference'])
+    obj.reindexObject(idxs=['reference', 'sortable_reference', 'SearchableText'])
```
phgross, line 84: Is the separation between `metadata` and `SearchableText` only for documents but not for dossiers? or why does `SearchableText` needs to be updated for dossiers? 🤔

PR #7208, opened 2021-10-13
Correctly set IDs for objects created over the API and when delegating tasks.

njohner:
With this PR we fix two issues where the IDs of created objects were not as expected:
1. For several portal types the IDs attributed when creating objects differed  when creation happened over the API or in the classic UI. 
2. The IDs for the subtasks created when delegating a task (both over the API and in the classic UI) did not respect the intended schema `task-1`.

These two issues were actually unrelated. In the first case, the issue was that a few portal types provided the `INameFromTitle` behavior, which actually does not provide an implementation of so that `INameFromTitle(obj).title` would be `None`. This led to [different defaults](https://github.com/plone/plone.app.content/blob/master/plone/app/content/namechooser.py#L35-L38) kicking in for the API than for the classic UI. For the API, an `id` is provided in the request, which is then used, while for the classic UI, it defaulted to the `portal_type`. We have now fixed this issue by adding a default `IDefaultNameFromTitle` behavior, returning the `portal_type` of the object, i.e. behaving as was the case until now in the classic UI.

The second issue simply came from actually passing the `id` as parameter to `createContent` when creating the subtasks when delegating a task. This led to bypassing the `ITaskNameFromTitle` behavior an using the passed `id` instead.

For [CA-1908]

## Checklist
- [x] Changelog entry

[CA-1908]: https://4teamwork.atlassian.net/browse/CA-1908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-1908-2.bugfix
+++ b/changes/CA-1908-2.bugfix
@@ -1,0 +1,1 @@
+Do not bypass NameFromTitle for subtasks created when delegating a task. [njohner]
--- a/changes/CA-1908.bugfix
+++ b/changes/CA-1908.bugfix
@@ -1,0 +1,1 @@
+Correct IDs for ContactFolders, InboxContainers, Inboxes, CommitteeContainers and TemplateFolders created over the API. [njohner]


opengever/core/upgrades/20211013081350_replace_name_from_title_behavior_with_default_implementation/upgrade.py
```diff
@@ -0,0 +1,9 @@
+from ftw.upgrade import UpgradeStep
+
+
+class ReplaceNameFromTitleBehaviorForTemplateFolders(UpgradeStep):
```
phgross, line 4: Do we need to update the `object_provides` index for those objects 🤔 ?

PR #7212, opened 2021-10-14
Use UIDs as tokens in attachable_documents_vocabulary.

njohner:
Because the frontend does not know the`IntId`s of the documents it wants to link to the created subtasks when delegating a task, we modify the corresponding vocabulary to use `UID`s as tokens instead. We leave the values as `IntIds` as this is necessary to create the relations in the backend afterwards.

I guess this is a breaking change...

For [CA-2358]

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [x] api-change label added

[CA-2358]: https://4teamwork.atlassian.net/browse/CA-2358?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-2358.other
+++ b/changes/CA-2358.other
@@ -1,0 +1,1 @@
+Use UIDs as tokens for documents when delegating a task. [njohner]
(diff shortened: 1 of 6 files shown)


deiferni (approved):
LGTM 👍

Is any of our apps affected by this change?

PR #7210, opened 2021-10-14
Implement webactions with scope context.

tinagerber:
Until now, there were only global webactions. Now there are also webactions with the scope `context`. Such a webaction is only displayed on an object if it has been explicitly activated for this object.

For [CA-2860]

## Checklist

_Everything has to be done/checked. Checked but not present means the author deemed it unnecessary._

- [x] Changelog entry
- [x] Link to issue (Jira or GitHub) and backlink in issue (Jira)

_Only applicable should be left and checked._

- API change:
  - [x] Documentation is updated
  - [x] API Changelog entry (see [guide](https://4teamwork.atlassian.net/wiki/spaces/4TEAM/pages/451248812/API+Changelog+Guidelines))

[CA-2860]: https://4teamwork.atlassian.net/browse/CA-2860?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7213, opened 2021-10-15
Allow activating and deactivating web actions that are not owned by the user

tinagerber:
Follow up for #7210 

`_check_ownership` must be overridden, otherwise an Unauthorized will be thrown if a user other than the owner tries to activate or deactivate the webaction.

For [CA-2860]

[CA-2860]: https://4teamwork.atlassian.net/browse/CA-2860?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/api/webactions.py
+++ b/opengever/api/webactions.py
@@ -250,6 +250,11 @@ class ContextWebActionsPost(WebActionLocator):
         self.request.response.setStatus(204)
         return _no_content_marker
 
+    def _check_ownership(self, action):
+        # Must be overridden because users can activate a web action
+        # even if they are not the owner.
+        return
+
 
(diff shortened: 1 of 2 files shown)


opengever/api/webactions.py
```diff
@@ -250,6 +250,9 @@ def reply(self):
         self.request.response.setStatus(204)
         return _no_content_marker
 
+    def _check_ownership(self, action):
```
njohner, line 253: Maybe add a comment as to why this is overwritten here

njohner (approved):
A comment in the code would be helpful I think, otherwise LGTM

PR #7215, opened 2021-10-18
Update sortable_reference and SearchableText for all childs when reference_prefix has changed.

phgross:
Currently only the `reference` index has been updated, but the recently introduced `sortable_reference` index was not updated yet. This PR fixes that. 

Besides that I realized that the reference number is also part of the SearchableText respectively Metadata index, therefore we also need to reindex those if a reference_prefix is changed.

For [CA-3042]

## Checklist
- [x] Changelog entry

[CA-3042]: https://4teamwork.atlassian.net/browse/CA-3042?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-3042.bugfix
+++ b/changes/CA-3042.bugfix
@@ -1,1 +1,1 @@
-Also update sortable_reference index when reference_prefix has changed. [phgross]
+Also update sortable_reference and SearchableText when reference_prefix has changed. [phgross]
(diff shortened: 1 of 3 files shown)


njohner (approved):
LGTM

PR #7216, opened 2021-10-18
Make workspace invitation mail content customizable.

phgross:
This PR adds an additional registry entry, to customize the invitation mail content. 

The setting is empty by default, therefore the default mail content (which also provides translations) is used. The customization flag does not support translations.

The mail content is customized only in exceptional cases, therefore the flag is not set in the policy generator.

For [CA-3038]

## Checklist

- [x] Changelog entry

[CA-3038]: https://4teamwork.atlassian.net/browse/CA-3038?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/workspace/participation/invitation_mailer.py
+++ b/opengever/workspace/participation/invitation_mailer.py
@@ -1,8 +1,11 @@
 from opengever.activity.mailer import Mailer
 from opengever.ogds.base.actor import ActorLookup
 from opengever.ogds.base.utils import get_current_admin_unit
+from opengever.workspace.interfaces import IWorkspaceSettings
 from opengever.workspace.participation import serialize_and_sign_payload
+from plone import api
 from plone.app.uuid.utils import uuidToObject
+from Products.CMFPlone.utils import safe_unicode
 from Products.Five.browser.pagetemplatefile import ViewPageTemplateFile
 from zope.i18n import translate
 from zope.i18nmessageid import MessageFactory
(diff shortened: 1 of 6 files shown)


opengever/workspace/participation/invitation_mailer.py
```diff
@@ -41,22 +43,33 @@ def send_invitation(self, invitation):
...
+            'title': target_workspace.title,
+            'user': inviter.get_label(with_principal=False),
+            'platform': get_current_admin_unit().public_url,
+            'accept_url': accept_url}
+
+        custom_mail_content = api.portal.get_registry_record(
+            name='custom_invitation_mail_content',
+            interface=IWorkspaceSettings)
+        if custom_mail_content:
+            content = custom_mail_content.format(**content_variables)
```
deiferni, line 57: how graceful will this fail variables are not used or used wrong? should we add a fallback of some sort? (It is fine for me if we don't and you merge this as is but still i'd like to raise the question).

opengever/workspace/tests/test_invitation.py
```diff
@@ -85,3 +87,41 @@ def test_adding_invitation_sends_mail(self):
 
         link = "{}/@@my-invitations/accept?invitation={}".format(self.workspace.absolute_url(), payload)
         self.assertIn(link, mail.as_string().decode('quoted-printable'))
+
+    def test_use_customized_mail_content_if_configured(self):
+        self.login(self.workspace_admin)
+
+        custom_mail_template = u"""
```
deiferni, line 94: maybe add a test with only a subset of the variables to address above question?

opengever/workspace/participation/invitation_mailer.py
```diff
@@ -41,22 +43,33 @@ def send_invitation(self, invitation):
...
+        content_variables = {
+            'title': target_workspace.title,
+            'user': inviter.get_label(with_principal=False),
+            'platform': get_current_admin_unit().public_url,
+            'accept_url': accept_url}
+
+        custom_mail_content = api.portal.get_registry_record(
+            name='custom_invitation_mail_content',
+            interface=IWorkspaceSettings)
+        if custom_mail_content:
```
deiferni, line 56: is this guaranteed unicode or could we run into encoding issues?

opengever/workspace/participation/invitation_mailer.py
phgross: Not used variables are ignored by string formatting and works like expected (I adjusted my test to also test this behavior). If you try to use a key which is not supported, it will raise, but thats what you expect in my opinion. The custom_mail flag is under control, so if you did something wrong, you want to be informed about it via sentry, so you can fix it. 

Switching back to the custom mail would be wrong IMO.

PR #7223, opened 2021-10-28
Allow to include subtasks in task reports

tinagerber:
This PR introduces a new parameter `include_subtasks` to include subtasks in task reports. The parameter is only used in the new frontend, because in the listings of the classic frontend, unlike the listings of the new frontend, subtasks are displayed.

I decided to sort the subtasks by path so that subsubtasks appear below the correct subtask. This has the consequence that for sequential tasks the order is not necessarily correct. But since determining the correct order for sequential tasks would require retrieving the Plone task, I chose not to include this.

For [CA-2844]

## Checklist

- [x] Changelog entry

[CA-2844]: https://4teamwork.atlassian.net/browse/CA-2844?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-2844.feature
+++ b/changes/CA-2844.feature
@@ -1,0 +1,1 @@
+Allow to include subtasks in task reports. [tinagerber]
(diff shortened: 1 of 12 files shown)


lukasgraf (approved):
LGTM 👍 
I also tested it locally (for Excel + PDF, old and new UI), and it works as expected. 🎉 
The ordering edge case with subtasks + sequential tasks is ok like that in my opinion.

PR #7227, opened 2021-11-01
Add validation of the end date for meetings

tinagerber:
Since the end date is validated in the meetings app, this repeatedly led to import errors. Therefore, the end date is now also validated in Gever.

The end date can now no longer be before the start date. For meetings that have an invalid end date set, the end date is set to None.

For [CA-3113]

## Checklist

- [x] Changelog entry

[CA-3113]: https://4teamwork.atlassian.net/browse/CA-3113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/core/upgrades/20211028173917_fix_committee_meeting_end_date/upgrade.py
+++ b/opengever/core/upgrades/20211028173917_fix_committee_meeting_end_date/upgrade.py
@@ -1,0 +1,13 @@
+from opengever.core.upgrade import SchemaMigration
+from opengever.meeting.model.meeting import Meeting
+
+
+class FixCommitteeMeetingEndDate(SchemaMigration):
+    """Fix committee meeting end date.
+    """
+
+    def migrate(self):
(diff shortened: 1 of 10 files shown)


opengever/core/upgrades/20211028173917_fix_committee_meeting_end_date/upgrade.py
```diff
@@ -0,0 +1,13 @@
+from opengever.core.upgrade import SchemaMigration
+from opengever.meeting.model.meeting import Meeting
+
+
+class FixCommitteeMeetingEndDate(SchemaMigration):
```
njohner, line 5: This should not be a `SchemaMigration` IMO. This is a standard `SQLUpgradeStep`

njohner (approved):
See comment. Otherwise LGTM

PR #7230, opened 2021-11-02
Nj/ca 2367/persist defaults

njohner:
In this PR we use nightly maintenance jobs to persist missing default values for the `IClassification` fields and `preserved_as_paper`, which were not persisted for objects created over the REST-API. While we have fixed [the underlying issue](https://github.com/4teamwork/opengever.core/pull/7041), we had not persisted the missing static defaults, as the upgrade would have been very costly without the nightly maintenance jobs. 

For [CA-2367]

## Checklist
- [x] Changelog entry


opengever/core/upgrades/20211026132728_persist_missing_default_values/upgrade.py
```diff
@@ -0,0 +1,44 @@
...
+classification = IClassification.get("classification")
+privacy_layer = IClassification.get("privacy_layer")
+public_trial = IClassification.get("public_trial")
+preserved_as_paper = IDocumentMetadata.get("preserved_as_paper")
+
+
+class PersistMissingDefaultValues(UpgradeStep):
+    """Persist missing default values.
+    """
+
```
phgross, line 17: IMHO this upgradestep should be deferrable. I know, the potential long running job will be done over night, but still iterating over all objects will take some time and its not absolutely essential.

phgross (changes requested):
LGMT, just the thing mentioned.

njohner:
@phgross did you want another look or can I merge? It was just the missing `deferrable` flag...

phgross (approved):

PR #7232, opened 2021-11-03
Implement a two state workflow for todos

elioschmutz:
This PR removes the `completed` field from workspace todos and implement a new two state workflow which provides an `active` and a `completed` state.

To make toggling of the todo-state in the frontend easier, we introduce a new `@toggle` endpoint to toggle the state. 

For [CA-1679]

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [x] api-change label added
- New functionality:
- DB-Schema migration
- New translations

[CA-1679]: https://4teamwork.atlassian.net/browse/CA-1679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/core/upgrades/20211105114219_add_completed_state_for_todos/workflows.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0"?>
+<object name="portal_workflow" meta_type="Plone Workflow Tool">
+
+    <object name="opengever_workspace_todo" meta_type="Workflow" remove="True" />
```
phgross, line 4: 🤔

opengever/testing/fixtures.py
```diff
@@ -2278,9 +2276,9 @@ def create_todos(self):
             .having(
                 text=u"Do some cleanups",
                 deadline=date(2016, 9, 2),
-                responsible=self.workspace_member.getId(),
-                completed=True)
+                responsible=self.workspace_member.getId())
             .within(self.todolist_introduction)
+            .in_state('opengever_workspace_todo--STATUS--completed')
```
phgross, line 2281: Use here `COMPLETED_TODO_STATE`

changes/CA-1679-2.other
```diff
@@ -0,0 +1 @@
+Introduce '@toggle' endpoint for workspace todos. [elioschmutz]
```
phgross, line 1: I do not understand, why a `@toggle` endpoint is necessary? 

Toggling a state often only leads to strange behavior in the UI, if there are errors or slow connections. Why is the endpoint necessary respectively why don't you just use the existing @transition endpoint?

elioschmutz: Because it makes a lot easier in the frontend. We don't have to know something about workflow transitions nor with all the errors that can happen if using the wrong transition. If we would do it without the `@toggle` endpoint, we would have to know the possible transitions and states to be able to toggle a todo over the `@transition` endpoint. Doing it generic by fetching the possible transitions would be even worse.

In other listings, we do a full context request if moving over a list item to get all transitions. Doing this for todos would just reduce usability and we would increase complexity for a non complex feature.

You have different options in the frontend to not make it strange.

So I thing it's better to have a `@toggle` endpoint than doing it directly over a transition.

docs/public/dev-manual/api/workspace/todos.rst
```diff
@@ -87,6 +87,32 @@ Ein erledigtes ToDo wieder öffnen
           "...": "..."
       }
 
+Ein ToDo-Status umdrehen (togglen)
```
phgross, line 90: `umdrehen` passt irgendwie nicht, ich würde wohl `Ein ToDo-Status togglen (öffnen/abschliessen)` schreiben.

opengever/api/todo.py
```diff
@@ -79,3 +80,9 @@ def _is_todo_list(self, obj):
 
     def _get_changes_text(self, obj):
         return obj.title if IToDoList.providedBy(obj) else ''
+
+
+class ToggleToDoCompleted(ContentPatch):
+    def reply(self):
+        self.context.toggle()
+        return super(ToggleToDoCompleted, self).reply()
```
phgross, line 88: I suspect this would allow me to change other todo metadata by using the @toggle endpoint:
```
        response = browser.open(
            self.todo, view="@toggle", method='PATCH', data={'title': 'neuer Titel'}
            headers={'Accept': 'application/json', 'Prefer': 'return=representation'}).json
```

is that a problem?

changes/CA-1679-2.other
phgross: ok ...

phgross (approved):
👍 besides the things mentioned.

opengever/core/upgrades/20211105114219_add_completed_state_for_todos/workflows.xml
elioschmutz: removed 👍

opengever/api/todo.py
elioschmutz: oh 👀  good point! That should not be possible, even if it's not a problem itself. Thanks 👍

elioschmutz: I updated the commit and added a test. I do no longer accept a body for the `@toggle` endpoint.

tinagerber: @elioschmutz I think you should use POST here

elioschmutz: @tinagerber sure? I was thinking about using POST as well, but the definition of POST creating a resource and PATCH is updating a resource. Since I just update it, I need PATCH, no?

elioschmutz: I'll update and will use `POST` instead `PATCH` since we use `POST` for all other action-endpoints

elioschmutz: @njohner what do you think? `POST` or `PATCH` for the `@toggle` endpoint? `PATCH` provides also the `Prefer` header which I want to use. So I prefer using `PATCH` over `POST`

njohner: Well, my vote would go against the toggle endpoint, and use the workflow transition instead. That being said, I'd also go for POST here. `POST` is not only used for creation IMO, but for all other endpoints that do something else than modify the metadata of a resource. For example workflow transition is a POST, `checkout`, `checkin`, `trash`, `@link-to-workspace`, etc. I don't think we have any dedicated endpoint that is a `PATCH`, do we?

njohner: To be honest I do not have a strong opinion against `@toggle`...

elioschmutz: Ok, thanks @njohner and @tinagerber . Because @tinagerber argued the same way as you @njohner, I'll go with the ?`POST` endpoint. The `@toggle` endpoint I'd like to keep.

PR #7235, opened 2021-11-05
Extend API error representation

phgross:
To provide translated and specific error message in the GEVER frontend, we need to extend the default error representation. This PR adds a Custom API errorHandling,  to return also a translated error message and further translated informations (ValidationErrors) if possible.

The custom handling only adds keys, so does not change the existing error format. A little example ...
_before:_
```json
{
  "message": "[{'fields': '' ... }]", 
  "type": "ValidationError",
}
```

_after:_
```json
{
  "message": "[{'fields': '' ... }]",
  "translated_message": "Validierung fehlgeschlagen"
  "type": "ValidationError",
  "additional_data:": {
    "fields": [
      {
        "field_name": "receipient_mail",
        "type": "RequiredMissing",
        "translated_message": "Fehlende Eingabe"
      }
    ]
  }
}
```

Furthermore the PR changes the error raising at some places and raise an Exception with a MessageFactory object instead of plain string message. So the CustomHandling automatically translates the message and the translated error message is uses by the gever-ui (see https://github.com/4teamwork/gever-ui/pull/2020).

For [CA-408]

## Checklist
- [x] Changelog entry

- API change:
- New translations

[CA-408]: https://4teamwork.atlassian.net/browse/CA-408?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -8,6 +8,7 @@ API Changelog
 
 Breaking Changes
 ^^^^^^^^^^^^^^^^
+- Some error messages have been renamed, but the format how an error is returned stays the same, only the response now usually contains a translated error message and may contain additional metadata.
 - Toggling a Workspace Todos review state from active to completed and back can be done thorugh the newly introduced `@toggle` endpiont for todos. [elioschmutz]
 - Workspace Todos do no longer provide a completed-field. Completing a todo is now done through a workflow transition. [elioschmutz]
 - The ``completed`` field in the ``@listing`` is now longer supported, use the ``is_completed`` field instead.
--- a/opengever/api/deserializer.py
+++ b/opengever/api/deserializer.py
@@ -1,0 +1,72 @@
+from Acquisition import aq_inner
+from Acquisition import aq_parent
+from opengever.base.default_values import set_default_values
+from opengever.base.interfaces import IOpengeverBaseLayer
+from plone.dexterity.interfaces import IDexterityContent
+from plone.restapi.deserializer import json_body
+from plone.restapi.deserializer.dxcontent import DeserializeFromJson
+from plone.restapi.interfaces import IDeserializeFromJson
+from z3c.form.interfaces import IManagerValidator
+from zExceptions import BadRequest
+from zope.component import adapter
+from zope.component import queryMultiAdapter
+from zope.event import notify
+from zope.interface import implementer
+from zope.lifecycleevent import Attributes
+from zope.lifecycleevent import ObjectModifiedEvent
+
+
+@implementer(IDeserializeFromJson)
+@adapter(IDexterityContent, IOpengeverBaseLayer)
+class GeverDeserializeFromJson(DeserializeFromJson):
+    """Customization of the default deserializer, to return the message
+    factory object instead of the error message string, if a validation error
+    occurs.
+    It also makes sure to prefill values that were not set with default or
+    missing values.
+    """
+
+    def __call__(self, validate_all=False, data=None, create=False):
+
+        if data is None:
+            data = json_body(self.request)
+
+        # set default values
+        if create:
+            container = aq_parent(aq_inner(self.context))
+            set_default_values(self.context, container, data)
+
+        schema_data, errors = self.get_schema_data(data, validate_all)
+
+        # Validate schemata
+        for schema, field_data in schema_data.items():
+            validator = queryMultiAdapter(
+                (self.context, self.request, None, schema, None), IManagerValidator
+            )
+            for error in validator.validate(field_data):
+                errors.append({"error": error, "message": error.message})
+
+        if errors:
+            # Drop Python specific error classes in order to be able to better handle
+            # errors on front-end
+            for error in errors:
+                error["error"] = "ValidationError"
+
+            raise BadRequest(errors)
+
+        # We'll set the layout after the validation and even if there
+        # are no other changes.
+        if "layout" in data:
+            layout = data["layout"]
+            self.context.setLayout(layout)
+
+        # OrderingMixin
+        self.handle_ordering(data)
+
+        if self.modified and not create:
+            descriptions = []
+            for interface, names in self.modified.items():
+                descriptions.append(Attributes(interface, *names))
+            notify(ObjectModifiedEvent(self.context, *descriptions))
+
+        return self.context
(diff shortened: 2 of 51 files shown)


opengever/api/locales/en/LC_MESSAGES/opengever.api.po
```diff
@@ -0,0 +1,120 @@
...
+
+#. Default: "Copy of ${title}"
+#: ./opengever/api/copy_.py
+msgid "copy_of"
+msgstr "Copy of ${title}"
+
+#. Default: "It's not possible to have non-.docx documents as proposal documents."
+#: ./opengever/api/tus.py
+msgid "error_proposal_document_type"
+msgstr "It's not possible to have non-.docx documents as proposal documents."
```
njohner, line 25: Maybe easier: "Only docx documents are allowed as proposal documents."

opengever/api/locales/en/LC_MESSAGES/opengever.api.po
```diff
@@ -0,0 +1,120 @@
...
+
+#. Default: "Maximum dossier depth exceeded"
+#: ./opengever/api/upload_structure.py
+msgid "msg_max_dossier_depth_exceeded"
+msgstr "Maximum dossier depth exceeded"
+
+#. Default: "It's not possible to have no file in proposal documents."
+#: ./opengever/api/document.py
+msgid "msg_needs_file_in_proposal_document"
+msgstr "It's not possible to have no file in proposal documents."
```
njohner, line 75: Proposal documents require a file?

opengever/api/deserializer.py
```diff
@@ -0,0 +1,72 @@
...
+from zope.interface import implementer
+from zope.lifecycleevent import Attributes
+from zope.lifecycleevent import ObjectModifiedEvent
+
+
+@implementer(IDeserializeFromJson)
+@adapter(IDexterityContent, IOpengeverBaseLayer)
+class GeverDeserializeFromJson(DeserializeFromJson):
+    """Customization of the default deserializer, to return the message
+    factory object instead of the error message string, if an validation error
```
njohner, line 23: if an validation -> if a validation

opengever/api/deserializer.py
```diff
@@ -0,0 +1,72 @@
...
+    occurs.
+    It also makes sure not set values are prefilled with default or missing
```
njohner, line 25: It also makes sure to prefill values that were not set with default or missing values.

opengever/api/deserializer.py
njohner, line 23: if an validation -> if a validation

opengever/api/errors.py
```diff
@@ -11,11 +11,23 @@ class GeverErrorHandling(ErrorHandling):
 
     def render_exception(self, exception):
         result = super(GeverErrorHandling, self).render_exception(exception)
+        message = exception.message
 
         # Validation errors
-        if isinstance(exception.message, list):
+        if isinstance(message, list):
+            # Special handling for invitations etc.
+            if len(message) and isinstance(message[0], tuple):
```
njohner, line 19: It would be much nicer if we could rely on the error type here. These heuristics do not seem very maintainable 😅

opengever/api/errors.py
```diff
@@ -11,11 +11,23 @@ class GeverErrorHandling(ErrorHandling):
...
+                return result
+
+            # Handle TUS error
+            elif len(message) and not message[0].get('field'):
```
njohner, line 23: Ok this helps with robustness of the handling here

docs/public/dev-manual/api/api_changelog.rst
```diff
@@ -12,6 +12,7 @@ Breaking Changes
 
 Other Changes
 ^^^^^^^^^^^^^
+- The response returned by the API provides now in the most of the cases a translated error message.
```
njohner, line 15: - When an error is raised, the response now usually contains a translated error message and may contain additional metadata.

njohner (approved):
Overall LGTM 🎉 
I'm just a bit worried that the `GeverErrorHandling` might itself fail under certain cirumstances. I'd really try to make it as safe as possible. Also the heuristics in there are a bit of a mess, but I'm not sure whether this can be improved with a reasonable amount of work.

opengever/api/deserializer.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def __call__(self, validate_all=False, data=None, create=False):
+
+        if data is None:
+            data = json_body(self.request)
+
+        # set default values
+        if create:
+            container = aq_parent(aq_inner(self.context))
+            set_default_values(self.context, container, data)
```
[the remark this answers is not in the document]
phgross: I'm not sure why we patched instead of overwritten the Deserializer, maybe we can chat tomorrow ...

opengever/api/errors.py
phgross: This was my original variant, but unfortunately it is not possible, because most of the errors from the plone.restapi have already been rewritten to a BadRequest.

opengever/api/tests/test_participation.py
```diff
@@ -506,13 +506,10 @@ def test_cannot_delete_the_last_workspace_admin_role_assignment(self, browser):
                 method='DELETE',
                 headers=http_headers()
             )
-        self.assertEqual(
-            browser.json,
-            {
-                u'message': u'At least one principal must remain admin.',
```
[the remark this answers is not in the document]
phgross: Yes indeed, some error message will change but the "schema" how the error is returned will stay the same. I think that ist enough. But maybe we should move the API changelog to the breaking section and extend it with a little hint. What do you think?

opengever/api/deserializer.py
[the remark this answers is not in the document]
phgross: Checked, works without the Patch! 🎉

phgross:
Hi @njohner thanks for the review. I unterstand your concerns, I decided to move the error heuristic and specific handling in to a separate method and put it into a try/except, so we can make sure, that the error handling fails by extending the error message. Besides that it makes the heuristic a bit less complex.

Could you take another look?

opengever/api/tests/test_participation.py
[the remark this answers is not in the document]
njohner: Yeah I would. Just because parsing the message was the only option external applications had to provide translated error messages, so it is not unlikely that it has been done.

opengever/api/deserializer.py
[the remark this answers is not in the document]
njohner: Cool perfect.

opengever/api/errors.py
njohner: Ok then, and at least it's commented, so the intention is at least clear.

PR #7236, opened 2021-11-05
Make workspace meeting start and end timezone aware.

njohner:
Start and end dates for `WorkspaceMeeting`s are sent as UTC from the frontend, but were stored as timezone naive on the objects. This leads to localization of the datetime when generating the PDF for the meeting to show the UTC start and end times. I chose to save the fields as localized datetimes on the objects instead.

Note that `UTCDatetime` fields are already stored as timezone aware in the classic UI, so would probably have been broken without the deserializer. As they were used only for meetings, which are not implemented in the new UI, this problem was not noticed before.

The upgrade step uses somewhat of a hack. By reindexing the full object we change the modified date, so that when checking whether the metadata has changed, we never reach the comparison with previous value for the start and end Fields, which would fail because we are comparing timezone naive and aware datetimes.

For [CA-3150]

## Checklist

- [x] Changelog entry
- API change:
  - If breaking:
    - [ ] api-change label added

[CA-3150]: https://4teamwork.atlassian.net/browse/CA-3150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-3150.bugfix
+++ b/changes/CA-3150.bugfix
@@ -1,0 +1,1 @@
+Make workspace meeting start and end timezone aware. [njohner]
--- a/docs/schema-dumps/opengever.workspace.meeting.schema.json
+++ b/docs/schema-dumps/opengever.workspace.meeting.schema.json
@@ -16,14 +16,14 @@
             "title": "Beginn",
             "format": "datetime",
             "description": "",
-            "_zope_schema_type": "Datetime"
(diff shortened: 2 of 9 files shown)


phgross (approved):
👍 LGTM

PR #7239, opened 2021-11-09
Add flags to office connector url for attach action.

tinagerber:
⚠️  Requires new OfficeConnector version.

The flags are used to tell the OfficeConnector which things are not needed in the mail. The interpretation of the flags is done in the OfficeConnector.

For [CA-1815]

## Checklist

- [x] Changelog entry

[CA-1815]: https://4teamwork.atlassian.net/browse/CA-1815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/officeconnector/helpers.py
+++ b/opengever/officeconnector/helpers.py
@@ -131,6 +131,33 @@ def get_auth_plugin(context):
     return plugin
 
 
+def _parse_flag(request, key):
+    return request.form.get(key, 'true').lower() not in ('n', 'no', 'f', 'false', 'off', '0')
+
+
+def parse_flags(request, action):
+    if action != 'attach':
+        return
+
+    flags = {}
+    if request['REQUEST_METHOD'] == 'GET':
+        flags['attach'] = _parse_flag(request, 'attach')
+        flags['bcc'] = _parse_flag(request, 'set_bcc')
+        flags['links'] = _parse_flag(request, 'links')
+    elif 'BODY' in request:
+        try:
+            data = json.loads(request['BODY'])
+        except ValueError:
+            data = {}
+
+        flags['attach'] = data.get('attach', True) is not False
+        flags['bcc'] = data.get('set_bcc', True) is not False
+        flags['links'] = data.get('links', True) is not False
+
+    flag_list = ['-{}'.format(key) for key, value in flags.items() if not value]
+    return ",".join(flag_list)
+
+
 def create_oc_url(request, context, payload):
     auth_plugin = get_auth_plugin(context)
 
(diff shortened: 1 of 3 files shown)


opengever/officeconnector/helpers.py
```diff
@@ -131,6 +131,28 @@ def get_auth_plugin(context):
     return plugin
 
 
+def parse_flags(request, action):
+    if action != 'attach':
+        return
+
+    flags = {}
+    if request['REQUEST_METHOD'] == 'GET':
+        flags['attach'] = json.loads(request.form.get('attach', 'true'))
```
buchi, line 140: Parsing a boolean value with `json.loads()`  seems not appropriate to me. It's also easy to produce an error by providing an invalid value. I would use something like `request.form.get('attach', 'true').lower() in ['true', '1', 'yes', 'on']`

opengever/officeconnector/helpers.py
```diff
@@ -131,6 +131,28 @@ def get_auth_plugin(context):
...
+        flags['bcc'] = json.loads(request.form.get('set_bcc', 'true'))
+        flags['links'] = json.loads(request.form.get('links', 'true'))
+    else:
+        data = json.loads(request['BODY'])
```
buchi, line 144: Handle `ValueError` if body contains invalid data.

opengever/officeconnector/helpers.py
```diff
@@ -131,6 +131,28 @@ def get_auth_plugin(context):
...
+        flags['attach'] = data.get('attach', True)
+        flags['bcc'] = data.get('set_bcc', True)
+        flags['links'] = data.get('links', True)
+
+    flag_list = []
+    for key, value in flags.items():
+        if not value:
+            flag_list.append('-{}'.format(key))
```
buchi, line 152: you could use a list comprehension which seems more elegant to me, but this is just a personal preference:
`['-{}'.format(key) for key, value in flags.items() if not value]`

buchi (changes requested):
LGTM, but the parsing of the flags should be more robust. Either fallback to the default if bad data is provided or return a BadRequest response.

opengever/officeconnector/helpers.py
tinagerber: This would result in invalid data being classified as false, which is not the default. I now check the false values instead.

tinagerber:
@buchi parsing of the flags is now more robust, can you have another look?

opengever/officeconnector/helpers.py
buchi: Hmm, I don't get your point. If there's invalid data in the body then `json.loads(request['BODY'])` will raise a `ValueError` which will result in a 500 server error. Thus I would catch the `ValueError`:
```python
try:
    data = json.loads(request['BODY'])
except ValueError:
    data = {}
```

buchi (approved):

PR #7240, opened 2021-11-11
Update local roles after moving document when creating forwarding.

njohner:
Somehow this fixes an issue where creating a forwarding leads to the documents being moved into the forwarding appearing twice in the catalog: once with the old path and once with the new. It seems to have something to do with reindexing and event order, but I wasn't really able to pin it down.

I could not reproduce the issue in a test, probably because the issue is linked to collective indexing.

For [CA-3085]

## Checklist

- [x] Changelog entry

[CA-3085]: https://4teamwork.atlassian.net/browse/CA-3085?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-3085.bugfix
+++ b/changes/CA-3085.bugfix
@@ -1,0 +1,1 @@
+Update local roles after moving document when creating forwarding. [njohner]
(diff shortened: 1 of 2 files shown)


phgross (approved):
👍

PR #7248, opened 2021-11-18
Update ordnungssystem.xlsx template for the repository_minimal profile

elioschmutz:
Updates the `ordnungssystem.xlsx` according to https://4teamwork.slack.com/archives/C01BHEGBVU1/p1637219145165400

Value `Noch nicht geprüft` from the `public_trial` field has been removed.


phgross (approved):
👍

PR #7249, opened 2021-11-18
Remove ftw.tika dependency and uninstall tika profile.

phgross:
The package dependency is still necessary and will removed later, when the profile is uninstalled everywhere (see CA-1158 ).

For [CA-1212]

## Checklist
- [x] Changelog entry

[CA-1212]: https://4teamwork.atlassian.net/browse/CA-1212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-1212.other
+++ b/changes/CA-1212.other
@@ -1,0 +1,1 @@
+Remove ftw.tika dependency and uninstall tika profile. [phgross]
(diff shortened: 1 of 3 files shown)


njohner (approved):
Did you verify that the uninstall works properly and removes the transform as well as the product?

phgross:
@njohner yes I checked that the transform and transform policy is correctly removed, but I was not aware of the "product ftw.tika" . Thank you for the hint. I now uninstall the product, which also runs the uninstall profile.

PR #7257, opened 2021-11-25
Split upgrade in two to mitigate memory issues.

deiferni:
The upgrade to persist some missing default values with nightly jobs has a relatively high memory consumption as it loads many objects into memory.

As the upgrade operates on two distinct sets of objects we split the upgrade into two separate upgrades so that in combination with intermediate commits the memory can be freed after the first upgrade has been run.

We introduce a heuristic to detect whether the new upgrade should be run or not based on checking if the split upgrade with id "20211026132728" is already installed. For this purpose we also manually backdate our new upgrade to id "20211026132727" so that it will be picked up and run immediately before the upgrade we have split into two.

Two cases are possible:
  - The upgrade 20211026132728 has already been installed. This means it has been installed in its previous version before the split and this new upgrade 20211026132727 should not be executed as it has already been executed as part of 20211026132728.
  - The upgrade 20211026132728 is not yet installed. This means we are installing the split versions of the upgrades for the first time and thus can go ahead with this upgrade.

Manually tested:
- ✅  &nbsp;Upgrade on a deployment where it is picked up as non-orphan and executed
- ✅  &nbsp;Upgrade on a deployment where it is picked up orphan and skipped

For [CA-2746](https://4teamwork.atlassian.net/browse/CA-2746)

## Checklist

- [x] Changelog entry


lukasgraf (approved):
LGTM 👍

PR #7259, opened 2021-11-25
Add closed state for workspace meetings.

tinagerber:
A workspace meeting can now be opened and closed.

Currently, the Lawgiver cannot fully map the desired workflow. Since the automatic creation of the workflow is nevertheless practical, I have kept the specification in, but written a test that ensures that the manual changes in the workflow are not overwritten.

For [CA-2461]

## Checklist

- [x] Changelog entry

[CA-2461]: https://4teamwork.atlassian.net/browse/CA-2461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/core/locales/de/LC_MESSAGES/plone.po
```diff
@@ -349,6 +359,21 @@ msgstr "Teammitglied"
...
+
+#. Default: "close"
+#: opengever/core/profiles/default/workflows/opengever_workspace_meeting/specification.txt
+msgid "opengever_workspace_meeting--TRANSITION--close--active_closed"
+msgstr "Abschliessen"
+
+#. Default: "reopen"
+#: opengever/core/profiles/default/workflows/opengever_workspace_meeting/specification.txt
+msgid "opengever_workspace_meeting--TRANSITION--reopen--closed_active"
+msgstr "Wieder öffnen"
```
phgross, line 375: We use "Wieder eröffnen" everywhere else - i would call it that for the meetings as well.

opengever/core/locales/de/LC_MESSAGES/plone.po
```diff
@@ -465,6 +490,11 @@ msgstr "Aktiv"
 msgid "reactivate"
 msgstr "reaktivieren"
 
+#. Default: "reopen"
+#: opengever/core/profiles/default/workflows/opengever_workspace_meeting/specification.txt
+msgid "reopen"
+msgstr "Wieder öffnen"
```
phgross, line 496: se above.

opengever/core/tests/test_lawgiver_workflows.py
```diff
@@ -27,3 +30,53 @@ class TestWorkspaceWorkflow(GeverWorkflowTest):
 
 class TestWorkspaceDocumentWorkflow(GeverWorkflowTest):
     workflow_name = 'opengever_workspace_document'
+
+
+class TestWorkspaceMeetingWorkflow(GeverWorkflowTest):
+    workflow_name = 'opengever_workspace_meeting'
+
+    def test_workflow_definition_up_to_date(self):
```
phgross, line 38: I would choose another way for this test. Because I consider the string replacement quite fragile and not very readable. I would just check the installed workflow permissions.

Something like this should work and checks the same:
```python
workflow = self.portal.portal_workflow['opengever_workspace']

# check roles for permission
self.assertEqual(
    ('Administrator', 'Manager', 'WorkspaceAdmin', 'WorkspaceMember')
    workflow.states['opengever_workspace--STATUS--active'].permission_roles['opengever.task: Edit task'])

# check for acquire and roles
self.assertEqual(
    {'acquired': 1, 'roles': []},
    workflow.states['document-state-draft'].getPermissionInfo('Delete objects'))
```

phgross (changes requested):
LGTM, besides the tests mentioned ...

opengever/core/tests/test_lawgiver_workflows.py
tinagerber: The purpose of this test is to check if the definition.xml is up to date. If I now only check these three permissions, that would make the test kind of useless. I have rewritten the test a bit so that it is less error-prone. I also wrote another test: test_add_modify_and_delete_agendaitem_permission

tinagerber:
@phgross I improved the test and wrote another test. Can you rereview?

PR #7262, opened 2021-11-25
Improve complete successor task endpoint when using urls or relative paths as document references

elioschmutz:
Extends the `documents` payload parameter of the `@complete-successor-task` endpoint to also accept URLs.

We do this by reusing the already existing method `relationfield_value_to_object` which provides resolving a value to an object by multiple ways, also by URL.

By using this method, we change the behavior of resolving paths. Currently, the endpoint expects physical paths for the `documents` parameter. But physical paths are for internal use only. It now supports relative paths which is the correct way of resolving objects paths provided by a client. [breaking change]

For [CA-3193]

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [x] api-change label added
- New functionality:
- DB-Schema migration
- New translations

[CA-3193]: https://4teamwork.atlassian.net/browse/CA-3193?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CA-3197]: https://4teamwork.atlassian.net/browse/CA-3197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/api/relationfield.py
+++ b/opengever/api/relationfield.py
@@ -1,5 +1,6 @@
 from opengever.api.utils import get_obj_by_path
 from opengever.base.interfaces import IOpengeverBaseLayer
+from opengever.base.oguid import Oguid
 from plone.dexterity.interfaces import IDexterityContent
 from plone.restapi.deserializer.relationfield import RelationChoiceFieldDeserializer
 from plone.restapi.interfaces import IFieldDeserializer
@@ -36,6 +37,9 @@ def relationfield_value_to_object(value, context, request):
         elif value.startswith("/"):
             # Resolve by path
             return get_obj_by_path(portal, value.lstrip("/")), "path"
+        elif Oguid.is_oguid(value):
+            # Resolve by OGUID
+            return Oguid.parse(value).resolve_object(), "oguid"
         else:
             # Resolve by UID
             catalog = getToolByName(context, "portal_catalog")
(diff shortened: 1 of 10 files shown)


opengever/base/oguid.py
```diff
@@ -51,6 +52,11 @@ def parse(cls, oguid):
         admin_unit_id, int_id = parts[0], int(parts[1])
         return cls(admin_unit_id, int_id)
 
+    @classmethod
+    def is_oguid(cls, oguid):
+        # Mathes characters except : and / followed by a colon (:) and any numbers
```
phgross, line 57: matches

opengever/api/relationfield.py
```diff
@@ -36,6 +37,9 @@ def relationfield_value_to_object(value, context, request):
         elif value.startswith("/"):
             # Resolve by path
             return get_obj_by_path(portal, value.lstrip("/")), "path"
+        elif Oguid.is_oguid(value):
+            # Oguid
```
phgross, line 41: ✂️ or `Resolve by OGUID`

phgross (approved):
👍 besides the things mentioned

PR #7266, opened 2021-11-29
Make WOPI discovery requests more robust and fail safe.

phgross:
In some cases the office-online server does not return a key `net-zone` and the `editable_extensions ` raises an KeyError (see https://sentry.4teamwork.ch/organizations/sentry/issues/79611/?project=17&referrer=slack). 
This leads to the fact that no more actions are displayed on the documents, even if actually only the Office Online Edit action would not be available. 

I added a try/ecxept on the `editable_extensions` getter, do be more robust in that case.

For [CA-3268]

## Checklist

- [x] Changelog entry

[CA-3268]: https://4teamwork.atlassian.net/browse/CA-3268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


njohner (approved):
LGTM

PR #7267, opened 2021-11-29
Support invalid participation IDs in listing endpoint.

njohner:
This is to handle 2 situations:
- There are participations on a dossier that are not of the type of the currently activated contact feature (for example on dev.onegovgever.ch/rk, I activated KuB but there are plone participations that were not migrated)
- There are participations for a contact that does not exist anymore (currently contacts can be deleted in KuB).

Both these cases should not really happen in production as I understand it because:
- There will be some sort of migration when switching to KuB contacts (https://4teamwork.atlassian.net/browse/CA-2605)
- KuB plans on implementing deactivation of contacts instead of deletion.

But still, we can't afford that the whole dossier listing fails because of an unknown participant ID.

For [CA-2606]

## Checklist
- [ ] Changelog entry : no CL entry IMO

[CA-2606]: https://4teamwork.atlassian.net/browse/CA-2606?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/dossier/indexers.py
+++ b/opengever/dossier/indexers.py
@@ -343,8 +343,11 @@ class ParticipationIndexHelper(object):
             source = ContactsSource(api.portal.get())
         else:
             source = UsersContactsInboxesSourceBinder()(api.portal.get())
-        term = source.getTermByToken(participant_id)
-        return term.title
+        try:
+            term = source.getTermByToken(participant_id)
+            return term.title
+        except LookupError:
+            return participant_id
 
     def index_value_to_label(self, value):
         """Returns a translated label of the form 'participant label|role label'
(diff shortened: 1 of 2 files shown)


deiferni (approved):
LGTM 👍

Thoughts:
- should we log when this happens
- should we even log this to sentry as it should not happen as per PR description 🤔 

Please feel free to :shipit: this as is. We can follow up on the thoughts tomorrow and amend later.

njohner:
Well if we log to sentry, it will log all the time from dev. Until we have a migration path...
We could maybe take up in the migration story, to start logging to sentry once we have a migration and tested it on dev?

PR #7268, opened 2021-11-29
Add @kub endpoint allowing to resolve KuB entities by their ID.

njohner:
This endpoint basically proxies requests to the KuB service. This allows the frontend to resolve KuB entities without communicating directly with KuB.
Note that we currently need to parse the ID of the KuB contact to then request it from the correct endpoint from the KuB service. This will not be necessary anymore once KuB offers a `@resolve` endpoint (see https://4teamwork.atlassian.net/browse/HG-1995). Once we have that new endpoint we could also use it in the `KuBContactsSource` where we currently still use the `search` endpoint to resolve a given ID (because the other endpoints do not return the `text` and `TypeId` fields, which we use in the source as value and token respectively).

For [CA-2606]

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [ ] api-change label added

[CA-2606]: https://4teamwork.atlassian.net/browse/CA-2606?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/public/dev-manual/api/users.rst
+++ b/docs/public/dev-manual/api/users.rst
@@ -366,3 +366,56 @@ Mit dem ``@reactivate-local-group`` Endpoint kann eine lokale, inaktive Gruppe w
 
       HTTP/1.1 204 No content
       Content-Type: application/json
+
+
+KuB Kontakte
+============
+
+Mit dem ``@kub`` Endpoint können Kontakte aus dem KuB geholt werden. Der Endpoint steht nur auf Stufe PloneSiteRoot zur Verfügung und erwartet als Pfad Parameter die UID des Kontaktes:
+
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+      GET /@kub/person:1234abdc HTTP/1.1
+      Accept: application/json
+
+**Beispiel-Response**:
+
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+      {
+        "addresses": [],
+        "canton": null,
+        "country": "",
+        "countryIdISO2": "",
+        "created": "2021-11-14T00:00:00+01:00",
+        "dateOfBirth": null,
+        "description": "",
+        "emailAddresses": [],
+        "firstName": "Julie",
+        "fullName": "Dupont Julie",
+        "id": "0e623708-2d0d-436a-82c6-c1a9c27b65dc",
+        "languageOfCorrespondance": "fr",
+        "maritalStatus": 2,
+        "memberships": [],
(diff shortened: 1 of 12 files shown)


opengever/kub/client.py
```diff
@@ -44,3 +50,17 @@ def get_by_id(self, _id):
...
             raise LookupError()
         return res[0]
+
+    def get_full_entity_by_id(self, _id):
+        url = self.get_resolve_url(_id)
+        resp = self.session.get(url)
+        resp.raise_for_status()
+        return resp.json()
+
+    def get_resolve_url(self, _id):
```
phgross, line 60: Ist there a follow up story for the new `@resolve` endpoint, or maybe add a comment here 🤔

opengever/kub/client.py
```diff
@@ -44,3 +50,17 @@ def get_by_id(self, _id):
...
+        if ":" not in _id:
+            raise LookupError
+        id_type, uid = _id.split(":", 1)
+        if id_type not in ENDPOINT_BY_TYPE:
+            raise LookupError
+        return u'{}{}/{}'.format(self.kub_api_url, ENDPOINT_BY_TYPE[id_type], uid)
```
phgross, line 66: Is there a slash missing between the api_url and the endpoint name, or can we guarantee that the api_url always has an ending slash.

docs/public/dev-manual/api/users.rst
```diff
@@ -366,3 +366,56 @@ Mit dem ``@reactivate-local-group`` Endpoint kann eine lokale, inaktive Gruppe w
 
       HTTP/1.1 204 No content
       Content-Type: application/json
+
+
+KuB Kontakte
+============
+
+Mit dem ``@kub`` Endpoint können Kontakte aus dem KuB geholt werden. Der Endpoint steht nur auf Stufe PloneSiteRoot zur Verfügung und erwartet als Pfad Parameter die UID des Kontaktes:
```
phgross, line 374: KuB (Kontakt- und Behördenverzeichniss)

phgross (approved):
LGTM 👍 just a detail.

opengever/kub/client.py
njohner: You're right I'll make a story

njohner: https://4teamwork.atlassian.net/browse/CA-3303

opengever/kub/client.py
njohner: Yes `self.kub_api_url` has a trailing slash (it's a property defined with trailing slash)

PR #7270, opened 2021-12-01
Fix sending notification mails from or to users with long mail addresses.

phgross:
Pass in from and to mail adresses when sending mails to avoid problems for users with long fullname or mail addresses.

possibly needs backport to 2021.22.x 

For [CA-3277]

## Checklist

- [x] Changelog entry

[CA-3277]: https://4teamwork.atlassian.net/browse/CA-3277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/activity/mailer.py
+++ b/opengever/activity/mailer.py
@@ -7,6 +7,7 @@ from opengever.base.model import get_locale
 from opengever.mail.utils import make_addr_header
 from opengever.ogds.models.service import ogds_service
 from plone import api
+from Products.MailHost.MailHost import _mungeHeaders
 from threading import local
 import transaction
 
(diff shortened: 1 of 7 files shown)


opengever/activity/mailer.py
```diff
@@ -37,8 +37,14 @@ def process_mail_queue():
...
-            mail = mail_queue.get()
-            mailhost.send(mail)
+            mail, mail_to, mail_from = mail_queue.get()
+
+            # The MailHost `send` method overwrites the FROM and TO header
+            # in the Message, if you pass in to and from adresses. But we
+            # need to pass in the adresses to avoid problems with long or
+            # fullnames with umlauts. Therefore we directly use the `_send`
+            # method.
+            mailhost._send(mail_from, mail_to, mail.as_string())
```
lukasgraf: I wonder if we should skip the call to _mungeHeaders alltogether. I would probably just call it with `mto=None` and `mfrom=None`, and ignore the `mto` and `mfrom` it returns, but otherwise let it do to the message whatever it does.

lukasgraf (approved):
LGTM! 👍

PR #7278, opened 2021-12-02
Fix task documents method/property collision

elioschmutz:
This PR fixes an issue where we have a conflict between a method called `documents` on a task (introduced in https://github.com/4teamwork/opengever.core/pull/7217) and a property callled `documents`.

Bildschirmfoto 2021-12-02 um 16 16 21

Since the property wins over a method, we have had an issue when rendering the `task/browser/overview.py` because it expects a function under `documents`.

Bildschirmfoto 2021-12-02 um 16 27 21

The main issue is when delegating tasks. The transition-extender passes `documents` coming from the `transition_params` further to the dexterity's [createContent](https://github.com/plone/plone.dexterity/blob/2.2.8/plone/dexterity/utils.py#L150) method which will then add a property `documents` to the created content, even if it's not defined in a schema-definition.

Since the refactoring of https://github.com/4teamwork/opengever.core/pull/7217, we also define a method on the task object called `documents`. So the user visible issue exists since this refactoring.

To fix the issue we:

- rename the `documents` method to `task_documents``
- no longer forward the `documents` transition-param to the `createContent` method

We decided to not create an upgrade step for this issue. It would be a quite expensive upgradestep without any benefit (right now...) => This is still up for discussion if we want an upgrade-step or not

For [CA-3272]

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

[CA-3272]: https://4teamwork.atlassian.net/browse/CA-3272?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-3272-2.bugfix
+++ b/changes/CA-3272-2.bugfix
@@ -1,0 +1,1 @@
+Task delegation does no longer set an unwanted documents-property on the subtask. [elioschmutz]
--- a/changes/CA-3272.bugfix
+++ b/changes/CA-3272.bugfix
@@ -1,0 +1,1 @@
+Fix task overview in old ui for tasks created by task delegation. [elioschmutz]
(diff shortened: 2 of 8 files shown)


njohner (commented):
LGTM, just a typo in your second commit message: porperties -> properties

njohner (approved):

PR #7279, opened 2021-12-03
Improve teamraum meeting PDF style.

njohner:
With this PR we try to improve the layout and style of the Teamraum meeting PDFs. The main changes are that we allow tables to get split over several pages and we add borders to the tables. I include 2 PDFs showing the difference between the old style and the new style:

[old_style.pdf](https://github.com/4teamwork/opengever.core/files/7649764/old_style.pdf)
[new_style.pdf](https://github.com/4teamwork/opengever.core/files/7649765/new_style.pdf)

Note that the output PDF does not respect the column widths set by the user. This is because the `wdith` attribute is only presentational hint. `weasyprint` does have a parameter that should force respecting these attributes (https://doc.courtbouillon.org/weasyprint/stable/api_reference.html#cmdoption-p), which I tried and did not help. 🤷 

For [CA-3249]

## Checklist

- [x] Changelog entry

[CA-3249]: https://4teamwork.atlassian.net/browse/CA-3249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/workspace/browser/templates/meeting_minutes.pt
+++ b/opengever/workspace/browser/templates/meeting_minutes.pt
@@ -41,14 +41,25 @@ li > p {
   margin-top: 0.25em;
   margin-bottom: 0.25em;
 }
-table, figure {
+figure {
   page-break-inside: avoid;
+}
+table {
+  margin-top: 10pt;
+  margin-bottom: 15pt;
+  border-collapse: collapse;
+  page-break-inside: auto;
 }
 th {
   text-align: left;
 }
 td, th {
+  padding-left: 10pt;
   padding-right: 10pt;
+  border: 1px solid black;
+}
+table.borderless td, table.borderless th {
+  border: none;
 }
 a {
   text-decoration: none;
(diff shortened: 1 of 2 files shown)


opengever/workspace/browser/templates/meeting_minutes.pt
```diff
@@ -41,15 +41,25 @@ li > p {
...
 th {
   text-align: left;
 }
 td, th {
+  padding-left: 10pt;
   padding-right: 10pt;
 }
+table:not(.borderless) td, table:not(.borderless) th {
+  border: 1px solid black;
+}
```
elioschmutz, line 62:
```css
table td, table th {
  border: 1px solid black;
}
table td.borderless, table th.borderless {
  border: none;
}
```

define the default styling first and then the special styling (here `borderless`)

elioschmutz (approved):

PR #7284, opened 2021-12-06
Handle KuB feature when creating a document from template.

njohner:
We add support for using KuB contacts as recipients when creating a document from a template. The data of that recipient is then passed as docproperties when creating the document.  Note that we are using the same docproperties with the same keys as were used with SQL contacts. This should allow existing templates to still work after a switch to KuB. Also note that KuB currently does not return a `primaryEmail` and `primaryUrl`, but that should change in the near future.

Note that we do not support KuB contacts in the classic-UI and therefore display an error message when a user tries to create a document from a template with the KuB feature active. This will not prevent creation from a document
from a template, but it does not allow to select a recipient either.

The config endpoint now distinguishes the three contact feature types, so returns a string (`plone`, `sql`, `kub`) instead of a boolean. This is a small breaking change.

For [CA-2452]

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [x] api-change label added

[CA-2452]: https://4teamwork.atlassian.net/browse/CA-2452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/kub/tests/test_docprops.py
+++ b/opengever/kub/tests/test_docprops.py
@@ -1,0 +1,71 @@
+from opengever.kub.docprops import KuBEntityDocPropertyProvider
+from opengever.kub.entity import KuBEntity
+from opengever.kub.testing import KuBIntegrationTestCase
+import requests_mock
+
+
+@requests_mock.Mocker()
+class TestKuBEntityDocPropertyProvider(KuBIntegrationTestCase):
+
+    def test_docproperties_for_kub_person(self, mocker):
+        self.mock_get_full_entity_by_id(mocker, self.person_julie)
+        entity = KuBEntity(self.person_julie, full=True)
+        properties = KuBEntityDocPropertyProvider(entity).get_properties()
+        self.assertDictEqual(
+            {'ogg.address.city': None,
+             'ogg.address.country': None,
+             'ogg.address.street': None,
+             'ogg.address.zip_code': None,
+             'ogg.contact.description': u'',
+             'ogg.contact.title': None,
+             'ogg.email.address': None,
+             'ogg.person.academic_title': u'',
+             'ogg.person.firstname': u'Julie',
+             'ogg.person.lastname': u'Dupont',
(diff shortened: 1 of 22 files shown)


opengever/kub/tests/test_docprops.py
```diff
@@ -0,0 +1,71 @@
...
+        self.assertDictEqual(
+            {'ogg.address.city': None,
+             'ogg.address.country': None,
+             'ogg.address.street': None,
+             'ogg.address.zip_code': None,
+             'ogg.contact.description': u'',
+             'ogg.contact.title': None,
+             'ogg.email.address': None,
+             'ogg.person.academic_title': u'',
+             'ogg.person.firstname': None,
```
phgross, line 23: Why is this `None` should be `Julie`, not? 🤔

opengever/kub/tests/test_docprops.py
```diff
@@ -0,0 +1,71 @@
...
+             'ogg.address.zip_code': None,
+             'ogg.contact.description': u'',
+             'ogg.contact.title': None,
+             'ogg.email.address': u'Jean.dupon@example.com',
+             'ogg.organization.name': u'4Teamwork',
+             'ogg.orgrole.department': u'',
+             'ogg.orgrole.description': u'',
+             'ogg.orgrole.function': u'CEO',
+             'ogg.person.academic_title': u'',
+             'ogg.person.firstname': None,
```
phgross, line 66: Same here

phgross (approved):
LGTM 👍 Just the one test i do not unterstand ...

opengever/kub/tests/test_docprops.py
njohner: Ah yes, thanks! I did not realise that KuB did not use the field names in the serialised data, but instead used CamelCased names...

PR #7287, opened 2021-12-08
Add @my-substitutes, @substitutes and @out-of-office API endpoints

tinagerber:
With `@my-substitutes` we can list, add and delete substitutes of the current user.
With the `@substitutes` endpoint we can list the substitutes of any user.
With the `@out-of-office` endpoint, we can show and edit the absence settings of the current user.
The `@actors` endpoint indicates whether an actor is absent.

For [CA-2274]

## Checklist

- [x] Changelog entry

- API change:
- DB-Schema migration
- New translations

[CA-2274]: https://4teamwork.atlassian.net/browse/CA-2274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -17,6 +17,9 @@ Other Changes
 - @xhr-upload: new endpoint to upload documents as a multipart/form-data xhr request. [elioschmutz]
 - Include is_completed in sql task serialization.
 - ``@listing``: Add retention_expiration column.
+- New endpoints ``@my-substitutes`` and ``@substitutes`` are added (see :ref:`substitutes`).
+- A new endpoint ``@out-of-office`` is added (see :ref:`out-of-office`).
+- Include is_absent in actors serialization.
 
 
 2021.24.0 (2021-11-30)
--- a/docs/public/dev-manual/api/substitutes.rst
+++ b/docs/public/dev-manual/api/substitutes.rst
@@ -1,0 +1,133 @@
+.. _substitutes:
+
+Stellvertreter
+==============
+
+Mit dem ``@my-substitutes`` Endpoint können für den aktuellen Benutzer Stellvertreter aufgelistet, hinzugefügt und gelöscht werden. Mit dem ``@substitutes`` Endpoint  können für einen beliebigen Benutzer Stellvertreter aufgelistet werden. Die Endpoints stehen nur auf Stufe PloneSite zur Verfügung.
+
+
+Eigene Stellvertreter auflisten:
+--------------------------------
+Mittels eines GET Request können Stellvertretungen des aktuellen Benutzers abgefragt werden. Es werden alle, also global über den ganzen Mandanten-Verbund, zurückgegeben.
+
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+       GET /@my-substitutes HTTP/1.1
+       Accept: application/json
+
+
+**Beispiel-Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+      {
+          "@id": "http://localhost:8081/fd/@my-substitutes",
+          "items": [
+              {
+                  "@id": "http://localhost:8081/fd/@my-substitutes/peter.mueller",
+                  "@type": "virtual.ogds.substitute",
+                  "substitution_id": 3,
+                  "substitute_userid": "peter.mueller",
+                  "userid": "kathi.barfuss"
+              },
+              {
+                  "@id": "http://localhost:8081/fd/@my-substitutes/nicole.kohler",
+                  "@type": "virtual.ogds.substitute",
+                  "substitution_id": 3,
+                  "substitute_userid": "nicole.kohler",
+                  "userid": "kathi.barfuss"
+              },
+          ],
+          "items_total": 2
+      }
+
+Stellvertreter eines Benutzers auflisten:
+-----------------------------------------
+Mittels eines GET Request können Stellvertretungen eines Benutzers abgefragt werden. Dabei wird die User-ID des Benutzers als Pfad-Parameter erwartet. Es werden alle, also global über den ganzen Mandanten-Verbund, zurückgegeben.
+
+
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+       GET /@substitutes/kathi.barfuss HTTP/1.1
+       Accept: application/json
+
+
+**Beispiel-Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+      {
+          "@id": "http://localhost:8081/fd/@substitutes/kathi.barfuss",
+          "items": [
+              {
+                  "@id": "http://localhost:8081/fd/@substitutes/peter.mueller",
+                  "@type": "virtual.ogds.substitute",
+                  "substitution_id": 3,
+                  "substitute_userid": "peter.mueller",
+                  "userid": "kathi.barfuss"
+              },
+              {
+                  "@id": "http://localhost:8081/fd/@substitutes/nicole.kohler",
+                  "@type": "virtual.ogds.substitute",
+                  "substitution_id": 3,
+                  "substitute_userid": "nicole.kohler",
+                  "userid": "kathi.barfuss"
+              },
+          ],
+          "items_total": 2
+      }
+
+
+Stellvertretung hinzufügen:
+---------------------------
+Eine Stellvertretung des aktuellen Benutzers kann mittels POST Request hinzugefügt werden. Dabei wird die User-ID als Parameter erwartet.
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+       POST /@my-substitutes HTTP/1.1
+       Accept: application/json
+
+       {
+        "userid": "peter.mueller"
+       }
+
+
+**Beispiel-Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 204 No content
+
+
+Stellvertretung entfernen:
+--------------------------
+Eine bestehende Stellvertretung des aktuelllen Benutzers kann mittels DELETE Request wieder gelöscht werden. Als Pfad-Parameter wird die User-ID der Stellvertretung erwartet.
+
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+       DELETE /@my-substitutes/peter.mueller HTTP/1.1
+       Accept: application/json
+
+
+**Beispiel-Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 204 No Content
--- a/opengever/api/out_of_office.py
+++ b/opengever/api/out_of_office.py
@@ -1,0 +1,73 @@
+from datetime import datetime
+from opengever.ogds.base.utils import ogds_service
+from plone import api
+from plone.restapi.deserializer import json_body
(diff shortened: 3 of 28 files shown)


docs/public/dev-manual/api/delegates.rst
```diff
@@ -0,0 +1,87 @@
...
+                  "@type": "virtual.ogds.my_delegate",
+                  "delegate_actorid": "peter.mueller"
+              },
+              {
+                  "@id": "http://localhost:8081/fd/@my-delegates/nicole.kohler",
+                  "@type": "virtual.ogds.my_delegate",
+                  "delegate_actorid": "nicole.kohler"
+              },
+          ],
+          "items_total": 3
```
njohner, line 43: "items_total": 2

opengever/api/delegates.py
```diff
@@ -0,0 +1,81 @@
...
+from plone.protect.interfaces import IDisableCSRFProtection
+from plone.restapi.deserializer import json_body
+from plone.restapi.services import Service
+from zExceptions import BadRequest
+from zope.interface import alsoProvides
+from zope.interface import implements
+from zope.publisher.interfaces import IPublishTraverse
+
+
+def serialize_my_delegate_to_json(delegate):
```
njohner, line 13: Why not write a serializer for the Delegate model, see for example https://github.com/4teamwork/opengever.core/blob/master/opengever/api/serializer.py#L336-L389?

opengever/ogds/base/actor.py
```diff
@@ -148,6 +149,10 @@ def get_link(self, with_icon=False):
     def is_active(self):
         return True
 
+    @property
+    def is_absent(self):
```
njohner, line 153: Hum I find that a bit debatable. This is a property that is very specific for users, I don't know if it really belongs on the actor. Do we plan on showing the absence in the frontend directly on the avatars?

opengever/core/upgrades/20211202155051_add_delegations/upgrade.py
```diff
@@ -0,0 +1,29 @@
...
+from sqlalchemy.schema import Sequence
+from sqlalchemy import Integer
+
+
+class AddDelegations(SchemaMigration):
+    """Add delegations.
+    """
+
+    def migrate(self):
+        # self.op.create_table(
```
njohner, line 18: ?

njohner (commented):
Ok I haven't reviewed in details, because I have a few general remarks:
- "delegates" is not the correct term in my opinion, more standard would be substitute, deputy, surrogate... Delegate can be confused with delegating. termdat says it should be "substitute" in english and "suppléant" in French for Gever. So please use those.
- I don't like that the `@my-delegates` and `@out-of-office` endpoints are for the current user. That will quickly lead to problems, i.e. as soon as we realise that it would be nice for someone to see if someone else is out of office, or maybe see who the substitute is for another person... I'd find it better that they both take the userid as path parameter.
- I'm not sure `is_absent` belongs on the actor, but if we absolutely need it...
- lastly splitting the "Add @my-delegates API endpoints." commit into two would make review a bit easier. One commit for the model, maybe one for the manager and one for the endpoint...

tinagerber:
@lukasgraf FYI: After a discussion with @njohner , we decided to call the endpoints @substitutes, since TERMDAT also suggests this for "Stellvertreter".

tinagerber:
@njohner As discussed. I've renamed the endpoints to `@my-substitutes` and added another endpoint `@substitutes`. Can you have another look?

opengever/api/out_of_office.py
```diff
@@ -0,0 +1,73 @@
...
+
+            if 'absent_to' in data:
+                absent_to = data['absent_to']
+                user.absent_to = data['absent_to']
+                if absent_to is not None:
+                    absent_to = datetime.strptime(absent_to, '%Y-%m-%d').date()
+            else:
+                absent_to = user.absent_to
+
+            self.validate_dates(absent_from, absent_to)
```
njohner, line 61: That whole block is fishy, and I think it should be:
```
        if 'absent_from' in data or 'absent_to' in data:
            if 'absent_from' in data:
                if absent_from is not None:
                    absent_from = datetime.strptime(data['absent_from'], '%Y-%m-%d').date()
                user.absent_from = absent_from
            else:
                absent_from = user.absent_from
            if 'absent_to' in data:
                if absent_to is not None:
                    absent_to = datetime.strptime( data['absent_to'], '%Y-%m-%d').date()
                user.absent_to = absent_to
            else:
                absent_to = user.absent_to
            self.validate_dates(absent_from, absent_to)
```

opengever/api/out_of_office.py
```diff
@@ -0,0 +1,73 @@
...
+
+        userid = api.user.get_current().getId()
+        user = ogds_service().fetch_user(userid)
+
+        if 'absent' in data:
+            user.absent = data['absent']
+
+        if 'absent_from' in data or 'absent_to' in data:
+            if 'absent_from' in data:
+                user.absent_from = data['absent_from']
```
njohner, line 46: That seems wrong to me, `data['absent_from']` is a string and you actually want to save a date on user no?

opengever/api/tests/test_actors.py
```diff
@@ -344,6 +358,76 @@ def test_returns_null_actor_for_invalid_actor_id(self, browser):
...
+        sql_user = User.query.get(self.regular_user.getId())
+        sql_user.absent = True
+        create_session().flush()
+
+        browser.open(url, headers=self.api_headers)
+        self.assertEqual(200, browser.status_code)
+        self.assertTrue(browser.json['is_absent'])
+
+    @browsing
+    def test_is_not_absent_if_absent_is_true_but_today_is_not_between_absent_dates(self, browser):
```
njohner, line 413: huh, that is not the behaviour I'd expect.

opengever/ogds/base/actor.py
njohner: As discussed we will need that to more easily handle absent users in vocabularies

opengever/ogds/models/substitute.py
```diff
@@ -0,0 +1,33 @@
...
+
+class Substitute(Base):
+    """Substitute model
+    """
+
+    query_cls = SubstituteQuery
+
+    __tablename__ = 'substitutes'
+
+    substitute_id = Column("id", Integer, Sequence("substitute_id_seq"), primary_key=True)
```
njohner, line 31: So this here should actually be `substitution_id` no? We have one row per substitution in that table, a substitution consists of a user and his substitute.

docs/public/dev-manual/api/substitutes.rst
```diff
@@ -0,0 +1,133 @@
...
+-----------------------------------------
+Mittels eines GET Request können Stellvertretungen eines Benutzers abgefragt werden. Dabei wird die User-ID des Benutzers als Pfad-Parameter erwartet. Es werden alle, also global über den ganzen Mandanten-Verbund, zurückgegeben.
+
+
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+       GET /@my-substitutes HTTP/1.1
```
njohner, line 60: @substitutes/kathi.barfuss

docs/public/dev-manual/api/substitutes.rst
```diff
@@ -0,0 +1,133 @@
+.. _substitutes:
+
+Stellvertreter
+==============
+
+Mit dem ``@my-substitutes`` Endpoint können für den aktuellen Benutzer Stellvertreter aufgelistet, hinzugefügt und gelöscht werden. Mit dem ``@substitutes`` Endpoint  können für den aktuellen Benutzer Stellvertreter aufgelistet werden. Die Endpoints stehen nur auf Stufe PloneSite zur Verfügung.
```
njohner, line 6: ``@substitutes`` Endpoint  können für den aktuellen Benutzer -> ``@substitutes`` Endpoint  können für ein gewissen Benutzer

opengever/api/substitutes.py
```diff
@@ -0,0 +1,123 @@
...
+            data['@id'] = '{}/@my-substitutes/{}'.format(
+                api.portal.get().absolute_url(),
+                self.context.substitute_userid)
+
+        return data
+
+
+class MySubstitutesGet(Service):
+
+    def reply(self):
```
njohner, line 43: ✂️ 
You don't need to define `reply` here, the only thing you need is:
```
def get_item_query(self):
    userid = self.get_userid()
    return SubstituteManager().list_substitutes_for(userid)
```

opengever/api/tests/test_substitutes.py
```diff
@@ -0,0 +1,198 @@
...
+            browser.open(self.portal, view='@my-substitutes', method='POST',
+                         data=json.dumps({}),
+                         headers=self.api_headers)
+
+        self.assertEqual(
+            {u"message": u"Property 'userid' is required", u"type": u"BadRequest"},
+            browser.json)
+
+    @browsing
+    def test_post_my_substitutes_raisesif_userid_is_invalid(self, browser):
```
njohner, line 130: raisesif -> raises_if

docs/public/dev-manual/api/api_changelog.rst
```diff
@@ -13,9 +13,11 @@ Other Changes
 ^^^^^^^^^^^^^
 - Disposition serialization contains now responses. [phgross]
 - @xhr-upload: new endpoint to upload documents as a multipart/form-data xhr request. [elioschmutz]
-
 - Include is_completed in sql task serialization.
 - ``@listing``: Add retention_expiration column.
+- New endpointa ``@my-substitutes`` and ``@substitutes`` are added (see :ref:`substitutes`).
```
njohner, line 18: endpointa -> endpoints

njohner (commented):
Just a few more comments and I think you currently have a bug...

opengever/api/out_of_office.py
tinagerber: It doesn't matter if a date object or a date string is assigned, sqlalchemy can handle both. But I changed it now anyway, so it's clearer.

opengever/api/tests/test_actors.py
tinagerber: This is the behavior I agreed with @all4manu  to make it as simple as possible for frontend and backend. Otherwise it would need an additional validation that the flag cannot be set at the same time as the dates.

opengever/api/substitutes.py
tinagerber: ?? I don't understand what you mean.

tinagerber:
@njohner can you have another look? One comment I didn't understand.

opengever/api/substitutes.py
njohner: Sorry, my bad...

opengever/api/out_of_office.py
njohner: ah ok 👍

opengever/api/tests/test_actors.py
njohner: well no, I'd say you can set the flag at the same time as dates, I'd simply have thought the flag wins. I find the current status weird. I'd go either for an `and`, i.e. `is_absent` is True if `absent` and in the date range. Or I'd go for an `or`, i.e. `is_absent` is True if the `absent` or if in the date range.

njohner: Simplest IMO would be to only have a POST and not a PATCH... Then it would be very easy, no need to check what the status was before and whether it makes sense with the new changes

njohner (approved):
LGTM

PR #7289, opened 2021-12-08
Docker for testserver

jone:
### Goal

The goal is to be able to run the testserver in docker, so that:
1. it can be easily used for testing other components (e.g. RIS)
2. devs who are developing services using GEVER can run the testserver locally without having to run buildout.

### Scope

This pull request makes it possible to run the testserver with docker-compose. It is not yet very optimized and the build process is not automated.

### Details

- The testserver is integrated in the `docker-compose.yml`, so that it can be easily run with docker compose and shares the services (such as sablon).
- The testserver has to be changed so that it supports the case that solr is running on a different host than `localhost`, so that the solr container can be accessed over the docker compose network.
- The Dockerfile of the testserver currently needs to re-copy the code (`/opengever`), because this PR contains code that is not yet incorporated in the latest ogcore image. Once the PR is merged and the image is re-published with the new code, this copy statement can be removed from the Dockerfile.

### Usage

```
$ cd opengever.core
$ docker-compose up testserver
# have patience
$ ./bin/testserverctl isolate
```

Access http://localhost:55001 and have fun.

### Caveats

Booting the testserver is still very slow. We could optimize that by building the fixture into an image. I think this should be an image on top of this one (e.g. `testserver-ogfixture`), so that we can have a similar one with the RIS fixture.
The challenge here is that building the testserver fixture requires other services such as `sablon`, which cannot be access during build time (at least I couldn't figure it out).
Dumping the databases with the fixture could be done with the existing `DBCacheManager` class.

## Checklist

- [X] Changelog entry

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

### DIFF
--- a/docker/testserver/Dockerfile
+++ b/docker/testserver/Dockerfile
@@ -1,0 +1,20 @@
+FROM 4teamwork/ogcore:latest
+
+USER root
+WORKDIR /app
+
+COPY ./docker/testserver/requirements-testserver.txt /app/
+RUN --mount=type=cache,target=/root/.cache \
+    pip install \
+    --extra-index-url https://buildout:buildout@psc.4teamwork.ch/simple \
+    -r requirements-testserver.txt \
+    -c requirements-core.txt \
+    -c requirements-deployment.txt
+
+HEALTHCHECK --start-period=2m CMD /app/bin/testserverctl connectiontest
+COPY ./bin/testserverctl ./docker/testserver/bin/testserver /app/bin
+COPY ./docker/testserver/docker-entrypoint.sh /app/
+
+USER plone
+EXPOSE 55001
+EXPOSE 55002
--- a/README.rst
+++ b/README.rst
@@ -1230,6 +1230,12 @@ The connection from the ``testserverctl`` to the XMLRPC-Server can be tested wit
 This will result in a "Connection refused" error as long as the testserver is starting and will do nothing when the server is ready for the first ``isolate`` or ``zodb_setup``.
 This can be used as docker healthcheck.
 
+Testserver in docker
+~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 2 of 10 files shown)


docker/testserver/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+COPY ./docker/testserver/requirements-testserver.txt /app/
+RUN --mount=type=cache,target=/root/.cache \
+    pip install \
+    --extra-index-url https://buildout:buildout@psc.4teamwork.ch/simple \
+    -r requirements-testserver.txt \
+    -c requirements-core.txt \
+    -c requirements-deployment.txt
+
+FROM build-stage
```
buchi, line 14: This doesn't make sense. We use a build stage for ogcore to reduce image size and to not have dev libraries installed in the final image. For the testserver there's no need for a build stage as only a few additional Python packages are installed and luckily without requiring the dev packages.
Thus remove this line and the `AS build-stage` part in line 1.

docker/testserver/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+HEALTHCHECK --start-period=2m CMD /app/bin/testserverctl connectiontest
+COPY ./bin/testserverctl ./docker/testserver/bin/testserver /app/bin
+COPY ./docker/testserver/docker-entrypoint.sh /app/
+
+# <TEMPORARY>
+# ./opengever is copied so that we can include local development changes of the testserver.
+# This should be removed when the dockerization of the testserver is finished.
+COPY ./opengever ./opengever
```
buchi, line 23: Remove this before merging.

buchi (changes requested):
👍🏻 LGTM!

Just a few change requests, see comments.

buchi (approved):

PR #7291, opened 2021-12-09
Add support for external activities

lukasgraf:
This adds support for creating external activities via a new `@external-activities` endpoint.

This endpoint allows to create activities of kind 'external-activity' over the REST API. These activities never have an internal `resource`, and therefore need an `external_resource_url`.

They allow for free-form text for their label, title, summary and description.

They also need to have their `notification_recipients` explicitly specified. Currently, the list of notification recipients always needs to be exactly the authenticated user - therefore, no notifications for other users can be created using this endpoint.

Because now not every activity has a `resource` with an OGUID, handling for this had to be added in several places. For activities without a resource, the `@@resolve_notification` view will redirect to the `external_resource_url`. For this to work properly, a change in `gever-ui` will also be necessary.

For [CA-3139](https://4teamwork.atlassian.net/browse/CA-3139)

### Notification Settings
Screenshot 2021-12-09 at 12 25 24

### Notification Examples
Screenshot 2021-12-09 at 12 29 13

Screenshot 2021-12-09 at 12 28 48

## Checklist

- [x] Changelog entry

- API change:
  - [x] API Changelog entry
- DB-Schema migration
- New translations

### DIFF
--- a/opengever/api/tests/test_external_activities.py
+++ b/opengever/api/tests/test_external_activities.py
@@ -1,0 +1,202 @@
+from datetime import datetime
+from ftw.testbrowser import browsing
+from ftw.testing import freeze
+from opengever.activity.model import Activity
+from opengever.testing import IntegrationTestCase
+import json
+
+
+class TestExternalActivitiesPost(IntegrationTestCase):
+
+    features = ("activity",)
+
+    @browsing
+    def test_creating_external_activity_requires_authentication(self, browser):
+        url = "%s/@external-activities" % self.portal.absolute_url()
+        with browser.expect_http_error(code=401):
+            browser.open(
+                url,
+                method="POST",
+                data=json.dumps({}),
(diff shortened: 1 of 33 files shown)


opengever/api/tests/test_external_activities.py
```diff
@@ -0,0 +1,204 @@
...
+from opengever.activity.model import Activity
+from opengever.testing import IntegrationTestCase
+import json
+
+
+class TestExternalActivitiesPost(IntegrationTestCase):
+
+    features = ("activity",)
+
+    HEADERS = {"Accept": "application/json", "Content-Type": "application/json"}
```
phgross, line 13: ✂️ - it's never used

phgross (approved):
Nice work, looks good 👍

I think the restriction that only custom notifications can be generated will probably be lifted soon. But currently that is good enough.

PR #7292, opened 2021-12-09
Fix batching issues in sharing view.

njohner:
Because the @sharing endpoint is batched, the sharing view did not display search results correctly when there were already many principals with local roles set on the context. This prevented from adding new local roles on certain contexts over the sharing view.

When there are too many search results, so that not all are displayed (i.e. more than 25 search results), then a warning is displayed, the user could then further restrict his search by typing some more...

Moreover I had to change the ordering of the results, to make sure the already existing settings are returned first (so that if the batch size is large enough). Otherwise it might be that the search results take over some of the spots of the existing settings. If the frontend were then to post these, it would actually be removing existing settings (yes that is a bug that we currently have, although it probably never happens).
Screenshot 2021-12-09 at 14 33 25

For [CA-3317]

## Checklist
- [x] Changelog entry

[CA-3317]: https://4teamwork.atlassian.net/browse/CA-3317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/sharing/locales/de/LC_MESSAGES/opengever.sharing.po
```diff
@@ -54,15 +54,16 @@ msgstr "Lokale Rolle erfolgreich gespeichert"
 msgid "message_save_failed"
 msgstr "Lokale Rollen konnten nicht gespeichert werden."
 
+#. Default: "Only the 25 first search results are visible."
+#: ./opengever/sharing/browser/sharing.py
+msgid "not_all_search_results_visible"
+msgstr "Nor fie 25 ersten Suchresultaten sind dargestellt."
```
phgross, line 60: Es werden nur die 25 ersten Suchresultate dargestellt.

phgross (approved):
👍

opengever/sharing/locales/de/LC_MESSAGES/opengever.sharing.po
njohner: 🤣

njohner: sorry about that

PR #7299, opened 2021-12-14
Implement @solrsearch filtering by @id through a POST request

elioschmutz:
This PR extends the `@solrsearch` endpoint with a new `fq`: `url` (or you can also use `@id` as an alias).

This allows us to directly filter objects by its `@id`. 

Since queries can be very long and an url is limited by the client, server or proxy, we now allow `POST` requests against the `@solrsearch` endpoint. This allows us to use the request-body instead the query-string to pass arguments.

The `POST` endpoint is an alias for the `GET` endpoint. So it behaves exactly the same with exactly the same feature-set. The only difference is, that we pass JSON with the `POST` endpoint.

Here is a practical example of how to use the `POST` endpoint in combination with the `@id` filter:

**Request**
```http
POST /plone/@solrsearch HTTP/1.1
Content-Type: application/json
Accept: application/json

{
    "fl":"Title",
    "fq": [
        "@id:http://localhost:8080/fd/sitzungen/committee-1/submitted-proposal-419/document-36",
        "@id:http://localhost:8080/fd/sitzungen/committee-1/submitted-proposal-419/document-34"
        ]
}
```

**Response**
```json
{
    "@id": "http://localhost:8080/fd/@solrsearch",
    "facet_counts": {},
    "items": [
        {
            "@id": "http://localhost:8080/fd/sitzungen/committee-1/submitted-proposal-419/document-36",
            "UID": "5829913fdab744adb3f5838312ce01bd"
        },
        {
            "@id": "http://localhost:8080/fd/sitzungen/committee-1/submitted-proposal-419/document-34",
            "UID": "cac0830361404d9c82ecd730a9337bd0"
        }
    ],
    "items_total": 2,
    "rows": 25,
    "start": 0
}
````

For [CA-3044]

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

[CA-3044]: https://4teamwork.atlassian.net/browse/CA-3044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/public/dev-manual/api/searching.rst
+++ b/docs/public/dev-manual/api/searching.rst
@@ -108,10 +108,13 @@ darzustellen, kann ``metadata_fields=_all`` verwendet werden.
    handelt, zur Verfügung.
 
 
-Solr Suche
-----------
+.. _solr-search-get:
+
+Solr Suche GET
+--------------
 Um direkt eine Suchabfrage an den Solr Service abzusetzen, steht der ``@solrsearch`` Endpoint zur Verfügung. Für die Abfrage kann SOLR API Syntax und dessen Parameter verwendet werden. Der Endpoint liefert ein Liste von Treffern zurück mit den im Parameter ``fl`` definierten Felder. Die folgende Parameter sind zurzeit vom Endpoint unterstützt:
 
+Achtung: Eine URL hat je nach Client/Server/Proxy eine maximale länge. Für grössere Queries muss der Endpoint als POST-Request abgefragt werden.
 
 Query
 ~~~~~
@@ -270,6 +273,34 @@ Der Endpoint stellt die Standard-Paginierung gem :ref:`Kapitel Paginierung <batc
 Breadcrumbs
 ~~~~~~~~~~~
 Wird eine ``@solrsearch`` Anfrage mit ``breadcrumbs=1`` Parameter ergänzt, so werden die einzelnen Suchtreffer unter dem Key ``breadcrumbs`` mit den Breadcrumb Informationen ergänzt. Diese Ergänzung ist nur bei kleinen Batchsizes (maximal 50) erlaubt.
+
+
+Solr Suche POST
+---------------
+Eine Suche kann auch über einen POST-Request an den ``@solrsearch`` Endpoint abgesetzt werden. Der Hauptvorteil eines POST-Requests ist, dass es keine Limitierung beim Payload gibt. Bei einem GET-Request wird die Länge der URL je nach Client/Server/Proxy limitiert.
+
+Grundsätzlich funktioniert der POST-Endpoint gleich wie der GET-Endpoint. Der Body vom Reqeust wird in JSON geschrieben.
+
+Query
+~~~~~
+Beispiel für eine Suche mit diversen Kriterien als Übersicht für die Verwendung des POST-Endpoints. Genauere Details zu den jeweiligen Parametern finden Sie unter :ref:`Solr Suche GET <solr-search-get>`
+
+.. sourcecode:: http
+
+  POST /plone/@solrsearch HTTP/1.1
+  Content-Type: application/json
+  Accept: application/json
+
+  {
+    "q": "Kurz",
+    "fl": "UID,Title",
+    "fq": [
+      "portal_type:opengever.document.document",
+      "path_parent:/fd/os/dossier-1"
+    ],
+    "facet": true,
+    "facet.field": "responsible"
+  }
 
 
 Teamraum Solr Suche
--- a/docs/public/dev-manual/api/searching.rst
+++ b/docs/public/dev-manual/api/searching.rst
@@ -138,11 +138,20 @@ Filters
 ~~~~~~~
 ``fq``: Filtern nach einem bestimmten Wert eines Feldes.
 
-Beispiel für ein Suchabfrage gefiltert nach portal_type nur für Dokumente und Dossiers
+Beispiele für gefilterte Suchabfragen
+
+**Filtern nach ``portal_type`` nur für Dokumente und Dossiers**
 
 .. sourcecode:: http
 
   GET /plone/@solrsearch?fq=portal_type:(opengever.document.document%20OR%20opengever.dossier.businesscasedossier) HTTP/1.1
+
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: opengever/api/tests/test_utils.py [hunk 1]
```diff
@@ -0,0 +1,48 @@
+from opengever.api.utils import recursive_encode
+from opengever.testing import IntegrationTestCase
+
+
+class TestRecursiveEncode(IntegrationTestCase):
```
njohner, line 5: That could probably be a `unittest.TestCase` or a `FunctionalTestCase`

opengever/api/solr_query_service.py
```diff
@@ -387,10 +390,21 @@ def __init__(self, context, request):
         self.response_fields = None
         self.facets = []
 
+    @property
+    def request_payload(self):
+        """Returns the request payload depending on the http request method.
+        """
```
njohner, line 415: Can we cache that, otherwise we will be doing the recursive encoding 3 times for a solr search... something like:
```
        if getattr(self, "_request_payload", None):
              return self._request_payload
        if self.request.method == 'GET':
            self._request_payload = self.request.form
        elif self.request.method == 'POST':
            # JSON always returns unicode strings. We need to encode the values
            # to utf-8 to get the same encoding as with a GET reqeust.
            self._request_payload = recursive_encode(json_body(self.request))
        return getattr(self, "_request_payload", None)

File: docs/public/dev-manual/api/searching.rst [hunk 3]
```diff
@@ -108,10 +108,13 @@ darzustellen, kann ``metadata_fields=_all`` verwendet werden.
...
 
-Solr Suche
-----------
+.. _solr-search-get:
+
+Solr Suche GET
+--------------
 Um direkt eine Suchabfrage an den Solr Service abzusetzen, steht der ``@solrsearch`` Endpoint zur Verfügung. Für die Abfrage kann SOLR API Syntax und dessen Parameter verwendet werden. Der Endpoint liefert ein Liste von Treffern zurück mit den im Parameter ``fl`` definierten Felder. Die folgende Parameter sind zurzeit vom Endpoint unterstützt:
 
+Achtung: Eine URL hat je nach Client/Server/Proxy eine maximale länge. Für grössere Queries muss der Endpoint mit als POST-Request abgefragt werden.
```
njohner, line 117: mit als -> als

opengever/api/tests/test_solrsearch.py
```diff
@@ -796,3 +797,116 @@ def test_blacklisted_attributes_stats_are_skipped(self, browser):
...
+
+        self.assertEqual(
+            {u'message': u'Solr is not enabled on this GEVER installation.',
+             u'type': u'BadRequest'}, browser.json)
+
+    @browsing
+    def test_raises_internal_error_for_invalid_queries(self, browser):
+        self.login(self.regular_user, browser=browser)
+        with browser.expect_http_error(500):
+            url = u'{}/@solrsearch?q=OR'.format(self.portal.absolute_url())
```
njohner, line 842: should be `url = u'{}/@solrsearch'.format(self.portal.absolute_url())`, query is in the body

opengever/api/solrsearch.py
```diff
@@ -105,6 +109,33 @@ def add_path_parent_filters(self, filters):
         if (requested_parent_paths):
             filters.extend(make_filters(path_parent=requested_parent_paths))
 
+    def add_url_filters(self, filters):
+        """Beside of the 'path_parent' and 'path' filter, we provide an 'url' (alias '@id')
```
njohner, line 113: Beside of the -> Beside the

opengever/api/solrsearch.py
```diff
@@ -105,6 +109,33 @@ def add_path_parent_filters(self, filters):
...
+        filter to filter by specific urls.
+
+        Usualy, a frontend does not know the physical path of a resource but it always
+        knows the url (@id) of a resource. So this filter can be used to directly
+        fetching specific objects by its 'url'.
```
njohner, line 118: fetching specific objects by its 'url' -> fetch specific objects by their 'url'.

opengever/api/solrsearch.py
```diff
@@ -105,6 +109,33 @@ def add_path_parent_filters(self, filters):
...
+        """
+        requested_urls = []
```
njohner, line 120: `requested_paths` would be better

opengever/api/tests/test_solrsearch.py
```diff
@@ -910,3 +926,19 @@ def test_fl_query(self, browser):
...
+        payload = {
+            'fq': [
+                '@id:{}'.format(self.document.absolute_url()),
+                'url:{}'.format(self.task.absolute_url())
+            ]
+        }
+        browser.open(url, method='POST', data=json.dumps(payload), headers=self.api_headers)
+        search_on_context = browser.json
+
+        self.assertEqual(2, search_on_context['items_total'])
```
njohner, line 944: I think it would also work if you provide relative paths, without any changes. Can you give it a quick try, and if it works, add a test?

njohner (approved):
LGTM.
As a sidenote, you have solved at least part of https://4teamwork.atlassian.net/browse/CA-2000, please update / close that story

opengever/api/solr_query_service.py
elioschmutz: 👍  I'll use the `@memoize`-decorator for a per-request caching

opengever/api/tests/test_solrsearch.py
elioschmutz: Yeah, it should also work with relative paths. But I think I let it as it is since the `fq` attribute-names are `@id` and `url` which implicitly says that we expect urls...

elioschmutz:
> LGTM. As a sidenote, you have solved at least part of https://4teamwork.atlassian.net/browse/CA-2000, please update / close that story

👍  I updated the story and referenced this PR

PR #7302, opened 2021-12-15
Extend KuBContactSource with ogds users.

njohner:
For both dossier participations and recipients ogds users should be available when KuB feature is enabled. This was already the case for SQL contacts and needs to also be supported with KuB contacts.

Note that when creating a document from template, after selecting a recipient from the OGDS, you could select an address, phone number, url and mail address when using SQL contacts. This is not possible anymore, as we have decided not to implement that form in the new UI, and instead use default address, phone number, URL and mail for KuB contacts. For OGDS users there is no default for these fields, so I simply chose to take the first on of each...

For [CA-3350]

## Checklist

- [x] Changelog entry

[CA-3350]: https://4teamwork.atlassian.net/browse/CA-3350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/api/tests/test_templatefolder.py
+++ b/opengever/api/tests/test_templatefolder.py
@@ -241,7 +241,7 @@ class TestDocumentFromTemplatePostWithKubFeatureEnabled(KuBIntegrationTestCase):
         self.assertItemsEqual(self.expected_doc_properties, properties)
 
     @browsing
-    def test_creates_document_from_template_with_recipient(self, mocker, browser):
+    def test_creates_document_from_template_with_kub_recipient(self, mocker, browser):
         self.login(self.secretariat_user, browser)
 
         data = {'template': self.docprops_template.UID(),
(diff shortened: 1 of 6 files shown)


opengever/api/tests/test_templatefolder.py
```diff
@@ -273,6 +273,44 @@ def test_creates_document_from_template_with_recipient(self, mocker, browser):
...
 
+    @browsing
+    def test_creates_document_from_template_with_ogds_recipient(self, mocker, browser):
+        self.login(self.secretariat_user, browser)
+
+        data = {'template': self.docprops_template.UID(),
+                'title': u'New d\xf6cument',
+                'recipient': self.regular_user.getId()}
+
+        self.mock_get_full_entity_by_id(mocker, self.person_jean)
```
phgross, line 284: ist this still necessary?

phgross (approved):
👍 LGTM

opengever/api/tests/test_templatefolder.py
njohner: no indeed

PR #7303, opened 2021-12-16
Fix copy document from workspace as new version when gever document is trashed.

njohner:
Sometimes the document on the Gever side cannot be found, for example because it has been trashed, or maybe moved to a dossier where the current user does not have access. In such cases, when copying a document back from a workspace as a new version, we will instead create a new document.

For [CA-3323]

## Checklist

- [x] Changelog entry

[CA-3323]: https://4teamwork.atlassian.net/browse/CA-3323?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-3323.bugfix
+++ b/changes/CA-3323.bugfix
@@ -1,0 +1,1 @@
+Fix copy document from workspace as new version when gever document is trashed. [njohner]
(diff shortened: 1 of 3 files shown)


phgross (approved):
👍 LGTM

PR #7304, opened 2021-12-16
Make sure customproperty default values are initalized, when content is created.

phgross:
This PR makes sure that custom properties are initialized correctly with default values, if they're provide one. 

The following content creation are handled:

_already ensured, implemented_
 - content creation via form (old and new UI)
 - mail creation via Inbound Mail
 - document creation via Drag&Drop TUS upload (neues UI)
 - document creation via Drag&Drop QuickUpload (old UI)
 - document creation Dokument via Mail-Attachment extraction (old and new UI)

_newly ensured_
 - content creation via API
 - dossier creation via Drag&Drop folder upload
 - content creation via OGGBundle import
 - content creation via `Transporter` (used in multi adminunit communication)
 - document creation from template (old and new UI)
 - dossier creation from template (old and new UI)
 
The concerns about dynamic defaults which lead to errors in content creation could be defused. Because we will only allow "Managers" to define dynamic defaults. This is already being ensured.

For [CA-2851]

## Checklist
- [x] Changelog entry

- New functionality:

[CA-2851]: https://4teamwork.atlassian.net/browse/CA-2851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/base/transport.py
+++ b/opengever/base/transport.py
@@ -4,6 +4,7 @@ from opengever.base.request import dispatch_json_request
 from opengever.base.security import elevated_privileges
 from opengever.ogds.base.utils import decode_for_json
 from opengever.ogds.base.utils import encode_after_json
+from opengever.propertysheets.creation_defaults import initialize_customproperties_defaults
 from opengever.task.reminder import Reminder
 from opengever.task.task import ITask
 from plone.dexterity.interfaces import IDexterityContent
@@ -157,6 +158,7 @@ class DexterityObjectCreator(object):
     # XXX use plone.api
     def create_in(self, container):
         obj = createContent(self.portal_type, title=self.title)
+        initialize_customproperties_defaults(obj)
         notify(ObjectCreatedEvent(obj))
 
         # insert data from collectors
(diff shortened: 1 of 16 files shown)


opengever/base/transport.py
```diff
@@ -164,6 +165,8 @@ def create_in(self, container):
         for name, collector in collectors:
             collector.insert(self.data[name])
 
+        initialize_customproperties_defaults(obj)
```
lukasgraf, line 168: I think we patched `createContent()` (used just above) to set the "regular" default values, right?
So these get initialized before the `ObjectCreatedEvent`, but custom property default values get set after that event has already been fired. Not sure if that could be a problem.

lukasgraf (approved):
LGTM 👍 
Just a small question.

opengever/base/transport.py
phgross: `createContent` is not patched, but i still moved the custom property initialization above the event notify.

PR #7305, opened 2021-12-16
Add support for custom property facets to `@solrsearch` endpoint

lukasgraf:
Custom property fields can now be faceted as well using the `@solrsearch` endpoint (except ones of type `text`).

The names for the Solr fields of custom properties, and in which listings they should appear, can be retrieved using the existing `@listing-custom-fields` endpoint.

For [CA-2613](https://4teamwork.atlassian.net/browse/CA-2613)

## Checklist

- [x] Changelog entry

- API change:
  - [x] API Changelog entry

### DIFF
--- a/opengever/api/solr_query_service.py
+++ b/opengever/api/solr_query_service.py
@@ -14,6 +14,7 @@ from opengever.base.vocabulary import wrap_vocabulary
 from opengever.dossier.indexers import ParticipationIndexHelper
 from opengever.globalindex.browser.report import task_type_helper as task_type_value_helper
 from opengever.propertysheets.definition import SolrDynamicField
+from opengever.propertysheets.storage import PropertySheetSchemaStorage
 from opengever.task.helper import task_type_helper
 from opengever.tasktemplates.content.templatefoldersschema import sequence_type_vocabulary
 from plone import api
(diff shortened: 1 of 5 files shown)


opengever/api/solr_query_service.py
```diff
@@ -13,7 +13,10 @@
 from opengever.base.vocabulary import wrap_vocabulary
 from opengever.dossier.indexers import ParticipationIndexHelper
 from opengever.globalindex.browser.report import task_type_helper as task_type_value_helper
+from opengever.propertysheets.assignment import get_document_assignment_slots
+from opengever.propertysheets.assignment import get_dossier_assignment_slots
```
tinagerber, line 17: ✂️

tinagerber (approved):
I was wondering if it makes sense to have facets for textline fields. But we could also solve that in the frontend if we don't want them.

lukasgraf:
@tinagerber yes, I was wondering the same thing. I checked though, Solr *can* do it properly if a facet for a `textline` field is requested, so it does work. But I agree, it doesn't make a whole lot of sense.

The case I could see is that they might be using it for some simple freetext keyword-like field. That would need to be textline (because they can't define keywords ahead of time), and facets would actually be useful for searching and to eliminate similarly spelled "keywords".

But if we don't want it, we'd have to do it in the frontend anyway - probably with some help of the backend. I don't think the `@listing-custom-fields` endpoint currently gives enough information to identify these textline fields clearly. I'll merge as is for now, and also check back with Phippu.

PR #7308, opened 2021-12-17
Fix save PDF under for workspace documents.

njohner:
Workspace documents have their own workflow, which has only a single state. This breaks save PDF under, as the function sets the document to shadow state and then transitions it back to active state when the PDF has been posted back by bumblebee. This transition did not exist and led to the function failing.

For [CA-3389]

## Checklist

- [x] Changelog entry
- New functionality:
  - [ ] for `document` also works for `mail`: 👎  not available for mails

[CA-3389]: https://4teamwork.atlassian.net/browse/CA-3389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-3389.bugfix
+++ b/changes/CA-3389.bugfix
@@ -1,0 +1,1 @@
+Fix save PDF under for workspace documents. [njohner]
--- a/opengever/api/tests/test_save_document_as_pdf.py
+++ b/opengever/api/tests/test_save_document_as_pdf.py
@@ -193,3 +193,15 @@ class TestSaveDocumentAsPdfPost(IntegrationTestCase):
         self.assertEqual(len(children["added"]), 1)
         created_document = children["added"].pop()
         self.assertEqual(0, IAnnotations(created_document).get(PDF_SAVE_SOURCE_VERSION_KEY))
+
+    @browsing
+    def test_save_document_as_pdf_post_does_not_set_shadow_state_for_workspace_documents(self, browser):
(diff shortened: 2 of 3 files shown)


lukasgraf (approved):
LGTM, but:
If we need a shadow state until the PDF is posted back, won't this be a problem for the workspace documents? And if we don't, why do we have it at all for regular documents? 😁 

I found [the code](https://github.com/4teamwork/opengever.core/blob/c8daffede38b243fe38bc6f14964cd777e9609e4/opengever/document/forms.py#L291-L292) that does it, and [the PR](https://github.com/4teamwork/opengever.core/pull/5042) that originally introduced it, but I honestly couldn't quite figure out what we want to achieve with the shadow state here.

Feel free to :shipit: if it's a non-isssue.

njohner:
I think the shadow state is only there to avoid that other users might see an empty document. But as bumblebee normally answer pretty fast, we decided (discussed with @phgross ) that this was not really an issue. Introducing the workflow would be messy, because states are named differently for workspace documents (because the workflow is generated with lawgiver), so we would need to introduce `if` clauses to handle them separately (`as_shadow_document` and `leave_shadow_state` use state and transition names).

lukasgraf:
I see, thanks for the clarification! 👍

PR #7310, opened 2021-12-20
Use custom Solr distribution containing Log4j 2.17.0

buchi:
Currently there's no official Solr distribution that contains the latest security fixes for Log4j. Thus we use our own for now.

For [CA-3339](https://4teamwork.atlassian.net/browse/CA-3339)

## Checklist

- [x] Changelog entry


lukasgraf (approved):
LGTM 👍

PR #7312, opened 2021-12-20
Introduce PropertySheetsManager role and custom permission 

lukasgraf:
This introduces a new permission `opengever.propertysheets: Manage PropertySheets` that is used to protect `POST` and `DELETE` to the `@propertysheets` endpoint. This permission is assigned to the `Manager` and a new `PropertySheetsManager` role.

Therefore, the `PropertySheetsManager` role can be assigned to a group of users that should be able to manage propertysheets (setting dynamic defaults is still only allowed for the `Manager` role though).

The `@config` endpoint also includes a new key `is_propertysheets_manager` to indicate whether the user is allowed to manage property sheets _(and the respective functionality should be displayed in the frontend)_.

For [CA-2953](https://4teamwork.atlassian.net/browse/CA-2953)

## Checklist

- [x] Changelog entry

- API change:
  - [x] API Changelog entry

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -626,6 +626,7 @@ Users
 - ``self.manager``: ``admin``
 - ``self.meeting_user``: ``herbert.jager``
 - ``self.member_admin``: ``david.meier``
+- ``self.propertysheets_manager``: ``propertysheets.manager``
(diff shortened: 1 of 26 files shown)


phgross (approved):
👍 LGTM

PR #7315, opened 2021-12-22
Create journal PDF when offering dossiers instead of when resolving them

lukasgraf:
This moves the dossier journal PDF creation to the point when a dossier is **offered** in a disposition (instead of when resolving the dossier).

The journal PDF will be created using a new nightly job.

In addition, there will only ever be **one** journal PDF for the main dossier (instead of one for each (sub)dossier)). Therfore all the journal entries are listed in a single table in one PDF, ordered by timestamp. In order to identify the (sub)dossier, there is a new column "Dossier" containing the reference number.

[Example PDF for a Dossier with a Subdossier](https://github.com/4teamwork/opengever.core/files/7763563/Dossier.Journal.Ein.Hauptdossier.22.12.2021.10.20.pdf)

For [CA-3276](https://4teamwork.atlassian.net/browse/CA-3276)

## Checklist

- [x] Changelog entry


phgross (approved):
LGTM 👍

PR #7317, opened 2021-12-23
WOPI check_file_info(): Fall back to document owner if creator isn't set.

lukasgraf:
The `Creator` for documents imported via OGGBundle may not be set (`Creator()` returns empty string).

This fix makes `OwnerId` fall back to `getOwner().getId()` in such a case, in order to make OfficeOnline happy.

For [CA-3318](https://4teamwork.atlassian.net/browse/CA-3318)

## Checklist

- [x] Changelog entry


tinagerber (approved):
I assume that you have made sure that an owner is set for documents imported via OGGBundle.

lukasgraf:
Yep, I confirmed this with an imported bundle on `lab`. 👍

PR #7320, opened 2022-01-03
Inbound mail: Add support for sender address aliases

lukasgraf:
This adds a new registry setting `IInboundMailSettings.sender_aliases` of type dict that can be used to alias sender addresses for incoming mails to a GEVER user. Mails coming from those senders will then be created with the respective user instead of getting rejected as unknown sender.

The lookup for an alias in the `email address -> userid` mapping is **case-insensitive** (because users might easily have a irregular casing configured in their MUA). The lookup of a user based on the `userid` however is **case-sensitive** (`getUserById()`) - but since we control that aspect, this shouldn't be an issue. 

The regular user lookup via sender email address has precedence - if the sender address can be resolved to a user, any potential entries in alias table will be ignored.

The PR to configure the aliases for RefBl is at 4teamwork/opengever.refbl#7

For [CA-2598](https://4teamwork.atlassian.net/browse/CA-2598)

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-2598.feature
+++ b/changes/CA-2598.feature
@@ -1,0 +1,1 @@
+Inbound mail: Add support for sender address aliases [lgraf]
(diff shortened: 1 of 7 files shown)


buchi (approved):
LGTM

See inline comment about handling mappings to invalid userids.

PR #7326, opened 2022-01-07
Exclude exceptions that happen in normal usage from sentry.

njohner:
I did not like the idea of excluding errors using the `RAVEN_DISABLE_EXCEPTIONS` environment variable from `ftw.raven`, notably because it only allows to exclude whole classes of exceptions and because it is not easily visible in the code. Also this did not seem like a very future proof solution for when we move away from ftw.raven. 

Instead I chose to not report to sentry all exceptions inheriting from a newly introduced `NotReportedException` class. This allows to specifically disregard certain exceptions in sentry, while still returning the correct error codes and types over the API.

I went over all sentry issues and checked which ones should be disregarded. For a few of them, notably validation errors which are all handled generically (https://github.com/4teamwork/opengever.core/blob/master/opengever/api/deserializer.py#L51-L57), it was not easily possible to avoid reporting them to sentry, so I left it as is. There are also some that I decided not to disregard, because I actually think they should not happen in normal circumstances, and should be avoided in the gever-ui (for example adding twice the same participant on a dossier). As I've gone over the issues I've made a list of a few that we should really look at (see [CA-2755]).

Here the sentry issues that will get resolved with this PR:
- https://sentry.4teamwork.ch/organizations/sentry/issues/75285
- https://sentry.4teamwork.ch/organizations/sentry/issues/79698
- https://sentry.4teamwork.ch/organizations/sentry/issues/80604
- https://sentry.4teamwork.ch/organizations/sentry/issues/75589
- https://sentry.4teamwork.ch/organizations/sentry/issues/76770
- https://sentry.4teamwork.ch/organizations/sentry/issues/72929
- https://sentry.4teamwork.ch/organizations/sentry/issues/68672
- https://sentry.4teamwork.ch/organizations/sentry/issues/73541
- https://sentry.4teamwork.ch/organizations/sentry/issues/77707
- https://sentry.4teamwork.ch/organizations/sentry/issues/79565
- https://sentry.4teamwork.ch/organizations/sentry/issues/80091
- https://sentry.4teamwork.ch/organizations/sentry/issues/80346
- https://sentry.4teamwork.ch/organizations/sentry/issues/79564

For [CA-2755]

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [ ] api-change label added

[CA-2755]: https://4teamwork.atlassian.net/browse/CA-2755?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1467,3 +1467,9 @@ which can be used in upgrade steps:
             indexer.add_by_obj(obj)
 
 This will register the corresponding jobs to the ``NightlyMaintenanceJobsProvider``.
+
+
+API Error handling
+------------------
+
+Errors, especially client errors, are a normal part of the API. ZPublisher's ``HTTPResponse`` will set the proper error codes while ``Plone.rest`` will serialize these errors back to the client. This allows to simply raise errors such as ``BadRequest`` in the API Services, and the rest will happen automatically. This does not prevent the error from being raised and therefore be handled by ``ftw.raven`` and logged to sentry. Specific exceptions that we know will happen in normal Gever operations should not be reported to sentry, which can be easily achieved by raising an exception inheriting from ``NotReportedException``.
--- a/changes/CA-2755.other
+++ b/changes/CA-2755.other
@@ -1,0 +1,1 @@
+Avoid reporting normal API exceptions in sentry. [njohner]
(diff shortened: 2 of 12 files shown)


tinagerber (approved):
LGTM. I wonder if it would be good to mention the concept in the README or API doc or something.

njohner:
@tinagerber , right I'd actually planned on doing so, but seems I forgot. Will amend.

lukasgraf (approved):
LGTM! 👍

PR #7330, opened 2022-01-12
Add `@propertysheet-metaschema` endpoint

lukasgraf:
This endpoint produces a JSON schema for propertysheet definitions, to be used by the frontend to render add/edit forms .

The JSON schema is generated from a zope.schema that `POST @propertysheets` requests are validated against.

The way that errors are handled and formatted in `POST @propertysheets` is not final yet - this will be unified in a next step.

#### Review Notes

It's probably easiest to first take a glance at the [final metaschema](https://github.com/4teamwork/opengever.core/pull/7330/files#diff-238a09a233531eb8d48f7eddaee530a59fdc12870955968ba8c1a00ea2af2fcdR34-R148), and possibly at the [JSON schema produced from it](https://github.com/4teamwork/opengever.core/pull/7330/files#diff-1016374353e066c013da338f0003faebb12381702213d8cfcb5a658adf078da9), to get an idea of where this is going.

## Checklist

- [x] Changelog entry

- API change:
  - [x] API Changelog entry
- New translations


opengever/base/schema.py
```diff
@@ -66,6 +67,48 @@ def __init__(self, columns=tuple(), vocabulary_depends_on=(),
...
+            else:
+                yield pytype
+
+    def _validate(self, value):
+        if not value:
+            return
+
+        if type(value) not in self.allowed_pytypes:
+            raise WrongType(
+                'Default value %r of type %r not allowed for its '
```
njohner, line 106: That is a bit awkward, as this is a generic field type, and you raise an exception specific to its use for default values. But we can address that if we ever reuse that Field type somewhere else.

njohner (approved):
LGTM 🎉

opengever/base/schema.py
lukasgraf: Oh, you're right, good catch.
This sort of foreshadows the headaches I'm gonna have once I implement error handling (or rather, consistent error _formatting_ and serialization).

The challenge will be turning rather generic validatiob exceptions into specific / helpful error messages that stand on their own, for the whole form. Here I took a cheeky shortcut it seems 😉 

I'll address this in some way during the error handling PR 👍

PR #7331, opened 2022-01-13
`@propertysheets` GET: Return same format as POST, instead of JSON schema

lukasgraf:
Use the same custom format that POST requests accept as input for GET responses (and the POST response upon successful creation).

This format, unlike the serialization to a JSON schema, is much more compact, and more importantly, **information-preserving**. For example:

In propertysheet definitions, we distinguish between `texline` and `multiple_choice` (of textline) fields on the field type level. In JSON schema, these both get serialized to a field with a `type` of `string`, but the second has a vocabulary.

Also, by using the same JSON format for serialization and deserialization with the `@propertysheets` endpoint, it will finally follow the behavior that any response it gives, will
- result in the creation of a equivalent sheet if POSTed
- not result in a meaningful modification if sent as a PATCH payload

The JSON schema serialization of a specific propertysheet is instead moved to `@schema/virtual.propertysheet.<sheet_id>`.

## Checklist

- [x] Changelog entry

- API change:
  - [x] API Changelog entry
  - If breaking:
    - [x] api-change label added
        _(deemed unnecessary, propertysheets API was only used internally so far)_
    - [ ] Scrum master is informed _(same)_

### DIFF
--- a/changes/CA-2954-1.feature
+++ b/changes/CA-2954-1.feature
@@ -1,0 +1,1 @@
+Expose propertysheet JSON schemas under `@schema` endpoint. [lgraf]
--- a/changes/CA-2954-2.feature
+++ b/changes/CA-2954-2.feature
@@ -1,0 +1,1 @@
+GET @propertysheets: Return same format as POST, instead of JSON schema. [lgraf]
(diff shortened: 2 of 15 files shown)


njohner (approved):
LGTM!

PR #7332, opened 2022-01-14
Fix file paths in OGG bundles and improve documentation

njohner:
When creating an OGG bundle, the file paths written to the bundle were relative, which led to the files not being found during bundle import (it would require copying the files over into the bundle itself to work).
I've also added an example in the documentation.

For [CA-3289]

## Checklist

- [x] Changelog entry


njohner:
@buchi is that a change that is OK with you? I just realised that currently creating and importing a bundle would not work because of the file paths, except if you copy the files over into the bundle. Maybe this is intentional because of how you do such imports for productive deployments?

buchi:
I would prefer the possibility to specify the base path on import time and still export relative paths. Absolute file paths probably only work if you do the export and import on the same server.

njohner:
Ok I updated so that the path to the files directory can be specified on import time. it defaults to the bundle path.

changes/CA-3289.other
```diff
@@ -0,0 +1 @@
+Make file paths absolute paths in OGG bundles. [njohner]
```
buchi, line 1: The changelog entry still reflects the previous approach.

buchi (approved):
👍🏻 LGTM

njohner:
I don't quite remember what the issue was here, but we had discussed this and found there were issues with that approach. I'll close it

PR #7333, opened 2022-01-16
Add PATCH support for `@propertysheets` endpoint

lukasgraf:
This implements PATCH support for the `@propertysheets` endpoint.

PATCH semantics only apply to the topmost level (assignments and fields) - if either of those keys is omitted, it will be preserved as-is. Omitting a field from the list of fields however will **delete** it. The field list as a whole therefore needs to be modified by the caller, and always be sent in its entirety.

Dynamic defaults also need some special handling for PATCH:

We also allow non-Managers to PATCH propertysheets that contain dynamic defaults (which otherwise may only be set by managers), if those dynamic defaults already existed before and don't get modified by the PATCH request.

This is necessary to allow regular PropertySheetsManagers to allow saving the form with modifications to other things that don't concern the dynamic defaults.

## Checklist

- [x] Changelog entry

- API change:
  - [x] API Changelog entry

### DIFF
--- a/opengever/propertysheets/api/base.py
+++ b/opengever/propertysheets/api/base.py
@@ -1,0 +1,75 @@
+from opengever.propertysheets.metaschema import IPropertySheetDefinition
+from opengever.propertysheets.storage import PropertySheetSchemaStorage
+from plone.restapi.services import Service
+from zExceptions import BadRequest
+from zExceptions import NotFound
+from zope.interface import implementer
+from zope.publisher.interfaces import IPublishTraverse
+
+
+@implementer(IPublishTraverse)
+class PropertySheetLocator(Service):
+    """Locates a propertysheet definition by its sheet_id.
+
+    This is a Service base class for all services that need to look up a
+    propertysheet definition via a /@propertysheets/{sheet_id} style URL.
+
+    It handles
+    - extraction of the {sheet_id} path parameter
+    - error response for incorrect number of path parameters
+    - validation of {sheet_id}
+    - return of a 404 Not Found response if that sheet doesn't exist
+    - retrieval of the respective sheet
+    in a single place so that not every service has to implement this again,
+    and we ensure consistent behavior across all services.
+
+    Because the GET service supports both individual retrieval as well as
+    listing, this needs to be handled here as well.
+    """
+
+    def __init__(self, context, request):
+        super(PropertySheetLocator, self).__init__(context, request)
+        self.params = []
+
+    def publishTraverse(self, request, sheet_id):
+        # Consume any path segments after /@propertysheets as parameters
+        self.params.append(sheet_id)
+        return self
+
+    def get_sheet_id(self):
+        sheet_id_required = getattr(self, 'sheet_id_required')
+
+        if sheet_id_required:
+            if len(self.params) != 1:
+                raise BadRequest(
+                    'Must supply exactly one {sheet_id} path parameter.')
+        else:
+            # We'll accept zero (listing) or one (get by id) params, but not more
+            if len(self.params) > 1:
+                raise BadRequest(
+                    'Must supply either exactly one {sheet_id} path parameter '
+                    'to fetch a specific property sheet, or no parameter for a '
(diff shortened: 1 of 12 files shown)


opengever/propertysheets/tests/test_schema_definition_patch.py
```diff
@@ -0,0 +1,313 @@
...
+            )
+        self.assertDictContainsSubset(
+            {
+                u"message": u"The 'id' of an existing sheet must not be changed.",
+                u"type": u"BadRequest",
+            },
+            browser.json,
+        )
+
+    @unittest.expectedFailure
```
lukasgraf, line 227: This is a test that describes how I wish the system would behave - but it currently doesn't.
Removing a propertysheet field doesn't immediately lead to deletion of old data from objects. So, removing a field and re-adding it *quickly* would preserve existing data.

But as soon as those objects get edited again while the field is temporarily missing, the old data gets overwritten. I'll create a follow up story for this.

njohner: I don't have a strong opinion that data should be preserved.
Something I am wondering on the other hand, is what happens if I patch a choice filed, changing the allowed values and then some objects have a value that is not in the vocabulary anymore?

opengever/propertysheets/api/base.py
```diff
@@ -120,16 +120,30 @@ def validate_fields(self, fields):
...
-            dynamic_defaults = PropertySheetSchemaDefinition.DYNAMIC_DEFAULT_PROPERTIES
-            if any(p in field_data for p in dynamic_defaults):
+            self.validate_dynamic_defaults(field_data, existing_dynamic_defaults)
+
+        return errors
+
+    def validate_dynamic_defaults(self, field_data, existing_dynamic_defaults):
+        """Require Manager role for any kind of dynamic defaults, unless
+        it's a PATCH request and they already existed and didn't get modified.
+        """
```
njohner, line 130: You could start with `if api.user.has_permission('cmf.ManagePortal'): return`. Might be a bit more efficient to check that just once and skip the rest if the user is manager... would also allow to remove the `else` clause below...

njohner (approved):
LGTM

opengever/propertysheets/api/patch.py
```diff
@@ -0,0 +1,87 @@
...
+
+    Syntax:
+
+    PATCH /@propertysheets/question HTTP/1.1
+    {
+        "fields": [{"name": "foo", field_type": "text", "required": true}],
+    }
+    """
+
+    sheet_id_required = False
```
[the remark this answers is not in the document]
lukasgraf: Absolutely. Thanks!

opengever/propertysheets/api/patch.py
```diff
@@ -0,0 +1,87 @@
...
+
+            seen = set()
+            duplicates = []
+            for name in [each["name"] for each in fields]:
+                if name in seen:
+                    duplicates.append(name)
+                seen.add(name)
+            if duplicates:
+                raise BadRequest(
+                    u"Duplicate fields '{}'.".format("', '".join(duplicates))
```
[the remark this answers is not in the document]
lukasgraf: Yes (to both). I think it already gets validated, but was difficult to extract so far. But I'm currently moving that code around anyway with the error handling, and centralising as much as possible, so I might do it in the error handling PR.

opengever/propertysheets/tests/test_schema_definition_patch.py
lukasgraf: Good point. Probably sadness ensues. 😢 
I'll check what exactly happens in this case. What we decided to do for the first case, is to disallow changing the `name` and `field_type` of an existing field in the edit form, by making those read-only in the user-interface.

Potentially this could be handled similarly, by only allowing to add to vocabularies for existing fields, but not remove from them.

opengever/propertysheets/api/base.py
lukasgraf: Good point. I'll change that with the next PR 👍

PR #7336, opened 2022-01-18
Serialize `@propertysheet` errors in a frontend-friendly way

lukasgraf:
This overhauls the error handling in the `@propertysheets` APIs in a way that the client-side errors get serialized in a way where they contain a deliberately crafted `translated_message`.

We do this to enable the frontend to just display this one message at the top of the form, without any further processing of the response or highlighting of the offending widget.

The message must therefore contain enough context so that the user can identify where the error is, and what he needs to change to fix it.

This generated `translated_message` is a multi-line string (three, sometimes four lines), where each line gets progressively more specific / local.

Example:
```
    The form contains errors:
    Field 3 ('location'):
    Parameter 'title': Value is too long
    (Max: 128 characters, Actual length: 135 characters)
```

In order to achieve this, we wrap certain validation errors in some custom higher level exceptions, to which we attach the necessary context info to build a rich `translated_message` later:

- `FieldValidationError` (for errors in the `fields` key)
- `AssignmentValidationError` ((for errors in the `assignments` key)
- `SheetValidationError` (everything else)

These then get processed by adapters that know how the data attached to these exceptions is structured, and they extract the necessary context information from them.

## Checklist

- [x] Changelog entry

- API change:
  - [x] API Changelog entry
  - If breaking:
    - [x] api-change label added
    - [ ] #delivery channel notified about breaking change _(deemed unnecessary)_
    - [ ] Scrum master is informed _(deemed unnecessary)_
- New translations

### DIFF
--- a/opengever/propertysheets/api/base.py
+++ b/opengever/propertysheets/api/base.py
@@ -4,6 +4,7 @@ from opengever.propertysheets.api.error_serialization import ErrorSerializer
 from opengever.propertysheets.definition import PropertySheetSchemaDefinition as PSDefinition
 from opengever.propertysheets.exceptions import AssignmentAlreadyInUse
 from opengever.propertysheets.exceptions import AssignmentValidationError
+from opengever.propertysheets.exceptions import FieldValidationError
 from opengever.propertysheets.exceptions import SheetValidationError
 from opengever.propertysheets.metaschema import IFieldDefinition
 from opengever.propertysheets.metaschema import IPropertySheetDefinition
@@ -112,8 +113,32 @@ class PropertySheetWriter(PropertySheetLocator):
     """Base class for @propertysheets endpoints that create or modify sheets.
     """
 
+    def get_fields(self, data, existing_dynamic_defaults=()):
+        fields = data.get("fields")
+
+        if fields:
+            errors = self.validate_fields(fields, existing_dynamic_defaults)
+            if errors:
+                raise FieldValidationError(errors)
+
+            seen = set()
+            duplicates = []
+            for name in [each["name"] for each in fields]:
+                if name in seen:
+                    duplicates.append(name)
+                seen.add(name)
+            if duplicates:
+                raise BadRequest(
+                    u"Duplicate fields '{}'.".format("', '".join(duplicates))
+                )
+
+        return fields
+
     def validate_fields(self, fields, existing_dynamic_defaults=()):
         errors = []
+
+        if not isinstance(fields, list):
+            raise BadRequest(u"Missing or invalid field definitions.")
 
         for field_no, field_data in enumerate(fields):
             # Cast JSON strings to their appropriate Python types (unicode or
--- a/opengever/propertysheets/api/base.py
+++ b/opengever/propertysheets/api/base.py
@@ -173,6 +173,9 @@ class PropertySheetWriter(PropertySheetLocator):
         """Require Manager role for any kind of dynamic defaults, unless
         it's a PATCH request and they already existed and didn't get modified.
         """
+        if api.user.has_permission('cmf.ManagePortal'):
+            return
+
         dynamic_default_types = PSDefinition.DYNAMIC_DEFAULT_PROPERTIES
 
         for name, value in field_data.items():
@@ -181,11 +184,10 @@ class PropertySheetWriter(PropertySheetLocator):
             if (field_data['name'], name, value) in existing_dynamic_defaults:
                 # Existing dynamic default that is left unchanged - allowed
                 continue
-            else:
-                # New or modified dynamic default - managers only
-                if not api.user.has_permission('cmf.ManagePortal'):
-                    raise Unauthorized(
-                        'Setting any dynamic defaults requires Manager role')
+
+            # New or modified dynamic default - managers only
+            raise Unauthorized(
+                'Setting any dynamic defaults requires Manager role')
 
     def get_assignments(self, data, sheet=None):
         assignments = data.get("assignments")
--- a/opengever/propertysheets/api/base.py
+++ b/opengever/propertysheets/api/base.py
@@ -3,6 +3,7 @@ from opengever.api.validation import scrub_json_payload
 from opengever.propertysheets.api.error_serialization import ErrorSerializer
 from opengever.propertysheets.definition import PropertySheetSchemaDefinition as PSDefinition
 from opengever.propertysheets.exceptions import AssignmentAlreadyInUse
+from opengever.propertysheets.exceptions import AssignmentValidationError
 from opengever.propertysheets.exceptions import SheetValidationError
 from opengever.propertysheets.metaschema import IFieldDefinition
 from opengever.propertysheets.metaschema import IPropertySheetDefinition
@@ -154,6 +155,17 @@ class PropertySheetWriter(PropertySheetLocator):
                 if not api.user.has_permission('cmf.ManagePortal'):
                     raise Unauthorized(
                         'Setting any dynamic defaults requires Manager role')
+
+    def get_assignments(self, data, sheet=None):
+        assignments = data.get("assignments")
+        assignment_errors = self.validate_assignments(assignments, sheet=sheet)
+        if assignment_errors:
+            raise AssignmentValidationError(assignment_errors)
+
+        if assignments is not None:
+            assignments = tuple(assignments)
+
+        return assignments
 
     def validate_assignments(self, assignments_data, sheet=None):
         if not assignments_data:
--- a/opengever/propertysheets/api/base.py
+++ b/opengever/propertysheets/api/base.py
@@ -1,6 +1,9 @@
 from opengever.api.validation import get_validation_errors
 from opengever.api.validation import scrub_json_payload
+from opengever.propertysheets.api.error_serialization import ErrorSerializer
 from opengever.propertysheets.definition import PropertySheetSchemaDefinition as PSDefinition
+from opengever.propertysheets.exceptions import AssignmentAlreadyInUse
+from opengever.propertysheets.exceptions import SheetValidationError
 from opengever.propertysheets.metaschema import IFieldDefinition
 from opengever.propertysheets.metaschema import IPropertySheetDefinition
 from opengever.propertysheets.storage import PropertySheetSchemaStorage
@@ -10,6 +13,7 @@ from plone.restapi.services import Service
 from zExceptions import BadRequest
 from zExceptions import NotFound
 from zExceptions import Unauthorized
+from zope.annotation import IAnnotations
 from zope.component import getMultiAdapter
 from zope.interface import implementer
 from zope.publisher.interfaces import IPublishTraverse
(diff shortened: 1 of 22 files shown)


opengever/propertysheets/api/error_serialization.py
```diff
@@ -0,0 +1,282 @@
...
+        field = IPropertySheetDefinition['assignments']
+        error_location = translate(
+            field.title,
+            context=self.request,
+        ) + u':'
+        return error_location
+
+    def get_error_short_text(self):
+        validation_exc = self.get_validation_error()
+        error_description = self.extract_error_description(validation_exc)
```
njohner, line 246: ✂️

opengever/propertysheets/api/error_serialization.py
```diff
@@ -0,0 +1,282 @@
...
+
+        if isinstance(validation_exc, AssignmentAlreadyInUse):
+            # Provide a custom, more helpful error message for this
+            error_description = translate(
+                _(u'msg_assignment_already_in_use',
+                  default=u"Assignment '${assignment}' is already in use.",
+                  mapping={'assignment': validation_exc.message['assignment']}),
+                context=self.request,
+            )
```
njohner, line 255:
```
else:
    error_description = self.extract_error_description(validation_exc)
```

opengever/propertysheets/api/error_serialization.py
```diff
@@ -0,0 +1,282 @@
...
+    def get_error_location(self):
+        # We currently only raise SheetValidationError for ID field errors.
+        field = IPropertySheetDefinition['id']
+        error_location = translate(
+            field.title,
+            context=self.request,
+        ) + u':'
+        return error_location
+
+    def get_error_short_text(self):
```
njohner, line 277: ✂️ I think this is the same as in the baseclass

opengever/propertysheets/locales/de/LC_MESSAGES/opengever.propertysheets.po
```diff
@@ -146,3 +151,25 @@ msgstr "Titel"
 #: ./opengever/propertysheets/metaschema.py
 msgid "label_values"
 msgstr "Wertebereich"
+
+#. Default: "Assignment '${assignment}' is already in use."
+#: ./opengever/propertysheets/api/error_serialization.py
+msgid "msg_assignment_already_in_use"
+msgstr "Der Slot '${assignment}' ist bereits belegt."
```
njohner, line 166: I don't really like "Slot" that we've been using in German. Why not something like Zuordnung, Zuteilung or Zuweisung

opengever/propertysheets/tests/test_api_error_formatting.py
```diff
@@ -654,3 +654,655 @@ def test_rejects_non_string_term_type_for_values(self, browser):
...
+        # Then attempt to patch it
+        browser.open(
+            view="@propertysheets/%s" % sheet_id,
+            method="PATCH",
+            data=json.dumps(data),
+            headers=self.api_headers,
+        )
+
+    @browsing
+    def test_rejects_missing_sheet_id(self, browser):
```
njohner, line 705: ✂️ ? These are tests for the PATCH endpoint where modifying the ID is anyway forbidden, and you seem to be testing POST here

opengever/propertysheets/tests/test_api_error_formatting.py
```diff
@@ -654,3 +654,655 @@ def test_rejects_non_string_term_type_for_values(self, browser):
...
+        self.assertDictContainsSubset(
+            {
+                u"message": u"Must supply exactly one {sheet_id} path parameter.",
+                u"type": u"BadRequest",
+            },
+            browser.json,
+        )
+
+    @browsing
+    def test_rejects_sheet_id_not_matching_pattern(self, browser):
```
njohner, line 727: Huh, I thought this would raise `BadRequest("The 'id' of an existing sheet must not be changed.")`

opengever/propertysheets/tests/test_api_error_formatting.py
```diff
@@ -654,3 +654,655 @@ def test_rejects_non_string_term_type_for_values(self, browser):
...
+                    u"Das Formular enth\xe4lt Fehler:",
+                    u"ID:",
+                    u"Ung\xfcltiger Bezeichner",
+                ]),
+            },
+            browser.json,
+        )
+
+    @browsing
+    def test_rejects_sheet_id_that_is_too_long(self, browser):
```
njohner, line 753: Same here

opengever/propertysheets/tests/test_api_error_formatting.py
```diff
@@ -654,3 +654,655 @@ def test_rejects_non_string_term_type_for_values(self, browser):
...
+                    u"ID:",
+                    u"Wert ist zu lang.",
+                    u"(Maximum: 32 Zeichen. Tats\xe4chliche L\xe4nge: 34 Zeichen)",
+                ]),
+            },
+            browser.json,
+        )
+
+    @browsing
+    def test_rejects_sheet_id_that_is_a_python_keyword(self, browser):
```
njohner, line 775: and here.

opengever/propertysheets/tests/test_api_error_formatting.py
```diff
@@ -654,3 +654,655 @@ def test_rejects_non_string_term_type_for_values(self, browser):
...
+    def test_rejects_assignments_already_in_use(self, browser):
+        self.login(self.propertysheets_manager, browser)
+
+        data = deepcopy(self.VALID_SAMPLE_PAYLOAD)
+
+        # IDossier.default is already used by a fixture object
+        data['assignments'] = [u"IDossier.default"]
+
+        with browser.expect_http_error(400):
+            self.patch_sheet(browser, data)
```
njohner, line 849: same

njohner (approved):
LGTM! 🎉

opengever/propertysheets/api/base.py
```diff
@@ -106,7 +114,7 @@ class PropertySheetWriter(PropertySheetLocator):
     def validate_fields(self, fields, existing_dynamic_defaults=()):
         errors = []
 
-        for field_data in fields:
+        for field_no, field_data in enumerate(fields):
```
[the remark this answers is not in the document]
lukasgraf: Yes, this is actually somewhat intentional.
Later, we'll probably want to use the field number to produce machine-readable field selectors like `fields[3].title` along with the response, to indicate to the frontend which field contains the error (the idea of this addressing scheme being "Key `fields`, Item `3` in that array, Key `title`). For that we'd want zero-based indexing.

So I chose to do the `field_no + 1` in the place where we render a human-facing number. (I'm not considering developers human in that context) 😝

opengever/propertysheets/api/error_serialization.py
```diff
@@ -0,0 +1,282 @@
...
+        error_location = _(
+            u'msg_propertysheet_field_location',
+            default=u"Field ${field_no} ('${field_name}'):",
+            mapping={
+                'field_no': self.propertysheet_field_no + 1,
+                'field_name': self.propertysheet_field_name,
+            })
+        return translate(error_location, context=self.request)
+
+    def get_error_short_text(self):
```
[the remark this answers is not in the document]
lukasgraf: Ah yes, that's why I pulled it up into the superclass in the first place 🤦  Thanks!

opengever/propertysheets/locales/de/LC_MESSAGES/opengever.propertysheets.po
lukasgraf: I actually worked with "Zuweisung" in an earlier version. But I wasn't happy with it at all, because it barely adds any clarity, and it sounds weird when used in a sentence like that. The three you suggested work fine as simply the label for the field, but when used in a sentence it falls apart IMHO:

"Die Zuweisung 'IDocument.default' wird bereits verwendet" sounds very strange to my ears. So I just went with the technical English name, since the audience is rather technical anyway, and it's sightly less awkward (IMHO).

But I agree, it's definitely not ideal, the wording probably needs a second pass in some other places as well. I just don't think the ones you mentioned are an unequivocal improvement in this case.

opengever/propertysheets/tests/test_api_error_formatting.py
lukasgraf: You're right, I accidentally POST here instead of PATCHing - fixed that.

The rest is correct though: A sheet ID _in the URL_ is always required for PATCH. It's used to address the resource that is supposed to be mutated. An `id` in the payload (JSON body) however is what would indicate that the client wants to modify the `id`, and that's indeed forbidden.

But that code is only ever reached once the sheet ID has been extracted from the URL, validated for lexical correctness, and resolved to an actual sheet. That is all being done by the [`PropertySheetLocator`](https://github.com/4teamwork/opengever.core/pull/7336/files#diff-22949404b71524e015383695cfaf120aff9a54293b557e43c9be8e4faf1a1c10L35) before the actual contents of the payload are even getting looked at.

opengever/propertysheets/tests/test_api_error_formatting.py
lukasgraf: Nope, for the reason described above: Changing a sheet ID would be done with `{"id": "new_id"}` in the body. This however rejects a malformed it as part of the URL scheme that addresses a particular sheet, e.g. `PATCH /@propertysheets/$$$`

opengever/propertysheets/tests/test_api_error_formatting.py
lukasgraf: Indeed 😉

opengever/propertysheets/tests/test_api_error_formatting.py
lukasgraf: Are you sure you meant to comment on this line? This doesn't seem like the same issue like the previous comments.

opengever/propertysheets/api/error_serialization.py
lukasgraf: I don't think anything can really be eliminated here. The changes as suggested by you would really just _add_ an `else` clause here, which I don't think is what you wanted to suggest, because if I remember correctly, you don't like unnecessary `else` clauses either 😉

opengever/propertysheets/locales/de/LC_MESSAGES/opengever.propertysheets.po
njohner: Works for me. I was just a but "surprised" when I did the French translations that German was using the english "Slot", while english was using "Assignment". In French I finally went for a direct translation of "Assignment" and not Slot. That's just why I was asking here. We could also use "Slot" in German, French and English...

opengever/propertysheets/tests/test_api_error_formatting.py
njohner: Ok thanks 👍 . I did think something was raising earlier, but didn't know what 😄 .

opengever/propertysheets/tests/test_api_error_formatting.py
njohner: yes indeed

PR #7337, opened 2022-01-18
Improve upgrade memory usage

njohner:
With this PR we modify the `NightlyIndexer` and `DefaultValuePersister` to store UIDs instead of IntIDs in their queues. This allows to iterate over brains instead of objects in the upgrade steps (although the only upgrade currently using `DefaultValuePersister` still needs to fetch the objects to decide whether a given objects needs persisting or not).
As storing the UIDs in the queue is less memory efficient than storing the IntIDs, we have left `NightlyWorkflowSecurityUpdater` untouched, i.e. it still stores IntIDs in its queue, as it is anyway necessary to get the objects during the upgrade step to check for placeful workflows.
I also had to write a migration for maintenance jobs from `NightlyIndexer` and `DefaultValuePersister` that might still be in queues but with IntIDs instead of the UIDs that are now necessary for the execution of these jobs.

For [CA-3289]

## Checklist
- [x] Changelog entry

[CA-3289]: https://4teamwork.atlassian.net/browse/CA-3289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/core/upgrades/20211221135724_add_dossier_type_index/upgrade.py
+++ b/opengever/core/upgrades/20211221135724_add_dossier_type_index/upgrade.py
@@ -20,5 +20,5 @@ class AddDossierTypeIndex(UpgradeStep):
 
         with NightlyIndexer(idxs=["dossier_type"],
                             index_in_solr_only=True) as indexer:
-            for obj in self.objects(query, 'Index dossier_type in Solr'):
-                indexer.add_by_obj(obj)
+            for brain in self.brains(query, 'Index dossier_type in Solr'):
+                indexer.add_by_brain(brain)
--- a/changes/CA-3289.other
+++ b/changes/CA-3289.other
@@ -1,0 +1,1 @@
+Improve memory usage for upgrade steps using nightly maintenance jobs. [njohner]
--- a/opengever/core/tests/test_upgrade.py
+++ b/opengever/core/tests/test_upgrade.py
@@ -272,7 +272,6 @@ class TestDefaultValuePersister(IntegrationTestCase):
         runner.execute_pending_jobs()
 
     def test_persists_only_passed_fields(self):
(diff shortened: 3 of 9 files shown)


opengever/core/upgrades/20211221135724_add_dossier_type_index/upgrade.py
```diff
@@ -20,5 +20,5 @@ def __call__(self):
 
         with NightlyIndexer(idxs=["dossier_type"],
                             index_in_solr_only=True) as indexer:
-            for obj in self.objects(query, 'Index dossier_type in Solr'):
-                indexer.add_by_obj(obj)
+            for brain in self.brains(query, 'Index dossier_type in Solr'):
+                indexer.add_by_obj(brain)
```
lukasgraf, line 24: 💥 This fails with `TypeError: ('Could not adapt', <opengever.base.brain.TranslatedTitleBrain object at 0x116543ad0>, <InterfaceClass plone.uuid.interfaces.IUUID>)` on a deployment that is upgraded from a state before the `Index dossier_type in Solr` was executed.

lukasgraf (changes requested):

opengever/core/upgrades/20211221135724_add_dossier_type_index/upgrade.py
njohner: Sorry, had testsed several times and combinations, but did some last minute modifications and thought I'd make a final test after the review...
Anyway thanks for spotting this. I've tested again the migration as well as the upgrade steps after the changes. I think now everything works fine

lukasgraf (approved):
LGTM 👍

PR #7340, opened 2022-01-24
Add parameter to include full representation in @actors endpoint

tinagerber:
This saves some requests in the frontend, because the @actors endpoint can now alse be used to fetch the user/group/... info if needed.

For [CA-2616]

## Checklist

- [x] Changelog entry

- API change:


njohner (approved):
LGTM

PR #7341, opened 2022-01-24
Use relative paths for the @solrsearch path_parent filter query.

elioschmutz:
This PR changes the `path_parent` filter query.

Since the physical path of an object is only for internal use, we have to change the `path_parent` filter query to accept relativ paths.

For [CA-2174]

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [x] api-change label added
- New functionality:
- DB-Schema migration
- New translations

[CA-2174]: https://4teamwork.atlassian.net/browse/CA-2174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -8,7 +8,7 @@ API Changelog
 
 Breaking Changes
 ^^^^^^^^^^^^^^^^
-
+- ``@solrsearch:``: Change ``path_parent`` filter query to no longer expect physical paths but relative paths instead.
 
 Other Changes
 ^^^^^^^^^^^^^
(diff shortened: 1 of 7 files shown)


docs/public/dev-manual/api/api_changelog.rst
```diff
@@ -8,7 +8,7 @@ API Changelog
 
 Breaking Changes
 ^^^^^^^^^^^^^^^^
-
+- ``@solrsearch:``: Change ``relative_path`` filter query to no longer expect physical paths but relative paths instead.
```
njohner, line 11: Change ``relative_path`` -> Change ``path_parent``?

njohner (approved):
LGTM

PR #7342, opened 2022-01-25
Use static fixtures for workspace fixtures

elioschmutz:
This PR uses static uids for the workspace testing fixtures. It seems that we've forgotten it while adding the fixtures.

We need this change to make frontend testing fixtures static.


njohner (approved):
LGTM

PR #7344, opened 2022-01-26
Actor portrait_url from different sources

elioschmutz:
This PR implements an  `IActorSettings` registry interface with a `user_avatar_image_source_type` property.

This allows us to configure the portrait_url lookup between the three modes:

- plone only
- portal only
- plone with fallback to portal

The `@actors` endpoint respects this setting and returns the expected portrait url.

For [CA-1938]

## Checklist

- [x] Changelog entry

- API change:
  - If breaking:
    - [ ] api-change label added
- New functionality:
- DB-Schema migration
- New translations

[CA-1938]: https://4teamwork.atlassian.net/browse/CA-1938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-1938.feature
+++ b/changes/CA-1938.feature
@@ -1,0 +1,1 @@
+Actor portrait_url respects the IActorSettings to choose the portrait_url from plone or the portal. [elioschmutz]
--- a/opengever/base/interfaces.py
+++ b/opengever/base/interfaces.py
@@ -490,3 +490,25 @@ class IWhiteLabelingSettings(Interface):
 
     logo_src = schema.Bytes(title=u'Logo image',
                             description=u'Format must be png, height must be 30px')
+
+
+AVATAR_SOURCE_PLONE_ONLY = 'plone_only'
+AVATAR_SOURCE_PORTAL_ONLY = 'portal_only'
+AVATAR_SOURCE_AUTO = 'auto'
+
+
+def get_user_avatar_image_sources():
+    return SimpleVocabulary.fromValues([
+        AVATAR_SOURCE_PLONE_ONLY,
+        AVATAR_SOURCE_PORTAL_ONLY,
+        AVATAR_SOURCE_AUTO,
+        ])
+
+
+class IActorSettings(Interface):
+
+    user_avatar_image_source = schema.Choice(
+        title=u"User avatar image source",
+        source=get_user_avatar_image_sources(),
+        default=AVATAR_SOURCE_PLONE_ONLY,
+    )
--- a/opengever/base/tests/test_casauth.py
+++ b/opengever/base/tests/test_casauth.py
@@ -3,6 +3,7 @@ from ftw.builder import create
 from opengever.base.casauth import get_cas_server_url
 from opengever.base.casauth import get_cluster_base_url
 from opengever.base.casauth import get_gever_portal_url
+from opengever.base.casauth import get_user_avatar_url
 from opengever.base.interfaces import IPortalSettings
 from opengever.ogds.base.utils import get_current_admin_unit
 from opengever.testing import IntegrationTestCase
@@ -87,3 +88,13 @@ class TestGEVERPortalURL(IntegrationTestCase):
 
         self.assertEquals(
             'http://example.com/my-special-portal', get_gever_portal_url())
+
+
+class TestUserAvatarUrl(IntegrationTestCase):
+
+    def test_user_avatar_url_contains_the_md5_encoded_username(self):
+        get_current_admin_unit().public_url = 'http://example.com'
+        get_current_admin_unit().unit_id = 'foobar'
+        self.assertEquals(
+            'http://example.com/portal/media/avatars/04f98100995b2f5633210e10f21ee022',
+            get_user_avatar_url('foo.bar'))
--- a/opengever/api/tests/test_actors.py
+++ b/opengever/api/tests/test_actors.py
@@ -2,10 +2,17 @@
 from datetime import datetime
 from ftw.testbrowser import browsing
 from ftw.testing import freeze
+from OFS.Image import Image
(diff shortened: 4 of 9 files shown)


opengever/base/tests/test_casauth.py
```diff
@@ -87,3 +88,12 @@ def test_use_the_portal_url_in_the_plone_registry_if_set(self):
...
         self.assertEquals(
             'http://example.com/my-special-portal', get_gever_portal_url())
+
+
+class TestUserAvatarUrl(IntegrationTestCase):
+
+    def test_user_avatar_url_contains_the_md5_encoded_username(self):
+        get_current_admin_unit().public_url = 'http://example.com/foobar'
+        self.assertEquals(
+            'http://example.com/foobar/portal/media/avatars/04f98100995b2f5633210e10f21ee022',
```
lukasgraf, line 98: This doesn't look right. If `http://example.com/foobar` is the URL of one admin unit (and `http://example.com/qux` one of another), the portal would be at `http://example.com/portal`, one the same level as the admin units.

Like for `dev.onegovgever.ch/fd` and `dev.onegovgever.ch/fd` for example.

I suspect the problem doesn't lie with `get_user_avatar_url()` though, but with `get_gever_portal_url()` instead. Which to this day I don't fully understand. Why exactly do we need the portal URL [configured redundantly](https://github.com/4teamwork/opengever.core/blob/33328cef67c9ffb1ca12232c58397f989ac6c169/opengever/base/casauth.py#L47) in `IPortalSettings` (instead of just deriving it from [`get_cas_server_url()`](https://github.com/4teamwork/opengever.core/blob/33328cef67c9ffb1ca12232c58397f989ac6c169/opengever/base/casauth.py#L53-L61)?)

opengever/base/interfaces.py
```diff
@@ -490,3 +490,27 @@ class IWhiteLabelingSettings(Interface):
...
+        AUTO_USER_AVATAR_IMAGE_SOURCE_TYPE,
+        ])
+
+
+class IActorSettings(Interface):
+
+    user_avatar_image_source_type = schema.Choice(
+        title=u"User avatar image source type",
+        source=get_user_avatar_image_source_types(),
+        default=DEFAULT_USER_AVATAR_IMAGE_SOURCE_TYPE,
```
lukasgraf, line 515: Just set this to `PLONE_ONLY_...` here instead of introducing a new constant and indirection?

opengever/base/interfaces.py
```diff
@@ -490,3 +490,27 @@ class IWhiteLabelingSettings(Interface):
 
     logo_src = schema.Bytes(title=u'Logo image',
                             description=u'Format must be png, height must be 30px')
+
+
+PLONE_ONLY_USER_AVATAR_IMAGE_SOURCE_TYPE = 'plone_only'
```
lukasgraf, line 495: Can you remove the `_type` suffix from all of these (from the registry setting name too, and the vocab below, etc..). And just call them `..._source` or `.._sources` respectively. I think it's more than clear without it, and names get really long otherwise 😜

changes/CA-1938.feature
```diff
@@ -0,0 +1 @@
+Avatar portrait_url respects the IActorSettings to choose the portrait_url from plone or the portal. [elioschmutz]
```
lukasgraf, line 1: `Avatar portrait_url` > `Actor portrait_url`

lukasgraf (approved):
Some minor things, but otherwise LGTM 👍 

But I think this would be a good chance to at least look into whether we really need the `IPortalSettings.portal_url` (see my comment), and possibly create a follow up story to remove it if it's not needed.

opengever/base/interfaces.py
elioschmutz: 👍

opengever/base/tests/test_casauth.py
```diff
@@ -87,3 +88,12 @@ def test_use_the_portal_url_in_the_plone_registry_if_set(self):
 
         self.assertEquals(
             'http://example.com/my-special-portal', get_gever_portal_url())
+
+
+class TestUserAvatarUrl(IntegrationTestCase):
+
+    def test_user_avatar_url_contains_the_md5_encoded_username(self):
+        get_current_admin_unit().public_url = 'http://example.com/foobar'
+        self.assertEquals(
```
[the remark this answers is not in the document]
elioschmutz: what's with local consistency?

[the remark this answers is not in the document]
buchi: For me local consistency is less important in this case, but it's nothing important. If you like it more this way, then just keep it.

opengever/base/tests/test_casauth.py
elioschmutz: The reason why `foobar` is included in the url is because the `unit_id` is missing on the admin-unit configuration in the test. I added it to make it clear

elioschmutz: @lukasgraf in summary: we do not really use the `IPortalSettings` in our customer deployments. We use it i.e. in our own deployments like the demo-deployment. It then points to the 4tw cas auth server on a different domain.

We could remove this setting and derive it from the `get_cas_server_url()` as you suggested already. But I think that's not correct, even if it probably works for the next 10 years 😄 

The `get_gever_portal_url()` has a fallback to `base-url-without-admin-unit/portal` which is our default configuration. If we set the `portal_url` in the setting, it will use this instead. This method gives us the URL to the portal, so the Ianus. But the CAS could be anything else. So using the `get_cas_server_url()` would just work because it also points to the portal, but it could also point to any other CAS.

I suggest to just leave it as it is. What do you think?

PR #7347, opened 2022-01-28
Add a url_parent/@id_parent filter for the @solrsearch endpoint

elioschmutz:
Implements new filter-queries for the `@solrsearch` endpoint according to https://github.com/4teamwork/opengever.core/pull/7299/commits/a65022472c3f7996bf8eed18c3cfb8a13d8fa7e1

It implements these two new filter queries:

- `url_parent`
- `@id_parent`

The filter-queries works like the `path_parent` filter but can be used with an url instead a relative path.

For [CA-3583]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-3583]: https://4teamwork.atlassian.net/browse/CA-3583?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-3583.feature
+++ b/changes/CA-3583.feature
@@ -1,0 +1,1 @@
+The @solrsearch results can now be filtered by ``@id_parent`` or ``url_parent``. [elioschmutz]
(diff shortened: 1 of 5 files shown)


njohner (approved):
LGTM

PR #7350, opened 2022-02-01
Avoid unnecessary security updates

buchi:
If a new object is added or an existing one is renamed, there's no need to update the security or perform any of the other index updates in this event handler. Further if an object is copied, the security is now updated only once. It was updated twice before, because the object is renamed after beeing added to the container.

This speeds up object creation and copying of objects significantly. E.g. bundle import time is reduced by approximately 30%.

For [CA-3625](https://4teamwork.atlassian.net/browse/CA-3625)

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-3625.other
+++ b/changes/CA-3625.other
@@ -1,0 +1,1 @@
+Avoid unnecessary security updates when adding new content or renaming existing content. [buchi]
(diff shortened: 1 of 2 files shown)


njohner (approved):
LGTM! Would be nice to ship that in SG before the repository migration of BDGS...

PR #7351, opened 2022-02-02
Replace dossier comments field with IResponseSupported behavior

tinagerber:
For dossiers, the comment field is removed and the IResponseSupported behavior is used instead. 

When creating a dossier from a template, a response is created if the comment field of the template contains anything.

I have completely removed the comment field in the classic UI, the comment feature is only available in the new UI.

For [CA-3327]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-3327]: https://4teamwork.atlassian.net/browse/CA-3327?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/core/upgrades/20220127165920_migrate_dossier_comments/upgrade.py
```diff
@@ -0,0 +1,16 @@
...
+
+class MigrateDossierComments(UpgradeStep):
+    """Migrate dossier comments.
+    """
+
+    deferrable = True
+
+    def __call__(self):
+        query = {'object_provides': IDossierMarker.__identifier__}
+        with DossierCommentsMigrator() as indexer:
```
njohner, line 15: `indexer` -> `comment_migrator` or something?

opengever/api/templatefolder.py
```diff
@@ -135,6 +139,12 @@ def validate_keywords(self, dossier_template, keywords_data):
                 if keyword not in allowed_keywords:
                     raise BadRequest("Keyword '{}' is not allowed".format(keyword))
 
+    def add_comment(self):
+        if IDossierTemplate(self.dossier_template).comments:
```
njohner, line 143: So on dossier templates we keep the `comments` field as it was before? And then it will have as creator the user creating the dossier? That might be a bit confusing. @phgross any thoughts?

njohner (commented):
A few points I'm unsure about:
- Does it make sense to have the `comments` field on the dossier templates?
- Are the responses displayed in the classic UI? Can users add responses?
- Did you check whether removing the field will break upgrade steps? 

Otherwise LGTM

opengever/api/templatefolder.py
tinagerber: That's what I discussed with Erich. 
I'm not quite sure about the creator either, I'll clarify this with Erich.

tinagerber: Erich thinks the dossier creator should be set.

njohner: Ok. I think I'd have gone for removing the comment field on the dossier templates. But fine with me

tinagerber:
@njohner 
- Comments are no longer supported in the classic UI, I have discussed this with PO's
- The only upgrade step that uses the comments field only accesses the annotations, so should not be a problem

Can you have another look?

njohner (approved):
LGTM

PR #7353, opened 2022-02-02
Fix propertysheets `date` field serialization and default

lukasgraf:
This fixes several issues found when adding more tests for the propertysheets `date` field type:

#### `CustomPropertiesStorageImpl`: Don't convert to json_compatible in `__getattr__`:

No longer pass returned properties through `json_compatible()` when reading them from annotations storage, e.g. with a read access to `IDocumentCustomProperties(doc).custom_properties`.

This was originally done to make the task transporter happy, but that doesn't really work, and caused issues with date fields:

- It didn't _really_ work for the transporter, because `json_compatible()` may be a no-op for Python fields that have a direct equivalent in JSON, but actually performs a serialization for types like dates that don't. Those would then need to be deserialized on the other side. But since `json_compatible()` is a one way "serialization" only, a proper serialization/deserialization would have to be done in the transporter (but won't, see below).

 - It caused issues, because for one, setting data again that has been read from `__getattr__` would wreck any non-JSON native types like dates.

It also broke the assumption that on this abstraction layer we're deadling with the Python types implied by the respective `zope.schema` fields.

#### Propertysheets widget: Don't pass widget value through `json_compatible()`:

This may have seemed to work for field types that have a direct JSON equivalent, but in the case of a date field, this will turn a Python native `date` into a string, and lead to the widget failing in display mode when it tries to parse that "date". It's also not necessary, and most likely a relict from the time where the implementation was based on a JSONField.

#### Don't attempt to transport custom properties across admin units:

Instead, skip custom properties and just make sure they don't break the transporter.

But they will not be transported, because we can't know what custom propertysheets exist on the other side. Fields might be missing, or even exist with a different type.

Also, since `document_types` and `dossier_types` are customizable, the other side might not even have the same assignment slots available.

I discussed this with @phgross, and we both came to the same conclusion - it just doesn't make sense to transport custom properties accross admin units. They are very domain-specific and local _by definition_, and any workaround to transport them would likely to be wrong. If we were to do this, we would need a very clearly defined use case first.

For [CA-3506](https://4teamwork.atlassian.net/browse/CA-3506)

## Checklist

- [x] Changelog entry

### DIFF
--- a/opengever/base/utils.py
+++ b/opengever/base/utils.py
@@ -179,6 +179,26 @@ def make_persistent(data):
         return data
 
 
+def make_nonpersistent(data):
+    """Recursively turn a nested data structure of persistent lists and dicts
+    into one using plain Python lists and dicts.
+    """
+    if isinstance(data, PersistentDict):
+        new = dict()
+        for key, value in data.items():
+            new[make_nonpersistent(key)] = make_nonpersistent(value)
+        return new
+
+    elif isinstance(data, PersistentList):
+        new = list()
+        for value in data:
+            new.append(make_nonpersistent(value))
+        return new
+
+    else:
+        return data
+
+
 def rewrite_path_list_to_absolute_paths(request):
     """If request contains a paths:list param, rewrite these paths so
     they're always absolute (start with the Plone site).
--- a/changes/CA-3506-1.bugfix
+++ b/changes/CA-3506-1.bugfix
@@ -1,0 +1,1 @@
+Custom properties: Don't prematurely make returned values json_compatible(). [lgraf]
--- a/changes/CA-3506-2.bugfix
+++ b/changes/CA-3506-2.bugfix
@@ -1,0 +1,1 @@
+Custom properties: Fix rendering z3c.form widgets in display mode for date fields. [lgraf]
--- a/changes/CA-3506-3.bugfix
+++ b/changes/CA-3506-3.bugfix
@@ -1,0 +1,1 @@
+Fix setting static defaults for propertysheet date fields. [lgraf]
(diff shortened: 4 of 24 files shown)


opengever/base/utils.py
```diff
@@ -179,6 +179,26 @@ def make_persistent(data):
         return data
 
 
+def make_nonpersistent(data):
+    """Recursively turn a nested data structure of persistent lists and dicts
+    into one using plain Python lists and dicts.
+    """
```
deiferni, line 185: Should we also handle any of the `BTree` data structures? Or use a duck-typing approach to figure out which collection type the object is? E.g. `hasattr(data, 'items')` to detect a mapping of any kind instead of explicit type checks.

My gut feeling would prefer the duck typing approach. Als an alternative you could use a logging/throwing safeguard if we receive any unhandled subclass of `Persistent`.

deiferni (approved):
LGTM 👍 .

One comment with some thoughts for your consideration. Feel free to :shipit:  as is though.

✨ 👍  for not transporting properties.

opengever/base/utils.py
lukasgraf: YAGNI, I think.

We don't have any BTrees in custom properties right now, so any code that handles BTrees specifically would literally be dead code. And even with duck typing, it just wouldn't be used (in the generic sense that is intended) for now, so we wouldn't know whether it would actually work properly with BTrees. Unless I'd write a unit test for it that *does* exercise the code with BTrees, even though that's not something used by any code as of now.

The docstring says what it does, and it does exactly that (the exact reverse of the `make_persistent()` above, BTW).

PR #7354, opened 2022-02-03
Propertysheets: Make sure preprocessing of static defaults doesn't choke on non-string types.

lukasgraf:
Because the new frontend apprently omits empty fields on **add**, but sends them with `null` on **edit**, the static `default` for a propertysheet field may receive `None`, instead of a string.

`strptime` then raises a `TypeError` instead of a `ValueError` which also needs to be handled.

For [CA-3506](https://4teamwork.atlassian.net/browse/CA-3506)

## Checklist

- [ ] Changelog entry _(deemed unnecessary, because the bug was introduced and fixed in the same release)_


deiferni (approved):
👀  :shipit:

PR #7355, opened 2022-02-03
Suppress creation of todo completion activities during upgrades:

lukasgraf:
We suppress creation of activities in the `todo_review_state_changed()` handler if it gets called during an upgrade to avoid failing upgrade steps.

This is because the creation of activities will emit SQL statements (queries, as well as inserts) which are likely to fail:

They will use the current models SQL (for Activities, Notifications, Users, ...) as present in the already updated code, but may run against an SQL database that hasn't had its schema fully upgraded yet.

(There may be later upgrade steps pending that would upgrade it and bring it in line with the schema on the Python models).

This has specifically been a problem with the upgrade [`20211105114219_add_completed_state_for_todos`](https://github.com/4teamwork/opengever.core/blob/master/opengever/core/upgrades/20211105114219_add_completed_state_for_todos/upgrade.py), which triggers state transitions for todos, and then fails e.g. because the `absent` field on the `User` model, or `external_resource_url` field on the `Activitiy` model isn't there yet.

Additionally, it's probably undesirable anyway to dispatch notifications to users during our upgrades.

With this change, I was able to sucessfully run through upgrades from `2021.12.1` to current master (2021.12.1 is the oldest real OnPremise deployment around).

For [CA-3574](https://4teamwork.atlassian.net/browse/CA-3574)

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-3574.bugfix
+++ b/changes/CA-3574.bugfix
@@ -1,0 +1,1 @@
+Suppress creation of todo completion activities during upgrades to avoid failing upgrades. [lgraf]
--- a/opengever/workspace/subscribers.py
+++ b/opengever/workspace/subscribers.py
@@ -1,3 +1,4 @@
+from ftw.upgrade.interfaces import IDuringUpgrade
 from opengever.base.placeful_workflow import assign_placeful_workflow
 from opengever.base.response import COMMENT_RESPONSE_TYPE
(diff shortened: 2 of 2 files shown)


njohner (approved):
LGTM!

PR #7356, opened 2022-02-03
Fix content-sharing for multiple recipients.

phgross:
With https://github.com/4teamwork/opengever.core/pull/7270 I broke the content-sharing for multiple recipients. Because we pass in now the `mail_to` manually we need to make sure its a list and not a comma separated string.
 
For [CA-3631]

## Checklist
- [x] Changelog entry

[CA-3631]: https://4teamwork.atlassian.net/browse/CA-3631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/activity/mailer.py
+++ b/opengever/activity/mailer.py
@@ -86,7 +86,7 @@ class Mailer(object):
         We therefore defer the dispatching of notification mails until the
         very end of the transaction to work around this issue.
         """
-        mail_queue.put(msg, mail_to, mail_from)
+        mail_queue.put(msg, [each.strip() for each in mail_to.split(',')], mail_from)
 
     def prepare_mail(self, subject=u'', to_userid=None, to_email=None,
                      cc_email=None, from_userid=None, data=None):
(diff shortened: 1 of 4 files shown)


opengever/activity/mailer.py
```diff
@@ -86,7 +86,7 @@ def send_mail(self, msg, mail_to, mail_from):
         We therefore defer the dispatching of notification mails until the
         very end of the transaction to work around this issue.
         """
-        mail_queue.put(msg, mail_to, mail_from)
+        mail_queue.put(msg, mail_to.split(', '), mail_from)
```
njohner, line 89: Are we sure here that it's always nicely separated by coma and whitespace? or should we maybe do something like `[each.strip() for each in mail_to.split(',')]`?

njohner (approved):
LGTM

opengever/activity/mailer.py
phgross: At the moment yes, but we can not guarantee it. I'll change it.

Issue #2033, opened 2016-07-19
Potential for infinite recursion in acquire_field_value() from restricted vocabularies


lukasgraf:
When acquiring field values from parents, the code for restricted vocabularies attempts something along the lines of

```
field.get(field.interface(obj))
```

at some point ([current code](https://github.com/4teamwork/opengever.core/blob/c1b24e294cadbaae4b515277dd0d1d1677e6ae2c/opengever/base/behaviors/utils.py#L137) | [code currently being refactored](https://github.com/4teamwork/opengever.core/blob/6213f9aa1a7e007f268e9f646198a2cc8128bd66/opengever/base/acquisition.py#L27)).

If no value for that field has been persisted on the object, and AttributeStorage is being used, the [`DexterityContent.__getattr__` fallback](https://github.com/plone/plone.dexterity/blob/2.1.2/plone/dexterity/content.py#L193-L205) will kick in, and fall back to `field.default`.

Once the `IClassification`/ `ILifeCycle` field default values have been rewritten to context aware `defaultFactory`'s, those **will automatically get called** by accessing `field.default` (thanks to `zope.schema`s [`DefaultProperty` descriptor](https://github.com/zopefoundation/zope.schema/blob/cd3d77c98455a1f872e18c28730dc5c77719dff4/src/zope/schema/_bootstrapfields.py#L70-L73)).

That means our function for acquiring a default value will be invoked recursively. Whether or not it will be infinite recursion, I can't say for sure yet. But we'll probably notice quite quickly.

A potential fix is to use [`get_persisted_value_for_field()`](https://github.com/4teamwork/opengever.core/blob/6213f9aa1a7e007f268e9f646198a2cc8128bd66/opengever/base/default_values.py#L99-L142) instead, making sure we avoid triggering the fallback to `field.default` - but that will only work correctly once non-persisted default values have been migrated / persisted.

PR #7358, opened 2022-02-07
Remove propagation and restriction of ILifecycle and IClassification fields.

njohner:
When modifying a field with a restricted vocabulary on an object we do not propagate that change to children anymore. Until now we were propagating to children but only within a depth of 2. This does not make much sense, but lifting that restriction would lead to serious performance issues. Also it might not always be what the user intends as the user might also only want to change the allowed values for future content. We therefore decided not to propagate such changes anymore.

Of course if we do not propagate changes in the values anymore, we also need to lift the restrictions on the children, otherwise we would end up with objects having invalid values for the corresponding fields. We therefore remove the restrictions of allowed values for the `classification`, `privacy_layer` and `retention_period` fields.

I've also checked all the client policies for the registry entry `IRetentionPeriodRegister.is_restricted` that is being removed in this PR. It was only used in SG, where I have cleaned it up: https://github.com/4teamwork/opengever.sg/pull/522

Note that this also led to an infinite recursion appearing in one of the tests, which was foreseen as a possibility in https://github.com/4teamwork/opengever.core/issues/2033. I've fixed it now, as suggested in that issue.

For [CA-3241]

## Checklist

- [x] Changelog entry

[CA-3241]: https://4teamwork.atlassian.net/browse/CA-3241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/ech0147/tests/test_import.py
```diff
@@ -188,16 +190,15 @@ def test_import_with_invalid_field_data(self, browser):
...
                         .having(title_de=u'Vertr\xe4ge und Vereinbarungen',
                                 title_fr=u'Contrats et accords',
-                                title_en=u'Contracts and agreements',
-                                classification='classified')
+                                title_en=u'Contracts and agreements')
                         .within(root))
 
+        # we modify the mapping used so that an invalid value is set for
+        # the classification field on the object being created.
+        INV_CLASSIFICATION_MAPPING["confidential"] = "invalid"
```
lukasgraf, line 201: Doesn't this mutation to the `INV_CLASSIFICATION_MAPPING` possibly affect other tests? Maybe back it up and restore it to its original contents afterwards with `INV_CLASSIFICATION_MAPPING.update(backup)`?

lukasgraf (approved):
LGTM 👍 🎉  🧹🧹🧹 
Just one minor point.

opengever/ech0147/tests/test_import.py
njohner: Yeah I actually wasn't too sure either, but thought that isolation took care of that for me 😅 . I'll do as you suggest.

PR #7360, opened 2022-02-08
Add new @config property: has_dossier_propertysheet_registered

elioschmutz:
This PR adds a new `@config` property: `has_dossier_propertysheet_registered`.

This is true if the deployment has one or more propertysheet definitions for dossiers

We use this in the frontend to show a card for custom properties for dossiers.

For [CA-3325]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-3325]: https://4teamwork.atlassian.net/browse/CA-3325?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/api/tests/test_config.py
+++ b/opengever/api/tests/test_config.py
@@ -5,6 +5,7 @@ from ftw.testbrowser import browsing
 from mock import patch
 from opengever.base.interfaces import IUserSnapSettings
 from opengever.private import enable_opengever_private
+from opengever.propertysheets.storage import PropertySheetSchemaStorage
 from opengever.testing import IntegrationTestCase
 from opengever.testing.readonly import ZODBStorageInReadonlyMode
 from pkg_resources import get_distribution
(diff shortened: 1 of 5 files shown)


opengever/api/tests/test_config.py
```diff
@@ -268,6 +269,26 @@ def test_is_propertysheets_manager_is_false_for_regular_user(self, browser):
...
+               .named("businesscase_dossier_schema")
+               .assigned_to_slots(u"IDossier.dossier_type.businesscase"))
+
+        self.login(self.regular_user, browser=browser)
+        browser.open(self.config_url, headers=self.api_headers)
+        self.assertEqual(browser.status_code, 200)
+        self.assertTrue(browser.json.get(u'has_propertysheet_definitions'))
+
+    @browsing
+    def test_has_propertysheet_definitions_is_true_if_the_deployment_does_not_defines_property_sheets(self, browser):
```
njohner, line 285: is_true -> is_false

njohner (approved):
Just a copy/paste fail in your test name, otherwise LGTM

elioschmutz:
> The question is how useful this is, since the point is to display a dossier card or not. If you have only document custom properties, the card will be displayed anyway. `has_dossier_propertysheet_definitions` would be more useful

@tinagerber that's a good point 👍  I'll update the PR

elioschmutz:
I updated the PR according to @tinagerber suggestion

elioschmutz:
@njohner could you take another look since I updated the PR with the new implementation?

njohner (approved):
Well still LGTM. But to follow up on Tina's comment, I was actually wondering why we needed this at all, as you probably get the propertysheet when getting the context in the UI, so you can decide based on that whether to display the card or not no? Because just looking at the flag will still lead you to display an empty card most of the time, because there might be a property sheet only for certain dossier types. I hadn't commented, because I thought I let it up to the frontenders to decide whether they need that config entry or not...

elioschmutz:
@njohner yes, you're completely right! Thanks!

@all4manu the `custom_properties` is filled if the obj has custom properties. This information should be enough (or even more helpful than this @config-property) to decide wether we need to render the card or not. The case where a deleted field gets still serialized can be ignored IMO.

Bildschirmfoto 2022-02-09 um 11 36 24

So I'll close this PR again because it'll just blow up the `@config`

PR #7363, opened 2022-02-14
Do not set security header for responses without content type

buchi:
For responses without a body it doesn't make sense to set the securtiy headers. Also fixes setting the wrong CSP for 304 responses, which breaks the classic frontend if HTTP caching is enabled.

Fix for: https://github.com/4teamwork/opengever.core/pull/6894

## Checklist

- [ ] Changelog entry


lukasgraf (approved):
LGTM 👍

PR #7364, opened 2022-02-15
Mention setuptools/buildout to use with bootstrap.

deiferni:
Currently we need to use very specific versions to be able to bootstrap a local GEVER. This should be mentioned in the readme


njohner (approved):
Merci!

PR #7365, opened 2022-02-16
Reindex 'filename' when creating a new document version:

lukasgraf:
When a new version is saved, the file's type may have changed (e.g. from .pdf to .docx), and the filename therefore is updated, and needs to be reindexed.

In particular this was the case in a situation where a PDF was reverted (which creates a new version) to a version that was a Word-Document. In this case the filename in Solr was outdated, because unlike the catalog, metadata in Solr doesn't get updated unless the index is explicitly listed.

❗ FYI: I decided to not add an upgrade step for this issue. I couldn't easily find a way to recognize just the documents that exhibit this specific problem, which would mean reindexing `filename` for all documents. And that is just not justified IMHO for a problem this rare and minor. Also, the situation is self-healing, in the sense that the `filename` in Solr will eventually get updated the next time the document's metadata is updated.

For [CA-3677](https://4teamwork.atlassian.net/browse/CA-3677)

## Checklist

- [x] Changelog entry

### DIFF
--- a/opengever/document/versioner.py
+++ b/opengever/document/versioner.py
@@ -35,7 +35,14 @@ class Versioner(object):
         """Creates a new version in CMFEditions.
         """
         self.repository.save(obj=self.document, comment=comment)
-        self.document.reindexObject(idxs=['UID', 'approval_state'])
+        self.document.reindexObject(
+            idxs=[
+                'UID',
+                'approval_state',
+                'filename',
+                'file_extension',
+            ]
+        )
 
     def has_initial_version(self):
         return self.get_history_metadata() != []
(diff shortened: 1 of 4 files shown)


opengever/document/versioner.py
```diff
@@ -35,7 +35,7 @@ def create_version(self, comment):
         """Creates a new version in CMFEditions.
         """
         self.repository.save(obj=self.document, comment=comment)
-        self.document.reindexObject(idxs=['UID', 'approval_state'])
+        self.document.reindexObject(idxs=['UID', 'approval_state', 'filename'])
```
njohner, line 38: I think you also need `file_extension` in this list of indexes.

PR #7367, opened 2022-02-21
Switch `@navigation` endpoint to Solr queries.

lukasgraf:
Switch `@navigation` endpoint to Solr queries.

This is in preparation for later adding the `dossier_type`, which is only indexed in Solr. Also, building the navigation by querying Solr instead of the ZCatalog should be noticably faster.

## Checklist

- [x] Changelog entry


buchi (approved):
👍🏻 LGTM

PR #7368, opened 2022-02-23
Add `@raw-dossier-types` endpoint:

lukasgraf:
This endpoint returns the raw contents of the `dossier_types` VDEX vocabulary.

They are returned

- in **inherent order** (even though orderSignificant may be False)
- **unfiltered** (hidden terms are included)

This is needed in order for the frontend to be able to map dossier types to colors in a deterministic way that is stable over time.

Because dossier types are mapped to a fixed-length color palette, and types that have already been mapped to a color once MUST keep that color, we need to provide the frontend with a list of dossier types where historical terms keep their position, and new terms only ever are appended at the end.

This endpoint facilitates this, together with the following rules for editing the `dossier_types.vdex` file:

- New terms must only be **appended at the end**
- Terms are **never removed**, but set to hidden instead (registry).
  They will then be omitted from the regular vocabularies / sources, which are elephant vocabs, but not this endpoint.
- Renames are ok (if the type is supposed to keep its color)

---

**Note**: _This is implemented as a separate endpoint, because if we were to just change `orderSignificant=True` in the VDEX file, the actual vocabularies (as returned by `@sources` for example) would then return terms in their inherent order, instead of sorting them alphabetically (which is not what we want for widgets etc..)._ 

## Checklist

- [x] Changelog entry

- API change:
  - [x] API Changelog entry

### DIFF
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -12,6 +12,8 @@ Breaking Changes
 Other Changes
 ^^^^^^^^^^^^^
 
+- ``@raw-dossier-types``: New endpoint that returns dossier types in unaltered order.
+
 
 2022.4.0 (2022-02-16)
 ---------------------
--- a/opengever/api/raw_dossier_types.py
+++ b/opengever/api/raw_dossier_types.py
@@ -1,0 +1,63 @@
+from collective.vdexvocabulary.term import VdexTerm
+from plone import api
+from plone.restapi.interfaces import ISerializeToJson
+from plone.restapi.services import Service
+from zope.component import getMultiAdapter
+from zope.component import getUtility
+from zope.schema.interfaces import IVocabularyFactory
+from zope.schema.vocabulary import SimpleVocabulary
+
+
+class RawDossierTypesGet(Service):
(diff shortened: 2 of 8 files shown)


docs/public/dev-manual/api/api_changelog.rst
```diff
@@ -12,6 +12,8 @@ Breaking Changes
 Other Changes
 ^^^^^^^^^^^^^
 
+- ``@raw-dossier-types``: New endpoint that returns dossier types in inherent order.
```
njohner, line 15: "unaltered" instead of "inherent"?

opengever/api/raw_dossier_types.py
```diff
@@ -0,0 +1,63 @@
...
+from zope.component import getUtility
+from zope.schema.interfaces import IVocabularyFactory
+from zope.schema.vocabulary import SimpleVocabulary
+
+
+class RawDossierTypesGet(Service):
+    """API Endpoint that returns the raw contents of the dossier_types VDEX
+    vocabulary. They are returned
+
+    - in inherent order (even though orderSignificant may be False)
```
njohner, line 15: here too maybe "unaltered" would be better?

njohner (approved):
LGTM

PR #7369, opened 2022-02-23
Allow admins to delete deactivated workspaces

tinagerber:
Workspaces can only be deleted if they are not linked to a dossier. Therefore, two additional changes were necessary: 
- It must be possible to unlink a workspace when the workspace is deactivated
- It must be possible to unlink a workspace when a dossier is closed

For [CA-1967]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations


opengever/api/linked_workspaces.py
```diff
@@ -128,6 +128,16 @@ def reply(self):
         return self.reply_no_content()
 
 
+class RemoveDossierReferencePost(Service):
+
+    def reply(self):
+        # Disable CSRF protection
+        alsoProvides(self.request, IDisableCSRFProtection)
+        self.context.external_reference = u''
```
njohner, line 131: I would at least check that the user executing the action is `WorkspaceMember` or `WorkspaceAdmin` no? A `WorkspaceGuest` should not be allowed to do that.
But basically I'd say we do not want to document that endpoint, because we only want this endpoint to be used from within Gever when unlinking a workspace.

Also we could add a soft-check that this comes from Gever by checking the `User-Agent` header (check that it contains `OPENGEVER_APICLIENT_USER_AGENT`).

I quickly discussed that with @lukasgraf

changes/CA-1967-1.bugfix
```diff
@@ -1 +1 @@
-Allow unlinking workspaces even if the workspace is deactivated. [tinagerber]
+Allow unlinking workspaces even if the workspace is deactivated or dossier is closed. [tinagerber]
```
njohner, line 1: commit title has a typo `workdpaces`

docs/public/dev-manual/api/api_changelog.rst
```diff
@@ -12,6 +12,7 @@ Breaking Changes
 Other Changes
 ^^^^^^^^^^^^^
 - Add new endpoint ``@remove-dossier-reference`` (see :ref:`docs <remove-dossier-reference>`)
+- ``@unlink-workspace``: Allow unlinking workdpaces even if the dossier is closed.
```
njohner, line 15: here is the typo as well

opengever/base/browser/folder_buttons_availability.py
```diff
@@ -116,7 +116,6 @@ def is_link_to_workspace_available(self):
 
     def is_unlink_workspace_available(self):
         return (self._is_main_dossier()
-                and self._can_modify_dossier()
```
njohner, line 119: This means the action will be available for users who can only view the dossier but are not allowed to modify it (even if the dossier is open). And they can probably also execute the action no? That seems wrong to me

opengever/api/tests/test_delete.py
```diff
@@ -224,3 +238,30 @@ def test_guests_cannot_permanently_delete_document(self, browser):
...
+        self.workspace.external_reference = u'a dossier UID'
+        api.content.transition(
+            self.workspace,
+            'opengever_workspace--TRANSITION--deactivate--active_inactive')
+        self.assert_cannot_delete(self.workspace, browser, code=403)
+
+    @browsing
+    def test_workspace_admin_cannot_permanently_delete_workspace(self, browser):
+        self.login(self.workspace_admin, browser)
+        self.assert_cannot_delete(self.workspace, browser, code=403)
```
njohner, line 267: Shouldn't you check with an inactive workspace here?

opengever/workspace/delete.py
```diff
@@ -94,3 +95,14 @@ def verify_may_delete(self, main=True):
...
+
+@adapter(IWorkspace)
+class WorkspaceDeleter(BaseWorkspaceContentDeleter):
+
+    permission = 'opengever.workspace: Delete Workspace'
+
+    def verify_may_delete(self, main=True):
+        self.check_delete_permission()
+        if self.context.external_reference:
+            raise Forbidden
```
njohner, line 108: `raise Forbidden()`

njohner (commented):
Overall LGTM, except for comments. we can quickly discuss that tomorrow if you want!

opengever/api/tests/test_delete.py
tinagerber: No, the admin cannot delete active workspaces. I have adjusted the title of the test.

opengever/api/linked_workspaces.py
tinagerber: The endpoint is now protected with the Unlink workspace permission.

opengever/base/browser/folder_buttons_availability.py
tinagerber: It will now check for the Unlink workspace permission.

tinagerber:
@njohner I have now added the Unlink workspace permission. Can you have another look?

opengever/core/profiles/default/workflows/opengever_dossier_workflow/definition.xml
```diff
@@ -296,6 +305,14 @@
         name="opengever.meeting: Add Proposal Comment"
         acquired="False">
     </permission-map>
+    <permission-map
+        name="opengever.workspaceclient: Unlink Workspace"
+        acquired="False">
+      <permission-role>Editor</permission-role>
```
njohner, line 311: Should we also have "Contributor" here?

opengever/api/tests/test_delete.py
```diff
@@ -224,3 +238,30 @@ def test_guests_cannot_permanently_delete_document(self, browser):
...
+    @browsing
+    def test_admins_can_permanently_delete_inactive_workspaces(self, browser):
+        self.login(self.administrator, browser)
+        api.content.transition(
+            self.workspace,
+            'opengever_workspace--TRANSITION--deactivate--active_inactive')
+        self.assert_can_delete(self.workspace, browser)
+
+    @browsing
+    def test_workspace_admin_cannot_permanently_delete_active_workspace(self, browser):
```
njohner, line 256: Ok I guess here you could test with an inactive workspace no? A WorkspaceAdmin should not have the right to delete an inactive workspace or should he? Makes the test more stringent.

njohner (approved):
Just 2 small things, otherwise LGTM

opengever/core/profiles/default/workflows/opengever_dossier_workflow/definition.xml
tinagerber: Since a contributor is not allowed to edit an active dossier, I would say no.

PR #7370, opened 2022-02-23
Default value acquisition: Skip intermediate objects missing attribute.

lukasgraf:
When acquiring default values, the [`acquire_field_value()`](https://github.com/4teamwork/opengever.core/blob/cfa50aed2f6acc85fe81bf6c7e1bcd11be0d15b0/opengever/base/acquisition.py#L10-L51) function may encounter objects in the chain that do claim to provide the field, but are missing the actual attribute because it never got persisted properly.

Because we now [use `get_persisted_value_for_field()`](https://github.com/4teamwork/opengever.core/blob/cfa50aed2f6acc85fe81bf6c7e1bcd11be0d15b0/opengever/base/acquisition.py#L43) to fetch the field value, we need to account for the `AttributeError` it [may raise](https://github.com/4teamwork/opengever.core/blob/cfa50aed2f6acc85fe81bf6c7e1bcd11be0d15b0/opengever/base/default_values.py#L112) (unlike the `field.get()` that was used before).

For [CA-2878](https://4teamwork.atlassian.net/browse/CA-2878)

❗ This needs to be backported to `2022.4-stable`.

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-2878.bugfix
+++ b/changes/CA-2878.bugfix
@@ -1,0 +1,1 @@
+Default value acquisition: Skip intermediate objects missing attribute. [lgraf]
--- a/opengever/base/acquisition.py
+++ b/opengever/base/acquisition.py
@@ -40,7 +40,12 @@ def acquire_field_value(field, container):
                 # could not adapt
                 pass
             else:
-                value = get_persisted_value_for_field(obj, field)
+                try:
+                    value = get_persisted_value_for_field(obj, field)
+                except AttributeError:
+                    # has field, but missing attribute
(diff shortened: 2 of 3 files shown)


buchi (approved):
👍🏻 LGTM

lukasgraf:
Merging this even though one test check is still "pending". Task [#891418](https://ci.4teamwork.ch/builds/482875/tasks/891418) has finished with SUCCESS, but somehow CI Governor doesn't report the status back to GitHub. (And I've re-run tests twice for this PR, same issue)

PR #7371, opened 2022-02-24
NightlyWorkflowSecurityUpdater: Gracefully skip objs that can't be resolved

lukasgraf:
If an object has been deleted for example, it's intid or UID won't resolve any more, resulting in `None`, and we must not attempt to run `update_security_for(None)`, because that will result in an `AttributeError` (via `ftw.upgrade`).

This blocked nightly jobs (only on `02-digs-test`) when updating SG test. I've made these changes in the source checkout of `opengever.core` and ran nightly jobs again, and then they all ran through successfully.

Sentry: https://sentry.4teamwork.ch/organizations/sentry/issues/82311

❗ Needs backport to `2022.4-stable`.

For [CA-2878](https://4teamwork.atlassian.net/browse/CA-2878)

## Checklist

- [x] Changelog entry

- Bug fixed:

### DIFF
--- a/changes/CA-2878-1.bugfix
+++ b/changes/CA-2878-1.bugfix
@@ -1,0 +1,1 @@
+NightlyWorkflowSecurityUpdater: Gracefully skip objs that can't be resolved. [lgraf]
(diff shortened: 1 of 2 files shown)


njohner (approved):
LGTM

PR #7373, opened 2022-02-24
Include dossier_type in responses of various endpoints

lukasgraf:
This includes the `dossier_type` in responses of various endpoints:
- `@favorites` (only if the are resolved, which the frontend does in the relevant situations AFAIK)
- `@breadcrumbs`
- `@navigation`
- JSON summary representations (which leads to it getting included in `relatedDossiers`) 

This is so that the frontend can render the dossier icon in the appropriate color.

## Checklist

- [x] Changelog entry

- API change:
  - [x] API Changelog entry


opengever/base/model/favorite.py
```diff
@@ -149,6 +156,7 @@ def serialize(self, portal_url, resolve=False):
             'admin_unit': AdminUnit.query.get(self.admin_unit_id).title,
             'review_state': self.review_state,
             'is_subdossier': self.is_subdossier,
+            'dossier_type': dossier_type,
```
njohner, line 159: If I understand well `dossier_type` will be `None` if we do not resolve the favorite, but it will also be `None` if we do, but the dossier does not have a `dossier_type`. That is pretty confusing for an external application no? I wonder if the should not instead only include `dossier_type` in the response when the object is a dossier and it is resolved.

PR #7375, opened 2022-02-24
Update changelog after backports to 2022.4.1.

lukasgraf:
Update changelog after backports to `2022.4.1`.

I _think_ this should be the way to do it. I'm coping over the `2022.4.1` section from the `2022.4-stable` branch here, and deleting the two respective news fragments. But I'm not 100% how towncrier will handle this, might need double checking.

For [CA-2874](https://4teamwork.atlassian.net/browse/CA-2874)

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -10,6 +10,16 @@ Versions are of the form MAJOR.MINOR.PATCH. Each MINOR release (MAJOR.MINOR.0) i
 
 .. towncrier release notes start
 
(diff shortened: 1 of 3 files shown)


njohner:
I thought we decided simply not to handle backports in the main changelog...

lukasgraf:
You're right, I remember now. We decided to keep the changelogs in a way that they reflect each respective branch's lineage. And when creating a backport-release from a `-stable` branch, towncrier does it correctly automatically.

PR #7377, opened 2022-02-25
Fix dossier_types vocabulary for testserver

elioschmutz:
We already mocked the `dossier_types` vdex in the `testserver` to get more than one dossier type.

We mocked it by just registering a function as a vocabulary factory. The PR: https://github.com/4teamwork/opengever.core/pull/7368/ introduced a new endpoint called `@raw-dossier-types` which uses the `getTerms` function from the VDEX-Object instance. So the endpoint was broken in e2e tests because our factory function has not provided such a function:

Bildschirmfoto 2022-02-25 um 15 19 02

We now use the `MockDossierTypes` object which was also introduced in the PR #7368 . This object properly registers the vocabulary as a vdex vocabulary.

In addition, we have to use the global site manager to register the vocabulary, otherwise the testserver will raise a `TypeError: can't pickle _Element objects` on startup after committing the transaction.

Bildschirmfoto 2022-02-25 um 15 22 01

For [CA-3324]

## Checklist

- [ ] Changelog entry (⚠️ no changelog because it's a fix in this release and for testing only)

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-3324]: https://4teamwork.atlassian.net/browse/CA-3324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/core/testserver.py
+++ b/opengever/core/testserver.py
@@ -14,6 +14,7 @@ from opengever.core.testing import OpengeverFixture
 from opengever.core.testserver_zope2server import ISOLATION_READINESS
 from opengever.dossier.interfaces import IDossierType
 from opengever.testing.helpers import incrementing_intids
+from opengever.testing.helpers import MockDossierTypes
 from plone import api
 from plone.app.testing import applyProfile
 from plone.app.testing import FunctionalTesting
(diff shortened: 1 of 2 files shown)


njohner (approved):
LGTM, but is probably not necessary anymore if we revert the changes from https://github.com/4teamwork/opengever.core/pull/7368...

PR #7380, opened 2022-02-28
Provide `@reference-number` expansion and view

deiferni:
Provide an expandable view `@reference-number` that returns the formatted reference number, a sortable reference number and the unformatted reference number parts.

This helps client applications that need to use the raw reference number or parts of the raw reference.

I did not find an interface for which to register a single service. Instead i found that the `IReferenceNumber` adapter is registered for `ISiteRoot` and `IDexterityContent`. Thus the new service and adapter
mirror the adapter registrations of `IReferenceNumber`. The same service and adapter are registered twice for different interfaces.

This can be used in a client application when raw reference numbers or sortable reference numbers are needed.

For [HG-2215](https://4teamwork.atlassian.net/browse/HG-2215)

## Checklist

- [x] Changelog entry

- API change:

### DIFF
--- a/docs/public/dev-manual/api/reference_number.rst
+++ b/docs/public/dev-manual/api/reference_number.rst
@@ -1,0 +1,45 @@
+.. _reference_number:
+
+Aktenzeichen
+============
+
+Über den ``@reference-number`` Endpoint kann Details über das Aktenzeichen und seine Komponenten eines beliebigen Inhalts abgefragt werden.
+
+**Beispiel-Request**:
+
+   .. sourcecode:: http
+
+       GET /ordnungssystem/fuhrung/vertrage-und-vereinbarungen/dossier-1/document-14 HTTP/1.1
+       Accept: application/json
+
+
+**Beispiel-Response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Content-Type: application/json
+
+      {
+        "@id": "http://nohost/plone/ordnungssystem/fuhrung/vertrage-und-vereinbarungen/dossier-1/document-14/@reference-number",
+        "parts": {
+            "document": ["14"],
+            "dossier": ["1"],
+            "repository": ["1", "1"],
+            "repositoryroot": [""],
+            "site": ["Client1"],
+        },
+        "reference_number": "Client1 1.1 / 1 / 14",
+        "sortable_reference_number": "client00000001 00000001.00000001 / 00000001 / 00000014"
+      }
+
+
+Das Aktenzeichen kann beim Abfragen eines Inhaltes über den ``expand``-Parameter eingebettet werden,
+so dass keine zusätzliche Abfrage nötig ist.
+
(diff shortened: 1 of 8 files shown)


opengever/api/tests/test_reference_number.py
```diff
@@ -0,0 +1,272 @@
+from ftw.testbrowser import browsing
+from opengever.testing import IntegrationTestCase
+
+
+class TestPloneSiteExpandableReferenceNumber(IntegrationTestCase):
```
njohner, line 5: I guess you could have avoided copying the same tests 5 times by defining the object you are working on in the `setUp` method and then inheriting from this class in all the other test classes and overwriting just `setUp`

docs/public/dev-manual/api/reference_number.rst
```diff
@@ -0,0 +1,45 @@
+.. _reference_number:
+
+Aktenzeichen
+============
+
+Über den ``@reference-number`` Endpoint kann Details über das Aktenzeichen und seine Komponenten eines eines beliebigen Inhalts abgefragt werden.
```
njohner, line 6: eines eines -> eines

njohner (approved):
Apart from comments, LGTM

opengever/api/reference_number.py
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        if not expand:
+            return result
+
+        ref_num = IReferenceNumber(self.context)
+        reference = {
+            "parts": ref_num.get_numbers(),
+            "reference_number": ref_num.get_number(),
+            "sortable_reference_number": ref_num.get_sortable_number(),
```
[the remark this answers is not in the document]
deiferni: The functionality taking care of expansion will do the conversion to JSON. In here it is enough if we return something that is json serializable, which we are doing. The numbers are strings or unicode and the parts are a dict.

opengever/api/reference_number.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class ReferenceNumberGet(Service):
+    """API Endpoint to return reference number information.
+
+    GET folder-1/@reference-number HTTP/1.1
+    """
+
+    def reply(self):
+        stats = ReferenceNumber(self.context, self.request)
```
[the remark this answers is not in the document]
deiferni: indeed. thanks!

opengever/api/tests/test_reference_number.py
deiferni: Yeah i was considering it. But this has the disadvantage that in case of a failure it is really hard to figure out which test exactly is failing.
But you're right there is a lot of duplication. I will try to introduce a base-class that provides some assertion helpers.

deiferni:
@njohner thanks for the review! i have implemented the changes or commented. i'd be glad for a second glance … and also a good for :shipit: as you're about to cut a release and i don't want to interfere without 🆗  from the team responsible.

PR #7386, opened 2022-03-07
Fix removing inbox mails.

njohner:
With https://github.com/4teamwork/opengever.core/pull/7204, deletion of documents in the inbox was fixed, but it did not really fix the issue it was supposed to fix, as they could not delete E-mails (see https://4teamwork.atlassian.net/browse/CA-3021). This is now fixed here. Indeed `review_history` is accessed in the [workflow transition endpoint](https://github.com/4teamwork/opengever.core/blob/master/opengever/api/transition.py#L107), and that was not supported for inbox mails (although it is in the normal mail workflow).

Besides that, there were other issues in the inbox mail workflow and also in the private mail workflow which are fixed as well.

For [CA-3808]

## Checklist

- [x] Changelog entry
- Upgrade-Steps:
- Bug fixed:
  - [ ] Resolved any Sentry issues caused by this bug -> https://sentry.4teamwork.ch/organizations/sentry/issues/82707
- New functionality:

[CA-3808]: https://4teamwork.atlassian.net/browse/CA-3808?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lukasgraf (approved):
LGTM 👍

PR #7391, opened 2022-03-11
Add dossier checklist feature flag

tinagerber:
A feature flag is needed so that the checklist feature can be sold.

For [CA-3329]

## Checklist

- [x] Changelog entry

- API change:

[CA-3329]: https://4teamwork.atlassian.net/browse/CA-3329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/core/upgrades/20220311082313_add_dossier_checklist_feature_flag/upgrade.py
+++ b/opengever/core/upgrades/20220311082313_add_dossier_checklist_feature_flag/upgrade.py
@@ -1,0 +1,9 @@
+from ftw.upgrade import UpgradeStep
+
+
+class AddDossierChecklistFeatureFlag(UpgradeStep):
+    """Add dossier_checklist feature flag.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
--- a/opengever/dossier/__init__.py
+++ b/opengever/dossier/__init__.py
@@ -1,4 +1,11 @@
+from plone import api
 from zope.i18nmessageid import MessageFactory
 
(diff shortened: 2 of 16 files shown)


opengever/dossier/__init__.py
```diff
@@ -1,4 +1,11 @@
+from plone import api
 from zope.i18nmessageid import MessageFactory
 
 _ = MessageFactory("opengever.dossier")
 
+
+def is_dossier_checklist_feature_enabled():
+    # # Avoid circular imports
```
njohner, line 8: -> # Avoid circular imports

njohner (commented):
Hiding the checklist field in the form that way only works for the classic UI no? So the UI does the same, look at the flag to hide the field, but basically it's there and could very well be filled over the API?

opengever/core/upgrades/20220311082313_add_dossier_checklist_feature_flag/upgrade.py
```diff
@@ -0,0 +1,9 @@
+from ftw.upgrade import UpgradeStep
+
+
+class AddDossierChecklistFeatureFlag(UpgradeStep):
```
njohner, line 4: Upgrade belongs in the other commit.

njohner (approved):
Just the upgrade step that is in the wrong commit. Otherwise LGTM

PR #7392, opened 2022-03-14
Add translated titles to Solr and use them in `@navigation`

lukasgraf:
This adds the translated title fields (`title_de`, `title_en`, `title_fr`) to the Solr schema, and make use of them in the `@navigation` endpoint.

### Implementation Details

- Added fields (`title_de`, `title_en`, `title_fr`) to Solr schema

  The fields were added with

    - `type="text"`
      I briefly considered introducing `text_de`, etc. field types, but unlike in `collective.solr` days, we don't have any language specific analyzers any more for the `text` type.

    - `indexed="true"`
      Having these fields indexed is not strictly necessary yet, because we don't query them. But if we at some point decide to query them, this avoids having to reindex them all.

    - `stored="true"`
      Because we need to retrieve the fields, they need to be stored. `useDocValuesAsStored` is `true` by default, but unlike the `string` type, the `text` doesn't have `docValues="true"`, so we need still to make them stored to be retrievable.

  An upgrade step is included to reindex them. It's `deferrable`, and uses the `NightlyIndexer` to reindex those fields. The changes in `@navigation` were tested to not fail (fall back to Title()) with both the new Solr fields completely absent, and with them being present, but not indexed yet.

- Added support for translated titles in `OGSolrDocument`

  Support for the translated title fields was added to `OGSolrDocument` (instead of the higher level `OGSolrContentListingObject`) to have the OGSolrDocument behave as much as a catalog brain as possible. Therefore, the implementation also mirrors the quirks regarding string types:
  - `OGSolrDocument.title_de` will be unicode
  - `OGSolrDocument.Title` will be UTF-8
  - `OGSolrContentListingObject.Title()` will be UTF-8 as well

  Fallback behavior is also mirrored from what brains do: If `OGSolrDocument.title_de` is falsy, it will fall back to `OGSolrDocument.Title` (so whatever title got written at indexing time).

### Caveats

In order for the fallback to `OGSolrDocument.Title` to work, queries that include `title_de` etc. in their `fl` must always also include `Title`. Otherwise `Title` won't be returned by Solr, and we'll have to fall back to `None` / `null`.

### Not addressed
This does **NOT** address:
- The Solr `Title` index containing content in whatever language was active at indexing time (same is true for the respective Catalog index)
- Full text search for repo folder titles: Since the `Title` index only contains content in one language, and the translated title fields are not yet concatenated into a searchable field with all the languages, content titles will only be found in the language that was active during indexing time. Not a big deal for repo folders IMHO.

For [CA-3847](https://4teamwork.atlassian.net/browse/CA-3847)

## Checklist

- [x] Changelog entry
- Upgrade-Steps:
- API change:


opengever/base/tests/test_translated_title.py
```diff
@@ -371,6 +372,14 @@ def test_solr_data_doesnt_contain_keys_for_objects_without_translated_title_supp
         self.assertNotIn('title_fr', solr_data)
         self.assertNotIn('title_en', solr_data)
 
+    def test_og_solr_document_returns_none_for_objects_without_translated_title_support(self):
+        self.login(self.regular_user)
+
+        og_solr_doc = OGSolrDocument(solr_data_for(self.dossier))
+        self.assertEquals(None, og_solr_doc.title_de)
```
njohner: self.assertIsNone?

opengever/base/tests/test_translated_title.py
```diff
@@ -582,6 +620,15 @@ def test_all_solr_stored_values(self):
                 self.titles[lang],
                 solr_data['title_{}'.format(lang)])
 
+    def test_all_og_solr_document_values(self):
```
njohner, line 623: I guess that test would have belonged to the previous commit, but it doesn't matter

njohner (approved):
Huh, I thought that we would need to do something to get the reference number as prefix for the title, but it seems I was wrong, or I'm did not see where that is done 😅 . Seems almost too easy...

But LGTM! 🎉

PR #7393, opened 2022-03-15
Update DEFAULT_DASHBOARD_CARDS with the responsibleToggleFilters

elioschmutz:
This PR updates the DEFAULT_DASHBOARD_CARDS with the responsibleToggleFilters for the `DashboardCard`.

This is required since we no longer provide the `myDossiersOnly` property in the frontend for more flexibility in configuring the cards.

For [CA-3682]

This is a follow-up PR for https://github.com/4teamwork/gever-ui/pull/2189

## Checklist

- [ ] Changelog entry (not required because it's a follow-up pr in the same release)

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-3682]: https://4teamwork.atlassian.net/browse/CA-3682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/base/interfaces.py
+++ b/opengever/base/interfaces.py
@@ -387,7 +387,7 @@ DEFAULT_DASHBOARD_CARDS = [
         'title_en': 'My dossiers',
         'title_fr': 'Mes dossiers',
         'componentName': 'DossiersCard',
-        'myDossiersOnly': True
+        'responsibleToggleFilters': ['myDossiers', 'anyParticipation'],
     },
     {
         'id': 'substitute_dossiers',


opengever/base/interfaces.py
```diff
@@ -387,7 +387,7 @@ class IOGMailSettings(Interface):
         'title_en': 'My dossiers',
         'title_fr': 'Mes dossiers',
         'componentName': 'DossiersCard',
-        'myDossiersOnly': True
+        'responsibleToggleFilters': ['myDossiers', 'anyParticipation'],
```
njohner, line 390: So this property should always be a list with 2 items in it?

njohner (approved):
LGTM

opengever/base/interfaces.py
elioschmutz: No, it can also be i.e. empty or even contain a list of all possible filters: `['all', 'myDossiers', 'anyParticipation']`

PR #7394, opened 2022-03-15
Testserver: add docker-compose example

jone:
The goal is to make it easier to run the gever testserver with docker compose when working on third party projects (such as ris and vertragsmanagement) because using docker is much simpler when not working directly on opengever.core.

This pull request basically adds a working, minimal docker compose example only containing the testserver and adds some convenience.

Changes:

- A `docker-compose.testserver.yml` is provided, so that it can be copied to other projects. It contains a minimal working example with everything needed for the testserver. It also contains a commented example of how to mount a fixture module using a volume, because that's what third party projects usually do.
- The testserver image is extended to have sane defaults for service `ENV`s using the common service names and ports. This simplifies the docker compose file so that we do not have to configure the services into the testserver; one less part copied around in the projects.
- The `opengever.core` version is no printed when starting the testserver so that it easier to realize that the images are outdated.

Please build a new image of `4teamwork/ogcore:latest` and `4teamwork/ogtestserver`. There is no ogcore build from this year.

## Checklist

- [X] Changelog entry

- Upgrade-Steps:
  - [X] SQL Operations do not use imported model (see [docs](https://4teamwork.atlassian.net/wiki/spaces/4TEAM/pages/994344994/Upgrade-Steps))
  - [ X Make it deferrable if possible
  - [X] Execute as much as possible conditionally
  - DB-Schema migration
    - [X] All changes on a model (columns, etc) are included in a DB-schema migration.
    - [X] Constraint names are shorter than 30 characters (`Oracle`)
- API change:
  - [X] Documentation is updated
  - [X] API Changelog entry (see [guide](https://4teamwork.atlassian.net/wiki/spaces/4TEAM/pages/451248812/API+Changelog+Guidelines))
  - If breaking:
    - [X] api-change label added
    - [ X #delivery channel notified about breaking change
- Bug fixed:
- New functionality:
- [x] Change could impact client installations, client policies need to be adapted
- New translations

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1237,6 +1237,10 @@ Testserver in docker
 You can run testserver with docker-compose: ``docker-compose up testserver``.
 See the `testerver docker readme <docker/testserver/README.md>`_.
 
+If you are using the testserver in another project and want to have a docker-compose file there,
+see the ``docker-compose.testserver.yml`` file for a minimal working example.
+It contains commented example on how insert your custom fixture as volume.
+
 
 Testing Inbound Mail
 --------------------
--- a/changes/testserver.feature
+++ b/changes/testserver.feature
@@ -1,0 +1,1 @@
+Improve docker / docker compose support for testserver. [jone]
--- a/docker-compose.testserver.yml
+++ b/docker-compose.testserver.yml
@@ -1,0 +1,33 @@
+# This example docker compose file contains a minimal
+# working example of a testserver configuration.
+# This file can be copied to other projects which need
+# the testserver.
+
+services:
+  testserver:
+    image: 4teamwork/ogtestserver:latest
+    ports:
+      - 55001:55001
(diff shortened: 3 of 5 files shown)


README.rst
```diff
@@ -1237,6 +1237,10 @@ Testserver in docker
 You can run testserver with docker-compose: ``docker-compose up testserver``.
 See the `testerver docker readme <docker/testserver/README.md>`_.
 
+If you are using the testserver in another project and want to have a docker-compose file there,
+see the ``docker-compose.testserver.yml`` file for a minimal working example.
+It contains commented example on how insert your custom fixture as volume.
```
buchi, line 1242:
```suggestion
It contains a commented example on how to insert your custom fixture as volume.
```

buchi (approved):
👍🏻 LGTM

PR #7396, opened 2022-03-16
Allow nesting tasktemplatefolders.

njohner:
With this PR we add support for nesting `TaskTemplateFolder`s. This PR does not implement the actual usage of nested `TaskTemplateFolder`s which will come in a separate PR. Instead, the focus here is to have nesting work in the `TemplateFolder`. Classic UI will offer only partial support for that feature.
Here the main changes that were introduced:
- Allow nesting tasktemplatefolders
- Add new field `is_subtasktemplatefolder`, which is necessary to only show the main `tasktemplatefolders` in various listings (also in the Gever-UI)
- Add `is_tasktemplatefolder_nesting_enabled` feature flag.
- Do not show subtasktemplatefolders in TaskTemplateFoldersTab. This means sub-TaskTemplateFolders are not listed anywhere in the classic UI, we therefore also display a warning when the user views a `TaskTemplateFolder` containing nested `TaskTemplateFolder`s.
- We also index `getObjPositionInParent` for sub TaskTemplateFolders, as `TaskTempalte`s and `TaskTemplateFolder`s can be ordered in the new UI.
- Finally nested `TaskTemplateFolder`s should always have a coherent workflow state, so that they can be triggered or not as a whole. We therefore make the workflow transitions only available on the main `TaskTemplateFolder`s but make them recursive. We also add an event handler to make sure that a newly created sub `TaskTemplateFolder` has the same workflow status as its parent.

There was also a bug with `sequence_type` field in the listing endpoint, which would throw an error when requested for any other object than `TaskTemplate`s

For [CA-3586]

## Checklist
- [x] Changelog entry
- Upgrade-Steps:
 
[CA-3586]: https://4teamwork.atlassian.net/browse/CA-3586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/base/indexes.py
```diff
@@ -30,6 +31,7 @@
 # one of thoes interfaces.
 CONTENTS_SUPPORTING_OBJ_POSITION_IN_PARENT = (
     ITaskTemplate,
+    ITaskTemplateFolderSchema,
```
phgross, line 34: Do we need an upgradestep for this 🤔

phgross (approved):
LGTM 👍 

just one uncertainty ...

opengever/base/indexes.py
njohner: No because we only index it for subtasktemplatefolders, see indexer below. Because there are None yet, we don't need an upgrade step.

phgross: 👍 makes sense

PR #7397, opened 2022-03-16
Teamraum visible users

elioschmutz:
This PR implements a special protection for users, groups, vocabularies, sources and endpoints within a teamraum.

This is required because a teamraum has a complete different security policy than a gever has. 

For [CA-3592]

## Checklist

- |x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-3592]: https://4teamwork.atlassian.net/browse/CA-3592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/workspace/__init__.py
```diff
@@ -7,6 +7,17 @@
 
 WHITELISTED_TEAMRAUM_GLOBAL_SOURCES = []
 
+WHITELISTED_TEAMRAUM_PORTAL_TYPES = [
```
buchi, line 10: make this a `set` for a more efficient containment check

opengever/base/visible_users_and_groups_filter.py
```diff
@@ -0,0 +1,89 @@
+from opengever.ogds.models.service import ogds_service
+from opengever.workspace import is_workspace_feature_enabled
+from plone import api
+
+
+class VisibleUsersAndGroupsFilter:
+    """This singleton object provides functions to check if a
```
buchi, line 7:
```suggestion
    """Provides functions to check if a
```

The singleton part is not relevant here nor does this class implement the singleton pattern.

opengever/base/visible_users_and_groups_filter.py
```diff
@@ -0,0 +1,89 @@
...
+    specific user or group should be visible to the currently logged in user
+    or not.
+
+    In general, every user can get information about every other user and group
```
buchi, line 11: I would try to shorten the rest of the docstring, e.g.:

"In GEVER user and group information is exposed to all users independant of their access rights.
In teamraum users are only allowed to get user information about users that participate in workspaces the user has access to."

buchi (commented):
LGTM, just some minor suggestions, mainly wording...
I still have to do a performance check.

elioschmutz:
@buchi 👍  I updated all of your change requests.

buchi:
@elioschmutz I've done some basic performance checks. It's not as bad as expected because in teamraum there are not many calls to the protected endpoints.

I've noticed that adding todo gets noticeably slower and `can_access_principal()` is called multiple times. Thus I would suggest to cache `get_whitelisted_principals()` on the request, which is quite simple und should be sufficient.

Unfortunately the `@view.memoize_contextless` decorator can't be used as the VUAGF doesn't have a `request` attribute (could be added though). But you could do something like this:
```
def get_whitelisted_principals(self):
    request = getRequest()
    allowed_users_and_groups = getattr(request, '_vuagf_allowed_users_and_groups', None)
    if allowed_users_and_groups is None:
        allowed_users_and_groups = set(self.extract_allowed_users_and_groups())
        allowed_users_and_groups |= set(self.extract_group_members(allowed_users_and_groups))
        allowed_users_and_groups.add(api.user.get_current().getId())
        request._vuagf_allowed_users_and_groups = allowed_users_and_groups
    return allowed_users_and_groups
```

buchi:
An acronym for VisibleUsersAndGroupsFilter 😉

elioschmutz:
👍 @buchi I added caching. See my last commit

buchi (approved):

PR #7398, opened 2022-03-16
Accept groups in `notification_recipients` for `@external-activities` endpoint 

lukasgraf:
This adds support for the `notification_recipients` parameter on `@external-activities` to also accept group IDs, resolve them to their members, and dispatch notification to those users.

For this, sending notifications to users _other than yourself_ had to be allowed in that endpoint for users with the necessary privileges. The new permission `opengever.api: Notify Arbitrary Users` allows this (assigned to the new `PrivilegedNotificationDispatcher` role).

In addition, I improved the error handling in that endpoint. Most problems should only cause soft errors, that are reported in an `{'errors': [...]}` list in the response, but still with status `201 Created`.

For [CA-3274](https://4teamwork.atlassian.net/browse/CA-3274)

## Checklist

- [x] Changelog entry
- API change:
  - [x] API Changelog entry

### DIFF
--- a/opengever/api/tests/test_external_activities.py
+++ b/opengever/api/tests/test_external_activities.py
@@ -3,6 +3,7 @@ from ftw.testbrowser import browsing
 from ftw.testing import freeze
 from opengever.activity.model import Activity
 from opengever.testing import IntegrationTestCase
+from plone import api
 import json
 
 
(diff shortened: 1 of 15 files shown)


opengever/api/tests/test_external_activities.py
```diff
@@ -312,3 +315,85 @@ def test_creating_external_activity_validates_schema_fields(self, browser):
...
+            'userid': 'david.meier',
+            }], browser.json['errors'])
+
+        activity = Activity.query.all()[-1]
+        self.assertItemsEqual(
+            [user.userid for user in group.users],
+            [notification.userid for notification in activity.notifications],
+        )
+
+    @patch('opengever.activity.mail.PloneNotificationMailer.dispatch_notification')
```
njohner, line 369: I don't think you need the patching here -> ✂️

njohner (approved):
LGTM

PR #7408, opened 2022-03-25
Fix reference_prefixes update mechanism for removed documents.

phgross:
It's not guaranteed that the user which changes the reference_prefix on a repositoryfolder, are allowed to access all content inside this folder, even if its a Administrator, for example if a document is in the removed-state. Therefore we need to reindex the objects with elevated_privileges.

For [CA-3880]

## Checklist

- [x] Changelog entry

[CA-3880]: https://4teamwork.atlassian.net/browse/CA-3880?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/repository/subscribers.py
+++ b/opengever/repository/subscribers.py
@@ -1,3 +1,4 @@
+from opengever.base.security import elevated_privileges
 from opengever.document.behaviors import IBaseDocument
 from opengever.repository.interfaces import IDuringRepositoryDeletion
 from plone import api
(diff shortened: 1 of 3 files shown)


opengever/repository/subscribers.py
```diff
@@ -32,15 +33,16 @@ def update_reference_prefixes(obj, event):
...
-            idxs = ['reference', 'sortable_reference']
-            if IBaseDocument.providedBy(obj):
-                idxs.append('metadata')
-            else:
-                idxs.append('SearchableText')
-
-            obj.reindexObject(idxs=idxs)
+        with elevated_privileges():
+            for child in children:
+                obj = child.getObject()
```
njohner, line 38: You could also have used `_unrestrictedGetObject`, but fine with me

njohner (approved):
LGTM

opengever/repository/subscribers.py
phgross: ah yes that would also have been possible, I let it stay like that, I tested it so extensively.

phgross:
Backported to `2022.4-stable` not release yet.

PR #7413, opened 2022-03-31
Add @task-template-structure endpoint

elioschmutz:
This PR introduces a new endpoint `@task-template-structure` to fetch the full structure of a task-template folder. This allows us to easely render a nested form in the frontend without doing lots of requests or manually building a tree in the frontend.

For [CA-3725]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

### DIFF
--- a/opengever/api/templatefolder.py
+++ b/opengever/api/templatefolder.py
@@ -19,9 +19,11 @@ from opengever.ogds.models.service import ogds_service
 from opengever.task.task import ITask
 from opengever.tasktemplates.sources import TaskResponsibleSourceBinder
 from plone import api
+from plone.dexterity.interfaces import IDexterityContainer
 from plone.protect.interfaces import IDisableCSRFProtection
 from plone.restapi.deserializer import json_body
 from plone.restapi.interfaces import IFieldDeserializer
+from plone.restapi.interfaces import ISerializeToJson
 from plone.restapi.interfaces import ISerializeToJson
 from plone.restapi.services import Service
 from plone.supermodel import model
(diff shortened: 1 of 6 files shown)


opengever/api/templatefolder.py
```diff
@@ -399,3 +401,22 @@ def _validate_related_documents(self, data):
...
+    def reply(self):
+        return self.recursive_serialize(self.context)
+
+    def recursive_serialize(self, obj):
+        result = queryMultiAdapter((obj, self.request), ISerializeToJson)()
+
+        if IDexterityContainer.providedBy(obj):
+            items = []
+
+            for child in obj.listFolderContents():
```
njohner, line 418: I don't think you will have the sorting correct here. We want to sort according to `getObjPositionInParent` at least for sequential tasktemplatefolders. See `TaskTree`

njohner (commented):
Just a thought: the whole API could be more consistent if we had one endpoint `@trigger-task-template` defined for `IDossierMarker`. Now a GET on this endpoint would return the structure as you define it in this PR, but it would replace the interactive actors with the correct ones (i.e. current user and dossier responsible). POST of course would accept exactly what GET returns and create the tasks accordingly.

opengever/api/templatefolder.py
elioschmutz: `listFolderContents()` already respects the position in parent. So I think there is no need to change here

elioschmutz:
@njohner I update the PR with another test.

As we discussed, we go with this PR without any change because it makes sense to have such a structure endpoint.

Instead of adding a GET/POST to the `trigger` endpoint, we'll introduce a new endpoint to create nested tasks without any relation to tasktemplates. So the frontend will use this new endpoint here to render the form.

njohner (approved):
LGTM

PR #7414, opened 2022-04-01
 Add response to parent task when subtask is closed or cancelled

tinagerber:
Journal entries are created only for newly closed or canceled tasks.

Bildschirmfoto 2022-04-01 um 17 15 46

For [CA-3573]

## Checklist

- [x] Changelog entry

[CA-3573]: https://4teamwork.atlassian.net/browse/CA-3573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-3573.feature
+++ b/changes/CA-3573.feature
@@ -1,0 +1,1 @@
+Add response to parent task when subtask is closed. [tinagerber]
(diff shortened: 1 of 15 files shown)


njohner (approved):
LGTM, I was just a bit confused by the links to the subtasks having themselves also an icon of the state of the subtask, but the current state, so completely unrelated to the transition. Anyway it was already like that, and I guess it's just me 😅

PR #7416, opened 2022-04-04
Add @process endpoint supporting nested task sequences.

njohner:
In this PR we implement a new endpoint `@process` which allows to create task sequences. This endpoint is a substitute for the `@trigger-task-template`, but allowing more flexibility. It notably supports nested task sequences and the created process does not need to come from a `TaskTemplateFolder`. The old functionality and endpoint is kept for now, so that we do not have an API change. A few notable points:
- The `@process` can now validate the data of subtasks directly using the `ITask` schema, as we do not have half the data coming from the corresponding `TaskTemplate` which forced us to validate everything by hand before.
- All fields from `ITask` are now supported for the subtasks
- Validation, deadline calculation, data preprocessing and process creation are now all recursive and support any nesting depth

A few things that still need to be addressed:
- The interfaces used for marking Sequential and Parallel processes (`IFromSequentialTasktemplate`, `IFromParallelTasktemplate`) are not sufficient anymore as is, because we were relying on the main task and the subtasks to provide the same interface. With nesting this is not possible anymore, because a sequential `TasktemplateFolder` can be nested into a parallel one, so that the created task should have both interfaces, `IFromParallelTasktemplate` because it is part of a parallel process because of its parent, and `IFromSequentialTasktemplate` because it is the parent of a sequential set of tasks. This leads to various issues and needs to be looked at, notably:
  - Initial states of the created Tasks are not yet as expected (This is at least partially linked to the next point)
  - tasktemplate predecessor is not set correctly
- Check start next
- Check event handlers
- Documentation

For [CA-3725]

## Checklist
- [ ] Changelog entry
- API change:
  - If breaking:
    - [ ] api-change label added

### DIFF
--- a/opengever/api/process.py
+++ b/opengever/api/process.py
@@ -2,7 +2,9 @@ from opengever.api.add import get_validation_errors
 from opengever.api.task import deserialize_responsible
 from opengever.base.source import DossierPathSourceBinder
 from opengever.ogds.base.utils import get_current_org_unit
+from opengever.task import _ as task_mf
 from opengever.task.task import ITask
+from opengever.tasktemplates import _ as tasktemplate_mf
 from opengever.tasktemplates.content.templatefoldersschema import sequence_type_vocabulary
 from opengever.tasktemplates.tasktemplatefolder import ProcessCreator
 from opengever.tasktemplates.tasktemplatefolder import ProcessDataPreprocessor
@@ -13,14 +15,15 @@ from plone.restapi.interfaces import IFieldDeserializer
 from plone.restapi.interfaces import ISerializeToJson
 from plone.restapi.services import Service
 from plone.supermodel import model
+from z3c.relationfield.relation import RelationValue
 from z3c.relationfield.schema import RelationChoice
 from z3c.relationfield.schema import RelationList
 from zExceptions import BadRequest
 from zope import schema
+from zope.component import getUtility
 from zope.component import queryMultiAdapter
 from zope.interface import alsoProvides
-from opengever.task import _ as task_mf
-from opengever.tasktemplates import _ as tasktemplate_mf
+from zope.intid.interfaces import IIntIds
 
 
 class IProcessSchema(model.Schema):
@@ -131,6 +134,7 @@ class ProcessPost(Service):
         if errors:
             self.raise_bad_request(errors)
 
+        self.make_related_documents_relation_values(data)
         self.fill_task_container_data(data["process"])
 
         data = data_processor.recursive_set_deadlines()
@@ -145,6 +149,13 @@ class ProcessPost(Service):
         # have to make sure to return the correct id.
         result['@id'] = task.absolute_url()
         return result
+
+    def make_related_documents_relation_values(self, data):
+        relations = []
+        for doc in data.get("related_documents"):
+            relations.append(RelationValue(getUtility(IIntIds).getId(doc)))
+
+        data["related_documents"] = relations
 
     def fill_task_container_data(self, data):
         """The task containers have some automatic values for certain fields,
--- a/opengever/api/process.py
+++ b/opengever/api/process.py
@@ -1,0 +1,205 @@
+from opengever.api.add import get_validation_errors
+from opengever.api.task import deserialize_responsible
+from opengever.base.source import DossierPathSourceBinder
+from opengever.ogds.base.utils import get_current_org_unit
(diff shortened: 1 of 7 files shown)


opengever/tasktemplates/tasktemplatefolder.py
```diff
@@ -182,8 +175,25 @@ def preprocess_data(self):
 
     def _recursive_preprocess_data(self, data):
         self.replace_interactive_actors(data)
-        for item in data.get("items", []):
-            self._recursive_preprocess_data(item)
+        if self.has_children(data):
```
phgross, line 166: As discussed, I'm not sure this calculation makes sense here. But I think it is ok for now and can be adjusted in a separate PR.

opengever/tasktemplates/tasktemplatefolder.py
```diff
@@ -238,14 +238,27 @@ def create_main_task(self, data):
...
         return main_task
 
+    @staticmethod
+    def set_state(task, state):
+        initial_state = api.content.get_state(task)
+        wftool = api.portal.get_tool('portal_workflow')
+        wf_id = wftool.getWorkflowsFor(task)[0].id
+        wftool.setStatusOf(wf_id, task, {'review_state': state})
+        wftool.getWorkflowsFor(task)[0].updateRoleMappingsFor(task)
+        return initial_state
```
phgross, line 258: When changing the `state` like this, its not guaranteed that the representation in the globalindex is updated. Maybe extend the tests to make it sure.

opengever/api/process.py
```diff
@@ -0,0 +1,205 @@
...
+    @staticmethod
+    def is_task_container(data):
+        if "items" in data or "sequence_type" in data:
+            return True
+        return False
+
+    @staticmethod
+    def raise_bad_request(errors):
+        errors = [{"field": each[0], "error": each[1]} for each in errors]
+        raise BadRequest(errors)
```
phgross, line 188: Are the errors already translated and can be shown correctly in the UI?

phgross (approved):
Nice work 👍 

Only a few question marks and comments.

opengever/tasktemplates/tasktemplatefolder.py
njohner: Yes I agree. I think we would still need the recursive calculation, but maybe not add the 5 days at the end, so would be an easy fix

opengever/tasktemplates/tasktemplatefolder.py
njohner: We don't really want the global index to be updated, as this is a temporary change. In create_subtasks we change the state to task-state-in-progress but a few lines after we set the state back to its initial state.
I'll add a test none the less, as there is a risk that this suddenly gets used differently in the future.

opengever/api/process.py
njohner: Formatting the errors as I do here allows them to be recognised as validation errors in our error handling and therefore be "translated" (basically just the "invalid input" is translated). Frontend will have a hard time showing the errors correctly though, as in the recursive schema, each field appears multiple times, with no real way of distinguishing them. We would need to invest an unreasonable amount of time here IMO if we want to do something better, knowing that the frontend does not handle any form validations correctly currently (it only has its own form validation but does not handle validation errors returned by the API).

opengever/api/tests/test_process.py
```diff
@@ -0,0 +1,551 @@
...
+            }
+        }
+
+        with browser.expect_http_error(400):
+            browser.open('{}/@process'.format(
+                         self.dossier.absolute_url()),
+                         data=json.dumps(data),
+                         headers=self.api_headers)
+
+        self.assertEqual(u'Inputs not valid',
```
njohner, line 380: Here you see that the error is translated

opengever/api/tests/test_process.py
```diff
@@ -0,0 +1,551 @@
...
+                         browser.json[u'translated_message'])
+
+        self.assertItemsEqual(
+            [{u'field': u'responsible_client',
+              u'translated_message': u'None',
+              u'type': u'responsible_client'},
+             {u'field': u'responsible',
+              u'translated_message': u'None',
+              u'type': u'responsible'}],
+            browser.json[u'additional_metadata']['fields'])
```
njohner, line 390: with `additional_metadata` that is only somewhat helpful

PR #7420, opened 2022-04-07
Allow moving TR documents to trash after copying them back to GEVER

lukasgraf:
Corresponding Frontend PR: 4teamwork/gever-ui#2219

---

This adds support for allowing Teamraum documents to be **moved to the trash** after copying them back to GEVER via the `@copy-document-from-workspace` endpoint.

I changed the endpoint's error responses so that they include a `translated_message`. These will be displayed by the frontend when moving the document to the trash fails for some reason (and in some other cases as well):

Screenshot 2022-04-07 at 15 13 01

In order to cover the case where only moving the teamraum document to trash failed, but everything else was successful, the transaction is explicitly committed in that case, to avoid rolling back the already successful transfer of the document to GEVER.

Note: This feature currently does not work when developing locally. See follow up issue [CA-3986](https://4teamwork.atlassian.net/browse/CA-3986)

For [CA-3594](https://4teamwork.atlassian.net/browse/CA-3594)

## Checklist

- [x] Changelog entry

- API change:

- New functionality:

### DIFF
--- a/changes/CA-3594.feature
+++ b/changes/CA-3594.feature
@@ -1,0 +1,1 @@
+Add support for trashing TR documents after retrieving them back to GEVER. [lgraf]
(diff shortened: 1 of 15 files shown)


phgross (approved):
👍 LGTM

PR #7427, opened 2022-04-12
Solr with docker

buchi:
Run Solr with Docker instead of installing it with Buildout.

This has several advantages:
- The required services for development can be started with one command `docker-compose up -d`.
- No locally installed JDK required.
- Better isolation. Solr for development does not interfere with Solr for running tests.

The Solr image has been extended to support the creation of multiple cores on startup.
Further the configuration of each core is updated on container startup. Before this change, the configuration of a core was never updated even if there were changes added to the image.

For [CA-XXXX]

## Checklist

- [ ] Changelog entry

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -217,22 +217,10 @@ with these contents:
 respective user in our development LDAP tree.
 
 
-Installing and activating Solr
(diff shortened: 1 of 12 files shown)


njohner (approved):
LGTM

PR #7429, opened 2022-04-13
Add missing french translations

elioschmutz:
Add missing french translations.

Slack: https://4teamwork.slack.com/archives/C01BHEGBVU1/p1649836363595599

For [CA-4011]

[CA-4011]: https://4teamwork.atlassian.net/browse/CA-4011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/workspaceclient/locales/fr/LC_MESSAGES/opengever.workspaceclient.po
+++ b/opengever/workspaceclient/locales/fr/LC_MESSAGES/opengever.workspaceclient.po
@@ -16,7 +16,7 @@ msgstr ""
 
 #: ./opengever/workspaceclient/linked_workspaces.py
 msgid "Not moved to trash: Document was retrieved, but workspace document could not be moved to trash."
-msgstr ""
+msgstr "Non déplacé dans la corbeille: le document a été récupéré, mais le document de l’espace de travail n’a pas pu être déplacé vers la corbeille"
 
 #. Default: "Document retrieved from teamraum"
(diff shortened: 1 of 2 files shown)


opengever/workspaceclient/locales/fr/LC_MESSAGES/opengever.workspaceclient.po
```diff
@@ -61,4 +61,4 @@ msgstr "Lié à l'espace de travail ${workspace_title}."
 #. Default: "Unlinked workspace ${workspace_title}."
 #: ./opengever/workspaceclient/linked_workspaces.py
 msgid "label_workspace_unlinked"
-msgstr ""
+msgstr "Espace de travail non lié ${workspace_title}"
```
lukasgraf, line 64: My french is not good at all, but I think this might be wrong. The context is a journal entry, so the phrase should be in the past tense, not present.

So "Unlinked workspace ${workspace_title}." could also be phrased as "The workspace ${workspace_title} **has been** unlinked".

elioschmutz: Oh, good point 👁️

YBourqui: L'espace de travail n'a pas été lié

PR #7430, opened 2022-04-13
Support nested task process in the sequence handling

phgross:
This PR introduces 4 new marker interfaces wich replaces the former IFromSequentialTaskTemplate and IFromParallelTaskTemplate. Because since the introduction of nested tasktemplates with different sequence types we need to store two information on a task: is the task part of a sequential/parallel process? and contains the task a sequential/parallel process. The former interfaces still need to be around and can be removed later,  when the current version is deployed everywhere.
The PR contains an upgradestep which replaces the marker interface on existing task process.

For [CA-3725]

## Checklist
- [x] Changelog entry
- Upgrade-Steps:
  - [ ] Make it deferrable if possible _(not possible - but should not be that expensive)_
- New functionality:
  - [x] for `task` also works for `forwarding` _(tasktemplates only works with tasks)_


opengever/task/task.py
```diff
@@ -671,11 +671,36 @@ def open_next_task(self):
...
+        if IContainParallelProcess.providedBy(self):
+            for child in self.objectValues():
+                if ITask.providedBy(child):
+                    child._open_planned_task()
+        elif IContainSequentialProcess.providedBy(self):
+            for child in self.objectValues():
+                if ITask.providedBy(child):
+                    child._open_planned_task()
+                    # start only the first task if sequential
+                    return
```
njohner, line 703: What if the child is itself a process? Shouldn't this be recursive?

PR #7432, opened 2022-04-14
`@tasktree` endpoint properly handles nested tasks

elioschmutz:
This PR updates the `@tasktree` endpoint to properly handle nested tasks.

The previous implementation expected to have the main-task as the context and only one nesting level. This is no longer the case since we provide multi-nesting for by using the `@process` endpoint.

This PR removes the main-task specific attribute `is_task_addable_in_main_task` and adds two attributes: `is_task_addable` and `is_task_addable_before` in a consistent way for each item within the task-tree.

For [CA-3725]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [x] api-change label added
- Bug fixed:
- New functionality:
- New translations

### DIFF
--- a/opengever/api/tasktree.py
+++ b/opengever/api/tasktree.py
@@ -7,6 +7,7 @@ from opengever.base.solr import OGSolrContentListing
 from opengever.task import TASK_STATE_PLANNED
 from opengever.task.task import ITask
 from opengever.tasktemplates.interfaces import IContainSequentialProcess
+from opengever.tasktemplates.interfaces import IPartOfSequentialProcess
 from plone.restapi.interfaces import IExpandableElement
 from plone.restapi.services import Service
 from zope.component import adapter
@@ -32,18 +33,23 @@ class TaskTree(object):
         if not expand:
             return result
         result['tasktree']['children'] = self.task_tree()
-        result['tasktree']['is_task_addable_in_main_task'] = self.is_task_addable_in_main_task()
         return result
 
-    def is_task_addable_in_main_task(self):
-        main_task = self.get_main_task()
-        for fti in main_task.allowedContentTypes():
-            if fti.id == main_task.portal_type:
+    def is_task_addable_in_container(self, container):
+        if not IContainSequentialProcess.providedBy(container):
+            return False
+
+        for fti in container.allowedContentTypes():
+            if fti.id == container.portal_type:
                 return True
         return False
 
-    def is_task_addable_before(self, obj):
-        return (obj.review_state() == TASK_STATE_PLANNED) and self.is_task_addable_in_main_task()
+    def is_task_addable_before(self, solr_item, parent_context):
+        if IPartOfSequentialProcess.__identifier__ not in solr_item.object_provides \
+                or not parent_context:
+            return False
+        return solr_item.review_state() == TASK_STATE_PLANNED \
+            and self.is_task_addable_in_container(parent_context)
 
     def get_main_task(self):
         main_task = self.context
@@ -52,6 +58,39 @@ class TaskTree(object):
             main_task = parent
             parent = aq_parent(main_task)
         return main_task
+
+    """We need to know if we can add subtasks to a specific task and if it possible
+    to add a new task before a specific task.
+
+    We can't just get out this information from the task objects due to
+    perforamnce reasons. We get all tasks by quering the solr which gives us
+    just solr-items.
+
+    We try to do as much as possible with just those items. But we need to
+    lookup the obejct to check if we can add tasks within the object. Since
+    leaf tasks do never allow adding new tasks, we can restrict the object
+    lookup to IContainProcess-tasks only. This reduces the object lookup to
+    a minimum.
+    """
+    def extend_tree_with_addable_information(self, tree, solr_items_per_url, parent_context=None):
+        solr_item = solr_items_per_url.get(tree.get('@id'))
+        is_task_addable_before = False
+        is_task_addable = False
+        children = tree.get('children')
+        if children:
+            context = solr_item.getObject()
+            is_task_addable_before = self.is_task_addable_before(solr_item, parent_context)
+            is_task_addable = self.is_task_addable_in_container(context)
+
+            for child in children:
+                self.extend_tree_with_addable_information(child, solr_items_per_url, context)
+
+        else:  # is a leafnode, we do not want to get the context of leave-nodes due to performance reasons
+            is_task_addable_before = self.is_task_addable_before(solr_item, parent_context)
+            is_task_addable = False
+
+        tree['is_task_addable_before'] = is_task_addable_before
+        tree['is_task_addable'] = is_task_addable
 
     def task_tree(self):
         main_task = self.get_main_task()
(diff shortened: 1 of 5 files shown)


opengever/api/tasktree.py
```diff
@@ -64,25 +104,27 @@ def task_tree(self):
             },
             object_provides=ITask.__identifier__,
         )
-        fieldlist = ['Title', 'portal_type', 'path', 'review_state']
+        fieldlist = ['Title', 'portal_type', 'path', 'review_state', 'object_provides']
```
phgross, line 107: just do be sure, the `object_provides` is not part of the serialized return value, by mistake?

opengever/api/tests/test_tasktree.py
```diff
@@ -9,44 +12,209 @@ def test_get_tasktree_for_task(self, browser):
...
+        self.assertFalse(browser.json['children'][0]['children'][2]['is_task_addable'],
+                         "Tasks should not be addable on a subtask")
 
     @browsing
     def test_is_task_addable_before(self, browser):
         self.login(self.regular_user, browser=browser)
-        browser.open(self.sequential_task, view="@tasktree", method="GET", headers=self.api_headers)
-        self.assertFalse(browser.json['children'][0]['children'][0]['is_task_addable_before'])
-        self.assertTrue(browser.json['children'][0]['children'][1]['is_task_addable_before'])
+        data = {
```
phgross, line 131: isn't that exactly the same data as on line 40?

phgross (approved):
LGTM besides the things mentioned.

opengever/api/tasktree.py
elioschmutz: that's ok. We don't need (don't want) this information in the serialized value.

opengever/api/tests/test_tasktree.py
elioschmutz: yes, it is 😅 ... I could have merged this two tests... I'll do it

opengever/api/tasktree.py
```diff
@@ -32,18 +33,23 @@ def __call__(self, expand=False):
...
             return result
         result['tasktree']['children'] = self.task_tree()
-        result['tasktree']['is_task_addable_in_main_task'] = self.is_task_addable_in_main_task()
         return result
 
-    def is_task_addable_in_main_task(self):
-        main_task = self.get_main_task()
-        for fti in main_task.allowedContentTypes():
-            if fti.id == main_task.portal_type:
+    def is_task_addable_in_container(self, container):
```
njohner, line 38: This has become confusing. Before that method was really checking whether tasks were addable in the main task, now it is not checking that tasks are addable in the container. I mean tasks can be added in a container that does not provide the `IContainSequentialProcess`

opengever/api/tasktree.py
```diff
@@ -53,6 +59,39 @@ def get_main_task(self):
...
+    """We need to know if we can add subtasks to a specific task and if it possible
+    to add a new task before a specific task.
+
+    We can't just get out this information from the task objects due to
+    perforamnce reasons. We get all tasks by quering the solr which gives us
+    just solr-items.
+
+    We try to do as much as possible with just those items. But we need to
+    lookup the obejct to check if we can add tasks within the object. Since
+    leaf tasks do never allow adding new tasks, we can restrict the object
```
njohner, line 71: I don't think that is true. I can add a task in a task, whether it is a leaf task or not

opengever/api/tasktree.py
elioschmutz: All this functions here are in context of the tasktree. And we do not allow to add tasks in a container if it does not provide the `IContainSequentialProcess` interface.

Do you have a suggestion of how we could make this function more clear?

opengever/api/tasktree.py
elioschmutz: Same here, this function is in context of the tasktree and we do not allow to add new tasks to a leaf-node by the tree itself. Possibly we can add new tasks through the default add menu, but here we're talking about this wonderful feature of adding new tasks with this amazing add-buttons between tasks in the task-tree widget of the gever-ui.

opengever/api/tasktree.py
```diff
@@ -53,6 +59,39 @@ def get_main_task(self):
...
+    lookup to IContainProcess-tasks only. This reduces the object lookup to
+    a minimum.
+    """
+    def extend_tree_with_addable_information(self, tree, solr_items_per_url, parent_context=None):
+        solr_item = solr_items_per_url.get(tree.get('@id'))
+        is_task_addable_before = False
```
njohner, line 77: is_task_addable_before = self.is_task_addable_before(solr_item, parent_context)

opengever/api/tasktree.py
```diff
@@ -53,6 +59,39 @@ def get_main_task(self):
...
+        is_task_addable = False
+        children = tree.get('children')
+        if children:
+            context = solr_item.getObject()
+            is_task_addable_before = self.is_task_addable_before(solr_item, parent_context)
```
njohner, line 82: ✂️

opengever/api/tasktree.py
```diff
@@ -53,6 +59,39 @@ def get_main_task(self):
...
+            is_task_addable = self.is_task_addable_in_container(context)
+
+            for child in children:
+                self.extend_tree_with_addable_information(child, solr_items_per_url, context)
+
+        else:  # is a leafnode, we do not want to get the context of leave-nodes due to performance reasons
+            is_task_addable_before = self.is_task_addable_before(solr_item, parent_context)
```
njohner, line 89: ✂️

opengever/api/tasktree.py
```diff
@@ -53,6 +59,39 @@ def get_main_task(self):
...
+        children = tree.get('children')
+        if children:
+            context = solr_item.getObject()
+            is_task_addable_before = self.is_task_addable_before(solr_item, parent_context)
+            is_task_addable = self.is_task_addable_in_container(context)
+
+            for child in children:
+                self.extend_tree_with_addable_information(child, solr_items_per_url, context)
+
+        else:  # is a leafnode, we do not want to get the context of leave-nodes due to performance reasons
```
njohner, line 88: You can drop that whole else statement

opengever/api/tasktree.py
njohner: How about rewriting this block as :
```
For sequential processes we allow adding subtasks at specific positions
in the process. For this we need to find out whether a given task contains
a sequential process and tasks can be added in it, as well as whether a
task can be added before a given task.

For performance reasons we avoid retrieving the objects and work only
with the solr-items whenever possible. But we need to lookup the obejct
to check if we can add tasks to the contained process the object. 
This is of vourse never the case for leaf tasks, so we can restrict
the object lookup to IContainProcess-tasks only. This reduces the object 
lookup to a minimum.
```

opengever/api/tasktree.py
elioschmutz: 👌

opengever/api/tasktree.py
elioschmutz: ... or `is_task_addable_in_sequential_task_container` ?

PR #7437, opened 2022-04-19
Optimize reindex of contained objects

buchi:
Improve performance when renaming dossiers.

`containing_dossier` and `containing_subdossier` indexes and metadata are removed from the catalog.
Having `containing_dossier` and `containing_subdossier` in Solr is sufficient. Reindexing these indexes and metadata is done for all contained objects and is expensive. Especially reindexing metadata always means reindexing all metadata.

`RelatedDocumentsCatalogTableSource` and the generation of Excel reports are now implemented using Solr.
Also various tests had to be switched to `SolrIntegrationTesting`.

Reindexing in Solr was further optimized by avoiding to fetch any child objects. 

Before these optimizations, renaming a dossier could take several minutes. E.g. renaming a dossier that contains ~1000 items took approximately 30 seconds. After optimizing, it takes ~3 seconds.

For [CA-3750](https://4teamwork.atlassian.net/browse/CA-3750)

## Checklist

- [x] Changelog entry

- Upgrade-Steps:


opengever/base/reporter.py
```diff
@@ -23,7 +24,7 @@ def value(input_string):
 def readable_author(author):
     """Helper method which returns the author description,
     instead of the userid"""
-    actor = Actor.lookup(author)
+    actor = Actor.lookup(six.ensure_text(author))
```
lukasgraf, line 27: This fails if `author` is None (which is the default for new documents).

opengever/base/solr.py
```diff
@@ -58,6 +67,13 @@ def _translated_title(self, lang):
         # On Catalog brains, title_* attributes are unicode. Mirror this here.
         return safe_unicode(self.get(field_name))
 
+    def _as_datetime(self, field):
+        value = self.data.get(field)
+        try:
+            return datetime.strptime(value, "%Y-%m-%dT%H:%M:%SZ")
```
lukasgraf, line 73: I've ran into this as well, but also with the changed field, those contents are different yet again. Because it's a datetime instead of a date, Solr returns it as `u'2022-04-21T14:05:58.825Z'` (note the three digits for milliseconds after the decimal points). Unless I'm missing something, this can't be easily parsed with Python's [`strptime()`](https://docs.python.org/2.7/library/datetime.html#strftime-and-strptime-behavior), which only know `%f` for `Microsecond as a decimal number, zero-padded on the left.`.

I therefore went with `dateutil.parser.parse(value)`, which handles pretty much anything you throw at it, except None.

opengever/api/tests/test_listing.py
```diff
@@ -764,7 +764,7 @@ def test_filter_by_start_date(self, browser):
         items = browser.json['items']
         start_dates = list(set(map(lambda x: x['start'], items)))
         self.assertEqual(1, len(start_dates))
-        self.assertEqual('2016-01-01T00:00:00Z', start_dates[0])
+        self.assertEqual('2016-01-01T00:00:00', start_dates[0])
```
lukasgraf, line 767: This probably warrants an API changelog entry 😢

lukasgraf (approved):
LGTM! 👍 🎉 
Just a few minor points.

opengever/api/tests/test_listing.py
buchi: This is not a negligible change as expected. I've quickly checked parsing with JS and without the `Z` the date is interpreted as local time instead of UTC.
Bildschirmfoto 2022-04-28 um 07 43 08

Another annoying thing is that the dossier representation returns just a date without time information:
```json
"start": "2016-01-01"
```
So we are inconsistent anyway. The proper solution would be to return dates only in listings as well.

lukasgraf: Yes, I noticed those inconsistencies as well.

For that reason, I kept the date parsing in my PR in the Excel export code, instead of doing it on the ContentListingObject, in order to not affect anything outside the Excel export. I already had to substantially refactor large parts of code that also get used in `@listing` and `@solrsearch` (but those _should_ be pure refactorings), so I didn't want to also introduce a change in behavior in all that.

I agree that it should be made consistent (return dates for dates, not datetimes, same string format and TZ offset everywhere), but I would rather not do it as part of a excel export or performance optimization PR.

opengever/base/reporter.py
buchi: This is now handled.

buchi: no longer part of this PR

PR #7446, opened 2022-04-27
Excel export: Support additional columns and custom properties

lukasgraf:
This adds support for exporting additional columns (ones not shown by default) and custom properties in the new frontend for Excel exports.

The behavior in the classic UI should remain the same (there should still be the same set of default columns, and if present, a tabbed view table config should be respected for filtering and ordering included columns), short of slight cosmetic changes.

This applies to Excel exports from **dossier** and **document** listings.

In addition, **proposal** Excel exports also had to be switched to Solr, because the `containing_(sub)dossier` indexes get dropped from the catalog in #7437.

---

The way it works is that for the `dossier_report` and `document_report` views, a `columns` parameter can optionally be specified in the request.

- If not present, the set of default columns is returned. In the classic UI, that set gets ordered and filtered using the tabbed view config, if one is present. In the new frontend, the default columns are ordered according to their definition order in `column_settings`.

- If a `columns` parameter is given, it is treated as an explicit, ordered list of columns to be included. The frontend may pass its column names 1:1 (e.g. `responsible_fullname`), and those get aliased to actual field/index names where necessary.

For [CA-2239](https://4teamwork.atlassian.net/browse/CA-2239)

## Checklist

- [x] Changelog entry

- New functionality:
- New translations


opengever/document/browser/report.py
```diff
@@ -1,64 +1,91 @@
...
+        },
+        {
+            'id': 'delivery_date',
+            'is_default': True,
+        },
+        {
+            'id': 'checked_out',
+            'is_default': True,
+            'title': _(u'label_document_checked_out_by'),
+            'transform': readable_author,
```
phgross, line 71: Is there a reason, why we can't reuse the existing transform, defined on the SolrField?

opengever/meeting/browser/report.py
```diff
@@ -1,48 +1,68 @@
...
+            'id': 'review_state',
+            'is_default': True,
+            'alias': 'review_state_label',
+            'transform': StringTranslater(None, 'plone').translate,
+        },
+        {
+            'id': 'issuer',
+            'is_default': True,
+            'title': _(u'label_issuer', default='Issuer'),
+            'transform': readable_author,
```
phgross, line 48: Same here and below, can we not reused the transform from the SolrListingField

opengever/base/browser/reporting_view.py
```diff
@@ -107,17 +107,25 @@ def __init__(self, *args, **kwargs):
...
-            sort='path asc',
             rows=1000,
-            fl=self.fields.get_query_fields(fields))
-
-        return [IContentListingObject(OGSolrDocument(d)) for d in resp.docs]
+            fl=self.fields.get_query_fields(fields) + ['path'])
+
+        # Sort results according to paths passed in request. Those should be
+        # sorted exactly as the user saw them in the UI (classic and geverui).
+        docs = sorted(
```
phgross, line 126: Will this possibly become a performance issue, with the planned change that exports of thousands of entries are  possible or should this not be a problem?

We would actually know the sorting criterion and could provide and use that instead of sorting manually by the paths.

phgross (approved):
LGTM 👍 - just some questions / thoughts

opengever/meeting/browser/report.py
lukasgraf: Short answer: Yes, we probably could.

And that's what I actually first planned to do. However, the transforms often have very slight differences between them, and I wanted to change the existing behavior as little as possible in this PR.

Otherwise it would have become very difficult to make sure I didn't break anything (in Excel exports, but also in the `@listing` and `@solrsearch` endpoints for example), if I change the behavior _and_ the tests in the same PR.

It's already huge, and at least for now, according to the tests, there's only a very small set of cosmetic differences in the Excel output.

And transforms are sometimes finicky and sometimes implemented pretty dumb. So defaulting to ALWAYS applying the transforms from the SolrListingFields in the Excel Exports was something that felt scary, and I didn't do in the end.

opengever/base/browser/reporting_view.py
lukasgraf: As discussed: I honestly don't expect this (the post-sorting according to paths in Python) to be a performance issue. Before this becomes a problem, I would first expect the long list of paths in the request, the time to produce the Excel, or the max number of rows in an Excel sheet to become a problem.

But what could be improved is the hard coded row limit of `1000`. In his PR, @buchi added a nice batching helper:

```python
def batched_solr_results(**kwargs):
    """Returns all Solr results in batches.
    """
    solr = getUtility(ISolrSearch)
    last_batch = False
    start = 0
    rows = kwargs.get('rows', 1000)
    while not last_batch:
        resp = solr.unrestricted_search(start=start, **kwargs)
        yield resp.docs
        if start + len(resp.docs) >= resp.num_found or not resp.docs:
            last_batch = True
        start += rows

    # Usage
    for batch in batched_solr_results(filters=filters, fl="..."):
        for doc in batch:
```

So this could be applied to the Solr query here, to remove the hard coded limit of 1000.

PR #7448, opened 2022-04-28
Set content type correctly in tus upload

njohner:
The content type is passed by the frontend to the `tus-upload` endpoint, but this content type comes from the browser or from uppy. This is therefore out of our control and does not work correctly for all content types that we support, notably for Cadwork files. We now set the content type in the backend, using the mimetypes registry, ensuring that content type is at least set correctly for all types supported by OfficeConnector. 

For [CA-4108]

## Checklist
- [x] Changelog entry
- Upgrade-Steps:

[CA-4108]: https://4teamwork.atlassian.net/browse/CA-4108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/core/upgrades/20220428110823_correct_content_type_for_cadwork/upgrade.py
+++ b/opengever/core/upgrades/20220428110823_correct_content_type_for_cadwork/upgrade.py
@@ -1,0 +1,48 @@
+from ftw.upgrade import UpgradeStep
+
+CADWORK_EXTENSIONS = [
+    '.2d',
+    '.2dc',
+    '.2dl',
+    '.2dm',
+    '.2dr',
+    '.3d',
+    '.3dc',
+    '.iv',
+    '.ivx',
+    '.ivz',
+    '.l2d',
+    '.lx2d',
+    '.lxz',
+]
+
+
+class AddCadworkMimeTypes(UpgradeStep):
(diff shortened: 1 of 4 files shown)


opengever/core/upgrades/20220428110823_correct_content_type_for_cadwork/upgrade.py
```diff
@@ -0,0 +1,46 @@
...
+    '.iv',
+    '.ivx',
+    '.ivz',
+    '.l2d',
+    '.lx2d',
+    '.lxz',
+]
+
+
+class AddCadworkMimeTypes(UpgradeStep):
```
phgross, line 20: Deferrable? 🤔

phgross (approved):
👍 LGTM, but I would make the upgrade step deferrable.

opengever/core/upgrades/20220428110823_correct_content_type_for_cadwork/upgrade.py
njohner: yes definitely. Sorry

PR #7451, opened 2022-04-29
Do not copy creator and creation date when transporting task documents.

njohner:
When closing a cross-client "information" task, documents can be copied to a dossier on the responsible client. In such cases, we also copied over the "creator" and "created" to the copied document. This is not consistent with normal copy of a document and is a bit confusing as in the journal it says that the document was added by the current user, whereas the creator is still the creator of the original document.

No upgrade step to fix existing such documents. Also I corrected this only for copying documents when closing a task and not generally in the transporter, as I think in most cases it makes sense to copy the Creator and creation date.

For [CA-4096]

## Checklist
- [x] Changelog entry

[CA-4096]: https://4teamwork.atlassian.net/browse/CA-4096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-4069.feature
+++ b/changes/CA-4069.feature
@@ -1,0 +1,1 @@
+Do not copy creator and creation date when transporting task documents. [njohner]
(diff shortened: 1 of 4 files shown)


phgross (approved):
👍

PR #7453, opened 2022-05-04
Fix moving an object with the api where not all path elements are accessible.

phgross:
Use our own `get_obj_by_path` in the `@move` endpoint to fix a traversal bug, when not all path elements are accessible.

For [CA-4138]

## Checklist
- [x] Changelog entry

[CA-4138]: https://4teamwork.atlassian.net/browse/CA-4138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-4138.bugfix
+++ b/changes/CA-4138.bugfix
@@ -1,0 +1,1 @@
+Fix moving an object with the api where not all path elements are accessible. [phgross]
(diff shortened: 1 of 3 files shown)


njohner (approved):
LGTM, but just wondering: you are testing moving a document out of a dossier where not all path elements are accessible. Does moving a document into such a dossier also work?

phgross:
@njohner yes that works to. But in that case, the `get_obj_by_path` does not matter, because the `@move` endpoint is called on the protected dossier itself, so its the regular traversing mechanism.

PR #7456, opened 2022-05-06
Reader cannot copy or move items anymore.

njohner:
A reader should not be allowed to move items out of a dossier. This is especially an issue when Reader permission is granted through a "for information" task, where the user only gets "Read" access to certain documents and should not be allowed to move them out of the dossier. Because it's the same permission for copy, both are now forbidden, but that should not be an issue.

For [CA-3929]

## Checklist

- [x] Changelog entry
- Upgrade-Steps:

[CA-3929]: https://4teamwork.atlassian.net/browse/CA-3929?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/document/move.py
```diff
@@ -53,3 +53,5 @@ def validate_movement(self, target):
                 raise Forbidden(
                     _(u'msg_docs_cant_be_moved_from_inbox_to_private_repo',
                       u'Documents within the inbox cannot be moved to the private repository.'))
+
+        return super(DocumentMovabiliyChecker, self).validate_movement(target)
```
tinagerber, line 57: I would move `super(DocumentMovabiliyChecker, self).validate_movement(target)` to the top without return, then it needs the line only once, as an error is thrown if the movement is not valid

opengever/dossier/move_items.py
```diff
@@ -387,6 +387,7 @@ def validate_movement(self, target):
             raise NotReportedForbidden(
                 _(u'msg_would_exceed_max_dossier_level',
                   u'This would exceed maximally allowed dossier depth.'))
+        return super(DossierMovabiliyChecker, self).validate_movement(target)
```
tinagerber, line 390: Thus, subdossier may be moved to repository folders without Move permisssion.

tinagerber:
We have also overridden the `@move` endpoint for ToDos. The permission check should probably go in there as well.

opengever/document/move.py
njohner: I purposefully didn't, because I did not want to screw up error check order. The missing permission error is more generic than the ones checked before.

tinagerber: You're right about that, but if you're just not allowed to move in general, it doesn't matter why you wouldn't be allowed to otherwise.

opengever/dossier/move_items.py
njohner: oh right 😓

njohner:
For ToDos the permission is checked because it calls `reply` from the parent class https://github.com/4teamwork/opengever.core/blob/master/opengever/api/todo.py#L56, which calls the movability checker https://github.com/4teamwork/opengever.core/blob/master/opengever/api/move.py#L61
What I don't understand is why anyone has the `Copy or Move` permission on the side of workspaces... but they do

tinagerber (approved):
https://github.com/4teamwork/ftw.lawgiver/blob/e7eadc4feef36af94e56c8fe94e0d1e931971a0a/ftw/lawgiver/lawgiver.zcml#L252 🤔 Is apparently wrong for all lawgiver workflows

opengever/api/copy_.py
```diff
@@ -33,6 +39,20 @@ def reply(self):
...
+
+        if not isinstance(source, list):
+            source = [source]
+
+        for item in source:
+            obj = self.get_object(item)
+            if not api.user.has_permission("Copy or Move", obj=obj):
+                raise NotReportedForbidden(
+                    _('copy_object_disallowed',
+                      default=u'You are not allowed to copy this object'))
```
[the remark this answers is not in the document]
njohner: I think it's fine as ist. All errors raised by the movability checker also don't give the object's title. Title is added only for the classic UI in the handling of the errors in the form. The frontend will actually do one request per object being moved and display the error in the task queue, which already contains the title of the object. So I'd leave it as is.

opengever/base/locales/de/LC_MESSAGES/opengever.base.po
```diff
@@ -463,6 +464,11 @@ msgstr "Information"
 msgid "message_title_warning"
 msgstr "Warnung"
 
+#. Default: "You are not allowed to move this object"
+#: ./opengever/base/adapters.py
+msgid "move_object_disallowed"
+msgstr "Sie sind nicht berechtigt dieses Objekt zu verschieben."
```
njohner, line 470: so here too `Sie sind nicht berechtigt, dieses...`

opengever/document/move.py
```diff
@@ -42,7 +44,7 @@ def validate_movement(self, target):
                 raise Forbidden(
                     _(u'msg_docs_cant_be_moved_from_repo_to_private_repo',
                       u'Documents within the repository cannot be moved to the private repository.'))
-            return
+            return super(DocumentMovabiliyChecker, self).validate_movement(target)
```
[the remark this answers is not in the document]
njohner: oh thanks

njohner:
Just to clarify concerning the `Copy or Move` permission on teamraum side: Basically that permission is granted to `Anonymous` on the Zope level and then inherited down, somehow that translates to `Authenticated` having that permission. I've had a look with philippe and we couldn't figure out where that comes from. But in the end it's a bit of an issue, because the `WorkspaceGuest`s have the permission, so they see the actions but the actions will fail because they don't have write permissions. I'll make an issue to have a look

njohner:
Here the follow-up issue: https://4teamwork.atlassian.net/browse/CA-4173

PR #7457, opened 2022-05-09
Remove flaky test.

njohner:
The test to copy a document from the second batch from a workspace does not make sense, as there is no such thing as a document from the second batch (the search is not sorted). So this test was flaky and I don't see any way to make a meaningful test for this. See https://ci.4teamwork.ch/builds/498333/tasks/936230

For [CA-3929]

## Checklist
- [x] Changelog entry

[CA-3929]: https://4teamwork.atlassian.net/browse/CA-3929?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-3929.other
+++ b/changes/CA-3929.other
@@ -1,0 +1,1 @@
+Remove flaky test. [njohner]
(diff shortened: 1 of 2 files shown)


tinagerber (approved):
You could also test if you can copy all 26 documents, but that probably doesn't make much more sense either.

njohner:
Yeah, true, but I think it's fine not to test that.

PR #7462, opened 2022-05-10
Excel mass export for documents and dossiers

elioschmutz:
This PR changes the report-view:

- it can now export more than 1000 objects. We do no longer define a limit
- it is now possible to choose selected items by passing a listing-name and the corresponding filter/sorting params instead of explicitly define them by paths

Bildschirmfoto 2022-05-11 um 21 21 19

- no longer sort the results locally when using paths. Use term-boosting to directly sort the results in solr instead: https://solr.apache.org/guide/6_6/the-standard-query-parser.html since we use a generator for the selected_items.

For [CA-2609]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-2609]: https://4teamwork.atlassian.net/browse/CA-2609?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-2609.feature
+++ b/changes/CA-2609.feature
@@ -1,0 +1,1 @@
+Excel export: Remove limit of 1000 items. [elioschmutz]
--- a/opengever/base/browser/reporting_view.py
+++ b/opengever/base/browser/reporting_view.py
@@ -5,6 +5,7 @@ from ftw.tabbedview.interfaces import IGridStateStorageKeyGenerator
 from opengever.base import _
 from opengever.base.behaviors.utils import set_attachment_content_disposition
 from opengever.base.reporter import DATE_NUMBER_FORMAT
+from opengever.base.solr import batched_solr_results
 from opengever.base.solr import OGSolrDocument
 from opengever.base.solr.fields import DateListingField
 from plone import api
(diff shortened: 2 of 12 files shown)


opengever/base/browser/reporting_view.py
```diff
@@ -112,17 +113,23 @@ def __init__(self, *args, **kwargs):
 
     def get_selected_items(self):
         paths = self.request.get('paths')
-
-        if not paths:
+        listing_name = self.request.get('name')
```
phgross, line 116: I would name the request parameter something more specific than `name`. Maybe `listing` or `listing_name`

changes/CA-2609.feature
```diff
@@ -0,0 +1 @@
+Excel export: Remove limit of 1000 items [elioschmutz]
```
phgross, line 1:
```suggestion
Excel export: Remove limit of 1000 items. [elioschmutz]
```

phgross (approved):
LGTM 👍 besides the things mentioned.

opengever/base/browser/reporting_view.py
elioschmutz: 👍 renamed to `listing_name`

PR #7467, opened 2022-05-17
Incresed listing performance when listing objects with empty fields

elioschmutz:
This PR fixes a performance issue in listings (and all kind of listing related actions, i.e. export listing items to .xlsx):

Solr does not return fields if they are empty, even if we request them. The current implementation: https://github.com/4teamwork/opengever.core/blob/5854d756bd2bb693d0967322ba3be21e8ccd2baa/opengever/base/contentlisting.py#L42 checks for the attribute-existence on the solr-doc. If the attribute does not exist, it will get the object and checks the existence on the object itself. 

This behavior is not desired when dealing with solr-docs. We know the fields and if it does not exist on the doc, it should just raise an attribute error without retrieving the object.

## Before
https://user-images.githubusercontent.com/557005/168813416-20a4d077-d886-4476-a68a-c630a5b44362.mov

## After
https://user-images.githubusercontent.com/557005/168813402-af9fdbe1-0264-4ba1-a4da-e4d0fbbc99ae.mov

## Tesing
I don't know how we could write a useful test for this case. I could mock an object and check if `getObject` have been called... but is 

For [CA-2609]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations


phgross (approved):
👍 LGTM

tinagerber:
What about the search? Could it possibly lead to a problem there?

elioschmutz:
@tinagerber good point. We usually do not show proposals of type `SubmittedProposal` in the search. But there seems to be another bug in general:

The livesearch will find them, but not the advanced search:

**Livesearch**
Bildschirmfoto 2022-05-19 um 10 54 48

**Advanced search**
Bildschirmfoto 2022-05-19 um 10 56 22

But this is not a bug related to my change. This is the current behavior. So it should be save for now to merge this one. I'll create a bug-report for this.

elioschmutz:
https://4teamwork.atlassian.net/browse/CA-4224

PR #7475, opened 2022-06-02
Add GitHub Action to build Docke image

buchi:
Builds a multiplatform Docker image and pushes it to Docker Hub.

For [CA-3995](https://4teamwork.atlassian.net/browse/CA-3995)

## Checklist

- [ ] Changelog entry


docker/core/requirements-core.txt
```diff
@@ -39,7 +39,7 @@ decorator==3.4.2
 diazo==1.1.3
 DocumentTemplate==2.13.6
 docutils==0.12
-docxcompose==1.3.3
+docxcompose==1.3.4
```
njohner, line 42: We probably would need to do this every time we update something in `versions.cfg`? I was not aware of that

njohner (approved):
Well, I can't really tell, but LGTM.

docker/core/requirements-core.txt
buchi: Yes, luckily there are not a lot of changes. I'm trying to keep an eye on this.

PR #7477, opened 2022-06-07
Fix vocabulary used for sharing Workspace content

njohner:
The `ActualWorkspaceMember` source, vocabulary and endpoint did not make much sense, as they basically returned the list of participants on a workspace. This was done by walking up the acquisition chain, but starting at the workspace and stopping at the portal, so basically just gathering the participants of the workspace (and workspace root, but that does not have any participants). Instead these vocabularies were always meant to be context dependent and walk up the acquisition chain. They are used for `ToDo` and `WorkspaceMeeting` responsibles as well as for sharing content from a workspace content. For `ToDo` and `WorkspaceMeeting` responsibles, things worked as expected as they are only allowed in the Workspace and not inside `WorkspaceFolder`s. The sharing itself was broken on other contents as the vocabulary basically allowed choosing all workspace participants, regardless of whether inheritance is blocked for example on the `WorkspaceFolder`. This used to work but got broken at some point, probably because the naming seemed to indicate the intent was to return the Workspace members. I have therefore also renamed the vocabulary, source and endpoint.

For [CA-3687]

## Checklist
- [x] Changelog entry
- API change:
  - If breaking:
    - [x] api-change label added

[CA-3687]: https://4teamwork.atlassian.net/browse/CA-3687?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/ogds/base/sources.py
```diff
@@ -645,9 +647,16 @@ def search_query(self):
...
 
     def _extend_query_with_workspace_filter(self, query):
-
-        userids = list(get_context_user_members_ids(self.context))
-        groupids = list(get_context_group_members_ids(self.context))
+        # When creating content, members need to be determined on the parent
+        # as the roles giving view permissions cannot be determined on the
+        # object being created.
+        request = getRequest()
+        if request.method == "POST" and request.getHeader("Content-Type") == "application/json":
```
tinagerber, line 654: This breaks the implementation with the `notify_all` parameter, since the `@share-content` endpoint is also a POST request.

tinagerber (changes requested):
LGTM, except for the content creation fix

njohner:
Thanks for the review, I now fixed the issue with `notify_all` and added a test.

PR #7478, opened 2022-06-08
Support importing dossier participations in bundle import.

phgross:
This adds support for an additional key `_participations` in the dossier bundle import which allows to import dossier participations. The participant ids and roles are not validated, this gives us more flexibility during migrations. Moreover, GEVER can handle Ids that no longer exist or are incorrect. 

For [CA-4037]

## Checklist
- [x] Changelog entry

[CA-4037]: https://4teamwork.atlassian.net/browse/CA-4037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/bundle/sections/participations.py
+++ b/opengever/bundle/sections/participations.py
@@ -1,0 +1,61 @@
+from collective.transmogrifier.interfaces import ISection
+from collective.transmogrifier.interfaces import ISectionBlueprint
+from collective.transmogrifier.utils import defaultMatcher
+from opengever.dossier.behaviors.participation import IParticipationAware
+from opengever.dossier.behaviors.participation import IParticipationAwareMarker
+from zope.interface import classProvides
+from zope.interface import implements
+import logging
+
+
+logger = logging.getLogger('opengever.setup.sections.set_dates')
+PARTICIPATIONS_KEY = '_participations'
+
+
+class ParticipationSection(object):
+    """Transmogrifier section which adds dossier participations.
+    """
+
+    implements(ISection)
+    classProvides(ISectionBlueprint)
+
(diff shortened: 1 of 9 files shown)


opengever/bundle/sections/participations.py
```diff
@@ -0,0 +1,60 @@
...
+            obj = self._get_obj(item)
+            participations = item.get(PARTICIPATIONS_KEY, [])
+            if participations:
+                self.create_participations(obj, participations)
+            yield item
+
+    def create_participations(self, obj, participations):
+        if not IParticipationAwareMarker.providedBy(obj):
+            logger.warn(u'Participations for object {} ignored, participation '
+                        'not supported for this type.'.format(obj))
```
njohner, line 41: I guess you should `return` here

opengever/dossier/participations.py
```diff
@@ -87,9 +87,10 @@ def validate_participant(self, participant_id):
             raise InvalidParticipantId(
                 u"{} is not a valid id".format(participant_id))
 
-    def add_participation(self, participant_id, roles):
-        self.validate_participant(participant_id)
-        self.validate_roles(roles)
+    def add_participation(self, participant_id, roles, validate=True):
```
njohner, line 90: Can you also add the validate parameter for the `SQLParticipationHandler`? I think it makes sense that all handlers support the same parameters.

njohner (approved):
Just some comments, otherwise LGTM

opengever/dossier/participations.py
phgross: ah yes, thanks for the hint

PR #7482, opened 2022-06-09
Always upload blobs to Solr from testserver

buchi:
Since https://github.com/4teamwork/opengever.core/pull/7427 we run Solr with Docker.
Because of that, blobs always have to be uploaded to Solr. But the testserver did not upload blobs if the Solr hostname was localhost.

## Checklist

- [ ] Changelog entry


njohner (approved):
LGTM

PR #7483, opened 2022-06-16
Support importing OGDS Users in a bundle import

phgross:
With this PR the bunde import now also provide the possibility to import SQL objects, currently only OGDS Users are supported. But I implemented a generic way, so that support for other SQL Objects could be easily added.

Used for [CA-4038], where the goal is to import old, inactive users into the ogds.

## Checklist
- [x] Changelog entry

[CA-4038]: https://4teamwork.atlassian.net/browse/CA-4038?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/bundle/sections/resolveguid.py
+++ b/opengever/bundle/sections/resolveguid.py
@@ -3,6 +3,7 @@ from collective.transmogrifier.interfaces import ISection
 from collective.transmogrifier.interfaces import ISectionBlueprint
 from opengever.base.interfaces import IReferenceNumberFormatter
 from opengever.base.interfaces import IReferenceNumberSettings
+from opengever.base.schemadump.config import GEVER_SQL_TYPES
 from opengever.base.schemadump.config import PARENTABLE_TYPES
 from opengever.base.schemadump.config import ROOT_TYPES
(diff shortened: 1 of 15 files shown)


opengever/bundle/sections/resolveguid.py
```diff
@@ -212,11 +213,14 @@ def build_tree(self):
             elif parent_reference is not None:
                 roots.append(item)
 
-            elif item['_type'] in ROOT_TYPES:
+            elif item['_type'] in ROOT_TYPES or item['_type'] in GEVER_SQL_TYPES:
                 # Repo roots and workspace roots are the only types
```
njohner, line 217: Also adapt the comment here to include ogds users

opengever/bundle/sections/resolveguid.py
```diff
@@ -212,11 +213,14 @@ def build_tree(self):
...
                 # without a parent pointer
                 roots.append(item)
```
njohner, line 219: Why to we need the item in the roots?

njohner (approved):
LGTM

PR #7486, opened 2022-06-21
Optimize groups API

buchi:
The previous implementation performed terribly with a large amount of group members (e.g. 10'000 members).
We now avoid fetching the user object for each group member which greatly improves the performance.

Further we also allow to recreate deleted local groups.

For [CA-1039](https://4teamwork.atlassian.net/browse/CA-1039)

## Checklist

- [x] Changelog entry

### DIFF
--- a/opengever/api/serializer.py
+++ b/opengever/api/serializer.py
@@ -32,6 +32,7 @@ from opengever.repository.interfaces import IRepositoryFolder
 from plone import api
 from plone.dexterity.interfaces import IDexterityContainer
 from plone.dexterity.interfaces import IDexterityContent
+from plone.restapi.batching import HypermediaBatch
 from plone.restapi.interfaces import IFieldSerializer
 from plone.restapi.interfaces import IJsonCompatible
 from plone.restapi.interfaces import ISerializeToJson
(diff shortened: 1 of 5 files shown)


opengever/api/serializer.py
```diff
@@ -341,6 +344,36 @@ def __call__(self):
...
+        # Get group member ids in a more performant way than in plone.restapi
+        # which calls `group.getGroupMemberIds()` which is incredibly slow
+        # for a larger amount of group members.
+        gtool = api.portal.get_tool(name='portal_groups')
+        members = gtool.getGroupMembers(self.context.getId())
+        batch = HypermediaBatch(self.request, members)
+        users_data = {
+            "@id": batch.canonical_url,
+            "items_total": batch.items_total,
+            "items": sorted(batch),
```
njohner, line 371: Does sorting the batch work? Shouldn't you sort `members` before making the batch?

opengever/api/serializer.py
```diff
@@ -341,6 +344,36 @@ def __call__(self):
...
+            "email": group.getProperty("email"),
+            "title": group.getProperty("title"),
+            "description": group.getProperty("description"),
+            "roles": group.getRoles(),
+        }
+
+    def member_data(self):
+        # Get group member ids in a more performant way than in plone.restapi
+        # which calls `group.getGroupMemberIds()` which is incredibly slow
+        # for a larger amount of group members.
```
njohner, line 364: I see we basically avoid fetching the users, so that we do not check whether the member exists or not. 👍

njohner (approved):
LGTM

opengever/api/serializer.py
buchi: Yes it should. `HypermediaBatch` is an iterable as it implements an `__iter__()` method and `sorted` works for iterables. Btw. the code is more or less copied from plone.restapi with the main difference that the group member ids are fetched directly without getting the member object for each id.
https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/serializer/group.py#L48

PR #7487, opened 2022-06-21
Add possibility to bundle import to set obj creator and actor for journal entries 

phgross:
The original goal of the story was to be able to define the creator of journal entries during a bundle import. We decided to extend this, so that the creator can also be set. It would be confusing if the listing under creator would say something different than in the journal.

For [CA-4039]

## Checklist
- [x] Changelog entry

[CA-4039]: https://4teamwork.atlassian.net/browse/CA-4039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/bundle/sections/manual_journal_entries.py
+++ b/opengever/bundle/sections/manual_journal_entries.py
@@ -1,0 +1,25 @@
+from collective.transmogrifier.interfaces import ISection
+from collective.transmogrifier.interfaces import ISectionBlueprint
+from opengever.journal import ManualJournalActor
+from zope.interface import classProvides
+from zope.interface import implements
+
+
+class ManualJournalEntrySection(object):
+    """Add request flag to create journal entries with the items creator
+    if given in the bundle.
+    """
+
+    classProvides(ISectionBlueprint)
+    implements(ISection)
+
+    def __init__(self, transmogrifier, name, options, previous):
+        self.previous = previous
+
+    def __iter__(self):
+        for item in self.previous:
+            if item.get('_creator'):
+                with ManualJournalActor(item.get('_creator')):
+                    yield item
+            else:
+                yield item
--- a/changes/CA-4039.feature
+++ b/changes/CA-4039.feature
@@ -1,0 +1,1 @@
+Set creator during bundle import and use creator as journal entry actor. [phgross]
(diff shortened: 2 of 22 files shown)


opengever/bundle/sections/manual_journal_entries.py
```diff
@@ -0,0 +1,25 @@
...
+    classProvides(ISectionBlueprint)
+    implements(ISection)
+
+    def __init__(self, transmogrifier, name, options, previous):
+        self.previous = previous
+
+    def __iter__(self):
+        for item in self.previous:
+            if item.get('_creator'):
+                with ManualJournalActor(item.get('_creator')):
```
elioschmutz, line 22: What kind of journal entries will be created during the bundle-import?

With this change every journal-entry would be made by the `_creator`. I think, only the "create" journal-entry should be made by this user. The remaining entries should be created by the current user (zopemaster?). 

Otherwise, we could potentially add entries in the name of a user who was not responsible for a specific change. Depending on the change, this could be really bad.

elioschmutz (commented):
👍 

Please take a closer look at the `ManualJournalActor`. I think it should only be `ManualJornalCreatorActor`...

opengever/bundle/sections/manual_journal_entries.py
phgross: During creation of an object are several journal entries created. Here an example of a dossier :
Bildschirmfoto 2022-06-24 um 11 47 12

The only goal of a bundle import is to create objects. IMHO those additional actions (like seting dossier metadata or participations) are part of the "creation" and should therefore also be made in the name of the creator. The creator of a dossier has not only created the dossier cover, but also defined those metadata and participations. So it would be strange if he has only created the dossier or not?

In my first commit only the creation entry, was made in the name of the creator, but that seemed to be wrong for me. Maybe a quick 📞 ?

elioschmutz (approved):
As discussed per phone-call:

It's ok as it is. The workflow-transitions will not be logged, thus, there will be no issue with "not-allowed-transitions" by a specific user.

PR #7489, opened 2022-06-23
Add test that filing number is serialized for dossiers.

njohner:
Filing number is already serialized when the feature is enabled and the filing number is set. So we just add a test.

For [CA-4374]

## Checklist
- [] Changelog entry -> no CL for that


phgross (approved):
👍

PR #7490, opened 2022-06-23
Solr: language specific indexing and use of eDisMax query parser

buchi:
Enable language detection and index text fields with language specific analyzers. In general we perform stemming to avoid using wildcards in queries and to be able to switch to the eDisMax parser.

Fields are still indexed using the canonical name. The `langid` update processor then detects the language and appends the language suffix to field name. However querying needs to be done against the field with the language suffix.

Using language specific fields in combination with atomic updates can result in outdated index data if another language is detected during reindexing. To avoid this, we delete all language specific fields before indexing the canonical field. This is handled in ftw.solr: https://github.com/4teamwork/ftw.solr/pull/189

We switch to the eDisMax parser instead of the default Lucene parser. Instead of construction a query with ftw.solr's `make_query` we pass the query unmodified to Solr.

We enforce the eDisMax parser in Solr so we don't have to this everywhere we make search requests. All other eDisMax configuration is done in Solr as well.

When installing Solr with Buildout, a new version of ftw.recipe.solr is needed, which support subdirectories in the Solr conf directory: https://github.com/4teamwork/ftw.recipe.solr/pull/19

Switching to language specific indexing requires reindexing of Solr. This can be done with the DIH (Data Import Handler) from Solr to Solr. Support to force reindexing even though the Solr major version didn't change is added in: https://github.com/4teamwork/opengever-deployments/pull/186

For [CA-4241](https://4teamwork.atlassian.net/browse/CA-4241)

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations


solr-conf/managed-schema
```diff
@@ -313,6 +328,111 @@
...
+        <!-- Case insensitive stop word removal.
+        -->
+        <filter class="solr.StopFilterFactory"
+                ignoreCase="true"
+                words="lang/stopwords_en.txt"
+                />
+        <filter class="solr.LowerCaseFilterFactory"/>
+        <filter class="solr.EnglishPossessiveFilterFactory"/>
+        <filter class="solr.KeywordMarkerFilterFactory" protected="protwords.txt"/>
+        <!-- Optionally you may want to use this less aggressive stemmer instead of PorterStemFilterFactory:
```
phgross, line 377: ✂️ ?

solr-conf/managed-schema
```diff
@@ -313,6 +328,111 @@
...
+        <tokenizer class="solr.StandardTokenizerFactory"/>
+        <filter class="solr.SynonymGraphFilterFactory" synonyms="synonyms.txt" ignoreCase="true" expand="true"/>
+        <filter class="solr.StopFilterFactory"
+                ignoreCase="true"
+                words="lang/stopwords_en.txt"
+                />
+        <filter class="solr.LowerCaseFilterFactory"/>
+        <filter class="solr.EnglishPossessiveFilterFactory"/>
+        <filter class="solr.KeywordMarkerFilterFactory" protected="protwords.txt"/>
+        <!-- Optionally you may want to use this less aggressive stemmer instead of PorterStemFilterFactory:
```
phgross, line 392: ✂️ ?

solr-conf/solrconfig.xml
```diff
@@ -756,6 +756,31 @@
...
       -->
+    <lst name="invariants">
+      <!-- Enforce eDisMax parser-->
+      <str name="defType">edismax</str>
+      <!-- Query fields -->
+      <str name="qf">
+          Title_de^10 Title_en^10 Title_fr^10 Title_general^10
+          Description_de^5 Description_en^5 Description_fr^5 Description_general^5
+          SearchableText_de SearchableText_en SearchableText_fr SearchableText_general
+          sequence_number_string^1000 metadata^5
```
phgross, line 767: Should the `reference` field not also part of the queried fields with a rather high weighting and probably also in the phrase fields?

phgross (commented):
LGTM, but the following things are unclear to me:
* How to we guarantee reindexing the data when updating to this gever version? As far as i know the `reindex-solr` script only checks for a solr version change.
* IMHO the `@solrsearch` endpoint will change how it works, we should add a note in the API changelog.
* We now use only the eDisMax query parser, does that have an effect for the listing text filter? Do we need to change/adapt the gever-ui?

solr-conf/managed-schema
```diff
@@ -313,6 +328,111 @@
...
+         cross-language defaults: it tokenizes with StandardTokenizer,
+   removes stop words from case-insensitive "stopwords.txt"
+   (empty by default), and down cases.  At query time only, it
+   also applies synonyms. -->
+    <fieldType name="text_general" class="solr.TextField" positionIncrementGap="100">
+      <analyzer type="index">
+        <charFilter class="solr.MappingCharFilterFactory" mapping="mapping-FoldToASCII.txt"/>
+        <tokenizer class="solr.StandardTokenizerFactory"/>
+        <filter class="solr.StopFilterFactory" ignoreCase="true" words="stopwords.txt" />
+        <!-- in this example, we will only use synonyms at query time
```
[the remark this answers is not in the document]
buchi: I would keep those. It's copied from the Solr examples and may help in further tuning...

solr-conf/solrconfig.xml
buchi: It wasn't so far, but maybe makes sense.

phgross: yes I would

phgross (approved):
👍 LGTM

please add a note to https://4teamwork.atlassian.net/wiki/spaces/4TEAM/pages/123077675/GEVER+Releases+Versions-Spezifische+Punkte that solr-reindexing is required for the upcoming 2022.13 release.

PR #7491, opened 2022-06-24
The @journal endpoint can now be filtered and searched.

elioschmutz:
This PR extends the `@journal` to make it searchable and filterable.

For [CA-4301]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-4301]: https://4teamwork.atlassian.net/browse/CA-4301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-4301.feature
+++ b/changes/CA-4301.feature
@@ -1,0 +1,1 @@
+The @journal endpoint can now be filtered and searched. [elioschmutz]
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -11,6 +11,8 @@ Breaking Changes
 
 Other Changes
 ^^^^^^^^^^^^^
+- ``@journal``: Provides filtering and searching.
+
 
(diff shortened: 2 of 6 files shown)


elioschmutz:
> LGTM, but I wonder how that will perform. I tried to have a look and it seems that dossiers usually do not have that many journal entries, so it should be fine. But we do make like 3 copies of the list of journal entries hier before returning. Maybe we could use some `yield` here and there instead.
> 
> But anyway, I think we will have another look if we get performance issues, so 👍

I think there should be no performance issues.

It is a fixed list coming from the annotations. We already have had reversed the list. So I just iter one more time over a list without any heavy computations. So it should easily handle thousands of entries without any performance issues.

PR #7494, opened 2022-06-28
Add new @transfer-number PATCH endpoint.

njohner:
For [CA-4354]

## Checklist

- [x] Changelog entry
- API change:
  - If breaking:
    - [ ] api-change label added
- New translations

### DIFF
--- a/opengever/api/tests/test_disposition.py
+++ b/opengever/api/tests/test_disposition.py
@@ -3,6 +3,7 @@ from ftw.testbrowser import browsing
 from ftw.testing import freeze
 from opengever.disposition.delivery import DeliveryScheduler
 from opengever.disposition.interfaces import IAppraisal
+from opengever.disposition.interfaces import IDisposition
 from opengever.testing import IntegrationTestCase
 from plone import api
 import json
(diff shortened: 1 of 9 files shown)


opengever/api/tests/test_disposition.py
```diff
@@ -351,3 +352,114 @@ def test_raises_if_uid_is_not_part_of_disposition(self, browser):
              'createtreatydossiers000000000001 is not part of the disposition',
              'type': 'BadRequest'},
             browser.json)
+
+
+class TestDispositionPatch(IntegrationTestCase):
+
+    @browsing
+    def test_only_archivist_can_patch_transfer_number(self, browser):
```
elioschmutz, line 360: Is the test-title correct? Don't you test a general patch-request for dispositions?

elioschmutz (approved):

opengever/api/tests/test_disposition.py
njohner: I'd say it's correct. If you look closely, the PATCH done with the records manager only modifies the title, while the transfer_number remains unchanged. Then the second PATCH done with the archivist actually modifies the transfer_number.

PR #7496, opened 2022-06-28
Improve DELETE @participations error handling

tinagerber:
If an error occurs on a subfolder when removing a workspace participation, this is now visible in the error message.
In addition, a subscriber can no longer be removed if the participant to be removed is authorized on a subfolder on which the user attempting to remove the participant is not authorized.

For [CA-3708]

## Checklist

- [x] Changelog entry

- API change:


opengever/api/participations.py
```diff
@@ -149,15 +150,29 @@ def read_params(self):
...
         current context.
         """
         if obj.get_context_with_local_roles() == obj and self.find_participant(token, obj):
             manager = ManageParticipants(obj, self.request)
-            manager._delete(Actor.lookup(token).actor_type, token)
+            try:
+                manager._delete(Actor.lookup(token).actor_type, token)
+            except Unauthorized:
+                raise BadRequest(
+                    _(u'msg_cannont_manage_member_in_subfolder',
```
njohner, line 165: cannont -> cannot

opengever/api/participations.py
```diff
@@ -158,9 +159,10 @@ def recursive_delete_participation(self, obj, token):
             manager = ManageParticipants(obj, self.request)
             manager._delete(Actor.lookup(token).actor_type, token)
 
-        for folder in obj.listFolderContents(
-                contentFilter={'portal_type': 'opengever.workspace.folder'}):
-            self.recursive_delete_participation(folder, token)
+        with elevated_privileges():
+            folders = obj.listFolderContents(contentFilter={'portal_type': 'opengever.workspace.folder'})
+        for folder in folders:
+            self.recursive_delete_participation(folder, token, depth + 1)
```
njohner, line 180: `depth + 1` should be in the second commit, but that might be a pain to rebase correctly, so you can leave as is ;-)

njohner (approved):
LGTM

opengever/api/participations.py
tinagerber: It was feasible 😄

PR #7498, opened 2022-07-04
Editable journal entries

elioschmutz:
🚧 

For [CA-4298]

## Checklist

- [ ] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations


opengever/api/tests/test_journal.py
```diff
@@ -344,3 +344,180 @@ def test_raises_without_edit_permission(self, browser):
...
+
+        manager.add_manual_entry('information', 'is an agent')
+
+        entry = browser.open(
+            self.dossier.absolute_url() + '/@journal',
+            method='GET',
+            headers=http_headers(),
+        ).json.get('items')[-1]
+
+        with browser.expect_http_error(500):
```
njohner, line 494: I don't think that should be a 500.

opengever/api/tests/test_journal.py
```diff
@@ -344,3 +344,180 @@ def test_raises_without_edit_permission(self, browser):
...
```
njohner, line 517: Here too, a BadRequest or so would be better

opengever/journal/manager.py
```diff
@@ -64,6 +69,33 @@ def count(self):
...
 
+    def lookup(self, entry_id):
+        for journal_entry in self.list():
+            if journal_entry.get('id') == entry_id:
+                return journal_entry
+
+        raise AttributeError("Entry '{}' does not exist.".format(entry_id))
+
+    def remove(self, entry_id):
+        self.list().remove(self.lookup(entry_id))
```
njohner, line 80: We should only be allowed to remove manual entries

njohner (commented):
Overall looks good, but a few minor things. Importantly you need to prevent deleting automatic journal entries. This is essential for traceability in Gever.

elioschmutz:
@njohner thanks for the review. I pushed again in the meantime and changed the `is_manual_entry` to `is_editable`. The backend should tell the frontend which items are editable. This should not be part of the frontend. What do you think about this?

I'll go through your review and will update it soon 👍

njohner:
Yes I've seen. 👍

opengever/api/tests/test_journal_manager.py
```diff
@@ -116,3 +118,45 @@ def test_remove_invalid_entry_will_raise(self, browser):
...
+    def test_can_modify_entry(self, browser):
+        self.login(self.regular_user, browser)
+
+        manager = JournalManager(self.dossier)
+        manager.clear()
+
+        manager.add_manual_entry('information', 'is an agent')
+        entry = manager.list()[0]
+
+        self.assertEqual(u'is an agent', entry.get('comments'))
```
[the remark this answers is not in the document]
njohner: ah ok sorry 👍

opengever/api/tests/test_journal.py
```diff
@@ -278,3 +283,273 @@ def test_can_search_by_title(self, browser):
...
+
+    @browsing
+    def test_raises_when_id_does_not_exist(self, browser):
+        self.login(self.regular_user, browser=browser)
+
+        with browser.expect_http_error(404):
+            url = '{}/@journal/{}'.format(self.dossier.absolute_url(), 'invalid')
+            browser.open(url, method='DELETE',
+                         headers={'Accept': 'application/json'})
+
```
njohner, line 356: Could you add a test `test_raises_when_deleting_a_non_manual_journal_entry(self, browser)`

PR #7504, opened 2022-07-06
Journal entry refactoring

elioschmutz:
This PR refactors the journal entry handling implementation by:

- introducing a base `JournalService` class which provides general functions for all journal endpoints
- introducing a `JournalManager` class which handles the journal entries in the annotations by providing useful functions to `add`, `list`, `count` entries. It also provides methods to add auto-/manual journal entris. Later on, this class will be extended with a `remove` and `update` method to remove or update journal entries.
- remove the unused `ManualJournalEntry` class

This refactoring makes it much easier to implement the upcoming new feature to remove and update manual journal entries.

For [CA-4298]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-4298]: https://4teamwork.atlassian.net/browse/CA-4298?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/journal/entry.py
```diff
@@ -18,88 +18,111 @@
...
-    def get_category_title(self):
-        voca_factory = getUtility(
-            IVocabularyFactory,
-            name='opengever.journal.manual_entry_categories')
-
-        return voca_factory(self.context).getTerm(self.category).title
+class JournalEntry(object):
+    @classmethod
+    def create():
+        raise NotImplementedError
```
njohner, line 91: instatiate the error

opengever/journal/entry.py
```diff
@@ -18,88 +18,111 @@
...
+
+
+class AutoJournalEntry(JournalEntry):
```
njohner, line 94: I personally don't like that refactoring. You make new classes, that can't be instantiated and have a single method. The only advantage is that they inherit from `JournalEntry`, but they actually don't inherit anything, and moreover the `create` method does not have the same signature for the two subclasses. I would either:
- delete these 3 classes and simply have two functions `create_manual_journal_entry` and `create_auto_journal_entry`
- Move some code into the `JournalEntry` base class, for example
```
class JournalEntry(object):
    def serialize_documents():
         .....

    def create(data):
        entry_obj = self.prepare_data(data)
        notify_journal_event(entry_obj)
   
    def prepare_data(self, data).
        raise NotImplementedError()
```

njohner (commented):
I'm not against some refactoring, but not a big fan of this one. Maybe a quick call?

elioschmutz:
@njohner @phgross I rewrote the PR and was able to completely remove the `ManualJournalEntry` class. We now only have the newly introduced `JournalManager`(@njohner do you still prefer `JournalStorage`?).

@njohner removing the `ftw.journal` seems to be a good idea. It really makes it just more complex than it is. But as discussed, I'd prefer to finish first the feature before I try to go further with refactoring.

opengever/journal/manager.py
```diff
@@ -0,0 +1,28 @@
+from ftw.journal.config import JOURNAL_ENTRIES_ANNOTATIONS_KEY
+from opengever.journal.entry import ManualJournalEntry
+from zope.annotation.interfaces import IAnnotations
+
+
+DEFAULT_COMMENT_CATEGORY = 'information'
+
+
+class JournalManager(object):
```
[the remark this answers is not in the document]
elioschmutz: Completely removing the `ftw.journal` would be great, but I think it's not as easy as it looks like. I think there will be edge cases we do not see right now. Removing it in this PR seems to be a too big change for what we need here.

Extending `ftw.journal` with this functionality would be possible but do we really want it? 

Should we talk once again about what we really want? And where?

opengever/api/journal.py
```diff
@@ -141,9 +139,7 @@ def _item_title(self, item):
         return translate(item.get('action').get('title'), context=self.request)
 
     def _data(self):
-        annotations = IAnnotations(self.context)
-        items = deepcopy(annotations.get(JOURNAL_ENTRIES_ANNOTATIONS_KEY, []))
-        return items
+        return deepcopy(JournalManager(self.context).list())
```
njohner, line 142: I wonder whether the `JournalManager` should not always return a copy of the journal entries in the `list` method, or do we use the `list` method to then modify it anywhere?

njohner (approved):
LGTM

njohner:
> @njohner @phgross I rewrote the PR and was able to completely remove the `ManualJournalEntry` class. We now only have the newly introduced `JournalManager`(@njohner do you still prefer `JournalStorage`?).
> 
> @njohner removing the `ftw.journal` seems to be a good idea. It really makes it just more complex than it is. But as discussed, I'd prefer to finish first the feature before I try to go further with refactoring.

- `JournalManager` is fine. Storage would have been better if we moved all of this into `ftw.journal`.
- As discussed, we do not want to invest too much time here, so we keep `ftw.journal` until the day we make a service out of it.

opengever/api/journal.py
elioschmutz: yes, we need it later to remove/update items from the list. I could implement a `def _list()` wich returns the original data and `def list()` which returns the `deepcopy`?

njohner: You can also leave as is, both are fine

PR #7505, opened 2022-07-07
Fix data_files

tinagerber:
Since we now have a subfolder in the solr-conf folder, the current method of including these files in the package no longer worked. This prevented the opengever.core egg from being installed.

For [CA-4082]
Follow up for #7490


buchi (approved):
👍🏻 LGTM

PR #7508, opened 2022-07-08
Journal api refactoring

elioschmutz:
This PR refactors hand harmonizes the `@journal` endpoint:

- We now use the `IFieldDeserializer` for every `IManualJournalEntry`-field, not only for the documents. This allows us to send the `category` as it comes from the `vocabulary`: `{ "token": "information" }`.
- The `POST` method expects a `comment` attribute, but the `GET` returns a `comments` attribute for each journal entry. The naming is now harmonized and will always be `comment`.

For [CA-4298]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [x] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-4298]: https://4teamwork.atlassian.net/browse/CA-4298?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/api/tests/test_journal.py
+++ b/opengever/api/tests/test_journal.py
@@ -23,8 +23,10 @@ class TestJournalPost(IntegrationTestCase):
     @browsing
     def test_add_journal_entry(self, browser):
         self.login(self.regular_user, browser)
-        payload = {'comment': 'example comment', 'category': 'information',
-                   'related_documents': [self.document.absolute_url()]}
+        payload = {
+            'comment': 'example comment',
+            'category': {'token': 'information'},
+            'related_documents': [self.document.absolute_url()]}
 
         browser.open(
             self.dossier.absolute_url() + '/@journal',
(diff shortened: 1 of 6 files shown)


opengever/api/tests/test_journal.py
```diff
@@ -96,23 +75,6 @@ def test_post_raises_when_document_lookup_failed(self, browser):
                 headers=http_headers(),
             )
 
-        self.assertEqual(
-            {u'message': u'Could not resolve object for UID=https://not-existing',
-             u'type': u'ValueError'},
-            browser.json)
-
-    @browsing
-    def test_add_journal_entry_via_api_is_xss_safe(self, browser):
```
njohner, line 105: is it not xss safe anymore?

tinagerber:
@elioschmutz Doesn't the intergever also use this endpoint?

tinagerber:
Ah, probably intergever only uses the POST endpoint

PR #7510, opened 2022-07-08
Bump ftw.upgrade to handle dead brains in security updates.

njohner:
For [CA-4112]

## Checklist

- [x] Changelog entry

[CA-4112]: https://4teamwork.atlassian.net/browse/CA-4112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


versions.cfg
```diff
@@ -105,7 +105,7 @@ ftw.testbrowser = 1.30.1
 ftw.testing = 1.20.2
 ftw.tika = 2.9.0
 ftw.tokenauth = 1.1.0
-ftw.upgrade = 3.3.0
+ftw.upgrade = 3.3.1
```
phgross, line 108: Docker image pinning should be updated as well.

phgross (changes requested):

versions.cfg
njohner: ah right thanks!

PR #7517, opened 2022-07-12
Improve perfomance in KuB contact source.

phgross:
With the current implementation, we had some serious performance issues, when accessing a large dossier listing with thousands of KuB-participations. Because the participation facet has led to thousands of separate kub-reqest (one separate request for each participations).

We solve the issue by using a separated, newly introduced KuB endpoint called `labels` (needs https://github.com/4teamwork/kub/pull/130), which returns the label for all contacts (people, organization and memberships) and create a mapping of those. This mapping will be cached for one hour.

For [CA-3978]

Needs  backport to LTS `2022.4` and `2022.11`

## Checklist

- [x] Changelog entry

[CA-3978]: https://4teamwork.atlassian.net/browse/CA-3978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/kub/client.py
+++ b/opengever/kub/client.py
@@ -1,6 +1,8 @@
 from opengever.base.sentry import log_msg_to_sentry
 from opengever.kub.interfaces import IKuBSettings
 from plone import api
+from plone.memoize import ram
+from time import time
 import requests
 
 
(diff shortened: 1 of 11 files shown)


opengever/kub/client.py
```diff
@@ -51,3 +53,18 @@ def get_by_id(self, _id):
 
     def get_resolve_url(self, _id):
         return u'{}resolve/{}'.format(self.kub_api_url, _id)
+
+    # Cache kub labels for an hour
+    @ram.cache(lambda *args: time() // (60 * 60))
+    def get_kub_id_label_mapping(self):
+        return self._fetch_kub_labels()
+
+    def _fetch_kub_labels(self):
```
njohner, line 62: The naming would rather suggest that this simply does `return self.session.get(u'{}labels'.format(self.kub_api_url))`. The rest would then go into `get_kub_id_label_mapping`. 
Otherwise you could also improve the naming here, maybe something like `_prepare_kub_label_mapping`

njohner (approved):
LGTM

PR #7519, opened 2022-07-13
Retrieve comment from event for DocumenVersionCreatedActivity.

njohner:
The comment is already stored on the corresponding event (`ObjectCheckedInEvent`), so we can avoid retrieving the latest version of the document in the activity by simply getting the comment from the event. This is more efficient and fixes issues with testing, as retrieving a version makes a transaction savepoint which is not supported by the testserver.

So to be clear, this fixes the failing gever-ui tests...

For [CA-4235]

## Checklist
- [ ] Changelog entry -> no cl entry

[CA-4235]: https://4teamwork.atlassian.net/browse/CA-4235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/document/interfaces.py
+++ b/opengever/document/interfaces.py
@@ -31,6 +31,8 @@ class IObjectCheckoutCanceledEvent(IObjectEvent):
 class IObjectRevertedToVersion(IObjectEvent):
     """The object was reverted back to a specific version."""
 
+    comment = schema.Text(title=u'journal comment')
+
 
 class IFileCopyDownloadedEvent(IObjectEvent):
     """Event interface for event.FileDownloadedEvent."""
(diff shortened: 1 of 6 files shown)


opengever/document/interfaces.py
```diff
@@ -31,6 +31,8 @@ class IObjectCheckoutCanceledEvent(IObjectEvent):
 class IObjectRevertedToVersion(IObjectEvent):
     """The object was reverted back to a specific version."""
 
+    comment = schema.Text(title=u'journal comment')
```
tinagerber, line 34:
```suggestion
    comment = schema.Text(title=u'Journal comment')

tinagerber (approved):

PR #7521, opened 2022-07-14
Change title of the journal entry for restoring a document version

tinagerber:
Follow up for #7515 
I overlooked the journal entry.

For [CA-4280]


njohner (approved):
LGTM

PR #7522, opened 2022-07-16
Fix docker requirements

buchi:
Add test to verify that Buildout and Docker versions match and update versions that do not match.

Some ancient versions used with Buildout were bumped as well as we already use newer versions with Docker:
- Use a more recent version of alembic which has approx. the same age as our SQLAlchemy version.
- Bump psycopg2 to next minor version.
- Bump pytz to latest version.
- Bump SQLAlchemy-Utils to latest version supported with Python 2.7

For [CA-XXXX]

## Checklist

- [x] Changelog entry


njohner (approved):
🎉 Thanks LGTM!
Could you also remove the following line from the PR template: https://github.com/4teamwork/opengever.core/blob/master/.github/pull_request_template.md?plain=1#L24

PR #7527, opened 2022-07-21
Add parameter to deactivate the workspace in the @unlink-workspace endpoint

tinagerber:
One can now unlink a workspace and deactivate it at the same time.

For [CA-3514]

## Checklist

- [x] Changelog entry

- API change:

[CA-3514]: https://4teamwork.atlassian.net/browse/CA-3514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-3514.feature
+++ b/changes/CA-3514.feature
@@ -1,0 +1,1 @@
+Add parameter to deactivate the workspace in the @unlink-workspace endpoint. [tinagerber]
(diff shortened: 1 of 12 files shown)


njohner (commented):
I guess it's ok, but I'm not a fan of having a single request doing to different things with no transactionality. I'd say unlinking should fail if the user does not have the permission to inactivate it, would be much cleaner IMO. We'd need to either first inactivate (I think this is currently not possible), or first find out whether the user has the necessary permissions, or we could relink the dossier on the workspace side when the inactivation fails.

tinagerber:
@njohner As discussed, I have included a check to see if the transition is available. Can you have another look?

njohner (approved):
LGTM

PR #7530, opened 2022-07-26
Fix upgrade step enabling all languages for Teamraum deployments.

njohner:
Sorry about that one. I tested it extensively and switched the condition during testing but it seems I then committed with the wrong condition. I'll backport it before the Thursday SaaS release

For [CA-4083]

## Checklist

- [ ] Changelog entry

[CA-4083]: https://4teamwork.atlassian.net/browse/CA-4083?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/core/upgrades/20220707162158_enable_all_languages_for_teamraum/upgrade.py
```diff
@@ -13,7 +13,7 @@ class EnableAllLanguagesForTeamraum(UpgradeStep):
     }
 
     def __call__(self):
-        if is_workspace_feature_enabled():
+        if not is_workspace_feature_enabled():
```
phgross, line 16: I would touch the ugradestep, so we can make sure that the step is run correctly, also for the already updated deployments.

phgross (approved):

opengever/core/upgrades/20220707162158_enable_all_languages_for_teamraum/upgrade.py
njohner: Yes I thought I'd do the two teamraums from monday by hand, but you're right, it's actually better to touch the upgrade step.

PR #7536, opened 2022-08-04
Fix excel export if multiple custom fields with same name exits.

phgross:
Currently it is not possible to assign multiple propertysheets to one slot, therefore we need to duplicate the field configuration in some cases. This has led to errors in the Excel report.

The current assertion can be ignored, if multiple fields are defined with the same name but are equal.

For [CA-4537]

## Checklist
- [x] Changelog entry

[CA-4537]: https://4teamwork.atlassian.net/browse/CA-4537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-4537.bugfix
+++ b/changes/CA-4537.bugfix
@@ -1,0 +1,1 @@
+Fix excel export if multiple customfields with same name are configured. [phgross]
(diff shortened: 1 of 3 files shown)


lukasgraf (approved):
LGTM! 👍

PR #7540, opened 2022-08-09
Policytemplate: Always configure ALL shared service URLs in SaaS buildouts

lukasgraf:
Policytemplate: Always configure ALL shared service URLs:

For SaaS-Servers, we've decided that all GEVER shared services (Sablon, PDFLatex, MSGConvert and WeasyPrint) will *always* be deployed on all servers (teamraum-servers as well as GEVER-servers).

Therefore we simplify the buildouts by always configuring all service URLs, irrespective of whether they're actually needed or not.

In addition, we remove the `parts -= gems` hack from the policytemplate, since it's not necessary anymore.

For [CA-485](https://4teamwork.atlassian.net/browse/CA-485)

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-485.other
+++ b/changes/CA-485.other
@@ -1,0 +1,1 @@
+Policytemplate: Always configure ALL shared service URLs. [lgraf]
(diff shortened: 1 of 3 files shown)


njohner (approved):
LGTM

PR #7544, opened 2022-08-11
Move version pinnings for QA tools into base-qa.cfg

lukasgraf:
Move version pinnings for QA tools into `base-qa.cfg`:

When we keep them in our version.cfg, our explicit `pycodestyle=2.7.0` pinning gets overwritten by the `pycodestyle <2.9.0` constraint from https://github.com/4teamwork/ftw-buildouts/blob/master/test-versions-plone-4.cfg#L23

This causes a version confict during buildout:

```
Updating flake8.
Version and requirements information containing pycodestyle:
  [versions] constraint on pycodestyle: <2.9.0
  Requirement of flake8-debugger==3.2.0: pycodestyle
  Requirement of flake8==3.9.2: pycodestyle<2.8.0,>=2.7.0
While:
  Updating flake8.
Error: There is a version conflict.
We already have: pycodestyle 2.8.0
```

Somehow buildout can't seem to resolve those constraints, even though they don't contradict each other. Moving the QA tools version pins to `base-qa.cfg` leads to them being extended later and avoids this issue.


njohner (approved):
I guess we will see 😸

PR #7552, opened 2022-08-24
Fix user lookup by email for inbound mail

lukasgraf:
To avoid fuzzy matches when looking up the user by email address for inbound mails, the `pas_search` call [was extended with `exact_match=True`](https://github.com/4teamwork/opengever.core/commit/28d28fda2c1753d81662a4639eff123c86b0938e) to accomplish this.

However, it turns out that the `ActiveDirectoryMultiPlugin` [refuses to do searches](https://github.com/4teamwork/Products.LDAPMultiPlugins/blob/3c8a9aeaa125e23de919312891942bb07ead5ffb/Products/LDAPMultiPlugins/ActiveDirectoryMultiPlugin.py#L341-L348) with `exact_match=True`, unless the query criteria contain the user's id or login. For the `LDAPMultiPlugin` this [works fine](https://github.com/4teamwork/Products.LDAPMultiPlugins/blob/3c8a9aeaa125e23de919312891942bb07ead5ffb/Products/LDAPMultiPlugins/LDAPMultiPlugin.py#L188-L205), but the AD plugin raises a ValueError that later gets supressed, and so no matches are returned.

We therefore change the strategy to let `pas_search` return fuzzy matches again, and filter them post-search to only pick case-insensitive, but otherwise exact matches.

For [CA-4607](https://4teamwork.atlassian.net/browse/CA-4607)

## Checklist

- [x] Changelog entry

⚠️ Needs backport to `2022.15.x` ⚠️

### DIFF
--- a/opengever/mail/browser/inbound.py
+++ b/opengever/mail/browser/inbound.py
@@ -54,21 +54,32 @@ class GeverMailInbound(MailInbound):
         acl_users = api.portal.get_tool('acl_users')
         pas_search = getMultiAdapter((self.context, self.request),
                                      name='pas_search')
-        users = pas_search.searchUsers(email=sender_email, exact_match=True)
-        if len(users) > 0:
-            user = acl_users.getUserById(users[0].get('userid'))
-            if not hasattr(user, 'aq_base'):
-                user = user.__of__(acl_users)
-            return user
+        matches = pas_search.searchUsers(email=sender_email, exact_match=False)
 
-        elif sender_email.lower() in sender_aliases:
+        if len(matches) > 0:
+            # Matches from pas_search are fuzzy (and exact_match=True may only
+            # be used when querying for 'id' or 'login'). We therefore filter
+            # for exact (but case-insensitive) matches ourselves.
+            for user_record in matches:
+                user = self._get_user_by_id(acl_users, user_record['userid'])
+                user_email = user.getProperty('email')
+                if user_email.lower() == sender_email.lower():
+                    return user
+
+        if sender_email.lower() in sender_aliases:
             target_userid = sender_aliases[sender_email.lower()]
             log.info('Sender address %r is an alias for userid %r' % (
                 sender_email, target_userid))
-            user = acl_users.getUserById(target_userid)
+            user = self._get_user_by_id(acl_users, target_userid)
             if user:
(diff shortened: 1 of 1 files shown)


opengever/mail/browser/inbound.py
```diff
@@ -72,3 +84,25 @@ def get_user(self):
         log.warn('Sender address: %r' % sender_email)
         log.warn('No users found in %r for that address.' % acl_users)
         raise UnknownSender(self.msg())
+
+    def _get_email(self, acl_users, user_record):
+        """Get email address from a pas_search result record, while respecting
+        a potential schema mapping.
+
+        Unfortunately, pas_search result records contain unmapped attribute
+        names (raw, as the appear in LDAP/AD). It accepts mapped names
```
njohner, line 93: as the appear -> as they appear

opengever/mail/browser/inbound.py
```diff
@@ -72,3 +84,25 @@ def get_user(self):
...
+        its schema, and look up the email address using the proper name.
+        """
+        email_attr_name = 'email'
+
+        plugin = acl_users[user_record['pluginid']]
+        if ILDAPMultiPlugin.providedBy(plugin):
+            schema = plugin._getLDAPUserFolder().getSchemaDict()
+            for entry in schema:
+                if entry['public_name'] == 'email':
+                    email_attr_name = entry['ldap_name']
```
njohner, line 106: I think you can add a `break` here.

njohner (approved):
LGTM, just wondering how painful it would be to add a regression test?

lukasgraf:
@njohner I've had to change the strategy a bit again, would you mind doing another review?
_(Need to fetch the actual user for each potential match, because only `user.getProperty('email')` works reliable. For the AD plugin, the email isn't returned in the result records)._

I've tested the code above as-is on DIGS TEST in SG, and it does work there.

njohner (approved):
LGTM.
I think that this code will not be necessary anymore once we implement our OGDS pas plugin no? Maybe you can add a comment in the corresponding Jira issue (well I'm not sure where that would be, I don't think any story covers actually using the new plugin, but somewhere here https://4teamwork.atlassian.net/browse/ROAD-2458)

PR #7553, opened 2022-08-25
Allow moving proposals outside main dossier during meeting migration.

njohner:
Handle meeting dossiers containing proposals during meeting migration. For this we need to allow moving the proposals, as we replace the meeting dossier with a normal dossier, effectively moving the proposal.

I did not add a test, as this is not used in Gever, only in the migration of the meetings.

For [CA-3912]

## Checklist

- [x] Changelog entry

[CA-3912]: https://4teamwork.atlassian.net/browse/CA-3912?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-3912.other
+++ b/changes/CA-3912.other
@@ -1,1 +1,1 @@
-Support disabled meetings feature. [njohner]
+Add support for meetings migration. [njohner]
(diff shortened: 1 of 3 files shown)


lukasgraf (approved):
LGTM! 👍

PR #7558, opened 2022-08-30
Remove IRestrictedDossier behavior and addable_dossier_types field

tinagerber:
This was a custom implementation for Zug and is no longer needed as the use case can now be covered with custom fields.

For [CA-3962]

## Checklist

- [x] Changelog entry

[CA-3962]: https://4teamwork.atlassian.net/browse/CA-3962?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lukasgraf:
@tinagerber I haven't had time for a full review yet, sorry. But be careful with this one regarding broken interfaces.

In some cases, Plone persists interface name or even actual interface class objects in the DB, and if you just remove the code for those interfaces without first removing the interfaces from the objects, you get nasty errors afterwards because loading the objects will attempt to look up `opengever.base.interfaces.IDontExistAnymore` etc..

And I thought this was the case for behavior interfaces, but I could be wrong.

tinagerber:
@lukasgraf The dossiers that had this behavior have been migrated, so there should be no more objects with this behavior.

lukasgraf (approved):
Ah, I see, thanks. LGTM 👍 then :)

PR #7562, opened 2022-09-09
Add username, groupname and external_id columns

tinagerber:
I did not migrate `username`/`groupname` for LDAP users/groups, because I think the mapping should be in the OGDS sync service. 

For local groups I set `groupid` as `groupname`, next OGDS sync should set `groupname`. As long as `groupname` can't be edited, that should be enough.

Also I am not 100% convinced that `username` and `groupname` should be unique, could probably lead to problems as inactive users/groups stay in the DB.

For [CA-4567]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:

[CA-4567]: https://4teamwork.atlassian.net/browse/CA-4567?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


changes/CA-4567-2.feature
```diff
@@ -0,0 +1 @@
+Add groupname and external_id columns to group model. [tinagerber]
```
lukasgraf, line 1: Double changelog?

opengever/ogds/models/user.py
```diff
@@ -28,6 +28,9 @@ class User(Base):
     __tablename__ = 'users'
 
     userid = Column(String(USER_ID_LENGTH), primary_key=True)
+    username = Column(String(USER_ID_LENGTH), unique=True)
```
lukasgraf, line 31: I think you're right, `username` shouldn't have a unique constraint. After all, that's exactly why we're doing this - the username isn't a reliable primary key for users, and may be re-used for different accounts.

`external_id` should be unique however I think, at least at a given point in time.

opengever/ogds/models/user.py
```diff
@@ -28,6 +28,9 @@ class User(Base):
...
+    external_id = Column(String(USER_ID_LENGTH), unique=True)
```
lukasgraf, line 32: Both `username` and `external_id` should be required (`nullable=False`). And as a consequence, everywhere where we create entries in the `users` or `groups` tables, we need to prefill `username` and `external_id` with whatever we write to `userid`. (Same for groups).

lukasgraf (changes requested):
> I did not migrate username/groupname for LDAP users/groups, because I think the mapping should be in the OGDS sync service.

I don't quite understand what you mean by that.

changes/CA-4567-2.feature
tinagerber: One is for the user model and the other for the group model 😄

tinagerber:
@lukasgraf As discussed, I did not set an index for username and groupname because you are implementing a case insensitive index. The failing test also fails in the nightlies and has nothing to do with this implementation.
Can you have another look?

lukasgraf (approved):
LGTM 👍

buchi (approved):

PR #7565, opened 2022-09-12
Sort dossier participations by label in @participations endpoint

tinagerber:
Participants are now sorted by `participant_title` in the `@participations` endpoint.

During the implementation I noticed that the endpoint with KuB participations is extremely slow. Therefore I optimized the KuBActor so that KuB is only requested when necessary.

A request of the endpoint with 5 KuB participations took more than 1.5 seconds. Now the request is about a factor of 45 faster.

For [CA-4678]

## Checklist

- [x] Changelog entry

- API change:

[CA-4678]: https://4teamwork.atlassian.net/browse/CA-4678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/api/dossier_participations.py
+++ b/opengever/api/dossier_participations.py
@@ -1,8 +1,6 @@
 from contextlib import contextmanager
 from opengever.api.actors import serialize_actor_id_to_json_summary
-from opengever.api.batch import SQLHypermediaBatch
 from opengever.base.interfaces import IOpengeverBaseLayer
-from opengever.contact import is_contact_feature_enabled
 from opengever.dossier.behaviors.participation import IParticipationAware
 from opengever.dossier.behaviors.participation import IParticipationAwareMarker
 from opengever.dossier.interfaces import IDossierParticipants
@@ -62,16 +60,14 @@ class Participations(object):
             return result
 
         handler = IParticipationAware(self.context)
-        participations = handler.get_participations()
+        participation_data = [IParticipationData(participation)
+                              for participation in handler.get_participations()]
+        participation_data = sorted(participation_data, key=lambda data: data.participant_title)
 
-        if is_contact_feature_enabled():
-            batch = SQLHypermediaBatch(self.request, participations, 'id')
-        else:
-            batch = HypermediaBatch(self.request, participations)
+        batch = HypermediaBatch(self.request, participation_data)
 
         items = []
-        for participation in batch:
-            data = IParticipationData(participation)
+        for data in batch:
             items.append({
                 "@id": '{}/@participations/{}'.format(
                     self.context.absolute_url(), data.participant_id),
--- a/opengever/ogds/base/actor.py
+++ b/opengever/ogds/base/actor.py
@@ -499,16 +499,7 @@ class KuBContactActor(Actor):
 
     def get_label(self, with_principal=True):
         mapping = KuBClient().get_kub_id_label_mapping()
-        name = mapping.get(self.identifier, self.identifier)
-
-        if with_principal:
-            # XXX Primary_email is currently not returned by the KuB search
-            # endpoint. This should change once we have an endpoint to actually
-            # resolve the contact ids.
-            email = self.contact.get("primary_email", "")
-            if email:
-                return u'{} ({})'.format(name, email)
-        return name
+        return mapping.get(self.identifier, self.identifier)
 
     def get_profile_url(self):
         # XXX This should be an URL pointing to KUB, e.g.
--- a/opengever/ogds/base/actor.py
+++ b/opengever/ogds/base/actor.py
@@ -38,6 +38,7 @@ from opengever.contact.models import Person
 from opengever.contact.utils import get_contactfolder_url
 from opengever.inbox.utils import get_inbox_for_org_unit
 from opengever.kub import is_kub_feature_enabled
+from opengever.kub.client import KuBClient
 from opengever.kub.entity import KuBEntity
 from opengever.ogds.base import _
 from opengever.ogds.base.browser.userdetails import UserDetails
@@ -48,6 +49,7 @@ from opengever.ogds.models.service import ogds_service
 from opengever.ogds.models.team import Team
 from plone import api
 from plone.dexterity.utils import safe_unicode
+from plone.memoize import instance
 from Products.CMFCore.interfaces._tools import IMemberData
 from Products.CMFCore.utils import getToolByName
 from Products.PluggableAuthService.interfaces.authservice import IPropertiedUser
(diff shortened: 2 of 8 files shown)


tinagerber:
The failing test also fails in the nightlies and has nothing to do with this implementation.

opengever/ogds/base/actor.py
```diff
@@ -484,24 +486,29 @@ class KuBContactActor(Actor):
...
     actor_type = 'kubcontact'
 
-    def __init__(self, identifier, contact=None):
-        super(KuBContactActor, self).__init__(identifier)
-        self.contact = contact
-
     def corresponds_to(self, user):
         return False
 
+    @instance.memoize
```
njohner, line 492: Hum I don't think that is a good idea. Basically you now get the date for a kub entry only once, then never again. What if the data is modified in KuB?

opengever/ogds/base/actor.py
```diff
@@ -759,14 +766,8 @@ def create_contact_actor(self, contact=None):
 
         return ContactActor(self.identifier, contact=contact)
 
-    def create_kub_contact_actor(self, contact=None):
-        if not contact:
-            try:
-                contact = KuBEntity(self.identifier)
-            except LookupError:
-                return self.create_null_actor()
```
njohner, line 767: What happens now when the identifier does not exist in KuB? Shouldn't we return the null actor in that case?

opengever/ogds/base/actor.py
```diff
@@ -499,16 +499,7 @@ def contact(self):
...
-
-        if with_principal:
-            # XXX Primary_email is currently not returned by the KuB search
-            # endpoint. This should change once we have an endpoint to actually
-            # resolve the contact ids.
-            email = self.contact.get("primary_email", "")
-            if email:
-                return u'{} ({})'.format(name, email)
-        return name
+        return mapping.get(self.identifier, self.identifier)
```
njohner, line 502: Ok I'm confused. I though primary_email exists in KuB. Is it never returned?

opengever/api/dossier_participations.py
```diff
@@ -62,16 +60,14 @@ def __call__(self, expand=False):
             return result
 
         handler = IParticipationAware(self.context)
-        participations = handler.get_participations()
+        participation_data = [IParticipationData(participation)
+                              for participation in handler.get_participations()]
+        participation_data = sorted(participation_data, key=lambda data: data.participant_title)
 
-        if is_contact_feature_enabled():
-            batch = SQLHypermediaBatch(self.request, participations, 'id')
```
njohner, line 68: SQL contacts do no longer exist anymore right?

njohner (commented):
Sorry I'm a bit confused by this PR right now. Maybe we can have a quick call tomorrow

opengever/api/dossier_participations.py
tinagerber: I think at the moment they are still around, but not for much longer. SQL contacts could certainly be sorted more optimally, but since they won't be around much longer anyway, I didn't want to spend extra time sorting them more efficiently.

opengever/ogds/base/actor.py
tinagerber: No. There is a field `primaryMail` and in there then a field `email` which would contain the email, but `primary_email` does not exist. Therefore nothing changes if we leave it out. We could also correct this, but it is not quite correct to use the email address anyway, and so the KuB request is not necessary anymore.

opengever/ogds/base/actor.py
tinagerber: Yes, you are right. I have corrected this, now it checks if the identifier exists in the label mapping and creates a KubActor or a Null-Actor depending on it.

opengever/ogds/base/actor.py
tinagerber: No, it is cached as long as the instance of the Actor exists. This was also the case before, the query was simply made in the `create_kub_contact_actor ` method. Now the request is only made when the data is actually needed.

njohner: Ah yes, right sorry 👍

opengever/ogds/base/actor.py
njohner: ok works for me, we will reintroduce it if necessary and then do it correctly...

njohner: ok works for me, we will reintroduce it if necessary and then do it correctly...

opengever/ogds/base/actor.py
```diff
@@ -484,24 +486,29 @@ class KuBContactActor(Actor):
...
-
-        # XXX Primary_email is currently not returned by the KuB search
-        # endpoint. This should change once we have an endpoint to actually
-        # resolve the contact ids.
-        email = self.contact.get("primary_email", "")
-        if with_principal and email:
-            return u'{} ({})'.format(name, email)
-        else:
-            return name
+        mapping = KuBClient().get_kub_id_label_mapping()
```
njohner, line 501: Ah I see the trick now, the label mapping itself is cached for an hour, so that avoids actually making a request to KuB.

opengever/ogds/base/actor.py
```diff
@@ -759,14 +766,11 @@ def create_contact_actor(self, contact=None):
...
-        if not contact:
-            try:
-                contact = KuBEntity(self.identifier)
-            except LookupError:
-                return self.create_null_actor()
-
-        return KuBContactActor(self.identifier, contact=contact)
+    def create_kub_contact_actor(self):
+        mapping = KuBClient().get_kub_id_label_mapping()
+        if not mapping.get(self.identifier):
```
njohner, line 762: This might be problematic and we might have to actually try to get the KubEntity when the identifier is not in the mapping, to avoid having null actors for actors that were created in the last hour. On the other hand we might have kub entities that get deleted, but the identifiers are still somewhere in gever and then we would always make requests to KuB for nothing. So for now, I agree that it's good as it is.

opengever/api/dossier_participations.py
njohner: Ah I understand, because you materialised the query above by sorting, you do no longer need an SQLHypermediaBatch.

opengever/ogds/base/actor.py
njohner: Just as a side note, why do you say it's not correct to use the email address? it's what we have always done for contacts (see https://github.com/4teamwork/opengever.core/blob/master/opengever/ogds/base/actor.py#L427-L428)

njohner (approved):
LGTM

opengever/ogds/base/actor.py
tinagerber: Ah ok, I didn't know that, I just meant because it's not the identifier. The SQL contacts do not show the email.

PR #7569, opened 2022-09-14
Implement OGDS auth plugin

lukasgraf:
This adds an implementation for an OGDS PAS plugin that aims to eventually replace the functionally we currently use from the LDAP/AD plugins of `P.LDAPMultiPlugins`.

It implements the following interfaces:

- `IUserEnumerationPlugin`
- `IGroupEnumerationPlugin`
- `IGroupsPlugin`
- `IPropertiesPlugin`

Results are cached, and SQL queries on hot paths (like `enumerateUsers('current.user')`) are backed by SQL indexes on the appropriate columns.

All lookups that this plugin does (by user ID or group ID) are case-insensitive.

**Notes:**

- `IGroupIntrospection` is not supported yet. In user-facing usage, I coulnd't find a case where this is needed. The groups tab in the "Groups and Users" Plone control panel seems to need it however, so that's currently broken when this Plugin is installed. I held off from implementing this already because I'm looking into a way to avoid having to completely [duplicate the `_findGroup()` and `_verifyGroup()` methods](https://github.com/collective/Products.PloneLDAP/blob/25ec0a42ab49ac9c5bbfff74eee571961b2cb2a1/Products/PloneLDAP/plugins/base.py#L49-L138).

- I have done some simple tests with `gever-ui`, but no in-depth testing yet

- The fact that we now have `userid`, `username` and `external_id` columns in OGDS is not handled yet by this plugin (all lookups still use `userid`).

For [CA-4543](https://4teamwork.atlassian.net/browse/CA-4543)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1365,6 +1365,8 @@ This plugin serves as a replacement for the LDAP/AD PAS plugins to enumerate use
 
 The plugin does not perform authentication itself. It therefore requires another ``IAuthenticationPlugin`` to be present, activated and capable to authenticate users for the given deployment.
 
+For programmatic installation during setup, the ``install_ogds_auth_plugin`` helper function in ``opengever.ogds.auth.plugin`` may be used to perform the steps listed above.
+
 
 Content creation
 ~~~~~~~~~~~~~~~~
--- a/README.rst
+++ b/README.rst
@@ -220,7 +220,7 @@ respective user in our development LDAP tree.
 Solr
 ~~~~
 
-Solr is provided as a Docker image and started with other services using `docker-compose`. 
+Solr is provided as a Docker image and started with other services using `docker-compose`.
 
 
 Activating Solr update chain
@@ -1350,6 +1350,21 @@ Example for a ``configuration.json``:
         ]
       }
     }
+
+OGDS PAS Plugin
+^^^^^^^^^^^^^^^
+
+This plugin serves as a replacement for the LDAP/AD PAS plugins to enumerate users and groups from OGDS instead of LDAP. Because it's still experimental, it's not installed by default. In order to install it, and have it function as intended, the following needs to be done:
+
+- Make sure a plugin is present that can perform authentication (e.g. ``cas_auth``)
+- Add an instance of "OGDS Authentication Plugin" in ZMI
+- In the "Cache" tab of the plugin, associate it with "RAMCache"
+- In the "Activate" tab of the plugin, enable all its capabilities
+- Move the OGDS plugin to the top of the list for properties plugins (acl_users -> plugins -> Properties Plugins -> move ``ogds_auth`` to the top)
+- Disable all of the LDAP plugin's capabilities
+
+The plugin does not perform authentication itself. It therefore requires another ``IAuthenticationPlugin`` to be present, activated and capable to authenticate users for the given deployment.
+
 
 Content creation
 ~~~~~~~~~~~~~~~~
--- a/opengever/ogds/auth/plugin.py
+++ b/opengever/ogds/auth/plugin.py
@@ -222,7 +222,7 @@ class OGDSAuthenticationPlugin(BasePlugin, Cacheable):
         query = (
             select([groups.c.groupid])
             .select_from(groups.join(groups_users))
-            .where(groups_users.c.userid == principal_id)
+            .where(func.lower(groups_users.c.userid) == principal_id.lower())
             .where(groups.c.active == True)
         )
 
--- a/opengever/ogds/auth/plugin.py
+++ b/opengever/ogds/auth/plugin.py
@@ -86,6 +86,8 @@ class OGDSAuthenticationPlugin(BasePlugin):
         )
         if id:
             query = query.where(User.userid == id)
+        if isinstance(max_results, int):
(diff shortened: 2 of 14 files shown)


README.rst
```diff
@@ -1351,6 +1351,21 @@ Example for a ``configuration.json``:
...
+OGDS PAS Plugin
+^^^^^^^^^^^^^^^
+
+This plugin serves as a replacement for the LDAP/AD PAS plugins to enumerate users and groups from OGDS instead of LDAP. Because it's still experimental, it's not installed by default. In order to install it, and have it function as intended, the following needs to be done:
+
+- Make sure a plugin is present that can perform authentication (e.g. ``cas_auth``)
+- Add an instance of "OGDS Authentication Plugin" in ZMI
+- In the "Cache" tab of the plugin, associate it with "RAMCache"
+- In the "Activate" tab of the plugin, enable all its capabilities
+- Move the OGDS plugin's to the top of the list for properties plugins (acl_users -> plugins -> Properties Plugins -> move ``ogds_auth`` to the top)
```
njohner, line 1363: plugin's -> plugin

opengever/ogds/auth/plugin.py
```diff
@@ -40,7 +40,18 @@ def addOGDSAuthenticationPlugin(self, id_, title=None, REQUEST=None):
...
 class OGDSAuthenticationPlugin(BasePlugin, Cacheable):
-    """Plone PAS plugin for authentication against OGDS.
+    """Plone PAS plugin to enumerate users and groups from OGDS.
+
+    This plugin aims to replace parts of the functionality provided by the
+    LDAP/AD plugins from P.LDAPMultiPlugins, so that we can eventually
+    remove those from our stack.
+
+    It acts as a user/group source backed by OGDS, capable of enumerating
+    users, groups group memberships and properties.
```
njohner, line 77: groups group -> groups, group

njohner (approved):
LGTM,

I was just wondering about the caching, is it as it is for the LDAP plugin, i.e. cached for an hour and can be invalidated TTW? If needed we could probably quite easily only invalidate the cache when the OGDS is synced (we have the sync timestamp) or when it is otherwise modified (invited users in Teamraum).

opengever/ogds/auth/plugin.py
```diff
@@ -0,0 +1,400 @@
...
+            return cached_info
+
+        principal_id = safe_unicode(principal_id)
+
+        groups = Group.__table__
+        query = (
+            select([groups.c.groupid])
+            .select_from(groups.join(groups_users))
+            .where(func.lower(groups_users.c.userid) == principal_id.lower())
+            .where(groups.c.active == True)
```
buchi, line 315: Use `sqlalchemy.sql.expression.true()` or `# noqa`

buchi (commented):
LGTM but I think we definitely need support for user lookup by `login`/`username`.

lukasgraf:
@njohner regarding caching:
Yes, it works pretty much the same as with the LDAP/AD plugins: By inheriting from `OFS.Cache.Cachable` you get a `[Cache]` tab in ZMI for free, with which you can (should) associate the plugin with the (site wide) `RAMCache` that exists by default:

Screenshot 2022-09-20 at 18 23 06

There you can invalidate the cache TTW.

In cache settings, the default of 3600 seconds could be changed TTW:
 
Screenshot 2022-09-20 at 18 23 20

Invalidation based on the timestamp in OGDS might be an option. But I don't think it's trivial, because IF a memory cached result is available, we want to return that immediately, without hitting SQL at all (in order to check the timestamp).

opengever/ogds/auth/plugin.py
```diff
@@ -75,6 +78,22 @@ def enumerateUsers(self, id=None, login=None, exact_match=False,
...
 
+        view_name = self.getId() + '_enumerateUsers'
+        criteria = {
+            'id': id,
+            'login': login,
+            'exact_match': exact_match,
+            'sort_by': sort_by,
+            'max_results': max_results,
+        }
+        criteria.update(kw)
```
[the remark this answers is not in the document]
lukasgraf: I did that deliberately. I did actually consider this, but decided against it: Caching incorrect results for a particular cache key can have security implications here. If we accidentally poison the cache for a particular query with incorrect results, that can lead to problematic outcomes.

Because the cached results are functionally dependent on the exact arguments to those methods, just constructing the cache key criteria from the method arguments as early as possible (before mutating or interpreting them - short of turning some of them like `PloneUser` objects into string representations) should always be correct. And the pattern for caching accross the different methods then always stays the same, and is simpler to reason about.

opengever/ogds/auth/plugin.py
```diff
@@ -216,10 +222,12 @@ def getGroupsForPrincipal(self, principal, request=None):
             .where(groups_users.c.userid == principal_id)
             .where(groups.c.active == True)
         )
-        results = tuple([
-            groupid.encode('utf-8')
+
+        # Omit groups with non-ASCII names
```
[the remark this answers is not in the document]
lukasgraf: Defininitely. Though I would still keep the handling in this plugin, just to make sure it always returns results that don't break Plone.

buchi (approved):

PR #7574, opened 2022-09-20
Fix exit code of QA tests

buchi:
Prior to this change only the exit code of isort was used and flake8 violations were not reported as failures.

For [CA-XXXX]

## Checklist

- [ ] Changelog entry


lukasgraf (approved):
LGTM 👍

PR #7575, opened 2022-09-20
Allow cancellation of tasks in progress

tinagerber:
The transition already existed but only for task templates. The transition is not available for cross-client tasks that are processed in  in a dossier on the responsible tenant.

For [CA-4666]

## Checklist

- [x] Changelog entry

[CA-4666]: https://4teamwork.atlassian.net/browse/CA-4666?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-4666.feature
+++ b/changes/CA-4666.feature
@@ -1,0 +1,1 @@
+Allow cancellation of tasks in progress. [tinagerber]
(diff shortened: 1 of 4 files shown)


njohner (commented):
I think you need to make sure to disallow this transition for cross-tenant tasks that are being treated in a dossier on the responsible tenant.

njohner (approved):
LGTM

PR #7577, opened 2022-09-21
Make max dossier depth restriction less strict

lukasgraf:
Don't enforce max dossier depth if a structure of same nesting level as the new one already exists.

For example: If a dossier already contains a subdossier with a sub-sub-dossier that already exceeds the max depth by 2, then it will be allowed to create more content on that level that exceeds the depth by up to 2 levels (but not more).

**Note:**
I tried to switch `Dossier.get_subdossiers()` to Solr instead of ZCatalog queries, because it is now (often) used when viewing a dossier, and having to determine whether to show subdossiers as addable or not.

However, I had to abort that attempt because `get_subdossiers()` is used in so many places, that ~80 tests failed, and among those 13 `Functional` test cases, which would not easily be converted to `SolrIntegrationTestCases` in the scope of this story.

For [CA-4662](https://4teamwork.atlassian.net/browse/CA-4662)

## Checklist

- [x] Changelog entry

### DIFF
--- a/opengever/dossier/base.py
+++ b/opengever/dossier/base.py
@@ -201,7 +201,27 @@ class DossierContainer(Container):
             interface=IDossierContainerTypes,
             default=100,
         )
-        return self._get_dossier_depth() + additional_depth <= max_depth + 1
+        max_depth_respected = (
+            self._get_dossier_depth() + additional_depth <= max_depth + 1)
+
+        if max_depth_respected:
+            return True
+
+        # Max depth would technically be exceeded.
+        # However, if a dossier structure that exceeds the max depth already
+        # exists at the same level, we'll allow new structures with the same
+        # additional depth (or lower).
+        path = '/'.join(self.getPhysicalPath())
+        subdossiers = self.get_subdossiers(unrestricted=True)
+        if subdossiers:
+            max_existing_additional_depth = max((
+                sub.getPath()[len(path):].count('/') for sub in subdossiers
+            ))
+
+            if additional_depth <= max_existing_additional_depth:
+                return True
+
+        return False
 
     def has_subdossiers(self):
         return len(self.get_subdossiers()) > 0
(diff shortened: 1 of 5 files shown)


opengever/dossier/base.py
```diff
@@ -201,7 +201,27 @@ def is_dossier_structure_addable(self, additional_depth=1):
...
+
+        if max_depth_respected:
+            return True
+
+        # Max depth would technically be exceeded.
+        # However, if a dossier structure that exceeds the max depth already
+        # exists at the same level, we'll allow new structures with the same
+        # additional depth (or lower).
+        path = '/'.join(self.getPhysicalPath())
+        subdossiers = self.get_subdossiers()
```
njohner, line 215: you should add `unrestricted=True` here.

opengever/dossier/base.py
```diff
@@ -201,7 +201,27 @@ def is_dossier_structure_addable(self, additional_depth=1):
...
+        if subdossiers:
+            max_existing_additional_depth = max((
+                sub.getPath()[len(path):].count('/') for sub in subdossiers
```
njohner, line 218: Can there be issues with trailing "/" ?

opengever/dossier/tests/test_move_items.py
```diff
@@ -1595,6 +1597,13 @@ def test_raises_when_dossier_depth_would_be_exceeded(self):
         DossierMovabiliyChecker(self.subsubdossier).validate_movement(
             self.empty_dossier)
 
+        # a dossier containing a subdossier can be even be moved to a place
```
njohner, line 1600: can be even be -> can even be

njohner (approved):
Just a few comments, otherwise LGTM

opengever/dossier/base.py
lukasgraf: Good point, thanks!

opengever/dossier/base.py
lukasgraf: There shouldn't, because the path should always be constructed the same way, but I think I'll make sure anyway using `rstrip('/')`

PR #7578, opened 2022-09-26
Optimize building labels for large `participants` facets

lukasgraf:
Optimize building labels for large `participants` facets:

When processing a large facet (~100k entries) of `participants`, the extraction is slow because a lot of work is repeated for each and every single entry.

This optimization adds some shortcuts that split the index value into participant_id and role only once, and then directly check in the KuB client's cached ID to label mapping if a label can be found for that participant_id.

Only if no label can be found (not cached, unknown ID, ...) it delegates to the classic implementation.

This speeds the extraction up by roughly 50% (or from ~6s to ~3s for a known pathological case).

For [CA-4752](https://4teamwork.atlassian.net/browse/CA-4752)

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-4752.other
+++ b/changes/CA-4752.other
@@ -1,0 +1,1 @@
+Optimize building labels for large `participants` facets. [lgraf]
--- a/opengever/api/solr_query_service.py
+++ b/opengever/api/solr_query_service.py
@@ -96,16 +96,20 @@ class SolrQueryBaseService(Service):
             if not solr_facet:
                 continue
 
-            facet_counts[facet_name] = {}
-            for facet, count in solr_facet.items():
-                if field.hide_facet(facet):
-                    continue
-                facet_counts[field.field_name][facet] = {
-                    "count": count,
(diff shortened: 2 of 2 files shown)


lukasgraf:
@elioschmutz if you want to do it inside `participant_id_to_label()`, you need to somehow store a reference to a `KubClient` to avoid instantiating it 100'000 times. But only if `is_kub_feature_enabled()` is True, otherwise it would raise an error. So you probably need to memoize `is_kub_feature_enabled()` as well.

Can't memoize it on the `ParticipationIndexHelper` instance, because that gets newly instantiated every time as well (unlike in my implementation). So you'd probably need to make `KuBClient` a singleton, using a module global registry. That's not a one line change any more at that point.

lukasgraf:
As discussed during today's review: We currently don't optimize performance for the pathological case of ~100k facets, but instead will look into whether we can avoid using facets alltogether for those cases.

PR #7580, opened 2022-09-27
Add object_sid column to OGDS users

lukasgraf:
Once the OGDS auth plugin is used and the  ActiveDirectoryMultiPlugin is removed, we need to keep the `objectSid` in OGDS for the OneOffixx integration to still work and expose it in the member properties via the OGDS auth plugin.

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration

### DIFF
--- a/changes/CA-4734.other
+++ b/changes/CA-4734.other
@@ -1,0 +1,1 @@
+Add users.object_sid column to OGDS. [lgraf]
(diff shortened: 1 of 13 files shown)


changes/CA-4734.other
```diff
@@ -0,0 +1 @@
+Add users.object_sid column to OGDS. [lgraf]
```
njohner, line 1: Commit description: 
- `the  ActiveDirectoryMultiPlugin` seems to have two spaces
- `to keep the objectSid to OGDS` -> `to store the objectSid in the OGDS` or something like that

njohner (approved):
LGTM

PR #7581, opened 2022-09-28
Add `display_name` column to OGDS user model

lukasgraf:
The LDAP/AD attributes `displayName` or `cn` are sometimes mapped to the `fullname` property. This is especially the case for customers that actively manage the display names of their users, e.g. to also contain the user's department or their role.

If it exists, we store it in OGDS in the `display_name` column. This is therefore a customer-controlled display name, and different from the fullname we build in the `User.fullname()` method by concatenating first and last name.

With that display ending up in the `fullname` property, Plone will use it e.g. when searching for users in the sharing tab, or to display it in the personal tools menu.

Also contains a bugfix for accessing SQLAlchemy RowProxy results.

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration

### DIFF
--- a/changes/CA-4734-3.other
+++ b/changes/CA-4734-3.other
@@ -1,0 +1,1 @@
+Add `display_name` column to OGDS user model. [lgraf]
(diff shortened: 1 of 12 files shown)


njohner (approved):
LGTM

Just for info, you could change the base branch of this PR to `lg/CA-4734/ogds-object-sid` so that the two first commit would not appear here. github then handles everything automatically when the base branch gets merged into master.

PR #7582, opened 2022-09-28
Add `organization` column to OGDS user model:

lukasgraf:
This column is intended to track a user's organization in a simpler way than the existing directorate and department columns.

In order to be populated, it will need to be mapped in the LDAP schema to an AD/LDAP attribute first, with a public_name of 'organization'. Then it will be synced to OGDS.

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - [x] SQL Operations do not use imported model 
  - DB-Schema migration

### DIFF
--- a/changes/CA-4734-6.other
+++ b/changes/CA-4734-6.other
@@ -1,0 +1,1 @@
+Add `organization` column to OGDS user model. [lgraf]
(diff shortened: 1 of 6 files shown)


njohner (approved):
LGTM.
You linked the wrong Jira story though.

PR #7584, opened 2022-09-29
Consider time zone when determining the end date of a dossier

tinagerber:
When determining the end date for dossiers, the time zone was not taken into account. If a document was last modified on 10.10.22 at 00:15, therefore then the date 09.10. came out, because the modified datetime is stored as utc, and from this the date was generated. Now, when converting the datetime to a date, the local time zone is respected.

For [CA-4774]

## Checklist

- [x] Changelog entry

[CA-4774]: https://4teamwork.atlassian.net/browse/CA-4774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-4774.bugfix
+++ b/changes/CA-4774.bugfix
@@ -1,0 +1,1 @@
+Consider time zone when determining the end date of a dossier. [tinagerber]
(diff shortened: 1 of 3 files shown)


lukasgraf (approved):
LGTM! 👍 Nice catch! 😉 🕐

PR #7586, opened 2022-09-29
 Include KuB organization phone number in membership docproperties.

lukasgraf:
Include KuB organization phone number in membership docproperties:

When serializing a KuB membership to DocProperties, this will also include the organization's phone number in `ogg.[scope].organization.phone.number` (where [scope] may currently be `sender` or `recipient`).

A side effect of this is that when serializing an `organization` entity, its phone number is now exposed twice in DocProperties: Once in `ogg.[scope].organization.phone.number` and once in `ogg.[scope.]phone.number`.

This is because the phone number is once included by the `KuBPhoneNumberDocPropertyProvider` at the upper, generic level (that exposes common properties in an entity type agnostic way), and once more by the `KuBOrganizationDocPropertyProvider` under `ogg.[scope].organization.phone.number`, to be able to distinguish the org's phone number from the person's.

For [CA-4703](https://4teamwork.atlassian.net/browse/CA-4703)

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-4703.feature
+++ b/changes/CA-4703.feature
@@ -1,0 +1,1 @@
+Include KuB organization phone number in membership docproperties. [lgraf]
--- a/opengever/base/docprops.py
+++ b/opengever/base/docprops.py
@@ -23,10 +23,15 @@ class BaseDocPropertyProvider(object):
 
         return tuple(name_or_iterable)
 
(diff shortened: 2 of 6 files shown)


buchi (approved):
👍🏻 LGTM

PR #7587, opened 2022-10-03
Include extra address lines from KuB contacts in DocProperties

lukasgraf:
Include extra address lines from KuB contacts in DocProperties:

This includes the KuB fields `addressLine1` and `addressLine2` in DocProperties for addresses as `.address.extra_line_1` and `.address.extra_line_2` respectively.

---

KuB labels these `Adresszusatz 1` in the UI, and names them `addressLine1` in the API. That feels incorrect to me. In common usage, "adress line 1" is usually non-optional (and refers to the street + number), and "address line 2" is optional, and only to be supplied if the first line is not specific enough on its own.

The KuB usage seems different - these are always optional, and always to be used in _addition_ to "street + number". Therefore I named them `extra_line_...` to make it clear that these can't be substituted for any of the other fields, but should be included in addition to them.

For [CA-4703](https://4teamwork.atlassian.net/browse/CA-4703)

## Checklist

- [x] Changelog entry


buchi (approved):
👍🏻 LGTM

PR #7592, opened 2022-10-13
Improve performance of @globalindex endpoint

tinagerber:
Since there was no index for predecessor_id, this meant that the @globalindex endpoint was very slow for Zug, a query in the inbox took over 10 seconds. 

Also, due to the NLSSORT settings, for Zug the review_state must have an index that supports linguistic sort. We had this problem before with task principal https://github.com/4teamwork/opengever.core/pull/5767.

For Zug, the change was made directly to the database, so the upgrade step does not need to be performed there.

For [CA-4836]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:

[CA-4836]: https://4teamwork.atlassian.net/browse/CA-4836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/core/upgrades/20221013095934_add_index_for_predecessor_id/upgrade.py
+++ b/opengever/core/upgrades/20221013095934_add_index_for_predecessor_id/upgrade.py
@@ -1,0 +1,13 @@
+from opengever.base.model import is_oracle
+from opengever.core.upgrade import SchemaMigration
+
+
+class AddIndexForPredecessorId(SchemaMigration):
+    """Add index for predecessor_id.
+    """
+
+    def migrate(self):
+        if is_oracle():
+            # The index already exists for Zug.
+            return
+        self.op.create_index('ix_tasks_predecessor_id', 'tasks', ['predecessor_id'])
--- a/changes/CA-4836.bugfix
+++ b/changes/CA-4836.bugfix
@@ -1,0 +1,1 @@
+Improve performance of @globalindex endpoint. [tinagerber]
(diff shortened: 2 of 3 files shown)


opengever/core/upgrades/20221013095934_add_index_for_predecessor_id/upgrade.py
```diff
@@ -0,0 +1,13 @@
+from opengever.base.model import is_oracle
+from opengever.core.upgrade import SchemaMigration
+
+
+class AddIndexForPredecessorId(SchemaMigration):
+    """Add index for predecessor_id.
+    """
+
+    def migrate(self):
+        if is_oracle():
```
elioschmutz, line 10: ok, we already have it for zug. But the if-statement is not meant to be used to detect zug-deployments, even if zug is the only deployment with an oracle-db.

I'd extend the if with a check for the existence of the index and add it if it does not exist.

elioschmutz (approved):

opengever/core/upgrades/20221013095934_add_index_for_predecessor_id/upgrade.py
tinagerber: I created the indexes manually for all Zug deployments so that I don't have to implement this for Oracle, where it works differently. That's why I prefer not to do that.

PR #7593, opened 2022-10-13
Support external_reference in dossier excel report

tinagerber:
We now also want to display "External Reference" in the custom table. Therefore, the column must also be available in the Excel report.

For [CA-4805]

## Checklist

- [x] Changelog entry

[CA-4805]: https://4teamwork.atlassian.net/browse/CA-4805?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/api/listing.py
```diff
@@ -38,18 +26,9 @@
...
-    'location',
-    'modified',
     'phone_office',
     'preselected',
-    'receipt_date',
-    'reference',
-    'responsible',
-    'review_state',
-    'sequence_number',
-    'start',
```
elioschmutz, line 52: why do you remove all these fields?

tinagerber: As stated in the commit: These fields have a mapping in opengever/base/solr/fields.py and are therefore no longer needed in this list.

elioschmutz: ah yes, sorry, I have not see the separate commit 👍

elioschmutz (approved):

PR #7596, opened 2022-10-21
No longer cache kub label mapping for an hour, use If-Modified-Since header instead

tinagerber:
With the `If-Modified-Since` header the labels are only returned if something has changed since the given date, otherwise a response with status code 304 is returned.
Since, if there have been no changes, the request is very fast, we therefore no longer have to cache the labels for an hour. Since with some requests (`@actors`endpoint) the label mapping is queried several times, I still cache the mapping for 5 seconds so that KuB is requested only once.

For [CA-4802]

## Checklist

- [x] Changelog entry

[CA-4802]: https://4teamwork.atlassian.net/browse/CA-4802?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/kub/client.py
+++ b/opengever/kub/client.py
@@ -1,8 +1,13 @@
+from datetime import datetime
 from opengever.base.sentry import log_msg_to_sentry
 from opengever.kub.interfaces import IKuBSettings
+from persistent.mapping import PersistentMapping
 from plone import api
 from plone.memoize import ram
+from time import mktime
 from time import time
+from wsgiref.handlers import format_date_time
+from zope.annotation import IAnnotations
 import requests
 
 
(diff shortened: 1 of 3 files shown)


opengever/kub/client.py
```diff
@@ -57,16 +72,24 @@ def get_resolve_url(self, _id):
         return u'{}resolve/{}'.format(self.kub_api_url, _id)
 
     # Cache kub labels for an hour
```
phgross, line 74: comment is now outdated.

opengever/kub/client.py
```diff
@@ -11,6 +16,16 @@
 
 class KuBClient(object):
 
+    STORAGE_MODIFIED_KEY = 'opengever.kub.storage.modified'
+    STORAGE_LABEL_MAPPING_KEY = 'opengever.kub.storage.label_mapping'
+
+    def __init__(self):
+        self._storage = IAnnotations(api.portal.get())
```
phgross, line 23: thats a potential write on read, but I don't see a possibility to avoid it, so ok.

phgross (approved):
LGTM 👍

opengever/kub/client.py
tinagerber: Yes we probably have this problem in general with the storages.

PR #7599, opened 2022-10-27
OfficeConnector cancel checkout

buchi:
Track if a document has pending changes and include this info in the checkout metadata for Office Connector.

We want to know if a document was modified since the last checkout and if there are changes that haven't been checked in yet. Unfortunately we can not determine this directly as Word documents are always modified on checkout and checkin because of DocProperties updates. Even if this is not the case, it's hard to determine if a document was really modified. We would have to diff the file contents of the current working copy with the last version. Therefore we add a marker when a new document version gets uploaded.

Requires a new Office Connector version (>1.16.0) to get the feature working, but older versions still work as before.
See: https://github.com/4teamwork/OfficeConnector/pull/392

For [CA-3864](https://4teamwork.atlassian.net/browse/CA-3864)

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-3864.other
+++ b/changes/CA-3864.other
@@ -1,0 +1,1 @@
+Add support for canceling checkout with Office Connector. [buchi]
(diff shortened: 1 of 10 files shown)


lukasgraf (approved):
LGTM 👍 
Changelog entry is missing though I think.

PR #7600, opened 2022-10-28
Fix sorting in OGDSUserListing when members of a group are listed.

njohner:
When listing users from a group, we join the `users` table with the `groups_users` table and we then need to make sure that columns are unambiguous (userid is in both tables). Before we had added the table designation in front of the column name for the sorting parameter, e.g. `users.userid`. Unfortunately this did not work, as the `sort_on` parameter [was simply dropped because the column was not recognised](https://github.com/4teamwork/opengever.core/blob/master/opengever/api/ogdslistingbase.py#L94-L95). If it had not been dropped it would have led to an error in SQL alchemy afterwards anyway.
Instead we now use a table statement to clearly define the sorting column.

For [CA-4852]

## Checklist
- [x] Changelog entry


lukasgraf (approved):
LGTM 👍 
Changelog entry is missing though 😉

njohner:
Ah sorry, forgot to checkin the file...

PR #7603, opened 2022-11-02
Add backend support to copy subdossiers to teamraum

lukasgraf:
This adds backend support for copying (sub)dossiers to teamraum by:
- adding a new `@prepare-copy-dossier-to-workspace` endpoint that will create the respective folder structure on the teamraum side
- allowing the `@copy-document-to-workspace` endpoint to also take teamraum folders as targets

The frontend will then call the `@prepare-copy-dossier-to-workspace` first (with `validate_only=True`) to validate preconditions (main dossier linked to workspace, no checked out documents), and display any errors to the user. It will then call it a second time to have it create an empty folder structure on the teamraum that mirrors the subdossier structure. The endpoint will then return a list of all documents that need to be copied by the frontend, and the teamraum folders they should be copied to.

The frontend will then copy these documents to their respective folders using the existing `@copy-document-to-workspace` endpoint.

For [CA-4726](https://4teamwork.atlassian.net/browse/CA-4726)

## Checklist

- [x] Changelog entry

- API change:
  - [x] API Changelog entry

### DIFF
--- a/opengever/api/linked_workspaces.py
+++ b/opengever/api/linked_workspaces.py
@@ -291,12 +291,14 @@ class PrepareCopyDossierToWorkspacePost(Service):
             uid = dossier.UID()
             msg = _(u'main_dossier_not_linked_to_workspace')
             return {
-                'msg': translate(msg, context=self.request),
-                'msgid': unicode(msg),
-                'obj_uid': uid,
-                'obj_title': dossier.title,
-                'obj_url': dossier.absolute_url(),
-                'obj_type': dossier.portal_type,
+                'translated_message': translate(msg, context=self.request),
+                'message': unicode(msg),
+                'additional_metadata': {
+                    'obj_uid': uid,
+                    'obj_title': dossier.title,
+                    'obj_url': dossier.absolute_url(),
+                    'obj_type': dossier.portal_type,
+                },
             }
 
     def validate_document(self, doc):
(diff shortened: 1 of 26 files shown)


opengever/api/linked_workspaces.py
```diff
@@ -186,12 +188,12 @@ def validate_data(self, data):
                 "Document can't be copied to a workspace because it's "
                 "currently checked out")
 
-        if not self.obj_contained_in(document, self.context):
+        if not folder_uid and not self.obj_contained_in(document, self.context):
```
njohner, line 191: I think that change is wrong. Whether we want to copy the document to the workspace or to a workspace folder, it still can only come from the current main dossier

njohner (approved):
Just a few small things, otherwise LGTM!
And thanks for fixing some testing issues 🎉 !

opengever/api/linked_workspaces.py
lukasgraf: You're right - I didn't realize `obj_contained_in()` is actually recursive.

opengever/dossier/actions.py
```diff
@@ -58,6 +58,23 @@ def is_move_items_available(self):
...
 
+@adapter(IDossierMarker, IOpengeverBaseLayer)
+class SubDossierListingActions(DossierListingActions):
+
+    def is_copy_dossier_to_workspace_available(self):
+        if not self.context.is_open():
+            return False
+        if not is_workspace_client_feature_available():
+            return False
+        if not api.user.has_permission('Modify portal content', obj=self.context):
```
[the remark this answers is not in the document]
lukasgraf: Me neither, to be honest. But it's consistent with the other teamraum-connect actions, and I'd rather open it up if we do find that it is too restrictive, if that turns out to be the case.

PR #7606, opened 2022-11-07
Use disposition creator's fullname in eCH0160 'ablieferndeStelle' field

lukasgraf:
According to comments in the code, this is what's supposed to be used in this field, instead of the current user's name.

This also fixes a unicode error when disposing a disposition on a deployment where the admin unit title contains non-ASCII characters.

For [CA-4889](https://4teamwork.atlassian.net/browse/CA-4889)

## Checklist

- [x] Changelog entry


njohner (approved):
LGTM

PR #7613, opened 2022-11-15
No longer grant permission to task responsible to add content to the dossier.

njohner:
Taks responsibles were granted the `Contributor` role on the task's parent container (usually the dossier). This is necessary as they need `View` permission on the dossier without getting `View` on its children. As a side-effect, the task responsible was allowed to add new objects into the dossier.
Here we add a new role `TaskResponsible`, which grants `View` and a few other permissions on the dossier, but does not grant permission to add new objects.

Because the `Sharing page: Delegate TaskResponsible role` is not assigned to any role, the `TaskResponsible` role is only displayed in the sharing view when permissions are disregarded, i.e. in "View" mode (except for managers). This means the role is displayed in the `Info` panel in the classic UI and in View mode in the sharing tab of the new frontend, but not in the edit mode in the new UI nor in the `sharing` view of the classic UI, see screenshots below.

TaskResponsible is shown in View mode
Screenshot 2022-11-17 at 15 29 12

TaskResponsible is not displayed in Edit mode
Screenshot 2022-11-17 at 15 29 18

TaskResponsible is shown in Info panel
Screenshot 2022-11-17 at 15 27 37

TaskResponsible is not displayed in sharing view
Screenshot 2022-11-17 at 15 27 47

For [CA-4395]

## Checklist
- [x] Changelog entry

- Upgrade-Steps:
  - [x] Make it deferrable if possible -> not deferrable
- New functionality:

[CA-4395]: https://4teamwork.atlassian.net/browse/CA-4395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/core/upgrades/20221117122801_update_workflows/upgrade.py
```diff
@@ -0,0 +1,21 @@
+from ftw.upgrade import UpgradeStep
+
+
+class UpdateWorkflows(UpgradeStep):
+    """Update workflows.
+    """
+
+    def __call__(self):
+        self.install_upgrade_profile()
+        self.update_workflow_security(
```
tinagerber, line 10: Is there a reason why you don't use the NightlyWorkflowSecurityUpdater here?

opengever/core/upgrades/20221117122801_update_workflows/upgrade.py
```diff
@@ -0,0 +1,21 @@
+from ftw.upgrade import UpgradeStep
+
+
+class UpdateWorkflows(UpgradeStep):
```
tinagerber, line 4: Since that somehow goes together, I would probably just create one upgrade step.

tinagerber:
For existing tasks, the role "Contributor" will now still be set. Have you thought about migrating these tasks?

opengever/core/profiles/default/rolemap.xml
```diff
@@ -105,6 +106,7 @@
       <role name="LimitedAdmin" />
       <role name="Manager" />
       <role name="Contributor" />
+      <role name="TaskManager" />
```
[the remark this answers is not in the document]
njohner: changed with `TaskResponsible`, must be a typo of some sort

opengever/core/profiles/default/workflows/opengever_dossier_workflow/definition.xml
```diff
@@ -233,6 +237,7 @@
         name="opengever.document: Checkout"
         acquired="False">
       <permission-role>Contributor</permission-role>
+      <permission-role>TaskResponsible</permission-role>
```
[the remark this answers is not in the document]
njohner: To be able to checkout you need the `opengever.document: Checkout` on the dossier and then the `Modify portal content` permission on the document. On the dossier level we give the same permission `TaskResponsible` on the dossier, and then for certain types we give `Editor` on the related documents. This then really allows to checkout the related documents. For Tasks where you get only `TaskResponsible` on the related documents, you will therefore not be able to check them out. So in short, yes, the `TaskResponsible` needs this permission on the dossier

opengever/task/localroles.py
```diff
@@ -127,12 +127,12 @@ def set_roles_on_distinct_parent(self):
             agency_principal = self.inbox_group_id
 
         self._add_local_roles(distinct_parent, principal, agency_principal,
-                              ('Contributor', ), reindex=False)
+                              ('TaskResponsible', ), reindex=False)
```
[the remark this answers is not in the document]
njohner: Because `Contributor` is a role that is attributed to certain groups on all sorts of objects on client deployments. I can think of cases where maybe being allowed to edit a repository folder but not add new ones could be desired. I have no idea whether that is the case on any productive deployment. So if we were to adapt the `Contributor` role, we would need to give to `Editor` role everything the `Contributor` can do, and then we would need to merge the groups and users that have one of these roles on the customer's deployments. When we could reuse the `Contributor` role for something new. This all seems a potential source of permission issues on productive deployments and I don't think we want to take that risk.
But overall I agree that the `Contributor` role as it is, gives pretty strange permisssions...

opengever/core/upgrades/20221117122801_update_workflows/upgrade.py
njohner: No sorry, I blindly let lawgiver generate the upgrade step. Will merge it with the other upgrade step.

opengever/core/upgrades/20221117122801_update_workflows/upgrade.py
njohner: See above

njohner:
> For existing tasks, the role "Contributor" will now still be set. Have you thought about migrating these tasks?

Yeah I had decided it didn't really matter. But I've now added an upgrade step.

njohner:
Thanks for the review, I think I've answered to all your comments or integrated the necessary changes.

opengever/core/upgrades/20221115154555_add_task_responsible_role/rolemap.xml
```diff
@@ -0,0 +1,26 @@
+<rolemap>
+
+  <roles>
+    <role name="TaskResponsible" />
+  </roles>
+
+  <permissions>
+    <!-- POLICY -->
+    <permission name="plone.restapi: Use REST API" acquire="True">
```
tinagerber, line 9: `CMFEditions: Save new version` is missing in the upgrade step

opengever/core/upgrades/20221123141742_update_task_assignment_role/upgrade.py
```diff
@@ -0,0 +1,27 @@
+from ftw.upgrade import UpgradeStep
+from opengever.base.role_assignments import ASSIGNMENT_VIA_TASK
+from opengever.base.role_assignments import ASSIGNMENT_VIA_TASK_AGENCY
+from opengever.base.role_assignments import RoleAssignmentStorage
+
+
+class UpdateTaskAssignmentRole(UpgradeStep):
```
tinagerber, line 7: This upgrade step could probably also be deferrable, no?

tinagerber (approved):

opengever/core/upgrades/20221115154555_add_task_responsible_role/rolemap.xml
njohner: 🐖 💵 
Thanks

opengever/core/upgrades/20221123141742_update_task_assignment_role/upgrade.py
njohner: yes definitely

njohner:
@tinagerber can you have a look at the 3 last commits 😅 . Philippe made me realise that inboxes were also concerned. Moreover my upgrade step was broken, because the list of roles in the assignments is not a `PersistentList`. So I have to actually set the `roles` anew for each assignment.

tinagerber (approved):

PR #7614, opened 2022-11-16
Sequential tasks: Only allow subtasks to be started when parent is in progress

lukasgraf:
In a nested task structure that contains a sequential subprocess, it was possible to manually start a subtask inside such a subprocess even though the containing parent task was still in the planned state.

This lead to an invalid task structure, causing the logic that is supposed start the next task when a task is closed to trip up, because it doesn't expect the next task to be already in progress.

Sentry: https://sentry.4teamwork.ch/organizations/sentry/issues/98191/

For [CA-4945](https://4teamwork.atlassian.net/browse/CA-4945)

## Checklist

- [x] Changelog entry

Bug fixed:

### DIFF
--- a/changes/CA-4945.bugfix
+++ b/changes/CA-4945.bugfix
@@ -1,0 +1,1 @@
+Sequential tasks: Only allow subtasks to be started when parent is in progress. [lgraf]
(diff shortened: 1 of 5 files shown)


phgross (approved):
👍 LGTM

PR #7615, opened 2022-11-17
Add `@@oc_checkout` view that redirects to oc: checkout URL

lukasgraf:
This view allows 3rd party applications to produce links that should directly trigger a checkout using OfficeConnector on a GEVER document.

In case a necessary precondition is not satisfied (e.g., document is already checked out by someone else), the view just redirects to the regular document view, which should give the user some indication as to why the checkout didn't work.

> **Note**: This registers a browser view with the same same as the [existing `oc_checkout` REST API endpoint](https://github.com/4teamwork/opengever.core/blob/d7ee0559ca3989fba3a55ded972f623acb585adc/opengever/officeconnector/configure.zcml#L70-L77).
> This technically isn't a problem, because one is a REST service and the other a Browser View (`plone.rest` can distingiush those via content negotiation), one is on the Site Root and the other only on documents, and one is POST while the other one is GET.
> Just thought I'd point it out.

For [CA-4900](https://4teamwork.atlassian.net/browse/CA-4900)

## Checklist

- [x] Changelog entry

- New functionality:
  _No, because mails are not supposed to be checked out_

### DIFF
--- a/changes/CA-4900.feature
+++ b/changes/CA-4900.feature
@@ -1,0 +1,1 @@
+Add ``@@oc_checkout`` view that redirects to oc: checkout URL. [lgraf]
--- a/opengever/officeconnector/browser/configure.zcml
+++ b/opengever/officeconnector/browser/configure.zcml
@@ -11,4 +11,11 @@
       permission="zope2.View"
       />
 
+  <browser:page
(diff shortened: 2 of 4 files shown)


njohner (approved):
LGTM

PR #7621, opened 2022-11-24
Also allow to deliver mails back to the predecessor task.

njohner:
There is no reason to restrict this to documents only.

For [CA-5004]

## Checklist

- [x] Changelog entry

[CA-5004]: https://4teamwork.atlassian.net/browse/CA-5004?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/task/validators.py
+++ b/opengever/task/validators.py
@@ -12,11 +12,14 @@ from zope.interface import Invalid
 
 
 def get_checked_out_documents(intids):
+    from opengever.mail.mail import IOGMailMarker
     checked_out = []
     intid_utility = getUtility(IIntIds)
     request = getRequest()
     for intid in intids:
         doc = intid_utility.getObject(intid)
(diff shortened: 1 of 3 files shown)


opengever/task/validators.py
```diff
@@ -12,11 +12,14 @@
 
 
 def get_checked_out_documents(intids):
+    from opengever.mail.mail import IOGMailMarker
```
tinagerber, line 15: I assume there is a problem with circular imports here, right?

tinagerber (approved):

opengever/task/validators.py
njohner: Yes something like that. I thought about moving the interface into `opengever.mail.interfaces`, but moving an interface is a bit of a pain.

PR #7622, opened 2022-11-25
Revoke roles when objects are removed from a task's related items.

njohner:
Local roles are added on documents that are related to a task. When closing that task, the local roles are revoked on the current list of related items, but if some objects were removed from the list of related items at some point, the local roles on these documents would never get revoked.
For newly added related items, the local roles were correctly updated in an event handler. To revoke the local roles on related items that get removed, we need to find out which objects were removed from the list of related items. This information is not accessible in the event handler, so we had to store the list of related items in the annotations on each task, allowing us to compare it with the `relatedItems` field on the task and revoking the local roles accordingly.

For [CA-5006]

## Checklist
- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration

[CA-5006]: https://4teamwork.atlassian.net/browse/CA-5006?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/task/localroles.py
+++ b/opengever/task/localroles.py
@@ -13,7 +13,14 @@ from opengever.globalindex.handlers.task import sync_task
 from opengever.meeting.proposal import IBaseProposal
 from opengever.ogds.base.actor import Actor
 from opengever.ogds.base.utils import get_current_org_unit
+from persistent.list import PersistentList
 from plone import api
+from zope.annotation.interfaces import IAnnotations
+from zope.component import getUtility
+from zope.intid import IIntIds
+
+
+RELATED_ITEMS_KEY = 'opengever.task.related_items'
 
 
 class LocalRolesSetter(object):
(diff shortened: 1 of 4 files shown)


opengever/task/localroles.py
```diff
@@ -144,6 +151,25 @@ def set_roles_on_distinct_parent(self):
...
+        if deleted_intids:
+            intids = getUtility(IIntIds)
+            for deleted_intid in deleted_intids:
+                try:
+                    obj = intids.getObject(deleted_intid)
+                except KeyError:
+                    # obj has been deleted
+                    continue
+                self._revoke_roles(obj)
+            annotations[RELATED_ITEMS_KEY] = PersistentList(to_ids)
```
lukasgraf, line 171: This (updating RELATED_ITEMS_KEY) is indented below the `if deleted_intids`. Shouldn't this always be updated? (Specifically also in the case where related items were added)

lukasgraf (approved):
LGTM 👍 apart from the comment about updating the RELATED_ITEMS_KEY.

PR #7623, opened 2022-11-25
Introduce a new option hide_members_for_guests on workspaces

phgross:
This PR adds a new option `hide members for guests`, which allows a workspace admin to hide the member data for guests.
Enabling the flag, which is disabled by default, means:
 - The share_content action is not available for guests of a workspace
 - The participant listing stays empty, and displays a corresponding message instead
 - All members of such a workspace are excluded of general user listings in other workspaces (for example in the `Add User` Select-field)

**Technical thoughts**
- `@participations` and `@workspace-content-members` endpoint raises unauthorized for Guests if the flag is enabled
- The VisibleUsersAndGroupsFilter skip workspaces with enabled flag, so that the hidden workspace members are not available in other workspaces.  To accomplish this, I introduced a new BooleanIndex `hide_member_details`.

For [CA-4950]

## Checklist
- [x] Changelog entry

- Upgrade-Steps:
- API change:
- New translations

[CA-4950]: https://4teamwork.atlassian.net/browse/CA-4950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -11,6 +11,8 @@ Breaking Changes
 
 Other Changes
 ^^^^^^^^^^^^^
+- Workspace and workspace folders serialization contains a new attribute ``can_access_members``.
+- ``@participations`` and ``@@workspace-content-members`` is no longer available for guests in workspaces with enabled ``hide_member_details`` option.
 
 2022.23.0 (2022-11-24)
 ----------------------
--- a/opengever/api/participations.py
+++ b/opengever/api/participations.py
@@ -116,6 +116,9 @@ class ParticipationsGet(ParticipationTraverseService):
     GET workspace/@participations HTTP/1.1
     """
     def reply(self):
+        if not self.context.access_members_allowed():
+            raise Forbidden
+
         token = self.read_params()
         if token:
             return self.prepare_response_item(self.find_participant(
--- a/opengever/api/workspace_content_members.py
+++ b/opengever/api/workspace_content_members.py
@@ -6,6 +6,7 @@ from plone.restapi.interfaces import ISerializeToJson
 from plone.restapi.services import Service
 from Products.CMFPlone.utils import safe_unicode
 from zExceptions import BadRequest
+from zExceptions import Forbidden
 from zope.component import getMultiAdapter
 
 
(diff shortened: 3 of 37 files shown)


opengever/api/participations.py
```diff
@@ -116,6 +116,9 @@ class ParticipationsGet(ParticipationTraverseService):
     GET workspace/@participations HTTP/1.1
     """
     def reply(self):
+        if not self.context.access_members_allowed():
```
njohner, line 119: Will that happen or will the gever-ui know not to use this endpoint? If it happens in normal situations, I'd raise a `opengever.api.not_reported_exceptions.Forbidden` instead. Otherwise I think this should be a `Forbidden`, not `Unauthorized`

opengever/api/workspace_content_members.py
```diff
@@ -13,6 +14,10 @@ class WorkspaceContentMembersGet(Service):
     def reply(self):
         if not is_within_workspace(self.context):
             raise BadRequest("'{}' is not within a workspace".format(self.context.getId()))
+
+        if not self.context.access_members_allowed():
+            raise Unauthorized
```
njohner, line 19: Same here, either `Forbidden` or `opengever.api.not_reported_exceptions.Forbidden`

opengever/workspace/hooks.py
```diff
@@ -0,0 +1,7 @@
+from opengever.core.catalog import add_catalog_indexes
+from opengever.workspace.config import INDEXES
+import logging
+
+
+def installed(site):
+    add_catalog_indexes(INDEXES, logging.getLogger('opengever.workspace'))
```
njohner, line 7: Belongs in the next commit.

docs/public/dev-manual/api/api_changelog.rst
```diff
@@ -11,6 +11,8 @@ Breaking Changes
 
 Other Changes
 ^^^^^^^^^^^^^
+- Workspace and workspace folders serialization contains a new attribute ``can_access_members``.
+- ``@participations`` and ``@@workspace-content-members`` is no longer available for guets in workspaces with enabled ``hide_member_details`` option.
```
njohner, line 15: guets -> guests

njohner (approved):
Just a few details, otherwise LGTM

opengever/api/participations.py
phgross: The gever-ui will know not to use the endpoint, I changed it to `Forbidden`

PR #7624, opened 2022-11-28
Make OGDS sync case insensitive

lukasgraf:
This makes the OGDS synchronisation case inensitive with regard to the user ID.

Therefore, when a user ID changes case in LDAP/AP, it will still be matched to the existing OGDS user (with original capitalization for its user ID) when updating user properties and group memberships.

For [CA-4834](https://4teamwork.atlassian.net/browse/CA-4834)

## Checklist

- [x] Changelog entry

### DIFF
--- a/opengever/ogds/base/sync/ogds_updater.py
+++ b/opengever/ogds/base/sync/ogds_updater.py
@@ -624,7 +624,6 @@ class OGDSUpdater(object):
         """
         ldap_keys = set(ldap_objects.keys())
         ogds_keys = set(ogds_objects.keys())
-        lower_ogds_keys = set([userid.lower() for userid in ogds_keys])
 
         ldap_objects_ci = CaseInsensitiveDict(ldap_objects)
         ogds_objects_ci = CaseInsensitiveDict(ogds_objects)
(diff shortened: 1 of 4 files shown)


opengever/ogds/base/sync/ogds_updater.py
```diff
@@ -22,6 +22,7 @@
 from sqlalchemy.sql.expression import false
 from sqlalchemy.sql.expression import or_
 from sqlalchemy.sql.expression import true
+from UserDict import UserDict
```
njohner, line 25: I did not know about that class. That's actually really useful

opengever/ogds/base/sync/ogds_updater.py
```diff
@@ -637,22 +641,33 @@ def update_mappings(self, ldap_objects, ogds_objects, pk='userid'):
...
 
-        deleted = ogds_active_keys - ldap_keys
-        existing = ldap_keys & ogds_keys
+        deleted = [k for k in ogds_active_keys if k not in ldap_objects_ci]
+        existing = [k for k in ogds_keys if k in ldap_objects_ci]
         modified = {}
         for key in existing:
-            diff = set(ldap_objects[key].items()) ^ set(ogds_objects[key].items())
+            # Case-insensitive lookup of the matching LDAP record
+            ldap_record = ldap_objects.get(key, ldap_objects_ci[key])
```
njohner, line 654: I think you can simply use `ldap_record = ldap_objects_ci[key]` no? I don't see why you would need to first try to get from `ldap_objects`, as `ldap_objects_ci` will anyway always first look for exact matches.

opengever/ogds/base/sync/ogds_updater.py
```diff
@@ -637,22 +641,33 @@ def update_mappings(self, ldap_objects, ogds_objects, pk='userid'):
...
+                if attributes:
+                    modified[key] = attributes
 
         added_mappings = [ldap_objects[a] for a in added_ignore_capitalization]
         deleted_mappings = [{pk: d, 'active': False} for d in deleted]
         modified_mappings = []
         for key, modified_attrs in modified.items():
             changes = {pk: key}
+            # Case-insensitive lookup of the matching LDAP record
+            ldap_record = ldap_objects.get(key, ldap_objects_ci[key])
```
njohner, line 673: Same here. `ldap_record = ldap_objects_ci[key]` should work

njohner (approved):
LGTM

opengever/ogds/base/sync/ogds_updater.py
lukasgraf: You're right, this is a remnant of before, where this wasn't guaranteed yet.

PR #7628, opened 2022-12-01
Use the membership's data for its docproperties.

njohner:
Instead of getting the phone number, URL, Email and Address either from the Person or the Organisation, we now get the data from the Membership itself. This is possible as a `Membership` now has `primaryEmail`, `primaryAddress`, `primaryPhoneNumber` and `primaryUrl` fields.

I also did a small change to avoid unnecessary requests to KuB.

For [CA-4738]

## Checklist
- [x] Changelog entry

[CA-4738]: https://4teamwork.atlassian.net/browse/CA-4738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/kub/docprops.py
```diff
@@ -18,8 +18,10 @@ def __init__(self, entity):
         elif entity.is_organization():
             self.organization = entity
         elif entity.is_membership():
-            self.person = KuBEntity("person:" + entity["person"]["id"])
-            self.organization = KuBEntity("organization:" + entity["organization"]["id"])
+            self.person = KuBEntity("person:" + entity["person"]["id"],
```
lukasgraf, line 21: This change probably means we can get rid of all those secondary mocked requests in tests (like [these](https://github.com/4teamwork/opengever.core/blob/master/opengever/kub/tests/test_docprops.py#L70-L71)), where we need to additionally mock not just the membership, but also one or both ends of the relationship, right?

Just would be a nice cleanup, I'm leaving it up to you whether you want to add this to this PR or not.

lukasgraf (approved):
LGTM 👍, just some potential for even more cleanup ;)

opengever/kub/docprops.py
njohner: Ah you're probably right, will do that.

PR #7633, opened 2022-12-07
Disable lock check in @tus-upload endpoint again

tinagerber:
With https://github.com/4teamwork/opengever.core/commit/51ed9dd4c66f1031525c571c2953a5c41aeba1a3 the lock check was enabled in the @tus-upload endpoint.
But the check does not work properly, so I have to disable it again. I have created a bug to address the problem:
https://4teamwork.atlassian.net/browse/CA-5107

For [CA-5093]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:

[CA-5093]: https://4teamwork.atlassian.net/browse/CA-5093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-5093.bugfix
+++ b/changes/CA-5093.bugfix
@@ -1,0 +1,1 @@
+Disable lock check in @tus-upload endpoint again. [tinagerber]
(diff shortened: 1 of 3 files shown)


njohner (approved):
Ouch that one hurts... Does only SaaS have that issue? I think this needs to go into the 2022.24 release. Otherwise we will get support all over the place. I might even argue to backport it to 2022.23 and update SaaS tomorrow and Monday.

PR #7634, opened 2022-12-08
Fix permission check for revive_bumblebee_preview action

tinagerber:
Since the permission check was wrong, it was only possible for users with global roles (i.e. administrators) to revive the preview. 

Furthermore, the permission should be available to all roles that have View Permission. A TaskResponsible does not necessarily have to be a Reader.

Only the first commit needs to be backported.

For [CA-5090]

## Checklist

- [x] Changelog entry

[CA-5090]: https://4teamwork.atlassian.net/browse/CA-5090?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-5090.bugfix
+++ b/changes/CA-5090.bugfix
@@ -1,0 +1,1 @@
+Fix permission check for revive_bumblebee_preview action. [tinagerber]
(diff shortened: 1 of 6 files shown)


phgross (approved):
👍  LGTM

PR #7643, opened 2022-12-15
Addressblock formatting: Filter any None values before joining lines.

lukasgraf:
Addressblock formatting: Filter any `None` values or empty strings before joining lines.

This may be needed when certain attributes required for a complete address are missing. This will result in incomplete addresses being produced, but that's still preferable to having the "Create document from template" action fail (or any other context, where that DocProperty is generated).

For [CA-5079](https://4teamwork.atlassian.net/browse/CA-5079)

## Checklist

- [ ] Changelog entry (_bug fixed in same release as it was introduced_) 


njohner (approved):
LGTM

PR #7646, opened 2022-12-19
Make address block docproperties more robust

lukasgraf:
This fixes address block data extraction for KuB entities without a `primaryAddress`, and also makes the address block docproperties more robust in general. 

For [CA-5079](https://4teamwork.atlassian.net/browse/CA-5079)

## Checklist

- [ ] Changelog entry _(bug fixed in same release as it was introduced)_


njohner (approved):
LGTM

PR #7655, opened 2023-01-10
Workspace touched date

elioschmutz:
This PR adds the touched date index for workspaces.

The `touched` field was only implemented for dossiers. In this PR, we refactor out the field into its own schema behavior called `ITouched`. We then generally use this behavior for dossiers and workspaces.

The indexer method is now also registered for all `IDexterityContent` objects and will only return the date if the `ITouched` behavior is provided. This makes the implementation more generic and reusable for different content types.

The upgradestep in this PR is not implemented as a nightly upgrade step because migrating the dossiers `touched` filed does not require a reindex, so this is blazing fast. For setting the initial touched date for workspaces we only need to reindex workspaces. There are no deployments with a huge amount of workspaces.

For [CA-4875]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-4875]: https://4teamwork.atlassian.net/browse/CA-4875?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/base/indexes.py
+++ b/opengever/base/indexes.py
@@ -3,6 +3,7 @@ from Acquisition import aq_inner
 from Acquisition import aq_parent
 from opengever.base.behaviors.changed import IChanged
 from opengever.base.behaviors.changed import IChangedMarker
+from opengever.base.behaviors.touched import ITouched
 from opengever.base.behaviors.translated_title import ITranslatedTitle
 from opengever.base.behaviors.translated_title import ITranslatedTitleSupport
 from opengever.base.interfaces import IReferenceNumber
@@ -160,3 +161,8 @@ def getObjPositionInParent(obj):
             return
         return ploneGetObjPositionInParent(obj)()
     return None
+
+
+@indexer(ITouched)
+def touched_indexer(obj):
+    return ITouched(obj).touched
--- a/opengever/core/upgrades/20230110082930_add_i_touched_behavior/upgrade.py
+++ b/opengever/core/upgrades/20230110082930_add_i_touched_behavior/upgrade.py
@@ -1,4 +1,5 @@
 from ftw.upgrade import UpgradeStep
+from opengever.base.behaviors.changed import IChanged
 from opengever.base.behaviors.touched import ITouched
 from opengever.dossier.behaviors.dossier import IDossier
 
@@ -10,6 +11,7 @@ class AddITouchedBehavior(UpgradeStep):
     def __call__(self):
         self.install_upgrade_profile()
         self.migrate_touched_to_new_behavior()
+        self.index_changed_on_todos_and_todo_lists()
         self.index_touched_on_worskspaces()
 
     def migrate_touched_to_new_behavior(self):
@@ -39,3 +41,10 @@ class AddITouchedBehavior(UpgradeStep):
             )[0].changed
 
             obj.reindexObject(idxs=['UID', 'touched'])
+
+    def index_changed_on_todos_and_todo_lists(self):
+        query = {'object_provides': ['opengever.workspace.interfaces.IToDo',
+                                     'opengever.workspace.interfaces.IToDoList']}
+        for obj in self.objects(query, "Index changed on todos and todolists."):
+            IChanged(obj).changed = obj.modified()
+            obj.reindexObject(idxs=['UID', 'changed'])
--- a/opengever/core/upgrades/20230110082930_add_i_touched_behavior/upgrade.py
+++ b/opengever/core/upgrades/20230110082930_add_i_touched_behavior/upgrade.py
@@ -10,6 +10,7 @@ class AddITouchedBehavior(UpgradeStep):
     def __call__(self):
         self.install_upgrade_profile()
         self.migrate_touched_to_new_behavior()
+        self.index_touched_on_worskspaces()
 
     def migrate_touched_to_new_behavior(self):
         TOUCHED_ANNOTATIONS_KEY = '{}.touched'.format(IDossier.__identifier__)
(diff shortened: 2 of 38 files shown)


opengever/base/indexes.py
```diff
@@ -160,3 +161,8 @@ def getObjPositionInParent(obj):
             return
         return ploneGetObjPositionInParent(obj)()
     return None
+
+
+@indexer(IDexterityContent)
```
njohner, line 166: Why don't you register this for `ITouched`?

opengever/core/upgrades/20230110082930_add_i_touched_behavior/upgrade.py
```diff
@@ -14,6 +14,9 @@ def __call__(self):
     def migrate_touched_to_new_behavior(self):
         query = {'object_provides': 'opengever.dossier.behaviors.dossier.IDossierMarker'}
         for obj in self.objects(query, "Migrate touched from IDossier to ITouched."):
+            if ITouched(obj).touched:
```
njohner, line 18: ? why is that necessary? That won't happen.

opengever/core/upgrades/20230110082930_add_i_touched_behavior/upgrade.py
```diff
@@ -23,3 +24,22 @@ def migrate_touched_to_new_behavior(self):
                 # Reassigning the value to the new behavior is enought and
                 # fast. So there is also no need for a nightly upgrade step.
                 ITouched(obj).touched = dossier_schema.touched
+
+    def index_touched_on_worskspaces(self):
+        query = {'object_provides': 'opengever.workspace.interfaces.IWorkspace'}
+        for obj in self.objects(query, "Index touched on workspaces."):
+            if ITouched(obj).touched:
+                continue  # already migrated
```
njohner, line 32: same here, the upgrade step will not be applied twice, so I don't understand that statement.

opengever/core/upgrades/20230110082930_add_i_touched_behavior/upgrade.py
```diff
@@ -23,3 +24,22 @@ def migrate_touched_to_new_behavior(self):
...
+
+            # Get the last modified object within the workspace or the workspace
+            # itself to determine the current touched date.
+            ITouched(obj).touched = self.catalog_unrestricted_search(
+                {
+                    'path': {'query': '/'.join(obj.getPhysicalPath())},
+                    'sort_on': 'modified',
```
njohner, line 39: I think you should be using `changed` not `modified` here. `modified` is also set when an object is reindexed for example, so it does not contain the information you want IMO.

njohner (commented):
Just a few comments, but otherwise LGTM!
Just wondering whether we should also introduce `touched` for `WorkspaceFolder`s?

elioschmutz:
> Just a few comments, but otherwise LGTM! Just wondering whether we should also introduce `touched` for `WorkspaceFolder`s?

Providing it for folders makes no sense at the moment since we do not show the modified/touched date of folders somewhere... (or I missed a view)

opengever/core/upgrades/20230110082930_add_i_touched_behavior/upgrade.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def migrate_touched_to_new_behavior(self):
+        query = {'object_provides': 'opengever.dossier.behaviors.dossier.IDossierMarker'}
+        for obj in self.objects(query, "Migrate touched from IDossier to ITouched."):
+            dossier_schema = IDossier(obj)
+            if hasattr(dossier_schema, 'touched'):
+                # We do not need to reindex because the field value does not change.
+                # Reassigning the value to the new behavior is enought and
+                # fast. So there is also no need for a nightly upgrade step.
+                ITouched(obj).touched = dossier_schema.touched
```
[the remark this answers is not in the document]
elioschmutz: You're right 👍  I updated the upgradestep. I now also remove the old value.

opengever/core/upgrades/20230110082930_add_i_touched_behavior/upgrade.py
elioschmutz: I removed it

opengever/base/indexes.py
elioschmutz: 👍 I updated

opengever/core/upgrades/20230110082930_add_i_touched_behavior/upgrade.py
```diff
@@ -0,0 +1,41 @@
...
+            # Get the last modified object within the workspace or the workspace
+            # itself to determine the current touched date.
+            ITouched(obj).touched = self.catalog_unrestricted_search(
+                {
+                    'path': {'query': '/'.join(obj.getPhysicalPath())},
+                    'sort_on': 'changed',
+                    'sort_order': 'descending',
+                    'sort_limit': 1
+                }
+            )[0].modified
```
njohner, line 39: `.changed`, not `.modified`

opengever/core/upgrades/20230110082930_add_i_touched_behavior/upgrade.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def index_touched_on_worskspaces(self):
+        query = {'object_provides': 'opengever.workspace.interfaces.IWorkspace'}
+        for obj in self.objects(query, "Index touched on workspaces."):
+            # Get the last modified object within the workspace or the workspace
+            # itself to determine the current touched date.
+            ITouched(obj).touched = self.catalog_unrestricted_search(
+                {
+                    'path': {'query': '/'.join(obj.getPhysicalPath())},
+                    'sort_on': 'changed',
```
njohner, line 35: Sorry I just had a closer look, and ToDos don't seem to have a `changed` date. I don't know why though, everything seems to have a changed date except ToDos and ToDoLists (even `meetingagendaitem` has a changed date). So I'm a bit unsure what you should do to set the initial `touched` date...
I'm pretty sure it simply got forgotten to add the Changed date do ToDos... we can have a quick call or discuss that quickly at daily tomorrow.

njohner (commented):
I'm still unsure as to the initial value of `touched`. Otherwise LGTM!

opengever/core/upgrades/20230110082930_add_i_touched_behavior/upgrade.py
elioschmutz: As discussed: The `IChanged` behavior is missing for todos and todolists. I'll add it and reindex it

elioschmutz: @njohner 👍 updated

njohner (approved):
Perfect thanks, LGTM

PR #7657, opened 2023-01-18
Policyless ogds auth plugin

elioschmutz:
🚧 

For [CA-5196]

## Checklist

- [ ] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations


lukasgraf (approved):
LGTM 👍

opengever/setup/creation/__init__.py
```diff
@@ -0,0 +1 @@
+OGDS_GROUP_KEY_REPLACE_MARKER_PREFIX = 'OGDS_SYNC_LOOKUP-'
```
buchi, line 1: Maybe call it `OGDS_SYNC_TEMP_ID_`? (and I would always use the same separator character)

opengever/setup/creation/orgunit.py
```diff
@@ -49,3 +66,12 @@ def create_unit(self, item):
         site = api.portal.get()
         site.acl_users.portal_role_manager.assignRoleToPrincipal(
             'Member', item['users_group_id'])
+
+    def get_or_create_group_by_name(self, group_name):
+        group = Group.query.filter(Group.groupname == group_name).one_or_none()
+        if not group:
+            external_id = '{}-{}'.format(OGDS_GROUP_KEY_REPLACE_MARKER_PREFIX, group_name)
```
buchi, line 73: The prefix already ends with a dash. Are you intentionally adding a second one?

buchi (approved):

PR #7661, opened 2023-01-31
Provides a "WWW-Authenticate" response header for unauthorized requests to the @scan-in endpoint.

elioschmutz:
This PR adds the `WWW-Authenticate` response header for unauthorized requests to the `@scan-in` endpoint.

Two notes to the reviewer:

1. The header could be added automatically by the `IChallengePlugin` of the PAS by plone by default. The `HTTPBasicAuthHelper` Plugin would already add the header. But we never reach this code even if the plugin is activated. We do not use the `challenge`-method of the `IChallengePlugins`s at this point .
2. The response object would provide a `addHeader`: `response.addHeader('WWW-Authenticate', 'Basic')`. But this is not compatible with the testbrowser of `ftw.testbrowser` which provides its own `TestResponse` object. I implemented a direct key assignment for the header. Otherwise we'd need to ship a new version of `ftw.testbrowser` to make it compatible in testing.

See https://docs.plone.org/4/en/old-reference-manuals/pluggable_authentication_service/challenges.html for more infos about the `IChallengePlugin` of in plone pas.

For [CA-5239]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-5239]: https://4teamwork.atlassian.net/browse/CA-5239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-5239.other
+++ b/changes/CA-5239.other
@@ -1,0 +1,1 @@
+Provides a "WWW-Authenticate" response header for unauthorized requests to the @scan-in endpoint. [elioschmutz]
--- a/opengever/api/tests/test_scanin.py
+++ b/opengever/api/tests/test_scanin.py
@@ -226,3 +226,12 @@ class TestScanIn(IntegrationTestCase):
 
         expected_message = 'The scan-in destination does not exist.'
         self.assertIn(expected_message, browser.json.get('error').get('message'))
+
+    @browsing
+    def test_scanin_provides_www_authenticate_header_if_401(self, browser):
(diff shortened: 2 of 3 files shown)


lukasgraf (approved):
LGTM 👍

PR #7667, opened 2023-02-08
Reindex the reference number of mails during move operations.

njohner:
It seems we had only registered an event handler to reindex the reference number of moved documents but not for mails.

For [CA-5274]

## Checklist
- [x] Changelog entry
- Upgrade-Steps:

[CA-5274]: https://4teamwork.atlassian.net/browse/CA-5274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/dossier/tests/test_move_items.py
+++ b/opengever/dossier/tests/test_move_items.py
@@ -85,7 +85,7 @@ class TestMoveItems(IntegrationTestCase, MoveItemsHelper):
             name='maximum_dossier_depth',
             value=2,
             interface=IDossierContainerTypes
-            )
+        )
 
         self.move_items([self.resolvable_dossier],
                         source=self.leaf_repofolder,
@@ -302,7 +302,6 @@ class TestMoveItemsUpdatesIndexAndMetadata(SolrIntegrationTestCase, MoveItemsHel
         u'sequence_number',
         u'sequence_number_string',
         u'sortable_title',
-        u'Subject',
         u'Title',
         u'trashed',
         u'UID',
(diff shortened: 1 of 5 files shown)


opengever/dossier/tests/test_move_items.py
```diff
@@ -519,7 +519,236 @@ def test_move_document_indexdata_update(self, browser):
         unchanged_indexdata = [
             'allowedRolesAndUsers',
             'is_subdossier',
-            ]
+        ]
```
njohner, line 522: The diff shown here is a bit weird, because the second part of the tests is always the same. I did not modify this test, instead added 3 new ones.

lukasgraf (approved):
LGTM 👍 - mind the merge freeze though, I almost forgot again :)

And, just to make sure I understand: This only adds _reindexing_ of the reference number that was missing for mails. That's all that's needed because the source of truth for the reference number (in the annotations, via ReferencePrefixAdapters and all that) is already correct, and does not need to be updated.

This is because for non-folderish objects like documents and mails their local, most specifc part is location-independent (sequence number), and the location dependent parts are stored/tracked on the parent containers. And those are already correct because merely by moving the mails they point to a new parent, and therefore docs/mails do not need an event handler that fiddles with the actual reference number storage, unlike folderish objects -> reindexing is all that's needed.

Correct?

njohner:
Yes I did not think that far, but you are right. The local reference number for mails and documents is already correct after a move, and the indexer will reconstruct the complete reference number by walking up the acquisition chain.
I'll change the title of the PR, commit and changelog entry to clarify that it's only reindexing.

PR #7668, opened 2023-02-08
Handle broken references when generating the meeting minutes PDF.

njohner:
Agendaitem in workspace meetings can reference documents, which can be deleted, so that the reference ends up being broken. With this PR we now handle broken references by simply skipping them.

For [CA-5312]

## Checklist
- [x] Changelog entry
- [x] Link to issue (Jira or GitHub) and backlink in issue (Jira)

[CA-5312]: https://4teamwork.atlassian.net/browse/CA-5312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7669, opened 2023-02-08
Improve workspace deleter tests.

njohner:
Workspace deleter tests did not have the workspace feature activated when they were written (https://github.com/4teamwork/opengever.core/commit/7cfedb5beceeef8ff66283a6b91438e5f26dac79) so that we were actually testing against the wrong permission. We therefore had to use manager permissions to delete objects, when in fact workspace members are allowed to delete documents. So we switch here to testing with a workspace member.

There is no related issue for this, I just stumbled across this error while working on https://github.com/4teamwork/opengever.core/pull/7668

## Checklist 
- [ ] Changelog entry -> no CL entry
- [ ] Link to issue (Jira or GitHub) and backlink in issue (Jira) -> no issue


lukasgraf (approved):
LGTM 👍

PR #7670, opened 2023-02-09
Allow to define a custom dossier resolution precondition.

njohner:
We introduce 2 registry records allowing to define an additional dossier resolution rule and the related error thrown when the rule is broken. The rule is a TALES expression. In order to make custom properties available in the custom resolve condition, I had to add custom properties as a property on the dossier.

For [CA-5228]

## Checklist
- [x] Changelog entry

[CA-5228]: https://4teamwork.atlassian.net/browse/CA-5228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-5228.feature
+++ b/changes/CA-5228.feature
@@ -1,0 +1,1 @@
+Allow to define a custom dossier resolution precondition. [njohner]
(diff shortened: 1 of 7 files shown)


lukasgraf (approved):
LGTM! 👍

PR #7672, opened 2023-02-16
Allow to define a custom dossier resolution after transition hook.

njohner:
SPDAG want to track the first resolution date on their dossiers. To avoid adding a new metadata field which will only be used by them but pollute all other deployments, we are adding the tools necessary to implement the desired behaviour with custom-fields. We are therefore introducing a registry entry `resolver_custom_after_transition_hook` allowing to define a TALES expression that will be executed in a post transition hook when resolving a dossier.

The actual field, upgrade step and Tales expressions that allow to define the desired behaviour for SPDAG are [defined in their policy](https://github.com/4teamwork/opengever.spdag/pull/30).

For [CA-5229](https://4teamwork.atlassian.net/browse/CA-5229)

## Checklist

- [x] Changelog entry

[CA-5229]: https://4teamwork.atlassian.net/browse/CA-5229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/dossier/resolve.py
+++ b/opengever/dossier/resolve.py
@@ -102,6 +102,19 @@ class ResolveDossierTransitionExtender(TransitionExtender):
 
     def after_transition_hook(self, transition, disable_sync, transition_params):
         AfterResolveJobs(self.context).execute()
+        self.execute_custom_after_transition_hook(transition, transition_params)
+
+    def execute_custom_after_transition_hook(self, transition, transition_params):
+        custom_after_transition_hook = api.portal.get_registry_record(
+            'resolver_custom_after_transition_hook', IDossierResolveProperties)
+
+        if not custom_after_transition_hook:
+            return
+
+        portal = api.portal.get()
+        ec = createExprContext(folder=self.context, portal=portal, object=self.context)
+        expr = Expression(custom_after_transition_hook)
+        expr(ec)
 
 
 class LockingResolveManager(object):
--- a/changes/CA-5229.feature
+++ b/changes/CA-5229.feature
@@ -1,0 +1,1 @@
+Allow to define a custom dossier resolution after transition hook. [njohner]
--- a/opengever/core/upgrades/20230214171647_add_resolver_custom_after_transition_hook_registry_field/registry.xml
+++ b/opengever/core/upgrades/20230214171647_add_resolver_custom_after_transition_hook_registry_field/registry.xml
@@ -1,0 +1,3 @@
+<registry purge="False">
+  <records interface="opengever.dossier.interfaces.IDossierResolveProperties" />
+</registry>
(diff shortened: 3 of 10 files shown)


opengever/dossier/resolve.py
```diff
@@ -115,6 +115,7 @@ def execute_custom_after_transition_hook(self, transition, transition_params):
         ec = createExprContext(folder=self.context, portal=portal, object=self.context)
         expr = Expression(custom_after_transition_hook)
         expr(ec)
+        self.context.reindexObject()
```
lukasgraf, line 118: Couldn't this go into the `Dossier.set_custom_property()` method, possibly with a kwarg `reindex=False`? Or is just invoking the reindexing operation from a Tales expression in any way causing an issue?

lukasgraf (approved):
LGTM 👍 - just a question about where to do the reindexing.

opengever/dossier/resolve.py
njohner: no it could go into `Dossier.set_custom_property()`, but my thought was that if we change some other attribute in the tales expression, it would not get reindexed. So the reindexing issue is not really related to modifying a custom property, but to modifying something in that transition hook. Of course, currently this difference is hypothetical, as custom properties are the only things we can modify on the dossier in the TALES expression.

lukasgraf: Interesting, I was going down the same line of thinking but with a different conclusion :) If the expression would cause a modification on an object _other_ than the dossier (also very hypothetical), the reindexing of the dossier wouldn't help. So it sort of needs to be the responsibility of whatever mechanism is used to cause the modification to also do the reindexing.

But TBH, this is a really minor point and doesn't really make a difference as of now, so it's fine with me either way.

njohner: Yes I also thought about that. I mean it is probably possible to access the documents in the dossier in the TALES expression (with `get_contained_documents`) and do something with them.
So I'd say yes, I can change it. Also note that I did not find a way to only reindex the custom field, so I would still be reindexing the full object.

lukasgraf: Yeah, that full `reindexObject()` is proably undoing some of the performance gains we got from introducing the nightly AfterResolveJobs (for SPD at least). It did find an [old upgrade step](https://github.com/4teamwork/opengever.core/blob/master/opengever/core/upgrades/20210615143823_reindex_customproperties_for_dossier/upgrade.py#L18-L29) where custom properties were indexed a bit more selectively, but that one doesn't really inspire confidence either in that being "the right way" ;)

njohner:
@lukasgraf I've added a commit allowing to reindex a specific custom field. I then added the possibility to reindex the modified custom field in `set_custom_property`, which allowed me to not reindex the object in the custom transition hook. Can you have another look at the last 4 commits?

lukasgraf (approved):
Nice catch! LGTM 👍

PR #7673, opened 2023-02-16
Fix LDAP properties plugin upgrade step to work in deployments without LDAP.

lukasgraf:
This upgrade step assumed that there's always an `ldap` plugin in GEVER deployments. This is not the case anymore however for policyless deployments, so it broke upgrades on those.

For [CA-5355](https://4teamwork.atlassian.net/browse/CA-5355)

## Checklist

- [x] Changelog entry

[CA-5355]: https://4teamwork.atlassian.net/browse/CA-5355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-5355.bugfix
+++ b/changes/CA-5355.bugfix
@@ -1,0 +1,1 @@
+Fix LDAP properties plugin upgrade step to work in deployments without LDAP. [lgraf]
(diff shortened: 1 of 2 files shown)


njohner (approved):
LGTM

PR #7675, opened 2023-02-17
Fix ordering of subtasks in @tasktree for nested task processes.

njohner:
Subtasks in sequential tasks are ordered according to position in parent while they are ordered according to creation date in parallel tasks. For nested tasktemplates, with a mixture of parallel and sequential processes, the ordering was broken as it was determined once and for all on the main task. This is fixed here by determining the sort order for subtasks at each nesting level.

For [CA-5160]

## Checklist
- [x] Changelog entry

[CA-5160]: https://4teamwork.atlassian.net/browse/CA-5160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/api/tasktree.py
+++ b/opengever/api/tasktree.py
@@ -1,6 +1,7 @@
 from Acquisition import aq_parent
 from ftw.solr.interfaces import ISolrSearch
 from ftw.solr.query import make_filters
+from ftw.solr.query import make_path_filter
 from opengever.base.browser.navigation import make_tree_by_url
 from opengever.base.interfaces import IOpengeverBaseLayer
 from opengever.base.solr import OGSolrContentListing
@@ -23,6 +24,9 @@ class TaskTree(object):
     def __init__(self, context, request):
         self.context = context
         self.request = request
+        self.solr = getUtility(ISolrSearch)
+        self.fieldlist = [
+            'Title', 'portal_type', 'path', 'review_state', 'object_provides']
 
     def __call__(self, expand=False):
         result = {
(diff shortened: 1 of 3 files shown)


opengever/api/tasktree.py
```diff
@@ -86,22 +90,31 @@ def extend_tree_with_addable_information(self, tree, solr_items_per_url, parent_
...
             },
             object_provides=ITask.__identifier__,
         )
-        fieldlist = ['Title', 'portal_type', 'path', 'review_state', 'object_provides']
+        is_sequential = IContainSequentialProcess.__identifier__ in item.get("object_provides")
         sort = 'getObjPositionInParent asc' if is_sequential else 'created asc'
-        resp = solr.search(
-            filters=filters, start=0, rows=1000, sort=sort,
-            fl=fieldlist)
+        resp = self.solr.search(
```
elioschmutz, line 104: Do we know if the current `item` has subtasks or not here?

The current implementation would perform 6 solr-requests for the following structure, but 2 would be enough...:

- Main Task
  - Subtask1
  - Subtask2
    - Subtask2.1
    - Subtask2.2
    - Subtask2.3
  - Subtask3

elioschmutz (commented):
Looks good in general, but I think, we should optimize the recursive query. What do you think?

opengever/api/tasktree.py
njohner: Yes you're right, I've fixed that in a new commit.

elioschmutz (approved):

PR #7677, opened 2023-02-21
Nj/ca 5261/handle deleted documents

njohner:
[Handle missing resources when deleting a workspace member.(https://github.com/4teamwork/opengever.core/commit/b9b052693b6940ca4ac21151b84edb8f65cb690a) 

We were not deleting resources and subscription when deleting an object, which led to resources that cannot be resolved. As we do not want to delete all resources with invalid Oguids, we need to handle resources with invalid Oguids when deleting a participant from a workspace. Starting now we also clean-up the resources and subscriptions when deleting an object
For [CA-5261]

## Checklist
- [x] Changelog entry
- Bug fixed:


lukasgraf (approved):
LGTM! 👍 - :shipit:

PR #7678, opened 2023-02-22
Fix various issues arising when the ContactFolder is missing.

njohner:
With this PR we fix various issues arising when the `ContactFolder` is missing. This should become the standard case in the future, as new deployments get created without a `ContactFolder` and the `ContactFolder` will get deleted for older deployments (migration towards KuB). We therefore delete the `ContactFolder` from the fixture and adapt the tests accordingly. We then fix a few issues so that things don't break down when the `ContactFolder` is missing notably:
- We don't throw an error anymore in `get_contactfolder_url` when the `ContactFolder` is missing
- Modify `get_profile_url` for Team Actors to return None when the `ContactFolder` is missing
- We also readd the list of teams on the userdetails view when the `ContactFolder` is missing. I'm not sure why that got dropped, I think listing the teams without a URL makes more sense than not showing the teams at all.

For [CA-5216]

## Checklist
- [x] Changelog entry
- Bug fixed:

[CA-5216]: https://4teamwork.atlassian.net/browse/CA-5216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/api/tests/test_dossier_participations.py
+++ b/opengever/api/tests/test_dossier_participations.py
@@ -1,4 +1,5 @@
 from ftw.testbrowser import browsing
+from opengever.contact.tests import create_contacts
 from opengever.dossier.behaviors.participation import IParticipationAware
 from opengever.dossier.interfaces import IDossierParticipants
 from opengever.dossier.participations import IParticipationData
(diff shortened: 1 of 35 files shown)


opengever/api/tests/test_dossier_participations.py
```diff
@@ -226,13 +227,7 @@ def test_participants_are_sorted_by_title(self, mocker, browser):
...
 
 
-class TestParticipationsPost(IntegrationTestCase):
-
-    def setUp(self):
-        super(TestParticipationsPost, self).setUp()
-        self.valid_participant_id = self.regular_user.getId()
-        with self.login(self.regular_user):
-            self.valid_participant_id2 = 'contact:{}'.format(self.franz_meier.getId())
+class ParticipationsPostTestMixin():
```
njohner, line 230: I extract this to a mixin to avoid creating the ContactFolder and contacts in `TestParticipationsPostWithKubFeatureEnabled`

njohner:
Yes no details view for teams in the classic ui. No listing for the teams either. All of that was on the contact folder. In the new UI everything should work, there are team listings, detail views, etc.

PR #7681, opened 2023-02-28
Fix @tasktree for users lacking view permission on main task.

njohner:
We now simply return an empty tree when the user has no permissions on the main task.

_PR-Description: should contain all the information necessary for an outsider to understand the change without looking at the code or the issue!_

We now simply return an empty tree when the user has no permissions on the main task. Raising an error and displaying it in the gever-ui was not easily implementable, as this is an expansion, and raising an error in an expansion will make the full request fail.

For [CA-5160]

## Checklist
- [x] Changelog entry
- Bug fixed:
  - [x] Resolved any Sentry issues caused by this bug : https://sentry.4teamwork.ch/organizations/sentry/issues/96052/events/7122cd767f6944eb8323b60e6a8618dc/?project=17&statsPeriod=14d


elioschmutz (approved):
👍 

A changelog is missing. And have you checked it with the new ui? Does it work as expected?

njohner:
Yes sorry, changelog got lost when I rebased...

lukasgraf (approved):
LGTM 👍

PR #7682, opened 2023-03-01
GitHub Actions: use scoped cache in builder.

jone:
The GitHub actions cache supports scopes. By using scopes for different images in the builder step, the caches are more efficient and this should speed up image building.


njohner (approved):
LGTM

PR #7683, opened 2023-03-02
Add @solrlivesearch endpoint and improve solr configuration. 

njohner:
The behavior of the livesearch in Gever is currently pretty confusing, as only searches for full tokens will lead to results. This typically leads to having 0 results after typing a few characters, but results will appear again later, once a word is complete.
We propose here to fix this by specifically adding wildcards to the terms of the query and doing some tokenization as it will not happen for wildcard queries.

I also switch the default operator the `AND`, IMO this is the expected behavior, getting fewer results when searching for more terms (this is also what google does). I had to set the `q.op` parameter and not the `mm`, as this otherwise prevents from using operators in the queries.

I've also added boosting of recently `changed` objects, which got lost when we switched to the edismax parser. I think it is ok to use `changed` which is defined for all relevant portal types (everything except `ContactFolder`, `TemplateFolder`, `InboxContainer`, `CommitteeContainer`, `PrivateFolder`, `PrivateRoot`, `RepositoryRoot`, `WorkspaceRoot`).

I also add the possibility to simply retrieve the preprocessed query, which could be used when the user hits enter in the livesearch to jump to a normal search with the preprocessed query and display that to the user.

I also switch the metadata field type to `text_general`. Having the same tokenization for metadata as for other text fields
that are queried by default allows consistent results for the @LiveSearch endpoint, notably to query reference numbers when using the grouped_by_three formatter. Indeed the reference number of all objects except documents and mails is in the `SearchableText`, but for documents and mails it's only in the `metadata`

Tests are failing because we would need to build and use a new image of solr to include the changes in solr configuration.

For [CA-5477] and [CA-5158]

## Checklist
- [x] Changelog entry

[CA-5158]: https://4teamwork.atlassian.net/browse/CA-5158?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CA-5477]: https://4teamwork.atlassian.net/browse/CA-5477?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-5158.feature
+++ b/changes/CA-5158.feature
@@ -1,0 +1,1 @@
+Add @solrlivesearch endpoint and improve solr configuration. [njohner]
--- a/opengever/api/configure.zcml
+++ b/opengever/api/configure.zcml
@@ -653,6 +653,22 @@
   <plone:service
       method="GET"
       for="zope.interface.Interface"
+      factory=".solrsearch.SolrLiveSearchGet"
+      name="@solrlivesearch"
+      permission="zope2.View"
+      />
+
+  <plone:service
+      method="POST"
+      for="zope.interface.Interface"
+      factory=".solrsearch.SolrLiveSearchGet"
+      name="@solrlivesearch"
+      permission="zope2.View"
+      />
+
+  <plone:service
+      method="GET"
+      for="zope.interface.Interface"
       factory=".solrsearch.TeamraumSolrSearchGet"
       name="@teamraum-solrsearch"
       permission="zope2.View"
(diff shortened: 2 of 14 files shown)


njohner:
Tests fail because the solr container does not include the changes I've made in the solr config

lukasgraf (approved):
LGTM 👍

PR #7686, opened 2023-03-13
ogcore docker image: Send ftw.upgrade's stats log to /dev/null:

lukasgraf:
We symlink ftw.upgrade's `upgrade_stats.csv` to /dev/null, because otherwise ftw.upgrade will try to create it during upgrades, and will fail because of insufficient permissions.

Because we don't actually care about the contents of this logfile, we can just discard its content instead of redirecting it somewhere else.

For [CA-5405](https://4teamwork.atlassian.net/browse/CA-5405)

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/CA-5405.other
+++ b/changes/CA-5405.other
@@ -1,0 +1,1 @@
+ogcore docker image: Send ftw.upgrade's stats log to /dev/null [lgraf]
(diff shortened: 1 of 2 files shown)


njohner (approved):
LGTM

PR #7689, opened 2023-03-14
Add duplicate-strategies to the @globalindex endpoint.

elioschmutz:
This PR implements an additional option called `duplicate_strategy` for the `@globalindex` endpoint.

This options allows us to change the behavior of which task should be returned for successor/predecessor task pairs.

We need this, because we sometimes need to change this behavior depending on the purpose of the request.

One use case is i.e. to get all tasks which I issued. In that case, I usually want the `predecessor_task` strategy which would always return the main task if there is also a predecessor task.

Providing this feature allows us to fix an issue in the frontend where the user does not see his tasks if requesting the `@globalindex` endpoint from a third admin unit.

For [CA-5446]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-5446]: https://4teamwork.atlassian.net/browse/CA-5446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/public/dev-manual/api/globalindex.rst
+++ b/docs/public/dev-manual/api/globalindex.rst
@@ -84,6 +84,17 @@ Facetten
 ~~~~~~~~
 - ``facets``: Liste von Feldern für die Facetten Wertebereiche zurückgegeben werden sollen.
 
+Duplikate
+~~~~~~~~~
+Bei mandantenübergreifenden Aufgaben wird nur eine der beiden Aufgabe zurückgebeben. Über die ``duplicate_strategy`` option kann gesteuert werden,
+welche Aufgabe bei Aufgaben-Pärchen zurückgegeben werden soll.
+
+Folgende Strategien stehen zur Verfügung:
+- ``duplicate_strategy``:
+  - ``local`` (Standard): Gibt immer die Aufgabe vom aktuellen Mandanten zurück. Wenn sich das Pärchen auf anderen Mandanten befindet, wird keine der beiden Aufgaben zurückgegeben.
+  - ``predecessor_task``: Gibt immer die Haupt-Aufgabe zurück, unabhängig davon, auf welchem Manden man sich aktuell befindet.
+  - ``successor_task``: Gibt immer die Nachfolge-Aufgabe zurück, unabhängig davon, auf welchem Manden man sich aktuell befindet.
+
 
 **Beispiel: Filtern nach erledigten und abgeschlossenen Aufgaben:**
 
@@ -111,4 +122,11 @@ Facetten
   .. sourcecode:: http
 
     GET /@globalindex?facets:list=review_state&facets:list=responsible HTTP/1.1
+    Accept: application/json
+
+**Beispiel: Bei Duplikaten immer die Hauptaufgabe zurückgeben:**
+
+  .. sourcecode:: http
+
+    GET /@globalindex?duplicate_strategy=predecessor_task HTTP/1.1
     Accept: application/json
--- a/opengever/globalindex/model/task.py
+++ b/opengever/globalindex/model/task.py
@@ -591,16 +591,13 @@ class TaskQuery(BaseQuery):
                 TaskPrincipal.principal.in_(principals))
         return self.filter(Task.task_id.in_(principal_query))
 
-    def avoid_duplicates(self, admin_unit_id=None):
+    def avoid_duplicates(self):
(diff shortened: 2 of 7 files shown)


opengever/globalindex/model/task.py
```diff
@@ -566,6 +566,11 @@ def get_main_task_title(self):
             return main_task.title
 
 
+AVOID_DUPLICATES_STRATEGY_CURRENT_ADMIN_UNIT = 'current_admin_unit'
```
lukasgraf, line 569: Rename to '...local'

opengever/globalindex/model/task.py
```diff
@@ -591,22 +596,51 @@ def restrict(self):
...
 
+        Available strategies are:
+
+        successor_task:
+        If a task has a successor task, always return the successor task
+
+        predecessor_task:
+        If a task has a predecessor, always return the predecessor task
+
+        fallback strategy:
```
lukasgraf, line 612: Document 3rd (local) strategy

opengever/globalindex/model/task.py
```diff
@@ -591,22 +596,51 @@ def restrict(self):
...
-                Task.admin_unit_id == get_current_admin_unit().id()
+        elif strategy == AVOID_DUPLICATES_STRATEGY_PREDECESSOR_TASK:
+            return self.filter(
+                or_(
+                    and_(Task.predecessor == None, Task.successors == None),  # noqa
+                    Task.successors != None  # noqa
+                )
+            )
+
+        else:
```
lukasgraf, line 634: As discussed: Just check with elif for all 3 strategies, and just raise NotImplementedError in the else if none match.

docs/public/dev-manual/api/globalindex.rst
```diff
@@ -84,6 +84,19 @@ Facetten
...
+~~~~~~~~~
+Bei mandantenübergreifenden Aufgaben wird nur eine Aufgabe zurückgebeben. Über die ``duplicate_strategy`` option kann gesteuert werden,
+welche Aufgabe bei Aufgaben-Pärchen zurückgegeben werden soll.
+
+Standardmässig wird immer die Aufgabe vom aktuellen Mandanten zurückgegeben.
+Wenn sich das Pärchen auf anderen Mandanten befindet, wird keine der beiden Aufgaben zurückgegeben.
+
+Folgende speziellen Strategien stehen zur Verfügung:
+- ``duplicate_strategy``:
+  - ``successor_task``: Gibt immer die Haupt-Aufgabe zurück, unabhängig davon, auf welchem Manden man sich aktuell befindet.
```
lukasgraf, line 97: Document 3rd (local) strategy, and switch Haupt/Nachfolge-Aufgabe.

lukasgraf (approved):
LGTM 👍

njohner (commented):
The code looks good to me, but I'm a bit unsure whether any of these strategies is what we really want. Can we maybe talk that over, see what listings we need and why a given strategy would make sense for that listing?

elioschmutz:
@njohner thanks for your feedback. I already talked with @lukasgraf but we can have a call as well if needed. I try to explain it here to have it documented for the future 

The `@globalindex` endpoint is used for two listings in the frontend: "My tasks" and "Overview of pending tasks" (Pendenzenkontrolle).

We use it mainly for the "My Tasks" listing. The "Overview of pending tasks" already filters all tasks by a specific org-unit. 

I the "My tasks" listing, we have two predefined filters for "Issued" and "Mine":

Bildschirmfoto 2023-03-16 um 09 44 16

The `predecessor_task`-strategy will be used for the "Issued"-filter (I am the issuer of the task) and the `successor_task`-strategy will be used for the "Mine"-fIlter (i am the responsible of the task).

We can't decide in the backend which strategy to use based on the query. It strongly depends on the use case of the listing itself. 

Here is an example of a query where we would not know what to do in the backen: if we list all "Mine" (i am the responsible) tasks filtered for a specific issuer. The backend would not know if it should return the predecessor or the successor. The frontend knows it because we used the "Mine" filter and not the generic `responsible` filter.

njohner (approved):
As discussed, LGTM

PR #7700, opened 2023-04-04
Defer TUS lock test by another 6 months.

njohner:
This has now been fixed in OfficeConnector, but we cannot reintroduce the check in opgenver.core before all customers have a new version of OfficeConnector installed.

## Checklist
- [ ] Changelog entry: No CL
- [ ] Link to issue (Jira or GitHub) and backlink in issue (Jira): No Jira


phgross (approved):
👍

PR #7704, opened 2023-04-11
Fix flaky test for task deadline default values.

njohner:
`datetime.today().date()` does not always return the same date as `date.today()`. We now consistently use date.today() for all deadline calculations.
I could have fixed the tests, but we already calculate the deadline for tasktemplates using `date.today()`, so decided to make it consistent instead.

This should fix the failing nightly build tests, see https://ci.4teamwork.ch/builds/555422/tasks/1097389

## Checklist
- [ ] Changelog entry


lukasgraf (approved):
LGTM 👍

PR #7716, opened 2023-05-02
Hide and skip inactive users in the @watchers endpoints.

phgross:
Inactive users are no longer displayed as observers and are also no longer available for selection.

For [CA-5400]

## Checklist
- [x] Changelog entry

[CA-5400]: https://4teamwork.atlassian.net/browse/CA-5400?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-5400.bugfix
+++ b/changes/CA-5400.bugfix
@@ -1,0 +1,1 @@
+Hide and skip inactive users in the @watchers endpoints. [phgross]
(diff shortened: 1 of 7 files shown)


njohner (commented):
If we don't display the inactive users as watchers we should also make sure that they do not get mails when a notification is sent. Is that the case? I haven't looked in detail but I think you'd need to adapt https://github.com/4teamwork/opengever.core/blob/master/opengever/activity/dispatcher.py#L29-L48

phgross:
@njohner you're right, could you take another look to the last two commits.

njohner (approved):
LGTM

PR #7717, opened 2023-05-04
Handle review_state mismatch between successor/predecessor pairs gracefully

phgross:
In some cases there is a mismatch between successor/predecessor pairs, which leads to a WorkflowException when trying to sync the workflow transition to the remote site, via remote_request. This cases needs to be fixed manually via debug mode.
To prevent this support effort, we try to fix this review_state mismatch automatically, when a workflow transition sync fails.

For [CA-5363]

## Checklist
- [x] Changelog entry

[CA-5363]: https://4teamwork.atlassian.net/browse/CA-5363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/task/response_syncer/workflow.py
+++ b/opengever/task/response_syncer/workflow.py
@@ -1,8 +1,10 @@
+from DateTime import DateTime
 from opengever.base.security import elevated_privileges
 from opengever.task.response_syncer import BaseResponseSyncerReceiver
 from opengever.task.response_syncer import BaseResponseSyncerSender
 from opengever.task.response_syncer import ResponseSyncerSenderException
 from plone import api
+from Products.CMFCore.WorkflowCore import WorkflowException
 from Products.CMFPlone.utils import safe_unicode
 from zope.event import notify
 from zope.lifecycleevent import ObjectModifiedEvent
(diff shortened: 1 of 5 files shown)


opengever/task/response_syncer/workflow.py
```diff
@@ -59,7 +61,51 @@ def _update(self, transition, text):
...
             # a snycing-loop
-            wftool.doActionFor(
-                self.context, transition, disable_sync=True, transition_params=data)
+            try:
+                wftool.doActionFor(
+                    self.context, transition, disable_sync=True, transition_params=data)
+            except WorkflowException:
+                missmatch_fixed = self._fix_review_state_missmatch()
+                if missmatch_fixed:
+                    wftool.doActionFor(
```
njohner, line 69: I'm a bit confused here. Isn't the case we mostly have that the successor task is closed and the state was not synced back to the predecessor? Then someone tries to close the task on the predecessor side. In that case `_fix_review_state_missmatch` would close the predecessor task, which is what we want, but then you try to close the task again here no? Wouldn't that throw an error?

njohner: Sorry, as discussed, we are in the syncer here, which I did not realise, so that will work in any case, as the set the state to the state of the object for which the transition is being executed, so the transition will then exist. So LGTM!

njohner (approved):
LGTM

PR #7718, opened 2023-05-04
Reindex maybe not properly indexed files

elioschmutz:
This PR reindexes files which may be not properly indexed.

The index is out of sync for files, mainly the filename index. Because we do not really know all the indexes which could be out of sync, we'll just reindex the full object.

The reason why these files where not properly indexed is not yet fully clear. It could have happened when introducing the `filename` index

Since most of the files are properly indexed, we do not use a nightly job for this upgradestep.

Slack-Reference: https://4teamwork.slack.com/archives/C01BHEGBVU1/p1683206204789769

## Before
Bildschirmfoto 2023-05-04 um 17 07 23

## After
Bildschirmfoto 2023-05-04 um 17 18 04

For [CA-5726]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-5726]: https://4teamwork.atlassian.net/browse/CA-5726?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/core/upgrades/20230504163300_reindex_not_properly_indexed_files/upgrade.py
+++ b/opengever/core/upgrades/20230504163300_reindex_not_properly_indexed_files/upgrade.py
@@ -1,0 +1,37 @@
+from ftw.solr.interfaces import ISolrSearch
+from ftw.upgrade import ProgressLogger
+from ftw.upgrade import UpgradeStep
+from opengever.base.solr import OGSolrDocument
+from opengever.core.upgrade import NightlyIndexer
+from zope.component import getUtility
+
+
+class ReindexNotProperlyIndexedFiles(UpgradeStep):
+    """Reindex not properly indexed files.
+    """
+
+    deferrable = True
+
+    def __call__(self):
+        self.install_upgrade_profile()
+        self.reindex()
+
+    def reindex(self):
+        solr = getUtility(ISolrSearch)
+        fq = [
+            "portal_type:(opengever.document.document OR ftw.mail.mail)",
+            "-filename:['' TO *]"]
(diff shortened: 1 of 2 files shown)


opengever/core/upgrades/20230504163300_reindex_not_properly_indexed_files/upgrade.py
```diff
@@ -0,0 +1,32 @@
...
+            "portal_type:(opengever.document.document OR ftw.mail.mail)",
+            "-filename:['' TO *]"]
+        fl = ['filename', 'path']
+
+        nrows = solr.search(fq=fq, rows=0).num_found
+        resp = solr.search(fq=fq, fl=fl, rows=nrows)
+
+        for doc in ProgressLogger('Maybe reindex not properly indexed files', resp.docs):
+            obj = OGSolrDocument(doc).getObject()
+            # We stillneed to check if the filename is really empty becuase the
```
phgross, line 28: typo: stillneed

phgross (commented):
Please, make it at least deferrable, this saves time and nerves during updaten. 

If possible i would also reuse the NightlyIndexer, makes updating even less painfull.

elioschmutz:
@phgross I updated the PR. It is now deferrable and it uses the `NightlyIndexer` 👍

phgross (approved):
👍 LGTM

PR #7720, opened 2023-05-08
Modify ech0147 test import data by adding two whitespaces to a document title

elioschmutz:
Having a document with two whitespaces in the document title will break the ech0147 import because the filename gets normalized when parsed by pyxb. The double whitespace will be reduced to a single whitespace. The file can no longer be found:

Pointer: https://github.com/4teamwork/opengever.core/blob/master/opengever/ech0147/browser/import.py#L99

## ZIP Content:
Bildschirmfoto 2023-05-08 um 17 04 18
Bildschirmfoto 2023-05-08 um 17 04 01

## Normalized Filename
Bildschirmfoto 2023-05-08 um 17 05 13

For [CA-5673]

## Checklist

- [ ] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-5673]: https://4teamwork.atlassian.net/browse/CA-5673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


elioschmutz:
The eCH-0147 defines the `pathFileName` as an `xs:token` which normalizes the whitespaces by default.

- `eCH-0147` interface (eCH-039 is the base of eCH-0147): https://www.ech.ch/sites/default/files/dosvers/hauptdokument/STAN_d_DEF_2017-09-25_eCH-0039_V3.1_E-Government-Schnittstelle.pdf
- `xs:token` definition: https://www.oreilly.com/library/view/xml-schema/0596002521/re95.html

So the import behaves correctly. The filename should already be normalized when exported.

PR #7723, opened 2023-05-16
Fix ech0147 import if umlauts are used in filenames

elioschmutz:
It was not possible to import an ech0147 zip-file if umlauts were used for the containing filenames.

According to the ech0147 which is based on the ech0139 - https://www.ech.ch/sites/default/files/dosvers/hauptdokument/STAN_d_DEF_2017-09-25_eCH-0039_V3.1_E-Government-Schnittstelle.pdf), the `filePathName` can contain any valid xml-string: https://www.data2type.de/xml-xslt-xslfo/xml-schema/datentypen-referenz/xs-token/ => See base-type "xs:string"

The python zipfile implementation provides a list of files based on its filenames in utf-8. The representation of umlauts can be either by directly using the unicode-sign (ä) or by using the base-sign and diaeresis (a + ¨) which is called "normal form D (NFD)": https://docs.python.org/2/library/unicodedata.html#unicodedata.normalize

```
>>> import unicodedata
>>> unicodedata.normalize('NFD', u'ä')
u'a\u0308'
>>> unicodedata.normalize('NFD', u'ä').encode('utf-8')
'a\xcc\x88'
>>> unicodedata.normalize('NFC', u'ä')
u'\xe4'
>>> unicodedata.normalize('NFC', u'ä').encode('utf-8')
'\xc3\xa4'
```
The zipfile-implementation uses utf-8 encoded strings in the NFD-form to represent its filenames.

Bildschirmfoto 2023-05-16 um 11 29 56

That means, we have to transform our own strings, which are usually represented by the NFC-normal-form to the NFD-normal-form to be able to lookup filenames with umlauts.

This PR also fixes an issue where an unicode error was raised when constructing the error-message of a not found file:

**Before**
Bildschirmfoto 2023-05-16 um 10 57 54

**After**
Bildschirmfoto 2023-05-16 um 10 58 17

For [CA-5673]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-5673]: https://4teamwork.atlassian.net/browse/CA-5673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


njohner (approved):
LGTM, though I think you mixed up the screenshots in the the PR description.

njohner (approved):
LGTM, though I think you mixed up the screenshots in the the PR description.

PR #7726, opened 2023-05-26
Implement IGroupIntrospection for OGDS auth plugin

buchi:
`IGroupIntrospection` is needed to allow setting permissions for OGDS groups in the sharing view. Also makes OGDS groups show up in the Plone control panel.

The methods aren't cached currently, as we expect they are used rarley. Caching could be easily done though.

`getGroups()` is implemented in a naive way and could be optimized to not query each group separately, but we are not expecting to use it.

For [CA-5766](https://4teamwork.atlassian.net/browse/CA-5766)

## Checklist

- [x] Changelog entry

[CA-5766]: https://4teamwork.atlassian.net/browse/CA-5766?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-5766a.other
+++ b/changes/CA-5766a.other
@@ -1,0 +1,1 @@
+Implement IGroupIntrospection for OGDS auth plugin. [buchi]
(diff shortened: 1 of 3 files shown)


phgross (approved):
👍

PR #7728, opened 2023-05-30
Index related items for documents and tasks

njohner:
With this PR we index related items for documents and tasks in Solr. This can be used for example to display contained and related documents of a task (see test `test_get_related_and_contained_documents`).

Tests are failing because we would need to build and use a new image of solr to include the changes in solr configuration.

For [CA-5463]

## Checklist
- [x] Changelog entry
- Upgrade-Steps:
- New functionality:
  - [ ] for `document` also works for `mail` -> mails do not have related items.

[CA-5463]: https://4teamwork.atlassian.net/browse/CA-5463?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/base/utils.py
+++ b/opengever/base/utils.py
@@ -328,6 +328,25 @@ def unrestrictedUuidToCatalogBrain(uuid):
     return result[0]
 
 
+def unrestrictedPathToCatalogBrain(path):
+    """Given a path, attempt to return a catalog brain.
+    """
+
+    site = getSite()
+    if site is None:
+        return None
+
+    catalog = api.portal.get_tool('portal_catalog')
+    if catalog is None:
+        return None
+
+    result = catalog.unrestrictedSearchResults(path=path)
+    if len(result) != 1:
+        return None
+
+    return result[0]
+
+
 def get_date_with_delta_excluding_weekends(start_date, days_delta):
     calculated_delta = 0
     end_date = start_date
--- a/opengever/mail/tests/test_extract_attachments.py
+++ b/opengever/mail/tests/test_extract_attachments.py
@@ -9,6 +9,8 @@ from opengever.propertysheets.storage import PropertySheetSchemaStorage
 from opengever.testing import FunctionalTestCase
 from opengever.testing import IntegrationTestCase
 from opengever.testing import obj2brain
+from opengever.testing import solr_data_for
+from opengever.testing import SolrIntegrationTestCase
 from pkg_resources import resource_string
 import transaction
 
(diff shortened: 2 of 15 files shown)


opengever/base/utils.py
```diff
@@ -328,6 +328,25 @@ def unrestrictedUuidToCatalogBrain(uuid):
...
+    """Given a path, attempt to return a catalog brain.
+    """
+
+    site = getSite()
+    if site is None:
+        return None
+
+    catalog = getToolByName(site, 'portal_catalog', None)
+    if catalog is None:
+        return None
```
elioschmutz, line 341: `api.portal.get_tool('portal_catalog')`

opengever/testing/helpers.py
```diff
@@ -241,8 +244,16 @@ def clear(self):
 
 class SolrTestMixin(object):
 
-    def commit_solr(self, after_commit=False):
-        SolrServer.get_instance().commit(after_commit=after_commit)
+    def commit_solr(self, avoid_blob_extraction=False):
```
elioschmutz, line 247: We loose the `after_commit` parameter. Is that a desired change?

opengever/mail/tests/test_extract_attachments.py
```diff
@@ -172,6 +161,32 @@ def test_creates_document_in_parent_submitted_proposal(self, browser):
...
+        self.assertEqual(1, len(children["added"]))
+        doc = children["added"].pop()
+
+        self.assertEquals('word_document', doc.Title())
+        self.assertEquals(doc.document_date, date.today())
+        self.assertEquals(obj2brain(doc).document_date, date.today())
+        self.assertEqual([mail], doc.related_items())
+        self.assertEqual([mail.UID()], solr_data_for(doc, "related_items"))
+
+
```
elioschmutz, line 189: I think these changes belongs to the previous commit.

elioschmutz (approved):
👍

opengever/testing/helpers.py
njohner: Yes we don't need it. We only used it to actually avoid blob extraction. Doing things in an `after_commit` hook in integration tests does not make sense, as we never commit, hence after commit hooks never get exectued.

opengever/mail/tests/test_extract_attachments.py
njohner: I don't think so, what was added here is to test that the `related_items` index in solr gets set properly on the document extracted from an E-mail. So that belongs in this commit.

PR #7739, opened 2023-06-14
Fix task responsible permissions

njohner:
`TaskResponsible` did not have the `Access contents information` permission on inactive and archived dossiers, which would lead to an error [when trying to set the local roles ](https://github.com/4teamwork/opengever.core/blob/master/opengever/task/localroles.py#L144-L147) on such dossiers, because `brain.getObject()` actually uses traversal on the last part of the path to check access permission, which requires the above mentioned permission. We fix this by giving the `Access contents information` to `TaskResponsible` and `Contributor`.

For [CA-5897]

## Checklist
- [x] Changelog entry
- Upgrade-Steps:
  - DB-Schema migration
- Bug fixed:

[CA-5897]: https://4teamwork.atlassian.net/browse/CA-5897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/core/upgrades/20230614121224_fix_dossier_permissions_for_task_responsible/upgrade.py
```diff
@@ -0,0 +1,13 @@
+from ftw.upgrade import UpgradeStep
+from opengever.core.upgrade import NightlyWorkflowSecurityUpdater
+
+
+class FixDossierPermissionsForTaskResponsible(UpgradeStep):
+    """Fix dossier permissions for task responsible.
+    """
+
```
phgross, line 8: Upgradestep should be deferrable.

phgross (approved):
Apart from the thing mentioned, LGTM 👍 

As discussed, the upgradestep could reindex only inactive and resolved dossiers, but not sure the speed up is relevant.

opengever/core/upgrades/20230614121224_fix_dossier_permissions_for_task_responsible/upgrade.py
njohner: No we make all upgrade steps which update workflows non deferrable, because we need to make sure we have them in the right order. 
If necessary we can split the upgrade step into one installing the upgrade profile which would be non deferrable and one that would be deferrable and update the workflow security. But as submitting the nightly jobs should be really fast, I don't think this is necessary.

PR #7745, opened 2023-06-21
System information endpoint

elioschmutz:
This PR introduces a new endpoint `@system-information`:
- `dossier_participation_roles`
- `property_sheets`

Bildschirmfoto 2023-06-21 um 10 28 56

ℹ️ why not using the already existing `@config` endpoint which already provides some info about the system?
Because the `@config`-endpoint is the main entry endpoint for the `gever-ui` and should be as fast as possible. The endpoint should only provide information which is required to let the ui properly run. The `@system-information`-endpoint is used for a dedicated information-view in the UI. This endpoint can return expensive data because it will not be called often. It should provide the information in a way that the frontend can easily display the data.

For [CA-5543]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-5543]: https://4teamwork.atlassian.net/browse/CA-5543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/api/system_information.py
+++ b/opengever/api/system_information.py
@@ -1,6 +1,11 @@
 from opengever.api.dossier_participations import available_roles
 from opengever.api.dossier_participations import primary_participation_roles
+from opengever.propertysheets.api.base import PropertySheetAPIBase
+from opengever.propertysheets.metaschema import IPropertySheetDefinition
+from opengever.propertysheets.storage import PropertySheetSchemaStorage
 from plone.restapi.services import Service
+from plone.restapi.types.interfaces import IJsonSchemaProvider
+from zope.component import getMultiAdapter
 
 
 class SystemInformationGet(Service):
@@ -9,6 +14,7 @@ class SystemInformationGet(Service):
     def reply(self):
         infos = {}
         self.add_dossier_participation_roles(infos)
+        self.add_property_sheet_information(infos)
         return infos
 
     def add_dossier_participation_roles(self, infos):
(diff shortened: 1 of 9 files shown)


opengever/api/tests/test_system_information.py
```diff
@@ -7,12 +7,25 @@
...
         self.login(self.manager, browser)
         api.portal.set_registry_record('primary_participation_roles', ['regard'], IDossierParticipants)
 
         self.login(self.regular_user, browser)
         browser.open(self.portal.absolute_url() + '/@system-information', headers=self.api_headers)
 
-        self.assertEqual(browser.status_code, 200)
-        self.assertEqual({}, browser.json)
+        self.assertEqual([
+            {u'active': True, u'token': u'final-drawing', u'primary': False, u'title': u'Final signature'},
```
njohner, line 29: Can you check that everything is OK with non-ASCII characters?

opengever/api/system_information.py
```diff
@@ -17,3 +23,27 @@ def add_dossier_participation_roles(self, infos):
         for role in available_participation_roles:
             role['primary'] = role.get('token') in primary_participation_role_ids
         infos['dossier_participation_roles'] = available_participation_roles
+
+    def add_property_sheet_information(self, infos):
+        # Get translated assignment values
+        assinments_schema = getMultiAdapter(
```
njohner, line 29: assinments_schema -> assignments_schema

opengever/api/system_information.py
```diff
@@ -17,3 +23,27 @@ def add_dossier_participation_roles(self, infos):
...
+        for [slot_id, title] in assinments_schema.get('items').get('choices'):
+            property_sheet_slot_translations[slot_id] = title
+
+        # Then serialize the property-sheets and use the translated assignment title
+        property_sheet_api = PropertySheetAPIBase()
+        property_sheet_api.request = self.request
+        propertysheets = {}
+        for sheet in PropertySheetSchemaStorage().list():
+            serialized_sheet = property_sheet_api.serialize(sheet)
+            serialized_sheet['assignments'] = map(
```
njohner, line 43: I see it's a PR where you give some love to `map` and `lambda`s...
I personally almost always prefer comprehensions, but of course it's up to you
```
serialized_sheet['assignments'] = [
    {'id': assignment_id, 
     'title': property_sheet_slot_translations.get(assignment_id, assignment_id)} 
    for assignment_id in serialized_sheet.get('assignments')]
```

opengever/api/tests/test_system_information.py
elioschmutz: I'm not sure if I can update the vdex-items on the fly...

njohner: Me neither. As discussed, maybe just test in a real deployment.

elioschmutz: As discussed, I can't really test it because I'd need to update the vdex for this case. Here a printscreen which demonstrates that it works properly with umlauts:

Bildschirmfoto 2023-06-21 um 16 49 58

njohner: thanks

PR #7746, opened 2023-06-21
Improve performance of @notifications endpoint.

njohner:
We now avoid getting all the users notifications from the OGDS DB, instead only getting the batch we actually want.
This should dramatically increase the performance of the `@notifications` endpoint for users with a large number of notifications (there is a user in SG with 150'000 notifications). Note that slow requests on the OGDS will block all other users from all tenants, so this is a critical fix which we will backport to our LTS and to the SG release.

For a user with 70'000 notifications, performance improvement was from 2s to 300ms.

For [CA-5939]

## Checklist
- [x] Changelog entry

[CA-5939]: https://4teamwork.atlassian.net/browse/CA-5939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-5939.other
+++ b/changes/CA-5939.other
@@ -1,0 +1,1 @@
+Improve performance of @notifications endpoint. [njohner]
--- a/opengever/activity/center.py
+++ b/opengever/activity/center.py
@@ -141,7 +141,7 @@ class NotificationCenter(object):
 
         return errors
 
-    def get_users_notifications(self, userid, only_unread=False, limit=None, badge_only=False):
+    def get_users_notifications(self, userid, only_unread=False, badge_only=False):
         query = Notification.query.by_user(userid)
         if only_unread:
             query = query.filter(Notification.is_read == false())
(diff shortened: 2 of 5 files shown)


changes/CA-5939.other
```diff
@@ -0,0 +1 @@
+Improve performance of @notifications endpoint. [njohner]
```
elioschmutz, line 1: also add it in the API-changelog

elioschmutz (approved):
🌟

changes/CA-5939.other
njohner: You think so? It's not really an API change IMO, so I'd rather not add that in the API-changelog.

njohner:
@elioschmutz can you have a look at the second commit?

elioschmutz (approved):

PR #7756, opened 2023-07-13
Introduce display mode in the @schema endpoint.

phgross:
Currently the @schema endpoint only returns information for the add or the edit form. This leads to the problem, that the lifecycle informations (ILifeCycle) are not visible for users, which does not have the `EditLifecycleAndClassification` permission. 

To fix this issue I introduced a display mode, similar than plone.dexterity already those for the DefaultView, that mode can be used by the gever-ui for the MetaDataPanel.

For [CA-6037]

## Checklist
- [x] Changelog entry
- API change:

[CA-6037]: https://4teamwork.atlassian.net/browse/CA-6037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/api/tests/test_schema.py
+++ b/opengever/api/tests/test_schema.py
@@ -91,6 +91,29 @@ class TestSchemaEndpoint(IntegrationTestCase):
         self.assertNotIn('privacy_layer', fieldset['fields'])
 
     @browsing
+    def test_add_schema_does_contain_protected_fields_if_display_mode(self, browser):
+        self.login(self.regular_user, browser)
+        response = browser.open(
+            self.leaf_repofolder, view='@schema/opengever.dossier.businesscasedossier',
+            method='GET',
+            headers=self.api_headers,
+        ).json
+        fieldset = self._get_schema_fieldset(response, "classification")
+        self.assertIn('classification', fieldset['fields'])
+        self.assertIn('privacy_layer', fieldset['fields'])
+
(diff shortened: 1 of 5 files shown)


opengever/api/tests/test_schema.py
```diff
@@ -90,6 +90,29 @@ def test_add_schema_does_not_contain_protected_fields(self, browser):
...
+        self.leaf_repofolder.manage_permission("Edit lifecycle and classification", roles=[])
+        response = browser.open(
+            self.leaf_repofolder,
+            view='@schema/opengever.dossier.businesscasedossier?mode=display',
+            method='GET',
+            headers=self.api_headers,
+        ).json
+        fieldset = self._get_schema_fieldset(response, "classification")
+        self.assertIn('classification', fieldset['fields'])
+        self.assertIn('privacy_layer', fieldset['fields'])
```
elioschmutz, line 114: these are the same fields as in the first request (without display mode). Is something missing here? I think, protected fields  should not be visible in display-mode

elioschmutz (approved):

opengever/api/tests/test_schema.py
phgross: Yes, but by default they're not part when user does not have the `Edit lifecycle and classification`  permission (see text above).

PR #7758, opened 2023-07-19
Add support for manual journal entries in OGG bundles.

njohner:
For the SPD Solothurn migration we also need to migrate journal entries.

For [CA-5758]

## Checklist
- [ ] Changelog entry

[CA-5758]: https://4teamwork.atlassian.net/browse/CA-5758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


phgross (approved):
LGTM 👍 

I wonder if we have an indexing problem, because the adding of the relation comes after the indexing or do we not index the relations anyways.

njohner:
Ah good point, but I don't think that we index the journal. So it should not be a problem if we add the journal entries after indexing the created objects.

PR #7763, opened 2023-08-14
Enable GEVER UI registry feature by default.

lukasgraf:
Enable GEVER UI registry feature by default.

Even though the classic UI still has many fans 😜 , there won't be any deployments any more where we don't also deploy the GEVER UI. Any this feature flag only controls the switcher action anyway. We therefore make it true by default, and we can kill the feature flag soon with https://4teamwork.atlassian.net/browse/CA-6036

I also removed the question from the policytemplate:
Since the flag is now enabled by default, we don't need the explicit registry entry any more. Disabling it is still possible, but should be so rare that it's not supported by the policytemplate any more.

For [CA-6039](https://4teamwork.atlassian.net/browse/CA-6039)

## Checklist

- [x] Changelog entry

[CA-6039]: https://4teamwork.atlassian.net/browse/CA-6039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-6039.other
+++ b/changes/CA-6039.other
@@ -1,0 +1,1 @@
+Enable GEVER UI registry feature by default. [lgraf]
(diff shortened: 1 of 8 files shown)


njohner (approved):
LGTM

PR #7766, opened 2023-08-16
Allow copying dossiers containing locked excerpt documents.

njohner:
Proposal excerpts, which get deposited and locked in the dossier from which the proposal originated, prevented copy/pasting the dossier. Copy/pasting of only the excerpt actually already worked, and there is no reason to prevent copying the dossier. We therefore fix this by removing the locked on the copy of the excerpt before renaming happens (which is what was raising an error before this change).

For [CA-5991]

## Checklist
- [x] Changelog entry

[CA-5991]: https://4teamwork.atlassian.net/browse/CA-5991?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/api/copy_.py
+++ b/opengever/api/copy_.py
@@ -7,7 +7,9 @@ from opengever.base import _ as base_mf
 from opengever.document.behaviors import IBaseDocument
 from opengever.document.handlers import _update_docproperties
 from opengever.document.handlers import DISABLE_DOCPROPERTY_UPDATE_FLAG
+from opengever.locking.lock import MEETING_EXCERPT_LOCK
 from plone import api
+from plone.locking.interfaces import ILockable
 from plone.restapi.deserializer import json_body
 from plone.restapi.services.copymove.copymove import Copy
 from zope.container.interfaces import INameChooser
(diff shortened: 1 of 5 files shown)


opengever/api/copy_.py
```diff
@@ -90,6 +92,14 @@ def recursive_rename_and_fix_creator(self, obj, docs_to_update):
                 # in bumblebee if its parent is also renamed, because it's path
                 # will have changed when the task queue is processed.
                 docs_to_update.add(obj)
+
+                # Locks are copied with the objects, which prevents their
+                # renaming. For excerpts the copy does not need to be locked
+                # so we remove the corresponding locks here.
+                lockable = ILockable(obj)
+                if lockable.locked():
+                    lockable.unlock(MEETING_EXCERPT_LOCK)
```
lukasgraf, line 101: Will `lockable.unlock(MEETING_EXCERPT_LOCK)` fail gracefully (i.e. just not do anything) if the object has a lock of another type on it, but none of type MEETING_EXCERPT_LOCK?

lukasgraf (approved):
LGTM 👍 , just a question.

PR #7770, opened 2023-08-17
Getting @linked-workspaces on subdossier returns workspaces linked to main dossier

njohner:
We need to cover the case where a user has permissions on a subdossier but not on the main dossier. The gever-ui would until now call the `@linked-workspaces` on the main dossier, leading to an error. Instead we now allow calling that endpoint on subdossiers, but it will return the workspaces linked to the main dossier. As workspaces can only be linked to main dossiers, this is fine.

For [CA-6104]

## Checklist
- [x] Changelog entry

[CA-6104]: https://4teamwork.atlassian.net/browse/CA-6104?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-6104.feature
+++ b/changes/CA-6104.feature
@@ -1,0 +1,1 @@
+Getting @linked-workspaces on subdossier returns workspaces linked to main dossier. [njohner]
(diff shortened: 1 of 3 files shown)


lukasgraf (approved):
LGTM! 👍 
And a pretty clean solution at that.

PR #7773, opened 2023-08-22
Add WordDelimiterGraphFilterFactory to all text indexers.

phgross:
Currently two strings combined with an underscore are not indexed separately as it was in former versions of GEVER.
I also added the reference field  to the searched fields, this improves the search results, when searching with a specific
reference-number.

For [CA-6136]

## Checklist
- [x] Changelog entry

[CA-6136]: https://4teamwork.atlassian.net/browse/CA-6136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/solr-conf/managed-schema
+++ b/solr-conf/managed-schema
@@ -323,6 +323,17 @@
         <charFilter class="solr.ICUNormalizer2CharFilterFactory"/>
         <charFilter class="solr.MappingCharFilterFactory" mapping="mapping-FoldToASCII.txt"/>
         <tokenizer class="solr.WhitespaceTokenizerFactory"/>
+        <filter class="solr.WordDelimiterGraphFilterFactory"
+                splitOnCaseChange="1"
+                splitOnNumerics="1"
+                stemEnglishPossessive="1"
+                generateWordParts="1"
+                generateNumberParts="1"
+                catenateWords="1"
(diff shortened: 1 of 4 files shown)


solr-conf/managed-schema
```diff
@@ -337,17 +348,39 @@
...
+                splitOnNumerics="1"
+                stemEnglishPossessive="1"
+                generateWordParts="1"
+                generateNumberParts="1"
+                catenateWords="1"
+                catenateNumbers="1"
+                catenateAll="0"
+                preserveOriginal="1"/>
+        <filter class="solr.FlattenGraphFilterFactory" />
+        <filter class="solr.KeywordRepeatFilterFactory" />
```
lukasgraf, line 363: KeywordRepeatFilterFactory is not in the same place for `text_general` as it is for the others. I think this should probably go after the `WhitespaceTokenizerFactory`?

lukasgraf:
LGTM 👍 - but could you also add a regression test for the underscore case, so we don't break it again?

lukasgraf (approved):
LGTM 👍 - but could you also add a regression test for the underscore case, so we don't break it again?

PR #7774, opened 2023-08-25
Fix extraction of attachments from mails in closed tasks.

njohner:
Extraction worked in the classic UI but not over the API because we checked the permission to add documents on the context instead of on the extraction target.

Test failures seem unrelated, probably due to the new solr  container that was pushed to the `latest` tag.

For [CA-6128]

## Checklist

- [x] Changelog entry

[CA-6128]: https://4teamwork.atlassian.net/browse/CA-6128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-6128.bugfix
+++ b/changes/CA-6128.bugfix
@@ -1,0 +1,1 @@
+Fix extraction of attachments from mails in closed tasks. [njohner]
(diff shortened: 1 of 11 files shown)


phgross (approved):
👍 LGTM

PR #7775, opened 2023-08-25
Add extract_attachments context action for mails.

njohner:
Currently the action to extract attachments from an E-mail always appears, whether it's possible to do so or not. We therefore introduce a context action that the frontend can use to only display the action when necessary.

For [CA-6128]

## Checklist
- [x] Changelog entry

[CA-6128]: https://4teamwork.atlassian.net/browse/CA-6128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-6128.other
+++ b/changes/CA-6128.other
@@ -1,0 +1,1 @@
+Add extract_attachments context action for mails. [njohner]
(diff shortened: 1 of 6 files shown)


phgross (approved):
👍 LGTM

PR #7779, opened 2023-09-07
Add new solr index `is_locked_by_copy_to_workspace` and provide it in serializations. 

elioschmutz:
This PR adds a new solr index `is_locked_by_copy_to_workspace` and provides it in listings.

This allows us to mark such contents in listings in the UI as locked.

We can't just reindex the object in the `@lock`/`@unlock` endpoints because it is also possible to unlock the content through the old ui. In addition, there is a function in the `ILocking` behavior which unlocks all locks. So we have to override the `TTWLockable` object to implement the reindex in the source of locking.

For [CA-5563]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-5563]: https://4teamwork.atlassian.net/browse/CA-5563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -11,6 +11,7 @@ Breaking Changes
 
 Other Changes
 ^^^^^^^^^^^^^
+- Exposes the ``is_locked_by_copy_to_workspace`` attribute for document serializers if the ``workspace_client`` feature is activated.
 - ``@solrsearch`` and ``@listing``: ``is_locked_by_copy_to_workspace`` is provided for documents if the ``workspace_client`` feature is activated.
 
 2023.12.0 (2023-09-08)
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -11,6 +11,7 @@ Breaking Changes
 
 Other Changes
 ^^^^^^^^^^^^^
+- @favorites: return ``is_locked_by_copy_to_workspace`` for resolved documents if the ``workspace_client`` feature is activated.
 - Exposes the ``is_locked_by_copy_to_workspace`` attribute for document serializers if the ``workspace_client`` feature is activated.
 - ``@solrsearch`` and ``@listing``: ``is_locked_by_copy_to_workspace`` is provided for documents if the ``workspace_client`` feature is activated.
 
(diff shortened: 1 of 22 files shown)


opengever/base/solr/fields.py
```diff
@@ -447,6 +447,10 @@ def hide_facet(self, facet):
         index='checked_out',
         accessor='checked_out_fullname',
     ),
+    ListingField(
```
njohner, line 450: You don't need that. This field does not have a mapping, instead, simply add it to the `OTHER_ALLOWED_FIELDS` in `opengever/api/listing.py`. I think that should do the trick

docs/public/dev-manual/api/api_changelog.rst
```diff
@@ -11,6 +11,7 @@ Breaking Changes
 
 Other Changes
 ^^^^^^^^^^^^^
+- Exposes the ``is_locked_by_copy_to_workspace`` attribute for resolved document favorite items if the ``workspace_client`` feature is activated.
```
njohner, line 14: I think that entry should be something like:
- ``@favorites``: return ``is_locked_by_copy_to_workspace`` for resolved documents if the ``workspace_client`` feature is activated.

njohner (approved):
Just a few comments, otherwise LGTM. Don't forget to remove the last commit before merging.

opengever/core/upgrades/20230911131841_reindex_is_locked_by_copy_to_workspace/upgrade.py
```diff
@@ -0,0 +1,24 @@
...
+    deferrable = True
+
+    def __call__(self):
+        self.install_upgrade_profile()
+        if not is_workspace_client_feature_enabled():
+            return
+        self.reindex()
+
+    def reindex(self):
+        query = {'object_provides': IDocumentSchema.__identifier__}
```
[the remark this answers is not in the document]
elioschmutz: 👍 I updated the upgradestep. I think it's a good performance boost. Most of the dossiers are not linked to a workspace.

PR #7783, opened 2023-09-18
Bundle import: Map principal names to ids in local roles

lukasgraf:
For principal names that appear in local roles (the `_permissions` property in bundle items), **map their names to ids** based on OGDS users / groups.

This is required to support new-style deployments with OGDS auth plugin and user/group names and ids that are distinct from each other (as opposed to legacy style deployments where they are identical).

**Notes:**
- This mapping will be applied during every bundle import, regardless of whether the deployment is new-style or legacy. On legacy deployments it will just not have any effects, since `name == id`
- Lookups by username/groupname are case-insensitive
- If a princpal name in local roles matches both a user and a group, the group is preferred
- Since we are doing case insensitive lookups, we require that usernames / groupnames in OGDS are case-insensitively unique. A check is therefore performed early to ensure all names in OGDS are  case-insensitively unique.

The check for case-insensitively unique names in OGDS can be skipped using the `--no-check-unique-principals` flag. If not skipped,  it will raise an exception if it fails, and report errors as follows:

```
Found ambiguous principal names in OGDS.
The following users/groups are not case-insensitively unique:

username "user1" matches multiple records:
  username: user1 (id: 33333)
  username: USER1 (id: 44444)

groupname "group1" matches multiple records:
  groupname: GROUP1 (id: 22222)
  groupname: group1 (id: 11111)

Skip this check by running import with --no-check-unique-principals flag
```

For [CA-6039](https://4teamwork.atlassian.net/browse/CA-6039)

## Checklist

- [x] Changelog entry

[CA-6039]: https://4teamwork.atlassian.net/browse/CA-6039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/bundle/tests/test_section_map_principal_names_to_ids.py
+++ b/opengever/bundle/tests/test_section_map_principal_names_to_ids.py
@@ -1,0 +1,324 @@
+from collective.transmogrifier.interfaces import ISection
+from collective.transmogrifier.interfaces import ISectionBlueprint
+from opengever.base.model import create_session
+from opengever.bundle.sections.bundlesource import BUNDLE_KEY
+from opengever.bundle.sections.map_principal_names_to_ids import MapPrincipalNamesToIDsSection
+from opengever.bundle.tests import MockBundle
+from opengever.bundle.tests import MockTransmogrifier
+from opengever.ogds.models.group import Group
+from opengever.ogds.models.user import User
+from opengever.testing import IntegrationTestCase
+from uuid import uuid4
+from zope.annotation import IAnnotations
+from zope.interface.verify import verifyClass
+from zope.interface.verify import verifyObject
+
+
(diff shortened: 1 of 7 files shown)


opengever/bundle/tests/test_section_map_principal_names_to_ids.py
```diff
@@ -0,0 +1,324 @@
...
+            active=False,
+        )
+        inactive_user = User(
+            userid='inactive-22222',
+            username='user1',
+            external_id=uuid4().hex,
+            active=False,
+        )
+
+        for group in (inactive_group, inactive_user):
```
njohner, line 295: group -> principal

njohner (approved):
LGTM

PR #7786, opened 2023-09-20
Fix ech-0147 import if using double spaces in filenames.

elioschmutz:
This PR is a follow-up of https://github.com/4teamwork/opengever.core/pull/7723

It fixes an issue in the ech-0147 import where the import have been failed if:

- the filenames have had double empty spaces in the filename. This is valid but will automatically be normalized by the pythons ZipFile implementation. Thus, the file references could not be looked up properly
- it also failed because the encoded strings are not always consistent. To fix this issue, we now use a path mapping between the `ZipFile` internal normalized file paths and the file paths defined in the xml

⚠️ Backport to `2023.8.x` required

For [CA-6204]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-6204]: https://4teamwork.atlassian.net/browse/CA-6204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/ech0147/utils.py
```diff
@@ -121,7 +153,9 @@ def create_document(container, document, zipfile):
     if document.files:
         file_ = document.files.file[0]
         try:
-            zipinfo = zipfile.getinfo(normalize_filename(file_.pathFileName))
+            zipinfo = zipfile.getinfo(path_mapping.get(
+                replace_double_spaces_with_single_spaces(
+                    safe_utf8(file_.pathFileName))))
```
njohner, line 158: This makes no sense to me. You have `file_.pathFileName`, which I guess can contain consecutive whitespaces, which you remove with `replace_double_spaces_with_single_spaces`, to then map it back with the mapping to a path containing the double spaces?

My guess is that `file_.pathFileName` does not contain any consecutive white spaces, because it went through the ech0147 stuff so got the consecutive white spaces merged. The Zip file does contain the consecutive white spaces, that is why you need to map the filename back. So here you can simply do `zipinfo = zipfile.getinfo(path_mapping.get(safe_utf8(file_.pathFileName)))`

njohner (commented):
Something is fishy here. 

First of all, `pathFileName` is of type `xs:token` which will merge consecutive whitespaces, so clearly filenames with conescutive whitespaces are not allowed in the ech-0147 standard. It would be allowed if `pathFileName` were of type `xs:string`, but it's not. It's nevertheless ok if we fix this on our side, as this is clearly dumb in the standard.

So I'm pretty sure that the file paths in the `message.xml` have consecutive whitespaces that are merged, but the real path of the files in the Zip file still have the double spaces. I don't think that Python's implementation of Zip will remove consecutive white spaces, that's why your mapping, as proof you are merging consecutive white spaces in your mapping for the paths from `zipfile.namelist()`.

opengever/ech0147/utils.py
elioschmutz: @njohner thanks! You're right! I messed all up. 

- the `zipfile.namelist` returns the real filenames but NFD-normal form encoded
- the processed xml-file provides the `pathFileName` which has merged consecutive whitepaces in the NFC-normal form encoded

I updated the mapping and also the docstrings. I think it's ok now...

Side-Infos:

- ech-0147 standard: https://www.ech.ch/sites/default/files/dosvers/hauptdokument/STAN_d_DEF_2017-09-25_eCH-0039_V3.1_E-Government-Schnittstelle.pdf
- xml data type references: https://www.data2type.de/xml-xslt-xslfo/xml-schema/datentypen-referenz/xs-token/

opengever/ech0147/utils.py
```diff
@@ -17,17 +18,39 @@
...
+    This makes it hard to lookup the files in the zip-file.
+
+    To be able to properly lookup the files, we'll create a path mapping which
+    maps normalized paths in the format:
+
+    utf-8 strings in the NFC-normal form without consecutive whitespaces to
+    the real underlying file path within the zipfile.
+    """
+    return {
+        normalize_filename(path.replace('  ', ' ')): path
```
njohner, line 51: This will not work for filenames with 3 consecutive white spaces. I'd use a regular expression for safety here:

```
import re
pattern = re.compile(" {2,}")
pattern.sub(" ", path)
```

njohner (approved):
Just one more comment, otherwise LGTM

PR #7787, opened 2023-09-21
Grant view permissions to archivists on offered dossiers

njohner:
With this PR we make sure that archivists have the needed permissions to inspect the dossiers of an offer.

- Because this requires reindexing of all the objects inside the offer, we set the permissions in a nightly job. 
- We also provide `has_dossiers_with_pending_permissions_changes` in the serialisation of the disposition so that a warning can be shown when permissions are not up to date yet. 
- To avoid that the archivists receive an E-mail for a newly created disposition that they can't evaluate yet because they don't have the permissions on the dossier, we delay creation of the activity to after updating the permissions on all offered dossiers.

For [CA-6167]

## Checklist
- [x] Changelog entry
- New translations

[CA-6167]: https://4teamwork.atlassian.net/browse/CA-6167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/core/upgrades/20230922085455_add_new_attributes_to_dispositions/upgrade.py
+++ b/opengever/core/upgrades/20230922085455_add_new_attributes_to_dispositions/upgrade.py
@@ -1,0 +1,17 @@
+from ftw.upgrade import UpgradeStep
+from opengever.disposition.interfaces import IDisposition
+from persistent.list import PersistentList
+from Products.CMFPlone.utils import safe_hasattr
+
+
+class AddNewAttributesToDispositions(UpgradeStep):
+    """Add new attributes to dispositions.
+    """
+
+    def __call__(self):
+        query = {'object_provides': IDisposition.__identifier__}
+        for disposition in self.objects(query, "Add storage for new attributes on dispositions."):
+            if not safe_hasattr(disposition, "_dossiers_with_missing_permissions"):
+                disposition._dossiers_with_missing_permissions = PersistentList()
+            if not safe_hasattr(disposition, "_dossiers_with_extra_permissions"):
+                disposition._dossiers_with_extra_permissions = PersistentList()
--- a/opengever/disposition/disposition.py
+++ b/opengever/disposition/disposition.py
@@ -232,6 +232,7 @@ class Disposition(Container):
     implements(IDisposition, IResponseSupported)
 
     destroyed_key = 'destroyed_dossiers'
+    creation_activity_recorded_key = 'creation_activity_recorded'
 
     def __init__(self, *args, **kwargs):
         super(Disposition, self).__init__(*args, **kwargs)
@@ -497,3 +498,11 @@ class Disposition(Container):
 
     def revoke_view_permissions_from_archivists_on_dossier(self, dossier):
         RoleAssignmentManager(dossier).clear_by_reference(self)
+
+    @property
+    def creation_activity_recorded(self):
+        return IAnnotations(self).get(self.creation_activity_recorded_key, False)
+
+    @creation_activity_recorded.setter
+    def creation_activity_recorded(self, value):
+        IAnnotations(self)[self.creation_activity_recorded_key] = value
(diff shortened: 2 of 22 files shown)


opengever/disposition/nightly_jobs.py
```diff
@@ -101,3 +104,56 @@ def run_job(self, job, interrupt_if_necessary):
...
+            transaction.commit()
+            self.logger.info("Added permissions on %r" % dossier)
+
+        while disposition.dossiers_with_extra_permissions:
+            uid = disposition.dossiers_with_extra_permissions.pop()
+            dossier = uuidToObject(uid)
+            if dossier is not None:
+                disposition.revoke_view_permissions_from_archivists_on_dossier(dossier)
+            transaction.commit()
+            self.logger.info("Added permissions on %r" % dossier)
```
lukasgraf: Added -> Revoked

opengever/disposition/disposition.py
```diff
@@ -497,3 +498,13 @@ def give_view_permissions_to_archivists_on_dossier(self, dossier):
 
     def revoke_view_permissions_from_archivists_on_dossier(self, dossier):
         RoleAssignmentManager(dossier).clear_by_reference(self)
+
+    @property
+    def creation_activity_recorded(self):
+        """True if the document has changes that haven't been checked-in yet.
```
lukasgraf, line 527: This docstring doesn't make any sense :P ✂️

opengever/disposition/disposition.py
```diff
@@ -276,6 +278,26 @@ def set_destroyed_dossiers(self, dossiers):
 
         IAnnotations(self)[self.destroyed_key] = value
 
+    @property
+    def dossiers_with_missing_permissions(self):
+        return self._dossiers_with_missing_permissions
+
+    @dossiers_with_missing_permissions.setter
+    def dossiers_with_missing_permissions(self, dossiers_with_missing_permissions):
+        self._dossiers_with_missing_permissions = dossiers_with_missing_permissions
```
lukasgraf, line 290: These setters should turn the passed sequence into a persistent list, e.g. `PersistentList(dossiers_with_missing_permissions)`. Right now, some tests do `disposition.dossiers_with_missing_permissions = []`, which sets it to a non-persistent regular Python list. That should not be possible, IMHO.

opengever/disposition/disposition.py
```diff
@@ -276,6 +278,26 @@ def set_destroyed_dossiers(self, dossiers):
...
+
+    @property
+    def dossiers_with_extra_permissions(self):
+        return self._dossiers_with_extra_permissions
+
+    @dossiers_with_extra_permissions.setter
+    def dossiers_with_extra_permissions(self, dossiers_with_extra_permissions):
```
lukasgraf, line 297: Also, for these setters I would just call the argument `dossiers`, to improve readability.

opengever/core/upgrades/20230922085455_add_new_attributes_to_dispositions/upgrade.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class AddNewAttributesToDispositions(UpgradeStep):
+    """Add new attributes to dispositions.
+    """
+
+    def __call__(self):
+        query = {'object_provides': IDisposition.__identifier__}
+        for disposition in self.objects(query, "Add storage for new attributes on dispositions."):
+            disposition._dossiers_with_missing_permissions = PersistentList()
```
lukasgraf, line 13: I think this should include a `safe_hasattr` check to make sure to not overwrite existing attributes, if we ever need to re-run this upgrade step.

lukasgraf (changes requested):
LGTM 👍 
Just some minor details.

opengever/disposition/nightly_jobs.py
njohner: This is fixed in a later commit

opengever/core/upgrades/20230922085455_add_new_attributes_to_dispositions/upgrade.py
njohner: Thanks, good point

njohner:
All done, thanks for the review.

PR #7791, opened 2023-10-02
Respect local dossier depth when uploading a dossier structure.

elioschmutz:
We can set a global allowed dossier depth in the registry. Adding a subdossier deeper than the configured limit is not possible. Beside of this global limit, we also have a local limit which can be higher than the actual global limit. Usually, this will happen if the user creates a dossier from a template. dossier templates can be deeper nested than the global limit. 

Such dossiers will automatically expand the allowed limit to the existing subdossier-depth. Means, the user can always create siblings, even if the depth is higher than the global limit.

Currently, we do not respect this local depth when uploading a folder structure. The `@upload-structure` endpoint only checks for the global limit.

This PR fixes this issue and uses the same mechanism for checking the allowed depth in the `@upload-structure` endpoint as we do it when manually adding a new dossier.

For [CA-6215]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-6215]: https://4teamwork.atlassian.net/browse/CA-6215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/dossier/utils.py
+++ b/opengever/dossier/utils.py
@@ -3,8 +3,10 @@ from Acquisition import aq_parent
 from opengever.dossier.behaviors.dossier import IDossierMarker
 from opengever.dossier.dossiertemplate.behaviors import IDossierTemplateMarker
 from opengever.dossier.dossiertemplate.behaviors import IDossierTemplateSchema
+from opengever.dossier.interfaces import IDossierContainerTypes
 from opengever.inbox.inbox import IInbox
 from opengever.workspace.interfaces import IWorkspace
+from plone import api
 from Products.CMFPlone.interfaces.siteroot import IPloneSiteRoot
 import unicodedata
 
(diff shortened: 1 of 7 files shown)


lukasgraf (approved):
LGTM 👍 
Great description in PR body and tests! 💚 

Tests with Solr are failing though.

elioschmutz:
@lukasgraf I had to do more changes than expected. If adding a dossier-structure on a private folder or a repository folder, we'll not have to check the local depth. But the way of calculation is different:

If within a dossier and talking about the `maximum_dossier_depth`, means the total number of subdossier-depth allowed. So if the `maximum_dossier_depth` is 1, it means, that we can have one subdossier within a dossier. But the dossier itself in the view of a repository folder or a private folder would have a depth of 2. The dossier (1) and the subdossier (2).
 
On the other hand, there is no need to check for the local depth if checking the allowed addable depth on a repository folder or a private folder.

So would you like to review once again?

opengever/dossier/utils.py
```diff
@@ -78,3 +80,16 @@ def supports_is_subdossier(obj):
     the is_subdossier index / method.
     """
     return IDossierMarker.providedBy(obj) or IDossierTemplateMarker.providedBy(obj)
+
+
+def assert_subdossier_depth_allowed(subdossier_depth=1):
```
lukasgraf, line 85: I would rename this to `check_subdossier_depth_allowed`.

As you write in the docstring, it's a check, not an assertion (which is assumed to be always true).

lukasgraf (approved):
Small detail, otherwise LGTM 👍

PR #7800, opened 2023-10-05
all_users_and_groups global source now also returns inactive groups.

njohner:
This will allow to create `RoleAssignmentReport`s for inactive groups. IMO this is the expected behaviour for that source. I also checked in the gever-ui and the source is only used for permission reports, so I consider this change non braking.

For [CA-6132]

## Checklist
- [x] Changelog entry
- API change:
  - If breaking:
    - [ ] api-change label added


lukasgraf (approved):
LGTM 👍

PR #7803, opened 2023-10-10
User different userid than username in testing fixture.

njohner:
With this PR we want to fix most remaining issues of having different usernames than userids. For this we:
- Modify the testing fixture so that the OGDS Users have different userids than usernames. While leaving the userids as they were and adding different usernames would most likely lead to fewer test failures, I decided instead to use the old userids as usernames and set something new for the userids. This has the advantage that it makes it apparent all the places we see the userids in the tests, notably in URLs and JSON responses from the API. This makes it easy to see where userids are visible to Users and external applications, which IMO they shouldn't.
- Activate the OGDSAuth plugin in the integration tests, as this is necessary when using different usernames from userids, as otherwise we get the userid as username everywhere the memberdata is used. Another way around this would be to actually use a different username than userid for the plone users in the tests, but this is far removed from the situation we will have on real deployments.

**Current Stand**
I've started fixing the tests, sometimes it's really the tests that need fixing, sometimes it's actually fixing things in the code. There are still a lot of failing tests and the process is pretty painful. 

Activating the OGDSAuth plugin itself also leads to ~40  test failures, I think these could be fixed in a separate PR for clarity (see first commit of https://github.com/4teamwork/opengever.core/pull/7808)

For [CA-6244]

## Checklist
- [ ] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-6244]: https://4teamwork.atlassian.net/browse/CA-6244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


njohner:
I'll leave that one open, in case you want to cherry-pick some things, but basically I think I did already cherry-pick out the stuff that was still useful.

buchi:
Most fixes should be in place now.

PR #7804, opened 2023-10-11
Remove role manager from sharing view of other contexts than dossier.

njohner:
with https://github.com/4teamwork/opengever.core/pull/7789/commits/3f0cce71eb3f96f2a4a6fd4f0f97654c87be1ade, `Role Manager` was added to lawgiver, which has as side-effect that it appears in the sharing view of diverse portal types, where it is not desired. This is fixed here, which, at least on the tested object types, recovers the roles displayed before that change in the sharing view.

For [CA-6155]

## Checklist
- [ ] Changelog entry -> no CL entry, it's in the same release as the PR introducing this issue


elioschmutz (approved):
👍 thanks!

lukasgraf (approved):
LGTM 👍 Thanks!

PR #7805, opened 2023-10-12
Bump ftw.recipe.solr to get extra-opts recipe option.

lukasgraf:
Bump `ftw.recipe.solr` to get `extra-opts` recipe option.

This is needed to use `extra-opts` to set `-Dsolr.disable.shardsWhitelist=true` as an additional option in opengever.zug buildouts.


njohner (approved):
LGTM

PR #7807, opened 2023-10-18
Fixes for feature giving local roles to dossier responsibles.

njohner:
Two things were broken in the new feature implemented in https://github.com/4teamwork/opengever.core/pull/7789:
- Permission check for editing the responsible field was not correct
- The event handler setting the local roles for the responsible broke the sharing view. I'm not sure why that was not picked up by any tests, as there are tests of the sharing view. Might be an event order issue?

For [CA-6155]

## Checklist
- [ ] Changelog entry -> no CL entry, the feature was not released yet


phgross (approved):
👍 LGTM

PR #7810, opened 2023-10-31
Add transferring office field to dispositions to overwrite the AktenbildnerName.

njohner:
Our default for the `AktebnbildnerName` in the SIP package only makes sense for deployments where we have one AdminUnit per Service, but not for deployments like Fribourg. That Metadata should indeed contain the name of the service which was responsible for the set of documents and therefore needs to be overwritable.

For [CA-6281]

## Checklist
- [x] Changelog entry

[CA-6281]: https://4teamwork.atlassian.net/browse/CA-6281?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/disposition/tests/test_disposition.py
+++ b/opengever/disposition/tests/test_disposition.py
@@ -349,6 +349,16 @@ class TestDispositionEditForm(IntegrationTestCase):
         self.assertEquals(['Item state changed.'], info_messages())
         self.assertFalse(self.disposition.has_sip_package())
 
+    @browsing
+    def test_can_overwrite_transferring_office(self, browser):
+        self.login(self.records_manager, browser)
+        self.assertEqual(u'Hauptmandant', self.disposition.transferring_office)
+
+        browser.open(self.disposition, view='edit')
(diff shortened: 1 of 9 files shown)


lukasgraf (approved):
LGTM 👍

opengever/disposition/tests/test_disposition.py
```diff
@@ -349,6 +349,16 @@ def test_sip_package_is_removed_on_close(self, browser):
         self.assertEquals(['Item state changed.'], info_messages())
         self.assertFalse(self.disposition.has_sip_package())
 
+    @browsing
+    def test_can_overwrite_transferring_office(self, browser):
+        self.login(self.records_manager, browser)
+        self.assertEqual(u'Hauptmandant', self.disposition.transferring_office)
```
lukasgraf, line 355: Failing test, looks like this should be `u'Hauptmand\xe4nt'`

lukasgraf (approved):

PR #7811, opened 2023-11-01
OGDS user and group labels display name instead of ID.

njohner:
OGDSUser and OGDSGroup labels now use the name instead of the id. The `__repr__` still uses the ids, but I think that is correct.

For [CA-6237]

## Checklist
- [x] Changelog entry

### DIFF
--- a/changes/CA-6237.other
+++ b/changes/CA-6237.other
@@ -1,0 +1,1 @@
+OGDS user and group labels display name instead of ID. [njohner]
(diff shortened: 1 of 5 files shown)


elioschmutz (approved):
Do we always have a groupname/username? Or is it possible to create users/groups only with an id? Should we use a fallback to the ids if there is no name?

njohner:
`groupname` and `username` are non-nullable. So there are always there

PR #7820, opened 2023-11-09
Use OGDS auth plugin in integration tests

lukasgraf:
This adds the OGDS auth plugin to integration tests, and disables the `IUserEnumeration` capability of the `source_users` plugin so that only the OGDS auth plugin is used to enumerate users in tests.

With this we should have the foundation for starting to split `userid` / `username` in testing.

For [CA-6237](https://4teamwork.atlassian.net/browse/CA-6237)

Follow up issue:
https://4teamwork.atlassian.net/browse/CA-6382

## Checklist

- [ ] Changelog entry (no user visible changes)

  - [x] Create follow-up stories and link them in the PR and Jira issue: [CA-6382](https://4teamwork.atlassian.net/browse/CA-6382)

[CA-6237]: https://4teamwork.atlassian.net/browse/CA-6237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CA-6382]: https://4teamwork.atlassian.net/browse/CA-6382?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/api/tests/test_users.py
+++ b/opengever/api/tests/test_users.py
@@ -40,6 +40,7 @@ class TestUsersGet(IntegrationTestCase):
                                        u'Member',
                                        u'Authenticated',
                                        u'principal:AuthenticatedUsers',
+                                       u'principal:projekt_a',
                                        u'principal:fa_users',
                                        u'Anonymous'],
              u'home_page': None,
--- a/opengever/api/tests/test_users.py
+++ b/opengever/api/tests/test_users.py
@@ -14,7 +14,7 @@ class TestUsersGet(IntegrationTestCase):
         browser.open('{}/@users'.format(self.portal.absolute_url()),
                      headers=self.api_headers)
 
-        self.assertEqual(22, len(browser.json))
+        self.assertEqual(21, len(browser.json))
 
     @browsing
     def test_enumarating_users_unauthorized_raises(self, browser):
--- a/opengever/readonly/tests/test_file_download.py
+++ b/opengever/readonly/tests/test_file_download.py
@@ -43,6 +43,10 @@ class TestFileDownloadInReadOnly(IntegrationTestCase):
     def test_downloading_doc_pdf_journaling_doesnt_cause_readonly_error(self, browser):
         self.login(self.regular_user, browser)
 
+        # Get other potential writes-on-read out of the way.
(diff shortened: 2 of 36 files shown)


opengever/readonly/tests/test_file_download.py
```diff
@@ -43,6 +43,10 @@ def test_file_download_journaling_doesnt_cause_readonly_error(self, browser):
     def test_downloading_doc_pdf_journaling_doesnt_cause_readonly_error(self, browser):
         self.login(self.regular_user, browser)
 
+        # Get other potential writes-on-read out of the way.
```
njohner, line 46: I guess we might have to fix that in the future if we want users to be able to login in read-only mode.

opengever/api/tests/test_users.py
```diff
@@ -14,7 +14,7 @@ def test_enumarating_user_is_possible_for_regular_user(self, browser):
         browser.open('{}/@users'.format(self.portal.absolute_url()),
                      headers=self.api_headers)
 
-        self.assertEqual(22, len(browser.json))
+        self.assertEqual(21, len(browser.json))
```
njohner, line 17: I just checked and the frontend uses the `@ogds-user-listing` endpoint to list users in the "Benutzende and Teams" card, so customers will still see the inactive users there, so that is 👍

njohner (approved):
Very nice and easy to review! 🎉 
As said in my comments, I think we might need to fix the write on read when logging-in in read-only mode.

opengever/readonly/tests/test_file_download.py
lukasgraf: Yes, but to be honest I'm not even sure it's actually caused by the login - at least not as it happens in production. It might be just an artifact of how login is performed during testing. Because as far as I remember, we fixed login during readonly in the classic setup (just don't write the last_login timestamp during readonly mode), but there were still some pending writes during tests that I didn't obeserve in production.

PR #7826, opened 2023-11-22
Always return the UID in when using the GeverSerializeToJsonSummary serializer

elioschmutz:
This PR extends the `GeverSerializeToJsonSummary` serializer to always return the UID.

The summary serializer will be used for related items, among other things, if requesting a context. We can define a list of `metadata_fields` which should be fetched in addition to the default fields. By this parameter, we're able to extend the fields with the `UID`:

`http://example.com/workspaces/workspace-5/meeting-5/agendaitem-537?metadata_fields:list=uid`
```
{
    "@id": "http://example.com/workspaces/workspace-5/meeting-5/agendaitem-537",
    "@type": "opengever.workspace.meetingagendaitem",
    "UID": "e6eede9d84204f5d8700daa972cf510a",
    "relatedItems": [
        {
            "@id": "http://example.com/fd/workspaces/workspace-5/folder-7/document-603",
            "@type": "opengever.document.document",
            "UID": "131fde8ea2c740108c3c926a14d0eab6",
            "...": "...",
        }
    ],
    "...": "..."
}
```

In patch-requests we can use the header `Prefer` header with `return=representation`. This will return the new context after updating it. Unfortunately, we can't use the `metadata_fields` parameter in this case and the response will only contain the default fields which does not include the UID.

This PR extends the default list to always return the UID, even in patch-requests.

For [CA-6151]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-6151]: https://4teamwork.atlassian.net/browse/CA-6151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lukasgraf (approved):
LGTM 👍 - but please have a look over any examples in the API docs, whether they need updating.

PR #7829, opened 2023-11-27
Blacklist uploaded file formats based on extension instead of mimetype.

lukasgraf:
Blacklist uploaded file formats based on extension instead of mimetype.

Blacklisting file formats by mimetypes doesn't work well because many file formats resolve to the MIME type `text/plain`, unless we specifically add them to our registry.

For [CA-5703](https://4teamwork.atlassian.net/browse/CA-5703)

## Checklist

- [x] Changelog entry

[CA-5703]: https://4teamwork.atlassian.net/browse/CA-5703?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-5703-1.other
+++ b/changes/CA-5703-1.other
@@ -1,0 +1,1 @@
+Blacklist uploaded file formats based on extension instead of mimetype. [lgraf]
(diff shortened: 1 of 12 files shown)


phgross (approved):
👍

PR #7831, opened 2023-11-29
Also support usernames for the responsible field in dossier API.

phgross:
With the recently introduced separation between userid and username, certain API endpoints now expect the userid to be specified. Since external applications such as the RIS have no knowledge of the GEVER internal userid, they must be adapted so that they can also handle the username.  This PR introduces this support for the dossier responsible field.

For [CA-6237]

## Checklist
- [x] Changelog entry
- [x] Link to issue (Jira or GitHub) and backlink in issue (Jira)
- API change:
  - [x] Documentation is updated _(imho not necessary)_
  - [x] API Changelog entry (see [guide](https://4teamwork.atlassian.net/wiki/spaces/4TEAM/pages/451248812/API+Changelog+Guidelines))

[CA-6237]: https://4teamwork.atlassian.net/browse/CA-6237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7843, opened 2023-12-05
Also support usernames for the responsible and issuer field in tasks API.

phgross:
Same as I already did for dossiers (#7831), we need also for task creation and modification.

With the recently introduced separation between userid and username, certain API endpoints now expect the userid to be specified. Since external applications such as the RIS have no knowledge of the GEVER internal userid, they must be adapted so that they can also handle the username.

For [CA-6237]

- [x] Changelog entry

- API change:

[CA-6237]: https://4teamwork.atlassian.net/browse/CA-6237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-6237-2.other
+++ b/changes/CA-6237-2.other
@@ -1,0 +1,1 @@
+Task responsible and issuer field supports now also usernames not just userids. [phgross]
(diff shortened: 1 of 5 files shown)


lukasgraf (approved):
LGTM! 👍 
With this change, a few tests that I needed to `skip()` should now be fixed (both in the backend and gever-ui). However, because it's a hassle to identify them one by one, I suggest that we do a cleanup pass after we've merged the first set of fixes and then
- unskip all the tests that are passing again
- replace update_user with a user from the fixture (where they already have been updated to have different userid/username)

PR #7845, opened 2023-12-05
Fix webactions provider for setups where userid and username are different

phgross:
For [CA-6237]

Note: I switched from regular_user to the dossier_manager in the tests, because the dossier_manager already has a different username than the userid.

## Checklist
- [x] Changelog entry


lukasgraf (approved):
LGTM 👍

PR #7864, opened 2023-12-18
Fix service user ids in hardcoded test keys

deiferni:
This is a follow-up for https://github.com/4teamwork/opengever.core/pull/7837. We should also adapt the `user_id` in the hardcoded service keys that are loaded into a fixture.

For [HG-4272](https://4teamwork.atlassian.net/browse/HG-4272)

## Checklist

- [x] Changelog entry

[HG-4272]: https://4teamwork.atlassian.net/browse/HG-4272?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


phgross (approved):
👍

PR #7865, opened 2023-12-19
Improve reporting view performance for path based queries.

phgross:
The current implementation which sorts using the supplied path sorting, was massively inperformant and led to Solr crashes. Instead we fetch the the current sorting of the listing and use this.

This drops the sorting support for the classic UI, as discussed at the daily that is not really a problem.

For [CA-6524]

## Checklist
- [x] Changelog entry

[CA-6524]: https://4teamwork.atlassian.net/browse/CA-6524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/base/browser/reporting_view.py
+++ b/opengever/base/browser/reporting_view.py
@@ -136,14 +136,17 @@ class SolrReporterView(BaseReporterView):
                 doc = OGSolrDocument(doc, fields=self.solr.manager.schema.fields)
                 yield IContentListingObject(doc)
 
-    def _extend_selected_items_query_by_paths(self, query, paths):
-        # Sort results according to paths passed in request. Those should be
-        # sorted exactly as the user saw them in the UI (classic and geverui).
-        query['sort'] = 'score asc'
-        query['query'] = 'path:({})'.format(
-            'OR '.join(['{}^{}'.format(escape(path), score_value)
-                       for score_value, path in enumerate(paths)])
-        )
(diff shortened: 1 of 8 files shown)


opengever/base/browser/reporting_view.py
```diff
@@ -136,14 +136,17 @@ def get_selected_items(self):
...
-            'OR '.join(['{}^{}'.format(escape(path), score_value)
-                       for score_value, path in enumerate(paths)])
-        )
+    def _extend_selected_items_query_by_paths(self, solr_query, paths):
+        # Try to fetch the current sort
+        try:
+            listing = queryMultiAdapter((self.context, self.request), name="GET_application_json_@listing")
+            listing.listing_name = self.corresponding_listing_name
+            query, filters, start, rows, sort, field_list, params = listing.prepare_solr_query(self.request.form)
+        except Exception:
```
elioschmutz, line 145: I think we should not sliently catch all exceptions. If there is something wrong, i.e. if something has changed in the `prepare_solr_query` it should raise that we can fix it.

elioschmutz (approved):

opengever/base/browser/reporting_view.py
phgross: You're right, forgot to replace it with the real Error.

PR #7866, opened 2023-12-27
Remove ftw.tika

buchi:
ftw.tika isn't used anymore for a long time but it's still a dependency and the Tika server is downloaded in the dev buildout.
Should be safe to remove it.

## Checklist

- [ ] Changelog entry


lukasgraf (approved):
LGTM 👍 
The GS profile should have been uninstalled already with https://github.com/4teamwork/opengever.core/pull/7249

PR #7867, opened 2024-01-03
Bump ftw.mail to version 2.8.0

buchi:
This drops the dependency on plone.directives.form and avoids pulling in the whole grok stack.

## Checklist

- [x] Changelog entry


lukasgraf (approved):
LGTM! 👍

PR #7868, opened 2024-01-08
Nested task templates: Also pass docs to subtasks that get started

lukasgraf:
Nested task templates: Also pass docs to subtasks that get started:

This will make sure that if a sequential task is started because it is the next one in line, and the user selected `pass_documents_to_next_task`, the documents will be passed (i.e. set as relatedItems) not just to that next task, but also to any subtasks that get started if the next task happens to be a task template.

So in short:
If the user selects `pass_documents_to_next_task`, the documents will always be passed to exactly the same tasks that automatically get started.

For [CA-6450](https://4teamwork.atlassian.net/browse/CA-6450)

## Checklist

- [x] Changelog entry

[CA-6450]: https://4teamwork.atlassian.net/browse/CA-6450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-6450.bugfix
+++ b/changes/CA-6450.bugfix
@@ -1,0 +1,1 @@
+Nested task templates: Also pass docs to subtasks that get started. [lgraf]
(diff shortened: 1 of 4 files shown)


buchi (approved):
👍🏻 LGTM

PR #7870, opened 2024-01-10
Fix tests on master

lukasgraf:
These tests were executed against an old base of `master`, where the fixture user's user ID was not renamed from `kathi.barfuss` to `regular_user` yet.

That's why the PR on GitHub passed at one point, but once merged into the current master where the rename already happened, they fail.


deiferni (approved):
LGTM 👍 😏

PR #7871, opened 2024-01-12
Fix find-links for Gitlab packages

buchi:
- No longer extend from `4teamwork-psc.cfg`. Also removes `isotoma.buildout.basicauth` extension which is no longer needed.

- Add `find-links` for each package as Gitlab's simple index consists of subpages that aren't processed by setuptools.

## Checklist

- [ ] Changelog entry

### DIFF
--- a/development.cfg
+++ b/development.cfg
@@ -9,6 +9,16 @@ extends =
     base-testserver.cfg
     base-qa.cfg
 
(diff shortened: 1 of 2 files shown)


lukasgraf (approved):
LGTM 👍

PR #7873, opened 2024-01-12
Fix performance when checking if a workspace folder can be trashed

buchi:
The current implementation recursively checks if all contained items of a workspace folder can be trashed.
This obviously gets slower, the more items a workspace folder contains and is already unusable if it contains just a few items.

Now, we only check if the workspace folder itself can be trashed and do the check for contained items while performing the trash operation.

For [CA-6605](https://4teamwork.atlassian.net/browse/CA-6605)

## Checklist

- [x] Changelog entry

[CA-6605]: https://4teamwork.atlassian.net/browse/CA-6605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-6605.bugfix
+++ b/changes/CA-6605.bugfix
@@ -1,0 +1,1 @@
+Fix performance when checking if a workspace folder can be trashed. [buchi]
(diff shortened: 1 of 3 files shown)


lukasgraf (approved):
LGTM 👍

PR #7874, opened 2024-01-16
Fix building Docker image

buchi:
We now require a Gitlab token for building the Docker image.

## Checklist

- [ ] Changelog entry


lukasgraf (approved):
LGTM 👍

PR #7875, opened 2024-01-18
Fix Docker image: remove ZCML include of ftw.tika in Docker image

buchi:
## Checklist

- [x] Changelog entry


lukasgraf (approved):
LGTM 👍

PR #7876, opened 2024-01-19
Also update `modified` when updating `changed` 

lukasgraf:
Also update `modified` when updating `changed` timestamp:

We rely on `modified` in `solr-sync` to determine which items are out of sync (present on both sides, but with potential difference in metadata contents). However, `modified` gets updated not nearly as consistently as we thought: For example, triggering a workflow transition will not update `modified`.

Therefore, we always update `modified` when we update the higher-level `changed` timestamp. This way, we should at least catch the most important user facing changes to objects.

This includes an upgrade-step which will touch `modified` for all items in the catalog where `changed` is newer than `modified`. It will defer the reindexing of `modified` (and all metadata) to a nightly job, and will only reindex `modified` in ZCatalog, but not Solr. Therefore, the next time `bin/instance solr-sync sync` is run, those objects will be picked up as being out of sync.    

For [CA-6552](https://4teamwork.atlassian.net/browse/CA-6552)

## Checklist

- [x] Changelog entry

- Upgrade-Steps:

[CA-6552]: https://4teamwork.atlassian.net/browse/CA-6552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-6552.bugfix
+++ b/changes/CA-6552.bugfix
@@ -1,0 +1,1 @@
+Also update modification date when updating changed timestamp. [lgraf]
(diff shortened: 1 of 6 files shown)


buchi (approved):
👍🏻 LGTM

PR #7878, opened 2024-01-21
Create site and run upgrades on startup

buchi:
Creates a new policyless deployment if no existing Plone site is found and runs pending upgrade steps on startup.
For now it's only enabled for Docker containers but it can be enabled for Buildout deployments as well by setting the environment variable `OG_AUTOSETUP=1`.

To ensure only one instance performs installation or upgrades in a multi ZEO client setup, a lock marker is written to the database. All instances not having the lock, wait until the lock is released before getting ready to serve requests. The lock is checked regularly with a maximum intervall of 5 minutes.

On the first start with multiple instances and an empty database, a ConflictError will probably occur. This is because Zope itself does setup its application object in the database but doesn't handle conflict resolution. Instances running into a ConflictError need to be restarted. Or better, use only one replica on initial setup:

```
docker compose up -d --scale ogcore=1
```

Automatic creation of a new deployment and execution of upgrade steps can be disabled by setting the appropriate environment variables `OG_SKIP_SITE_CREATION=1` or `OG_SKIP_UPGRADES=1`.

For [CA-XXXX]

## Checklist

- [ ] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

### DIFF
--- a/docker/core/Dockerfile
+++ b/docker/core/Dockerfile
@@ -127,6 +127,7 @@ COPY --from=go-crond-builder /go/src/github.com/webdevops/go-crond /usr/local/bi
 
 VOLUME /data
 WORKDIR /app
-ENV PATH="/app/bin:$PATH"
+ENV PATH="/app/bin:$PATH" \
+    OG_AUTOSETUP=1
 USER plone
 EXPOSE 8160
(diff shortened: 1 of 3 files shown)


buchi:
Superseded by #7920. Turns out we want more control when create a site and run upgrades.

PR #7889, opened 2024-02-04
Use Plone 4.3 image based on Alpine 3.18

buchi:
The previous base image was based on Alpine 3.14 which is EOL.

The new base image comes with various improvements described in https://github.com/4teamwork/plone-docker/pull/1

PostgreSQL 15.5 (libpq) and OpenLDAP 2.6.5 (libldap) provided by Alpine 3.18 are newer versions than officially supported by psycopg2 2.8.6 and python-ldap 2.4.45 but seem to work so far.

Bump cryptography to version 3.3.2 to include fix for CVE-2020-36242. We do not bump it in buildout `versions.cfg` because it requires OpenSSL 1.1 which is not available on CentOS 7.

## Checklist

- [ ] Changelog entry

### DIFF
--- a/changes/GH-7889.other
+++ b/changes/GH-7889.other
@@ -1,0 +1,1 @@
+Update Docker image to Alpine 3.18 and include backports of security fixes for various dependencies. [buchi]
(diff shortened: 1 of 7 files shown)


lukasgraf (approved):
LGTM! 👍

PR #7890, opened 2024-02-04
Bump dependencies 2024 01

buchi:
Bump various dependencies to include security fixes.

For [CA-XXXX]

## Checklist

_Everything has to be done/checked. Checked but not present means the author deemed it unnecessary._

- [ ] Changelog entry
- [ ] Link to issue (Jira or GitHub) and backlink in issue (Jira)

PR #7893, opened 2024-02-09
Pin picked production dependency versions

lukasgraf:
Pin versions of picked production dependencies.

These are some new transitive dependencies that are pulled in after the dependency bump in #7890 and haven't been pinned yet (and therefore preventing production buildouts with `allow-picked-versions = false` to run).

This should be all the versions required for production I think, all others currently listed as picked seem to be dev-dependencies.

For [CA-6686](https://4teamwork.atlassian.net/browse/CA-6686)

[CA-6686]: https://4teamwork.atlassian.net/browse/CA-6686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


buchi (approved):
👍🏻 LGTM

PR #7896, opened 2024-02-14
Add new field external_reference to teamroom todo 

KunzS85:
With this PR, we add a new text field for external references to the todo's of the teamroom. 
The aim of the change is to be able to provide links to external resources in the UI.
This leads to an extension of the attributes of the todo schema with a text field with the name external_reference.

For [CA-6482]

## Checklist

- [x] Changelog entry

- API change:

[CA-6482]: https://4teamwork.atlassian.net/browse/CA-6482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -13,8 +13,6 @@ Breaking Changes
 Other Changes
 ^^^^^^^^^^^^^
 - ``POST @dossier-transfers``: New endpoint to create dossier transfers.
-- Add 'external_reference' field to workspace todo 
-
 
 2024.3.0 (2024-02-09)
 ---------------------
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -13,6 +13,7 @@ Breaking Changes
 Other Changes
 ^^^^^^^^^^^^^
 - ``POST @dossier-transfers``: New endpoint to create dossier transfers.
+- Add 'external_reference' field to workspace todo 
 
 
 2024.3.0 (2024-02-09)
--- a/docs/public/dev-manual/api/workspace/todos.rst
+++ b/docs/public/dev-manual/api/workspace/todos.rst
@@ -20,8 +20,9 @@ Ein ToDo erstellen
         "@type": "opengever.workspace.todo",
         "title": "Bitte Dokument reviewen",
         "text": "Das wichtige Dokument muss angeschaut werden.",
-        "deadline": "2018-10-16"
+        "deadline": "2018-10-16",
         "responsible": "john.doe",
+        "external_reference": "www.einexternerinhalt.ch"
       }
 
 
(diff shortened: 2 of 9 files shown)


docs/public/dev-manual/api/workspace/todos.rst
```diff
@@ -20,8 +20,9 @@ Ein ToDo erstellen
         "@type": "opengever.workspace.todo",
         "title": "Bitte Dokument reviewen",
         "text": "Das wichtige Dokument muss angeschaut werden.",
-        "deadline": "2018-10-16"
+        "deadline": "2018-10-16",
         "responsible": "john.doe",
+        "external_reference": "www.einexternerinhalt.ch"
```
lukasgraf, line 25: It's generally not considered good practice to use domain names that you don't own in examples or tests.
While that domain might not exist today, it might at some point, and we might be stepping on someones toes without even realizing it.

Worst case scenarios include E2E accidentally making actual network requests to such domains, or emails being sent to their postmasters. Unlikely, but it's best not to even take the risk by using one of the example domains that [RFC 2606](https://datatracker.ietf.org/doc/html/rfc2606#section-3) specifically reserves for this purpose:
- `example.com`
- `example.net`
- `example.org`

(Or a subdomain of those, like `external.example.org`)

docs/public/dev-manual/api/api_changelog.rst
```diff
@@ -13,6 +13,7 @@ Breaking Changes
 Other Changes
 ^^^^^^^^^^^^^
 - ``POST @dossier-transfers``: New endpoint to create dossier transfers.
+- Add 'external_reference' field to workspace todo
```
lukasgraf, line 16: So we have two changelogs:
- The API changelog (this one), where we track significant changes to the REST API
- The regular changelog, where we track every nontrivial change to the application overall.

I technically wouldn't consider this an API change, therefore it doesn't need a changelog entry in the _API_ changelog.

If you consider the public API `GET /todo-xy` to be "Serializes all the fields of the todo object in the common response format" then that API didn't really change. The underlying data model did, but not necessarily the API itself.

This is a bit of a semantic point, because if we were to add a _required_ field to the data model, then that would indeed break POST (creation) requests from clients that don't include that field yet. So in that case, we would want to mention it in the API changelog as a a breaking change.

But for practical reasons, we don't really track every nonbreaking update to the data model in the _API_ changelog as well, because that would detract from its readability and defeat its purpose (a quick way for internal and external consumers of the API to check whether they need to make changes to their client code after an update).

So I would move the changelog entry from the API changelog to just the regular changelog. The regular changelog is in [HISTORY.rst](https://github.com/4teamwork/opengever.core/blob/master/HISTORY.rst), which is autogenerated during release from so called "news fragements" in the directory `/changes`. That directory is currently empty, because your's is the first change in this release :)

But here's a view of how `/changes` looked before the last release:
https://github.com/4teamwork/opengever.core/tree/b27d0c74d3a262e031690011a8640d19fa41d9bf/changes

So you can add a new file in there, with the filename being the Jira Ticket number, and the extension one of:
- `.other`
- `.bugfix`
- `.feature` (which I would pick in your case)

And the actual changelog line in the file would be `<changelog message> [<your-signature>]`.

lukasgraf (changes requested):
LGTM! 👍  - Just some very minor details 🍰 🌟 🍰

lukasgraf (approved):

PR #7897, opened 2024-02-16
Extend is_convertable in bumblebee service

KunzS85:
This PR extends the check for convertibility to PDF. If a file does not have a file extension, it will not be allowed for conversion via Bumblebee. 
Without this check, the file is sent for conversion but not processed and remains in the pickup folder. This must be avoided.

For [CA-6496]

## Checklist

- [x] Changelog entry

- New functionality:

[CA-6496]: https://4teamwork.atlassian.net/browse/CA-6496?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-6496.other
+++ b/changes/CA-6496.other
@@ -1,0 +1,1 @@
+ Extension of the is_convertabe method for suppressing the conversion to PDF in case of missing file extension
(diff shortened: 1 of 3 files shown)


lukasgraf (approved):
LGTM! 👍

PR #7909, opened 2024-03-07
Add ability to restrict workspace mails and documents actions for guests 

KunzS85:
This PR adds the option of restricting the action options for guests to the workspace. 
The aim of this is to ensure that guests can only view documents in the team room at this time and to make it as difficult as possible for them to print or download content. These restrictions relate primarily to the new UI.

For [CA-6417]

## Checklist

_Everything has to be done/checked. Checked but not present means the author deemed it unnecessary._

- [x] Changelog entry
- [x] Link to issue (Jira or GitHub) and backlink in issue (Jira)

_Only applicable should be left and checked._

- API change:
  - [x] API Changelog entry (see [guide](https://4teamwork.atlassian.net/wiki/spaces/4TEAM/pages/451248812/API+Changelog+Guidelines))
- New functionality:
  - [x] for `document` also works for `mail`
- New translations
  - [x] All msg-strings are unicode

[CA-6417]: https://4teamwork.atlassian.net/browse/CA-6417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7914, opened 2024-03-21
Extend the restriction of possible actions by guests

KunzS85:
Follow-up PR of: Add ability to restrict workspace mails and documents actions for guests #7909

For [CA-6417]

## Checklist

- [x] Changelog entry

[CA-6417]: https://4teamwork.atlassian.net/browse/CA-6417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/CA-6417.other
+++ b/changes/CA-6417.other
@@ -1,0 +1,1 @@
+Hide "share content" and "zipexport" actions for guest access on restricted workspaces [KunzS85]
--- a/opengever/document/actions.py
+++ b/opengever/document/actions.py
@@ -235,7 +235,8 @@ class BaseDocumentContextActions(BaseContextActions):
         return IBumblebeeable.providedBy(self.context)
 
     def is_share_content_available(self):
-        return is_within_workspace(self.context)
+        return (is_within_workspace(self.context)
+                and not is_restricted_workspace_and_guest(self.context))
 
     def is_trash_context_available(self):
         return self.file_actions.is_trash_context_action_available()
--- a/opengever/document/tests/test_actions.py
+++ b/opengever/document/tests/test_actions.py
@@ -263,10 +263,30 @@ class TestDocumentContextActions(IntegrationTestCase):
         ]
         self.assertEqual(expected_actions, self.get_actions(template))
 
+    def test_document_actions_in_workspace_with_guest(self):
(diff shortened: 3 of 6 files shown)


opengever/document/actions.py
```diff
@@ -246,6 +251,10 @@ def is_untrash_context_available(self):
     def is_unlock_available(self):
         return self.file_actions.is_unlock_available()
 
+    def is_save_document_as_pdf_available(self):
+        return (is_within_workspace(self.context)
```
lukasgraf, line 255: I think this (`is_within_workspace(self.context)`) might also disable this action for documents in GEVER, where it _should_ be available.

So I think this should just be
```
def is_save_document_as_pdf_available(self):
        return not is_restricted_workspace_and_guest(self.context)
```

lukasgraf (changes requested):
LGTM 👍 - just a small detail.

elioschmutz (commented):
A guest should not have the `Copy or Move` permission, so I wonder why the action is visible for guests. And a guest should generally not be able to save a document as a pdf because he would not have enough permission to do that. 

So I think these actions should be disabled in general for a guest and not only for guests in restricted workspaces.

opengever/document/actions.py
KunzS85: At first i had it the way you suggest, but some tests failed. So i changed it to the current way. But I'll check it more closely.

KunzS85: Due to the splitting of the issue, this method disappears from the class in this PR.

KunzS85:
The general restriction for guests regarding Copy/Move/Save as PDF via permission has been moved to a new bug issue (CA-6855)

changes/CA-6417.other
```diff
@@ -0,0 +1 @@
+Hide some more actions for guest access on restricted workspaces [KunzS85]
```
buchi, line 1: The changelog entry could be more specific:
```suggestion
Hide "share content" and "zipexport" actions for guest access on restricted workspaces [KunzS85]
```

buchi (approved):

PR #7916, opened 2024-03-22
Provide System Messages implemenation

Abdu-moustafa:
Provide widely System Message `CRUD `operations. Only the `manager `has access to the endpoints. The endpoints are defined as follows:

**GET**: This will fetch a list of all system messages or a single record when calling the URL with a system message `id `parameter. The endpoint will also deal with filtering. We can filter by `active`, which will return all active messages, or by `admin_unit_only`, which will return only messages to the `authenticated `user.
**POST**: Create a new System message.
**PATCH**: Update a record based on the record `id` parameter.
**DELETE**: Remove a record based on the record `id` parameter..
Additionally, we provide an additional data with the key `system_messages `on the` @config` endpoint.

For [CA-1852](https://4teamwork.atlassian.net/browse/CA-1852)

example GET request (list):  `@system-messages`
```python
{
  "@id": "http://localhost:8080/fd/@system-messages",
  "items": [
    {
      "@id": "http://localhost:8080/fd/@system-messages/29",
      "@type": "virtual.ogds.systemmessages",
      "admin_unit": "fd",
      "end": "2024-04-01T12:00:00+00:00",
      "id": 29,
      "start": "2024-03-26T12:00:00+00:00",
      "text": "EN",
      "text_de": null,
      "text_en": "EN",
      "text_fr": null,
      "type": "info"
    },
    {
      "@id": "http://localhost:8080/fd/@system-messages/12",
      "@type": "virtual.ogds.systemmessages",
      "admin_unit": null,
       ............
    },
 ]
} 
``` 

example GET message by message id :  `@system-messages/12`
```python
{
  "@id": "http://localhost:8080/fd/@system-messages/12",
  "@type": "virtual.ogds.systemmessages",
  "admin_unit": null,
  "end": "2024-03-27T12:00:00+00:00",
  "id": 12,
  "start": "2024-03-25T12:00:00+00:00",
  "text": "DE",
  "text_de": "DE",
  "text_en": "EN",
  "text_fr": "FR",
  "type": "warning"
}
``` 

example post request: `@system-messages`
example payload:
```python
{
  "admin_unit": "fd",
  "text_en": "EN",
  "text_fr": "fr",
  "text_de": "de",
  "start": "2024-03-26T12:00:00Z",
  "end": "2024-04-1T12:00:00Z",
  "type": "info"
}
``` 

example patch request: `@system-messages/12`

example payload:
```python
{
  "type": "error"
}
``` 
## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[CA-1852]: https://4teamwork.atlassian.net/browse/CA-1852?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -30,6 +30,12 @@ Nothing changed in this version.
 
 Other Changes
 ^^^^^^^^^^^^^
+- ``DELETE @system-messages/<id>``: New endpoint to delete a system message.
+- ``GET @system-messages/<id>``: New endpoint to fetch a system Message.
+- ``GET @system-messages``: New endpoint to list system messages.
+- ``POST @system-messages``: New endpoint to create system message.
+- ``GET @config`` : Include additional information about active system message
+- ``PATCH @system-messages/<id>``: New endpoint to update existing system message.
 - Include responsible_actor in tasktree response.
 - ``DELETE @dossier-transfers/<id>``: New endpoint to delete a dossier transfer.
 - ``GET @dossier-transfers/<id>``: New endpoint to fetch a dossier transfer.
--- a/opengever/base/systemmessages/api/base.py
+++ b/opengever/base/systemmessages/api/base.py
@@ -1,0 +1,15 @@
+from plone.restapi.interfaces import ISerializeToJson
+from plone.restapi.services import Service
+from zope.component import getMultiAdapter
+from zope.globalrequest import getRequest
+
+
+class SystemMessagesBase(Service):
+    """Base service class for all @system-messages endpoints.
+    """
+
+    def render(self):
+        return super(SystemMessagesBase, self).render()
+
+    def serialize(self, sys_msg):
+        return getMultiAdapter((sys_msg, getRequest()), ISerializeToJson)()
--- a/opengever/base/systemmessages/api/base.py
+++ b/opengever/base/systemmessages/api/base.py
@@ -1,7 +1,15 @@
+from opengever.base.date_time import utcnow_tz_aware
+from opengever.base.systemmessages.models import SystemMessage
+from opengever.ogds.base.utils import get_current_admin_unit
 from plone.restapi.interfaces import ISerializeToJson
 from plone.restapi.services import Service
+from sqlalchemy import and_
+from zExceptions import BadRequest
+from zExceptions import NotFound
 from zope.component import getMultiAdapter
 from zope.globalrequest import getRequest
+from zope.interface import implementer
+from zope.publisher.interfaces import IPublishTraverse
 
 
 class SystemMessagesBase(Service):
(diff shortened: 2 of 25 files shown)


opengever/base/systemmessages/api/patch.py
```diff
@@ -0,0 +1,68 @@
...
+        sys_msg_request_data = json_body(self.request)
+        local_instance = self.locate_message()
+
+        # We will serializer the system message instance `local_instance`
+        # pop (@id, @type, id, text) from the local_instance_serializer
+        # overwrite local_instance values with the request data `sys_msg_request_data`
+        # overwrite the serialized instance with the data from request payload
+        # send the data to the validation process
+
+        local_instance_serializer = self.serialize(local_instance)
```
lukasgraf, line 32: `existing_msg_serialized`?

opengever/base/systemmessages/api/patch.py
```diff
@@ -0,0 +1,68 @@
...
+        local_instance_serializer.pop("@id")
+        local_instance_serializer.pop("@type")
+        local_instance_serializer.pop("id")
+        local_instance_serializer.pop("text")
+
+        for key, value in sys_msg_request_data.items():
```
lukasgraf, line 38: `existing_msg_serialized.update(patch_data)`?

opengever/base/systemmessages/api/patch.py
```diff
@@ -0,0 +1,68 @@
...
+    """API endpoint to update an existing system message.
+
+    PATCH /@system-messages/25 HTTP/1.1
+    """
+
+    def reply(self):
+        # Disable CSRF protection
+        alsoProvides(self.request, IDisableCSRFProtection)
+
+        sys_msg_request_data = json_body(self.request)
```
lukasgraf, line 23: `patch_data`?

opengever/base/systemmessages/api/base.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        msg_id = self.msg_id
+        # Check if the message with the given ID exists in the SystemMessages
+        if not SystemMessages.get(msg_id):
+            # If not found, raise NotFound exception
+            raise NotFound
+
+        query = SystemMessages.query
+        query = query.filter(SystemMessages.id == msg_id)
+        # Get the first result from the query
```
lukasgraf, line 98: Again, these types of comments don't add much value 😁 - they simply duplicate what the code is already telling us.

If you're interested, this is a pretty interesting article about the use of comments (by the original author of Redis):
http://antirez.com/news/124

The article is well worth reading, but long. But the essence of it is:

> During my research I identified nine types of comments:
> 
> * Function comments
> * Design comments
> * Why comments
> * Teacher comments
> * Checklist comments
> * Guide comments
> * Trivial comments
> * Debt comments
> * Backup comments
> 
> The first six are, in my opinion, mostly very positive forms of commenting, while the **final three are somewhat questionable**. In the next sections each type will be analyzed with examples from the Redis source code.

lukasgraf (changes requested):
Great work! 🎉 

I have quite a lot of feedback, but that is to be expected for the first backend PR of this size! 😄 Overall it looks good to me, great job! 👍 

Tip: You should be able to re-run all your system-messages related tests tests locally quite quickly by doing

```bash
bin/test-cached -m opengever.base.systemmessages
```

instead of having to push every change to have it tested by Jenkins.

docs/public/dev-manual/api/api_changelog.rst
```diff
@@ -30,6 +30,13 @@ Nothing changed in this version.
 
 Other Changes
 ^^^^^^^^^^^^^
+- ``DELETE @system-messages/<id>``: New endpoint to delete a system message.
+- ``GET @system-messages/<id>``: New endpoint to fetch a system Message.
+- ``GET @system-messages``: New endpoint to list system messages.
+- ``GET @system-messages?active=true&current_admin_unit_only=true``: New endpoint to list filtered system messages (current_admin and active)
```
lukasgraf, line 36: This line in the changelog has now become outdated, since only `@system-messages?filters` is supported now.
Honestly, I would simply delete this line. The query string parameters are an implementation detail, and this level of detail is not really needed for a changelog.

lukasgraf (approved):
LGTM! 👍 

Just one line that needs to be removed from the API changelog.

PR #7924, opened 2024-04-04
Bump ftw.contentstats to version 1.4.0

buchi:
Add support for Docker-based deployments.

## Checklist

- [x] Changelog entry


docker/core/requirements-core.txt
```diff
@@ -275,6 +275,7 @@ requests==2.27.1
 RestrictedPython==3.6.0
 rfc6266==0.0.4
 roman==1.4.0
+scandir==1.10.0
```
lukasgraf, line 278: I think`scandir` also needs to be pinned in `versions.cfg`, no? Otherwise we'll get an error because of "picked versions"  when running buildout in production. Or am I missing something here?

lukasgraf (changes requested):

docker/core/requirements-core.txt
buchi: 👍🏻 added pinning in `versions.cfg`

lukasgraf (approved):

PR #7931, opened 2024-04-18
Restricted workspaces

elioschmutz:
This is a follow-up PR for #7909 

The PR disables some more ui-actions and protects endpoints for guests within a restricted workspace.

For [TI-16]

## Checklist

_Everything has to be done/checked. Checked but not present means the author deemed it unnecessary._

- [x] Changelog entry
- [x] Link to issue (Jira or GitHub) and backlink in issue (Jira)

- New functionality:
  - [x] for `document` also works for `mail`
  - [x] for `task` also works for `forwarding`

[TI-16]: https://4teamwork.atlassian.net/browse/TI-16?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7933, opened 2024-04-22
Remove save document as pdf action for guests in restricted workspaces

elioschmutz:
This PR removes the `Download as pdf` action for guests within a restricted workspace.

This is a follow-up PR for #7931 

We had tests for this case, but unfortunately, the tested document was nod a bumblebeeable document, thus the action was disabled by default in the tests. I split up this PR into two commits to make it clear what happens.

For [CA-6417]

## Checklist


lukasgraf (approved):
LGTM 👍 
Now I also understand why this was missed, if the `save_document_as_pdf` action didn't show up in tests before 😆

elioschmutz:
Jup, that was nasty 🙈

PR #7936, opened 2024-04-25
Dossier transfer create content

buchi:
Performs a dossier transfer at the given repository folder:
- First all documents are fetched and stored in a temporary directory.
- Then all contacts referenced in participations are created if they don't exist yet
- At last the dossier, its subdossiers and documents are created

Checking for existing contacts is done by comparing the `third_party_id`.
If the source doesn't have a `third_party_id`, its id is used instead.

If the creation of dossiers or documents fails, newly created contacts are kept, as they are created externaly in KuB.

For [TI-161](https://4teamwork.atlassian.net/browse/TI-161)

## Checklist

- [x] Changelog entry

- API change:

[TI-161]: https://4teamwork.atlassian.net/browse/TI-161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/kub/client.py
+++ b/opengever/kub/client.py
@@ -10,10 +10,13 @@ from time import time
 from urllib import urlencode
 from wsgiref.handlers import format_date_time
 from zope.annotation import IAnnotations
+import logging
 import requests
 
 
 KUB_API_VERSION = 'v2'
+
+logger = logging.getLogger('kub')
 
 
 class KuBClient(object):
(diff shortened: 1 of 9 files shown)


opengever/kub/client.py
```diff
@@ -99,3 +102,31 @@ def _update_kub_label_mapping(self):
...
+        try:
+            resp = self.session.post(url, json=data)
+            resp.raise_for_status()
+        except requests.RequestException:
+            logger.exception('Creating person failed')
+            return None
+        try:
+            return resp.json()
+        except requests.JSONDecodeError:
+            logger.exception('Creating person failed')
```
elioschmutz, line 117: Creating people was successful, but decoding is not possible...

elioschmutz (approved):
👍

opengever/kub/client.py
buchi: Could be, but I'm only considering it successful if I get the expected response.
The exception will be logged anyway, so it should be obvious to see what happened.

PR #7939, opened 2024-04-30
Disposition: Include document count and cumulative file size

lukasgraf:
Disposition: Include stats about document count and cumulative file size:

These stats are persisted on the disposition, and tracked by dossier. They get updated when dossiers are added or removed from the disposition. Since the dossiers are in a locked state 'offered' once added to a disposition, these stats should not get out of date.

Stats are queried from Solr, with one query per dossier.

For [TI-2](https://4teamwork.atlassian.net/browse/TI-2)

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
- API change:
  - [x] API Changelog entry

[TI-2]: https://4teamwork.atlassian.net/browse/TI-2?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/disposition/locales/de/LC_MESSAGES/opengever.disposition.po
```diff
@@ -142,6 +142,16 @@ msgstr "Angebot bearbeitet"
 msgid "label_dispositon_package_download"
 msgstr "Ablieferungspaket herunterladen"
 
+#. Default: "Documents"
+#: ./opengever/disposition/browser/excel_export.py
+msgid "label_docs_count"
+msgstr "Dokumente"
```
phgross, line 148: Anzahl Dokumente

phgross (approved):
LGTM 👍 

don't forget to update the fixture in the UI, if necessry.

opengever/disposition/locales/de/LC_MESSAGES/opengever.disposition.po
lukasgraf: As mentioned in Slack: I don't think this is beneficial in the Excel sheet. It gets more cumbersome because column names are longer, and it is already unambigously clear that "Documents" refers to a "Document count", because the column contains numbers.

PR #7948, opened 2024-05-08
Revert "Provide excel download for workspace users"

Abdu-moustafa:
Reverts 4teamwork/opengever.core#7938


Abdu-moustafa:
Worg implementation place it should be in workspace not under user and teams

PR #7951, opened 2024-05-08
Add dossier progress indexer

elioschmutz:
This PR adds a new solr-index `progress`.

We use this index for the checklist-feature of dossiers to track the progress of a dossier and list it in listings.

For [TI-3]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
- API change:


solr-conf/managed-schema
```diff
@@ -182,6 +182,7 @@
     <field name="trashed" type="boolean" indexed="true" stored="false" />
     <field name="watchers" type="string" indexed="true" stored="false" multiValued="true"/>
     <field name="participations" type="string" indexed="true" stored="true" multiValued="true"/>
+    <field name="progress" type="pfloat" indexed="true" stored="true" multiValued="false"/>
```
buchi, line 185: Storing the value is not necessary as fields of type `pfloat` have doc values.

```suggestion
    <field name="progress" type="pfloat" indexed="true" stored="false" multiValued="false"/>
```

opengever/dossier/behaviors/dossier.py
```diff
@@ -27,6 +27,11 @@
 
 LOG = logging.getLogger('opengever.dossier')
 
+CHECKLIST_OPEN_STATE = 'open'
+CHECKLIST_CLOSED_STATE = 'closed'
+
+CHECKLIST_CLOSED_STATES = [CHECKLIST_CLOSED_STATE]
```
buchi, line 33: Generally prefer a `set` to a `list` for containment checks. Time complexity is `O(1)` vs. `O(n)`.
Doesn't matter here as the list contains only one item

opengever/dossier/indexers.py
```diff
@@ -115,6 +116,22 @@ def dossier_type(obj):
     return IDossier(obj).dossier_type
 
 
+@indexer(IDossierMarker)
+def progress(obj):
+    """Returns the current progress of the dossier.
+
+    We need to differ between a checklist without any closed items or
+    a dossier without a checklist at all. We need this by returning "None"
```
buchi, line 124:
```suggestion
    a dossier without a checklist at all. We do this by returning "None"
```

opengever/dossier/tests/test_indexer.py
```diff
@@ -483,3 +484,52 @@ class TestParticipationIndexHelper(IntegrationTestCase):
...
+        self.dossier.reindexObject(["progress"])
+        self.commit_solr()
+
+        self.assertIsNone(solr_data_for(self.dossier, 'progress'))
+
+        self.activate_feature('dossier-checklist')
+        self.dossier.reindexObject(["progress"])
+        self.commit_solr()
+
+        self.assertEquals(1.0, solr_data_for(self.dossier, 'progress'))
```
buchi, line 509: Make a separate test for this? The method name is misleading...

buchi, line 509:
```suggestion
        self.assertEqual(1.0, solr_data_for(self.dossier, 'progress'))
```

opengever/dossier/tests/test_indexer.py
```diff
@@ -483,3 +484,52 @@ class TestParticipationIndexHelper(IntegrationTestCase):
...
+        self.dossier.reindexObject(["progress"])
+        self.commit_solr()
+
+        self.assertIsNone(solr_data_for(self.dossier, 'progress'))
+
+        IDossier(self.dossier).checklist = checklist_with_items
+        self.dossier.reindexObject(["progress"])
+        self.commit_solr()
+
+        self.assertEquals(1.0, solr_data_for(self.dossier, 'progress'))
```
buchi, line 535: Separate test?

opengever/dossier/tests/test_indexer.py
elioschmutz: i reorganized the tests. The feature is now activated by default and check if there is a progress indexed. After disabling the feature, I'll assert again to check if its `None`.

I also added another test to test the index at all.

opengever/dossier/tests/test_indexer.py
elioschmutz: same as above

PR #7950, opened 2024-05-08
Provide Excel Export for workspace participants

Abdu-moustafa:
Implement Endponit  for exporting workspace participants to Excel format.

For [TI-156](https://4teamwork.atlassian.net/browse/TI-156)

## Checklist

- [x] Changelog entry
- [x] Link to issue (Jira or GitHub) and backlink in issue (Jira)

[TI-156]: https://4teamwork.atlassian.net/browse/TI-156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7953, opened 2024-05-13
Remove any non ogdsuser from User excel export

Abdu-moustafa:
The Changes related to the PR [Workspace export participants](https://github.com/4teamwork/opengever.core/pull/7950/files).
We exclude any non ogds user from the Excel export

### DIFF
--- a/opengever/globalindex/browser/report.py
+++ b/opengever/globalindex/browser/report.py
@@ -107,6 +107,7 @@ class UserReport(BaseReporterView):
                     self.context.absolute_url())
 
         users = [ogds_service().fetch_user(user_id) for user_id in user_ids]
+        users = [user for user in users if user]
 
         reporter = XLSReporter(
             self.context.REQUEST,


opengever/globalindex/browser/report.py
```diff
@@ -107,6 +107,7 @@ def __call__(self):
                     self.context.absolute_url())
 
         users = [ogds_service().fetch_user(user_id) for user_id in user_ids]
+        users = [user for user in users if user]
```
elioschmutz, line 110:
```suggestion
        users = [ogds_service().fetch_user(user_id) for user_id in user_ids  if user]
```

Abdu-moustafa: @elioschmutz 
We need to access the users first so we can check if the user exists or not 
`users = [ogds_service().fetch_user(user_id) for user_id in user_ids  if user]` user is not accessiblehere

elioschmutz (approved):

PR #7955, opened 2024-05-15
Fix some dossier transfer issues

elioschmutz:
This PR adds some fixes for the dossier transfer:

- Properly handle ogds users as dossier participants when performing a dossier transfer
- Fix transfering emails
- Fix upgradestep
- Fix creation of people with memberships and/or username (remove it)
- Fix creation of organizations
- Add contact-guessing for people and organizations

For [TI-349]

## Checklist

[TI-349]: https://4teamwork.atlassian.net/browse/TI-349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/dossiertransfer/api/get.py
+++ b/opengever/dossiertransfer/api/get.py
@@ -1,3 +1,4 @@
+from ftw.mail.mail import IMail
 from opengever.base.behaviors.utils import set_attachment_content_disposition
 from opengever.base.security import elevated_privileges
 from opengever.document.behaviors import IBaseDocument
@@ -98,7 +99,11 @@ class DossierTransfersGet(DossierTransferLocator):
                     raise Unauthorized
 
             document = brains[0].getObject()
-            namedfile = document.file
+
+            if IMail.providedBy(document):
+                namedfile = document.message
+            else:
+                namedfile = document.file
 
         filename = getattr(namedfile, 'filename', u'document.bin').encode('utf-8')
         set_attachment_content_disposition(self.request, filename, namedfile)
--- a/opengever/dossiertransfer/api/perform.py
+++ b/opengever/dossiertransfer/api/perform.py
@@ -239,25 +239,8 @@ class PerformDossierTransfer(Service):
         for contact_id, contact_data in (
             self.metadata.get("content", {}).get("contacts", {}).items()
         ):
-            third_party_id = contact_data.get('thirdPartyId') or contact_id
-            existing = kub.list_people(filters={'third_party_id': third_party_id})
-            if existing is None:
-                raise InternalError('Checking for existing contact failed')
-
-            if existing['count'] == 1:
-                self.contact_id_mapping[contact_id] = "person:{}".format(
-                    existing["results"][0]["id"]
-                )
-                continue
-
-            if not contact_data.get('thirdPartyId'):
-                contact_data['thirdPartyId'] = contact_id
-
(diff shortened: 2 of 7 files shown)


opengever/dossiertransfer/api/get.py
```diff
@@ -98,7 +99,11 @@ def stream_document_blob(self):
                     raise Unauthorized
 
             document = brains[0].getObject()
-            namedfile = document.file
+
+            if IMail.providedBy(document):
+                namedfile = document.message
```
buchi, line 104: This way we loose the original file of an Outlook mail which is stored in the field `original_message`.
We can fix this in a later PR, but I think it's unexpected that the .msg file gets lost. And the converted .eml isn't necessarily a good representation.
The whole mail handling is a mess...

opengever/dossiertransfer/api/perform.py
```diff
@@ -239,25 +239,8 @@ def create_contacts(self):
...
-
-            if not contact_data.get('thirdPartyId'):
-                contact_data['thirdPartyId'] = contact_id
-
-            created_data = kub.create_person(contact_data)
-            if created_data is None:
-                raise InternalError('Creating contact failed')
-
-            self.contact_id_mapping[contact_id] = created_data['typedId']
+            self.contact_id_mapping[contact_id] = PersonContactSyncer(kub).sync(contact_id,
```
buchi: It doesn't make sense to instantiate this object in a loop when the only parameter has the same value every time.
Object instantiation (and destruction) is expensive. => instantiate it outside of the loop.

opengever/dossiertransfer/api/perform.py
```diff
@@ -240,8 +240,14 @@ def create_contacts(self):
...
             self.metadata.get("content", {}).get("contacts", {}).items()
         ):
-            self.contact_id_mapping[contact_id] = PersonContactSyncer(kub).sync(contact_id,
-                                                                                contact_data)
+            if contact_data.get('type') == 'person':
+                syncer = PersonContactSyncer(kub)
+            elif contact_data.get('type') == 'organization':
+                syncer = OrganizationContactSyncer(kub)
+            else:
+                raise NotImplementedError()
```
buchi, line 248: Maybe log a warning and just ignore unknown types?
Currently it will not work at all.

opengever/kub/client.py
```diff
@@ -117,6 +117,20 @@ def create_person(self, data):
             logger.exception('Creating person failed')
             return None
 
+    def create_organization(self, data):
+        url = '{}organizations'.format(self.kub_api_url)
+        try:
+            resp = self.session.post(url, json=data)
+            resp.raise_for_status()
+        except requests.RequestException:
+            logger.exception('Creating organization failed: {}'.format(resp.content))
```
buchi, line 126: You should not use `str.format` or f-strings for logging but instead printf style with positional parameters.
The advantages are:
- It's only evaluated if the required logging level is reached
- In log aggregation tools like Sentry you get only one event for all combinations of parameter values. With your approach you get a new (different) event for every new parameter value.

```suggestion
            logger.exception('Creating organization failed: %s', resp.content)
```

buchi (approved):
👍🏻 LGTM
Just a few minor things...

PR #7956, opened 2024-05-15
Inhance system information view with Docproperties

Abdu-moustafa:
Provide Docproperty information in `@system-information` view
For [TI-317](https://4teamwork.atlassian.net/browse/TI-317)

## Checklist

- [x] Changelog entry

[TI-317]: https://4teamwork.atlassian.net/browse/TI-317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/api/system_information.py
+++ b/opengever/api/system_information.py
@@ -24,6 +24,11 @@ class SystemInformationGet(Service):
             in primary_participation_roles(self.context)}
         for role in available_participation_roles:
             role['primary'] = role.get('token') in primary_participation_role_ids
+            role_id = role.get("token")
+            role['docproperty_key'] = {
+                "person": u"ogg.{}.person.*".format(role_id),
+                "organization": u"ogg.{}.organization.*".format(role_id)
+            }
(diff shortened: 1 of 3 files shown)


opengever/api/system_information.py
```diff
@@ -24,6 +24,11 @@ def add_dossier_participation_roles(self, infos):
             in primary_participation_roles(self.context)}
         for role in available_participation_roles:
             role['primary'] = role.get('token') in primary_participation_role_ids
+            role_id = role.get("token")
+            role['docproperty_key'] = {
+                "person": "ogg.{}.person.*".format(role_id),
```
phgross, line 29: Because its an id, and should not contain non ascii charaters the string formatting should be safe. but otherwise I would get into the habit of writing the key in unicode format so that you avoid unicode errors.

for example `u"ogg.{}.person.*".format(role_id),` because role_id is unicode as well.

phgross (approved):
LGTM 👍

PR #7960, opened 2024-05-24
Admin Unit Auth Plugin

buchi:
Provides a new PAS plugin for authenticating requests between admin units.
It replaces the existing "octopus tentacle" plugin.

The old plugin trusts requests from ip addresses stored in OGDS. This is not usable for containerized environments, as the ip address may change often.

The new plugin uses signed tokens with asymmetric cryptography. The private key, which is used to sign a token, is stored in the ZODB of the admin unit and the public key, which is used to verify a token, is stored in OGDS. This way all admin units can verify if a token was signed by a specific admin unit.

The token is transmitted in the `X-OGDS-AC` header and includes the signature, a timestamp, the admin unit id and the userid. The previously used header `X-OGDS-AUID` is no longer needed.

For new installations, the new plugin is installed instead of the old one.
On upgrade, the old octopus tentacle plugin gets removed and the new `admin_unit_auth` plugin is installed and activated. A new public/private key pair is generated on installation automatically. No manual setup work is necessary for the new plugin.

The plugin can be only installed if the admin unit already exists in OGDS. This is not the case when creating a policyless deployment. Therefore for policyless setups, the plugin is installed after the admin unit gets created during bundle import.

For [TI-61](https://4teamwork.atlassian.net/browse/TI-61)

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration

[TI-61]: https://4teamwork.atlassian.net/browse/TI-61?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


phgross (approved):
👍 LGTM

PR #7961, opened 2024-05-27
Provide docproperty_keys for fields

Abdu-moustafa:
Provide `docproperty keys` on `propertysheet fields`
Retalted to  [Enhance system information view with Docproperties](https://github.com/4teamwork/opengever.core/pull/7956)


phgross (changes requested):
Imho there is a test missing.

phgross (approved):

PR #7967, opened 2024-06-05
Fix access for workspace admin on export users

Abdu-moustafa:
- Fix workspace admin permission when export workspace users.
For [TI-506](https://4teamwork.atlassian.net/browse/TI-506)

## Checklist

- [x] Changelog entry

[TI-506]: https://4teamwork.atlassian.net/browse/TI-506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/TI-506.bugfix
+++ b/changes/TI-506.bugfix
@@ -1,0 +1,1 @@
+Provide ui-action to export workspace users by adding a new permission.[amo]
--- a/opengever/base/context_actions.py
+++ b/opengever/base/context_actions.py
@@ -69,6 +69,7 @@ class BaseContextActions(object):
         self.maybe_add_untrash_context()
         self.maybe_add_zipexport()
         self.maybe_add_save_minutes_as_pdf()
+        self.maybe_export_workspace_user()
         return self.actions
 
     def add_action(self, action):
@@ -243,6 +244,9 @@ class BaseContextActions(object):
         return False
 
     def is_add_save_minutes_as_pdf_available(self):
+        return False
+
+    def is_export_workspace_users_available(self):
         return False
 
     def maybe_add_add_invitation(self):
@@ -468,3 +472,7 @@ class BaseContextActions(object):
     def maybe_add_save_minutes_as_pdf(self):
         if self.is_add_save_minutes_as_pdf_available():
             self.add_action(u'save_minutes_as_pdf')
+
+    def maybe_export_workspace_user(self):
+        if self.is_export_workspace_users_available():
+            self.add_action(u'export_workspace_participators')
--- a/opengever/workspace/configure.zcml
+++ b/opengever/workspace/configure.zcml
@@ -160,7 +160,7 @@
       name="workspace_participants_report"
       for="*"
       class=".report.WorkspaceParticipantsReport"
-      permission="opengever.workspace.AccessAllUsersAndGroups"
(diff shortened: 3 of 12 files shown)


elioschmutz (changes requested):
This will not work as expected. The view itself is also protected: https://github.com/4teamwork/opengever.core/blob/master/opengever/workspace/configure.zcml#L163 and the user would lose the permission for inactive workspaces (see the workflow for more information). So this would end in the same issue.

So there are two options to solve this issue:

option 1: remove the permission in the `configure.zcml` and use `zoe2.View` instead. The you have to check for a specific role as it is already done for the user-report: https://github.com/4teamwork/opengever.core/blob/master/opengever/base/utils.py#L250 (`api.user.get_current().has_role('WorkspaceAdmin)`

option 2: introduce a new permission: `opengever.workspace.ExportWorkspaceParticipants` and use this permission for the view in the `.zcml` and completely remove the `check_permission`-implementation here.

In both cases, you should also update the UI because the export button is currently not visible for deactivated workspaces. So the UI should show the button if a ui-action `export_workspace_participants` is available. This action needs to be added in the backend first.

opengever/workspace/configure.zcml
```diff
@@ -160,7 +160,7 @@
       name="workspace_participants_report"
       for="*"
       class=".report.WorkspaceParticipantsReport"
-      permission="opengever.workspace.AccessAllUsersAndGroups"
+      permission="zope2.View"
```
elioschmutz, line 163: Follow-Up of my review comment: https://github.com/4teamwork/opengever.core/pull/7967#pullrequestreview-2104148458:

You've chosen option one, implementing as we did it already for the `UserReport`.

I thought about it and I'd prefer options 2. It is the preferred and default way how we do permission checks and there is no reason why we should implement a custom permission check based on roles.

You can protect this view with a new permission `opengever.workspace.ExportWorkspaceParticipants` and remove the `check_permissions` in the `report.py` since zope will do the permission-check for you.

The permission-check in the actions endpoint will no longer implement logic but only the permission check:

`return api.user.has_permission('opengever.workspace.ExportWorkspaceParticipants', obj=self.context)`

Don't forget to update the lawgiver-workflow. You can call me as soon as you need to update it.

changes/TI-506.bugfix
```diff
@@ -0,0 +1 @@
+Fix export workspace users did not work for workspace admin [amo]
```
elioschmutz, line 1: It was not broken in the backend, so you do not fix anything with this PR. The UI was not able to properly find out when to show the export button.

So in this PR you provide a ui action to tell the UI if the user has enough permission to export the user listing. I'd write something like: `Provide ui-action to export workspace users.`

opengever/base/context_actions.py
```diff
@@ -69,6 +69,7 @@ def get_actions(self):
         self.maybe_add_untrash_context()
         self.maybe_add_zipexport()
         self.maybe_add_save_minutes_as_pdf()
+        self.maybe_export_workspace_user()
```
elioschmutz, line 72:
```suggestion
        self.maybe_export_workspace_users()
```

elioschmutz (approved):
👍 thanks!

PR #7971, opened 2024-06-06
Fix Document date after drag and drop from E-Mail

Abdu-moustafa:
Fix Update document date on Drag & Drop from Outlook Mail.

For [TI-458](https://4teamwork.atlassian.net/browse/TI-458)

**Before:**

before

**After:**

after

## Checklist

- [x] Changelog entry

[TI-458]: https://4teamwork.atlassian.net/browse/TI-458?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/api/tests/test_tus.py
+++ b/opengever/api/tests/test_tus.py
@@ -5,6 +5,8 @@ from ftw.builder import Builder
 from ftw.builder import create
 from ftw.testbrowser import browsing
 from opengever.document.behaviors.customproperties import IDocumentCustomProperties
+from opengever.mail.mail import OGMail
+from opengever.mail.tests import MAIL_DATA
 from opengever.private.interfaces import IPrivateFolderQuotaSettings
 from opengever.propertysheets.storage import PropertySheetSchemaStorage
 from opengever.testing import IntegrationTestCase
@@ -331,3 +333,14 @@ class TestTUSUpload(IntegrationTestCase):
         upload_metadata = UPLOAD_METADATA + ',document_date MjAxNS0wOC0yMQ=='
         self.assert_tus_replace_succeeds(self.document, browser, headers={"Upload-Metadata": upload_metadata})
         self.assertEqual(current_date, self.document.document_date)
+
+    @browsing
+    def test_add_additional_metadata_for_ogmail(self, browser):
+        self.login(self.regular_user, browser)
+        mail = create(Builder('mail').with_message(MAIL_DATA))
+        mail_document_date_b64 = b64encode(mail.document_date.isoformat().encode()).decode()
+        upload_metadata = UPLOAD_METADATA + ',document_date {}'.format(mail_document_date_b64)
+        doc = self.assert_tus_upload_succeeds(
+            self.dossier, browser, upload_metadata=upload_metadata
+        )
+        self.assertEqual(mail.document_date, doc.document_date)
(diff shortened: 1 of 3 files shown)


elioschmutz (changes requested):
Please see the jira-comment from @phgross : https://4teamwork.atlassian.net/browse/TI-458?focusedCommentId=45534

The issue in this story is, that the document date for **mails** are not set properly. The TUS-Upload should use the Mail-Date as the document date, not the current date.

For testing: you should not adjust the existing test but add another test for mail to test if the correct date will be extracted.

opengever/api/tus.py
```diff
@@ -70,9 +70,15 @@ def fix_content_type(self, tus_upload):
     def add_additional_metadata(self, metadata, obj):
         if not metadata.get("mode", "create") == 'create':
             return
+        from opengever.mail.mail import OGMail
```
phgross, line 73: imports do not belong here

opengever/mail/tests/test_mail_metadata.py
```diff
@@ -128,6 +128,11 @@ def test_mail_catalog_metadata(self):
                           brain.checked_out,
                           'Checked out, should be empty')
 
+    def test_initialize_metadata_sets_document_date(self):
```
phgross, line 131: I think this test would still pass without your fix, because the tus upload ist not used when creating via `create_mail`.

phgross (changes requested):

changes/TI-458.other
```diff
@@ -0,0 +1 @@
+Ensure the document date displays accurately after drag-and-drop actions. [amo]
```
phgross, line 1: It's not a display issue. 


```suggestion
Fix document_date when uploading mails with DnD, use mail header date. [amo]
```

opengever/mail/tests/test_mail_metadata.py
Abdu-moustafa: i rremoved this test and provide a test on `test_tus.py`  `test_add_additional_metadata_for_ogmail` so we can test the `mail upload` via tus

changes/TI-458.other
```diff
@@ -0,0 +1 @@
+Fix document_date when uploading mails with DnD, use mail header date. [amo]
```
elioschmutz, line 1: the filename should be `.bugfix` not `.other`

opengever/api/tests/test_tus.py
```diff
@@ -331,3 +332,40 @@ def test_tus_replace_does_not_update_document_date(self, browser):
...
+            u'IDocument.default': {
+                u'languages': set([u'de', u'en']),
+                u'notrequired': u'Not required, still has default',
+            },
+        }
+        self.assertEqual(
+            expected_defaults,
+            IDocumentCustomProperties(uploaded_doc).custom_properties
+        )
+        self.assertEqual(mail.document_date, uploaded_doc.document_date)
```
elioschmutz, line 371: The test seems quite complicated and unrelated to your change.

Please, take a look at another test for documents: i.e.: `test_tus_upload_can_set_document_date` as as starting point and then adjust it.

In general, you need to upload a mail by tus and then check if the date is set correctly. To do this, you need to extend the `assert_tus_upload_succeeds`-method to allow another upload-data. Then you can pass `MAIL_DATA` from `from opengever.mail.tests import MAIL_DATA`.

The assert-statement should then validate, that the mail-header date will be used instead the metadata-date.

phgross (approved):
👍

PR #7978, opened 2024-06-13
Respect new file mimetype when retrieve a file from teamraum as a new version

elioschmutz:
When changing the mimetype of a linked workspace document in teamraum an then retrieve it in gever by creating a new version of a gever-document, the new version will be broken. 

We also have to set the mimetype of the document do be in sync with the blob data.

For [TI-366]

## Checklist

- [x] Changelog entry

[TI-366]: https://4teamwork.atlassian.net/browse/TI-366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/TI-366.bugfix
+++ b/changes/TI-366.bugfix
@@ -1,0 +1,1 @@
+Respect new file mimetype when retrieve a file from teamraum as a new version. [elioschmutz]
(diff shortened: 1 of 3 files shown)


buchi (approved):
👍🏻 LGTM

PR #7993, opened 2024-07-12
Add serialized source dossier to dossier transfer if possible

elioschmutz:
This PR adds the summary serialized brain of the `root` dossier if requesting dossier-transfers through the `@dossier-transfer` endpoint or directly requesting a transfer with `@dossier-transfer/1`.

The root-dossier will only be serialized if:
1. the root is stored on the current admin-unit
2. the user has permission to view the dossier

**Why don't we add the `root_item` to the dossier-transfer serializer?**
The reason is performance. A dossier-transfer is stored in the ogds. The serializer would need to know if the source dossier is stored on the current admin unit and would need to do a catalog query to lookup the dossier. We'd need to do this for every dossier transfer.

For [TI-706]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
- API change:

[TI-706]: https://4teamwork.atlassian.net/browse/TI-706?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/TI-706.other
+++ b/changes/TI-706.other
@@ -1,0 +1,1 @@
+Expose the serialized source dossier if possible when using the "@dossier-transfer" endpoint. [elioschmutz]
--- a/docs/public/dev-manual/api/api_changelog.rst
+++ b/docs/public/dev-manual/api/api_changelog.rst
@@ -12,7 +12,7 @@ Breaking Changes
 
 Other Changes
 ^^^^^^^^^^^^^
-
+- ``@dossier-transfer``: Returns the serialized root_item if it is located on the same admin unit and the current user has view permission.
 
 2024.10.0 (2024-07-15)
 ----------------------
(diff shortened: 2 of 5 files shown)


changes/TI-706.other
```diff
@@ -0,0 +1 @@
+Expose the serilaized source dossier if possible when using the "@dossier-transfer" endpoint. [elioschmutz]
```
phgross, line 1: typo: serilaized

docs/public/dev-manual/api/api_changelog.rst
```diff
@@ -12,7 +12,7 @@ Breaking Changes
 
 Other Changes
 ^^^^^^^^^^^^^
-
+- ``@dossier-transfer``: Return the serialized root_item if it is on the same admin unit and the current user has view permission.
```
phgross, line 15: Returns the serialized root_item if it is located on the same admin unit and the current user has view permission.

phgross (approved):
👍 besides the things mentioned.

PR #7999, opened 2024-07-18
Create Journal entry in source dossier after transfer

Abdu-moustafa:
Create a `journal entry` in the `source` Dossier after dossier `transfer`.

For [TI-710](https://4teamwork.atlassian.net/browse/TI-710)

**Before:**
Before

**After:**
After

## Checklist

- [x] Changelog entry

- New translations

[TI-710]: https://4teamwork.atlassian.net/browse/TI-710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/dossiertransfer/api/post.py
+++ b/opengever/dossiertransfer/api/post.py
@@ -5,14 +5,20 @@ from opengever.dossiertransfer.api.base import DossierTransfersBase
 from opengever.dossiertransfer.api.schemas import IDossierTransferAPISchema
 from opengever.dossiertransfer.model import DossierTransfer
 from opengever.dossiertransfer.model import TRANSFER_STATE_PENDING
+from opengever.journal.handlers import journal_entry_factory
 from opengever.ogds.base.utils import get_current_admin_unit
 from opengever.ogds.models.service import ogds_service
 from plone import api
 from plone.protect.interfaces import IDisableCSRFProtection
 from plone.restapi.deserializer import json_body
+from Products.CMFPlone.utils import safe_unicode
 from zExceptions import BadRequest
 from zExceptions import Unauthorized
+from zope.i18nmessageid import MessageFactory
 from zope.interface import alsoProvides
+
+
+_ = MessageFactory('opengever.dossiertransfer')
 
 
 class DossierTransfersPost(DossierTransfersBase):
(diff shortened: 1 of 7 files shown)


opengever/dossiertransfer/api/post.py
```diff
@@ -76,4 +82,32 @@ def reply(self):
...
+        if not root_obj:
+            return
+
+        target_user = safe_unicode(transferred_dossier.target.title)
+        source_dossier_title = safe_unicode(transferred_dossier.title)
+        transfer_id = safe_unicode(transferred_dossier.id)
+        title = _(
+            u'label_journal_entry_dossier_transferred_for_source_dossier',
+            default=u'Dossier with the title (${source_dossier_title}) and '
+            '(Transfer-ID: ${transfer_id}) was handed over to (${target_user})',
```
elioschmutz, line 102: Please use:
`Dossier was handed over to ${target_unit_title}. (Transfer-ID: ${transfer_id})` for english and
`Dossier wurde an ${target_unit_title} übergeben. (Transfer-ID: ${transfer_id})` for german

We're already in the source dossier here, so there is no need for adding the dossier title again. and there are too many brackets at all and the variables seems to be wrong. We do not use the `source_dossier_title` nor the `target_user`.

opengever/dossiertransfer/api/post.py
```diff
@@ -76,4 +82,29 @@ def reply(self):
...
+            default='Dossier was handed over to ${target_unit_title}. (Transfer-ID: ${transfer_id})',
+            mapping=dict(
+                transfer_id=transfer_id,
+                target_unit_title=target_unit_title
+            )
+        )
+
+        journal_entry_factory(
+            root_obj,
+            "Dossier transferred",
```
elioschmutz, line 109: You've already used this type name for the journal entry within the target dossier: https://github.com/4teamwork/opengever.core/pull/7996/files

Please, use another type name here to separate these two entreis.

elioschmutz (approved):
👍 

See my comment about the dossier-type and please fixup the translation-commit into the first commit. It's part of the first commit.

phgross (approved):

PR #8010, opened 2024-08-13
Make cancellation of subtasks recursive

Acrop146:
Fixed cancellation of nested tasks by making them recursive. Before only direct subtasks got cancelled. Now all levels of subtasks get cancelled as well.
 IDEA: I noticed that when you reopen a task, the subtasks dont automatically get reopened. Maybe we should change that. Thoughts?

Before:
546_Before

After:
546_After

For [TI-546]

## Checklist

- [X] Changelog entry

- New functionality:
  - [?] for `task` also works for `forwarding`


lukasgraf (changes requested):
LGTM! 👍 
But please add a regression test that covers the bug in question. Maybe in `opengever/task/tests/test_task.py`?

lukasgraf (approved):
Great, exactly what I had in mind! 😃

PR #8011, opened 2024-08-14
Add customFields support in IKUBSettings to include them in Docprops

Abdu-moustafa:
This PR introduces :

- Added configurable DocProperties
 - Introduced a new registry property to define which KuB fields are available as DocProperties 
 - Updated the DocProperties Crawlers to respect the new configuration without causing errors if a field is unavailable.

For [TI-691](https://4teamwork.atlassian.net/browse/TI-691)

**Default (Before):**
Before

**With Custom Doc Props (After)**

after

## Checklist

- [x] Changelog entry

[TI-691]: https://4teamwork.atlassian.net/browse/TI-691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/kub/docprops.py
+++ b/opengever/kub/docprops.py
@@ -1,6 +1,21 @@
 from datetime import datetime
 from opengever.base.addressblock.docprops import get_addressblock_docprops
 from opengever.base.docprops import BaseDocPropertyProvider
+from plone.registry.interfaces import IRegistry
+from zope.component import getUtility
+
+
+def get_additional_doc_properties():
+    registry = getUtility(IRegistry)
+    return registry['opengever.kub.interfaces.IKuBSettings.additional_docproperty_fields'] or []
+
+
+def update_default_fields_with_additional_doc_props(context, default_fields):
+    custom_values = context.get('customValues')
+    if custom_values:
+        for field in get_additional_doc_properties():
+            if field in custom_values:
+                default_fields.update({field: custom_values[field]})
 
 
 class KuBEntityDocPropertyProvider(object):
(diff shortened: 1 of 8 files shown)


opengever/kub/docprops.py
```diff
@@ -1,6 +1,22 @@
...
+from zope.component import getUtility
+
+
+def get_additional_doc_properties():
+    registry = getUtility(IRegistry)
+    return registry['opengever.kub.interfaces.IKuBSettings.additional_docproperty_fields'] or []
+
+
+def update_default_fields_with_additional_doc_props(context, default_fields):
+    custom_values = context.get('customValues', None)
```
lukasgraf, line 14: `None` is already the default for `dict.get()` - no need to specify it ✂️

opengever/kub/docprops.py
```diff
@@ -1,6 +1,22 @@
...
+    if custom_values:
+        for field in get_additional_doc_properties():
+            value = custom_values.get(field, None)
+            if value:
```
lukasgraf, line 18: This will not work for legitimate values that are falsy - for example the integer `0`.

I would instead write this in a way where if the dict with the custom properties contains an entry for the field, it always gets added to the docproperties, no matter what exactly the value is. For example:

```python
for field in get_additional_doc_properties():
    if field in custom_values:
        default_fields.update({field: custom_values[field]})
```

lukasgraf (changes requested):
LGTM! 🎉 

Just a few minor details.

PR #8012, opened 2024-08-14
Add dossier_review_state solr index.

elioschmutz:
This PR implements the `dossier_review_state` index. The index provide the state of the current dossier or the closest parent dossier for other content types.

The use case for this index is to filter by contents within a dossier of a specific state. I.e. all documents within an inactive dossier.

ℹ️ there is no need to recreate the solr docker image. The `dossier_review_state` is already defined within the solr-schema: https://github.com/4teamwork/opengever.core/blob/master/solr-conf/managed-schema#L144 but is not in use. I just reused this field for the new indexer.

For [TI-479]

## Checklist

- [x] Changelog entry

[TI-479]: https://4teamwork.atlassian.net/browse/TI-479?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/base/handlers.py
+++ b/opengever/base/handlers.py
@@ -25,6 +25,7 @@ from opengever.workspace.interfaces import IToDoList
 from opengever.workspace.interfaces import IWorkspace
 from opengever.workspace.interfaces import IWorkspaceMeeting
 from plone.app.workflow.interfaces import ILocalrolesModifiedEvent
+from Products.CMFCore.interfaces import IFolderish
 from Products.CMFPlone.interfaces import IPloneSiteRoot
 from sqlalchemy import and_
 from zope.component import getUtility
@@ -273,3 +274,18 @@ def update_touched_date_for_move_event(obj, event):
     """
     if obj == event.object:
         update_touched_date(obj, event)
+
+
+def recursive_update_dossier_review_state(context):
+    context.reindexObject(idxs=["dossier_review_state"])
+    if not IFolderish.providedBy(context):
+        return
+
+    for obj in context.listFolderContents():
+        if IDossierMarker.providedBy(obj):
+            continue
+        recursive_update_dossier_review_state(obj)
+
+
+def update_dossier_review_state(obj, event):
+    recursive_update_dossier_review_state(obj)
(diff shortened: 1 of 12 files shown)


opengever/base/handlers.py
```diff
@@ -273,3 +274,18 @@ def update_touched_date_for_move_event(obj, event):
...
+    if not IFolderish.providedBy(context):
+        return
+
+    for obj in context.listFolderContents():
+        if IDossierMarker.providedBy(obj):
+            continue
+        recursive_update_dossier_review_state(obj)
+
+
+def update_dossier_review_state_event(obj, event):
```
lukasgraf, line 290: Remove the `_event` from this method's name?

opengever/core/upgrades/20240814101526_add_dossier_review_state_index/upgrade.py
```diff
@@ -0,0 +1,18 @@
...
+from plone.dexterity.interfaces import IDexterityContent
+
+
+class AddDossierReviewStateIndex(UpgradeStep):
+    """Add dossier review state index.
+    """
+
+    def __call__(self):
+        query = {'object_provides': [
+            IDexterityContent.__identifier__,
```
lukasgraf, line 12: Can't we be more specific here? For example, do we really need to reindex repository roots and repository folders?

opengever/dossier/indexers.py
```diff
@@ -204,6 +204,31 @@ def containing_subdossier(obj):
...
+
+@indexer(IDexterityContent)
+def dossier_review_state(obj):
+    """Returns the review state of the current dossier or of the dossier of
+    which the current content is contained in.
+    """
+    if obj.portal_type not in TYPES_WITH_DOSSIER_REVIEW_STATE_INDEX:
+        return ''
+
+    while not ISiteRoot.providedBy(obj):
```
lukasgraf, line 225: I would write this as `while obj and not ISiteRoot.providedBy(obj) :`

Otherwise an `None` object might lead to an infinite loop (`aq_parent` returns `None` for objects that aren't acquisition wrapped)

lukasgraf (changes requested):
LGTM 👍 - just some minor details.

elioschmutz:
@lukasgraf could you take another look?

lukasgraf (approved):
LGTM! 👍

PR #8014, opened 2024-08-16
Improve workspace users vocabulary to properly handle no longer available users for restricted workspaces

Abdu-moustafa:
This PR:

- Refactored vocabulary handling to return 'not found' for invalid  users and exceptions. 
- Added tests to ensure correct term management and error handling.

For [TI-922](https://4teamwork.atlassian.net/browse/TI-922)

## Checklist
- [x] Changelog entry

[TI-922]: https://4teamwork.atlassian.net/browse/TI-922?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/TI-922.other
+++ b/changes/TI-922.other
@@ -1,0 +1,1 @@
+Improve workspace users vocabulary to properly handle no longer available users for restricted workspaces [amo]
--- a/opengever/workspace/tests/test_vocabularies.py
+++ b/opengever/workspace/tests/test_vocabularies.py
@@ -57,6 +57,7 @@ class TestPossibleWorkspaceFolderParticipantsVocabulary(IntegrationTestCase):
 
 
 class TestWorkspaceContentMembersVocabulary(IntegrationTestCase):
+    features = ('workspace', )
 
     def test_vocabulary_list_all_members_of_the_current_workspace(self):
         self.login(self.workspace_member)
@@ -135,3 +136,31 @@ class TestWorkspaceContentMembersVocabulary(IntegrationTestCase):
         self.assertEquals(
             self.workspace_guest.getId(),
             factory(context=workspace).getTerm(self.workspace_guest.getId()).token)
+
+    def test_returns_notfound_user_without_access_all_participants_permissions(self):
+        self.login(self.workspace_admin)
+        ogds_user = create(Builder('ogds_user').id(u'peter'))
+
+        ogds_user.session.flush()
+
+        factory = getUtility(
+            IVocabularyFactory,
+            name='opengever.workspace.WorkspaceContentMembersVocabulary'
+        )
+
+        term = factory(context=self.workspace).getTerm(ogds_user.userid)
+        self.assertEqual(ogds_user.userid, term.token)
+
+        self.login(self.workspace_guest)
+        term = factory(context=self.workspace).getTerm(ogds_user.userid)
+        self.assertEqual('<not-found>', term.token)
+
+    def test_vocabulary_for_non_existent_user(self):
+        self.login(self.workspace_member)
+        factory = getUtility(
+            IVocabularyFactory,
+            name='opengever.workspace.WorkspaceContentMembersVocabulary'
+        )
+        non_existent_user_id = 'nonexistent_user'
+        term = factory(context=self.workspace).getTerm(non_existent_user_id)
+        self.assertEqual(term.token, '<not-found>')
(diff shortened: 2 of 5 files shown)


opengever/workspace/vocabularies.py
```diff
@@ -74,7 +74,12 @@ def getTerm(self, value):
         try:
             return super(WorkspaceUserVocabulary, self).getTerm(value)
         except LookupError:
-            return AllUsersSource(api.portal.get()).getTermByToken(value)
+            try:
+                # Attempt to get the term from AllUsersSource
+                return AllUsersSource(api.portal.get()).getTermByToken(value)
+            except (LookupError):
+                # If term not found reutn  "not found" as a string
+                return "not found"
```
elioschmutz, line 82: please, return a `SimpleTerm` and a value of `<not-found>` : `return SimpleTerm(value, value, value)`

And let's do the following check here before calling the `AllUsersSource`. This let's us return the `not found` before even calling the vocabulary. We then don't need to catch the `LookupError` of the `AllUsersSource`. So something like that: 
  

```suggestion
            if not visible_users_and_groups_filter.can_access_principal(value):
                return SimpleTerm( `<not-found>`,  `<not-found>`,  `<not-found>`)
            return AllUsersSource(api.portal.get()).getTermByToken(value)
```

elioschmutz (commented):
Thanks! You did it as discussed in advanced 👍 . But I think, we should only return the not found user if the current user is not allowed to access all users. See my suggestion. In all other cases, the user **must** exist because we don't remove users from the ogds. So please, also update the test for this case.

How does it look in the UI?

opengever/workspace/tests/test_vocabularies.py
```diff
@@ -135,3 +135,35 @@ def test_meeting_attendees_persist_after_user_role_removal(self):
...
+            .with_roles('WorkspaceMember')
+        )
+        ogds_user = create(
+            Builder('ogds_user')
+            .id(inactive_user.getId())
+            .having(email='inactive@example.com',)
+        )
+        ogds_user.session.flush()
+        term = factory(context=self.workspace).getTerm(inactive_user.getId())
+        self.assertEqual('inactive_user', term.token)
```
elioschmutz, line 158: The title of the test tells me, that the vocabulary should return not-found for inactive user but it does not 😉 

It will return the not-found user only when the current logged in user does not have the `opengever.workspace: Access all users and groups` permission.

So, first of all, the test-class should activate and test with the workspace-feature:

```
class TestWorkspaceContentMembersVocabulary(IntegrationTestCase):

    features = ('workspace', )
```

then, remove the guest from the workspace and check the vocabulary with the admin and then with the member:

```
    self.login(self.workspace_admin)
    factory = getUtility(
            IVocabularyFactory,
            name='opengever.workspace.WorkspaceContentMembersVocabulary'
        )
    role_assignment_manager = RoleAssignmentManager(workspace)
    role_assignment_manager.clear_by_cause_and_principal(ASSIGNMENT_VIA_SHARING, 
self.workspace_guest.getId())
    term = factory(context=self.workspace).getTerm(inactive_user.getId())
    self.assertEqual('xxx', term.token)

    self.login(self.workspace_member)
    self.assertEqual('<not-found>', term.token)
```

elioschmutz (commented):
Please, update the PR-title and description with a more meaningful text.

I also wrote a suggestion for a better test case. The current test seems not to test the right things... 

And could you provide a printscreen from the UI how it handles the `<not-found>` user?

opengever/workspace/tests/test_vocabularies.py
Abdu-moustafa: Fixed as discussed

changes/TI-922.other
```diff
@@ -0,0 +1 @@
+Refactored vocabulary handling to return '<not found>' for invalid / inactive users. [amo]
```
elioschmutz, line 1:
```suggestion
Improve workspace users vocabulary to properly handle no longer available users for restricted workspaces [amo]
```

Please, also update the PR title.

elioschmutz (approved):
👍 

see my suggestion for the changelog and PR title.

changes/TI-922.other
Abdu-moustafa: Updated as suggested

PR #8019, opened 2024-08-22
Add field country (nationality) to person docprops

Abdu-moustafa:
Add `nationality` field to person docprops

For [TI-691](https://4teamwork.atlassian.net/browse/TI-691)

## Checklist

- [x] Changelog entry

[TI-691]: https://4teamwork.atlassian.net/browse/TI-691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/api/tests/test_templatefolder.py
```diff
@@ -357,7 +357,8 @@ def test_creates_document_from_template_with_kub_recipient(self, mocker, browser
             ('ogg.recipient.person.firstname', 'Jean'),
             ('ogg.recipient.person.lastname', 'Dupont'),
             ('ogg.recipient.person.salutation', 'Herr'),
-            ('ogg.recipient.phone.number', '666 666 66 66')]
+            ('ogg.recipient.phone.number', '666 666 66 66'),
+            ('ogg.recipient.person.nationality', 'schweiz'),]
```
elioschmutz, line 361: could you move the new property (also in all upcoming tests) to the other `ogg.recipient.person` properties to have them grouped properly?

opengever/kub/docprops.py
```diff
@@ -124,6 +124,7 @@ def _collect_properties(self):
             'lastname': self.context.get("officialName"),
             'date_of_birth': date_of_birth,
             'sex': self.context.get('sex'),
+            'nationality': self.context.get('country'),
```
elioschmutz, line 127: The `country` and the `nationality` are not exactly the same. For example, I'm from `Switzerland` (country) so my nationality is `swiss`.

Since `kub` defines the `country`, I'd also name it `country` in the docprops.

elioschmutz (changes requested):
the implementation is good 👍 

Please take a closer look at the naming.

opengever/api/tests/test_templatefolder.py
Abdu-moustafa: Modified as suggested

elioschmutz (approved):

PR #8020, opened 2024-08-22
Include user stats in content stats.

elioschmutz:
This PR includes the users stats into the content stats:

`@@content-stats-json?stat=user_stats`

```
{
   "total_active_unique_users":26,
   "total_active_users_logged_in_last_30_days":0,
   "total_active_users_logged_in_last_365_days":2,
   "total_active_users_never_logged_in":23
   "total_active_unique_spv_users":26,
}
```

For [TI-911]

## Checklist

- [x] Changelog entry


opengever/contentstats/active_user_stats.py
```diff
@@ -0,0 +1,48 @@
...
+            'total_active_users_logged_in_last_30_days': self.calc_total_users_logged_in_last_x_days(30),
+            'total_active_users_logged_in_last_365_days': self.calc_total_users_logged_in_last_x_days(365),
+            'total_active_users_never_logged_in': self.calc_total_users_never_logged_in(),
+            'total_active_spv_users': self.calc_total_active_unique_spv_users(),
+        }
+
+        return stats
+
+    def get_active_users_query(self):
+        return User.query.filter_by(active=True)
```
buchi, line 24: We should also have a metric with the total active users of the current org unit.

opengever/contentstats/active_user_stats.py
```diff
@@ -0,0 +1,48 @@
...
+from opengever.ogds.models.group import groups_users
+from opengever.ogds.models.user import User
+
+
+class UserStatsCalculator(object):
+    """Returns statistics for the active users on the system
+    """
+    def get_stats(self):
+        stats = {
+            'total_active_users': self.calc_total_active_unique_users(),
```
buchi, line 14: As all numbers are totals, I would strip the `total_` prefix from the key.

opengever/contentstats/active_user_stats.py
elioschmutz: Of the current org unit only? or of all users assigned to any of the available active org units? I think the latter is besser.

elioschmutz: @buchi what do you think? see my last comment.

elioschmutz: => I think only for the current org unit is not possible because we wouldn't be abe to tell the unique amount of users for one specific customer. A user of one org-unit could also be in the other. But this user should be counted only once.

opengever/contentstats/active_user_stats.py
```diff
@@ -0,0 +1,76 @@
...
+
+class UserStatsCalculator(object):
+    """Returns statistics for the active users on the system
+    """
+    def get_stats(self):
+        stats = {
+            'active_users': self.calc_active_unique_users(),
+            'active_users_logged_in_last_30_days': self.calc_users_logged_in_last_x_days(30),
+            'active_users_logged_in_last_365_days': self.calc_users_logged_in_last_x_days(365),
+            'active_users_logged_in_once_at_least_once': self.calc_users_logged_in_at_least_once(),
```
buchi, line 19:
```suggestion
            'active_users_logged_in_at_least_once': self.calc_users_logged_in_at_least_once(),
```

opengever/contentstats/active_user_stats.py
```diff
@@ -0,0 +1,76 @@
...
+            'active_users_never_logged_in': self.calc_users_never_logged_in(),
+            'active_users_in_any_org_unit': self.calc_active_unique_users_in_any_org_unit(),
+            'active_users_in_any_org_unit_of_current_admin_unit':
```
buchi, line 22:
```suggestion
            'active_users_of_current_admin_unit':
```

opengever/contentstats/active_user_stats.py
```diff
@@ -0,0 +1,76 @@
...
+    """Returns statistics for the active users on the system
+    """
+    def get_stats(self):
+        stats = {
+            'active_users': self.calc_active_unique_users(),
+            'active_users_logged_in_last_30_days': self.calc_users_logged_in_last_x_days(30),
+            'active_users_logged_in_last_365_days': self.calc_users_logged_in_last_x_days(365),
+            'active_users_logged_in_once_at_least_once': self.calc_users_logged_in_at_least_once(),
+            'active_users_never_logged_in': self.calc_users_never_logged_in(),
+            'active_users_in_any_org_unit': self.calc_active_unique_users_in_any_org_unit(),
```
buchi, line 21:
```suggestion
            'active_users_of_all_admin_units': self.calc_active_unique_users_in_any_org_unit(),
```

buchi (approved):

PR #8021, opened 2024-08-23
Use TaskResponses if removing a task comment

elioschmutz:
We have two response implementations. A `Response` and a `TaskResponse`. The `TaskResponse` is used for tasks only and provides some additional attributes.

When removing a task comment, we currently use a default `Response` object instead a `TaskResponse`. We have several code parts where we check for attributes which are only available on the `TaskResponse`. For example: https://github.com/4teamwork/opengever.core/blob/2fa9dbf08e4ff5eb621186788c398c59382a4c29/opengever/task/browser/close.py#L311 . This will not work for `Reponses` and will raise an error.

This PR fixes this issue by using the `TaskResponse` for deleted comments as we do it already for every other response on tasks.

Providing an upgrade step seems not to be necessary since it is a rare edge-case. The last committed response of an object needs to be a removed comment on a multi-admin unit task.

For [TI-992]

## Checklist

- [x] Changelog entry

[TI-992]: https://4teamwork.atlassian.net/browse/TI-992?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/TI-992.bugfix
+++ b/changes/TI-992.bugfix
@@ -1,0 +1,1 @@
+Fix an issue where changing remote task state was not possible if the remote task have had a deleted comment. [elioschmutz]
--- a/opengever/api/configure.zcml
+++ b/opengever/api/configure.zcml
@@ -994,6 +994,14 @@
       />
 
   <plone:service
+      method="DELETE"
+      name="@responses"
+      for="opengever.task.task.ITask"
+      factory=".task.TaskResponseDelete"
+      permission="cmf.ModifyPortalContent"
+      />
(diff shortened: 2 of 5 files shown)


lukasgraf (approved):
LGTM! 👍 
I agree that we can do without an upgradestep - it should be very rare, and if it happens, there's even a workaround for the user, from what I understand: Just add a regular response to the remote task.

PR #8024, opened 2024-08-27
Fix copying documents when deployed with Docker

buchi:
Load monkey patches as early as possible

Because of not using `z3c.autoinclude` with Docker, the order of loading ZCML is different. It turns out that applying the patch for `handleDynamicTypeCopiedEvent` after registering the event handler doesn't work and breaks copying documents.

For [TI-1009](https://4teamwork.atlassian.net/browse/TI-1009)

## Checklist

- [x] Changelog entry

[TI-1009]: https://4teamwork.atlassian.net/browse/TI-1009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/TI-1009.bugfix
+++ b/changes/TI-1009.bugfix
@@ -1,0 +1,1 @@
+Fix copying documents when deployed with Docker. [buchi]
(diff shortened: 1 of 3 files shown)


phgross (approved):
👍

PR #8025, opened 2024-08-27
Disable colorization if env variable has an empty value

buchi:
Allows us to disable colorization without removing the env variable which is more convenient for Docker Compose deployments.

## Checklist

- [x] Changelog entry


lukasgraf (approved):
LGTM! 👍

PR #8027, opened 2024-08-28
Resolve invitation error when user dose not exists any more

Abdu-moustafa:
This pull request addresses a recurring AttributeError encountered by users in Teamraum when handling invitations. The error was traced to an issue where a non-existent inviter was being referenced, leading to a failure in the invitation process.

For [TI-1036](https://4teamwork.atlassian.net/browse/TI-1036)

**Before:**

before

**After:**
after

## Checklist

- [x] Changelog entry

[TI-1036]: https://4teamwork.atlassian.net/browse/TI-1036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/TI-1036.other
+++ b/changes/TI-1036.other
@@ -1,0 +1,1 @@
+Resolve invitation error when user dose not exists any more. [amo]
--- a/opengever/api/tests/test_participation.py
+++ b/opengever/api/tests/test_participation.py
@@ -1561,6 +1561,40 @@ class TestMyInvitationsGet(IntegrationTestCase):
                 }
             ], response.get('items'))
 
+    @browsing
+    def test_inviter_resolution_with_non_existing_user(self, browser):
+        self.login(self.workspace_owner, browser)
+
+        with freeze(datetime(2018, 4, 30, 10, 30)):
+            iid = getUtility(IInvitationStorage).add_invitation(
+                self.workspace,
+                self.regular_user.getProperty('email'),
+                "Does not exist",
+                'WorkspaceGuest')
+
+        self.login(self.regular_user, browser)
+
+        response = browser.open(
+            self.portal,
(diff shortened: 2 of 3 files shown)


phgross (changes requested):
This is not the right way to fix this issues. 

The problem should be fixed by just using the Actor.lookup() instead of the `PloneUserActor` in the my-invitations view
.

Abdu-moustafa:
> This is not the right way to fix this issues.
> 
> The problem should be fixed by just using the Actor.lookup() instead of the `PloneUserActor` in the my-invitations view .

Move the logic to the `my-invitations`

opengever/workspace/tests/test_my_invitations.py
```diff
@@ -301,3 +302,24 @@ def test_get_group_dn_from_registry(self):
...
+                'inviter': 'nonexistent_user',
+                'recipient_email': 'test@example.com',
+                'roles': 'WorkspaceGuest',
+                'comment': u"Test invitation"
+            }
+
+            # Add the invitation to the storage
+            self.storage._write_invitations[invitation_id] = entry
+
+            # Resolve the inviter from the invitation
```
phgross, line 320: This test does not actually check whether the problem still exists or not. Instead of manually fetching the inviter and create the actor from it, you should call the my-invitations view with the test user.

phgross (changes requested):

opengever/workspace/tests/test_my_invitations.py
Abdu-moustafa: ✅

lukasgraf (approved):
LGTM 👍

PR #8030, opened 2024-08-29
Disable Z2 access log in Docker deployments

buchi:
We log requests with ftw.structlog which provides all the information from the Z2 log and more.

## Checklist

- [x] Changelog entry


lukasgraf (approved):
LGTM 👍

PR #8034, opened 2024-08-30
Sign workflow

elioschmutz:
This PR extends the document workflow with all the necessary states and transitions to be able to implement the sign process.

It does not yet implement the transition extenders and the transitions will only be available if the sign-feature is available.

The transition extenders will be implemented in an upcoming PR. This PR is meant for a pre-review for the naming, permissions and available states and transitions.

Bildschirmfoto 2024-08-30 um 09 51 08

ℹ️ the document-workflow is simplified in this diagram. Trash and shadow states and transitions are not visible.

For [TI-659]

[TI-659]: https://4teamwork.atlassian.net/browse/TI-659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/base/locales/en/LC_MESSAGES/plone.po
```diff
@@ -233,13 +233,34 @@ msgstr "Finalize"
 msgid "document-transition-remove"
 msgstr "Remove"
 
+msgid "document-state-signing"
+msgstr "Will be signed"
```
lukasgraf, line 237: -> `Being signed`

opengever/base/locales/fr/LC_MESSAGES/plone.po
```diff
@@ -171,6 +171,12 @@ msgstr "En suspend"
 msgid "document-state-removed"
 msgstr "Effacé"
 
+msgid "document-state-signing"
+msgstr "Sera signé"
```
lukasgraf, line 175: `En cours de signature`?

elioschmutz:
> LGTM 👍, but
> 
> No transition guards yet for moving the document into the `signing` state? No tests yet? I think tests like these could be useful already:
> 
> In the `signing` state, it should be impossible to
> 
> * Delete the document
> * Copy or move the document
> * Checkout the document
> * Revert the document to a previous version
> * Probably some other things I can't think of right now ;)

As discussed. I pushed the PR very quickly to get a early feedback about the naming of the states, transitions, permissions and the workflow in general. I pushed all the missing parts in the meantime.

So it should be ready for a final review now.

elioschmutz:
@lukasgraf could you take a final look?

lukasgraf (approved):
LGTM! 👍

PR #8037, opened 2024-09-06
Add translation for classification

pareiss:
Add document property document property 'ogg.document.classification' for storing the classification information documents can have in gever for use in office applications.

To create a document go to a dossier and click the green '+' button, then 'document from template'. Give a name and selecte the template 'Allgemeine Word-Vorlage mit Metadaten'. Click 'create' and then download the file. Copy it from linux to windows and open it in word. Click on 'Datei' -> 'Informationen' -> 'Eigenschaften' -> 'Erweiterte Eigenschaften' -> 'anpassen'. Then look for 'ogg.document.classification'. To see a change in the Document Properties you must first check out the file in GEVER and then download it.

Um ein neues Feld einzufügen. auf 'Einfügen' klicken und dann 'Schnellbaustein' unter 'Text'. Danach beim Feld 'Kategorie' 'Dokumentinformationen' auswählen, dann 'DocProperty' und hier bei 'Feldeigenschaften' die gewünschte DocProperty.

For [TI-892]

## Checklist

- [x] Changelog entry

[TI-892]: https://4teamwork.atlassian.net/browse/TI-892?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/dossier/docprops.py
```diff
@@ -1,6 +1,7 @@
 from datetime import date
```
lukasgraf, line 1: This commit message is a bit strange. `multilingual` has a very particular meaning in Plone, and it refers to something else, not just regular translations.
I would just use something like this as the commit message:
`Make sure classification DocProperty values are translated.`

opengever/workspace/locales/de/LC_MESSAGES/opengever.workspace.po
```diff
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
-"POT-Creation-Date: 2024-08-21 07:37+0000\n"
+"POT-Creation-Date: 2024-09-06 06:10+0000\n"
```
lukasgraf, line 4: This commit is unnecessary. ✂️ 

First, there are no significant changes to the `.po` files. Only timestamps were updated, which are irrelevant. And those timestamps always get updated whenever you run the script. So unless there are actual changes, don't commit those `.po` files which only have their timestamps changed.

The second thing is, there *shouldn't* actually be any any changes to the translations! In your change you only make use of existing translations and make sure they actually take effect. So you shouldn't expect any translation content to change. Which means, in that case, it's not even necessary to run the `bin/i18n-build` script or touch the `.po` files in any other way. You're using making sure _existing_ translations are used.

lukasgraf (changes requested):
Just two minor points, otherwise LGTM 👍

opengever/workspace/locales/de/LC_MESSAGES/opengever.workspace.po
pareiss: Yes, here a commit from another branch snuck in that was actually changing things in th .po files. I will remove the commit from this PR.

PR #8042, opened 2024-09-16
Makes reports available for teamraum

XenoY:
Currently there is no overview of all permissions for a user or a group in Teamraum. This already exists in GEVER. We need to allow access for teamraum to the corresponding sources. We splitted the role mapping for the titles inside the report for GEVER and Teamraum.

_Needs changes from gever-ui: https://github.com/4teamwork/gever-ui/pull/2803_

For [TI-115](https://4teamwork.atlassian.net/browse/TI-115)

## Checklist

- [x] Changelog entry
- New translations

[TI-115]: https://4teamwork.atlassian.net/browse/TI-115?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/ogds/base/sources.py
```diff
@@ -1016,6 +1016,7 @@ class AllFilteredGroupsSourcePrefixed(FilterMixin, AllGroupsSource):
     multi-query-source.
     """
     GROUP_PREFIX = 'group:'
+    gever_only = False
```
lukasgraf, line 1019: I don't think these should be necessary any more after Elio's recent changes in https://github.com/4teamwork/opengever.core/pull/8045

(Those changes BTW are also the reason that there's merge conflicts. I would suggest that you maybe have a quick look with Elio again about this)

opengever/workspace/__init__.py
```diff
@@ -19,6 +19,7 @@
     'opengever.workspace.todo',
     'opengever.workspace.todolist',
     'opengever.workspace.workspace',
+    'all_users_and_groups'
```
lukasgraf, line 22: This doesn't really belong here, since this is a list of _types_ that are whitelisted, not vocabularies/sources. Which also means that it probably wasn't needed in the first place, because it didn't have any effect 😀

lukasgraf (changes requested):
LGTM, mostly! 👍 - but please double check with Elio, since he has just recently made made changes to the involved vocabularies/sources.

opengever/workspace/__init__.py
XenoY: If i remove it, as it was before, i get an 404 on the `@globalsources/all_users_and_groups` request. Which is needed to select the user or group to make the report for. Maybe there is another place i can put it, but i think it is needed. 

`WHITELISTED_TEAMRAUM_PORTAL_TYPES` is referenced in https://github.com/4teamwork/opengever.core/blob/9a7e34fc9e434ca8baa5fa17df23c8ffa36c8331/opengever/api/schema/globalsources.py#L62-L65

image
image

lukasgraf: I see, you're right - it indeed acts as a list of sources. Sorry for the detour then, go ahead and keep it in 😉 👍

elioschmutz:
👍

elioschmutz (approved):

PR #7986, opened 2024-06-28
Implement group and teams watchers

elioschmutz:
Currently, only users are allowed as watchers for resources.

This PR extends the `@possible-watchers` and `@watchers` endpoint to allow groups and teams as watchers beside of users.

For [TI-407]

## Checklist

_Everything has to be done/checked. Checked but not present means the author deemed it unnecessary._

- [x] Changelog entry
- [x] Link to issue (Jira or GitHub) and backlink in issue (Jira)
- API change:
  - [x] Documentation is updated
  - [x] API Changelog entry (see [guide](https://4teamwork.atlassian.net/wiki/spaces/4TEAM/pages/451248812/API+Changelog+Guidelines))
  - If breaking:
    - [x] api-change label added
    - [ ] #delivery channel notified about breaking change
    - [ ] Scrum master is informed

[TI-407]: https://4teamwork.atlassian.net/browse/TI-407?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #8045, opened 2024-09-18
Fix possible watchers in teamraum

elioschmutz:
The PR #7986 extended the possible watchers from users only to watching users, groups and teams. This works fine for gever but not for teamraum.

The group source is not whitelisted in teamraum and teams do not exist in teamraum.

The PR fixes the issue by:
- Whitelist and protect the groups source for teamraum deployments
- only use the users and groups source for teamraum deployments and add the teams-source only for gever deployments.

For [TI-1221]

## Checklist

- [x] Changelog entry

[TI-1221]: https://4teamwork.atlassian.net/browse/TI-1221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/ogds/base/sources.py
+++ b/opengever/ogds/base/sources.py
@@ -960,6 +960,7 @@ class FilterMixin(object):
 class AllGroupsSource(BaseSQLModelSource):
 
     model_class = Group
+    gever_only = False
 
     def __init__(self, context, include_inactive_groups=False, **kwargs):
         super(AllGroupsSource, self).__init__(context, **kwargs)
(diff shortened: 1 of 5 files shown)


opengever/ogds/base/sources.py
```diff
@@ -968,13 +969,24 @@ def __init__(self, context, include_inactive_groups=False, **kwargs):
...
     @property
     def base_query(self):
         return self._base_query
 
     @property
     def search_query(self):
-        return self.base_query.order_by(
+        query = self.base_query
+        if not visible_users_and_groups_filter.can_access_all_principals():
+            query = query.filter(Group.groupid.in_(
```
lukasgraf, line 986: I am a bit worried that this might produce some massive `IN [...]` clauses in the resulting SQL statements. These can be bad for performance, or outright fail because the SQL statement gets too long. But from what I can tell:
1) This clause will only ever be added for Teamraum deployments
2) The clause should only contain users and groups with access to the same workspaces as the logged in user. So usually a smaller set than *all* users and groups in OGDS.

-> So, should be fine I guess?

elioschmutz: Yes, exactly. Only for tr-deployments and for non `WorkspaceCreator` user which are mainly external users.

We have this restriction already for the `AllUsersSource`: https://github.com/4teamwork/opengever.core/blob/661f5f0b0f39dea1a19daec10bffeab965187e28/opengever/ogds/base/sources.py#L583

It should be save in term of there will be probably no situation where we have a large list of users/groups.

PR #8050, opened 2024-09-20
Serialize pending signing job for documents

elioschmutz:
This PR makes the pending signing state available for the frontend.

The current implementation has some issues when working with different versions and trying to make it visible to the UI. If we revise a signed document or revert to an old version, we'd loose the information about the signed documents and signers.

Currently, we provide `sign_metadata` which is in fact the `pending_signing_job` state. It gives us information about a pending signing process like the job-id, redirect url and the signers. this could also contain the current state of signers, i.e. who has aborted or not yet signed the document.

But there should be another state, the `signatures`-state itself. This should give us a finalized info about which version is signed by which users. So as soon as the signing-process is finished, the `pending_signing_job` should be removed and all the relevant information should be stored in a list which can then be used when serializing versions or the context itself. 

This allows us to really mark signed versions in the UI, remember reverted signed versions and to separate a pending signing job which can change over time from the effectively signed persistent state.

So this PR introduces the following changes:
- moves all the pending signing job related data to its own persistent classes
- clears the pending signing job state after completing or aborting the signing process
- exposes the pending signing job state for serialized documents
- provides a fake sign service client to make e2e testing possible

I also renamed the annotations key. Since we do not have the sign-feature in production right now we can safely do that without an upgradestep.

For [TI-661]

## Checklist

- [x] Changelog entry

### DIFF
--- a/opengever/sign/utils.py
+++ b/opengever/sign/utils.py
@@ -1,6 +1,15 @@
+from opengever.ogds.models.user import User
 from opengever.sign.interfaces import ISignSettings
 from plone import api
+from sqlalchemy import func
 
 
 def is_sign_feature_enabled():
     return api.portal.get_registry_record('is_feature_enabled', interface=ISignSettings)
+
+
+def email_to_userid(email):
+    if not email:
+        return ''
+    user = User.query.filter(func.lower(User.email) == email.lower()).first()
+    return user.userid if user else ''
--- a/changes/TI-661.other
+++ b/changes/TI-661.other
@@ -1,0 +1,1 @@
+Document serialization provides pending_signing_job for a running sign process. [elioschmutz]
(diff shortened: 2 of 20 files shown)


opengever/sign/utils.py
```diff
@@ -1,6 +1,28 @@
...
 from opengever.sign.interfaces import ISignSettings
 from plone import api
+from sqlalchemy import or_
 
 
 def is_sign_feature_enabled():
     return api.portal.get_registry_record('is_feature_enabled', interface=ISignSettings)
+
+
+def emails_to_userids(emails):
```
lukasgraf, line 11: As discussed, simplify, make case insensitive ([example](https://github.com/4teamwork/opengever.core/blob/master/opengever/ogds/auth/plugin.py#L201))

lukasgraf (approved):
LGTM 👍

elioschmutz:
@lukasgraf could you take a final look? you already reviewed it once. But I think we discussed in the meantime some implementation details which I updated here in this PR. But it's long time ago...

lukasgraf (approved):
LGTM 👍

PR #8052, opened 2024-09-25
Change docs publishing server

combine-space:
Decommission Centos 7 Server seth.4teamwork.ch and use web06p.4teamwork.ch for future publishing of onegovgever docs.


lukasgraf (approved):
LGTM 👍

PR #8053, opened 2024-09-27
Implement and serialize signatures

elioschmutz:
This PR is a follow-up PR of #8050 

Currently, we only track pending signing jobs on a document. As soon as the document sign process is completed, we do not have any information about the signatories or which versions has been signed.

This PR adds a storage for `signature items` and provides it through the api.

A new signature item will be created and persisted as soon as the signing process will be completed (all the requested signatories have been signed the document). 

In a pending signing job, we're just working with email addresses and will provide actor-ids on the fly when serializing it. This is what we want because the underlying actor could still change. As soon as the signature job is completed, we need to make sure, that the underlying actor does no longer change. We do this by effectively store the actor-ids of internal users instead of their email. But we still keep the email addresses for external users. This guarantees that its not possible to implicitly change the signatory by assigning the email address to another user in the future.

For [TI-661]

## Checklist

- [x] Changelog entry

### DIFF
--- a/opengever/sign/signatory.py
+++ b/opengever/sign/signatory.py
@@ -1,0 +1,20 @@
+from persistent import Persistent
+from persistent.list import PersistentList
+from plone.restapi.serializer.converters import json_compatible
+
+
+class Signatories(PersistentList):
+    def serialize(self):
+        return json_compatible([signatory.serialize() for signatory in self])
+
+
+class Signatory(Persistent):
+    def __init__(self, userid='', email=''):
+        self.userid = userid
+        self.email = email
+
+    def serialize(self):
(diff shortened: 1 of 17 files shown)


opengever/sign/signatory.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class Signatories(PersistentList):
+    def serialize(self):
+        return json_compatible([signatory.serialize() for signatory in self])
+
+
+class Signatory(Persistent):
+    def __init__(self,
+                 userid='',
```
lukasgraf, line 13: formatting

lukasgraf (approved):

PR #8055, opened 2024-09-27
Allow downloading checked-out documents with no modifications

buchi:
Until now we prevented downloading documents with no versions that are checked out by another user. This doesn't make sense as this is only the case when the document hasn't been modified yet. As soon as the document is modified, the initial version is created and used for downloads.

For [TI-675](https://4teamwork.atlassian.net/browse/TI-675)

## Checklist

- [x] Changelog entry

[TI-675]: https://4teamwork.atlassian.net/browse/TI-675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/TI-675.bugfix
+++ b/changes/TI-675.bugfix
@@ -1,0 +1,1 @@
+Allow downloading checked-out documents with no modifications and only an initial version. [buchi]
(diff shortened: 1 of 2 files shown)


lukasgraf (approved):
LGTM! 👍

PR #8056, opened 2024-10-01
Fix logging in upgrade command

buchi:
Set log level for root logger instead of "opengever.setup" logger only. This way we also get the logs from ftw.upgrade at INFO level.

## Checklist

- [x] Changelog entry


lukasgraf (approved):
LGTM! 👍

PR #8058, opened 2024-10-03
Cron time window

buchi:
Create cron jobs in Docker deployments within a time window.

Cron jobs are now started within a time window to avoid starting them at the same time for each deployment.
By default the deployment number is used to spread the cron jobs equaly in 12 slots. This can be customized by setting the `CRON_START_SLOT` and `CRON_MAX_START_SLOTS` environment variables explicitly.
The time window is defined by start and end time and is currently not configurable.

## Checklist

- [x] Changelog entry


lukasgraf (approved):
LGTM! 👍

PR #8067, opened 2024-10-17
Fix dossier transfer if the target admin unit does not provide the same dossier_types or document_types.

elioschmutz:
This PR fixes an issue where it is not possible to transfer a dossier if the target admin unit does not provide the same dossier or document types.

We fix the issue by removing the not existing type before deserializing it. Means, we loose this information on the target admin unit. This is the desired behavior.

For [TI-1345]

## Checklist

- [x] Changelog entry

[TI-1345]: https://4teamwork.atlassian.net/browse/TI-1345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/dossiertransfer/tests/test_api_perform.py
+++ b/opengever/dossiertransfer/tests/test_api_perform.py
@@ -6,12 +6,15 @@ from ftw.testing import freeze
 from opengever.api.kub import OrganizationContactSyncer
 from opengever.api.kub import PersonContactSyncer
 from opengever.base.model import create_session
+from opengever.document.behaviors.metadata import IDocumentMetadata
+from opengever.dossier.behaviors.dossier import IDossier
 from opengever.dossier.behaviors.participation import IParticipationAware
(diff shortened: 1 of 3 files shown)


opengever/dossiertransfer/tests/test_api_perform.py
```diff
@@ -448,6 +451,214 @@ def test_creates_dossiers_and_documents_with_new_contacts(self, mocker, browser)
         )
         self.assertTrue(journal_entry_exists)
 
+    @requests_mock.Mocker()
+    @browsing
+    def test_overtake_dossier_type_if_it_exists_on_the_target(self, mocker, browser):
```
lukasgraf, line 456: Nitpick: `overtake` means "überholen (auf der Autobahn)" :wink:
Maybe `apply`?

lukasgraf (approved):
LGTM! 👍

PR #7958, opened 2024-05-23
Solr conf: use WordDelimiterGraph only on indexing

phgross:
This helps to speed up queries of alphanummeric terms, for example "B57.0.346" or similar.

For [TI-293]

## Checklist
- [x] Changelog entry
- [x] Link to issue (Jira or GitHub) and backlink in issue (Jira)

[TI-293]: https://4teamwork.atlassian.net/browse/TI-293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #8069, opened 2024-10-18
Fix Solr tests

buchi:
Required due to https://github.com/4teamwork/opengever.core/pull/7958
Didn't get noticed because no new ogsolr image was built, which is used in tests.

The regular `@solrsearch` endpoint is mainly affected by the changes, but it's no longer used in the UI, so we just remove the assertions as fixing it would reintroduce the performance problems. Currently we are only using the `@solrlivesearch` endpoint.

## Checklist

- [ ] Changelog entry


lukasgraf (approved):
LGTM 👍

PR #8075, opened 2024-11-03
Close template file after processing with Sablon

buchi:
Leaving it open can lead to: `ValueError: Can't commit with opened blobs.`

## Checklist

- [x] Changelog entry


lukasgraf (approved):
LGTM 👍

PR #8077, opened 2024-11-03
Fix broken JS registry for new installs

buchi:
With #8036 the order of the resources was changed so that the cooked javascript was broken.

## Checklist

- [x] Changelog entry


lukasgraf (approved):
LGTM! 👍 Thanks!

PR #8078, opened 2024-11-04
Fix isolation issue with tests using a mocked solr.

elioschmutz:
The `mock_solr`-method of `opengever/testing/integration_test_case.py` is not isolated at all. If used once, the solr connection is mocked for every upcoming test.

We have another solr-mock implementation within the disposition builder introduced in https://github.com/4teamwork/opengever.core/pull/7939 (https://github.com/4teamwork/opengever.core/pull/7939/commits/b38f4d82779b33fe19e68f5ba39af55294651ca3) which checks for an already running solr and mocks the response if there is no solr connection yet. 

Until now, there was no test using the `mock_solr` method before the disposition activity-tests which relies on these mocked results: https://github.com/4teamwork/opengever.core/blob/master/opengever/disposition/tests/test_activities.py#L36

Under some circumstances, the execution order changes and when running the disposition-tests, the solr was already mocked and have not returned the right results.

This PR fixes this issue by using a context-manager for the `mock_solr` method to properly isolate it between tests.

I run into this issue through the linked story below. But it is unrelated to the story and could have been happen after every other test change.

For [TI-661]

## Checklist

### DIFF
--- a/opengever/api/tests/test_listing.py
+++ b/opengever/api/tests/test_listing.py
@@ -59,11 +59,6 @@ class TestListingEndpointWithSolr(IntegrationTestCase):
 
     features = ('bumblebee', 'solr')
 
-    def setUp(self):
-        super(TestListingEndpointWithSolr, self).setUp()
-
-        self.solr = self.mock_solr(response_json={})
-
     def test_filesize_accessor_avoids_obj_lookup(self):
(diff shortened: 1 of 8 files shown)


lukasgraf (approved):
Nice! Looks good to me! 👍

PR #8081, opened 2024-11-06
Fix username userid issues

buchi:
Fix various issues when creating content in a policy:

- Sync OGDS as early as possible on new site setup
- Map responsible name to id when creating dossiers with a policy
- Map responsible name to id when creating dossiers with dossier builder
- Map group names to ids when assigning global roles in deployment setup
- Install OGDS auth plugin when there's no LDAP profile

## Checklist

- [x] Changelog entry

### DIFF
--- a/changes/GH-8081.other
+++ b/changes/GH-8081.other
@@ -1,0 +1,1 @@
+Map responsible name to id when creating content with a policy. [buchi]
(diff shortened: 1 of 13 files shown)


lukasgraf (approved):
LGTM! 👍 

With the most recent changes, this was sufficient to get my local setup working.

(The problem with the group memberships not being resolved properly was caused by the presence of the LDAP Plugin: When its `Mutable properties` capability is enabled, this interferes with the group lookups of the OGDS auth plugin)

PR #8086, opened 2024-11-14
Introduce a lookup table for objects having modified local roles.

elioschmutz:
This PR introduces a local roles tracking table which can be used to do ad-hoc reports about modified local-roles for users or objects.

The table stores all managed roles of each principal for each object having managed custom local roles.

For [TI-1540]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration

[TI-1540]: https://4teamwork.atlassian.net/browse/TI-1540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/TI-1540.other
+++ b/changes/TI-1540.other
@@ -1,0 +1,1 @@
+Introduce a lookup table for objects having modified local roles. [elioschmutz]
--- a/opengever/base/role_assignments.py
+++ b/opengever/base/role_assignments.py
@@ -1,5 +1,6 @@
 from opengever.base import _
 from opengever.base.oguid import Oguid
+from opengever.sharing.local_roles_lookup.manager import LocalRolesLookupManager
 from persistent.list import PersistentList
 from persistent.mapping import PersistentMapping
 from zope.annotation.interfaces import IAnnotations
@@ -476,6 +477,8 @@ class RoleAssignmentManager(object):
         for principal in owner_principals:
             self.context.manage_addLocalRoles(
                 principal, ['Owner'], verified=True)
+
+        LocalRolesLookupManager().update_lookup_table(self.context)
 
         if reindex:
             self.context.reindexObjectSecurity()
--- a/opengever/base/tests/test_database_table_declarations.py
+++ b/opengever/base/tests/test_database_table_declarations.py
@@ -6,6 +6,7 @@ from opengever.dossiertransfer.model import tables as dossier_transfer_tables
 from opengever.globalindex.model import tables as globalindex_tables
 from opengever.locking.model import tables as lock_tables
 from opengever.meeting.model import tables as meeting_tables
+from opengever.sharing.local_roles_lookup.model import tables as local_roles_lookup_tables
 from unittest import TestCase
 
 
@@ -29,5 +30,6 @@ class TestDatabaseTableDeclarations(TestCase):
             + dossier_transfer_tables
             + base_tables
             + system_messages_tables
+            + local_roles_lookup_tables
         )
         self.assertItemsEqual(expected_tables, Base.metadata.tables.keys())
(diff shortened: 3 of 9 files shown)


opengever/sharing/local_roles_modifications/model.py
```diff
@@ -0,0 +1,30 @@
...
+from sqlalchemy import String
+from sqlalchemy.schema import Sequence
+
+
+tables = [
+    'local_roles_modifications',
+]
+
+
+class LocalRolesModification(Base):
```
buchi, line 18: Weird name. Why modification? I would just call it `LocalRoles`.
Or if you want to be more specific: `LocalRolesLookup` or `LocalRolesCache`.
With modification in the name I would expect that the table contains an entry for earch modification.
E.g. if you add a role for a user on an object and then later add an additional role for the same user and object, the table should contain two entries. But that's not what we want.

opengever/sharing/local_roles_modifications/manager.py
```diff
@@ -0,0 +1,141 @@
...
+        return False
+
+    def reindex_modifications(self, context):
+        """Checks the local roles of the given context and adds or removes
+        principals to the lookup table if necessary.
+        """
+        if context.portal_type not in self.MANAGED_PORTAL_TYPES:
+            return
+
+        lookup_table_principal_ids = self.get_distinct_principal_ids_by_object_uid(context.UID())
```
buchi, line 80: Wouldn't it be much easier to just delete all entries for the given object and then add the current entries?

opengever/sharing/local_roles_modifications/manager.py
```diff
@@ -0,0 +1,141 @@
...
+        return self.model_cls.query.filter_by(
+            admin_unit_id=self.local_unit_id,
+            principal_id=principal_id)
+
+    def get_distinct_object_uids_by_principal(self, principal_id):
+        query = self._get_by_principal_query(principal_id)
+        query = query.with_entities(self.model_cls.object_uid).distinct()
+        return [item[0] for item in query]
+
+    def _get_by_principals_query(self, principal_ids):
```
buchi, line 123: For me it's not clear why this and other methods are marked as private/internal and others are not. In general I would not use `_` prefixes here as I'm not expecting that this class will be extendend.

opengever/sharing/local_roles_modifications/model.py
elioschmutz: 👍 thanks for the suggestion. I'll use `LocalRolesLookup`.

opengever/sharing/local_roles_modifications/manager.py
buchi: A composite primary key should be fine here. As long as you don't need a foreign key in another table and the values of the PK columns never change it's ok to use composite PKs.

opengever/sharing/local_roles_modifications/manager.py
```diff
@@ -0,0 +1,141 @@
...
+        local_roles = {}
+        for principal, roles in context.get_local_roles():
+            managed_roles = list(set(roles).intersection(self.MANAGED_ROLES))
+            if not managed_roles:
+                continue
+            local_roles[principal] = managed_roles
+        return local_roles
+
+    def get_entry_by_principal_and_object_uid(self, principal_id, object_uid):
+        return self.model_cls.query.filter_by(
```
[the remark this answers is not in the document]
elioschmutz: I use composite primary keys now.

elioschmutz:
@lukasgraf @buchi I updated the PR with your suggestions. The main difference is the use of composite primary keys and the easier update implementation.

Could you take a final look?

PR #8088, opened 2024-11-19
Implement RoleAssignmentReporter

elioschmutz:
This PR implements the `RoleAssignmentReporter` which is able to return a dict of objects having changed local roles.

The reporter creates a dict providing a list of objects, ordered by their paths, that define custom local roles.

Each item in the list contains:
- a serialized Solr item representing the object.
- a dictionary mapping each defined local role to a list of principals assigned that role.

The report can be filtered by a specific `principal_id`, with an option to include or exclude assigned groups for the given principal. Additionally, the report can be limited to a specific branch.

For [TI-1541]

## Checklist

- [x] Changelog entry

[TI-1541]: https://4teamwork.atlassian.net/browse/TI-1541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/sharing/local_roles_lookup/reporter.py
+++ b/opengever/sharing/local_roles_lookup/reporter.py
@@ -1,0 +1,193 @@
+from collections import defaultdict
+from ftw.solr.interfaces import ISolrSearch
+from ftw.solr.query import make_filters
+from opengever.base.solr import batched_solr_results
+from opengever.base.solr import OGSolrDocument
+from opengever.ogds.models.user import User
+from opengever.sharing.local_roles_lookup.manager import LocalRolesLookupManager
+from plone.app.uuid.utils import uuidToObject
+from zope.component import getUtility
+
+
+class RoleAssignmentReporter(object):
+    """Responsible for generating role assignment report data.
+    """
+
+    field_list = [
+        'UID',
+        'path',
+        'Title',
+        'Description',
+        'review_state',
+        'portal_type',
+        'reference',
+        'is_leafnode',
+    ]
+
+    sort_order = 'path asc'
+
+    def __init__(self):
+        self.solr_service = SolrQueryService()
+        self.local_roles_manager = LocalRolesLookupManager()
+
+    def __call__(self, *args, **kwargs):
+        return self.report_for(*args, **kwargs)
+
+    def report_for(self,
+                   principal_id=None,
+                   include_memberships=False,
+                   root=None,
+                   start=0,
+                   rows=25):
+        """Generates a report of objects with custom local role assignments.
+
+        The report contains a dictionary with:
+        - `items`: A list of objects, each containing:
+            - A serialized Solr document representing the object.
+            - A dictionary mapping local roles to the list of principals assigned
+                to each role.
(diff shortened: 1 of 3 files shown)


opengever/sharing/local_roles_lookup/reporter.py
```diff
@@ -0,0 +1,195 @@
...
+
+        principal_ids = [principal_id]
+        if include_memberships:
+            principal_ids.extend(self.resolve_user_memberships(principal_id))
+
+        return principal_ids
+
+    def resolve_user_memberships(self, principal_id):
+        ogds_user = User.query.filter_by(userid=principal_id).one_or_none()
+        return [group.groupid for group in ogds_user.groups] if ogds_user else []
```
lukasgraf, line 120: I think this should be limited to active groups.

opengever/sharing/local_roles_lookup/reporter.py
```diff
@@ -0,0 +1,195 @@
...
+            "role1": ["principal_1", "principal_2]
+            "role2": ["principal_1"]
+          }
+        }
+        """
+        uids = {}
+        for entry in self.local_roles_manager.get_entries(
+                uids_filter=uids_filter,
+                principal_ids_filter=principal_ids_filter):
+            uids.setdefault(entry.object_uid, {})
```
lukasgraf, line 142: This seems like a rather complicated use of `dict.setdefault`. If you want to use `setdefault`, you should assign its return value, that's the whole point ;-)

So:
```python
local_roles = uids.setdefault(entry.object_uid, {})

for role in entry.roles:
    principals = local_roles.setdefault(role, [])
    principals.append(entry.principal_id)
```

But I think what you actually want is to use defaultdicts in this case:
```python
>>> from collections import defaultdict
>>> import json
>>>
>>> uids = defaultdict(lambda: defaultdict(list))
>>> uids[12345]['reader'].append('peter')
>>> uids[12345]['reader'].append('hans')
>>> uids[12345]['writer'].append('admin')
>>> uids[77777]['writer'].append('admin')
>>>
>>> print(json.dumps(uids))
{
    "12345":
        {
            "reader": ["peter", "hans"],
            "writer": ["admin"]
        },
    "77777": 
        {
            "writer": ["admin"]
        }
}
```

But this will create instances of `defaultdict`s, not `dict`s. So depending on how you use that data structure down the line, that might not be desirable.

opengever/sharing/local_roles_lookup/reporter.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def search(self, filters, fields, sort, start, rows):
+        response = self.solr.search(filters=filters,
+                                    fl=fields,
+                                    sort=sort,
+                                    start=start,
+                                    rows=rows)
+        return response.docs, response.num_found
+
+    def fetch_all(self, filters, fields, sort):
```
lukasgraf, line 191: `fetch_all` isn't used at all?

lukasgraf (approved):
LGTM 👍 - just some comments.

PR #8089, opened 2024-11-19
Send 'checked out by' value for FavoritesGET

Acrop146:
Favorite view did not show checked out status for documents. This was corrected by adding 'checked_out' attribute to response for FavoriteGet class for @favorites endpoint.

Before:
1448_before

After (checked out by current user):
1448after2

After (checked out by different user):
1448_after

For [TI-1448]

## Checklist

- [x] Changelog entry


elioschmutz (approved):
👍 

just to note: this won't work for favorites on remote admin units, but that's ok. we don't have a solution to fix this yet.

Could you reflect this in the changelog and write a note in the jira story, that we're aware of this.

PR #8090, opened 2024-11-21
Add policy template for Compose deployments

buchi:
Extends the policy generator to allow creating policies for Docker Compose deployments.

Instructions for creating new deployments:
https://4teamwork.atlassian.net/wiki/spaces/4TEAM/pages/2045607937/Neues+Gever+Teamraum+Deployment+mit+Docker+Compose

The Compose files are split into multiple files:
- `compose-deployment.yaml` (Services required for each Deployment)
- `compose-geverui.yaml` (Frontend)
- `compose-ianus.yaml` (Portal)
- `compose-<server>-<num>-<name>.yaml` (Specific Deployment)

This allows easier customization without modifying the base yaml files.
Configuration settings are stored in `compose-<server>-<num>-<name>.env` and customizations should be made preferably in `compose-<server>-<num>-<name>.yaml`. 

## Checklist

- [x] Changelog entry


elioschmutz (approved):
👍 

Do we still need the old-style deployment creation?

buchi:
Old-style may be needed for existing policies when a new admin unit is added. As soon as all policies are converted to Compose, we can remove the old template.

PR #8091, opened 2024-11-22
Role assignment excel exporter

elioschmutz:
This PR implements the excel-export for the newly introduced ad-hoc role assignment reports.

The exporter now exports a row for each role, principal, object-uid combination which gives the user the possibility to do further filtering, sorting, pivoting in excel.

The exporter can be used with the same filters as the ad-hoc role assignment report itself. So it's possible to only export a specific branch or principal.

Important to note is the performance of the export. We expect < 5000 items (currently, the larges site has ~3000) having customized local roles. The exporter should easily handle this amount of items.

## Old export
Bildschirmfoto 2024-11-25 um 14 38 26

## New export
Bildschirmfoto 2024-11-25 um 14 36 31

For [TI-1546]

## Checklist

- [x] Changelog entry

[TI-1546]: https://4teamwork.atlassian.net/browse/TI-1546?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/sharing/local_roles_lookup/reporter.py
```diff
@@ -81,6 +83,56 @@ def report_for(self,
 
         return {'total_items': total_items, 'items': items}
 
+    def excel_report_for(self,
+                         principal_id=None,
+                         include_memberships=False,
+                         root=None):
+        """Generates a report of objects with custom local role assignments optimized
+        to export in excel. An item for each
```
lukasgraf, line 91: ✂️  `An item for each`

opengever/sharing/local_roles_lookup/reporter.py
```diff
@@ -81,6 +83,56 @@ def report_for(self,
...
+                        }
+                }
+            ]
+        """
+        principal_ids = self.expand_principal_ids(principal_id,
+                                                  include_memberships)
+        uids = self.get_distinct_uids(principal_ids)
+
+        if not uids:
+            yield iter([])
```
lukasgraf, line 113: This should be just `return`.

`yield iter([])` will make the generate return exactly 1 element in this case (a `list_iterator` object over an empty list), instead of behaving like a generator with zero elements.

opengever/sharing/local_roles_lookup/reporter.py
```diff
@@ -148,6 +200,41 @@ def role_assignments_by_uid(self, uids_filter=None, principal_ids_filter=None):
...
+                entry_with_principal_details2,
+            ]
+        }
+        """
+        uids = {}
+        for entry in self.get_entries_with_principal_details(
+                uids_filter=uids_filter,
+                principal_ids_filter=principal_ids_filter):
+
+            uids.setdefault(entry[0].object_uid, []).append(entry)
```
lukasgraf, line 223: Again, using a `defaultdict(list)` here would be more idiomatic.

opengever/sharing/local_roles_lookup/reporter.py
```diff
@@ -148,6 +200,41 @@ def role_assignments_by_uid(self, uids_filter=None, principal_ids_filter=None):
 
         return uids
 
+    def explode_local_role_entry_by_role(self, entry):
```
lukasgraf, line 203: `explode_local_role_entry_by_role` is not used anywhere?

lukasgraf (approved):
LGTM 👍 - just some comments-

PR #8094, opened 2024-11-26
Allow multiple principals when creating role assignment reports

elioschmutz:
This is a follow-up PR of #8088 

It extends the `RoleAssignmentReporter` to allow creating reports of multiple principals instead only of one principal.

In addition, the PR fixes an issue in testing where I used `assertItemsEqual` instead `assertEqual`. Using `assertItemsEqual` for nested dicts will not properly deep assert the values. It will lead to false positives.

For [TI-1541]

## Checklist

- [x] Changelog entry

[TI-1541]: https://4teamwork.atlassian.net/browse/TI-1541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/sharing/local_roles_lookup/reporter.py
+++ b/opengever/sharing/local_roles_lookup/reporter.py
@@ -34,7 +34,7 @@ class RoleAssignmentReporter(object):
         return self.report_for(*args, **kwargs)
 
     def report_for(self,
-                   principal_id=None,
+                   principal_ids=[],
                    include_memberships=False,
                    root=None,
                    start=0,
@@ -62,7 +62,7 @@ class RoleAssignmentReporter(object):
                 "total_items": 1
             }
         """
-        principal_ids = self.expand_principal_ids(principal_id,
+        principal_ids = self.expand_principal_ids(principal_ids,
                                                   include_memberships)
         uids = self.get_distinct_uids(principal_ids)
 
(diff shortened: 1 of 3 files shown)


opengever/sharing/local_roles_lookup/reporter.py
```diff
@@ -34,7 +34,7 @@ def __call__(self, *args, **kwargs):
         return self.report_for(*args, **kwargs)
 
     def report_for(self,
-                   principal_id=None,
+                   principal_ids=list(),
```
lukasgraf, line 37: Why `list()` and not just `[]`?

If you intended this to avoid the mutable keyword arguments issue, it doesn't work like that. `principal_ids=list()` would have exactly the same issue as `principal_ids=[]`.

If you need to avoid getting the same reference to a mutable keyword argument across multiple invocations of a function, you always need to use a pattern like this:

```python
def func(values=None):
    if valus is None:
        values = []
```

However, if you _don't_ intend to mutate your function arguments (which is preferrable anyway), it's best to use tuples, since they are already immutable:
```python
def func(values=()):
    ...
```

lukasgraf (approved):
LGTM 👍 - just a comment about mutable keyword args.

PR #8103, opened 2024-12-17
Restrict team watchers source to the current org unit

elioschmutz:
We introduced group and teams watchers in a previous PR: #7986  which allows the user to select a group or a team as a watcher. Currently, all teams within the OGDS are queried which is not the desired behavior. Each team is assigned to a specific Org-Unit. The watchers-source should respect this setting to only list teams of the current users org unit.

So this PR updates the sources to only list those teams instead of all teams across all admin units.

For [TI-1735]

## Checklist

- [x] Changelog entry

[TI-1735]: https://4teamwork.atlassian.net/browse/TI-1735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/TI-1735.other
+++ b/changes/TI-1735.other
@@ -1,0 +1,1 @@
+Restrict team watchers source to the current org-unit. [elioschmutz]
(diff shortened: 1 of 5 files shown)


buchi (approved):
👍🏻 LGTM

PR #8106, opened 2025-01-07
Set is_active filter to true for possible participants

Acrop146:
Inactive users in kub shouldn't be able to be found in the "possible participants" search in Gever Dossiers. 

To correct this behaviour, the "only_active" in the KubClient was set to True and the behavior of "is_active" in KuB was corrected in the following PR: https://github.com/4teamwork/kub/pull/356

Before: (Pascal Reiss is an inactive user in KuB):

1737_before

After:

1737_after

For [TI-1737]

## Checklist

- [x] Changelog entry

[TI-1737]: https://4teamwork.atlassian.net/browse/TI-1737?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/TI-1737.other
+++ b/changes/TI-1737.other
@@ -1,0 +1,1 @@
+Restrict possible participants search from finding inactive users in kub. [ran]
--- a/opengever/dossier/participations.py
+++ b/opengever/dossier/participations.py
@@ -158,7 +158,7 @@ class KuBParticipationHandler(PloneParticipationHandler):
     def __init__(self, context):
         self.context = context
         self.annotations = IAnnotations(self.context)
-        self.participant_source = KuBContactsSourceBinder()(self.context)
+        self.participant_source = KuBContactsSourceBinder(only_active=True)(self.context)
 
 
 class IParticipationData(Interface):


opengever/api/dossier_participations.py
```diff
@@ -232,6 +233,8 @@ def reply(self):
         query = safe_unicode(self.request.form.get('query', ''))
 
         handler = IParticipationAware(self.context)
+        if isinstance(handler.participant_source, KuBContactsSource):
+            handler.participant_source.kub_filters = handler.participant_source.get_kub_filters(only_active=True)
```
elioschmutz, line 237: With these changes, you deeply couple the this Service to the internal structure of the `handler`-adapter. You should try to solve it closer to the object where it belongs to. 

So in this case, it'd currently just set the attribute when creating the `participant_source`: https://github.com/4teamwork/opengever.core/blob/master/opengever/dossier/participations.py#L161

elioschmutz (commented):
Please take a look at my comment.

And could you validate that editing the participants is still possible even if there are participants which are inactive in the meantime:

- add participant A, participant B
- set participant B as inactive
- edit the participants in gever and add participant C
- remove participant A
- and last, try to also remove participant B

This should still be possible.

opengever/api/dossier_participations.py
Acrop146: Done

Acrop146:
Process:
add participant A, participant B
set participant B as inactive
edit the participants in gever and add participant C
remove participant A
and last, try to also remove participant B

-> Done, worked well.

elioschmutz (approved):
👍 thanks!

PR #8107, opened 2025-01-09
Implement auto-open of next task after closing the previous one

Abdu-moustafa:
**This PR:**

If we have a standard process for tasks, the current situation is that once a task is marked as erledigt, the next task will open automatically. This behavior is not ideal. With our new implementation, the next task will only open if the previous task is closed.

For [TI-1331](https://4teamwork.atlassian.net/browse/TI-1331)

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[TI-1331]: https://4teamwork.atlassian.net/browse/TI-1331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/api/tests/test_complete_successor_task.py
+++ b/opengever/api/tests/test_complete_successor_task.py
@@ -239,7 +239,7 @@ class TestCompleteSuccessorTaskPost(IntegrationTestCase):
     def test_pass_documents_to_next_task_if_selected(self, browser):
         self.login(self.regular_user, browser)
 
-        self.seq_subtask_1.task_type = 'correction'
+        self.seq_subtask_1.task_type = 'direct-execution'
         self.seq_subtask_1.sync()
 
         successor = accept_task_with_successor(
@@ -252,13 +252,14 @@ class TestCompleteSuccessorTaskPost(IntegrationTestCase):
 
         # Need to trick the transition controller into thinking its a remote
         # request so that it allows the resolve transition on the predecessor
-        with patch('opengever.task.browser.transitioncontroller.'
-                   'RequestChecker.is_remote',
-                   new_callable=PropertyMock) as mock_is_remote:
+        with patch(
+            'opengever.task.browser.transitioncontroller.'
+            'RequestChecker.is_remote',
+                new_callable=PropertyMock) as mock_is_remote:
             mock_is_remote.return_value = True
 
             request_body = json.dumps({
-                'transition': 'task-transition-in-progress-resolved',
+                'transition': 'task-transition-in-progress-tested-and-closed',
                 'text': 'I finished this task.',
                 'pass_documents_to_next_task': True,
             })
@@ -269,9 +270,10 @@ class TestCompleteSuccessorTaskPost(IntegrationTestCase):
                 headers=self.api_headers)
 
         # Predecessor and successor should have been closed
-        self.assertEqual('task-state-resolved',
+        self.assertEqual('task-state-tested-and-closed',
                          api.content.get_state(self.seq_subtask_1))
-        self.assertEqual('task-state-resolved', api.content.get_state(successor))
+        self.assertEqual('task-state-tested-and-closed',
+                         api.content.get_state(successor))
 
         # relatedItems are passed to sequential successor
         self.assertEqual(1, len(self.seq_subtask_2.relatedItems))
--- a/opengever/task/handlers.py
+++ b/opengever/task/handlers.py
@@ -101,8 +101,8 @@ def review_state_changed(task, event):
     # on the parent task. We need to ensure, that every parent task is in progress.
     if event.action not in ['task-transition-open-planned',
                             'task-transition-planned-open',
-                            'task-transition-rejected-skipped',
-                            'task-transition-reassign']:
+                            'task-transition-reassign',
+                            'task-transition-rejected-skipped']:
         task.maybe_start_parent_task()
 
     if event.action in FINAL_TRANSITIONS:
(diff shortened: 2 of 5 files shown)


opengever/task/tests/test_task_from_template.py
```diff
@@ -296,7 +269,7 @@ def test_record_activity_when_open_next_task(self):
         self.assertEquals(
             'task-state-resolved', api.content.get_state(self.subtask))
         self.assertEquals(
-            'task-state-open', api.content.get_state(subtask2))
+            'task-state-planned', api.content.get_state(subtask2))
```
elioschmutz, line 272: Read the test-function name to get an idea of what this test is doing: it should record an activity when the next task gets opened. But with your changes, the next task will not open with the `task-transition-in-progress-resolved` transition. I'ts still in the `planned`-state. So there should be no activity yet and the test should fail.

Please, check why the activity is recorded and adjust the test to reflect the new behavior:
- resolve subtask
- assert next task in planned state
- assert no new activity have been recorded
- close subtask
- assert next task in open state
- assert new activity have been recorded

opengever/task/tests/test_task_from_template.py
```diff
@@ -267,7 +240,7 @@ def test_handles_missing_permissions_on_next_task(self):
         self.assertEquals(
             'task-state-resolved', api.content.get_state(subtask1))
         self.assertEquals(
-            'task-state-open', api.content.get_state(subtask2))
+            'task-state-planned', api.content.get_state(subtask2))
```
elioschmutz, line 243: This test checks if the current user has enough permission to open the next task in a task sequence without permission.

So you should close the `subtask1` to have the same test as before.

opengever/task/handlers.py
```diff
@@ -101,7 +101,6 @@ def review_state_changed(task, event):
     # on the parent task. We need to ensure, that every parent task is in progress.
     if event.action not in ['task-transition-open-planned',
                             'task-transition-planned-open',
-                            'task-transition-rejected-skipped',
```
elioschmutz, line 104: Why do you need to remove this transition from the list?

opengever/api/tests/test_complete_successor_task.py
```diff
@@ -235,48 +235,6 @@ def test_approves_document_if_given(self, browser):
         self.assertEqual(1, len(approvals))
         self.assertEqual(IUUID(successor), approvals[0].task_uid)
 
-    @browsing
-    def test_pass_documents_to_next_task_if_selected(self, browser):
```
elioschmutz, line 239: Please, do not remove this test. It's purpose is to test multi admin unit task syncing for documents. You should rewrite the test to get the same purpose, adjusted to your new implementation

elioschmutz (changes requested):

opengever/task/handlers.py
Abdu-moustafa: @elioschmutz  If the previous task is skipped, the next task will open automatically, which we no longer want. Skipping the previous task will not cause any changes to the next task until the previous one is closed

opengever/task/tests/test_task_from_template.py
Abdu-moustafa: ✔️

opengever/task/tests/test_task_from_template.py
Abdu-moustafa: ✔️

opengever/api/tests/test_complete_successor_task.py
Abdu-moustafa: ✔️

opengever/api/tests/test_complete_successor_task.py
```diff
@@ -253,12 +253,12 @@ def test_pass_documents_to_next_task_if_selected(self, browser):
         # Need to trick the transition controller into thinking its a remote
         # request so that it allows the resolve transition on the predecessor
         with patch('opengever.task.browser.transitioncontroller.'
-                   'RequestChecker.is_remote',
-                   new_callable=PropertyMock) as mock_is_remote:
+                    'RequestChecker.is_remote',
+                    new_callable=PropertyMock) as mock_is_remote:
```
elioschmutz, line 257: ✂️  nothing changed here, please correct the indentation

opengever/api/tests/test_complete_successor_task.py
```diff
@@ -269,14 +269,16 @@ def test_pass_documents_to_next_task_if_selected(self, browser):
...
                          api.content.get_state(self.seq_subtask_1))
-        self.assertEqual('task-state-resolved', api.content.get_state(successor))
+        self.assertEqual('task-state-tested-and-closed',
+                         api.content.get_state(successor))
 
         # relatedItems are passed to sequential successor
         self.assertEqual(1, len(self.seq_subtask_2.relatedItems))
         self.assertEqual(self.document, self.seq_subtask_2.relatedItems[0].to_object)
 
+
```
elioschmutz, line 281: ✂️  remove the new line

opengever/task/handlers.py
elioschmutz: But this line is used to start the parent task if we work on a subtask, not to check if the next task should open or not.

I don't think you should remove this line. If it's required, please add it in a separate commit and show the purpose of this change with a test.

elioschmutz (approved):
👍 

please update the PR according to my feedback before merging it.

opengever/api/tests/test_complete_successor_task.py
Abdu-moustafa: ✔️

opengever/api/tests/test_complete_successor_task.py
Abdu-moustafa: ✔️

opengever/task/handlers.py
Abdu-moustafa: ✔️ added back

PR #8116, opened 2025-01-27
Use self-hosted action runner

buchi:
Improve building of Docker image:

- Use our self-hosted action runner which should be faster
- Use newer versions of Docker actions
- Allow running the action manually
- Build image for `master` branch with tag `edge` which can be used for deployment on dev systems.
- Only build ARM image for releases (tags) to have faster builds for `master` branch.
- Only build `lastest` tag if it's on the `master` branch. Should avoid overwriting `latest` with backport releases.

## Checklist

- [ ] Changelog entry

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -5,6 +5,8 @@ on:
   push:
     tags:
       - '*.*.*'
+    branches:
+      - master
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


elioschmutz (approved):
👍 

How much faster is it with the self hosted runner? Can you provide some stats?

buchi:
@elioschmutz A lot! You can see the stats at https://github.com/4teamwork/opengever.core/actions

PR #8117, opened 2025-01-27
Change the task template text field to RichtextField

Abdu-moustafa:
This PR:
Change `task template` text field from `Text` to RichText Field.  

For [TI-1333](https://4teamwork.atlassian.net/browse/TI-1333)

related to the FE PR: https://github.com/4teamwork/gever-ui/pull/2844

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[TI-1333]: https://4teamwork.atlassian.net/browse/TI-1333?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/api/tests/test_process.py
+++ b/opengever/api/tests/test_process.py
@@ -114,7 +114,7 @@ class TestProcessPost(IntegrationTestCase):
             "start_immediately": False,
             "process": {
                 "title": "New employee",
-                "text": "A new employee arrives.",
+                "text": "<p> A new employee arrives.</p>",
                 "sequence_type": "sequential",
                 "items": [
                     {
@@ -138,6 +138,7 @@ class TestProcessPost(IntegrationTestCase):
         self.assertEqual(1, len(children['added']))
         main_task = children['added'].pop()
 
+        self.assertEqual("<p> A new employee arrives.</p>", main_task.text.output)
         self.assertEqual(browser.json['@id'], main_task.absolute_url())
         self.assertEqual(u'New employee', main_task.title)
         self.assertEqual(self.regular_user.getId(), main_task.issuer)
--- a/opengever/core/upgrades/20250127123059_update_task_template_text_field_to_rich_text/upgrade.py
+++ b/opengever/core/upgrades/20250127123059_update_task_template_text_field_to_rich_text/upgrade.py
@@ -1,0 +1,23 @@
+from ftw.upgrade import UpgradeStep
+from plone.app.textfield import IRichTextValue
+from plone.app.textfield.value import RichTextValue
+
+
+class UpdateTaskTemplateTextFieldToRichText(UpgradeStep):
+    """Update task template text field to rich text.
+    """
+
+    def __call__(self):
+        query = {'object_provides': "opengever.tasktemplates.content.tasktemplate.ITaskTemplate"}
+        msg = "Change task template text field from text field to rich text field"
+
+        for task in self.objects(query, msg):
+            if IRichTextValue.providedBy(task.text):
+                continue
+
+            task.text = RichTextValue(
+                raw=task.text or "",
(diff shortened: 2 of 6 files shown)


elioschmutz (commented):
👍 looks good so far.

For testing, please add at least a test in: `opengever.api.tests.test_tasktemplate.py` where you test adding a tasktemplate with text and serializing it later on. You can refer to the existing tests for other content types as examples."

This is only the first part of tasktemplates. Please, take a look at the `@process` endpoint which is responsible for creating the effective task objects. This endpoints also provides a lot of tests in `TestProcessPost`. Please, extend these tests with rich text values to make sure it still works with the new implementation.

opengever/core/upgrades/20250127123059_update_task_template_text_field_to_rich_text/upgrade.py
```diff
@@ -0,0 +1,23 @@
...
+from plone.app.textfield.value import RichTextValue
+
+
+class UpdateTaskTemplateTextFieldToRichText(UpgradeStep):
+    """Update task template text field to rich text.
+    """
+
+    def __call__(self):
+        query = {'object_provides': "opengever.tasktemplates.content.tasktemplate.ITaskTemplate"}
+        msg = "Change task template text field from Text field to Richtext field"
```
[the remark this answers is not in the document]
Abdu-moustafa: ✔️

changes/TI-1333.other
```diff
@@ -0,0 +1 @@
+Change task template `text field` from Text field to RichText field. [amo]
```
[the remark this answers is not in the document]
Abdu-moustafa: ✔️

opengever/api/tests/test_process.py
```diff
@@ -666,7 +689,7 @@ def test_handles_teams_and_inbox_as_task_responsible(self, browser):
                         "issuer": self.regular_user.id,
                         "deadline": "2022-03-01",
                         "task_type": "direct-execution",
-                        "is_private": False,
+                        "is_private": False
```
elioschmutz, line 692: not related to this PR or commit. Please undo this change

opengever/api/tests/test_process.py
```diff
@@ -688,6 +711,10 @@ def test_handles_teams_and_inbox_as_task_responsible(self, browser):
 
         self.assertEqual(1, len(children['added']))
         main_task = children['added'].pop()
+
+        # Text (description) rendered as a richtext
+        self.assertTrue(IRichTextValue.providedBy(main_task.text))
+
```
elioschmutz, line 717: You don't need to add this assertion to every test. Read the test function name to get an idea of what should be tested in the specified function and then decide if you really need to do an assertion for the text or not. 

For example this test: `test_handles_teams_and_inbox_as_task_responsible` has nothing to do with the text and should not assert it. It's the same for  all other testcases instead of `test_create_simple_process` which generally tests the attributes created by the process -endopint.

So please only add the assertion to the `test_create_simple_process` method and effectively call the endpoint with a html value and then assert for that value

elioschmutz (commented):
👍 

one small thing

opengever/api/tests/test_process.py
Abdu-moustafa: ✔️

opengever/api/tests/test_process.py
Abdu-moustafa: ✔️

elioschmutz (approved):

PR #8122, opened 2025-01-30
Fix encoding issues in classic UI

buchi:
Since upgrading the Docker base image to Alpine 3.20 (ogcore 2024.15.0 and later) the HTML parser used in plone.app.theming doesn't detect the correct encoding when parsing a response. As we always use UTF-8, we set the encoding manually.

Not sure what causes the error, it seems only to happen with newer Alpine versions.

The behavior can be reproduced with the following snippet, which should return 'UTF-8', but returns 'ISO-8859-1':

```python
>>> from lxml.html import HTMLParser
>>> parser = HTMLParser()
>>> parser.feed(b'<div><h2>Schlagw\xc3\xb6rter</h2></div>')
>>> parser.close().getroottree().docinfo.encoding
'ISO-8859-1'
```

## Checklist

- [x] Changelog entry

### DIFF
--- a/docker/core/patches/p_a_theming_transform.py
+++ b/docker/core/patches/p_a_theming_transform.py
@@ -149,7 +149,7 @@ class ThemeTransform(object):
             return None
 
         try:
-            return getHTMLSerializer(result, pretty_print=False)
+            return getHTMLSerializer(result, pretty_print=False, encoding='utf-8')
         except (AttributeError, TypeError, etree.ParseError):
             return None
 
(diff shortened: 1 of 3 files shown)


docker/core/patches/p_a_theming_transform.py
```diff
@@ -0,0 +1,216 @@
+import logging
```
elioschmutz, line 1: is it possible to commit the file without changes and in a second commit the patch? This would make it easier to understand what you've changed.

elioschmutz (approved):

PR #8125, opened 2025-02-13
Remove persistent objects in sign module

elioschmutz:
This PR refactors the sign module to no longer use persistent objects when dealing with signature data. See https://github.com/4teamwork/opengever.core/pull/8120#discussion_r1950529343 for more information.

To store the values, we now use json strings and we deserialize it back to objects when loading it from the annotations.

This should make future development easier and less error prone. 

For [TI-1495]

## Checklist

[TI-1495]: https://4teamwork.atlassian.net/browse/TI-1495?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/sign/pending_editor.py
+++ b/opengever/sign/pending_editor.py
@@ -1,22 +1,54 @@
 from opengever.sign.utils import email_to_userid
-from persistent import Persistent
-from persistent.list import PersistentList
 from plone.restapi.serializer.converters import json_compatible
 
 
-class PendingEditors(PersistentList):
+class PendingEditors(list):
 
     @classmethod
     def from_emails(cls, emails):
         return cls([PendingEditor(email=email) for email in emails])
 
+    @classmethod
+    def from_json_object(cls, values):
+        if not values:
+            return cls()
+
+        return cls(
+            [PendingEditor.from_json_object(value) for value in values])
+
+    def to_json_object(self):
+        return [pending_editor.to_json_object() for pending_editor in self]
+
+    def __eq__(self, value):
+        return isinstance(value, PendingEditors) and value.to_json_object() == self.to_json_object()
+
+    def __ne__(self, value):
+        return not value == self
+
     def serialize(self):
         return json_compatible([pending_editor.serialize() for pending_editor in self])
 
 
-class PendingEditor(Persistent):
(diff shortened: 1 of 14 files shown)


opengever/sign/pending_editor.py
```diff
@@ -1,23 +1,55 @@
...
+        return not value == self
+
+    @classmethod
+    def from_json_object(cls, value):
+        return cls(
+            email=value.get('email')
+        )
+
+    def to_json_object(self):
+        return json_compatible({
```
buchi, line 49: Why is the `json_compatible` converter needed? dicts and strings are already json-serializable.

opengever/sign/pending_signature.py
```diff
@@ -27,6 +41,27 @@ def __init__(self,
...
+            email=value.get('email'),
+            status=value.get('status'),
+            signed_at=value.get('signed_at')
+        )
+
+    def to_json_object(self):
+        return json_compatible({
+            'email': self.email,
+            'status': self.status,
+            'signed_at': self.signed_at,
```
buchi, line 62: Not sure about this conversion. If `signed_at` can be a `datetime`, then there's also a reverse conversion needed in the `from_json_object()`method. If it's a string anyway, then why are you calling the `json_compatible` converter? It will not work if you do the conversion only in the `to_json_object()` method, you also have to do the reverse operation in the `from_json_object()` method.

opengever/sign/pending_editor.py
elioschmutz: To be safe... I thought it makes it less error prone if we once extend this object. If something is not compatible, it will raise here and we'll see it directly. 

Would you just remove it?

opengever/sign/pending_signature.py
elioschmutz: These are just strings. So same as above... I used it for safety reasons...

elioschmutz:
@buchi I removed the `json_compatible`. Could you take another look? Thanks!

buchi (approved):
👍🏻 LGTM

PR #8130, opened 2025-02-25
Change text field in task schema to RichField

Acrop146:
Due to making the "Beschreibung" field in tasks into a RichText field, the backend schema for tasks must be changed as well. Before the change, the delegate action threw a validation error. 

Delegate_Error

For [TI-2108]

## Checklist

- [x] Changelog entry


changes/TI-2108.bugfix
```diff
@@ -0,0 +1 @@
+Change text in task schema from TextField to Richtext. [ran]
```
elioschmutz, line 1: In my optionon, this is a little too technical. I'd begin with: `Fix ...` and try to explane what i really fixed with this bugfix. But I see, it's not easy in this case

elioschmutz (approved):

PR #8133, opened 2025-02-26
Fix pdf-dossier-tasks endpoint

Abdu-moustafa:
After providing the RichText field for the task, we encountered an issue with the `pdf-dossier-tasks` endpoint returning a `500 `error when accessing the description directly via `task.text`, which will no longer work with the rich text. We should be able to get the text (description) via `task.text.output`.

Since the `task history` also uses the same field `RichText`, we have an issue where, if the user has, for example, tables or other HTML tags in the history, the PDF is not displaying as expected—some fields get wrapped or disappear. After some debugging with @buchi , we fixed this problem by stripping the HTML from the history and displaying it as plain text.

For [TI-2113](https://4teamwork.atlassian.net/browse/TI-2113)

**After:**

[dossier-17.pdf](https://github.com/user-attachments/files/18988360/dossier-17.pdf)

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
  - DB-Schema migration
- API change:
  - If breaking:
    - [ ] api-change label added
- Bug fixed:
- New functionality:
- New translations

[TI-2113]: https://4teamwork.atlassian.net/browse/TI-2113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/latex/dossiertasks.py
+++ b/opengever/latex/dossiertasks.py
@@ -79,9 +79,11 @@ class DossierTasksLaTeXView(MakoLaTeXView):
             if deadline:
                 deadline = deadline.strftime(self.strftimestring)
 
+            task_description = task.text.output if task.text else u''
+
             task_data_list.append({
                 'title': self.convert_plain(task.title),
-                'description': self.convert_plain(task.text or ""),
+                'description': self.convert(task_description),
                 'sequence_number': task.get_sequence_number(),
                 'type': self.convert_plain(task.get_task_type_label()),
                 'completion_date': completion_date,
--- a/changes/TI-2113.other
+++ b/changes/TI-2113.other
@@ -1,0 +1,1 @@
+Fix pdf-dossier-tasks endpoint after provide Richtext for text (description) field. [amo]
--- a/opengever/latex/listing.py
+++ b/opengever/latex/listing.py
@@ -17,6 +17,10 @@ from Products.Five.browser.pagetemplatefile import ViewPageTemplateFile
 from zope.component import adapter
 from zope.interface import implementer
 from zope.interface import Interface
+import re
+
+
+STRIP_HTML_RE = re.compile('<[^<]+?>')
 
 
 class Column(object):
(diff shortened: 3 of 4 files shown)


opengever/latex/dossiertasks.py
```diff
@@ -79,9 +79,17 @@ def get_render_arguments(self):
             if deadline:
                 deadline = deadline.strftime(self.strftimestring)
 
+            task_description = ""
+
+            if task.text:
```
buchi, line 84: Why is this needed? Shouldn't task.text always be a `RichtTextValue`?

opengever/latex/dossiertasks.py
Acrop146, line 1: Some of my tests which generated a pdf were failing for the same reason once I corrected the new task builder in https://github.com/4teamwork/opengever.core/pull/8130, so I implemented a similiar solution. Please check out my changes and rebase if needed.

opengever/latex/dossiertasks.py
Abdu-moustafa: Yes, it is always a `RichTextValue` if we have a value. However, if `task.text` is `None`, we encounter the issue that a None value does not have an `output` attribute. With this check, we no longer have this issue, as it replaces the None value with an empty string and avoids checking for output.

opengever/latex/dossiertasks.py
```diff
@@ -79,9 +79,17 @@ def get_render_arguments(self):
...
+                if isinstance(task.text, unicode):
+                    task_description = task.text
+                else:
+                    task_description = task.text.output
```
elioschmutz, line 88: The `task.text` should be `None` or `RichTextValue`.  So I think it's enough if you just do:

```python
task_description = task.text.output if task.text else u''
```

elioschmutz (approved):
Please, update the PR according to my feedback and coordinate with @Acrop146 the merge order.

opengever/latex/dossiertasks.py
Abdu-moustafa: we decide to merge this branch to master and rebase the other branch

PR #8134, opened 2025-02-28
Update UpdateMetadata schema to include richtext and fix relevant tas…

Acrop146:
Due to making the "Beschreibung" field in tasks into a RichText field, the backend schema for tasks must be changed as well. Before the change, the delegate action threw a validation error. 

Delegate_Error

For [TI-2108]

## Checklist

- [x] Changelog entry

[TI-2108]: https://4teamwork.atlassian.net/browse/TI-2108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/testing/builders/dx.py
```diff
@@ -238,7 +239,10 @@ def __init__(self, session):
         self.arguments = {
             'responsible_client': 'org-unit-1',
             'responsible': TEST_USER_ID,
-            'issuer': TEST_USER_ID}
+            'issuer': TEST_USER_ID,
+            'text': RichTextValue(raw=u"",
+                                  mimeType='text/html',
+                                  outputMimeType='text/x-html-safe')}
```
elioschmutz, line 245: Shouldn't it be `None`? We've seen that the default empty value is `None` when creating a new task.

elioschmutz (approved):

PR #8135, opened 2025-03-03
Sign service uses SIGN_SERVICE_GEVER_URL environment variable to construct the callback links.

elioschmutz:
Currently, the sign client implementation uses the current url to generate the callback urls of the sign service.

This only works if the sign service has access to this url. In case of local development or advanced production deployments, we need a static url which is available for the sign-service.

Currently, we have this issue in development. The sign service runs within a docker container but the gever backend is not. The `ogsign` service can't access gever by `localhost`. The URL should point to the host ip address:

``` 
SIGN_SERVICE_GEVER_URL http://172.16.0.1:8080/fd
```  

For [TI-1931]

## Checklist

- [x] Changelog entry

[TI-1931]: https://4teamwork.atlassian.net/browse/TI-1931?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/TI-1931-2.other
+++ b/changes/TI-1931-2.other
@@ -1,0 +1,1 @@
+Sign service uses SIGN_SERVICE_GEVER_URL environment variable to construct the callback links. [elioschmutz]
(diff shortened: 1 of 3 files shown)


buchi (approved):
👍🏻 LGTM

PR #8141, opened 2025-03-14
Dossier responsible as dossier manager

elioschmutz:
This pull request refactors the [PR #7789](https://github.com/4teamwork/opengever.core/pull/7789), which introduced the feature `grant_role_manager_to_responsible` that automatically assigned the `Role-Manager` role to the responsible  of a dossier. The new approach replaces the `Role-Manager` role with the more appropriate `Dossier-Manager` role, which better matches the responsible user's permissions. This change addresses several issues, including confusion between the roles and the inability of a `Role-Manager` to protect dossiers. The feature can be controlled via a feature flag  `grant_dossier_manager_to_responsible`. The old feature  `grant_role_manager_to_responsible` and all the related code of this feature has been removed.

**Background:**
This pull request is part of the work done under [CA-6155](https://4teamwork.atlassian.net/browse/CA-6155). The goal of that story was to improve how permissions are granted to the responsible person of a dossier. Initially, we implemented a feature that automatically assigned the `Role-Manager` role to the responsible person. This role allowed the responsible person to adjust local roles and manage permissions for the dossier. However, there were some issues with this approach:

`Role-Manager` vs. `Dossier-Manager`: The `Role-Manager` role was used to manage permissions, but it couldn’t protect dossiers. However, a `Dossier-Manager` role exists, which is specifically designed to manage dossiers, including protecting them. The confusion between these two roles was unclear to users and not very intuitive. A Role-Manager could change permissions but could not "protect" a dossier, which seemed inconsistent. 

Lockout Issue: A `Role-Manager` could accidentally lock themselves out of a dossier if localroles where missconfigured or it was protected and they were not assigned any of the protection roles (reader, reader and writer, manager). This created confusion and potential access issues.

Clarifying the Roles: The key problem here was that the `Role-Manager` role didn’t fully match what the responsible person of a dossier actually needed. The `Dossier-Manager` role makes more sense because it grants the permissions needed to manage and protect a dossier, including accessing the dossier. The "Role-Manager" role was often misunderstood, especially when combined with the "Protect Dossier" feature.

**New Approach:**
The decision was made to refactor the previous feature by replacing the `Role-Manager` role with the `Dossier-Manager` role for the responsible person. The `Dossier-Manager` role better reflects what the responsible person needs to do, and it eliminates the confusion between these two roles. Additionally, this new implementation avoids the problem of a `Role-Manager` accidentally locking themselves out.

**This PR also introduces the following changes:**

- The `Dossier-Manager` role cannot be changed once the feature is activated; it will always be assigned to the responsible person of the dossier.
- A new checkbox called "Extend permissions" was added in the "Protect Dossier" schema, allowing dossier managers to prevent inheritance from being interrupted when protecting a dossier.
- All unnecessary code related to the `Role-Manager` role has been removed. The `Role-Manager` role itself will not be removed

## Dossier responsible permissions before:
Screenshot_20250324_142116

## Dossier responsible permissions after:
Screenshot_20250325_090737

For [TI-2137]

## Checklist

- [x] Changelog entry

- Upgrade-Steps:
 - API change:
  - If breaking:
    - [x] api-change label added

[TI-2137]: https://4teamwork.atlassian.net/browse/TI-2137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


buchi (approved):
👍🏻 LGTM

For me the feature of extending local roles by protecting a dossier is a bit weird. what kind of protection should this be? and it's already possible to extend local roles by editing access.

elioschmutz:
Using the default local roles settings is not possible and not desired. The default local roles management is way too complicated for default users. We had to implement the "dossier protection"-feature to make local roles management easy to understand and manage for default users. 

Extending the current local roles in context of dossier protection seems weird, but its the already known feature for users to manage the local permissions of a dossier.

PR #8144, opened 2025-03-24
Upgrade Solr to version 9.x

buchi:
This upgrades Solr from version 8.11.4 to the currently latest available version 9.8.1.

Beside using the new Solr version, the main configuration files `solrconfig.xml` and `managed-schema.xml` (renamed from `managed-schema`) have been updated. 

`solrconfig.xml` was taken from the techproducts sample config and the following customizations were reapplied:
- Use `SearchableText` as default query field
- Set default query operator to AND
- Enforce eDisMax parser with boosting of specific fields
- Detect language and map text to language specific fields (SearchableText, Description, Title)
- Enable text extraction with Tika (Solr Cell)
- Add replication handler for making backups in tests
- Add Data Import Handler for reindexing from another Solr instance
- Add JavaScript engine and script update processor for `solr-sync.js` script used for RIS.

`managed-schema.xml` was updated to schema version 1.7. Field types for primitive fields were taken from tech products sample config.

The Solr upgrade does **not** require any changes in the integration code (`opengever.core`, `ftw.solr`). Only some test fixes have been necessary. This means that the Solr upgrade can be deployed independently of a specific `opengever.core` version. However a full reindex is required.

Upgrading Docker deployments:
- Bump `ogsolr` image to version 9.8.1
- Update deployment with `--reindex-solr` option.
- It's also possible to just upgrade Solr without an `ogcore` update with the `--skip-upgrades` option. 

Upgrading Buildout deployments:
- The Solr version will be automatically updated when deploying an `opengever.core` release containing this change. To avoid upgrading Solr, pin down the Solr download url in the policy of the deployment.
- Solr 9.x requires Java 11 or later. Ensure that the correct Java version is provided by the OS before trying to upgrade Solr!

Already deployed to: gever.4teamwork.ch, teamraum.4teamwork.ch

For [TI-2274](https://4teamwork.atlassian.net/browse/TI-2274)

## Checklist

- [x] Changelog entry

[TI-2274]: https://4teamwork.atlassian.net/browse/TI-2274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docker/solr/create_and_update_cores.sh
+++ b/docker/solr/create_and_update_cores.sh
@@ -4,13 +4,16 @@ set -e
 CONFIG_SOURCE=/opt/solr/server/solr/configsets/ogsite
 CORES_DIR=/var/solr/data
 
-for core in ${SOLR_CORES}; do
-	coredir="$CORES_DIR/$core"
-	if [ ! -d "$coredir" ]; then
-		mkdir -p "$coredir/conf"
-		echo "name=$core" >"$coredir/core.properties"
-		cp -r "$CONFIG_SOURCE/"* "$coredir/conf/"
-	else
-		cp -r "$CONFIG_SOURCE/"* "$coredir/conf/"
-	fi
-done
+# If ZK_HOST is defined, then assume SolrCloud mode
+if [ -z ${ZK_HOST} ]; then
+	for core in ${SOLR_CORES}; do
+		coredir="$CORES_DIR/$core"
+		if [ ! -d "$coredir" ]; then
+			mkdir -p "$coredir/conf"
+			echo "name=$core" >"$coredir/core.properties"
+			cp -r "$CONFIG_SOURCE/"* "$coredir/conf/"
+		else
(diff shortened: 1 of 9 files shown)


docker/solr/create_and_update_cores.sh
```diff
@@ -4,13 +4,16 @@ set -e
...
-	coredir="$CORES_DIR/$core"
-	if [ ! -d "$coredir" ]; then
-		mkdir -p "$coredir/conf"
-		echo "name=$core" >"$coredir/core.properties"
-		cp -r "$CONFIG_SOURCE/"* "$coredir/conf/"
-	else
-		cp -r "$CONFIG_SOURCE/"* "$coredir/conf/"
-	fi
-done
+# If ZK_HOST is defined, the assume SolrCloud mode
```
elioschmutz, line 7:
```suggestion
# If ZK_HOST is defined, then assume SolrCloud mode
````

PR #8146, opened 2025-04-02
Fix upgradestep for grant dossier manager to responsible if there is no responsible

elioschmutz:
This is a follow-up PR of #8141 and fixes an issue in the upgradestep where we try to check the permission of the current dossier responsible. Every dossier should have a responsible. But it seems, that we have dossier without any responsible, maybe creaded manually by code.

For [TI-2137]

## Checklist

[TI-2137]: https://4teamwork.atlassian.net/browse/TI-2137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Acrop146 (approved):
LGTM 👍

PR #8161, opened 2025-05-27
Enable attach attachments to submitted / scheduled ris proposal

Abdu-moustafa:
This PR:

We have a case on the RIS site where, when a user tries to submit document (nachreichen) for a proposal (Antrag), a permission error occurs. To resolve this, we grant the "Modify content" permission to the user when the proposal is in the `submitted` or `approved` state.

- Enables the users to attch attachments for ris submitted proposal.
- Enables the users to attch attachments for ris approved proposal.

For [TI-2329](https://4teamwork.atlassian.net/browse/TI-2329)
Related RIS PR: [RIS PR](https://github.com/4teamwork/ris/pull/3772)
## Checklist

- [x] Changelog entry

[TI-2329]: https://4teamwork.atlassian.net/browse/TI-2329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/ris/tests/test_proposal_transition_schedule.py
```diff
@@ -59,3 +60,41 @@ def test_proposal_schedule_debug_action_sets_wokflow_state(self, browser):
         statusmessages.assert_message('Item state changed.')
 
         self.assert_workflow_state('proposal-state-scheduled', self.ris_proposal)
+
+    @browsing
+    def test_can_submit_attachments_to_scheduled_proposal(self, browser):
+        self.login(self.manager, browser)
```
elioschmutz, line 66: don't test with the manager but with an editor instead.

opengever/ris/tests/test_proposal_transition_submit.py
```diff
@@ -74,3 +75,41 @@ def test_proposal_submit_debug_action_sets_wokflow_state(self, browser):
         statusmessages.assert_message('Item state changed.')
 
         self.assert_workflow_state('proposal-state-submitted', self.ris_proposal)
+
+    @browsing
+    def test_can_submit_attachments_to_submitted_proposal(self, browser):
+        self.login(self.manager, browser)
```
elioschmutz, line 81: Same here

elioschmutz (commented):
The changes for the workflow are missing. You only added the upgradestep with the `definition.xml` but you also need to adjust the workflow for the default profile:

https://github.com/4teamwork/opengever.core/blob/master/opengever/core/profiles/default/workflows/opengever_ris_proposal_workflow/definition.xml

opengever/ris/tests/test_proposal_transition_schedule.py
Abdu-moustafa: ✔️

PR #8174, opened 2025-06-30
Fix dossier transfer dossier protection

elioschmutz:
This PR fixes an issue where the dossier protection settings have not been properly updated after transfering a dossier from one responsible to another thorugh the `@transfer-dossier` endpoint when the feature `grant_dossier_manager_to_responsible` is activated.

The `grant_dossier_manager_to_responsible` is responsible for granting the dossier manager role to the current responsible user. To make this habben, the feature implements an eventhandler which listens to the `IObjectModifiedEvent`. To make the eventhandler performant enough, it will only update the protection settings, if the `responsible` field has been changed. Means, if we trigger the `IObjectModifiedEvent`, we need to make sure that we also pass the changed attributes.

For [TI-2752]

## Checklist

- [x] Changelog entry


buchi (approved):
👍🏻 LGTM

PR #8187, opened 2025-09-10
Only users with 'Transfer Assignment' can see transfer responsible ac…

Acrop146:
Fribourg, which grants the dossier responsible the dossier manager role and wants explicitly that every user sees the "transfer responsible action". All other users from other customers should only be able to see the action if they have the permission to enact it.

For [TI-1604]

## Checklist

- [x] Changelog entry

[TI-1604]: https://4teamwork.atlassian.net/browse/TI-1604?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/dossier/actions.py
+++ b/opengever/dossier/actions.py
@@ -2,6 +2,7 @@ from opengever.base.context_actions import BaseContextActions
 from opengever.base.interfaces import IOpengeverBaseLayer
 from opengever.base.listing_actions import BaseListingActions
 from opengever.docugate import is_docugate_feature_enabled
+from opengever.dossier import is_grant_dossier_manager_to_responsible_enabled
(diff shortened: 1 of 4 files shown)


opengever/dossier/actions.py
```diff
@@ -188,7 +189,16 @@ def is_zipexport_available(self):
         return True
 
     def is_transfer_dossier_responsible_available(self):
-        return api.user.has_permission('Modify portal content', obj=self.context)
+        """
+        Fribourg wants everyone to see the action and get an Error if they don't
```
buchi, line 193: Should be a comment, not a doc string.

PR #8195, opened 2025-10-01
Escape html for solr listing document snippets.

elioschmutz:
This PR properly removes dangerous html tags from the solr listing document snippets.

We can't just search and replace reserved html characters and convert them into html entities (i.e. replacing < to &gt;) because the `snippets` includes text highlighting which is made with html.

We use the plone internal `safe_html` portal transform to fix it. This will remove dangerous tags which is what we want here.

For [TI-3167]

## Checklist

- [x] Changelog entry

[TI-3167]: https://4teamwork.atlassian.net/browse/TI-3167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/TI-3167.other
+++ b/changes/TI-3167.other
@@ -1,0 +1,1 @@
+- XSS: Escape html for solr listing document snippets. [elioschmutz]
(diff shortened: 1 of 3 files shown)


buchi (approved):
👍🏻 LGTM

PR #8239, opened 2026-05-05
Fix overriding portal types provider for content stats

buchi:
So far the portal types provider wasn't registered correctly and we didn't get the main dossiers counts.
Worked before with buildout because of z3c.autoinclude "magic".

## Checklist

- [x] Changelog entry

## Post-merge action

> [!IMPORTANT]


Acrop146 (approved):
LGTM 👍

PR #8249, opened 2026-07-03
Docuteam disposition api

elioschmutz:
This PR implements a new ui action and endpoint to directly send an sip to a remote archive ingest server.

Currently we do only provide the docuteam ingest provider: https://docs.docuteam.ch/de/actions/ingest/ 

### Configuration

- ``DISPOSITION_SIP_TO_ARCHIVE_PROVIDER`` - Archive provider to use. Currently only ``docuteam``
  is supported. 
- ``DISPOSITION_DOCUTEAM_INGEST_URL`` - Base URL of the Docuteam ingest endpoint.
- ``DISPOSITION_DOCUTEAM_INGEST_API_KEY`` - API key (token) for authenticating with the
  Docuteam ingest endpoint.

### Permission

A user who can download an SIP is also allowed to send it to the archive.

### Journal

A new response-entry is created if an sip is downloaded or send to the archive.

For [TI-3710]

## Checklist

- [x] Changelog entry

## Post-merge action

> [!IMPORTANT]
> - [ ] After this PR is merged, Dev on Kubernetes must be updated.

[TI-3710]: https://4teamwork.atlassian.net/browse/TI-3710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/opengever/disposition/archive_clients.py
+++ b/opengever/disposition/archive_clients.py
@@ -1,0 +1,52 @@
+from os import environ
+import logging
+import requests
+
+
+logger = logging.getLogger('opengever.disposition.archive_clients')
+
+
+class NullClient(object):
+    """Client which does nothing and can be used for testing
+    """
+    def deliver(self, *args, **kwargs):
+        return None, '0'
+
+
+class DocuteamClient(object):
+    def __init__(self):
+        self.ingest_url = environ.get('DISPOSITION_DOCUTEAM_INGEST_URL')
+        if not self.ingest_url:
+            raise ValueError('Missing required environment variable: DISPOSITION_DOCUTEAM_INGEST_URL')
+
+        self.api_key = environ.get('DISPOSITION_DOCUTEAM_INGEST_API_KEY')
+        if not self.api_key:
+            raise ValueError('Missing required environment variable: DISPOSITION_DOCUTEAM_INGEST_API_KEY')
+
+    def deliver(self, sip_package, filename='package.zip'):
+        url = '{}?token={}&format=ech0160'.format(
+            self.ingest_url, self.api_key)
+        response = requests.post(
+            url,
+            files={'package': (filename, sip_package.data, 'application/zip')},
+        )
+        response.raise_for_status()
+        try:
+            submission_id = response.json()['response'][0]['id']
+        except (ValueError, KeyError, IndexError):
+            submission_id = u'<unavailable>'
+        logger.info('SIP delivered to docuteam archive. Submission ID: %s', submission_id)
+        return response, submission_id
+
+
+archive_client_registry = {
+    'null_client': NullClient,
+    'docuteam': DocuteamClient,
+}
+
+
+def get_archive_client():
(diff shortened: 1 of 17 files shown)


opengever/disposition/archive_clients.py
```diff
@@ -0,0 +1,53 @@
...
+class NullClient(object):
+    """Client which does nothing and can be used for testing
+    """
+    def deliver(self, *args, **kwargs):
+        return None, '0'
+
+
+class DocuteamClient(object):
+    def __init__(self):
+        self.ingest_url = environ.get('DOCUTEAM_INGEST_URL')
```
buchi: In general these settings could also be stored in the registry with the advantage of not requiring a restart when changing a setting.
At least I would prefix all disposition related envs with `DISPOSITION_` . => `DISPOSITION_DOCUTEAM_INGEST_URL`.

opengever/disposition/archive_clients.py
```diff
@@ -0,0 +1,53 @@
...
+    'docuteam': DocuteamClient,
+}
+
+
+def get_archive_client():
+    provider_name = environ.get('SIP_TO_ARCHIVE_PROVIDER', '').strip().lower()
+    if not provider_name:
+        raise ValueError(
+            'SIP_TO_ARCHIVE_PROVIDER environment variable is not set.')
+    return archive_client_registry[provider_name]()
```
buchi, line 52: Handle unknown provider_name?
Not sure if the end user sees the error details if an exception is raised. But this is not good practice security wise. The end user should get a message like 'currently not available' or 'misconfigured'.

buchi (commented):
Check if using envs instead of registry settings is really what you want.

opengever/disposition/archive_clients.py
elioschmutz: Personally, I would keep envs here. That's what we did with the Sign Service as well. It's simply more practical in a Docker/Kubernetes environment. We would need an additional policy upgrade step with the registry to set the values. I wanted to avoid that.

opengever/disposition/archive_clients.py
elioschmutz: I've updated the error message to a more generic one without internal details.

In general, this is a mandatory deployment configuration if the feature is in use. If it is missing or points to an unknown provider, the system is misconfigured and needs to be fixed by an administrator/developer. In my opinion, this function should fail loudly when the deployment configuration is invalid.

opengever/disposition/archive_clients.py
elioschmutz: As discussed, I'll keep env vars but prefix it.

opengever/disposition/archive_clients.py
buchi: I agree, but the channel to notify a developer or an admin is `logger.error()`, not the end user.

buchi (approved):
👍🏻 LGTM

PR #8251, opened 2026-07-13
Fix sharing group filter to match on group name instead of group id

elioschmutz:
When searching for groups in the sharing view, the black-list/white-list filter was checking the wrong value. It compared the filter pattern against the group's internal ID instead of its name.

The fix changes the filter to check the group's actual name instead of its ID. This value was already being loaded elsewhere in the code for display purposes

## Before
https://github.com/user-attachments/assets/27b0e310-0866-4c6a-838a-e361222e6384

## After
https://github.com/user-attachments/assets/25357a3f-a8e6-483a-80fd-92ca560fa94e

For [TI-3930]

## Checklist

- [x] Changelog entry

## Post-merge action

> [!IMPORTANT]

[TI-3930]: https://4teamwork.atlassian.net/browse/TI-3930?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/TI-3930.bugfix
+++ b/changes/TI-3930.bugfix
@@ -1,0 +1,1 @@
+Fix group white-/black-list prefix filter in sharing view to match on the group name instead of the group id  [elioschmutz]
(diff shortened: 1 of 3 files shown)


buchi (approved):
👍🏻 LGTM

PR #8254, opened 2026-08-28
Prevent finalizing a trashed document via the finalize transition

elioschmutz:
Document.is_finalize_allowed() previously checked only the checkout state and pending approval tasks, but did not account for whether the document had been trashed.

Because trashing a document does not change its workflow state or the Modify Portal Content permission, the finalize transition remained available for documents in the trash. The same issue affected the finalize-and-sign transition, which reuses the same guard. 

I also updated other transitions which are not checking for the trashed property, like reopen an already finalized document or sign a finalized document.

For [TI-4033]

## Checklist

- [x] Changelog entry

## Post-merge action

> [!IMPORTANT]

[TI-4033]: https://4teamwork.atlassian.net/browse/TI-4033?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


opengever/document/document.py
```diff
@@ -427,7 +428,8 @@ def leave_shadow_state(self):
                 self, transition='document-transition-initialize')
 
     def is_finalize_allowed(self):
-        return not self.is_checked_out() and \
+        return not ITrasher(self).is_trashed() and \
```
all4manu: Suggestion: documents already have an `is_trashed` property (`opengever/document/base.py`), which the workflow itself uses as guard for the remove transition (`here/is_trashed`). Using it here keeps things consistent and we can drop the new `ITrasher` import on line 31.

```suggestion
        return not self.is_trashed and \
```

opengever/document/document.py
```diff
@@ -427,7 +428,8 @@ def leave_shadow_state(self):
                 self, transition='document-transition-initialize')
 
     def is_finalize_allowed(self):
```
all4manu, line 429: Observation (not a blocker): documents that were already finalized while in the trash stay like that. The "remove" transition only starts from the draft state, so these documents can never be removed from the trash anymore. Might be worth a quick check on production (trashed + final) whether such documents exist and whether we need a small cleanup/upgrade.

opengever/document/tests/test_document_workflow.py
```diff
@@ -90,6 +90,17 @@ def test_document_cannot_be_finalized_if_checked_out(self):
         self.assertEquals(Document.active_state,
                           api.content.get_state(obj=self.document))
 
+    def test_document_cannot_be_finalized_if_trashed(self):
```
all4manu, line 93: Suggestion: the same guard is also used by the workspace document workflow (finalize in workspaces) and by the "start signing from draft" transition. A short test for the workspace case would make sure this stays covered there too.

elioschmutz: These transitions are not available in teamraum

opengever/document/document.py
elioschmutz: The transition to restore is available for trashed documents.

elioschmutz:
@all4manu I've updated the PR. I also checked other transitions and found some more to fix. So please take another look.

all4manu (approved):
LGTM 👍

